PR #16, opened 2022-03-11
Use k8s namespace from extension settings

eyalb181:


aviramha:
Great. Please update change log.

PR #31, opened 2022-05-05
Move some functionality to separate files

eyalb181:
Move all socket management to separate file

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -1,34 +1,15 @@
-#![feature(c_variadic)]
+// #![feature(c_variadic)]
 
-use std::{
-    collections::HashMap,
-    env,
-    io::{Read, Write},
-    net::{IpAddr, Ipv4Addr, SocketAddr},
-    os::unix::io::AsRawFd,
-    sync::Mutex,
-    thread,
-    time::Duration,
-};
+use std::{sync::Mutex, thread, time::Duration};
 
 use ctor::ctor;
-use frida_gum::{interceptor::Interceptor, Gum, Module, NativePointer};
+use frida_gum::{interceptor::Interceptor, Error, Gum, Module, NativePointer};
 use futures::{SinkExt, StreamExt};
-use k8s_openapi::api::core::v1::Pod;
-use kube::{
-    api::{Api, Portforwarder, PostParams},
-    runtime::wait::{await_condition, conditions::is_pod_running},
-    Client,
-};
+use kube::api::Portforwarder;
 use lazy_static::lazy_static;
 use libc::{c_char, c_void, sockaddr, socklen_t};
 use mirrord_protocol::{ClientCodec, ClientMessage, DaemonMessage};
-use multi_map::MultiMap;
 use os_socketaddr::OsSocketAddr;
-use queues::{IsQueue, Queue};
-use rand::distributions::{Alphanumeric, DistString};
-use serde_json::json;
-use socketpair::{socketpair_stream, SocketpairStream};
 use tokio::{
     runtime::Runtime,
     select,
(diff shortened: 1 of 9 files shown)


mirrord-layer/src/sockets.rs
```diff
@@ -0,0 +1,152 @@
...
+
+pub struct Sockets {
+    connections: Mutex<MultiMap<i32, u16, ConnectionSocket>>,
+    data: Mutex<HashMap<u16, DataSocket>>,
+    connection_queue: Mutex<Queue<u16>>,
+    pending_data: Mutex<HashMap<u16, Vec<u8>>>
+}
+
+impl Sockets {
+    pub fn new() -> Self {
```
meowjesty, line 36: Minor thing, if `new` takes no arguments, then it should probably be an implementation of `default` instead.

```suggestion
impl Default for Sockets {
    fn default() -> Self { ... }
}
```

meowjesty (commented):
👍

mirrord-layer/src/sockets.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub write_socket: SocketpairStream,
+}
+
+pub struct Sockets {
+    connections: Mutex<MultiMap<i32, u16, ConnectionSocket>>,
+    data: Mutex<HashMap<u16, DataSocket>>,
+    connection_queue: Mutex<Queue<u16>>,
+    pending_data: Mutex<HashMap<u16, Vec<u8>>>
+}
+
```
aviramha, line 34: perhaps add some docs here

aviramha (changes requested):

mirrord-layer/src/lib.rs
```diff
@@ -89,47 +58,16 @@ unsafe extern "C" fn bind_detour(sockfd: i32, addr: *const sockaddr, addrlen: so
...
-    let real_fd = match CONNECTION_SOCKETS.lock().unwrap().get(&sockfd) {
-        Some(s) => s.real_read_socket_fd,
-        None => {
-            println!("No socket found for fd: {}", sockfd);
-            return -1;
-        }
-    };
-
-    libc::connect(real_fd, addr, addrlen)
+    SOCKETS.add_port_to_connection_socket(sockfd, port);
```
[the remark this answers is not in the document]
eyalb181: I had some theoretical logic there (where I cached the real socket) but I'd rather run into real behaviour and then decide what to do.

mirrord-layer/src/lib.rs
```diff
@@ -14,72 +10,45 @@ use std::{
...
+    let (sender, receiver) = channel::<i32>(1000);
+    *NEW_CONNECTION_SENDER.lock().unwrap() = Some(sender);
+    enable_hooks();
+    RUNTIME.spawn(poll_agent(pf, receiver));
+}
 
-    read_fd
+unsafe extern "C" fn socket_detour(_domain: i32, _socket_type: i32, _protocol: i32) -> i32 {
+    debug!("socket called");
+    SOCKETS.create_connection_socket()
```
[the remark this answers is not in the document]
eyalb181: But it doesn't just create the socket, it also adds it to the connection sockets map, and uses it internally later

mirrord-layer/src/lib.rs
```diff
@@ -172,138 +110,22 @@ unsafe extern "C" fn accept_detour(
         let os_addr: OsSocketAddr = socket_addr.into();
```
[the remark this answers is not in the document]
eyalb181: Implement what?

mirrord-layer/src/pod_api.rs
```diff
@@ -0,0 +1,93 @@
...
+                        }
+                    ],
+                    "command": [
+                        "./mirrord-agent",
+                        "--container-id",
+                        container_id,
+                        "-t",
+                        "60"
+                    ],
+                    "env": [{"name": "RUST_LOG", "value": "trace"}],
```
aviramha, line 75: RUST_LOG should be env controlled

mirrord-layer/src/sockets.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(PartialEq)]
+pub enum ConnectionState {
+    Bound,
+    Listening,
+}
+
+pub struct DataSocket {
+    pub connection_id: u16,
+    #[allow(dead_code)]
+    pub read_socket: SocketpairStream, /* Though this is unread, it's necessary to keep the
```
aviramha, line 37: We actually should read it so buffers wont fill.

mirrord-layer/src/sockets.rs
```diff
@@ -0,0 +1,222 @@
...
+                                        * socket open */
+    pub write_socket: SocketpairStream,
+    pub address: SocketAddr,
+}
+
+pub struct Sockets {
+    new_sockets: Mutex<HashMap<i32, Socket>>,
+    connections: Mutex<MultiMap<i32, u16, ConnectionSocket>>, // Keys are sockfd, port
```
aviramha, line 45: You can define types so code will be more readable even without docs - 
`type SockFd = i32` `type Port = u16`

aviramha (changes requested):
Very good job, few last comments ;)

mirrord-layer/src/sockets.rs
```diff
@@ -0,0 +1,240 @@
...
+            // NewTcpConnection
+            debug!("No socket found for connection_id: {}", connection_id);
+            self.pending_data
+                .lock()
+                .unwrap()
+                .insert(connection_id, data);
+        }
+    }
+
+    fn clear_data(&self, socket: &mut SocketpairStream) {
```
aviramha, line 224: no use of self, make this a module level function

aviramha (approved):

PR #32, opened 2022-05-06
Hook libuv uv__accept4, accept4

infiniteregrets:
An issue was encountered on the following system:
```
mehula@mehul-machine:~/mirrord$ uname -a
Linux mehul-machine 5.13.0-1022-azure #26~20.04.1-Ubuntu SMP Thu Apr 7 19:42:45 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```
where `LD_PRELOAD=/home/mehula/mirrord/target/debug/libmirrord.so node sample/app.js` would fail with the following error:
```
2022-05-05T17:46:41.820500Z DEBUG mirrord: send message to client 80
2022-05-05T17:46:43.970924Z DEBUG mirrord: new connection id: 0
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: accept EINVAL
    at TCP.onconnection (net.js:1497:24)
Emitted 'error' event at:
    at TCP.onconnection (net.js:1497:10)
```
The linux manpage describes the accept4 extension https://man7.org/linux/man-pages/man2/accept.2.html, which could have been the probably cause, but apparently was not. Further digging by Aviram:
```
1. on Linux that has accept4, node is compiled to use accept4.
2. it doesn't call accept4 (from libc) but it calls uv__accept4
3. checking uv code, unlike the regular accept, instead of calling libc, it actually calls the sys call by itself!
```

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use ctor::ctor;
-use frida_gum::{interceptor::Interceptor, Gum, Module, NativePointer};
+use frida_gum::{interceptor::Interceptor, Error, Gum, Module, NativePointer};
 use futures::{SinkExt, StreamExt};
 use k8s_openapi::api::core::v1::Pod;
 use kube::{
@@ -210,6 +210,15 @@ unsafe extern "C" fn accept_detour(
     read_fd
 }
 
+unsafe extern "C" fn accept4_detour(
+    sockfd: i32,
+    addr: *mut sockaddr,
+    addrlen: *mut socklen_t,
+    _flags: i32,
+) -> i32 {
+    accept_detour(sockfd, addr, addrlen)
+}
+
 async fn create_agent() -> Portforwarder {
     // Create Agent
     let client = Client::try_default().await.unwrap();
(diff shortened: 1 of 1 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -425,11 +425,38 @@ fn enable_hooks() {
...
         .unwrap();
 
-    interceptor
-        .replace(
-            Module::find_export_by_name(None, "accept").unwrap(),
-            NativePointer(accept_detour as *mut c_void),
-            NativePointer(std::ptr::null_mut::<c_void>()),
-        )
-        .unwrap();
+    match Module::find_export_by_name(None, "uv__accept4") {
```
aviramha, line 428: we don't want to hook the other functions if this is none, rather hook each one of those if not hooked already.

aviramha (changes requested):

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +425,40 @@ fn enable_hooks() {
...
```
aviramha, line 428: change to if let some, and maybe create a hashset of already hooked addresses so we won't hook twice?
also, just to be on the safe side, create a detour for accept4 because it accepts 1 more argument.

aviramha (changes requested):

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
+                    error!("accept already replaced");
+                }
+                _ => {
+                    error!("accept error: {:?}", e);
+                }
+            },
+            Ok(_) => {
+                debug!("accept hooked");
+            }
+        }
```
meowjesty, line 454: We're not doing proper error handling yet, so you can drop the `match` in favor of `unwrap`.

```rust
interceptor.replace(...).unwrap();
```

If you're doing this to help you debug, `Result` has a method (on nightly) to inspect errors: [inspect_err](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err). Just need to enable the feature `#![feature(result_option_inspect)]`.

meowjesty (commented):
Will be more thorough after I get it to run on the vm.

mirrord-layer/src/lib.rs
aviramha: That's actually my suggestion because on some systems accept4 can be same as accept and uv_accept4 can lead to accept4 directly, so we might encounter a situation where `Error::InterceptorAlreadyReplaced` is an okay situation.

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
+        match interceptor.replace(
+            addr,
+            NativePointer(accept4_detour as *mut c_void),
             NativePointer(std::ptr::null_mut::<c_void>()),
-        )
-        .unwrap();
+        ) {
+            Err(e) => match e {
+                Error::InterceptorAlreadyReplaced => {
+                    error!("accept already replaced");
```
aviramha, line 445: it's not an error, use debug

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
         .unwrap();
 
-    interceptor
-        .replace(
-            Module::find_export_by_name(None, "accept").unwrap(),
-            NativePointer(accept_detour as *mut c_void),
+    if let Some(addr) = Module::find_export_by_name(None, "accept") {
+        match interceptor.replace(
+            addr,
+            NativePointer(accept4_detour as *mut c_void),
```
aviramha, line 440: this should be `accept_detour`?

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
+
+        if let Some(addr) = Module::find_export_by_name(None, "uv__accept") {
+            match interceptor.replace(
+                addr,
+                NativePointer(accept_detour as *mut c_void),
+                NativePointer(std::ptr::null_mut::<c_void>()),
+            ) {
+                Err(e) => match e {
+                    Error::InterceptorAlreadyReplaced => {
+                        error!("uv__acept already replaced");
```
aviramha, line 464: typo + debug

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
+                    }
+                    _ => {
+                        error!("uv__acept error: {:?}", e);
```
aviramha, line 467: typo

mirrord-layer/src/lib.rs
```diff
@@ -425,11 +434,62 @@ fn enable_hooks() {
...
+                _ => {
+                    error!("accept error: {:?}", e);
+                }
+            },
+            Ok(_) => {
+                debug!("accept hooked");
+            }
+        }
+
+        if let Some(addr) = Module::find_export_by_name(None, "uv__accept") {
```
aviramha, line 456: you need to hook `uv__accept4` no?

aviramha (changes requested):

aviramha (approved):
Awesome work @infiniteregrets!

PR #34, opened 2022-05-08
Implement Mirrord CLI

aviramha:
Should merge only after https://github.com/metalbear-co/mirrord/pull/33

### DIFF
--- a/mirrord-cli/README.md
+++ b/mirrord-cli/README.md
@@ -1,0 +1,36 @@
+# mirrord-cli
+mirrord-cli is the actual binary that users use to run mirrord. The client helps you execute processes with mirrord injected to it.
+Right now injection is done using `LD_PRELOAD` on Linux and `DYLD_INSERT_LIBRARIES` on macOS.
+
+## Usage
+`mirrord exec --pod-name <POD_NAME> <BINARY> [BINARY_ARGS..]`
+
+## Compatibility
+### Linux
+`LD_PRELOAD` works only on dynamically linked binaries, so static binaries won't work. Most development use cases are dynamic binaries, so it should be okay.
+We do have plan to support static binaries, so let us know if you encountered a use case that isn't covered.
+
+### macOS
+`DYLD_INSERT_LIBRARIES` works on "unprotected" binaries, which are most development binaries anyway. To be more specific, the following use cases won't work:
+
+* setuid and/or setgid bits are set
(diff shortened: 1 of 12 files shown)


mirrord-cli/README.md
```diff
@@ -0,0 +1,34 @@
+# mirrord-cli
+mirrord-cli is the actual binary that users use to run mirrord. The client helps you execute processes with mirrord injected to it.
+Right now injection is done using `LD_PRELOAD` on Linux and `DYLD_INSERT_LIBRARIES` on macOS.
+
+
+## Compatability
```
eyalb181, line 6: Compatibility

mirrord-cli/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Linux
+`LD_PRELOAD` works only on dynamically linked binaries, so static binaries won't work. Most development use cases are dynamic binaries, so it should be ok.
```
eyalb181, line 8: okay*

mirrord-cli/README.md
```diff
@@ -0,0 +1,34 @@
...
+We do have plan to support static binaries, so let us know if you encountered a use case that isn't covered.
+
+### macOS
+`DYLD_INSERT_LIBRARIES` works on "unprotected" binaries, which are most development binaries anyway. To be more specific, such use cases won't work:
```
eyalb181, line 12:
```suggestion
`DYLD_INSERT_LIBRARIES` works on "unprotected" binaries, which are most development binaries anyway. To be more specific, the following use cases won't work:
```

mirrord-cli/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* setuid and/or setgid bits are set
+* restricted by entitlements
+* restricted segment
+
+[Source](https://theevilbit.github.io/posts/dyld_insert_libraries_dylib_injection_in_macos_osx_deep_dive/)
+
+Please let us know if you encountered a use case where it doesn't work for you, regardless if documented that it isn't supported, so we know there's demand to implement that use case.
```
eyalb181, line 20:
```suggestion
Please let us know if you encountered a use case where it doesn't work for you, regardless of whether it's documented that it isn't supported, so we know there's demand to implement that use case.
```

eyalb181 (changes requested):

PR #36, opened 2022-05-09
Namespace

aviramha:
Add ability to specify different namespace for agent & impersonated pod


mirrord-layer/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 use std::{sync::Mutex, thread, time::Duration};
 
 use ctor::ctor;
+use envconfig::Envconfig;
```
eyalb181, line 6: Why is this needed?

eyalb181 (approved):

PR #37, opened 2022-05-09
Ignore unix sockets in hooks

eyalb181:
Interfering with unix sockets broke our behavior when running in VSCode's debugger. This PR ignores unix sockets in hooks and instead maintains their default behavior.

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -7,7 +7,7 @@ use frida_gum::{interceptor::Interceptor, Error, Gum, Module, NativePointer};
 use futures::{SinkExt, StreamExt};
 use kube::api::Portforwarder;
 use lazy_static::lazy_static;
-use libc::{c_void, sockaddr, socklen_t};
+use libc::{c_int, c_void, fd_set, sockaddr, socklen_t, timeval};
 use mirrord_protocol::{ClientCodec, ClientMessage, DaemonMessage};
 use os_socketaddr::OsSocketAddr;
 use tokio::{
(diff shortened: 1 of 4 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -25,6 +25,7 @@ lazy_static! {
     static ref RUNTIME: Runtime = Runtime::new().unwrap();
     static ref SOCKETS: sockets::Sockets = sockets::Sockets::default();
     static ref NEW_CONNECTION_SENDER: Mutex<Option<Sender<i32>>> = Mutex::new(None);
+    static ref UNIX_FDS: Mutex<HashSet<i32>> = Mutex::new(HashSet::new());
```
aviramha, line 28: Future thinking, I think this should be `ignored_fds` as we might have other reasons to ignore fds.

mirrord-layer/src/lib.rs
```diff
@@ -41,13 +42,21 @@ fn init() {
     RUNTIME.spawn(poll_agent(pf, receiver));
 }
 
-unsafe extern "C" fn socket_detour(_domain: i32, _socket_type: i32, _protocol: i32) -> i32 {
+unsafe extern "C" fn socket_detour(domain: i32, socket_type: i32, protocol: i32) -> i32 {
     debug!("socket called");
+    if domain == 2 { // Unix socket
```
aviramha, line 47: there's a libc const probably?

mirrord-layer/src/lib.rs
aviramha, line 28: use the sockfd type? (if you did it, if not maybe create it now?)

aviramha (changes requested):

PR #43, opened 2022-05-09
Use native certificates on TLS for rustls

infiniteregrets:
Native certificates are installed using the following package: https://docs.rs/rustls-native-certs/latest/rustls_native_certs/index.html for a custom client. 
Reference: https://docs.rs/kube-client/0.71.0/src/kube_client/client/tls.rs.html#129


aviramha:
superseded by #47

PR #44, opened 2022-05-10
Refactor sockets

### LINKED ISSUE #41: mirrord-layer deadlocks if accept is called when there's no data
### Bug Description

When accept is called before a connection is read from the agent, it locks the connections struct and there's a deadlock.

### Steps to Reproduce

Simulate a delay so that accept is called before a connection is read from the agent.

### Backtrace

_No response_

MacOS

### Your Node version (`node --version`)

N/A

### Additional Info

_No response_

### LINKED ISSUE #42: When running with VSCode's Node debugger, mirrord hooks and breaks an internal socket
### Bug Description

VSCode's Node debugger uses a socket for some internal mechanism. Mirrord hooks it, breaking the flow. We should be able to overlook these sockets.

### Steps to Reproduce

Debug a node process in VSCode with mirrord-layer in LD_PRELOAD.

### Backtrace

_No response_

MacOS

### Your Node version (`node --version`)

v17.6.0

### Additional Info

_No response_

aviramha:
Refactor sockets - beforehand we created socketpair on each socket creation, now we only do it when necessary using delayed execution of bind,listen.

closes #41 closes #42

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -1,33 +1,93 @@
 // #![feature(c_variadic)]
 
-use std::{sync::Mutex, thread, time::Duration};
+use std::{
+    collections::HashMap,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    os::unix::io::RawFd,
+    thread,
+    time::Duration,
+};
 
 use ctor::ctor;
 use envconfig::Envconfig;
-use frida_gum::{interceptor::Interceptor, Error, Gum, Module, NativePointer};
+use frida_gum::{interceptor::Interceptor, Gum};
 use futures::{SinkExt, StreamExt};
 use kube::api::Portforwarder;
 use lazy_static::lazy_static;
 use mirrord_protocol::{ClientCodec, ClientMessage, DaemonMessage};
 use tokio::{
+    io::AsyncWriteExt,
+    net::TcpStream,
     runtime::Runtime,
     select,
     sync::mpsc::{channel, Receiver, Sender},
+    task,
 };
 use tracing::{debug, error};
 
+mod common;
 mod config;
 mod macros;
 mod pod_api;
 mod sockets;
-use config::Config;
 use tracing_subscriber::prelude::*;
+
+use crate::{
+    common::{HookMessage, Port},
+    config::Config,
+    sockets::{SocketInformation, CONNECTION_QUEUE},
+};
 
 lazy_static! {
     static ref GUM: Gum = unsafe { Gum::obtain() };
     static ref RUNTIME: Runtime = Runtime::new().unwrap();
-    static ref SOCKETS: sockets::Sockets = sockets::Sockets::default();
-    static ref NEW_CONNECTION_SENDER: Mutex<Option<Sender<i32>>> = Mutex::new(None);
+}
+
+pub static mut HOOK_SENDER: Option<Sender<HookMessage>> = None;
+
+#[derive(Debug)]
+enum TcpTunnelMessages {
+    Data(Vec<u8>),
+    Close,
+}
+
+#[derive(Debug, Clone)]
+struct ListenData {
+    ipv6: bool,
+    port: Port,
+    fd: RawFd,
+}
+
+async fn tcp_tunnel(mut local_stream: TcpStream, mut receiver: Receiver<TcpTunnelMessages>) {
+    loop {
+        select! {
+            message = receiver.recv() => {
+                match message {
+                    Some(TcpTunnelMessages::Data(data)) => {
+                        local_stream.write_all(&data).await.unwrap()
+                    },
+                    Some(TcpTunnelMessages::Close) => break,
+                    None => break
+                };
+            },
+            _ = local_stream.readable() => {
+                let mut data = vec![0; 1024];
+                match local_stream.try_read(&mut data) {
+                    Err(ref err) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                        continue
+                        },
+                    Err(err) => {
+                        debug!("local stream ended with err {:?}", err);
+                        break;
+                    }
+                    Ok(n) if n == 0 => break,
+                    Ok(_) => {}
+                }
+
+            }
+        }
+    }
+    debug!("exiting tcp tunnel");
 }
 
 #[ctor]
@@ -46,23 +106,27 @@ fn init() {
         config.agent_rust_log,
         config.agent_image,
     ));
-    let (sender, receiver) = channel::<i32>(1000);
-    *NEW_CONNECTION_SENDER.lock().unwrap() = Some(sender);
+    let (sender, receiver) = channel::<HookMessage>(1000);
+    unsafe {
+        HOOK_SENDER = Some(sender);
+    };
     enable_hooks();
     RUNTIME.spawn(poll_agent(pf, receiver));
 }
 
-async fn poll_agent(mut pf: Portforwarder, mut receiver: Receiver<i32>) {
+async fn poll_agent(mut pf: Portforwarder, mut receiver: Receiver<HookMessage>) {
     let port = pf.take_stream(61337).unwrap(); // TODO: Make port configurable
     let mut codec = actix_codec::Framed::new(port, ClientCodec::new());
+    let mut port_mapping: HashMap<Port, ListenData> = HashMap::new();
+    let mut active_connections = HashMap::new();
     loop {
         select! {
             message = receiver.recv() => {
                 match message {
-                    Some(sockfd) => {
-                        let port = SOCKETS.get_connection_socket_address(sockfd).unwrap().port();
-                        debug!("send message to client {:?}", port);
-                        codec.send(ClientMessage::PortSubscribe(vec![port])).await.unwrap();
+                    Some(HookMessage::Listen(msg)) => {
+                        debug!("received message from hook {:?}", msg);
+                        codec.send(ClientMessage::PortSubscribe(vec![msg.real_port])).await.unwrap();
+                        port_mapping.insert(msg.real_port, ListenData{port: msg.fake_port, ipv6: msg.ipv6, fd: msg.fd});
                     }
                     None => {
                         debug!("NONE in recv");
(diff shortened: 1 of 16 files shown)


mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
 use queues::{IsQueue, Queue};
 use socketpair::{socketpair_stream, SocketpairStream};
 use tracing::{debug, error};
 
-pub struct Socket {
-    pub read_fd: SockFd,
-    pub read_socket: SocketpairStream,
-    pub write_socket: SocketpairStream,
-}
+use crate::{
```
eyalb181, line 19: Why do you need use crate here, can't you just do 'mod'?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
-                    write_socket: socket.write_socket,
-                    address,
-                    state: ConnectionState::Bound,
-                },
-            );
-        } else {
-            error!("No socket found for fd: {}", sockfd);
-        }
-    }
+impl Eq for Socket {}
```
eyalb181, line 60: Why is this necessary?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
+#[derive(Default)]
+struct State {
+    sockets: HashSet<Socket>,
+}
 
-    pub fn get_data_socket_address(&self, sockfd: SockFd) -> Option<SocketAddr> {
-        let sockets = self.data.lock().unwrap();
-        sockets.get(&sockfd).map(|socket| socket.address)
+lazy_static! {
+    static ref STATE: Mutex<State> = Mutex::new(State::default());
```
eyalb181, line 80: Why not just keep a direct reference to the sockets?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
+}
+
+#[inline]
```
eyalb181, line 83: Why'd you decide to inline it?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
+fn is_ignored_port(port: Port) -> bool {
+    port == 0 || (port > 50000 && port < 60000)
+}
+
+fn socket(domain: c_int, type_: c_int, protocol: c_int) -> RawFd {
```
eyalb181, line 88: Why the underscore suffix for type_?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,368 @@
...
+            }
+            Some(socket) => socket,
+            None => {
+                debug!("bind: no socket found for fd: {}", &sockfd);
+                return unsafe { libc::bind(sockfd, addr, addrlen) };
+            }
+        }
+    };
+
+    let raw_addr = unsafe { OsSocketAddr::from_raw_parts(addr as *const u8, addrlen as usize) };
```
eyalb181, line 133: Why not move this to the beginning of the function? Then you could edit the socket as you find it with get_mut and not have to lock the State twice

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,361 @@
...
+lazy_static! {
+    static ref STATE: Mutex<State> = Mutex::new(State::default());
+}
+
+#[inline]
+fn is_ignored_port(port: Port) -> bool {
+    port == 0 || (port > 50000 && port < 60000)
+}
+
+fn socket(domain: c_int, type_: c_int, protocol: c_int) -> RawFd {
```
eyalb181, line 88: I'm going to add my interpretation of what the hooks do, let me know if I didn't get it right, and if I did please add it as function documentation.
For socket:
Create the socket, add it to State, and return it to the calling application.

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,361 @@
...
-            let mut buffer = [0; 1];
-            socket.read_socket.read_exact(&mut buffer).unwrap();
-            sockets.insert(sockfd, socket.address.port(), socket);
-            let mut queues = self.connection_queues.lock().unwrap();
-            let queue = queues.get_mut(&sockfd).unwrap();
-            queue.remove().unwrap()
-        } else {
-            error!("No socket found for fd: {}", sockfd);
-            0
+fn bind(sockfd: c_int, addr: *const sockaddr, addrlen: socklen_t) -> c_int {
```
eyalb181, line 116: If it's an unignored port, map the address the application bound to the socket.

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,361 @@
...
+            let ret = unsafe { libc::bind(sockfd, os_addr.as_ptr(), os_addr.len()) };
+            if ret != 0 {
+                error!(
+                    "listen: failed to bind socket ret: {:?}, addr: {:?}, sockfd: {:?}, errno: {:?}",
+                    ret, os_addr, sockfd, errno()
+                );
+                return ret;
+            }
+            let mut addr_len = os_addr.len();
+            let ret = unsafe { libc::getsockname(sockfd, os_addr.as_mut_ptr(), &mut addr_len) };
```
eyalb181, line 203: Why is this necessary?

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,373 @@
...
+                    OsSocketAddr::from(SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0))
+                }
+                _ => {
+                    // shouldn't happen
+                    debug!("unsupported domain");
+                    return libc::EINVAL;
+                }
+            };
+
+            let ret = unsafe { libc::bind(sockfd, os_addr.as_ptr(), os_addr.len()) };
```
eyalb181, line 194: So does this call the hook, not find the socket, then call the actual function?

mirrord-layer/src/lib.rs
```diff
@@ -1,151 +1,121 @@
...
-unsafe extern "C" fn accept4_detour(
-    sockfd: i32,
-    addr: *mut sockaddr,
-    addrlen: *mut socklen_t,
-    _flags: i32,
-) -> i32 {
-    accept_detour(sockfd, addr, addrlen)
+#[ctor]
+fn init() {
+    INIT.call_once(|| {
```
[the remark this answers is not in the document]
aviramha: erm, guess not

mirrord-layer/src/sockets.rs
aviramha: that's how you impl Eq which is required for HashSet.

mirrord-layer/src/sockets.rs
aviramha: I'm preparing for a more elaborate usecase, which might not exist! :)

aviramha: (fixing)

mirrord-layer/src/sockets.rs
aviramha: because the over head of calling the function is stupid while it's just a simple condition wrapped in a function for beauty reasons

mirrord-layer/src/sockets.rs
aviramha: Because we want to remove it anway (if ignore we want to remove it,?

mirrord-layer/src/sockets.rs
aviramha: If you look at above lines, we provide port:0, this sets an unspecified port and this is how we retrieve what port it is.

mirrord-layer/src/sockets.rs
aviramha: yup

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,506 @@
...
+
+unsafe extern "C" fn bind_detour(
+    sockfd: c_int,
+    addr: *const sockaddr,
+    addrlen: socklen_t,
+) -> c_int {
+    bind(sockfd, addr, addrlen)
+}
+
+fn listen(sockfd: RawFd, _backlog: c_int) -> c_int {
```
eyalb181, line 204: Bind the socket to a fake, local port, and subscribe to the agent on the real port. Messages received from the agent on the real port will later be routed to the fake local port.

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,506 @@
...
+            let ret = unsafe { libc::bind(sockfd, os_addr.as_ptr(), os_addr.len()) };
+            if ret != 0 {
+                error!(
+                    "listen: failed to bind socket ret: {:?}, addr: {:?}, sockfd: {:?}, errno: {:?}",
+                    ret, os_addr, sockfd, errno()
+                );
+                return ret;
+            }
+            let mut addr_len = os_addr.len();
+            let ret = unsafe { libc::getsockname(sockfd, os_addr.as_mut_ptr(), &mut addr_len) };
```
eyalb181, line 243: // Resolve the fake local port that the socket was bound to

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,506 @@
...
-            address,
-        };
-        self.data
-            .lock()
-            .unwrap()
-            .insert(read_fd, connection_id, data_socket);
-        read_fd
+    };
+
+    // We don't handle this socket, so restore state if there was any. (delay execute bind)
```
eyalb181, line 311: What do you mean here with "we don't handle this socket"? Because it's for outgoing communication?

aviramha: yes

mirrord-layer/src/sockets.rs
```diff
@@ -1,240 +1,506 @@
...
+    listen(sockfd, backlog)
+}
+
+fn connect(sockfd: RawFd, address: *const sockaddr, len: socklen_t) -> c_int {
+    debug!("connect called");
 
-        if let Some(data) = self.pending_data.lock().unwrap().remove(&connection_id) {
-            debug!("writing pending data for connection_id: {}", connection_id);
-            write_socket.write_all(&data).unwrap();
+    let socket = {
```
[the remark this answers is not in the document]
eyalb181: ?

mirrord-layer/src/sockets.rs
aviramha: I added it in the function doc?

eyalb181 (approved):

PR #47, opened 2022-05-11
Accept invalid certificates optionally

infiniteregrets:
Following changes have been made:
- Currently certificates can't be verified in the current version of rustls for IP addresses and this results in an `invalid dnsname` error. The latest commits on rustls suppress certificate verification https://github.com/rustls/rustls/pull/1032 and we use this by patching the dependencies. 
- Added a new CLI argument which accepts boolean value to accept/reject invalid certificates. This argument is directly related to the new `MIRRORD_ACCEPT_INVALID_CERTIFICATES` environment variable (default false)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,11 @@ members = [
     "mirrord-layer",
     "mirrord-cli",
 ]
+
+# latest commits on rustls suppress certificate verification 
+# https://github.com/rustls/rustls/pull/1032
+# so we patch crates.io to use the latest commits from rustls
+# this should be changed once a newer version of rustls is out
+
+[patch.crates-io]
+rustls = { git = 'https://github.com/rustls/rustls' }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -5,4 +5,7 @@ members = [
     "mirrord-agent",
     "mirrord-layer",
     "mirrord-cli",
-]
\ No newline at end of file
+]
+
+[patch.crates-io]
```
aviramha, line 10: Comment explaining why we do this (can link to this PR)

aviramha (commented):
Great job. I have left one small comment. Address it and let's merge it.

mirrord-layer/src/pod_api.rs
```diff
@@ -41,7 +44,15 @@ pub async fn create_agent(
     log_level: String,
     agent_image: String,
 ) -> Portforwarder {
-    let client = Client::try_default().await.unwrap();
+    let env_config = config::Config::init_from_env().unwrap();
+    let client = if env_config.accept_invalid_certificates {
+        let mut config = Config::infer().await.unwrap();
+        config.accept_invalid_certs = true;
+        debug!("Accepting invalid certificates");
```
aviramha, line 51: warning

aviramha (approved):
Great job, thanks!

PR #49, opened 2022-05-13
make accept_invalid_certificates a flag instead of an argument

aviramha:
Interface to use `accept_invalid_certificates` was to supply `true`/`false` value, while it makes more sense to have it as a flag.


infiniteregrets (approved):
Thank you!

PR #52, opened 2022-05-16
validate log levels

infiniteregrets:
Currently mirrord can accept any kind of arbitrary string for agent logs (-l/--agent-log-level <AGENT_LOG_LEVEL> ), which is not desirable as a simple typo in the log level can make it hard to determine why someone doesn't see the logs. Desirable behavior would be to error out on invalid log levels.
Valid log levels: trace, debug, info, warn, error (https://docs.rs/log/0.4.17/log/enum.Level.html)

### DIFF
--- a/mirrord-cli/src/main.rs
+++ b/mirrord-cli/src/main.rs
@@ -108,6 +108,12 @@ fn exec(args: &ExecArgs) -> Result<()> {
         std::env::set_var("MIRRORD_AGENT_NAMESPACE", namespace.clone());
     }
     if let Some(log_level) = &args.agent_log_level {
+        let levels = vec!["trace", "debug", "info", "warn", "error"];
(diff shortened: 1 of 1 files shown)


mirrord-cli/src/main.rs
```diff
@@ -108,6 +108,12 @@ fn exec(args: &ExecArgs) -> Result<()> {
         std::env::set_var("MIRRORD_AGENT_NAMESPACE", namespace.clone());
     }
     if let Some(log_level) = &args.agent_log_level {
+        let levels = vec!["trace", "debug", "info", "warn", "error"];
```
aviramha, line 111: I'm not sure if we should restrict that, as the rust env logger has very complex configurations you can use.. I think leaving it free text is better

aviramha (changes requested):

Issue #50, opened 2022-05-15
v2.0 Test Plan


eyalb181:
## 1. Core (Layer + Agent)

1. Happy Flow (Run for Node.js, FastAPI)
    1. Run a process that prints data received with the mirrord env variables
    2. Send three different requests to the impersonated pod
    3. For each request, assert that the correct data was printed
    4. Stop the process
    5. After 30 seconds, assert that the job and agent were both deleted from the cluster
2. Wrong Pod
    1. 1.1.a
    2. Send a request to a different pod in the cluster
    3. Assert that the request wasn’t printed
3. Agent Namespace
    1. Happy Flow
        1. Create a new k8s namespace (call it ‘test-namespace’)
        2. Run a process that prints data received with the mirrord env variables and MIRRORD_AGENT_NAMESPACE=test-namespace
        3. Assert that the job and the agent were created in the cluster under ‘test-namespace’
        4. 1.b-1.c
    2. Nonexistent namespace
        1. Run a process that prints data received with the mirrod env variables and MIRRORD_AGENT_NAMESPACE=<some nonexistent namespace>
        2. Assert that the process crashes with an indicative error
        3. Assert that a new job and agent weren’t created
4. Pod Namespace
    1. Happy Flow
        1. Create a new k8s namespace (call it ‘test-namespace’)
        2. Create a new nginx pod under test-namespace
        3. Run a process that prints data received with the mirrord env variables, IMPERSONATED_POD_NAME set to the name of the nginx pod, and IMPERSONATED_POD_NAMESPACE set to the name of the nginx namespace
        4. 1.1.b-1.1.c
    2. Nonexistent namespace
        1. Run a process that prints data received with the mirrord env variables and IMPERSONATED_POD_NAMESPACE set to a nonexistent namespace
        2. Assert that the process crashed with an indicative error
        3. Assert that a new job and agent weren’t created

## 2. CLI

1. Happy Flow (run for Node.js, FastAPI)
    1. Using the CLI, run a process that prints data received
    2. 1.1.b-1.1.e

## 3. VSCode Extension

1. Happy Flow (Run for Node.js, FastAPI. Test at least one on WSL)
    1. Click ‘enable mirrord’
    2. Set a breakpoint in the debugged process
    3. Start debugging the process
    4. Send a request to the impersonated pod
    5. Assert the breakpoint is hit with the correct data
2. Repeat use
    1. 3.1.a-3.1.e
    2. Stop debugging the process
[...]

PR #54, opened 2022-05-16
Add E2E tests

### LINKED ISSUE #35: Add E2E tests
Add a suite of E2E tests to test the integration between the layer and the agent.

infiniteregrets:
This PR closes #35 adding tests conforming to the suite provided in #50 

Current state of the tests:
validation tests:
- run the node application with mirrord exec in a forked process
- send GET, POST, PUT, and DELETE requests and keep a counter
- POST validates data.txt, PUT creates a file with the data received, DELETE reads the file created by PUT, validates the data and removes the file. 
- if all operations succeed, we exit with EXIT_SUCCESS
- since only a single job and pod is created, we verify that these resources exist and are removed after agent timeout finishes. 

namespace tests [agent]:
- create a new namespace 
- run a binary with agent namespace assigned the new namespace
- check if the job and the pod exist in the created namespace within the timeout limit 
- run a binary with agent namespace assigned a non-existent namespace 
- make sure the process crashes
- verify no resources have been created in either the default or the bad namespace 

namespace tests [pod]:
- create a new namespace 
- run a binary with pod namespace assigned the new namespace in the form of an env var
- verify the job exists


aviramha:
Great job! Now get it passing ;)

eyalb181 (changes requested):
So we're going to need a pretty major rewrite here. Here are my main critiques:

- We want the test code to be easy to understand, so bash isn't a great choice. Lets go with Rust, since we're already using it and it's best not to introduce new languages without sufficient reason.
- Lets use a testing framework, not just write freeform code. It'll give our tests structure, would print out legible output, let us e.g. skip certain tests in the future, etc.
- Strive to minimize tests to check one thing. Instead of checking all four methods in one test, write a test for each. We want our test results to focus us as much as possible on where the problem is.
- Strive to have as much of the testing logic in the test code itself, rather than in the app being tested.

e2e/src/tests.rs
```diff
@@ -0,0 +1,295 @@
...
+fn curl_request(url: &str, method: &str) {
+    // we can use the reqwest/hyper library here
+    Command::new("curl")
+        .args(["-X", method, "--data", TEXT, url])
+        .output()
+        .unwrap();
+}
+
+#[allow(dead_code)]
+fn validate_data(read: String, service_url: &str, server: &mut Child) {
```
aviramha, line 105: I don't like this function.. It's flow is very odd and weird. I suggest it'll get the stdout reader, server and service url
and then it would do something like
```rust
request(service_url, "GET")
assert_eq!(stdout.read.., "GET: Request completed"
```

eyalb181 (changes requested):

infiniteregrets:
Todo:
- remove CURL requests, switch to reqwest 
- change CLI commands to kube-rs fn calls
- change how validate_data function works
- add more assertions/panics

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+                    let mut stderr_reader = BufReader::new(server.stderr.take().unwrap());
+                    let mut err_stream = String::new();
+                    stderr_reader.read_to_string(&mut err_stream).unwrap();
+                    panic!("{}", err_stream);
+                } else {
+                    panic!("server exited with status code: {}", status);
+                }
+            }
+            Ok(None) => {} // server is running
+            Err(e) => panic!("{:?}", e),
```
aviramha, line 42: I'm not sure why you check that server is running here, we shouldn't care AFAIK

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+                    Some(pod.metadata.name.clone().unwrap())
+                } else {
+                    None
+                }
+            })
+            .next()
+            .unwrap();
+        assert!(pod_name.contains(&job_name));
+        delete_namespace(&client, namespace).await;
+        sleep(Duration::from_secs(10)).await; // Todo: check if we still need this sleep
```
aviramha, line 183: Why would we need a sleep at the end of the test?

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+        let mut config = Config::infer().await.unwrap();
+        config.accept_invalid_certs = true;
+        let client = Client::try_from(config).unwrap();
+
+        let nonexistent_namespace = "nonexistent-namespace";
+        let pod_name = get_nginx_pod_name(&client, "default").await.unwrap();
+        let env = HashMap::from([("MIRRORD_AGENT_NAMESPACE", nonexistent_namespace)]);
+        let mut server = start_server(&pod_name, vec!["node", "node-e2e/app.js"], env);
+        sleep(Duration::from_secs(5)).await;
+        match server.try_wait() {
```
aviramha, line 199: just call wait here with timeout instead of sleeping

e2e/src/utils.rs
```diff
@@ -0,0 +1,227 @@
...
+pub async fn delete_namespace(client: &Client, namespace: &str) {
+    let namespace_api: Api<Namespace> = Api::all(client.clone());
+    namespace_api
+        .delete(&namespace, &DeleteParams::default())
+        .await
+        .unwrap();
+}
+
+pub async fn reqwest_request(url: &str, method: &[u8]) {
+    let client = reqwest::Client::new();
```
aviramha, line 101: why not just pass the method as enum instead of bytes?

e2e/src/tests.rs
infiniteregrets: for example if the server errors out for some reason on start, it will write to it's stderr and on line 44 we are reading the stdout and it will block forever, we dont exactly know what could happen so we need to check if all went i.e. server is running or it errored out. its just a guess that the server will be up and running in 5 secs (worst case)

e2e/src/tests.rs
infiniteregrets: some of the tests are using the same namespace, if I dont sleep then if one test deleted the namespace and the other wants to create it immediately and somehow it wasnt deleted then we get an error from the server. however in this case its a unique namespace so we can probably remove it. ill check and update

e2e/src/tests.rs
infiniteregrets: there is no method for a timeout in the impl for the Child process struct?

e2e/src/tests.rs
infiniteregrets: fixed, we don't need the sleep there

e2e/src/utils.rs
infiniteregrets: done

e2e/src/tests.rs
aviramha: https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

e2e/src/tests.rs
aviramha: you should read the stdout when the process finishes, there's no value in reading it in real time.
flow should be:
1. start process
2. send requests, do stuff
3. kill/verify process is done
4. validate stderr/stdout of process

eyalb181 (changes requested):

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+use crate::utils::*;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[tokio::test]
+    // starts the node(express.js) api server, sends four different requests, validates data,
+    // stops the server and validates if the agent job and pod are deleted
+    async fn test_complete_node_api() {
+        let mut config = Config::infer().await.unwrap();
```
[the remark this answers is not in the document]
infiniteregrets: thanks! i completely ignored that. fixed

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+                if pod.metadata.name.clone().unwrap().contains(&job_name) {
+                    Some(pod.metadata.name.clone().unwrap())
+                } else {
+                    None
+                }
+            })
+            .next()
+            .unwrap();
+        assert!(pod_name.contains(&job_name));
+        sleep(Duration::from_secs(50)).await;
```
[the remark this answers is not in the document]
infiniteregrets: takes around ~35 to cleanup the job & pod, i will change the sleep. if you think it should be different then please let me know!

e2e/src/tests.rs
```diff
@@ -0,0 +1,288 @@
...
+    // stops the server and validates if the agent job and pod are deleted
+    async fn test_complete_node_api() {
+        let mut config = Config::infer().await.unwrap();
+        config.accept_invalid_certs = true;
+        let client = Client::try_from(config).unwrap();
+
+        let service_url = get_service_url(&client, "default").await.unwrap();
+        let pod_name = get_nginx_pod_name(&client, "default").await.unwrap();
+        let command = vec!["node", "node-e2e/app.js"];
+        let env = HashMap::from([("MIRRORD_AGENT_TTL", "0")]); // for adding more environment variables
```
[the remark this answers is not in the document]
infiniteregrets: yes, changed it to `HashMap::new()`

e2e/scripts/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -ex
+
+sudo apt-get update -y
+sudo apt-get install -y nodejs npm \
```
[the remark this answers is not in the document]
infiniteregrets: thanks for pointing out, fixed now

.github/workflows/ci.yaml
```diff
@@ -3,67 +3,90 @@ name: CI
...
+      - name: start minikube
+        uses: medyagh/setup-minikube@master
+        with:
+          container-runtime: containerd
+      - name: setup nginx
+        run: kubectl apply -f e2e/app.yaml
+      - name: install dependencies
+        run: sudo bash e2e/scripts/install.sh
+      - name: add rustfmt
+        run: rustup component add rustfmt --toolchain nightly-x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
infiniteregrets: didn't realize we could add this manually through the action, fixed now

e2e/src/utils.rs
```diff
@@ -0,0 +1,233 @@
...
+};
+use reqwest::{Method, StatusCode};
+use serde_json::json;
+use tokio::time::{sleep, Duration};
+
+static TEXT: &str = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+
+// target/debug/mirrord exec --pod-name pod_name  -c binary command
+pub fn start_server(pod_name: &str, command: Vec<&str>, env: HashMap<&str, &str>) -> Child {
+    let path = format!(
```
aviramha, line 28: relative path, use https://docs.rs/bintest/latest/bintest/

e2e/src/tests.rs
```diff
@@ -0,0 +1,276 @@
+use std::{
```
aviramha, line 1: move tests to tests dir

aviramha: and rename file to something like sanity

.github/workflows/ci.yaml
```diff
@@ -50,21 +50,46 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+      # For now, just evrify it compiles.
+      - run: cargo +nightly build --manifest-path=mirrord-layer/Cargo.toml
+
+  e2e:
+    runs-on: ubuntu-latest
```
eyalb181, line 72: Can we try and do a platform matrix here? Lets start with mac and ubuntu.

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+            .filter_map(|pod| {
+                if pod.metadata.name.clone().unwrap().contains(&job_name) {
+                    Some(pod.metadata.name.clone().unwrap())
+                } else {
+                    None
+                }
+            })
+            .next()
+            .unwrap();
+        assert!(pod_name.contains(&job_name));
```
eyalb181, line 80: Why are you asserting this?

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+                // panicked at 'Error: '`
+                if !error_stream.is_empty() {
+                    panic!("Error: {}", error_stream);
+                }
+            }
+        });
+
+        // since we are reading from the stdout, we could block at any point in case the server
+        // does not write to its stdout, so we need a timeout here
+        let validation_duration = Duration::from_secs(30);
```
eyalb181, line 49: I'd call this validation_timeout, not validation_duration. The validation duration should (hopefully) be a lot shorter.

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+                if pod.metadata.name.clone().unwrap().contains(&job_name) {
+                    Some(pod.metadata.name.clone().unwrap())
+                } else {
+                    None
+                }
+            })
+            .next()
+            .unwrap();
+        assert!(pod_name.contains(&job_name));
+        sleep(Duration::from_secs(35)).await;
```
eyalb181, line 81: As discussed (I think):
We want to try and not sleep the whole 35 seconds, if possible. So lets do something like this:

counter = 0
job_len = 1
while counter =< 35 && job_len > 0
{
job_len = pods_api...
sleep(1)
counter++
}
assert(job_len, 0)

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+
+        //verify cleanup
+        let updated_jobs = jobs_api.list(&ListParams::default()).await.unwrap();
+        assert_eq!(updated_jobs.items.len(), 0);
+
+        let updated_pods = pods_api.list(&ListParams::default()).await.unwrap();
+        assert_eq!(updated_pods.items.len(), 1); // only the nginx pod should exist
```
eyalb181, line 88: Maybe verify the name here too, to make the code clearer and more robust

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+    // we send a request to a different pod in the cluster (different namespace) and assert
+    // that no operation is performed as specified in the request by the server
+    // as the agent pod is impersonating the pod running in the default namespace
+    async fn test_different_pod_in_cluster() {
+        let client = setup_kube_client().await;
+
+        let namespace = "test-namespace";
+        create_namespace(&client, namespace).await;
+        create_nginx_pod(&client, namespace).await;
+        sleep(Duration::from_secs(2)).await; // we need a short sleep here, otherwise the test panics
```
eyalb181, line 101: Why does it panic without the sleep?

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+        assert_eq!(updated_pods.items.len(), 1); // only the nginx pod should exist
+    }
+
+    #[tokio::test]
+    // we send a request to a different pod in the cluster (different namespace) and assert
+    // that no operation is performed as specified in the request by the server
+    // as the agent pod is impersonating the pod running in the default namespace
+    async fn test_different_pod_in_cluster() {
+        let client = setup_kube_client().await;
+
```
eyalb181, line 97: Lets extract all this setup code to an external function instead of repeating it in each test

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+        let service_url = get_service_url(&client, "default").await.unwrap();
+        let pod_name = get_nginx_pod_name(&client, "default").await.unwrap();
+        let command = vec!["node", "node-e2e/app.js"];
+        let env: HashMap<&str, &str> = HashMap::new(); // for adding more environment variables
+        let mut server = start_node_server(&pod_name, command, env);
+
+        let mut stderr_reader = BufReader::new(server.stderr.take().unwrap());
+        let child_stdout = server.stdout.take().unwrap();
+
+        setup_panic_hook();
```
eyalb181, line 31: Why is this necessary?

.github/workflows/ci.yaml
eyalb181: Save this for later though - let's get this PR merged first.

tests/src/sanity.rs
```diff
@@ -0,0 +1,326 @@
...
+        let mut line = String::new();
+
+        let timeout_duration = Duration::from_secs(10);
+        timeout(timeout_duration, reader.read_line(&mut line))
+            .await
+            .unwrap()
+            .unwrap();
+
+        assert!(line.contains("Server listening on port 80"));
+        reqwest_request(service_url.as_str(), Method::PUT).await;
```
eyalb181, line 132: Lets try to keep all the implicit knowledge of what the API does in one place. You already have validate_requests, so maybe add validate_no_requests next to it, or something like that.

eyalb181: Just to elaborate - right now, if I make a change to how app.js work, I would have to go over the entire code and look for dependencies. So instead we should strive to cenralize all code that relies on the internal implementation of app.js in one place.

tests/src/utils.rs
```diff
@@ -0,0 +1,236 @@
...
+    let client = Client::try_from(config).unwrap();
+    client
+}
+
+// minikube service nginx --url
+pub async fn get_service_url(client: &Client, namespace: &str) -> Option<String> {
+    let service_api: Api<Service> = Api::namespaced(client.clone(), namespace);
+    let pod_api: Api<Pod> = Api::namespaced(client.clone(), namespace);
+    for pod in pod_api.list(&ListParams::default()).await.unwrap() {
+        if pod.metadata.name.unwrap().contains("nginx") {
```
eyalb181, line 53: Rather than querying all the pods and filtering them in memory, you can filter them in the api call and then just call .first(). See pod_api.rs in mirrord_layer for an example.

eyalb181: This can wait for after the merge.

tests/src/utils.rs
```diff
@@ -0,0 +1,236 @@
...
+// kubectl delete namespace name
+pub async fn delete_namespace(client: &Client, namespace: &str) {
+    let namespace_api: Api<Namespace> = Api::all(client.clone());
+    namespace_api
+        .delete(namespace, &DeleteParams::default())
+        .await
+        .unwrap();
+}
+
+pub async fn reqwest_request(url: &str, method: Method) {
```
eyalb181, line 102: A. try a clearer function name
B. as a rule, lets not have asserts in util functions, only in the tests themselves

tests/src/utils.rs
```diff
@@ -0,0 +1,236 @@
...
+            assert_eq!(res.status(), StatusCode::from_u16(405).unwrap())
+        }
+        _ => panic!("unexpected method"),
+    }
+}
+
+// kubectl apply -f e2e/app.yaml -n name
+pub async fn create_nginx_pod(client: &Client, namespace: &str) {
+    let deployment_api: Api<Deployment> = Api::namespaced(client.clone(), namespace);
+    let deployment = serde_json::from_value(json!({
```
eyalb181, line 122: Isn't this the same as app.yaml? Why not use it?

eyalb181 (changes requested):

tests/src/sanity.rs
infiniteregrets: so the pod name contains the name of the job, just making sure i selected the correct job

Cargo.toml
```diff
@@ -4,7 +4,11 @@ members = [
     "mirrord-protocol",
     "mirrord-agent",
     "mirrord-layer",
-    "mirrord-cli",
+    "mirrord-cli",    
+]
+
+exclude = [
```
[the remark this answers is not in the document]
infiniteregrets: i am doing this because one of the tests, i believe test_agent_image fails and iirc it was something related to manifest not found. I can rerun the tests with tests being a part of the workspace members and ping you if that error happens again

.github/workflows/ci.yaml
infiniteregrets: yes, i'll add macos

tests/src/sanity.rs
infiniteregrets: we cannot panic from inside the thread. even if we panic the test doesnt fail so we need to change how the default panic works

.github/workflows/ci.yaml
infiniteregrets: added to my todo list, will open an issue later

tests/src/sanity.rs
infiniteregrets: reduced the duration to 20 seconds

tests/src/sanity.rs
```diff
@@ -0,0 +1,284 @@
...
+                        let nginx_pod = updated_pods.items[0].metadata.name.clone().unwrap();
+                        assert!(nginx_pod.contains("nginx"));
+                        break;
+                    }
+                }
+            }),
+        )
+        .await
+        .unwrap()
+        .unwrap();
```
infiniteregrets, line 83: @aviramha @meowjesty do you guys think this timeout on job cleanup is okay? or is there a better way to do it?

meowjesty (commented):
Feedback regarding sleeps and watcher.

PR #55, opened 2022-05-16
Add watch on pod creation

infiniteregrets:
It looks like Job creation and pod listing happens too fast, such that the pod has not been created yet and this leads to the following panic:
```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', mirrord-layer/src/pod_api.rs:127:34
```
`let pod = pods.items.first().unwrap();`
We use the watch API and wait till the resource has been added. https://docs.rs/kube/latest/kube/struct.Api.html#method.watch

### DIFF
--- a/mirrord-layer/src/pod_api.rs
+++ b/mirrord-layer/src/pod_api.rs
@@ -1,13 +1,15 @@
 use envconfig::Envconfig;
+use futures::{StreamExt, TryStreamExt};
 use k8s_openapi::api::{batch::v1::Job, core::v1::Pod};
 use kube::{
     api::{Api, ListParams, Portforwarder, PostParams},
+    core::WatchEvent,
     runtime::wait::{await_condition, conditions::is_pod_running},
     Client, Config,
 };
 use rand::distributions::{Alphanumeric, DistString};
 use serde_json::json;
-use tracing::warn;
+use tracing::{error, warn};
 
 use crate::config;
 struct RuntimeData {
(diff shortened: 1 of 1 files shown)


mirrord-layer/src/pod_api.rs
```diff
@@ -120,6 +122,16 @@ pub async fn create_agent(
         .unwrap();
 
     let pods_api: Api<Pod> = Api::namespaced(client.clone(), agent_namespace);
+    let params = ListParams::default().labels(&format!("job-name={}", agent_job_name));
+    let mut stream = pods_api.watch(&params, "0").await.unwrap().boxed();
+    while let Some(status) = stream.try_next().await.unwrap() {
```
aviramha, line 127: need to add timeout on that

aviramha (changes requested):

mirrord-layer/src/pod_api.rs
infiniteregrets: not sure but what if the resource has not been added yet and if we add a timeout, wouldnt it break out of the loop without the pod being there? the docs say the default timeout is 295 seconds though

mirrord-layer/src/pod_api.rs
```diff
@@ -120,6 +122,16 @@ pub async fn create_agent(
...
+        match status {
+            WatchEvent::Added(_) => break,
+            WatchEvent::Error(_) => error!("Error watching pods"),
```
aviramha, line 130: break here/unwrap

mirrord-layer/src/pod_api.rs
aviramha: well, yes but we don't want it to wait indefinitely

aviramha (changes requested):

mirrord-layer/src/pod_api.rs
infiniteregrets: added a 10 second timeout

mirrord-layer/src/pod_api.rs
infiniteregrets: not sure wym, where?

mirrord-layer/src/pod_api.rs
```diff
@@ -120,6 +122,21 @@ pub async fn create_agent(
...
 
     let pods_api: Api<Pod> = Api::namespaced(client.clone(), agent_namespace);
+    let params = ListParams::default()
+        .labels(&format!("job-name={}", agent_job_name))
+        .timeout(10);
+    let mut stream = pods_api.watch(&params, "0").await.unwrap().boxed();
+    while let Some(status) = stream.try_next().await.unwrap() {
+        match status {
+            WatchEvent::Added(_) => break,
+            WatchEvent::Error(_) => {
```
aviramha, line 132: capture the error variable and add it to the error macro

aviramha (commented):
almost there!

aviramha (approved):
Awesome, Thanks!

PR #56, opened 2022-05-17
V2.0

meowjesty:
Alright my initial PR is adding just a hook for `open`, but in the process I also ended up refactoring the `select!` macro usage.

Before this becomes an even bigger PR, I'm pushing it to get some reviews.


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM rust:latest as build-env
```
aviramha, line 1: I wouldn't put the dockerfile at root directory, as it only builds the agent..

mirrord-agent/src/main.rs
```diff
@@ -94,6 +97,34 @@ struct PeerMessage {
...
+            );
+
+            let file = std::fs::File::open(path)?;
+            let file_fd = file.as_raw_fd();
+
+            debug!("handle_peer_message -> file is open with fd {file_fd:?}");
+
+            let open_file_message =
+                DaemonMessage::OpenFileResponse(FileOpenResponse { fd: file_fd });
+            stream.send(open_file_message).await?;
```
aviramha, line 120: You should call open, then if there's an error return the error, if it's success you just save the fd (the layer doesn't need the actual fd that the agent has)

mirrord-agent/src/main.rs
```diff
@@ -181,22 +214,25 @@ async fn start() -> Result<()> {
...
                         state.remove_peer(message.peer_id);
                         let ports = state.port_subscriptions.get_subscribed_topics();
                         packet_command_tx.send(SnifferCommand::SetPorts(ports)).await?;
                     },
                     ClientMessage::ConnectionUnsubscribe(connection_id) => {
                         state.connections_subscriptions.unsubscribe(message.peer_id, connection_id);
                     }
-
-
+                    ClientMessage::OpenFileRequest(_) => {
```
aviramha, line 231: we need to add a todo, maybe create a different enum for this match

mirrord-agent/src/sniffer.rs
```diff
@@ -202,14 +202,6 @@ impl PacketCodec for TCPManagerCodec {
 
     fn decode(&mut self, packet: pcap::Packet) -> Result<Self::Type, pcap::Error> {
         Ok(packet.data.to_vec())
```
aviramha, line 204: Thanks!

mirrord-layer/src/file.rs
```diff
@@ -0,0 +1,170 @@
...
+
+// https://users.rust-lang.org/t/problem-overriding-libc-functions-with-ld-preload/41516/4
+
+/// NOTE(alex): Regex that ignores system files + files in the current working directory.
+static IGNORE_FILES: SyncLazy<RegexSet> = SyncLazy::new(|| {
+    // NOTE(alex): To handle the problem of injecting `open` and friends into project runners (like
+    // in a call to `node app.js`, or `cargo run app`), we're ignoring files from the current
+    // working directory (as suggested by aviram).
+    let mut cwd_buffer = [0; libc::PATH_MAX as usize];
+    let cwd = unsafe { libc::getcwd(cwd_buffer.as_mut_ptr(), cwd_buffer.len()) };
```
aviramha, line 30: why not [current_dir](https://doc.rust-lang.org/std/env/fn.current_dir.html) ?

mirrord-layer/src/file.rs
```diff
@@ -0,0 +1,170 @@
...
+            Ok(_) => {}
+            Err(fail) => {
+                error!("open: failed to send open message: {fail:?}!");
+                return libc::EFAULT;
+            }
+        };
+
+        debug!("Blocking on file open operation!");
+
+        let file_open = RUNTIME.block_on(async {
```
aviramha, line 93: I'm pretty sure we don't want to use tokio runtime in hooks? I didn't actually think that's a possibility, so maybe we do and it'll make rest of the code easier..?
anyway you can do https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Receiver.html#method.blocking_recv

mirrord-layer/src/file.rs
```diff
@@ -0,0 +1,170 @@
...
+        // - send an open file request to `-agent` in `poll_agent`;
+        // - when the reply comes back, `send_tx.send(file)` in `poll_agent`;
+
+        file_open.fd
+    }
+}
+
+/// NOTE(alex): libc also has an `open64` function. Both functions point to the same address, so
+/// trying to intercept them all will result in an `InterceptorAlreadyReplaced` error.
+pub(super) unsafe extern "C" fn open_detour(path: *const c_char, oflag: c_int) -> RawFd {
```
aviramha, line 114: I would consider only hooking read operations atm? @eyalb181 UX wise it might be better initially

mirrord-layer/src/file.rs
```diff
@@ -0,0 +1,170 @@
...
+unsafe extern "C" fn fopen_detour(filename: *const c_char, mode: *const c_char) -> *mut FILE {
+    let filename_str = CStr::from_ptr(filename)
+        .to_str()
+        .expect("Failed converting filename from c_char!");
+    debug!("fopen_detour -> filename {filename_str}");
+
+    let file_ptr = if IGNORE_FILES.is_match(filename_str) {
+        libc::fopen(filename, mode)
+    } else {
+        libc::fopen(filename, mode)
```
aviramha, line 129: seems like a bug?

mirrord-layer/src/lib.rs
```diff
@@ -97,116 +98,175 @@ fn init() {
...
+                    fd: listen.fd,
+                },
+            );
+        }
+        Some(HookMessage::OpenFileHook(open)) => {
+            debug!("HookMessage::OpenFileHook {open:#?}");
+
+            // NOTE(alex): Lock the file handler and insert a channel that will be used to retrieve
+            // the file data when it comes back from `DaemonMessage::OpenFileResponse`.
+            open_file_handler.lock().unwrap().push(open.file_channel_tx);
```
aviramha, line 152: I wouldn't choose a vec mechanism as this won't work if we get two open at the same time (threads), it's kinda error prune. I'd just use a hashmap of request id to response id?

meowjesty:
I forgot to mention that the changes to `-agent` require using an updated image, in `vscode/launch.json`:

`"MIRRORD_AGENT_IMAGE": "meowjesty/mirrord-agent:latest",`

mirrord-layer/src/lib.rs
```diff
@@ -97,116 +98,175 @@ fn init() {
...
 
+/// Enables file and socket hooks.
 fn enable_hooks() {
-    let interceptor = Interceptor::obtain(&GUM);
-    sockets::enable_hooks(interceptor)
+    let mut interceptor = Interceptor::obtain(&GUM);
+    interceptor.begin_transaction();
+
+    sockets::enable_socket_hooks(&mut interceptor);
+    file::enable_file_hooks(&mut interceptor);
```
aviramha, line 269: Make file hooks optional by env var, to be confirmed with @eyalb181 (version wise and UX wise it'd be weird if a new version starts hooking file operations without a flag set, if we don't release it as major)

aviramha (commented):
Excellent work! I really liked the refactoring you did. I had some comments on design and some practical.

mirrord-layer/src/file.rs
meowjesty: Think both are roughly equivalent, but I'm changing it to use `tokio::oneshot` instead of `futures::oneshot` and apply your suggestion of using `blocking_recv`.

PR #58, opened 2022-05-17
Refactoring `mirrord-layer` `select!` body into functions.

meowjesty:
Still not the best solution, as error handling is (mostly) missing, but I believe this will be a quality of life improvement.

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -116,6 +116,7 @@ fn init() {
     RUNTIME.spawn(poll_agent(pf, receiver));
 }
 
+#[inline]
 async fn handle_hook_message(
     hook_message: HookMessage,
     port_mapping: &mut HashMap<Port, ListenData>,
@@ -142,6 +143,7 @@ async fn handle_hook_message(
     }
 }
 
+#[inline]
 async fn handle_daemon_message(
     daemon_message: DaemonMessage,
     port_mapping: &mut HashMap<Port, ListenData>,
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -1,13 +1,10 @@
-// #![feature(c_variadic)]
-
 use std::{
     collections::HashMap,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     os::unix::io::RawFd,
-    thread,
-    time::Duration,
 };
 
+use actix_codec::{AsyncRead, AsyncWrite};
 use ctor::ctor;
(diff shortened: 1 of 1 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -97,110 +94,129 @@ fn init() {
...
+        }
+        DaemonMessage::Close => todo!(),
+        DaemonMessage::LogMessage(_) => todo!(),
+    }
+}
+
 async fn poll_agent(mut pf: Portforwarder, mut receiver: Receiver<HookMessage>) {
     let port = pf.take_stream(61337).unwrap(); // TODO: Make port configurable
+
+    // NOTE(alex): `codec` is used to retrieve messages from the daemon (messages that are sent
```
aviramha, line 208: docstrings are welcome, but no need to put names ;P

aviramha (approved):

aviramha:
Thanks for the needed refactor!

mirrord-layer/src/lib.rs
meowjesty: Oops, sorry I use an extension to show docs in a tree like view, so tagging them like so helps me with finding stuff.

Do I drop only the `(alex)` part or do you want me to drop the tag (`NOTE`) as well?

aviramha: Yeah I assumed that's some sort of an extension. 
I think it'd be better to drop the tag as well, as most developers probably don't share the same extension and it's pretty implicit that a comment is a note ;p

eyalb181 (approved):

mirrord-layer/src/lib.rs
meowjesty: Alright, will drop those.

mirrord-layer/src/lib.rs
```diff
@@ -97,110 +94,129 @@ fn init() {
...
     unsafe {
         HOOK_SENDER = Some(sender);
     };
+
     enable_hooks();
+
     RUNTIME.spawn(poll_agent(pf, receiver));
 }
 
+async fn handle_hook_message(
```
aviramha, line 119: nitpick: add inline, compiler will probably do it automatically, but a hint in the direction is always good (one call site, this function is separated from caller only to be more readable. 
same for the other handle function.

aviramha (approved):
Approved, check the nitpick and I'll merge it.

mirrord-layer/src/lib.rs
meowjesty: Done.

PR #65, opened 2022-05-19
File open + read

meowjesty:
Draft PR of file operations, working `open` and `read` calls.


meowjesty:
Hey, I'm not a fan of this clippy warning:

```
error: returning the result of a `let` binding from a block
   --> mirrord-layer/src/files.rs:133:9
    |
132 |         let fd = libc::open(raw_path, oflag);
    |         ------------------------------------- unnecessary `let` binding
133 |         fd
    |         ^^
    |
```

I find that these "unnecessary" bindings really help out when using a debugger.

Is it okay to disable `clippy::let-and-return`?

aviramha:
> Hey, I'm not a fan of this clippy warning:
> 
> ```
> error: returning the result of a `let` binding from a block
>    --> mirrord-layer/src/files.rs:133:9
>     |
> 132 |         let fd = libc::open(raw_path, oflag);
>     |         ------------------------------------- unnecessary `let` binding
> 133 |         fd
>     |         ^^
>     |
> ```
> 
> I find that these "unnecessary" bindings really help out when using a debugger.
> 
> Is it okay to disable `clippy::let-and-return`?

if it's a return value then you should be able to still inspect it..

mirrord-agent/src/main.rs
```diff
@@ -58,6 +65,49 @@ impl Borrow<PeerID> for Peer {
...
+impl FileManager {
+    // TODO(alex) [mid] 2022-05-19: Need to do a better job handling errors here (and in `read`)?
+    // Ideally we would send the error back in the `Response`, to be handled by the mirrod-layer
+    // side, and converted into some `libc::X` value.
+    pub(crate) fn open(&mut self, path: PathBuf) -> Result<i32> {
+        debug!("FileManager::open -> Trying to open file {path:#?}.");
+
+        // TODO(alex) [low] 2022-05-18: To avoid opening the same file multiple times, we could
+        // search `FileManager::open_files` for a path that matches the one passed as a parameter.
+        let file = std::fs::File::open(path)?;
```
aviramha, line 82: why do we keep a fd instead of just File object?

aviramha: okay that's weird it seems that you're creating a File struct then you transform it into fd then you convert it back to File? why?

mirrord-agent/src/main.rs
```diff
@@ -58,6 +65,49 @@ impl Borrow<PeerID> for Peer {
...
+        Ok(fd)
+    }
+
+    pub(crate) fn read(&self, fd: i32, buffer_size: usize) -> Result<ReadFileResponse> {
+        let mut file = self.open_files.get(&fd).unwrap();
+
+        debug!("FileManager::read -> Trying to read file {file:#?}, with count {buffer_size:#?}");
+        debug!(
+            "FileManager::read -> File has len {:#?}",
+            file.metadata().unwrap().len()
```
aviramha, line 96: I don't think we should call another function just for the debug print?

mirrord-layer/src/files.rs
```diff
@@ -0,0 +1,194 @@
...
+        read_amount: read_file_response.read_amount,
+    };
+
+    Ok(read_file)
+}
+
+/// NOTE(alex): libc also has an `open64` function. Both functions point to the same address, so
+/// trying to intercept them all will result in an `InterceptorAlreadyReplaced` error.
+pub(super) unsafe extern "C" fn open_detour(raw_path: *const c_char, oflag: c_int) -> RawFd {
+    let path: PathBuf = CStr::from_ptr(raw_path)
```
aviramha, line 126: I'd move this code to the safe function, as we want as less unsafe as possible (same for other detour functions)

mirrord-layer/src/files.rs
```diff
@@ -0,0 +1,194 @@
...
+        }) {
+            Ok(fd) => fd,
+            Err(fail) => fail,
+        }
+    }
+}
+
+unsafe extern "C" fn read_detour(fd: c_int, buf: *mut c_void, count: size_t) -> ssize_t {
+    // NOTE(alex): We're only interested in files that are handled by `mirrord-agent`.
+    if let Some(managed_fd) = OPEN_FILES.lock().unwrap().get(&fd) {
```
aviramha, line 147: you're locking the `OPEN_FILES` mutex during the whole read, which is quite unnecessary imo.

aviramha (changes requested):
Looks great, few design/implementation details comments in line.

meowjesty:
Yeah, but it's so much easier to just look at the `local` view (or hover over the binding) to see the resulting value.

Plus I would argue that it can improve readability, we know what `open` returns, but for more complex returns I find that naming the result makes it clearer, an example:

```
return foo();
```

versus

```
let filename = foo();
return filename;
```

mirrord-agent/src/main.rs
meowjesty: Hmm, I'm just grabbing the fd from the file to use as an id for `open_files`, not sure what you mean by "convert it back".

aviramha:
> Yeah, but it's so much easier to just look at the `local` view (or hover over the binding) to see the resulting value.
> 
> Plus I would argue that it can improve readability, we know what `open` returns, but for more complex returns I find that naming the result makes it clearer, an example:
> 
> ```
> return foo();
> ```
> 
> versus
> 
> ```
> let filename = foo();
> return filename;
> ```

I get your point but I think deviating from standards (which are very well defined and shared in the Rust community) has a more significant cost than altering a variable name when debugging..

mirrord-agent/src/main.rs
aviramha: Ah, got it. I would recommend against it, Imagine a case where the "remote fd" is 3 and then layer's `read` gets called with fd 3. This could either be the remote fd or it might be a socket fd that exists locally.
It's okay if you pass this id as the remote fd, but on layer you need the hashset to be mapped by a number that is guaranteed to not be taken as a fd, so maybe create a fd just to assure you're not reusing something that exists. (on the layer)

mirrord-layer/src/files.rs
```diff
@@ -107,7 +113,28 @@ pub(crate) fn open(path: PathBuf, open_options: OpenOptionsInternal) -> Result<R
...
 
     let fake_file_name = CString::new(open_file.fd.to_string()).unwrap();
-    let local_file_fd = unsafe { libc::memfd_create(fake_file_name.as_ptr(), 0) };
+
+    let mode = (open_flags & O_RDONLY) | (open_flags | O_RDONLY) | O_CREAT;
+    // TODO(alex) [mid] 2022-05-24: Be very careful here, if a call to create the local fd fails,
+    // the remote one stays up (basically a leak). So I need a way to `close` the remote on failure
+    // here.
+    //
+    // NOTE(alex): The pair `shm_open`, `shm_unlink` are used to create a temporary file
```
aviramha, line 122: why do we care about flags though? we "tunnel" those flags to the agent and we hook all operations on the file (presumably) so the fact it's shm or memfd doesn't really matter?

meowjesty: Yeah, I'm just being cautious here, hoping to mimic what the file system would expect as much as possible.

mirrord-agent/src/file.rs
```diff
@@ -0,0 +1,113 @@
+use std::{
```
aviramha, line 1: P.S all the file ops need to run in the namespace of the container id given to us, as sniffer does. Right now you probably relay the calls to the pod host rather then to the pod ;)

PR #66, opened 2022-05-20
Vscode extension ci

eyalb181:


.github/workflows/ci.yaml
```diff
@@ -66,4 +66,37 @@ jobs:
...
-    - run: cargo +nightly build --manifest-path=mirrord-layer/Cargo.toml
\ No newline at end of file
+    - run: cargo +nightly build --manifest-path=mirrord-layer/Cargo.toml
+
+
+  release_vscode_ext:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-18.04
```
aviramha, line 76: I think we don't have to do the matrix, as we can cross-compile probably?

PR #70, opened 2022-05-22
Publish ext in ci

eyalb181:
Add a CI workflow to publish the VSCode extension for all supported platforms.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,6 +54,6 @@ jobs:
     - run: npm install -g vsce typescript esbuild
     - run: cargo +nightly build --release --manifest-path=mirrord-layer/Cargo.toml --target=${{matrix.target}}
     - run: mv target/release/libmirrord_layer.${{matrix.extension}} vscode-ext/
-    - run: cd vscode-ext && npm install && tsc && cat package.json && vsce package  {{github.event.inputs.version}} --allow-star-activation --target ${{matrix.code-target}} --pre-release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -16,4 +22,38 @@ jobs:
...
+            code-target: darwin-arm64    
+    runs-on: ${{matrix.os}}
+    steps:  
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: nightly
+    - run: npm install -g vsce typescript esbuild
+    - run: rustup target add ${{matrix.target}}
```
aviramha, line 54: you can specify it in the toolchain action:
https://github.com/marketplace/actions/rust-toolchain

aviramha (changes requested):

.github/workflows/release.yaml
```diff
@@ -16,4 +22,38 @@ jobs:
...
+    - run: cargo +nightly build --release --manifest-path=mirrord-layer/Cargo.toml --target=${{matrix.target}}
+    - run: mv target/release/libmirrord_layer.${{matrix.extension}} vscode-ext/
+    - run: cd vscode-ext && npm install && tsc && cat package.json && vsce package  {{github.event.inputs.version}} --allow-star-activation --target ${{matrix.code-target}} --pre-release
```
aviramha, line 57: this wouldn't work if we ran it on release?

PR #72, opened 2022-05-23
Fix fd leak

aviramha:
We had fd leak because we didn't hook close, meaning that if a socket got closed we'd still hold it in our `SOCKETS` struct.
This adds a hook for close to add removal of a fd from our data.


meowjesty (approved):
Looks good.

PR #79, opened 2022-05-25
Use local agent images

infiniteregrets:
Just a small change `"imagePullPolicy": "IfNotPresent"` as we should be able to use local images for mirrord-agent.
https://kubernetes.io/docs/concepts/containers/images/


aviramha (approved):
Great catch, Thank you!

PR #86, opened 2022-05-25
Ignore connections closed by the application

eyalb181:


mirrord-layer/src/lib.rs
```diff
@@ -180,8 +180,14 @@ async fn handle_daemon_message(
                 });
         }
         DaemonMessage::TCPData(msg) => {
-            if let Err(fail) = active_connections
-                .get(&msg.connection_id)
+            debug!("Received data from connection id {}", msg.connection_id);
+            let connection = active_connections
+                .get(&msg.connection_id);
+            if connection.is_none() {
```
aviramha, line 186: This is kinda quirky, add a todo to beautify it someday.

aviramha (approved):

mirrord-layer/src/lib.rs
aviramha: Using is_none then unwrapping is an anti pattern https://github.com/rust-lang/rust/issues/75661

PR #94, opened 2022-05-27
Add E2E tests

infiniteregrets:
This PR adds the following E2E tests:

- built with the latest agent image 
`docker build -t test . --file mirrord-agent/Dockerfile`
`minikube load image test:latest`
- **test_complete_node_api**: sends four different requests to the node server present in node-e2e
`GET, POST, PUT, DELETE`. The node process writes to it's stdout which is then checked and verified by the test. Everything runs in the default namespace. The test also verifies the agent job and pod are cleaned up.
- **test_different_pod_in_cluster**: A `PUT` request is sent to a different pod in the cluster and asserted that no action corresponding to the request takes place.
- **test_good_agent_namespace**: A new namespace is created and it is specified as the agent's namespace and four different requests are sent to verify the server writes to its stdout. 
- **test_nonexistent_agent_namespace**: A nonexistent namespace is specified as agent's namespace and it is asserted that the process panics. 
- **test_good_pod_namespace**: A new namespace is created and it is specified as the pod's namespace and four different requests are sent to verify the server writes to its stdout. 
- **test_bad_pod_namespace**: A nonexistent namespace is specified as pod's namespace and it is asserted that the process panics.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "mirrord-agent",
     "mirrord-layer",
     "mirrord-cli",
+    "tests"
 ]
 
 [workspace.package]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,3 +75,31 @@ jobs:
           key: ${{matrix.os}}
       # For now, just evrify it compiles.
       - run: cargo +nightly build --manifest-path=mirrord-layer/Cargo.toml
+
+  e2e:
+    runs-on: ubuntu-latest
+    name: e2e
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -7,6 +7,11 @@ members = [
     "mirrord-cli",
 ]
 
+# CI: test_agent_image fails if "tests" is no excluded from the workspace
```
aviramha, line 10: you can change the ci command so it explicitly ignores tests on that run?

aviramha (approved):
LGTM! Thanks

eyalb181 (approved):

aviramha (approved):

PR #96, opened 2022-05-30
Add support for Docker runtime 

### LINKED ISSUE #95: Add support for docker runtime
We currently only support Kubernetes clusters using the containerd runtime. We need to add support for docker runtime as well. The containerd-specific code is in the agent where we get the linux namespace from the pod name, and in the pod creation in the layer.

infiniteregrets:
Closes #95


mirrord-agent/src/namespace.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use anyhow::Result;
+use async_trait::async_trait;
+use nix::sched::setns;
+use serde::{Deserialize, Serialize};
+
+#[async_trait]
+pub trait Namespace: Send + Sync {
+    async fn get_namespace(&self) -> Result<String>;
+    fn set_namespace(&self, ns_path: String) -> Result<()> {
```
aviramha, line 14: why would we put set_namespace in a trait?

mirrord-agent/src/docker.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use bollard::{container::InspectContainerOptions, Docker};
+
+use crate::namespace::Namespace;
+
```
aviramha, line 6: put it also in runtime, maybe we need to split but this split isn't good anyway..

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,59 @@
-use std::{
-    fs::File,
-    os::unix::io::{IntoRawFd, RawFd},
-};
-
 use anyhow::{anyhow, Result};
```
eyalb181, line 1: If this file is the containerd equivalent of docker.rs, shouldn't it be called containerd.rs?

eyalb181 (changes requested):

mirrord-agent/src/sniffer.rs
```diff
@@ -236,12 +235,17 @@ pub async fn packet_worker(
...
     container_id: Option<String>,
+    container_runtime: Option<String>,
 ) -> Result<()> {
     debug!("setting namespace");
-    if let Some(container_id) = container_id {
-        let namespace = get_container_namespace(container_id).await?;
-        debug!("Found namespace to attach to {:?}", &namespace);
-        set_namespace(&namespace)?;
+    if let (Some(container_id), Some(container_runtime)) = (container_id, container_runtime) {
+        let runtime: Box<dyn Namespace> = match container_runtime.as_str() {
```
aviramha, line 242: I think we can have better design here than using Box + dyn. 
I'm trying to create an example, maybe @meowjesty will have an idea.

aviramha (changes requested):

mirrord-agent/src/namespace.rs
infiniteregrets: set_namespace is the same across all the types, thats why it is put in the trait. didnt want to repeat code

mirrord-agent/src/docker.rs
infiniteregrets: the only reason i split this was to not make everything so tightly coupled

aviramha: I get it, but this split doesn't make sense so I'd put it in one file then maybe consider re-organizing it.

mirrord-agent/src/namespace.rs
aviramha: it doesn't use self -> it can be a regular function.

aviramha:
P.S - Changelog

aviramha (changes requested):

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,65 @@
...
+    let pid = inspect_response
+        .state
+        .and_then(|state| state.pid)
+        .ok_or_else(|| anyhow!("No pid found"))?;
+
+    Ok(PathBuf::from(pid.to_string()))
+}
+
+async fn get_containerd_container_pid(container_id: String) -> Result<PathBuf> {
+    let (containerd_socket, default_namespace) = ("/run/containerd/containerd.sock", "k8s.io");
```
meowjesty, line 50: Move these strings outside, make them global `const` if they never change.

mirrord-agent/src/sniffer.rs
```diff
@@ -236,13 +237,29 @@ pub async fn packet_worker(
...
     container_id: Option<String>,
+    container_runtime: Option<String>,
 ) -> Result<()> {
     debug!("setting namespace");
-    if let Some(container_id) = container_id {
-        let namespace = get_container_namespace(container_id).await?;
-        debug!("Found namespace to attach to {:?}", &namespace);
-        set_namespace(&namespace)?;
+
+    let default_runtime = "containerd";
```
meowjesty, line 244: Same thing as my other comment, if this never changes, move it outside as a `const`.

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,65 @@
...
-struct LinuxInfo {
-    namespaces: Vec<Namespace>,
-}
-
-#[derive(Serialize, Deserialize, Debug)]
-struct Spec {
-    linux: LinuxInfo,
+impl Runtime {
+    pub async fn get_container_pid(container_id: &str, container_runtime: &str) -> Result<PathBuf> {
+        match container_runtime {
```
meowjesty, line 20: `match`ing on strings like this feels weird, maybe you could have an enum that wraps the string + container_id, like:

```suggestion
        match container_runtime {
            Runtime::Docker(container_id) => ...
            Runtime::Containerd(container_id) => ...
```

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,65 @@
...
+    let inspect_response = client
+        .inspect_container(&container_id, inspect_options)
+        .await?;
+
+    let pid = inspect_response
+        .state
+        .and_then(|state| state.pid)
+        .ok_or_else(|| anyhow!("No pid found"))?;
+
+    Ok(PathBuf::from(pid.to_string()))
```
meowjesty, line 46: I would leave converting the `pid` into a `PathBuf` on the caller side, so here we return just the `pid` as a number, and when some other function needs to call like `set_namespace`, then they can do the conversion there like:

```rust
set_namespace(PathBuf::from(pid));
```

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,65 @@
...
+        ..Default::default()
+    };
+    let request = with_namespace!(request, default_namespace);
+    let response = client.get(request).await?;
+    let process = response
+        .into_inner()
+        .process
+        .ok_or_else(|| anyhow!("No pid found"))?;
+
+    Ok(PathBuf::from(process.pid.to_string()))
```
meowjesty, line 64: Same as my other comment about `pid` -> `PathBuf` conversion.

mirrord-agent/src/sniffer.rs
```diff
@@ -236,13 +237,29 @@ pub async fn packet_worker(
...
+    let pid = match (container_id, container_runtime) {
```
meowjesty, line 245: Having an enum like:

```rust
enum Runtime {
    Docker(String),
    Containerd(String),
}
```

Would simplify this match, I think.

meowjesty (changes requested):
Thank you for figuring out the PID thing!

mirrord-agent/src/runtime.rs
infiniteregrets: the callee converts it to PathBuf because the `get_docker_container_pid` returns an i64 for the process however the containerd one returns a u32. I dont know if we should cast i64 to u32 but i think PIDs cant be negative. WDYT @aviramha

aviramha:
Gentle reminder to not forget adding this to the change log ;)

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,70 @@
...
     tonic::Request,
     with_namespace,
 };
 use nix::sched::setns;
-use serde::{Deserialize, Serialize};
+use tracing::debug;
 
 const CONTAINERD_SOCK_PATH: &str = "/run/containerd/containerd.sock";
 const DEFAULT_CONTAINERD_NAMESPACE: &str = "k8s.io";
+pub(crate) enum Runtime {}
```
aviramha, line 20: I don't see any reason this would be an Enum with current design (totally okay to have just public functions..)

mirrord-agent/src/runtime.rs
```diff
@@ -1,68 +1,65 @@
...
-#[derive(Serialize, Deserialize, Debug)]
-struct Namespace {
-    #[serde(rename = "type")]
-    ns_type: String,
-    path: Option<String>,
+pub async fn get_container_pid(container_id: &str, container_runtime: &str) -> Result<u64> {
+    match container_runtime {
+        "docker" => get_docker_container_pid(container_id.to_string()).await,
+        "containerd" => get_containerd_container_pid(container_id.to_string()).await,
+        _ => Err(anyhow!("Unknown runtime: {}", container_runtime)),
```
aviramha, line 24: Unsupported

aviramha (approved):
Great, thank you @infiniteregrets !

eyalb181 (changes requested):

eyalb181 (approved):

PR #102, opened 2022-06-01
Prompt outdated

eyalb181:
In the CLI or VS Code extension, check if the user's version is outdated, and if so prompt them to update.


vscode-ext/src/extension.ts
```diff
@@ -53,11 +55,34 @@ async function toggle(state: vscode.Memento, button: vscode.StatusBarItem) {
 	}
 }
 
+async function checkVersion(version: string) {
+	https.get('https://us-central1-mirrord.cloudfunctions.net/get-latest-version?source=vscode-ext', (res: any) => {
```
aviramha, line 59: const + add a domain (`version.mirrord.dev`?)

aviramha, line 59: add current version to the params

aviramha (changes requested):

mirrord-cli/src/main.rs
```diff
@@ -92,7 +94,15 @@ fn add_to_preload(path: &str) -> Result<()> {
     }
 }
 
+const CURRENT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
 fn exec(args: &ExecArgs) -> Result<()> {
+    let resp = reqwest::blocking::get("https://version.mirrord.dev/get-latest-version?source=2&currentVersion=0").unwrap().text().unwrap();
```
aviramha, line 100: this should fail silently/not crash

aviramha: ah tbh this should be on main after arg parse and not only on exec.
Also we should have a flag/environment variable that disables this check

mirrord-cli/src/main.rs
```diff
@@ -125,11 +126,35 @@ fn exec(args: &ExecArgs) -> Result<()> {
     Err(anyhow!("Failed to execute binary"))
 }
 
+const CURRENT_VERSION: &str = env!("CARGO_PKG_VERSION");
 fn main() {
     registry()
         .with(fmt::layer())
         .with(EnvFilter::from_default_env())
         .init();
+    let check_version = match option_env!("MIRRORD_CHECK_VERSION") {
```
aviramha, line 135: this checks the env at compile time

mirrord-cli/src/main.rs
```diff
@@ -125,11 +126,35 @@ fn exec(args: &ExecArgs) -> Result<()> {
...
+                .get(format!(
+                    "https://version.mirrord.dev/get-latest-version?source=2&currentVersion={}",
+                    CURRENT_VERSION
+                ))
+                .timeout(Duration::from_secs(1))
+                .send()
+            {
+                if let Ok(latest_version) = Version::parse(&result.text().unwrap()) {
+                    if latest_version > Version::parse(CURRENT_VERSION).unwrap() {
+                        println!("New mirrord version available: {}. To update, run: `curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash`", latest_version);
```
aviramha, line 151: add instruction on how to disable (env)

mirrord-cli/src/main.rs
aviramha, line 135: put it in it's own func

aviramha (changes requested):

aviramha (approved):

PR #111, opened 2022-06-03
Resolves #53

### LINKED ISSUE #53: mirrord extract panics on non-existent directories
```
mehula@mehul-machine:~/mirrord$ target/debug/mirrord extract blah 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', mirrord-cli/src/main.rs:70:52
```
It would be nicer to handle the Result type gracefully and return an error or maybe prompt the user whether they want to create the directory (only useful for CLI users)

thedanvail:
Adds a more graceful exit when the directory does not exist.

If the desired behavior is any different (for example, printing the absolute path of the nonexistent directory), I can update with the absolute path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,14 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+
+## 2.1.0
 ### Added
 - Prompt user to update if their version is outdated in the VS Code extension or CLI.
 - Add support for docker runtime, closes [#95](https://github.com/metalbear-co/mirrord/issues/95).
 - Add a keep-alive to keep the agent-pod from exiting, closes [#63](https://github.com/metalbear-co/mirrord/issues/63)
 
-## 2.0.0
-
+## 2.0.4
 Complete refactor and re-write of everything.
 - The CLI/VSCode extension now use `mirrord-layer` which loads into debugged process using `LD_PRELOAD`/`DYLD_INSERT_LIBRARIES`.
   It hooks some of the syscalls in order to proxy incoming traffic into the process as if it was running in the remote pod.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+- Added a graceful exit for system paths that do not exist.
 
 ## 2.1.0
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+### Changed
 - Added a graceful exit for system paths that do not exist.
 
 ## 2.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 ### Changed
-- Added a graceful exit for system paths that do not exist.
+- Added graceful exit for library extraction logic in case of error
 
 ## 2.1.0
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ## [Unreleased]
 ### Changed
 - Added graceful exit for library extraction logic in case of error
+- Refactor the CI by splitting the building of mirrord-agent in a separate job and caching the agent image for E2E tests.
 
 ## 2.1.0
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ## [Unreleased]
 ### Added
 - Prompt user to update if their version is outdated in the VS Code extension or CLI.
-- Set `MIRRORD_CHECK_VERSION` to false to make E2E tests not read update messages. 
+- Add support for docker runtime, closes [#95](https://github.com/metalbear-co/mirrord/issues/95).
 - Add a keep-alive to keep the agent-pod from exiting, closes [#63](https://github.com/metalbear-co/mirrord/issues/63)
 
 ## 2.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 - Added graceful exit for library extraction logic in case of error
 - Refactor the CI by splitting the building of mirrord-agent in a separate job and caching the agent image for E2E tests.
 
+### Fixed
+- Fix typos
+
 ## 2.1.0
 ### Added
 - Prompt user to update if their version is outdated in the VS Code extension or CLI.
--- a/mirrord-cli/src/main.rs
+++ b/mirrord-cli/src/main.rs
@@ -120,7 +120,7 @@ fn exec(args: &ExecArgs) -> Result<()> {
     if args.accept_invalid_certificates {
         std::env::set_var("MIRRORD_ACCEPT_INVALID_CERTIFICATES", "true");
     }
-    let library_path = extract_library(None).unwrap();
+    let library_path = extract_library(None)?;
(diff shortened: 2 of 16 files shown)


aviramha (changes requested):
Thanks for your contribution!
I have left few notes in the code.
Feel free to ask for clarifications/guidance if needed.
Also, you are more than welcome to consult with us in our Discord server.

thedanvail:
@aviramha I tried to take all of your comments into consideration as I updated using `anyhow` to handle an invalid path. 
Additional meta-information:
1. This is my first open source PR, so I apologize if I'm not quite following best practices
2. I found this issue through the `This Week In Rust` call for collaboration - just so you know!

aviramha:
@thedanvail 
1. Great! Thanks for choosing us to be your first contribution. Beware, OSS contributing can be addictive!
2. Cool! Thanks for letting us know :)

mirrord-cli/src/main.rs
```diff
@@ -117,7 +120,7 @@ fn exec(args: &ExecArgs) -> Result<()> {
     if args.accept_invalid_certificates {
         std::env::set_var("MIRRORD_ACCEPT_INVALID_CERTIFICATES", "true");
     }
-    let library_path = extract_library(None);
+    let library_path = extract_library(None).unwrap();
```
aviramha, line 123: you can use `?` here instead of unwrap

mirrord-cli/src/main.rs
```diff
@@ -138,7 +141,7 @@ fn main() {
     match cli.commands {
         Commands::Exec(args) => exec(&args).unwrap(),
         Commands::Extract { path } => {
-            extract_library(Some(path));
+            extract_library(Some(path)).unwrap();
```
aviramha, line 144: Change fn main to return `Result<()>` then you can use `?` here

aviramha (changes requested):
The context can be great, but a few more tweaks are needed to merge this.
Thanks!

mirrord-cli/src/main.rs
thedanvail: If I change the main function signature to return `Result<()>`, then we'd have to remove the `unwrap()`s/`?` to return the `Result` - or we can leave in the `unwrap()`s/`?` and leave out `Result` from the return type - which behavior would be preferred?

aviramha: `?` is a syntatic sugar that returns when a `Result`/`Option` has `Err`/`None` else it returns the inner value.
So the change would to add `?` and change the return type to `Result`
https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html

thedanvail: I understand that, but the issue is that, if we leave in the `Result` return type, running cargo check gives me the following:

<a href="https://ibb.co/xFV8zKR">Screen-Shot-2022-06-03-at-2-27-25-PM</a>

So the types don't match up if we add `Result` alongside unwrapping. I just want to make sure I'm making the changes you want - I apologize if I'm being difficult!

thedanvail: I understand that, but the issue is, if I run `cargo check` with the following code

```
fn main() -> Result<()> {
    registry()
        .with(fmt::layer())
        .with(EnvFilter::from_default_env())
        .init();
    prompt_outdated_version();

    let cli = Cli::parse();
    match cli.commands {
        Commands::Exec(args) => exec(&args)?,
        Commands::Extract { path } => {
            extract_library(Some(path))?;
        }
    }
}
```

I receive the following error:

<a href="https://ibb.co/kHv9Y4L">Screen-Shot-2022-06-03-at-2-44-03-PM</a>

aviramha: Sorry but the image isn't available.. you can upload directly to GitHub by pasting the image here.
Considering the snippet, I think I know what the issue is. You can just drop the `?` in both arms and it'd work because both `exec` and `extract_library` return `Result`, so no need to unwrap it and you can just return their return value.

thedanvail: Added in your changes - since `extract_library` returns a `Result<String>` instead of `Result<()>`, I just return a normal `Ok(())`. I think that should cover everything!

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+### Changed
+- Added a graceful exit for system paths that do not exist.
```
aviramha, line 10: Reword - "Added graceful exit for library extraction logic in case of error"

mirrord-cli/src/main.rs
```diff
@@ -60,19 +60,22 @@ const INJECTION_ENV_VAR: &str = "LD_PRELOAD";
...
+fn extract_library(dest_dir: Option<String>) -> Result<String> {
     let library_file = env!("CARGO_CDYLIB_FILE_MIRRORD_LAYER");
     let library_path = std::path::Path::new(library_file);
     let file_name = library_path.components().last().unwrap();
     let file_path = match dest_dir {
         Some(dest_dir) => std::path::Path::new(&dest_dir).join(file_name),
         None => temp_dir().as_path().join(file_name),
     };
-    let mut file = File::create(file_path.clone()).unwrap();
+    let mut file = File::create(&file_path)
```
aviramha, line 71: We can't know for sure the reason it failed is due to path not existing, so I'd just add using context the file path ("Path .. creation failed")

mirrord-cli/src/main.rs
```diff
@@ -136,9 +139,10 @@ fn main() {
 
     let cli = Cli::parse();
     match cli.commands {
-        Commands::Exec(args) => exec(&args).unwrap(),
+        Commands::Exec(args) => exec(&args),
         Commands::Extract { path } => {
             extract_library(Some(path));
```
aviramha, line 144: I think you would have warning here if you wont add ?

aviramha (changes requested):
Few small tweaks left :)
Sorry for being hard, but we want to make sure everything is great.
Also please fix conflicts with the main branch so we GH can run our automated tests.

mirrord-cli/src/main.rs
thedanvail: Nay - the function returns `Result<()>`, and because we added that to the `main() -> Result<()>` as well, this compiles as is. If I add the `?` operator - `Commands::Exec(args) => exec(&args)?,`, `cargo check` returns the following:

```
error[E0308]: `?` operator has incompatible types
   --> mirrord-cli/src/main.rs:142:33
    |
142 |         Commands::Exec(args) => exec(&args)?,
    |                                 ^^^^^^^^^^^^ expected enum `Result`, found `()`
    |
    = note: `?` operator cannot convert from `()` to `Result<(), anyhow::Error>`
    = note:   expected enum `Result<(), anyhow::Error>`
            found unit type `()`
help: try removing this `?`
    |
142 -         Commands::Exec(args) => exec(&args)?,
142 +         Commands::Exec(args) => exec(&args),
    | 
help: try wrapping the expression in `Ok`
    |
142 |         Commands::Exec(args) => Ok(exec(&args)?),
    |                                 +++            +

For more information about this error, try `rustc --explain E0308`.
error: could not compile `mirrord` due to previous error
```

thedanvail:
> Few small tweaks left :) Sorry for being hard, but we want to make sure everything is great. Also please fix conflicts with the main branch so we GH can run our automated tests.

No, please be as hard as you'd like! I'm trying to take as many notes as I can to get better at Rust and so I can be better at my next OSS PR. I sincerely appreciate your patience with me during this process!

I believe I have updated to reflect your suggestions and updated to fix the merge conflicts. I'm currently between a couple flights, so I didn't have time to check the status before the plane took off, but should be good now!

mirrord-cli/src/main.rs
aviramha: I think it'd be best to have `?` in the end of all arm branches (on exec + Extract) then outside the match branch (in the end of function) just to have `Ok(())` which means that if it got there, we're all good.

aviramha:
> > Few small tweaks left :) Sorry for being hard, but we want to make sure everything is great. Also please fix conflicts with the main branch so we GH can run our automated tests.
> 
> No, please be as hard as you'd like! I'm trying to take as many notes as I can to get better at Rust and so I can be better at my next OSS PR. I sincerely appreciate your patience with me during this process!
> 
> I believe I have updated to reflect your suggestions and updated to fix the merge conflicts. I'm currently between a couple flights, so I didn't have time to check the status before the plane took off, but should be good now!

Awesome, just see my last comment in the conversation about the `?`.

thedanvail:
Done and done!

thedanvail:
(Just updated with cargo fmt)

thedanvail:
I looked at the last CI run, and it looks like the `cargo clippy` issues are in `mirrord-layer`, which I did not touch in this PR. Let me know if I should do anything further.

aviramha:
Yes you're right. We're not binding clippy to a specific version so such cases can happen :)
Thanks for the contribution!

PR #103, opened 2022-06-01
File pr v3

meowjesty:
`openat` implementation missing.

Some of the changes in -agent come from @infiniteregrets branch (might be a bit outdated by now). Will update those as well after getting some review on  file + refactors I made.

Issue #101, opened 2022-05-31
Traffic steal


aviramha:
Allow user to "steal" incoming traffic from the remote pod.
First version of this feature will steal all incoming traffic.

aviramha:
I investigated how linkerd works as it's a good model on what we want to achieve - when we intercept connections we want to detect and understand if we want to "steal" it and later on we want to steal only specific requests based on filter.
Initially, I believed we could just use components from linkerd2-proxy codebase but I realized it was too coupled for its use case, so now my goal is to implement our own stuff using hyper.

eyalb181:
"Detect and understand if we want to "steal" it" - based on what?

PR #113, opened 2022-06-03
Refactor traffic

aviramha:
While working on #101 I started having too much duplicate code and non-sense between the mirroring API and stealing API and decided to refactor this so following the code and changing it will be easier.
The part that really messed me up was the layer, so I've converted most of the logic to a trait with "default" impl and a struct that does implementation for mirroring logic. 
I suggest this should be merged after #103 as @meowjesty dealt with too many rebases/merges and I'm a nice fella :)

Please review the API and overall logic, I think we should have the same for #103 code later on, and probably do the same in the agent side. Maybe we can split the work (I'll do layer, someone else do agent?)


mirrord-layer/src/tcp_mirror.rs
```diff
@@ -0,0 +1,170 @@
...
+                            Err(err) => {error!("writing to local stream err {err:?}"); break;}
+                        }
+                    },
+                    None => {
+                        debug!("tcp tunnel exiting due to remote stream closed");
+                        break;
+                    }
+                }
+            },
+            res = local_stream.read(&mut buffer) => {
```
eyalb181, line 41: Add documentation, something like: "Read the application's response from the socket and discard the data, so that the socket doesn't fill up."

mirrord-layer/src/tcp_mirror.rs
```diff
@@ -0,0 +1,170 @@
...
+impl TCPHandler for TCPMirrorHandler {
+    fn new() -> TCPMirrorHandler {
+        Self {
+            ports: HashSet::new(),
+            running: true,
+            connections: HashSet::new(),
+        }
+    }
+
+    fn is_running(&mut self) -> bool {
```
eyalb181, line 115: Why aren't the x_running functions implemented in the parent class?

eyalb181: Actually most of the other functions too. Shouldn't the only mirror-specific code be in tcp_tunnel?

eyalb181 (changes requested):

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,179 @@
...
+        self.send(TrafficIn::Data(data)).await
+    }
+
+    pub async fn tcp_close(&self, close: TCPClose) -> Result<()> {
+        self.send(TrafficIn::Close(close)).await
+    }
+}
+
+#[async_trait]
+pub trait TCPHandler {
```
meowjesty, line 81: Hmm, why do we need this trait?

mirrord-layer/src/tcp_mirror.rs
```diff
@@ -0,0 +1,170 @@
...
+impl Borrow<ConnectionID> for Connection {
+    fn borrow(&self) -> &ConnectionID {
+        &self.id
+    }
+}
+
+/// Handles traffic mirroring
+pub struct TCPMirrorHandler {
+    ports: HashSet<Listen>,
+    running: bool,
```
meowjesty, line 101: Not a big fan of holding `bool`s to represent On/Off states. Can't you "flip it around" and have some sort of `RunningTCPMirrorHandler` and a `StoppedTCPMirrorHandler` (if the stopped version even makes sense, like why do we care about a non-running thing)?

meowjesty: I would suggest turning this struct into a sort of FSM, and implement message handling (just a simple way of handling `event -> next state`, to decouple stateful mutations and have these look more like state transitions instead) to change states between stopped/running/whatever else.

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Create new TCP handler communicating using given TCPConfig
+    fn new() -> Self;
```
meowjesty, line 83: `new` with no arguments should be just a trait constraint on `Default`.

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// Should the run loop keep running
+    fn is_running(&mut self) -> bool;
+
+    /// Changes the state so is_running will return False
+    fn stop_running(&mut self);
+
+    fn ports(&mut self) -> &HashSet<Listen>;
+    fn ports_mut(&mut self) -> &mut HashSet<Listen>;
+
```
meowjesty, line 106: These functions look an awlful lot like `Deref` and `DerefMut` for the type.

Also, `fn ports` should be just `&self`, instead of `&mut self`.

meowjesty (commented):
Some initial comments on the refactor.

mirrord-layer/src/tcp.rs
aviramha: The idea is that we'll have some default implementation and then it'll be shared behavior between traffic mirror & steal

mirrord-layer/src/tcp_mirror.rs
aviramha: There's no use for Running/Stopped, just a way to share behavior between TCPHandler implementations. I can remove that by moving the loop to each implementatino

mirrord-layer/src/tcp.rs
aviramha: That's here to have shared behavior between mirror & steal

mirrord-layer/src/tcp.rs
meowjesty: Can't you "extract" shared behavior to an outer structure? An example of what I mean is:

```rust
struct TCPHandler {
    ports: HashSet<Listen>,
    different_behavior_part: TCPMode,
}

enum TCPMode {
    Mirror(...),
    Steal(....),
}
```

aviramha: But we'd need a shared interface for Mirror/Steal structs, which is TCPHandler right now?

eyalb181 (approved):

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+#[async_trait]
+pub trait TCPHandler {
+    fn ports(&mut self) -> &HashSet<Listen>;
+    fn ports_mut(&mut self) -> &mut HashSet<Listen>;
+
+    /// Returns true to let caller know to keep running
+    async fn handle_incoming_message(&mut self, msg: Option<TrafficHandlerInput>) -> Result<bool>
```
meowjesty, line 76: Function signature looks a bit weird to me:

1. Why not take the `Traffic<_>` here, instead of `Option<Traffic<_>>`?;
2. Why `Result<bool>` and not `Result<()>` and check `result.is_ok()` on the caller side?

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Handle NewConnection messages
+    async fn handle_new_connection(&mut self, conn: NewTCPConnection) -> Result<()>;
+
+    /// Connects to the local listening socket, add it to the queue and return the stream.
+    /// Find better name
+    async fn create_local_stream(&mut self, conn: &NewTCPConnection) -> Option<TcpStream> {
+        let listen_data = self.ports().get(&conn.destination_port)?;
+        let addr = match listen_data.ipv6 {
```
meowjesty, line 101: `match` on `bool` should just be `if else`.

mirrord-layer/src/tcp.rs
aviramha: Didn't understand 1 btw
We need to keep it running even after an error. Any suggestion how to mark error ignorable?

mirrord-layer/src/tcp.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Handle New Data messages
+    async fn handle_new_data(&mut self, data: TCPData) -> Result<()>;
+
+    /// Handle connection close
+    async fn handle_close(&mut self, close: TCPClose) -> Result<()>;
+
+    /// Handle listen request
+    async fn handle_listen(&mut self, listen: Listen) -> Result<()> {
```
meowjesty, line 126: Why return `Result` here if the inner operation can't fail?

Also, do we need this function at all, as it's basically a one-liner?

mirrord-layer/src/tcp_mirror.rs
```diff
@@ -0,0 +1,186 @@
...
+                },
+            }
+        }
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl TCPHandler for TCPMirrorHandler {
+    async fn handle_listen(&mut self, listen: Listen) -> Result<()> {
```
meowjesty, line 130: Another one-liner that can't error, no?

mirrord-layer/src/tcp.rs
meowjesty: 1. Leave the function caller responsible for checking if it's an option or not, and then it passes down just the thing, instead of `Option` thing.

Make a `LayerError::KeepRunning` or something like that, and check for it.

mirrord-layer/Cargo.toml
```diff
@@ -44,7 +44,8 @@ multi-map = "1.3.0"
 envconfig = "0.10.0"
 regex = "1.5"
 errno = "0.2.8"
-
+async-trait = "0.1.53"
+anyhow.workspace = true
```
meowjesty, line 48: 😢

mirrord-layer/src/tcp.rs
aviramha: `LayerError::KeepRunning` would be 1 error but if we have many different errors that we wish to ignore, you have a suggestion for a nicer approach?

mirrord-layer/src/tcp.rs
aviramha: welp the steal might have different logic? also that's why it's Result

mirrord-layer/src/tcp.rs
meowjesty: `map_err` for all ignorable errors? Not sure.

mirrord-layer/src/tcp.rs
meowjesty: Hmm, but if they can return in different ways, should they still be treated the same?

If A can never error, but B can, adding a function that displays something like "Hey, A may fail", is not very nice. I prefer that the function signature matches what's inside.

aviramha: What about using Infallible then?

meowjesty: Hmm, sure, but I'm still not quite certain that things with such potential different behavior should be shared. Feels a bit to much OOP.

mirrord-layer/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 #![feature(once_cell)]
 #![feature(result_option_inspect)]
 #![feature(const_trait_impl)]
+#![feature(type_alias_impl_trait)]
```
eyalb181, line 5: Looks unused?

eyalb181 (approved):

aviramha (commented):
LGTM

PR #115, opened 2022-06-05
Add detours for fcntl/dup system calls

### LINKED ISSUE #51: Add detours for dup/dup2/fcntl system calls
When mirrord exec is called with a simple flask API (https://gist.github.com/infiniteregrets/ba753ffb53ff8b98b2dd8f5051323919), where port 80/any other port is explicitly specified, the application seems to ignore that and instead assign a different port. This could perhaps be due to a call to `dup` or `fcntl` with `F_DUPFD/F_DUPFD_CLOEXEC` flags where mirrord is not aware of the new sockfd and probably getsockname returns a different port to the flask app. So we might need to add detours for the mentioned system calls to bring new sockfds to mirrord book-keeping. 
More context:
Strace of flask API: https://gist.github.com/infiniteregrets/09773cc3f1aee7983f95fc800ff8d969
Agent logs when enforcing port `3689` in the Flask Application: https://gist.github.com/infiniteregrets/7c42aeae359e3a55c756f2ba9c7e1213

infiniteregrets:
Closes #51. Added detours for fcntl/dup/dup2/dup3 system calls such that when either are called by a framework which needs a duplicate file descriptor, we can record that in our socket hashmap.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 - Prompt user to update if their version is outdated in the VS Code extension or CLI.
 - Add support for docker runtime, closes [#95](https://github.com/metalbear-co/mirrord/issues/95).
 - Add a keep-alive to keep the agent-pod from exiting, closes [#63](https://github.com/metalbear-co/mirrord/issues/63)
+- Add detours for fcntl/dup system calls, closes [#51](https://github.com/metalbear-co/mirrord/issues/51)
```
aviramha, line 17: Wrong place in the changelog :)

mirrord-layer/src/sockets.rs
```diff
@@ -193,12 +193,12 @@ fn bind(sockfd: c_int, addr: *const sockaddr, addrlen: socklen_t) -> c_int {
         return unsafe { libc::bind(sockfd, addr, addrlen) };
     }
 
-    socket.state = SocketState::Bound(Bound {
+    Arc::get_mut(&mut socket).unwrap().state = SocketState::Bound(Bound {
```
aviramha, line 196: `socket.get_mut` instead of calling the fully qualified name

mirrord-layer/src/sockets.rs
```diff
@@ -548,19 +541,90 @@ unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
...
+    }
+}
+
+unsafe extern "C" fn dup_detour(fd: c_int) -> c_int {
+    let dup_fd = libc::dup(fd);
+    if dup_fd == -1 {
+        error!("dup failed");
+        return dup_fd;
+    }
+    let mut sockets = SOCKETS.lock().unwrap();
```
aviramha, line 569: please move this part to a safe func

mirrord-layer/src/sockets.rs
```diff
@@ -548,19 +541,90 @@ unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
...
+    match cmd {
+        libc::F_DUPFD | libc::F_DUPFD_CLOEXEC => {
+            let mut sockets = SOCKETS.lock().unwrap();
+            if let Some(socket) = sockets.get(&fd) {
+                let fcntl_socket = socket.clone();
+                sockets.insert(fcntl_fd as RawFd, fcntl_socket);
+            }
+            fcntl_fd
+        }
+        _ => fcntl_fd,
```
aviramha, line 559: move `fcntl_fd` to the end of the function instead to avoid repeating.

mirrord-layer/src/sockets.rs
```diff
@@ -548,19 +541,90 @@ unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
     libc::close(fd)
 }
 
+unsafe extern "C" fn fcntl_detour(fd: c_int, cmd: c_int, arg: ...) -> c_int {
+    let fcntl_fd = libc::fcntl(fd, cmd, arg);
+    if fcntl_fd < 0 {
+        error!("fcntl failed");
+        return fcntl_fd;
+    }
+    match cmd {
```
aviramha, line 550: move this to safe block

aviramha (changes requested):
Awesome job Mehul! I have left few comments inline :)

mirrord-layer/src/sockets.rs
infiniteregrets: the get_mut method is only available as an associated function, so we have to use it like so. Otherwise I can wrap the socket in a Pin and then use it

CHANGELOG.md
aviramha: It should go to added, but you left it under a released version.

aviramha (approved):
LGTM! Now CI has to pass ;)

mirrord-layer/src/sockets.rs
```diff
@@ -533,17 +505,105 @@ unsafe extern "C" fn accept4_detour(
...
+}
+
+unsafe extern "C" fn dup2_detour(oldfd: c_int, newfd: c_int) -> c_int {
+    dup2(oldfd, newfd)
+}
+
+#[cfg(target_os = "linux")]
+fn dup3(oldfd: c_int, newfd: c_int, flags: c_int) -> c_int {
+    let dup3_fd = unsafe { libc::dup3(oldfd, newfd, flags) };
+    if dup3_fd == -1 {
```
eyalb181, line 574: The implementations for fcntl, dup, dup2, and dup3 are pretty much the same - any way to extract them to a common function?

eyalb181 (changes requested):

eyalb181 (approved):

PR #118, opened 2022-06-06
New bug template

### LINKED ISSUE #74: Revise issue templates
Right now the Bugfix template doesn't match the new design of mirrord (asks for node version) so we need to think what needs to be collected in case of a bug.
I'd assume:
1. distro/os
2. binary being tried to run
3. version of binary
4. hash of it?

instructions to run with debug and print trace?
Feel free to add more stuff a bug report (or other issue templates) should cover.

eyalb181:
Changed bug template to apply to the latest version of mirrord. Closes #74

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -11,7 +11,7 @@ body:
     id: description
     attributes:
       label: Bug Description
-      description: Please provide a clear and concise description of what the bug is.
+      description: Please provide a clear description of the bug.
       placeholder: The bug is...
     validations:
       required: true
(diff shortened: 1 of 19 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -32,19 +32,30 @@ body:
...
     validations:
       required: true
   - type: input
-    id: node_version
+    id: binary_type
     attributes:
-      label: Your Node version (`node --version`)
-      placeholder: ex. v16.13.1
+      label: Local process
+      description: What process are you trying to run with mirrord?
```
aviramha, line 51: Add "please run `file $(which PROCESSNAME)`" as it's very important for debugging

aviramha (approved):
Great, Thanks!

PR #122, opened 2022-06-07
Support running emulated processes on m1 using mirrord

aviramha:
This will download and use x64 binary when needed.

### DIFF
--- a/mirrord-cli/src/main.rs
+++ b/mirrord-cli/src/main.rs
@@ -1,4 +1,4 @@
-use std::{env::temp_dir, fs::File, io::Write, time::Duration};
+use std::{env::temp_dir, fs::File, io::Write, path::PathBuf, time::Duration};
 
 use anyhow::{anyhow, Context, Result};
 use clap::{Args, Parser, Subcommand};
@@ -69,12 +69,13 @@ const INJECTION_ENV_VAR: &str = "DYLD_INSERT_LIBRARIES";
 mod mac_aarch {
     use std::{
         io::{Cursor, Read},
-        path::Path,
+        path::{Path, PathBuf},
     };
-    use tracing::warn;
+
     use mach_object::{OFile, CPU_TYPE_X86_64};
     use reqwest;
     use search_path::SearchPath;
+    use tracing::warn;
 
     use super::*;
     pub fn is_binary_different_arch(binary_path: &String) -> bool {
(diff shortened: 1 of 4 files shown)


mirrord-cli/src/main.rs
```diff
@@ -64,6 +64,83 @@ const INJECTION_ENV_VAR: &str = "LD_PRELOAD";
...
+    };
+
+    use mach_object::{OFile, CPU_TYPE_X86_64};
+    use reqwest;
+    use search_path::SearchPath;
+
+    use super::*;
+    pub fn is_binary_different_arch(binary_path: &String) -> bool {
+        let search_path = SearchPath::new("PATH").unwrap();
+        let binary_path = match search_path.find(&Path::new(binary_path)) {
```
meowjesty, line 82: I would suggest combining these operations by using `and_then` and `map`, instead of having separate `match`es.

mirrord-cli/src/main.rs
```diff
@@ -64,6 +64,83 @@ const INJECTION_ENV_VAR: &str = "LD_PRELOAD";
...
+                false
+            }
+            Err(err) => {
+                warn!("Could not open binary: {err:?}");
+                false
+            }
+        }
+    }
+
+    pub fn extract_intel_binary() -> Result<String> {
```
meowjesty, line 121: Rename it to `extract_or_create_intel_binary` to make it a bit clearer what might happen here.

mirrord-cli/src/main.rs
```diff
@@ -64,6 +64,83 @@ const INJECTION_ENV_VAR: &str = "LD_PRELOAD";
...
+        }
+        info!("download x86 dylib for rosetta mode");
+        let download_url = format!("https://github.com/metalbear-co/mirrord/releases/download/{}/libmirrord_layer_mac_x86_64.dylib", env!("CARGO_PKG_VERSION"));
+        let mut response =
+            reqwest::blocking::get(download_url).context("Could not download intel binary")?;
+        let mut file = File::create(&file_path).context("Could not create intel binary")?;
+        response
+            .copy_to(&mut file)
+            .context("Could not write intel binary")?;
+        Ok(file_path.to_str().unwrap().to_string())
```
meowjesty, line 137: Why can't you return the path as `PathBuf`?

meowjesty (commented):
Some suggestions.

mirrord-cli/src/main.rs
aviramha: but then I can't add appropriate logs?

meowjesty: Think you still can do it.

aviramha: erm I'm still against it because as far as I see it, `and_then` and `map` are to avoid nesting, but if I'm just doing match -> return the code is "nicer" and has less nesting.

eyalb181 (approved):

mirrord-cli/src/main.rs
meowjesty: Nope, these types of functions are not to avoid nesting, they exist to let you work in a "monadic context" (sorta) for as long as possible. What you're doing here is very similar to the `match => None` anti-pattern actually, I'm surprised clippy doesn't complain. [See here](https://github.com/rust-lang/rust-clippy/issues/6).

When you're dealing with mutation, `match`ing`, `for` loops and the other more imperative forms are better, but this is not case here.

messense:
I'd suggest build [universal2](https://en.wikipedia.org/wiki/Universal_binary#Universal_2) binaries instead, that way it will work on both Intel and Apple Silicon Macs automatically and when running in emulated mode on Apple Silicon it just works™️.

aviramha:
> I'd suggest build [universal2](https://en.wikipedia.org/wiki/Universal_binary#Universal_2) binaries instead, that way it will work on both Intel and Apple Silicon Macs automatically and when running in emulated mode on Apple Silicon it just works™️.

That was me initial thought but couldn't get cargo to produce such binaries. Do you know how?

messense:
@aviramha Opened #140

PR #134, opened 2022-06-09
Override remote env #25

### LINKED ISSUE #25: Use remote service environment variables
When a process is ran using mirrord, we want to be able to set it's environment variables to mirror the remote service one's.

### Proposed UX

1. Configuration called REMOTE_ENV - boolean, default true.
2. OVERRIDE_ENV - list of key-(optional)value of environment variables. if value is not provided, use the "original" environment variable.

### Implementation

There are various ways we can implement it - as part of taking care of this issue, those ways should be considered (feel free to suggest more)

#### K8S API
Obtain environment variables via K8S API (as part of creating the agent).

Pros:
1. Doesn't need any change in the agent + agent protocol.
Cons:
1. if environment is changed post-execution or not via k8s-aware ways, we won't be able to track it.

#### Agent API

Obtain the environment variables via adding such api in the agent protocol.

Pros:
1. Environment variables will reflect better - this is exactly how the env is being set in the remote process
2. if env is changed during runtime, we can get updates (needs finer implementation, but this creates a path)

Cons:
?

I think Agent API will be better, but when we start working on this we should get to a decision.

meowjesty:
Added an env variable for mirrord (`MIRRORD_OVERRIDE_ENV`) that enables overriding environment variables with the remote ones.

Closes #25


mirrord-agent/src/main.rs
```diff
@@ -104,13 +108,79 @@ pub struct PeerMessage {
...
+    environ_path: PathBuf,
+    override_env_vars: HashMap<String, String>,
+) -> Result<HashMap<String, String>, ResponseError> {
+    debug!(
+        "select_env_vars -> environ_path {:#?} override_env_vars {:#?}",
+        environ_path, override_env_vars
+    );
+
+    let mut environ_file = tokio::fs::File::open(environ_path).await.map_err(|fail| {
+        ResponseError::FileOperation(FileError {
```
aviramha, line 125: erm leveraging the file operation errors seems a bit shady?

aviramha (changes requested):

mirrord-agent/src/main.rs
meowjesty: But this fails in a file operation though.

mirrord-layer/src/lib.rs
```diff
@@ -314,6 +320,22 @@ async fn handle_hook_message(
                 codec_result
             );
         }
+        HookMessage::OverrideEnvVarsHook(OverrideEnvVarsHook { override_env_vars }) => {
```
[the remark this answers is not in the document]
meowjesty: Not getting an unused warning, what do you mean?

CHANGELOG.md
```diff
@@ -7,13 +7,18 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+- Override the process environment variables with the remote equivalents by setting `MIRRORD_OVERRIDE_ENV`, like `MIRRORD_OVERRIDE_ENV="DB=fakedb;CONFIG=fake.config"`;
```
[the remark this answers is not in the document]
eyalb181: We're assuming they'd want to override all for 80% of the cases, but maybe let the same variable accept different values for all, whitelist, blacklist?
So e.g.:
MIRRORD_OVERRIDE_ENV: true
or
MIRRORD_OVERRIDE_ENV: PATH, HOME
or
MIRRORD_OVERRIDE_ENV: !(PATH, HOME)

mirrord-layer/src/lib.rs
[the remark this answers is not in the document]
meowjesty: Removed.

CHANGELOG.md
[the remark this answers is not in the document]
meowjesty: It now loads all env vars, and the user specifies a list to filter out (reverse of what it was doing).

mirrord-protocol/src/lib.rs
```diff
@@ -1,9 +1,30 @@
 #![feature(const_trait_impl)]
 #![feature(io_error_more)]
 #![feature(core_ffi_c)]
+#![feature(hash_drain_filter)]
```
eyalb181, line 4: Where are you using this?

eyalb181: Same question re OnceCell

meowjesty: Good catch, was using this while experimenting, and missed removing in the clean up.

CHANGELOG.md
```diff
@@ -9,13 +9,18 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 ### Changed
 - E2E small improvements, removing sleeps. (Fixes #99)
```
aviramha, line 11: Update this

mirrord-agent/src/main.rs
aviramha: Indeed but layer sent a env request but it's getting a file operation error.. we can discuss whether errors/request reflect lower level implementation or high level logic. IMO it's high level (i.e layer doesn't care what failed exactly, only that the request it had sent has failed)
but it doesn't matter too much :)

mirrord-agent/src/main.rs
```diff
@@ -160,6 +231,33 @@ async fn handle_peer_messages(
                 .send((peer_message.peer_id, file_request))
                 .await?;
         }
+        ClientMessage::OverrideEnvVarsRequest(OverrideEnvVarsRequest { filter_env_vars }) => {
```
aviramha, line 234: I think the name `OverrideEnvVarsRequest` is too "use case specific" while what really happens here is layer asking the agent for env vars, so I'd suggest to rename it to `GetEnvVars`

mirrord-agent/src/main.rs
```diff
@@ -160,6 +231,33 @@ async fn handle_peer_messages(
...
+            debug!(
+                "ClientMessage::OverrideEnvVarsRequest peer id {:?} requesting {:?}",
+                peer_message.peer_id, filter_env_vars
+            );
+
+            let container_runtime = container_runtime
```
aviramha, line 240: maybe start the abstraction with this PR and create a `EnvVarHandler` or smth similar that holds all the needed stuff (container id, runtime, etc?) and just handles the message directly (to reduce logic congestion here :P)
not top priority, but it's a refactor we would like to have soon anyway

aviramha (commented):
Looking great, I left few comments more about design/naming ;)

mirrord-layer/src/config.rs
```diff
@@ -25,4 +25,11 @@ pub struct Config {
 
     #[envconfig(from = "MIRRORD_FILE_OPS", default = "false")]
     pub enabled_file_ops: bool,
+
+    #[envconfig(from = "MIRRORD_OVERRIDE_ENV_VARS", default = "false")]
+    pub enabled_override_env_vars: bool,
+
+    /// Overrides local env variables with remote ones, except for those specified here.
+    #[envconfig(from = "MIRRORD_OVERRIDE_FILTER_ENV_VARS", default = "PATH;HOME")]
```
aviramha, line 33: tbh I'm not sure if when using the env var we want them to explicitly add those, I'd imagine that even if the env/cli argument is used the user wouldn't expect to need to specify those as it was "magically" added before.
I'd add PATH;HOME regardless of the env var, and maybe later on add a flag to remove those explicitly (can't imagine someone wanting the remote PATH?)

CHANGELOG.md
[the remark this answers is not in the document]
eyalb181: I don't think we have a reason to believe an "exclude list" use case would be more common than an "include list" use case, I think we might as well support both.
btw I support using the same env var for all of them (since you can't have override_env = false && filter = [x, y, z]).

mirrord-layer/src/lib.rs
```diff
@@ -139,7 +140,18 @@ fn init() {
     let enabled_file_ops = ENABLED_FILE_OPS.get_or_init(|| config.enabled_file_ops);
     enable_hooks(*enabled_file_ops);
 
-    RUNTIME.spawn(poll_agent(port_forwarder, receiver));
+    let _ = ENABLED_OVERRIDE_ENV_VARS.get_or_init(|| config.enabled_override_env_vars);
```
eyalb181, line 143: Why are you using _ here?

mirrord-layer/src/lib.rs
```diff
@@ -480,6 +527,20 @@ async fn poll_agent(mut pf: Portforwarder, mut receiver: Receiver<HookMessage>)
     let close_file_handler = Mutex::new(Vec::with_capacity(4));
 
     let mut ping = false;
+
+    if ENABLED_OVERRIDE_ENV_VARS.get().is_some() {
```
eyalb181, line 531: Isn't it enough that you've received the filter_env_vars argument?

eyalb181: Alternatively, since you're accessing the configuration here anyway, why not initialise the whole thing here and forgo the filter_env_vars param entirely?

eyalb181 (changes requested):

mirrord-layer/src/config.rs
meowjesty: Yeah, that's why I started out doing it the other way around (user has to specify what they want). 

I'll add these in some sort of default ignore when the env vars are loaded, and if someone ever asks to have more control we can have some other config like `OVERRIDE_EVERYTHING`.

mirrord-agent/src/main.rs
meowjesty: Yeah both approaches make total sense, and it is indeed a discussion between if we want layer to have a lower or higher level view of what happens in agent.

I would vote for lower level, but I'm game for either. 

So how should I proceed?

mirrord-agent/src/main.rs
meowjesty: `GetEnvVars` makes sense, I've tried out so many bad names for it.

mirrord-agent/src/main.rs
aviramha: I think calling it "EnvGetError" or a like is better, but wouldn't waste too much time on it atm.

mirrord-layer/src/lib.rs
```diff
@@ -449,12 +441,44 @@ async fn handle_daemon_message(
...
+                Ok(remote_env_vars) => {
+                    for (key, value) in remote_env_vars.into_iter() {
+                        unsafe {
+                            debug!(
+                                "DaemonMessage::GetEnvVarsResponse set key {:#?} value {:#?}",
+                                key, value
+                            );
+
+                            let setenv_result =
+                                libc::setenv(key.as_ptr().cast(), value.as_ptr().cast(), 1);
```
aviramha, line 457: https://doc.rust-lang.org/std/env/fn.set_var.html ?

mirrord-agent/src/main.rs
```diff
@@ -104,13 +107,81 @@ pub struct PeerMessage {
...
+
+    let mut environ_file = tokio::fs::File::open(environ_path).await.map_err(|fail| {
+        ResponseError::FileOperation(FileError {
+            operation: "open".to_string(),
+            raw_os_error: fail.raw_os_error(),
+            kind: fail.kind().into(),
+        })
+    })?;
+
+    let mut raw_env_vars = String::with_capacity(255);
```
aviramha, line 131: I think 255 is too small

aviramha (changes requested):

aviramha (approved):
LGTM

CHANGELOG.md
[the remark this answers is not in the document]
meowjesty: I think we support both now, and I've added a panic if exclude/include share values.

.vscode/launch.json
```diff
@@ -52,8 +58,12 @@
         "DYLD_INSERT_LIBRARIES": "${workspaceFolder}/target/debug/libmirrord_layer.dylib",
         "LD_PRELOAD": "${workspaceFolder}/target/debug/libmirrord_layer.so",
         "RUST_LOG": "warn,mirrord=debug",
-        "MIRRORD_AGENT_IMPERSONATED_POD_NAME": "",
+        "MIRRORD_AGENT_IMPERSONATED_POD_NAME": "apache-deployment-8477b45749-bs9qj",
+        "MIRRORD_AGENT_IMAGE": "meowjesty/mirrord-agent:latest",
```
aviramha, line 62: :)

.vscode/launch.json
```diff
@@ -33,8 +33,14 @@
...
```
aviramha, line 37: :))

tests/src/utils.rs
```diff
@@ -299,7 +299,8 @@ pub async fn test_server_init(
     // used by the CI, to load the image locally:
     // docker build -t test . -f mirrord-agent/Dockerfile
     // minikube load image test:latest
-    env.insert("MIRRORD_AGENT_IMAGE", "test");
+    // env.insert("MIRRORD_AGENT_IMAGE", "test");
+    env.insert("MIRRORD_AGENT_IMAGE", "meowjesty/mirrord-agent:latest");
```
aviramha, line 303: :)))

aviramha (approved):
Seems to be ready once CI is working!

mirrord-layer/src/lib.rs
```diff
@@ -375,6 +392,34 @@ async fn poll_agent(mut pf: Portforwarder, mut receiver: Receiver<HookMessage>)
 
     let mut ping = false;
 
+    if config.enabled_override_env_vars {
+        let env_vars_filter = HashSet::from(EnvVarsFilter(config.override_env_vars_exclude));
+        let env_vars_select = HashSet::from(EnvVarsFilter(config.override_env_vars_include));
+
+        let env_vars_intersection = env_vars_filter.intersection(&env_vars_select);
```
eyalb181, line 399: You shouldn't bother with the intersection - it doesn't make any sense for both INCLUDE and EXCLUDE to be populated at all at the same time, even if the values don't intersect. e.g. INCLUDE=a,b,c, EXCLUDE=x,y,z - if values a,b,c are included, then of course values x,y,z are excluded, there's no need to specify.

eyalb181: Also, there's no need for enabled_override_env_vars to be required for them to take effect - if I specified INCLUDE or EXCLUDE then naturally I want env vars to be enabled.

eyalb181 (changes requested):

eyalb181 (approved):

mirrord-agent/src/main.rs
meowjesty: As this PR took a bit too long, I'll leave this up as a TODO for now.

PR #139, opened 2022-06-10
[WIP] Fix openat relative paths + detours to support ls

### LINKED ISSUE #120: Support ls
Our files feature currently support most of the relevant syscalls, but we still need to add support for ls.

After #904, this is what is left to be done (prioritized)

- [x]  Go Linux x86-64 (same flow/integration test should work~)
- [x]  Python list directory
- [x]  Node list directory
- [ ] `ls` binary - macOS
- [ ] `ls` binary - Linux
- [ ]  All of the libc functions:
  - [X] opendir
  - [X] fdopendir
  - [X] readdir
  - [X] readdir_r
  - [ ] telldir
  - [ ] seekdir
  - [ ] rewinddir
  - [X] closedir
  - [x] dirfd

I think the best way would be to drop each as separate PRs, first few tasks will resolve most of the libc functions :)

infiniteregrets:
Closes #120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,15 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 ### Added
-- Add support for openat, closes [#120](https://github.com/metalbear-co/mirrord/issues/120)
+- Add support for openat, opendir, closedir, readdir, telldir, closes [#120](https://github.com/metalbear-co/mirrord/issues/120)
 
 ### Changed
+- Speed up agent container image building by using a more specific base image.
+- CI: Remove building agent before building & running tests (duplicate)
+
+## 2.2.1
+### Changed
+- Compile universal binaries for MacOS. (Fixes #131)
 - E2E small improvements, removing sleeps. (Fixes #99)
 
 ## 2.2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+### Added
+- Add support for openat, closes [#120](https://github.com/metalbear-co/mirrord/issues/120)
+
 ### Changed
 - E2E small improvements, removing sleeps. (Fixes #99)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+### Changed
+- E2E small improvements, removing sleeps. (Fixes #99)
 
 ## 2.2.0
 ### Added
--- a/mirrord-agent/src/file.rs
+++ b/mirrord-agent/src/file.rs
@@ -3,16 +3,17 @@ use std::{
     collections::HashMap,
     fs::{File, OpenOptions},
     io::{prelude::*, SeekFrom},
-    os::unix::prelude::OpenOptionsExt,
+    os::unix::prelude::{DirEntryExt, OpenOptionsExt},
     path::PathBuf,
 };
 
 use mirrord_protocol::{
-    CloseFileRequest, CloseFileResponse, FileError, FileRequest, FileResponse, OpenDirRequest,
-    OpenFileRequest, OpenFileResponse, OpenOptionsInternal, OpenRelativeFileRequest,
-    ReadFileRequest, ReadFileResponse, ResponseError, SeekFileRequest, SeekFileResponse,
-    WriteFileRequest, WriteFileResponse,
+    CloseFileRequest, CloseFileResponse, DirEntry, FileError, FileRequest, FileResponse,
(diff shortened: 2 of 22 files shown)


mirrord-protocol/src/codec.rs
```diff
@@ -187,6 +193,25 @@ pub enum FileResponse {
...
+    // Todo: Directory block and mutex are missing here. (we probably don't need them)
+    // Refer: https://code.woboq.org/userspace/glibc/sysdeps/posix/dirstream.h.html#__dirstream
+}
+
+pub struct CloseDirResponse {
+    pub errcode: i32,
+}
+
+pub struct ReadDirResponse {
+    pub stream: Vec<u8>,
```
aviramha, line 212: stream has a specific meaning in Rust ecosystem so that's a bad name imo, buffer/data would be better

mirrord-agent/src/file.rs
```diff
@@ -21,7 +21,7 @@ use crate::{error::AgentError, runtime::get_container_pid, sniffer::DEFAULT_RUNT
 // `HashMap<_, RemoteFile`>, where `RemoteFile` is a struct that holds both the `File` + `PathBuf`.
 #[derive(Debug, Default)]
 pub struct FileManager {
-    pub open_files: HashMap<i32, File>,
+    pub open_files: HashMap<i32, (File, PathBuf)>,
```
aviramha, line 24: Why the change?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+### Added
+- Add support for openat, closes [#120](https://github.com/metalbear-co/mirrord/issues/120)
```
eyalb181, line 10: Is it just openat? Because if so it doesn't close #120, which specifies ls as well. Maybe edit #120 to only specify openat, and open a separate issue for ls?

infiniteregrets: It will be for both, just a draft for now. Havent pushed all changes

mirrord-agent/src/file.rs
meowjesty: I would suggest changing this tuple into a `RemoteFile` struct:

```rust
struct RemoteFile {
    inner: std::fs::File,
    path: PathBuf,
}
```

And then you could implement many of the file functions in a similar fashion to Rust's `std` (or just keep the current implementation of `FileManager` and we do something better later).

mirrord-agent/src/file.rs
```diff
@@ -176,7 +180,55 @@ impl FileManager {
 
         debug!("FileManager::write -> Trying to close file {:#?}", file);
 
-        Ok(CloseFileResponse)
+        Ok(CloseFileResponse { result: 0 })
+    }
+
+    pub(crate) fn opendir(
```
meowjesty, line 186: Do we need this distinction? Shouldn't `opendir` be the same as `open`? Have you checked how Rust implements this?

mirrord-agent/src/file.rs
```diff
@@ -176,7 +180,55 @@ impl FileManager {
...
+            .map_err(|fail| {
+                ResponseError::FileOperation(FileError {
+                    operation: "open".to_string(),
+                    raw_os_error: fail.raw_os_error(),
+                    kind: fail.kind().into(),
+                })
+            })
+    }
+
+    pub(crate) fn readdir(&mut self, fd: i32) -> Result<ReadDirResponse, ResponseError> {
```
meowjesty, line 212: Same thing here, isn't a directory just a special case of a file?

mirrord-layer/src/file/hooks.rs
```diff
@@ -132,50 +135,37 @@ pub(super) unsafe extern "C" fn openat_detour(
...
-        // what).
-        let remote_fd = OPEN_FILES.lock().unwrap().get(&fd).cloned();
+        // Todo(check): AT_FDCWD - refers to the CWD of the calling process.
+        // Do we really need to manipulate paths here since we already have the path for the given
+        // fd in the agent? since AT_FDCWD refers to the context of the calling process, do
+        // we need to check if the remote fd even exists?
 
-        // Are we managing the relative part?
-        if let Some(remote_fd) = remote_fd {
+        if path.is_absolute() || fd == AT_FDCWD {
```
meowjesty, line 156: The docs on this handling are a bit confusing to me, but shouldn't these be handled differently?

> absolute path, it ignores the fd argument

> AT_FDCWD in the fd parameter, the current working directory is used

mirrord-layer/src/file/ops.rs
```diff
@@ -249,6 +251,102 @@ pub(crate) fn close(fd: RawFd) -> Result<c_int, LayerError> {
...
+
+    let OpenFileResponse { fd } = dir_channel_rx.blocking_recv()?;
+
+    let fake_local_dir_name = CString::new(fd.to_string())?;
+
+    // Todo: check if we need extra flags like O_DIRECTORY here
+
+    let local_dir_fd = unsafe {
+        let local_dir_fd = libc::shm_open(
+            fake_local_dir_name.as_ptr(),
```
meowjesty, line 286: This will cause the `*mut DIR` to be reported as not a directory. You can check for this on the sample side by looking at the file metadata (Rust has an `is_dir` method).

meowjesty (commented):
Leaving some initial comments.

PR #140, opened 2022-06-11
Create macOS universal2 binaries

messense:
Reference: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -111,7 +111,7 @@ jobs:
 
   release_gh:
     needs: [build_binaries, release_docker_image, release_vscode_ext]
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     permissions:
       packages: write
       contents: write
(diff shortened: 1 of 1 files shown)


eyalb181 (approved):
Thanks @messense!

aviramha:
Hey @messense, we figured out it doesn't work quite well for us atm as the CLI includes the dylib at compile time, meaning that arm binary will be used to execute x86 target, but it will have only arm binary in it. So we need to compile the dylib as a separate step or something else. If you have any suggestion we'd love to hear (currently looking into zig or just changing our CI/CD steps to enable the solution.)

messense:
Hmm, if I understand it correctly, it's because when running `mirrord exec node app.js -n my-app -p my-app-8vqpp`, mirrord is running in native arm64 mode while `node` is x86_64 node, right? I guess you could detect the architecture of node or other programs passed to mirrord then re-run mirrord with `arch -x86_64 mirrord`?

aviramha:
@messense That's quite interesting approach. Not sure if I like it better than manipulating our CI/CD. I was thinking of adding a new env var for compilation "MIRRORD_MIXED_DYLIB" that exists on arm64 compilation that has the location of a mixed dylib then it gets inserted into the library. Which approach do you think is better?

aviramha:
> Well, embedding a universal2 dylib is certainly better approach, but I am not sure whether Cargo bindeps feature supports it.

It doesn't, there's an issue for a RFC that would allow that: https://github.com/rust-lang/cargo/issues/10030
I'm thinking to overcome the issue by splitting the steps - compile the dylib as a step and then the cli as a different step, thus eliminating the need of the above RFC..

Issue #110, opened 2022-06-03
mirrord-layer panics without any relevant info if the agent exits 


infiniteregrets:
If mirrord-agent exits i.e. if layer reaches its timeout before the agent even started, mirrord-layer panics with an `Elapsed` error without giving any context what went wrong. It would be nice to remove the unwrap here: https://github.com/metalbear-co/mirrord/blob/main/mirrord-layer/src/pod_api.rs#L163 and handle the error gracefully providing good context and logs.

Issue #62, opened 2022-05-19
Support MIRRORD_AGENT_TTL as an argument in the CLI


eyalb181:
It's currently only usable as an environment variable.

thedanvail:
Planning on looking at this once #144 is merged.

Looking at it now, my approach will probably be to edit the `ExecArgs` struct to add in the `MIRRORD_AGENT_TTL` that will override an environment variable. If it's not in the CLI args, I'll default back to the environment setting or the default setting of 0.

aviramha:
Awesome @thedanvail. Assigning this to you :)

thedanvail:
Addressed in #156

aviramha:
Thank you!

PR #144, opened 2022-06-13
Better logging for Kube pod failures

thedanvail:
- Handles the error more gracefully if we receive from Kube's API when trying to spin up a pod
- Updated CHANGELOG with new changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 - CI: Remove building agent before building & running tests (duplicate)
 - CI: Add Docker cache to Docker build-push action to reduce build duration.
 - CD release: Fix universal binary for macOS
+- Refactor: Change protocol + mirrord-layer to split messages into modules, so main module only handles general messages, passing down to the appropriate module for handling.
 
 ## 2.2.1
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 - Speed up agent container image building by using a more specific base image.
 - CI: Remove building agent before building & running tests (duplicate)
 - CI: Add Docker cache to Docker build-push action to reduce build duration.
+- CD release: Fix universal binary for macOS
 
 ## 2.2.1
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Previous versions had CHANGELOG per component, we decided to combine all reposit
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+
+## 2.2.0
 ### Added
 - File operations are now available behind the `MIRRORD_FILE_OPS` env variable, this means that mirrord now hooks into the following file functions: `open`, `fopen`, `fdopen`, `openat`, `read`, `fread`, `fileno`, `lseek`, and `write` to provide a mirrored file system.
 - Support for running x64 (Intel) binary on arm (Silicon) macOS using mirrord. This will download and use the x64 mirrord-layer binary when needed.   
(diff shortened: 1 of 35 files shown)


thedanvail:
Hopefully this is the functionality that was requested from #110 - if not, can still make changes. Thought about using `Result` and adding onto `error.rs`, but neither seemed to quite fit the bill for this in my opinion.

aviramha (changes requested):
Thanks for your contribution! See comments inline.

thedanvail:
All right, I've switched over to using `anyhow` for attaching context to these errors. I'm signing off for the night, but should any additional changes be needed, please do not hesitate to let me know and I'll get on them tomorrow evening or possibly Wednesday.

thedanvail:
I see some checks failed, I'll look at those in the morning

aviramha:
LGTM, @meowjesty can you review this please?

aviramha (approved):

mirrord-layer/src/pod_api.rs
```diff
@@ -57,7 +58,7 @@ pub async fn create_agent(
     agent_namespace: &str,
     log_level: String,
     agent_image: String,
-) -> Portforwarder {
+) -> Result<Portforwarder> {
```
meowjesty, line 61: I would suggest returning just `PortForwarder` here, and change `unwrap`s inside this function to `expect` with a nice message giving additional context.

If this function fails we're not handling a potential recovery yet, so it's better to just crash on failures.

mirrord-layer/src/pod_api.rs
```diff
@@ -57,7 +58,7 @@ pub async fn create_agent(
...
     let env_config = config::Config::init_from_env().unwrap();
     let client = if env_config.accept_invalid_certificates {
         let mut config = Config::infer().await.unwrap();
```
meowjesty, line 64: For example here, I would suggest changing it to:

```rust
... = Config.infer().await.expect("Failed to infer config!");
```

Or some nicer message (I'm lacking in creativity here :P).

mirrord-layer/src/pod_api.rs
```diff
@@ -162,6 +163,14 @@ pub async fn create_agent(
 
     let _ = tokio::time::timeout(std::time::Duration::from_secs(20), running)
         .await
-        .unwrap();
-    pods_api.portforward(&pod_name, &[61337]).await.unwrap()
+        .with_context(|| {
```
meowjesty, line 166: And here you change this `with_context` to `expect`(same for every similar piece of code).

meowjesty (changes requested):
Hey Dan, thanks for the help in improving this! The lack of context kinda tripped me over a few times.

My main suggestion comes in the form that we probably don't need to have a `Result` here, as we just crash when this fails anyway. Just change some of the `unwrap`s into `expect`s (also no need to change every single one).

mirrord-layer/src/pod_api.rs
aviramha: I don't think we should call expect from this part of the code.. rather return an error and the top function should decide whether to panic (in this case, it should because it's a ctor function and can't do anything else but panic)

meowjesty: Hey Dan, talked with aviram, and we decided to go for the `Result` approach (sorry about potential confusion).

thedanvail: Haha it's all good, friends. Just to make sure I'm up to date - are there any changes you need before we merge this and I get started on #62 ? I updated the branch to be in line with `upstream/main`, so y'all just let me know if anything needs changing :)

mirrord-layer/src/pod_api.rs
```diff
@@ -162,6 +163,14 @@ pub async fn create_agent(
...
+            format!(
+                "Failed to receive a timely response from pod: {:?}",
+                pod_name
+            )
+        });
```
aviramha, line 171: you lack `?` at the end of the line to make it return the error.

aviramha (changes requested):
Last change and we can merge it :)

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## 2.2.1
 ### Changed
+- Removed `unwrap` from the `Future` that was waiting for Kube pod to spin up in `pod_api.rs`. (Fixes #110)
```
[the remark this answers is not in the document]
thedanvail: It is done, my captain!

mirrord-layer/src/pod_api.rs
thedanvail: Your wish is my command

aviramha:
Great, can you merge main into your branch so tests will run before merging?
Thanks!

eyalb181 (approved):

eyalb181:
Tests are passing, the e2e are just a bit flaky at the moment (need to look into it). Merging to main, thanks again @thedanvail !

PR #147, opened 2022-06-14
Fix universal binary for macOS

aviramha:


aviramha:
@messense would love your feedback on it.

.github/workflows/release.yaml
```diff
@@ -12,13 +12,6 @@ jobs:
           - os: ubuntu-18.04
```
eyalb181, line 12: Is there a reason to leave this as a matrix?

eyalb181 (approved):

.github/workflows/release.yaml
```diff
@@ -39,6 +32,43 @@ jobs:
...
+          MIRRORD_LAYER_FILE: /tmp/target/universal-apple-darwin/release/libmirrord_layer.dylib
+        run: cargo +nightly build --release -p mirrord --target=aarch64-apple-darwin
+      - name: build macOS x86-64 cli with universal dylib
+        env:
+          MIRRORD_LAYER_FILE: /tmp/target/universal-apple-darwin/release/libmirrord_layer.dylib
+        run: cargo +nightly build --release -p mirrord --target=x86_64-apple-darwin
+      - name: create universal macOS cli
+        run: lipo -create -output /tmp/target/universal-apple-darwin/release/mirrord target/aarch64-apple-darwin/release/mirrord target/x86_64-apple-darwin/release/mirrord
+      - name: change dir to tmp
+        run: cd /tmp
```
messense, line 64: I was under the impression that CWD is not preserved between steps?

mirrord-cli/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    if std::env::var("MIRRORD_LAYER_FILE").is_err() {
+        println!(
+            "cargo:rustc-env=MIRRORD_LAYER_FILE={}",
+            std::env::var("CARGO_CDYLIB_FILE_MIRRORD_LAYER").unwrap()
```
messense, line 5: Nice trick! I wonder if there are any size limits for env vars.

.github/workflows/release.yaml
aviramha: Good point, too bad upload artifacts is a very awful action.. I changed to copy the tmp target to current dir so it'll work now :)

.github/workflows/release.yaml
aviramha: erm, maybe when we want to have more archs

messense (approved):

PR #150, opened 2022-06-15
Refactor: Change protocol + mirrord-layer to split messages into modu…

aviramha:
…les, so main module only handles general messages, passing down to the appropriate module for handling.

### DIFF
--- a/mirrord-agent/src/main.rs
+++ b/mirrord-agent/src/main.rs
@@ -11,6 +11,7 @@ use std::{
 use error::AgentError;
 use futures::SinkExt;
 use mirrord_protocol::{
+    tcp::{DaemonTcp, LayerTcp},
     ClientMessage, ConnectionID, DaemonCodec, DaemonMessage, FileRequest, FileResponse, Port,
 };
 use tokio::{
@@ -113,14 +114,23 @@ async fn handle_peer_messages(
     peer_message: PeerMessage,
 ) -> Result<(), AgentError> {
     match peer_message.client_message {
-        ClientMessage::PortSubscribe(ports) => {
+        ClientMessage::Tcp(LayerTcp::PortUnsubscribe(port)) => {
             debug!(
-                "ClientMessage::PortSubscribe -> peer id {:#?} asked to subscribe to {:#?}",
-                peer_message.peer_id, ports
+                "ClientMessage::PortUnsubscribe -> peer id {:#?}, port {port:#?}",
+                peer_message.peer_id
             );
             state
                 .port_subscriptions
-                .subscribe_many(peer_message.peer_id, ports);
+                .unsubscribe(peer_message.peer_id, port);
+        }
+        ClientMessage::Tcp(LayerTcp::PortSubscribe(port)) => {
+            debug!(
+                "ClientMessage::PortSubscribe -> peer id {:#?} asked to subscribe to {:#?}",
+                peer_message.peer_id, port
+            );
+            state
+                .port_subscriptions
+                .subscribe(peer_message.peer_id, port);
 
             let ports = state.port_subscriptions.get_subscribed_topics();
             sniffer_command_tx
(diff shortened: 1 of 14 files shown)


mirrord-agent/src/main.rs
```diff
@@ -288,15 +298,15 @@ async fn start_agent() -> Result<(), AgentError> {
...
+                            SnifferOutput::NewTcpConnection(new_connection) => {
+                                debug!("SnifferOutput::NewTcpConnection -> connection {:#?}", new_connection);
                                 let peer_ids = state.port_subscriptions.get_topic_subscribers(new_connection.destination_port);
 
                                 for peer_id in peer_ids {
                                     state.connections_subscriptions.subscribe(peer_id, new_connection.connection_id);
 
                                     if let Some(peer) = state.peers.get(&peer_id) {
-                                        match peer.channel.send(DaemonMessage::NewTCPConnection(new_connection.clone())).await {
+                                        match peer.channel.send(DaemonMessage::Tcp(DaemonTcp::NewConnection(new_connection.clone()))).await {
```
meowjesty, line 309: `if let Err(fail) = ...` if we don't care about the `Ok` portion.

mirrord-agent/src/main.rs
```diff
@@ -305,13 +315,13 @@ async fn start_agent() -> Result<(), AgentError> {
...
-                            SnifferOutput::TCPClose(close) => {
-                                debug!("SnifferOutput::TCPClose -> close {:#?}", close);
+                            SnifferOutput::TcpClose(close) => {
+                                debug!("SnifferOutput::TcpClose -> close {:#?}", close);
                                 let peer_ids = state.connections_subscriptions.get_topic_subscribers(close.connection_id);
 
                                 for peer_id in peer_ids {
                                     if let Some(peer) = state.peers.get(&peer_id) {
-                                        match peer.channel.send(DaemonMessage::TCPClose(close.clone())).await {
+                                        match peer.channel.send(DaemonMessage::Tcp(DaemonTcp::Close(close.clone()))).await {
```
meowjesty, line 324: Same here.

`if let Err(fail) = ...` if we don't care about the `Ok` portion.

mirrord-agent/src/main.rs
```diff
@@ -321,13 +331,13 @@ async fn start_agent() -> Result<(), AgentError> {
...
-                            SnifferOutput::TCPData(data) => {
-                                debug!("SnifferOutput::TCPData -> data");
+                            SnifferOutput::TcpData(data) => {
+                                debug!("SnifferOutput::TcpData -> data");
                                 let peer_ids = state.connections_subscriptions.get_topic_subscribers(data.connection_id);
 
                                 for peer_id in peer_ids {
                                     if let Some(peer) = state.peers.get(&peer_id) {
-                                        match peer.channel.send(DaemonMessage::TCPData(data.clone())).await {
+                                        match peer.channel.send(DaemonMessage::Tcp(DaemonTcp::Data(data.clone()))).await {
```
meowjesty, line 340: And here.

`if let Err(fail) = ...` if we don't care about the `Ok` portion.

mirrord-layer/src/sockets.rs
```diff
@@ -261,12 +262,12 @@ fn listen(sockfd: RawFd, _backlog: c_int) -> c_int {
                 return ret;
             }
             let sender = unsafe { HOOK_SENDER.as_ref().unwrap() };
-            match sender.blocking_send(HookMessage::Listen(Listen {
+            match sender.blocking_send(HookMessage::Tcp(HookMessageTcp::Listen(Listen {
```
meowjesty, line 265: > `if let Err(fail) = ...` if we don't care about the `Ok` portion.

mirrord-protocol/src/lib.rs
```diff
@@ -3,7 +3,10 @@
 #![feature(core_ffi_c)]
 
 pub mod codec;
+pub mod common;
 pub mod error;
+pub mod tcp;
 
 pub use codec::*;
+pub use common::*;
```
meowjesty, line 11: Hmm, do we need this module right now just for some type aliases? I would just move those here.

meowjesty (commented):
I can see this building on top of the previous refactor.

Can't spot anything that I would ask for changes (at least not until I have to use this :P).

eyalb181 (approved):

mirrord-agent/src/main.rs
aviramha: I'll take care of this in next pr

mirrord-agent/src/main.rs
aviramha: same x3

mirrord-protocol/src/lib.rs
aviramha: good idea

PR #151, opened 2022-06-15
Handle openat/open relative paths 

infiniteregrets:
Following changes have been made in the pull request:
- if open is called with a relative path, we just propagate that call to libc otherwise the call goes to agent 
- if openat is called with an absolute path or `AT_FDCWD` then we send the call to open otherwise we check our hashmap for existing descriptors if we dont have a dirfd we make the libc call 
- formatted python files with black
- a test to verify read-write file operation

### DIFF
--- a/mirrord-agent/src/file.rs
+++ b/mirrord-agent/src/file.rs
@@ -57,7 +57,7 @@ impl FileManager {
                 })
             })?;
 
-        let fd = self.index_allocator.next_index().ok_or_else(|| {
+        let fd = self.index_allocator.next_index().ok_or({
             error!("FileManager::open -> Failed to allocate file descriptor");
             ResponseError::FileOperation(FileError {
                 operation: "open".to_string(),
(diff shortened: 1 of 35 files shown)


tests/python-e2e/ops.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+
+TEXT = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
+
+
+class FileOpsTest(unittest.TestCase):
+    def test_read_write_family(self):
```
aviramha, line 7: how does this test ensure it's running remotely?

mirrord-agent/src/file.rs
```diff
@@ -17,11 +17,15 @@ use tracing::{debug, error};
 
 use crate::{error::AgentError, runtime::get_container_pid, sniffer::DEFAULT_RUNTIME, PeerID};
 
-// TODO: To help with `openat`, instead of `HashMap<_, File>` this should be
-// `HashMap<_, RemoteFile`>, where `RemoteFile` is a struct that holds both the `File` + `PathBuf`.
+#[derive(Debug)]
+pub struct RemoteFile {
```
aviramha, line 21: why do we need it..?

aviramha: when openat is used, we just need a different object for it, but that object doesn't need a path either..

aviramha (changes requested):

mirrord-agent/src/file.rs
```diff
@@ -36,10 +47,28 @@ impl FileManager {
...
+                    .map_err(|err| {
+                        error!(
+                            "FileManager::open -> Failed to get metadata for file {:#?}",
+                            err
+                        );
+                        err
+                    })
+                    .map(|metadata| {
+                        debug!("FileManager::open -> Got metadata for file {:#?}", metadata);
+                        if metadata.is_dir() {
```
eyalb181, line 66: Stylewise, I would suggest
```suggestion
                        let value = if metadata.is_dir() { RemoteFile::Directory(path)} else {RemoteFile::File(file)};
                        self.open_files.insert(fd, value);
```

mirrord-agent/src/file.rs
```diff
@@ -17,11 +18,21 @@ use tracing::{debug, error};
 
 use crate::{error::AgentError, runtime::get_container_pid, sniffer::DEFAULT_RUNTIME, PeerID};
 
-// TODO: To help with `openat`, instead of `HashMap<_, File>` this should be
-// `HashMap<_, RemoteFile`>, where `RemoteFile` is a struct that holds both the `File` + `PathBuf`.
+/// Unqiue file descriptor generator
+pub(crate) fn get_fd() -> usize {
```
eyalb181, line 22: Might be clearer to call this generate_fd, get implies you're retrieving an existing value

mirrord-agent/src/file.rs
```diff
@@ -113,65 +192,87 @@ impl FileManager {
...
+                );
+
+                WriteFileResponse {
+                    written_amount: write_amount,
+                }
+            })
+        } else {
+            return Err(ResponseError::NotFound);
+        }
+        .map_err(|fail| {
```
eyalb181, line 266: Why isn't this inside the if clause (rather than chained to the whole if/else)?

eyalb181 (approved):

aviramha (approved):

PR #153, opened 2022-06-16
Refactor agent side

aviramha:
Continuing my refactor spree, agent side mostly now.


mirrord-agent/src/main.rs
```diff
@@ -160,70 +161,99 @@ async fn handle_peer_messages(
...
+    pub async fn start(
+        sender: Sender<PeerMessage>,
+        id: PeerID,
+        stream: TcpStream,
+        receiver: Receiver<DaemonMessage>,
+        pid: Option<u64>,
+    ) -> Result<(), AgentError> {
+        let file_manager = FileManager::new(pid);
+        let stream = actix_codec::Framed::new(stream, DaemonCodec::new());
+        let mut peer_handler = PeerHandler {
```
eyalb181, line 191: Why doesn't this method take self? And then called with PeerHandler{sender, ...}.start()?

mirrord-agent/src/main.rs
```diff
@@ -160,70 +161,99 @@ async fn handle_peer_messages(
...
+            file_manager,
+            stream,
+            receiver,
+        };
+        peer_handler.handle_loop().await?;
+
+        peer_handler
+            .sender
+            .send(PeerMessage {
+                client_message: ClientMessage::Close,
```
eyalb181, line 203: Is this the agent sending the layer a layer->agent message?

mirrord-agent/src/main.rs
aviramha: It could, but it require self to be `Send` which is useless because the caller never really needs to own the struct itself

mirrord-agent/src/main.rs
aviramha: erm it's a bit off but yeah :F I think we can fix it after I refactor the sniffer part.

mirrord-agent/src/main.rs
```diff
@@ -160,70 +161,99 @@ async fn handle_peer_messages(
...
                     }
-                    None => ()
                 }
 
             }
         }
+        Ok(())
+    }
+    /// Handle incoming messages from the peer.
+    async fn handle_peer_message(&mut self, message: ClientMessage) -> Result<(), AgentError> {
```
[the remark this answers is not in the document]
aviramha: handle_peer_messages wont exist once I'm done refactoring this hopefully :P

aviramha:
@eyalb181 do I have your approval?

eyalb181 (approved):

mirrord-agent/src/main.rs
```diff
@@ -160,70 +161,99 @@ async fn handle_peer_messages(
...
+                peer_id: id,
+            })
+            .await?;
+
+        Ok(())
+    }
+
+    async fn handle_loop(&mut self) -> Result<(), AgentError> {
+        loop {
+            select! {
```
meowjesty, line 213: Think there is a futures thing that could be used to handle just 2 futures, so we could drop the `select!` macro from here.

meowjesty: Not sure if it is an exact replacement for the macro though.

meowjesty (commented):
+1 to the refactor.

Only have small things to comment on it right now, will probably have more once we start using it ;P

mirrord-agent/src/main.rs
aviramha: in next PR I believe I add more stuff to it, so can't really use it.

PR #156, opened 2022-06-17
Add CLI arg for Agent TTL

thedanvail:
Adding in a CLI arg for an Agent TTL

### DIFF
--- a/mirrord-cli/src/main.rs
+++ b/mirrord-cli/src/main.rs
@@ -53,7 +53,7 @@ struct ExecArgs {
     #[clap(value_parser)]
     pub binary: String,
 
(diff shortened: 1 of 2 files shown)


mirrord-cli/src/main.rs
```diff
@@ -53,6 +53,10 @@ struct ExecArgs {
     #[clap(value_parser)]
     pub binary: String,
 
+    /// Agent TTL (overrides environment variable)
```
eyalb181, line 56: Parentheses unnecessary here, essentially all the CLI args override env vars

eyalb181 (changes requested):

PR #159, opened 2022-06-20
Fix once cell renaimings

infiniteregrets:
This PR fixes the compilation issue as with the latest release of rust, the stdlib was updated to use better naming for once cell https://github.com/rust-lang/rust/pull/98165


aviramha (approved):
Thank you so much @infiniteregrets

PR #165, opened 2022-06-21
Refactor agent

aviramha:


mirrord-agent/src/sniffer.rs
```diff
@@ -93,12 +93,73 @@ fn is_closed_connection(flags: u16) -> bool {
     0 != (flags & (TcpFlags::FIN | TcpFlags::RST))
 }
 
-struct ConnectionManager {
-    sessions: SessionMap,
-    index_allocator: IndexAllocator<ConnectionID>,
-    ports: HashSet<u16>,
+pub struct TcpManagerCodec {}
```
meowjesty, line 96: Empty struct could be turned into just a marker struct (also add a bunch of derives to it if it's going to be just a marker).

mirrord-agent/src/sniffer.rs
```diff
@@ -93,12 +93,73 @@ fn is_closed_connection(flags: u16) -> bool {
...
+    capture.set_datalink(Linktype::ETHERNET)?;
+    // Set a dummy filter that shouldn't capture anything. This makes the code easier.
+    capture.filter(DUMMY_BPF, true)?;
+    capture = capture.setnonblock()?;
+
+    Ok(capture)
 }
 
+struct TcpPacketData {
+    data: Vec<u8>,
```
meowjesty, line 130: I would rename this to `bytes`, otherwise I foresee some `let _ = data.data.len()` call.

meowjesty (commented):
First look.

mirrord-agent/src/sniffer.rs
```diff
@@ -111,207 +172,343 @@ fn format_bpf(ports: &[u16]) -> String {
...
+#[derive(Debug)]
+pub struct SnifferCommand {
+    client_id: ClientID,
+    command: SnifferCommands,
+}
+
+pub struct TCPSnifferAPI {
+    client_id: ClientID,
+    sender: Sender<SnifferCommand>,
+    enabled: bool,
```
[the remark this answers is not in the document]
aviramha: How would we benefit from that if we can't implement the variant?

mirrord-agent/src/sniffer.rs
aviramha: Can you elaborate? I'm not sure I follow

meowjesty: Just saying that if this struct has no fields, drop the `{}` (and feel free to add derive Clone, Debug, and so on).

mirrord-agent/src/sniffer.rs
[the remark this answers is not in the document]
meowjesty: Then why do we need this bool?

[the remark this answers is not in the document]
aviramha: It's for the drop..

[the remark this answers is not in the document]
aviramha: done

mirrord-agent/src/util.rs
```diff
@@ -75,7 +77,8 @@ where
         }
     }
 
-    /// Removes a topic and all of it's clients
+    // /// Removes a topic and all of it's clients
+    #[allow(dead_code)]
```
eyalb181, line 81: We should maybe document whenever we use allow(dead_code)

mirrord-agent/tests/blackbox.rs
```diff
@@ -26,23 +26,26 @@ mod tests {
...
-            .stdout(Stdio::piped())
-            .stderr(Stdio::piped())
             .spawn()
             .expect("mirrord-agent failed to start");
         // Wait for agent to listen
         sleep(Duration::from_millis(2000)).await;
         let stream = TcpStream::connect("127.0.0.1:61337")
             .await
             .expect("connection to agent failed");
+        println!("..");
```
eyalb181, line 36: Was this intentionally left in? Don't we want proper logging in tests?

eyalb181 (changes requested):

eyalb181 (approved):

meowjesty (commented):
lgtm +1

PR #166, opened 2022-06-23
Reusable Index

aviramha:
Built on top of #165


aviramha:
Closed, wont merge. I came to realize this is not worth it as the odds of reaching u64 max in a single run are improbable.

PR #167, opened 2022-06-23
Refactor sockets (-layer)

meowjesty:
Trying to refactor the socket code in `mirrord-layer` to have a more straightforward "flow", and not have `listen` be such a big function.

### DIFF
--- a/mirrord-layer/src/socket.rs
+++ b/mirrord-layer/src/socket.rs
@@ -7,7 +7,6 @@ use std::{
     sync::{Arc, LazyLock, Mutex},
 };
 
-use errno::{set_errno, Errno};
 use libc::{c_int, sockaddr, socklen_t};
 use mirrord_protocol::Port;
 use os_socketaddr::OsSocketAddr;
@@ -100,6 +99,25 @@ pub struct Socket {
     pub state: SocketState,
 }
 
+impl Socket {
+    fn get_connected_remote_address(&self) -> Result<SocketAddr, LayerError> {
+        if let SocketState::Connected(connected) = &self.state {
+            Ok(connected.remote_address)
+        } else {
+            Err(LayerError::SocketInvalidState(self.state.clone()))
+        }
+    }
+
+    fn get_local_address(&self) -> Result<SocketAddr, LayerError> {
+        match &self.state {
+            SocketState::Initialized => Err(LayerError::SocketInvalidState(self.state.clone())),
+            SocketState::Bound(bound) => Ok(bound.address),
+            SocketState::Listening(listening) => Ok(listening.address),
+            SocketState::Connected(connected) => Ok(connected.local_address),
+        }
+    }
+}
+
 impl TryFrom<&Socket> for OsSocketAddr {
     type Error = LayerError;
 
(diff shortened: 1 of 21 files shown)


mirrord-layer/src/socket/ops.rs
```diff
@@ -0,0 +1,210 @@
...
+    error::LayerError,
+    message::HookMessage,
+    tcp::{HookMessageTcp, Listen},
+    HOOK_SENDER,
+};
+
+/// Create the socket, add it to `MIRROR_SOCKETS` if it's successful, plus protocol and domain
+/// matches our expected type (Tcpv4/v6).
+///
+/// Otherwise the socket is added to `BYPASS_SOCKETS`.
```
aviramha, line 23: why do we need `BYPASS_SOCKETS`?

meowjesty: Good intuition, we don't need it. I started up with the assumption that we would need to capture every socket, but I don't think we need to do so.

aviramha:
I love the socket2 usage. Great idea!

mirrord-layer/src/socket/ops.rs
```diff
@@ -0,0 +1,242 @@
...
+            state: SocketState::Initialized,
+        };
+
+        MIRROR_SOCKETS.lock().unwrap().insert(fd, mirror_socket);
+
+        Ok(fd)
+    }
+}
+
+fn mirror_address(address: SocketAddr) -> Result<SocketAddr, LayerError> {
```
aviramha, line 55: maybe doc string above?

mirrord-layer/src/socket/ops.rs
```diff
@@ -0,0 +1,242 @@
...
+    error::LayerError,
+    message::HookMessage,
+    tcp::{HookMessageTcp, Listen},
+    HOOK_SENDER,
+};
+
+/// Create the socket, add it to `MIRROR_SOCKETS` if it's successful, plus protocol and domain
+/// matches our expected type (Tcpv4/v6).
+///
+/// Otherwise the socket is added to `BYPASS_SOCKETS`.
```
aviramha, line 18: remove bypass from docs also :)

mirrord-layer/src/socket/ops.rs
```diff
@@ -0,0 +1,242 @@
...
+        Type::from(type_),
+        Some(Protocol::from(protocol)),
+    )?;
+
+    let fd = socket.as_raw_fd();
+
+    // We don't handle non Tcpv4 sockets
+    if !((domain == libc::AF_INET) || (domain == libc::AF_INET6) && (type_ & libc::SOCK_STREAM) > 0)
+    {
+        warn!(
```
aviramha, line 36: not a warning.. this is completely fine. either trace or debug.

mirrord-layer/src/socket/ops.rs
```diff
@@ -0,0 +1,242 @@
...
+                .insert(accepted_fd, accepted_socket);
+        }
+
+        Ok((accepted_fd, accepted_address))
+    } else {
+        Err(LayerError::LocalFdNotFound(sockfd))
+    }
+}
+
+pub(super) fn dup(sockfd: RawFd) -> Result<RawFd, LayerError> {
```
aviramha, line 220: docstring

aviramha (commented):
Very well done! few nitpicks.

mirrord-layer/src/socket.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use mirrord_protocol::Port;
+use socket2::Socket;
+
+use crate::error::LayerError;
+
+pub(crate) mod hooks;
+pub(crate) mod ops;
+
+pub(crate) static MIRROR_SOCKETS: LazyLock<Mutex<HashMap<RawFd, MirrorSocket>>> =
```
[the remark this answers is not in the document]
meowjesty: A recent PR changed the name of [`SyncLazy`](https://github.com/rust-lang/rust/issues/74465).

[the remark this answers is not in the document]
aviramha: gee, cutting edge!

meowjesty:
Closing this as it'll require more work to get thing behaving nicely. Left some comments around for future use:

The refactor of sockets takes more work than just moving the implementation of functions around.

1. The use of `socket2` helps with safety around `libc` calls, but it also introduces potential deadlock and ownership issues;
2. Deadlocks are easy to trigger because we end up calling our own safe wrapper functions, which also lock the `MIRROR_SOCKETS` map of managed sockets. If we had a way of pre-loading the `libc` functions we could avoid these issues, but we don't;

PR #173, opened 2022-06-28
E2E - Use real app

### LINKED ISSUE #149: Change e2e to use a "real app"
Right now the E2E uses something we found online (http echo) but that means that the TCP/HTTP session behavior can be different leading to odd reactions.
I think building our own docker image for the test would be better. It should be something that accepts the same APIs that our tests use, have files that we expect to have, etc.

Long story:
Right now we work on TCP level, and therefore if the application running locally and the application running remotely are different, one can close the TCP connection for each HTTP request and the other can reuse it.
By using the same app we eliminate randomness.

infiniteregrets:
Make the E2E tests use real deployment for consistency with the local process. Closes #149


infiniteregrets:
Log - 
```
test tests::test_complete_node_api ... pods: ObjectList { metadata: ListMeta { continue_: None, remaining_item_count: None, resource_version: Some("518"), self_link: None }, items: [] }
```

PR #175, opened 2022-06-29
Add From trait impl for LayerError

infiniteregrets:
Better error handling for File operations. Marking as draft for now, need to check what should various error types map to on Linux or if we even need these mappings.


infiniteregrets:
What kind of mapping are we looking for between LayerError and libc errorn nos?

mirrord-layer/src/error.rs
```diff
@@ -53,3 +53,31 @@ pub enum LayerError {
...
+            LayerError::SendErrorHookMessage(_) => libc::EBADMSG,
+            LayerError::SendErrorConnection(_) => libc::EBADMSG,
+            LayerError::SendErrorLayerTcp(_) => libc::EBADMSG,
+            LayerError::RecvError(_) => libc::EBADMSG,
+            LayerError::Null(_) => libc::EINVAL,
+            LayerError::TryFromInt(_) => libc::EINVAL,
+            LayerError::LocalFDNotFound(_) => libc::EBADF,
+            LayerError::EmptyHookSender => libc::ENOENT,
+            LayerError::NoConnectionId(_) => libc::ECONNREFUSED,
+            LayerError::IO(_) => libc::EIO,
```
meowjesty, line 77: IO errors should probably use [`raw_os_error`](https://doc.rust-lang.org/std/io/struct.Error.html#method.raw_os_error). You could `unwrap_or` it to return `-1` for the `None` case.

mirrord-layer/src/error.rs
```diff
@@ -53,3 +53,31 @@ pub enum LayerError {
...
+
+//Todo: https://stackoverflow.com/questions/39150216/implementing-a-trait-for-multiple-types-at-once
+
+// what should the mapping be between LayerError and integer types
+
+// mapping based on - https://man7.org/linux/man-pages/man3/errno.3.html
+
+impl From<LayerError> for i32 {
+    fn from(error: LayerError) -> Self {
+        match error {
```
meowjesty, line 65: Log the error with `error!` before the `match` so we get a bit more context.

mirrord-layer/src/file/hooks.rs
```diff
@@ -34,14 +35,16 @@ pub(super) unsafe extern "C" fn open_detour(raw_path: *const c_char, open_flags:
...
+        let open_result = open(path.clone(), open_options);
 
-        open_result
-            .map_err(|fail| {
-                error!("Failed opening file with {:#?}", fail);
+        match open_result {
+            Ok(fd) => fd,
+            Err(fail) => {
+                error!("Failed opening file {:#?} with {:#?}", path, fail);
+                set_errno(Errno(fail.into()));
```
meowjesty, line 44: Have you tried using `set_errno` in the `From` impl instead of in each function?

mirrord-layer/src/file/hooks.rs
```diff
@@ -34,14 +35,16 @@ pub(super) unsafe extern "C" fn open_detour(raw_path: *const c_char, open_flags:
         libc::open(raw_path, open_flags)
     } else {
         let open_options = OpenOptionsInternalExt::from_flags(open_flags);
-        let open_result = open(path, open_options);
+        let open_result = open(path.clone(), open_options);
 
-        open_result
-            .map_err(|fail| {
-                error!("Failed opening file with {:#?}", fail);
+        match open_result {
```
meowjesty, line 40: Instead of `match`, you could do `map_err(From::from)`, something like this:

```rust
let (Ok(result) | Err(result)) = open.map_err(From::from);

result
```

And for functions where the safe wrapper doesn't return some number type, you could do `operantion.map(|something| .... )` before `map_err` to convert the result into the appropriate return value.

meowjesty (commented):
Nice! Leaving some notes on how to make the code even shorter.

mirrord-layer/src/error.rs
infiniteregrets: Does that mean we remove the log message from the hook? that would make the `map_err` ignore the argument in the closure

mirrord-layer/src/error.rs
infiniteregrets: I dont think there is any point right now? where is this error being generated? it would always return None because as of now we are not using any `raw_os_error` related functions? if we do follow this, shouldn't we do the same for all errors

mirrord-layer/src/error.rs
```diff
@@ -53,3 +56,40 @@ pub enum LayerError {
     #[error("mirrord-layer: Failed inserting listen, already exists!")]
     ListenAlreadyExists,
 }
+
+// mapping based on - https://man7.org/linux/man-pages/man3/errno.3.html
+
+macro_rules! impl_from {
+    ($type:ty, $return:expr) => {
```
meowjesty, line 63: Just avoid using keywords here, I would go for `return_type` and `return_value`, instead of `type` and `return`.

meowjesty (approved):
That's it, just rename the macro args and I think it's fine.

PR #180, opened 2022-07-01
E2e revamp

aviramha:
Refactored E2E, only HTTP Mirroring on node is tested (need to fix some bugs to enable other)


tests/src/sanity.rs
```diff
@@ -1,552 +1,556 @@
...
+            .envs(env)
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .unwrap();
+        println!(
+            "executed mirrord with args {args:?} pid {}",
+            server.id().unwrap()
+        );
+        // We need to hold temp dir until the process is finished
```
meowjesty, line 213: Document this on the `TestProcess` struct also.

meowjesty (commented):
Looks good as far as I can see.

meowjesty (approved):
👍

Issue #100, opened 2022-05-31
Add ability to specify container name to impersonate


aviramha:
Today, we pick up the first container in the pod we impersonate. This might not be the case for service mesh setups or more elaborate setups, so we need to let the user specify container name and maybe add a warning when we just assume and use the first one.
Also, we should consider ignoring linkerd/envoy containers by default (unless specified?)
Edit:

Status:
- [x]  @camerondurham added basic functionality controlled by env var.
- [x]  Need to expose that functionality from the cli
- [x]  ^^ VS Code Extension

camerondurham:
I'd like to work on this issue. Thanks!

aviramha:
> I'd like to work on this issue. Thanks!

Great! Let us know if you need any help 😊

aviramha:
Thanks @camerondurham . I'm updating the ticket to reflect what needs to be added.

aviramha:
@camerondurham Do you you wish to contribute the other open items as well? (making sure the assigned state is correct ;) )

camerondurham:
Hey @aviramha, I'm interested in finishing the other tasks as well if that's alright! I was running into a bug with the extension selection so sending the draft PR for now and can provide more context in `#dev` too...

aviramha:
That's more than alright! Feel free to consult with us on `#dev`

aviramha:
@camerondurham
Hey, are you still working on this issue? No pressure :) Just so we can mark this as available for others if not.

camerondurham:
@aviramha Following up from convo in`#dev`.

I can help with minor changes to fix existing features added but if it's more than a couple days work I probably won't have time.

What I've done so far:

1. (PR #182) Added a flag to the CLI in (here: https://github.com/metalbear-co/mirrord/pull/182/files#diff-e7e961df0a87503db868f6db719d10b5f06c962d7b1939079055514a118844bc)
2. (PR #185) Added another quickselect window to the VSCode extension when starting to debug (here: https://github.com/metalbear-co/mirrord/pull/185/files#diff-71da5b8ace00ececa231c3f58ec52d2876a05efaf23404acf1c24cbebbad2bea) 

Thanks for following up, I should've asked this sooner!

aviramha:
Thanks @camerondurham - My fault for not tracking the issues/linking to this.

PR #182, opened 2022-07-03
Add MIRRORD_IMPERSONATED_CONTAINER_NAME var

camerondurham:
### Details

This CR adds an optional `MIRRORD_IMPERSONATED_CONTAINER_NAME` to specify container name within pod for #100.

This doesn't implement ignoring linkerd/envoy pods yet. Can iterate on this draft based on feedback. I suppose the best way to ignore containers like this is to filter out common container names for linkerd/envoy within a pod?

A couple things that are missing, I'll convert to draft if these are blockers.

1. add ability to set this variable in VSCode extension
2. Ignore linkerd/envoy container names

### Testing

These are low effort tests and not really best practice since they won't be in CI. I can try this weekend adding a mock failure run to [sanity.rs](https://github.com/metalbear-co/mirrord/blob/main/tests/src/sanity.rs#L203-L209).

1. local GH action run manually setting `MIRRORD_IMPERSONATED_CONTAINER_NAME=test` ([link](https://github.com/camerondurham/mirrord/runs/7167446938?check_suite_focus=true))
2. local GH action run not setting `MIRRORD_IMPERSONATED_CONTAINER_NAME` ([link](https://github.com/camerondurham/mirrord/actions/runs/2604311840))
3. local GH action sanity check failed when setting `MIRRORD_IMPERSONATED_CONTAINER_NAME=testgarbage` ([link](https://github.com/camerondurham/mirrord/runs/7167505799?check_suite_focus=true#step:15:76))

Example failure:

https://github.com/camerondurham/mirrord/runs/7167505799?check_suite_focus=true#step:15:76

```
stderr 7282: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: no container named testgarbage found in namespace=default, pod=http-echo-gw7u7b2-59cdf9cf46-pfqwp', mirrord-layer/src/pod_api.rs:49:18
```

Thanks!

### DIFF
--- a/mirrord-layer/src/pod_api.rs
+++ b/mirrord-layer/src/pod_api.rs
@@ -22,15 +22,38 @@ struct RuntimeData {
 }
 
 impl RuntimeData {
-    async fn from_k8s(client: Client, pod_name: &str, pod_namespace: &str) -> Self {
+    async fn from_k8s(
+        client: Client,
+        pod_name: &str,
+        pod_namespace: &str,
+        container_name: &Option<String>,
+    ) -> Self {
         let pods_api: Api<Pod> = Api::namespaced(client, pod_namespace);
         let pod = pods_api.get(pod_name).await.unwrap();
         let node_name = &pod.spec.unwrap().node_name;
         let container_statuses = pod.status.unwrap().container_statuses.unwrap();
-        let container_info = container_statuses
-            .first()
-            .unwrap()
-            .container_id
+        let container_info_map = container_statuses
+            .into_iter()
+            .map(move |cs| (cs.name.clone(), cs))
+            .collect::<std::collections::HashMap<_, _>>();
+
+        let container_info = if let Some(container_name) = container_name {
+            &container_info_map
+                .get(container_name)
+                .with_context(|| {
+                    format!(
+                        "no container named {} found in namespace={}, pod={}",
+                        &container_name, &pod_namespace, &pod_name
+                    )
+                })
+                .unwrap()
+                .container_id
+        } else {
+            warn!("No container name specified, defaulting to first container found");
+            &container_info_map.iter().next().unwrap().1.container_id
+        };
+
+        let container_info = container_info
             .as_ref()
             .unwrap()
             .split("://")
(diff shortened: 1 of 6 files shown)


mirrord-layer/src/pod_api.rs
```diff
@@ -22,15 +22,38 @@ struct RuntimeData {
...
+    ) -> Self {
         let pods_api: Api<Pod> = Api::namespaced(client, pod_namespace);
         let pod = pods_api.get(pod_name).await.unwrap();
         let node_name = &pod.spec.unwrap().node_name;
         let container_statuses = pod.status.unwrap().container_statuses.unwrap();
-        let container_info = container_statuses
-            .first()
-            .unwrap()
-            .container_id
+        let container_info_map = container_statuses
```
aviramha, line 35: I think maybe just looping would be better, this is a bit of a nitpick but creating the hashmap, cloning data then just getting one key is overkill then to just check on the iteration.
Possibly https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

mirrord-layer/src/pod_api.rs
```diff
@@ -22,15 +22,38 @@ struct RuntimeData {
...
+                .with_context(|| {
+                    format!(
+                        "no container named {} found in namespace={}, pod={}",
+                        &container_name, &pod_namespace, &pod_name
+                    )
+                })
+                .unwrap()
+                .container_id
+        } else {
+            warn!("No container name specified, defaulting to first container found");
```
aviramha, line 52: I don't think it's a warning :)

aviramha (changes requested):
Awesome! This looks great. I left two comments, one I consider a nitpick and the other is UX but small so would be required to get merged.

mirrord-layer/src/pod_api.rs
camerondurham: You're totally right, `find()` would be simpler and this is pretty convoluted :)

mirrord-layer/src/pod_api.rs
camerondurham: Ah ok. Should it just be a println! instead or just not print anything? 

Happy to change it either way!

aviramha: `info!` seems suitable.

aviramha (approved):
Great, Thank you.

PR #183, opened 2022-07-03
Fix unwraps in fileops

infiniteregrets:
Handle a few unwraps in lib.rs for fileops, such that mirrord does not panic when opening a "non existent" file and provides error messages nicely.


aviramha:
Can you re enable the file ops tests ?

.github/workflows/ci.yaml
```diff
@@ -206,7 +206,7 @@ jobs:
       - name: setup cluster requirements
         run: brew install hyperkit docker minikube
       - name: start minikube
-        run: minikube start --driver=hyperkit --container-runtime=docker
+        run:  minikube start --driver=hyperkit --container-runtime=docker
```
aviramha, line 209: I think this is a typo? (why do we need an extra space?)

mirrord-layer/src/error.rs
```diff
@@ -52,4 +55,20 @@ pub enum LayerError {
 
     #[error("mirrord-layer: Failed inserting listen, already exists!")]
     ListenAlreadyExists,
+
+    #[error("mirrord-layer: Failed to get a `Lock` on resource!")]
```
aviramha, line 59: Failed to lock resource IMO

mirrord-layer/src/lib.rs
```diff
@@ -256,82 +257,82 @@ async fn handle_daemon_message(
...
             write_file_handler
-                .lock()
-                .unwrap()
+                .lock()?
                 .pop()
-                .unwrap()
-                .send(write_file.unwrap())
-                .unwrap();
+                .ok_or(LayerError::SendErrorFileResponse)?
+                .send(write_file?)
```
aviramha, line 313: why do you uwnrap the write_file? we want it to send it to the function so it can return the appropriate error..

tests/src/sanity.rs
```diff
@@ -507,8 +507,7 @@ mod tests {
 
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    #[ignore] // TODO: Fix this test
-    pub async fn test_file_ops(#[future] service: EchoService) {
+    pub async fn test_file_ops_(#[future] service: EchoService) {
```
aviramha, line 510: why `_` at the endo f the function?

tests/src/sanity.rs
```diff
@@ -521,7 +520,6 @@ mod tests {
         .await;
         let res = process.child.wait().await.unwrap();
         assert!(res.success());
```
aviramha, line 522: we want to assert the std err, but just let it be more permissive (allow specific text)

aviramha (changes requested):

tests/src/sanity.rs
infiniteregrets: should it be a separate function? what kind of abstraction are you looking for to handle this? it makes sense to keep it a separate function because its specific to python only

aviramha: yup, feel free to make it a standalone function or w.e you think

mirrord-layer/src/lib.rs
```diff
@@ -256,82 +257,82 @@ async fn handle_daemon_message(
...
             open_file_handler
-                .lock()
-                .unwrap()
+                .lock()?
                 .pop()
-                .unwrap()
-                .send(open_file.unwrap())
-                .unwrap();
+                .ok_or(LayerError::SendErrorFileResponse)?
+                .send(open_file?)
```
aviramha, line 274: don't propagate?

mirrord-layer/src/lib.rs
```diff
@@ -256,82 +257,82 @@ async fn handle_daemon_message(
...
             seek_file_handler
-                .lock()
-                .unwrap()
+                .lock()?
                 .pop()
-                .unwrap()
-                .send(seek_file.unwrap())
-                .unwrap();
+                .ok_or(LayerError::SendErrorFileResponse)?
+                .send(seek_file?)
```
aviramha, line 302: don't propogatE?

mirrord-layer/src/lib.rs
```diff
@@ -256,82 +257,82 @@ async fn handle_daemon_message(
...
             close_file_handler
-                .lock()
-                .unwrap()
+                .lock()?
                 .pop()
-                .unwrap()
-                .send(close_file.unwrap())
-                .unwrap();
+                .ok_or(LayerError::SendErrorFileResponse)?
+                .send(close_file?)
```
aviramha, line 324: same

aviramha (changes requested):

meowjesty (approved):
👍 



Sorry for sending you on this quest! Here is your reward 🥇

tests/src/sanity.rs
```diff
@@ -88,6 +88,10 @@ mod tests {
             assert!(self.stderr.lock().unwrap().is_empty());
         }
 
+        fn assert_python_fileops_stderr(&self) {
+            assert!(self.stderr.lock().unwrap().contains("OK"));
```
aviramha, line 92: this is a very weak assertion.. we want to know mirrord layer didn't report errors as well. if tests passed by mirrord reached an unexpected scenario we want to fail :)

aviramha (changes requested):

aviramha (approved):
LGTM! Thanks

PR #185, opened 2022-07-04
Add quickselect for impersonated container name, fix CLI env var name

camerondurham:
### Details

Working on remaining action items for #100: updating CLI and VSCode Extension. 

This is kind of a large, messy PR combining both the Rust CLI + Typescript changes. I'm happy to break this apart more.

#### Testing extension
1. `cd vscode-ext && vsce package`
2. Command Palette > Install extension from `.vsxi`
3. Select `mirrord-2.3.1.vsix`

#### Using quickselect flow

~~**WIP:** There's a bug in the quickselect handling after selecting the pod name. I'm not quite sure what I'm doing wrong but think it has something to do with not bad multi-select handling. Will study [this sample](https://github.com/microsoft/vscode-extension-samples/blob/main/quickinput-sample/src/multiStepInput.ts) a bit more.~~

**EDIT**: I think I found the bug, I needed to return the `vscode.DebugConfiguration` in all branches.

[Unlisted YT Screen Recording (HD)](https://youtu.be/Ct3Ny7ISVTk)

**Low Quality Screen Rec/Gif**

animation_opt

#### Verify New CLI Flag

Installed CLI locally with `cargo install --path .`, will continue to work on testing against a local pod.

The new flag is a little verbose... `--impersonated-container-name <IMPERSONATED_CONTAINER_NAME>`. Happy to shorten it and/or adjust implementation handling.

### DIFF
--- a/vscode-ext/README.md
+++ b/vscode-ext/README.md
@@ -23,4 +23,4 @@ mirrord lets you easily mirror traffic from your Kubernetes cluster to your loca
 
 - You can control the namespace mirrord will find pods by changing the impersonated pod namespace by clicking the settings button next to the Enable/Disable mirrord button
 - You can also control in which k8s namespace the mirrord-agent will spawn using the same setting button.
-- You can control which container name in the impersonated pod namespace
+- You can control which container to impersonate within the impersonated pod
--- a/vscode-ext/README.md
+++ b/vscode-ext/README.md
@@ -23,3 +23,4 @@ mirrord lets you easily mirror traffic from your Kubernetes cluster to your loca
 
 - You can control the namespace mirrord will find pods by changing the impersonated pod namespace by clicking the settings button next to the Enable/Disable mirrord button
 - You can also control in which k8s namespace the mirrord-agent will spawn using the same setting button.
+- You can control which container name in the impersonated pod namespace
--- a/vscode-ext/package.json
+++ b/vscode-ext/package.json
@@ -42,12 +42,14 @@
 		"watch": "tsc -watch -p ./",
 		"pretest": "npm run compile && npm run lint",
 		"lint": "eslint src --ext ts",
+		"format": "eslint src --ext ts --fix",
 		"test": "node ./out/test/runTest.js",
 		"vscode:prepublish": "npm run esbuild-base -- --minify",
 		"esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/main.js --external:vscode --format=cjs --platform=node",
 		"esbuild": "npm run esbuild-base -- --sourcemap",
 		"esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
-		"test-compile": "tsc -p ./"
+		"test-compile": "tsc -p ./",
+		"package": "vsce package"
 	},
 	"devDependencies": {
 		"@types/glob": "^7.2.0",
@@ -63,7 +65,8 @@
 		"glob": "^7.2.0",
 		"mocha": "^9.1.3",
 		"prettier": "2.5.1",
-		"typescript": "^4.5.4"
+		"typescript": "^4.5.4",
+		"vsce": "^2.9.2"
 	},
 	"dependencies": {
 		"@kubernetes/client-node": "^0.16.3",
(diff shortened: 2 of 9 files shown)


vscode-ext/package.json
```diff
@@ -42,12 +42,14 @@
...
 		"lint": "eslint src --ext ts",
+		"format": "eslint src --ext ts --fix",
 		"test": "node ./out/test/runTest.js",
 		"vscode:prepublish": "npm run esbuild-base -- --minify",
 		"esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/main.js --external:vscode --format=cjs --platform=node",
 		"esbuild": "npm run esbuild-base -- --sourcemap",
 		"esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
-		"test-compile": "tsc -p ./"
+		"test-compile": "tsc -p ./",
+		"package": "vsce package"
```
camerondurham, line 52: Added `vsce` dependency and this script to generate a `vsix` for local extension testing. Is there maybe a better way or should I maybe update a dev guide with these steps?

vscode-ext/src/extension.ts
```diff
@@ -159,12 +163,23 @@ class ConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+						// eslint-disable-next-line @typescript-eslint/naming-convention
 						'MIRRORD_AGENT_NAMESPACE': globalContext.workspaceState.get('agentNamespace', 'default'),
 						// eslint-disable-next-line @typescript-eslint/naming-convention
 						'MIRRORD_FILE_OPS': globalContext.workspaceState.get('fileOps', 'false')
 					}
 				};
 				config.env[environmentVariableName] = path.join(libraryPath, libraryName);
+
+				// let user select container name if there are multiple containers in the pod
+				const pod = pods.body.items.find(p => p.metadata!.name === podName!);
```
camerondurham, line 176: Working on this part still. I think the bug is I shouldn't be nesting these quickselect windows. There's an issue with the program not starting debug session after selecting the container name. I'll keep working on this but sending draft PR in case anyone notices an obvious bug in this part.

vscode-ext/README.md
```diff
@@ -23,3 +23,4 @@ mirrord lets you easily mirror traffic from your Kubernetes cluster to your loca
 
 - You can control the namespace mirrord will find pods by changing the impersonated pod namespace by clicking the settings button next to the Enable/Disable mirrord button
 - You can also control in which k8s namespace the mirrord-agent will spawn using the same setting button.
+- You can control which container name in the impersonated pod namespace
```
eyalb181, line 26: You can control which container to impersonate within the impersonated pod

vscode-ext/package.json
eyalb181: So VSCode lets you [debug extensions](https://code.visualstudio.com/api/get-started/your-first-extension), which I find more convenient for all but the very last phase of testing. That said, it could still be useful to have vsce package available in addition. 
And it would be great if you could add it all to the dev guide!

vscode-ext/vscode-ext
```diff
@@ -6,5 +6,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+- Extension provides quick select for setting impersonated container name (environment variable `MIRRORD_IMPERSONATED_CONTAINER_NAME`)
```
eyalb181, line 9: We don't maintain this file - we document all changes in the main mirrord CHANGELOG file.

eyalb181 (changes requested):
Great stuff! Appreciate you doing some code style housekeeping too. Left a few minor comments.

vscode-ext/package.json
camerondurham: Ah thanks! This was my first time working on a VSCode extension and I didn't know that. I would PR `metalbear/mirrord.dev` repo (https://github.com/metalbear-co/mirrord.dev/tree/main/content/docs/developer) for adding to dev guide?

vscode-ext/README.md
camerondurham: Thanks! Fixed this wording in latest commit.

vscode-ext/vscode-ext
camerondurham: Removed this change. I'll just keep the change in that main CHANGELOG.

vscode-ext/package.json
camerondurham: We discussed this on `#dev`. Consensus seems to be new page should be added to developer docs for VSCode Extension development.

eyalb181 (approved):
LGTM :)

Issue #184, opened 2022-07-04
Python example breaks on mirroring traffic on Apple Silicon


aviramha:
### Bug Description

When running a python app with mirrord, we encounter a bug introduced by hooking `fcntl`. 
It seems that our hook messes with the `var_args` of `fcntl`, causing it to break on M1/Apple Silicon. 
This is probably a bug upstream in frida, so this is mostly a tracking to link between the two issues.

### Steps to Reproduce

`mirrord exec --pod-name nginx-deployment-66b6c48dd5-phrpl /usr/local/bin/python3 -- -m http.server`

### Backtrace

```shell
2022-06-29T10:12:27.857622Z ERROR mirrord_layer::sockets: fcntl failed
    app.run(host="0.0.0.0", port=8080)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/flask/app.py", line 920, in run
2022-06-29T10:12:27.857693Z ERROR mirrord_layer::sockets: fcntl failed
    run_simple(t.cast(str, host), port, self, **options)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/werkzeug/serving.py", line 1065, in run_simple
2022-06-29T10:12:27.859086Z ERROR mirrord_layer::sockets: fcntl failed
    srv = make_server(
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/werkzeug/serving.py", line 849, in make_server
2022-06-29T10:12:27.859691Z ERROR mirrord_layer::sockets: fcntl failed
    return ThreadedWSGIServer(
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/werkzeug/serving.py", line 716, in init
2022-06-29T10:12:27.859794Z ERROR mirrord_layer::sockets: fcntl failed
    self.socket = socket.fromfd(fd, address_family, socket.SOCK_STREAM)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/socket.py", line 545, in fromfd
2022-06-29T10:12:27.859896Z ERROR mirrord_layer::sockets: fcntl failed
    nfd = dup(fd)
OSError: [Errno 22] Invalid argument
```

### Your operating system and version

[...]

PR #188, opened 2022-07-06
Fix fcntl m1

aviramha:
Workaround for #184


aviramha:
> Weird that this is needed, but ok. 👍

Yup, I hope we'll fix it upstream soon.

PR #189, opened 2022-07-06
Load environment variables before process starts

aviramha:


meowjesty (approved):
Minor nit using `panic` instead of `unreachable`, but it's fine either way. 👍

Issue #27, opened 2022-04-28
Remote network support


aviramha:
Add support for accessing network resources from the remote service context.
This includes dns resolving.

This will add an option that once enabled, all outgoing traffic will be sent from the remote agent context instead of the local context.

Some example use cases:

- User wants their local process to make an API call to another service in the k8s cluster. When the local process makes a request to another service, mirrord intercepts it and executes it through the agent instead. In the future we'd probably want very granular configuration here - so e.g. be able to toggle the feature only for certain destination hostnames, or even certain HTTP methods/parameters. This isn't part of the current requirements but might be worth thinking about when designing the feature.
- User wants their local process to read data from a database on the k8s cluster. When the local process makes a request to the database, mirrord should intercept it and execute it through the agent instead. Note that this would often rely on environment variables being overridden with their remote values, so that the local process would have the correct connection string.
- Users wants their local process to make an API call to an external API accessible from the k8s cluster (e.g. a hosted database).

aviramha:
I think this issue should be broken into few parts, here's a suggestion:

1. DNS Resolving - This should be done by hooking `getaddrbyname` and `getaddrinfo`. Should be a flag. (`MIRRORD_ENABLE_DNS` ?)
2. TCP Outgoing - We already hook most of the relevant functions, we need to handle `connect` flow, then connect from the remote agent, probably have a local "proxy" (using DuplexStream?) and `copy_bidirectional` of tokio. (`MIRRORD_ENABLE_TCP`?)
3. UDP Outgoing - This is a bit more "complicated" as it's not a stream, but might be even "easier" on some level. There's a bind that would probably happen if the application wants to receive data also. Let's start discussing it once we're close to finish 2.

eyalb181:
When would it make sense to turn on ENABLE_TCP without ENABLE_DNS?

aviramha:
> When would it make sense to turn on ENABLE_TCP without ENABLE_DNS?
When the services are accessible from your existing network I guess? but my thought was to split it into a feature flag just so it can be introduced partially, but ofc we can have a united flag for networking.

eyalb181:
Okay, wanted to make sure there wasn't a use case I was missing here. I wouldn't split it into two flags, it seems like unnecessary complexity.

aviramha:
We still lack some UDP flows AFAIK based on what @t4lz said (recvto/sendto flow)

eyalb181:
My mistake, reopened.

[...]

PR #190, opened 2022-07-08
DNS resolving

meowjesty:
PR for solving part of #27 [DNS resolving](https://github.com/metalbear-co/mirrord/issues/27#issuecomment-1154072686).

Hooks `getaddrinfo`, `freeaddrinfo` pair.


tests/node-e2e/remote_dns/test_remote_dns_resolving_works.mjs
```diff
@@ -0,0 +1,27 @@
+import { createServer } from "net";
+
+console.log(">> test_remote_remote_dns_resolving_works");
+
+const server = createServer();
+server.on("connection", handleConnection);
```
aviramha, line 6: not sure how this tests it?

aviramha (changes requested):

tests/node-e2e/remote_dns/test_remote_dns_resolving_works.mjs
meowjesty: Yeah, this was just my "does enabling the feature causes a crash?" test.

mirrord-agent/src/main.rs
```diff
@@ -238,6 +253,46 @@ impl ClientConnectionHandler {
                     .send(DaemonMessage::GetEnvVarsResponse(env_vars_result))
                     .await?
             }
+            ClientMessage::GetAddrInfoRequest(GetAddrInfoRequest {
+                node,
+                service,
+                hints,
+            }) => {
```
aviramha, line 260: I prefer this to be in a function as this enum is huge and should have 1-2 lines per variant

aviramha (approved):

PR #195, opened 2022-07-10
Downgrade es5-ext so we don't get flagged as malicious

eyalb181:
es5-ext 10.54+ is contains [protestware](https://medium.com/checkmarx-security/new-protestware-found-lurking-in-highly-popular-npm-package-d46f8ba67e36) which is flagged as malicious by the VSCode marketplace. Downgrade the dependency so we don't get flagged.
image


aviramha:
Created issue https://github.com/medikoo/es5-ext/issues/186 for tracking in case they decide to change it so we can update..

aviramha (approved):

PR #203, opened 2022-07-12
Add "ephemeral containers" feature 

### LINKED ISSUE #172: Add support for using ephemeral containers
https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/

We should use these for the mirrord-agent instead of jobs with TTL=0. Since they're currently in beta, they should be implemented with a feature flag.

infiniteregrets:
This PR adds support for the Kubernetes beta feature called "ephemeral containers" and closes #172


Cargo.toml
```diff
@@ -37,9 +37,9 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 futures = "0.3"
 thiserror = "1"
-k8s-openapi = { version = "0.15", features = ["v1_22"] }
+k8s-openapi = { version = "0.15", features = ["v1_24"] }
 reqwest = { version = "0.11", features = ["blocking"] }
-kube = { version = "0.73", default-features = false, features = ["runtime", "derive", "client", "ws", "rustls-tls"] }
+kube = { version = "0.73.1", default-features = false, features = ["runtime", "derive", "client", "ws", "rustls-tls"] }
```
aviramha, line 42: we prefer to not specify patch version.. any reason to do so here?

mirrord-layer/src/error.rs
```diff
@@ -64,6 +65,21 @@ pub enum LayerError {
 
     #[error("mirrord-layer: Unmatched pong!")]
     UnmatchedPong,
+
+    #[error("mirrord-layer: Feailed to get `KubeConfig`!")]
```
aviramha, line 69: typo

mirrord-layer/src/pod_api.rs
```diff
@@ -107,126 +110,191 @@ pub async fn create_agent(config: LayerConfig) -> Result<Portforwarder> {
...
-                    ]
+    let pod_name = if env_config.ephemeral_container {
+        warn!("Ephemeral Containers is an experimental feature
+              >> Refer https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/ for more info");
+        let ephemeral_container: EphemeralContainer = serde_json::from_value(json!({
+            "name": "mirrord-agent",
+            "image": agent_image,
+            "imagePullPolicy": image_pull_policy,
+            "target_container_name": impersonated_container_name,
+            "securityContext": {
```
aviramha, line 125: we can drop the privlege

mirrord-layer/src/pod_api.rs
```diff
@@ -107,126 +110,191 @@ pub async fn create_agent(config: LayerConfig) -> Result<Portforwarder> {
...
+            .fields(&format!("metadata.name={}", "mirrord_agent"))
+            .timeout(20);
+
+        let mut stream = pod_api
+            .watch(&params, "0")
+            .await
+            .map_err(|e| LayerError::KubeError(e))?
+            .boxed();
+
+        while let Some(status) = stream.try_next().await? {
```
aviramha, line 178: timeout?

mirrord-layer/src/pod_api.rs
```diff
@@ -107,126 +110,191 @@ pub async fn create_agent(config: LayerConfig) -> Result<Portforwarder> {
...
-                                runtime_data.container_id,
-                                "--container-runtime",
-                                runtime_data.container_runtime,
-                                "-t",
-                                "30",
-                            ],
-                            "env": [{"name": "RUST_LOG", "value": agent_rust_log}],
-                        }
-                    ]
+    let pod_name = if env_config.ephemeral_container {
```
aviramha, line 117: split the logic into two functions (ephemeral and old method) as this is getting too long

mirrord-layer/src/pod_api.rs
```diff
@@ -107,126 +110,191 @@ pub async fn create_agent(config: LayerConfig) -> Result<Portforwarder> {
...
+                "ephemeralcontainers",
+                &env_config.impersonated_pod_name,
+                &PostParams::default(),
+                to_vec(&ephemeral_containers_subresource).unwrap(),
+            )
+            .await
+            .map_err(|e| LayerError::KubeError(e))?;
+
+        let params = ListParams::default()
+            .fields(&format!("metadata.name={}", "mirrord_agent"))
```
aviramha, line 169: maybe generate a unique name so it wont wait for *any* agent

aviramha (changes requested):
Looks great, left few comments in line.
Need to add tests. (just a case probably?)

mirrord-layer/src/pod_api.rs
aviramha: are you sure it's a timeout for the loop and not for the initial request?

mirrord-layer/src/pod_api.rs
```diff
@@ -72,62 +74,158 @@ impl RuntimeData {
...
             .to_lowercase()
     );
+    agent_name
+}
 
-    let agent_image = agent_image.unwrap_or_else(|| {
-        concat!("ghcr.io/metalbear-co/mirrord:", env!("CARGO_PKG_VERSION")).to_string()
-    });
+async fn create_ephemeral_container_agent(
+    config: LayerConfig,
```
aviramha, line 135: can pass it as reference btw

mirrord-layer/src/pod_api.rs
```diff
@@ -72,62 +74,158 @@ impl RuntimeData {
...
+        .fields(&format!("metadata.name={}", "mirrord_agent"))
+        .timeout(10);
+
+    let mut stream = pods_api
+        .watch(&params, "0")
+        .await
+        .map_err(LayerError::KubeError)?
+        .boxed();
+
+    while let Some(status) = stream.try_next().await? {
```
aviramha, line 196: to re-iterate, if you won't have timeout here and we'll get events that aren't error or added, this will run forever.

aviramha (changes requested):

mirrord-layer/src/pod_api.rs
aviramha: seems overkill

aviramha (approved):

PR #205, opened 2022-07-13
Layer refactor

aviramha:
pending merge of #190


mirrord-layer/src/file.rs
```diff
@@ -106,3 +117,299 @@ impl OpenOptionsInternalExt for OpenOptionsInternal {
...
+            "HookMessage::OpenFileHook path {:#?} | options {:#?}",
+            path, open_options
+        );
+
+        self.open_queue.push_back(file_channel_tx);
+
+        let open_file_request = OpenFileRequest { path, open_options };
+
+        let request = ClientMessage::FileRequest(FileRequest::Open(open_file_request));
+        codec.send(request).await.map_err(From::from)
```
meowjesty, line 219: Why do you need most of these `map_err(From::from)`? Can't you do `Ok(...?)` instead?

```rust
Ok(codec.send(request).await?)
```

mirrord-layer/src/file.rs
```diff
@@ -106,3 +117,299 @@ impl OpenOptionsInternalExt for OpenOptionsInternal {
...
+        self.close_queue.push_back(file_channel_tx);
+
+        let close_file_request = CloseFileRequest { fd };
+
+        let request = ClientMessage::FileRequest(FileRequest::Close(close_file_request));
+        codec.send(request).await.map_err(From::from)
+    }
+}
+type FileResult<T> = Result<T, ResponseError>;
+type ResponseChannel<T> = oneshot::Sender<FileResult<T>>;
```
meowjesty, line 363: Move these type aliases next to the other type alias up above.

mirrord-layer/src/file.rs
```diff
@@ -106,3 +117,299 @@ impl OpenOptionsInternalExt for OpenOptionsInternal {
             })
     }
 }
+
+type ResponseDeque<T> = VecDeque<ResponseChannel<T>>;
```
meowjesty, line 121: Kind of an extreme type idea, but maybe adding a marker trait here to guarantee that this cannot be used with any type would be nice?

Illustration:

```rust
trait FileOperation {}

type ResponseDeque<T: FileOperation> = ...;

impl FileOperation for OpenFileResponse {}

...
```

This would prevent possible (but very unlikely) misuses of the type.

meowjesty (commented):
Have only some very minor points to make. Glad that the huge amount of parameters was moved elsewhere.

mirrord-layer/src/file.rs
aviramha: The map felt more cool then wrapping it in an Ok, guess it's a style-decision..

mirrord-layer/src/file.rs
aviramha: I think that would be an overkill and might impact compile time with no real value..

aviramha:
moved here https://github.com/metalbear-co/mirrord/pull/207

PR #209, opened 2022-07-14
Refactor remote result

meowjesty:
Refactored `ResponseError` to try and simplify error handling that should happen when errors are passed from `-agent` to `-layer`.


mirrord-agent/src/file.rs
```diff
@@ -84,248 +84,191 @@ impl FileManager {
...
-            file.write(&write_bytes)
-                .map(|write_amount| {
+            .ok_or(ResponseError::NotFound(fd))
+            .and_then(|remote_file| {
+                if let RemoteFile::File(file) = remote_file {
                     debug!(
-                        "FileManager::write -> wrote {:#?} bytes to fd {:#?}",
-                        write_amount, fd
+                        "FileManager::write -> Trying to write file {:#?}, with bytes {:#?}",
+                        file, write_bytes
```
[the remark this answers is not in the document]
meowjesty: Agree, but this was not an actual change, just moved these and it picked up as so.

Will remove some of these debug messages as part of the PR.

mirrord-layer/src/file/ops.rs
```diff
@@ -188,8 +186,7 @@ pub(crate) fn fdopen(
 pub(crate) fn read(fd: usize, read_amount: usize) -> Result<ReadFileResponse, LayerError> {
     debug!("read -> trying to read valid file {:?}.", fd);
 
-    let (file_channel_tx, file_channel_rx) =
-        oneshot::channel::<Result<ReadFileResponse, ResponseError>>();
+    let (file_channel_tx, file_channel_rx) = oneshot::channel::<RemoteResult<ReadFileResponse>>();
```
aviramha, line 189: Nitpick: no need for explicit type IMO.

mirrord-layer/src/file/ops.rs
```diff
@@ -100,8 +99,7 @@ pub(crate) fn openat(
         "openat -> trying to open valid file {:?} with relative dir {:?}.",
         path, relative_fd
     );
-    let (file_channel_tx, file_channel_rx) =
-        oneshot::channel::<Result<OpenFileResponse, ResponseError>>();
+    let (file_channel_tx, file_channel_rx) = oneshot::channel::<RemoteResult<OpenFileResponse>>();
```
aviramha, line 102: Nitpick: no need for explicit type IMO.

mirrord-layer/src/lib.rs
```diff
@@ -119,19 +115,16 @@ async fn handle_hook_message(
                 service,
                 hints,
             });
-            let codec_result = codec.send(request).await;
-
-            trace!("HookMessage::GetAddrInfo codec_result {:#?}", codec_result);
+            let _codec_result = codec.send(request).await;
```
aviramha, line 118: why not `?` it

meowjesty: This would change behavior, as we're currently just "eating" the error (note how these functions don't return `Result`).

Should I go for it (refactor `handle_x` to deal with these `codec.send` errors)?

aviramha (approved):

PR #211, opened 2022-07-15
Last layer refactor I promise

aviramha:
- Refactored `mirrord-layer/lib.rs` - no more passsing many arguments
- Changed `addrinfo` to `VecDeque` - fixes a potential bug (loss of order)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Use Kubernetes beta feature `Ephemeral Containers` to mirror traffic with the `-
 - Refactored error handling in mirrord-layer.
 - E2E: Collect minikube logs and fix collecting container logs
 - E2E: macOS use colima instead of minikube.
-- Refactored `mirrord-layer/lib.rs` - no more passsing many arguments! :)
+- Refactored `mirrord-layer/lib.rs` - no more passing many arguments! :)
 
 ### Fixed
 - Handle unwraps in fileops to gracefully exit and enable python fileops tests.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,11 @@ Use Kubernetes beta feature `Ephemeral Containers` to mirror traffic with the `-
 - Refactored error handling in mirrord-layer.
 - E2E: Collect minikube logs and fix collecting container logs
 - E2E: macOS use colima instead of minikube.
+- Refactored `mirrord-layer/lib.rs` - no more passsing many arguments! :)
 
 ### Fixed
 - Handle unwraps in fileops to gracefully exit and enable python fileops tests.
+- Changed `addrinfo` to `VecDeque` - fixes a potential bug (loss of order)
 
 ## 2.4.1
 
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -86,7 +86,16 @@ where
     pub codec: actix_codec::Framed<T, ClientCodec>,
     ping: bool,
     tcp_mirror_handler: TcpMirrorHandler,
+    // TODO: Starting to think about a better abstraction over this whole mess. File operations are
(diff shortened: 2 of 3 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -80,89 +79,106 @@ fn init() {
...
+    }
 
-            let request = ClientMessage::GetAddrInfoRequest(GetAddrInfoRequest {
+    async fn handle_hook_message(&mut self, hook_message: HookMessage) {
+        match hook_message {
+            HookMessage::Tcp(message) => {
+                self.tcp_mirror_handler
+                    .handle_hook_message(message, &mut self.codec)
+                    .await
+                    .unwrap();
```
infiniteregrets, line 113: Remove unwrap?

infiniteregrets (commented):
We should probably change from unit to a result maybe

aviramha:
> We should probably change from unit to a result maybe

What do you mean?

infiniteregrets:
@aviramha 
I meant, `handle_hook_message` could probably return a Result so we can propagate errors.. and remove unwraps, not sure if that's to be done here. Sorry for not providing more context before!

mirrord-layer/src/lib.rs
```diff
@@ -80,89 +79,106 @@ fn init() {
...
+    ping: bool,
+    tcp_mirror_handler: TcpMirrorHandler,
+    file_handler: FileHandler,
+    getaddrinfo_handler_queue: VecDeque<ResponseChannel<Vec<AddrInfoInternal>>>,
+}
 
-            getaddrinfo_handler.lock().unwrap().push(hook_channel_tx);
+impl<T> Layer<T>
+where
+    T: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin + Send,
```
meowjesty, line 95: Do we need this `T` generic? Don't we create only 1 type of `Layer<T>` (if you can see us needing a different `T` in the future, it's ok)?

mirrord-layer/src/lib.rs
```diff
@@ -80,89 +79,106 @@ fn init() {
...
+{
+    fn new(codec: actix_codec::Framed<T, ClientCodec>) -> Layer<T> {
```
meowjesty, line 97: This function is only ever called once, with one type of `codec`, so I would suggest moving it to a `Default` trait implementation.

meowjesty (changes requested):
Unless I'm misunderstanding something, `Layer` doesn't need to be generic, other than that it's looking good.

mirrord-layer/src/lib.rs
meowjesty: Forget about this suggestion, `codec` cannot be made with default arguments, as it takes a configurable `port`.

mirrord-layer/src/lib.rs
meowjesty: Forgot about the pain around the `Frame(whatever)` type.

aviramha:
> @aviramha I meant, `handle_hook_message` could probably return a Result so we can propagate errors.. and remove unwraps, not sure if that's to be done here. Sorry for not providing more context before!

Yeah that can be a nice improvement but let's leave it out of scope for now and maybe let new contributors do it as it's fairly easy and not very "important"

CHANGELOG.md
```diff
@@ -22,9 +22,11 @@ Use Kubernetes beta feature `Ephemeral Containers` to mirror traffic with the `-
 - Refactored error handling in mirrord-layer.
 - E2E: Collect minikube logs and fix collecting container logs
 - E2E: macOS use colima instead of minikube.
+- Refactored `mirrord-layer/lib.rs` - no more passsing many arguments! :)
```
eyalb181, line 25: passing**

eyalb181 (approved):

PR #215, opened 2022-07-25
Traffic steal

aviramha:


mirrord-layer/src/tcp_mirror.rs
```diff
@@ -177,4 +178,41 @@ impl TcpHandler for TcpMirrorHandler {
...
+            ClientCodec,
+        >,
+    ) -> Result<(), LayerError> {
+        codec
+            .send(ClientMessage::Tcp(LayerTcp::PortUnsubscribe(close.port)))
+            .await
+            .map_err(From::from)
+    }
+
+    async fn handle_listen(
```
eyalb181, line 196: handle_listen and handle_listen_close are pretty much the same in tcp_steal and tcp_mirror, I'd try to deduplicate with generics

eyalb181 (changes requested):
Made a few minor comments. Additionally, we should try and reduce the code duplication between tcp_steal and tcp_mirror in the layer (from what I understand they're essentially the same with different enum types). I wouldn't necessarily block the PR on it, but if you decide not to handle it please open a separate issue.

mirrord-agent/src/main.rs
```diff
@@ -294,6 +317,11 @@ impl ClientConnectionHandler {
             LayerTcp::PortUnsubscribe(port) => self.tcp_sniffer_api.port_unsubscribe(port).await,
         }
     }
+
+    async fn handle_client_steal_tcp(&mut self, message: LayerStealTcp) -> Result<(), AgentError> {
```
meowjesty, line 321: Is this a WIP? Otherwise why not have the `send` call directly in `ClientMessage::TcpSteal` handling?

mirrord-agent/src/steal.rs
```diff
@@ -0,0 +1,374 @@
...
+
+struct SafeIpTables {
+    inner: iptables::IPTables,
+    chain_name: String,
+}
+
+const IPTABLES_TABLE_NAME: &str = "nat";
+
+impl SafeIpTables {
+    pub fn new() -> Result<Self> {
```
meowjesty, line 36: Add some docs here specifying which flags/configuration are being used for iptables (no need to say what they do, I think it's enough help just so that no one has to read the function implementation just to get this info).

mirrord-agent/src/steal.rs
```diff
@@ -0,0 +1,374 @@
...
+        )
+        .map_err(|e| AgentError::IPTablesError(e.to_string()))?;
+        Ok(Self {
+            inner: ipt,
+            chain_name,
+        })
+    }
+
+    pub fn add_redirect(&mut self, redirected_port: Port, target_port: Port) -> Result<()> {
+        let rule = format!(
```
meowjesty, line 57: This is the same `rule` in `remove_redirect`, so maybe extract it and share? I would suggest a helper function that returns the rule `String`.

mirrord-layer/src/config.rs
```diff
@@ -29,6 +29,9 @@ pub struct LayerConfig {
     #[envconfig(from = "MIRRORD_AGENT_TTL", default = "0")]
     pub agent_ttl: u16,
 
+    #[envconfig(from = "MIRRORD_AGENT_STEAL_TRAFFIC", default = "false")]
```
meowjesty, line 32: Hmm maybe add consistency between names, and have it be `TCP_STEAL_TRAFFIC` (also apply the same namimg everywhere).

mirrord-agent/src/steal.rs
```diff
@@ -0,0 +1,374 @@
+use std::{
```
meowjesty, line 1: Have this module be under a `tcp` module, so we'll end up having a `tcp::sniffer`, `tcp::steal`, and `tcp::outgoing` (or something similar).

With this, should probably move some of the more fundamental (and shared) types to the `tcp.rs` file itself.

mirrord-layer/src/lib.rs
```diff
@@ -42,6 +43,7 @@ mod pod_api;
 mod socket;
 mod tcp;
 mod tcp_mirror;
+mod tcp_steal;
```
meowjesty, line 46: Same module suggestion I've given. It would be nicer to have a `tcp` module, and put these `tcp_` under it (also removing the `tcp_` part from names).

meowjesty (changes requested):
Initial review, still gotta look a bit deeper.

aviramha:
Thanks for the comments @meowjesty ! There's a lot of room to improve on the names and design, feel free to suggest more changes (will address the current comments asap)

mirrord-agent/src/steal.rs
aviramha: that's a good suggestion, but I think very soon we'll have the logic for UDP and it'll be part of the same module..?

meowjesty: Makes sense, if they're going to be sharing the same `struct`s and functions then leaving as-is is fine. Maybe when we start working on udp and the differences begin to appear (if any) we can refactor as modules then.

mirrord-agent/Cargo.toml
```diff
@@ -35,7 +35,11 @@ thiserror.workspace = true
 dns-lookup.workspace = true
 num-traits = "0.2"
 bollard = "0.13"
-tokio-util = "0.7"
+tokio-util.workspace = true
+rand.workspace = true
+streammap-ext.workspace = true
+iptables = "0.5.0"
```
meowjesty, line 41: Just `0.5` is probably fine.

mirrord-layer/src/config.rs
```diff
@@ -29,6 +29,9 @@ pub struct LayerConfig {
     #[envconfig(from = "MIRRORD_AGENT_TTL", default = "0")]
     pub agent_ttl: u16,
 
+    #[envconfig(from = "MIRRORD_AGENT_TCP_STEAL_TRAFFIC", default = "false")]
+    pub agent_steal_tcp_traffic: bool,
```
meowjesty, line 33: `agent_tcp_steal_traffic` to be consistent with the envconfig. Or change the envconfig.

mirrord-protocol/src/tcp.rs
```diff
@@ -38,3 +38,12 @@ pub enum DaemonTcp {
     Data(TcpData),
     Close(TcpClose),
 }
+
+/// Messages related to Steal Tcp handler from client.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum LayerStealTcp {
```
meowjesty, line 44: Consistency, the `DaemonMessage` is `TcpSteal`, so change this to `LayerTcpSteal`.

mirrord-protocol/src/codec.rs
```diff
@@ -260,6 +261,7 @@ impl From<dns_lookup::AddrInfo> for AddrInfoInternal {
 pub enum DaemonMessage {
     Close,
     Tcp(DaemonTcp),
+    StealTcp(DaemonTcp),
```
meowjesty, line 264: `TcpSteal`

mirrord-protocol/src/steal.rs
```diff
@@ -0,0 +1,16 @@
...
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum StealClientMessage {
+    PortSteal(Port),
+    TCPData(TCPData),
+    CloseConnection(TCPClose),
+}
+
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum StealDaemonMessage {
+    NewConnection(NewTCPConnection),
```
meowjesty, line 13: `TcpNewConnection`

mirrord-agent/src/steal.rs
```diff
@@ -0,0 +1,407 @@
...
+            read_streams: StreamMap::default(),
+            connection_index: 0,
+        })
+    }
+
+    pub async fn handle_loop(
+        &mut self,
+        mut rx: Receiver<LayerStealTcp>,
+        mut listener: TcpListener,
+    ) -> Result<()> {
```
meowjesty, line 135: Is this time for using [`ControlFlow`](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html) or should we leave it for a refactor in the future (I would vote for a future refactor)?

meowjesty (approved):
Mostly rename some stuff for consistency's sake.

mirrord-agent/src/steal.rs
aviramha: not sure why control flow is relevant here though - the loop is inside the function?

mirrord-protocol/src/lib.rs
```diff
@@ -10,7 +10,7 @@ use std::{collections::HashSet, ops::Deref};
 pub use codec::*;
 pub use error::*;
 
-pub type ConnectionID = u16;
+pub type ConnectionID = u64;
```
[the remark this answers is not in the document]
aviramha: I wanted to dissolve the "reuse mechanism" so u64 (no chance of conflict)

mirrord-layer/src/lib.rs
```diff
@@ -174,29 +176,42 @@ where
     // Stores a list of `oneshot`s that communicates with the hook side (send a message from -layer
     // to -agent, and when we receive a message from -agent to -layer).
     getaddrinfo_handler_queue: VecDeque<ResponseChannel<Vec<AddrInfoInternal>>>,
+
+    pub tcp_steal_handler: TcpStealHandler,
```
DmitryDodzin, line 180: how about unifying `tcp_mirror_handler` & `tcp_steal_handler` to `tcp_handler`? under something like `Box<dyn TcpHandler>` since both implement `TcpHandler` it can be better than initializing both since only one will be in use

aviramha: That's actually something I tried to do but wasn't sure how exactly (not very experienced with `dyn`). The reason I didn't put any more effort into making it like that is very soon it wont be either, but both based on configurations so no reason to bother so hard ;)

eyalb181 (approved):

mirrord-layer/src/pod_api.rs
```diff
@@ -209,6 +209,12 @@ async fn create_ephemeral_container_agent(
     let ephemeral_container: EphemeralContainer = serde_json::from_value(json!({
         "name": mirrord_agent_name,
         "image": agent_image,
+        "securityContext": {
```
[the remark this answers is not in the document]
aviramha: mm yeah we can do that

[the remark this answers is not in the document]
aviramha: thinking about it I think we should leave as is as it only complicates.. maybe we'll change it later on.

PR #216, opened 2022-07-27
remove ephemeral containers feature

aviramha:
Few reasons we decided to remove for now:

1. spec cluttering
2. can't support all features (filesystem, traffic stealing) using current feature

### DIFF
--- a/tests/src/sanity.rs
+++ b/tests/src/sanity.rs
@@ -77,8 +77,6 @@ mod tests {
         GoHTTP,
     }
     pub enum Agent {
-        #[cfg(target_os = "linux")]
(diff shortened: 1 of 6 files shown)


tests/src/sanity.rs
```diff
@@ -77,8 +77,6 @@ mod tests {
         GoHTTP,
     }
     pub enum Agent {
```
eyalb181, line 79: Looks like this enum can be removed entirely

eyalb181 (changes requested):

eyalb181 (approved):

PR #219, opened 2022-07-27
Restore ephemeral containers feature

infiniteregrets:
- revert the previous commit that removed the feature 
- change path to `proc/pic/root` for fs access when pid is None
- allow fs tests to run if the default path exists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 
 ### Added
-- Add a flag for the agent, `--ephemeral`, to correctly refer to the filesystem i.e. refer to filesystem path as `/proc/1/root` when the flag is on, otherwise `/`.
+- Add a flag for the agent, `--ephemeral-container`, to correctly refer to the filesystem i.e. refer to root path as `/proc/1/root` when the flag is on, otherwise `/`.
 
 ### Changed
 - Assign a random port number instead of `61337`. (Reason: A forking process creates multiple agents sending traffic on the same port, causing addrinuse error.)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,12 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ### Added
 - Add a flag for the agent, `--ephemeral`, to correctly refer to the filesystem i.e. refer to filesystem path as `/proc/1/root` when the flag is on, otherwise `/`.
 
+### Changed
+- Assign a random port number instead of `61337`. (Reason: A forking process creates multiple agents sending traffic on the same port, causing addrinuse error.)
+
 ### Fixed
 - Fix filesystem tests to only run if the default path exists.
+
 
 ## 2.5.0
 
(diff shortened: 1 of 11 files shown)


mirrord-agent/src/file.rs
```diff
@@ -76,7 +76,7 @@ impl FileManager {
     pub fn new(pid: Option<u64>) -> Self {
         let root_path = match pid {
             Some(pid) => PathBuf::from("/proc").join(pid.to_string()).join("root"),
-            None => PathBuf::from("/"),
+            None => PathBuf::from("/proc").join("1").join("root"),
```
aviramha, line 79: I don't like this - I prefer that when it's ephermal container we just pass pid 1 as the pid..

aviramha (changes requested):
Small change

mirrord-agent/src/file.rs
```diff
@@ -73,10 +73,16 @@ impl FileManager {
         }
     }
 
-    pub fn new(pid: Option<u64>) -> Self {
+    pub fn new(pid: Option<u64>, ephemeral: bool) -> Self {
```
aviramha, line 76: I don't think it's needed; just set `pid` to `1` on ephemeral from the outside.

mirrord-layer/src/lib.rs
```diff
@@ -51,6 +52,9 @@ pub static mut HOOK_SENDER: Option<Sender<HookMessage>> = None;
 
 pub static ENABLED_FILE_OPS: OnceLock<bool> = OnceLock::new();
 
+pub static CONNECTION_PORT: LazyLock<u16> =
```
aviramha, line 55: why does it need to be global? just pass it.

aviramha (changes requested):

CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
...
 ## [Unreleased]
 
-### Removed
-- Removed Ephemeral Containers until they become better supported.
+
+### Added
+- Add a flag for the agent, `--ephemeral`, to correctly refer to the filesystem i.e. refer to filesystem path as `/proc/1/root` when the flag is on, otherwise `/`.
+
+### Fixed
+- Fix filesystem tests to only run if the default path exists.
```
aviramha, line 16: add port random

mirrord-agent/src/main.rs
```diff
@@ -183,10 +183,20 @@ impl ClientConnectionHandler {
...
         stream: TcpStream,
         pid: Option<u64>,
+        ephemeral: bool,
         sniffer_command_sender: Sender<SnifferCommand>,
         cancel_token: CancellationToken,
     ) -> Result<(), AgentError> {
-        let file_manager = FileManager::new(pid);
+        let file_manager = match pid {
+            Some(_) => FileManager::new(pid),
+            None => {
```
aviramha, line 192: I think this can be formatted better (https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html)
`None if ephemeral`
`None if !ephemeral`

aviramha (approved):
Small nitpick. LGTM! Thanks @infiniteregrets

aviramha:
Great, Thank you!

PR #226, opened 2022-07-29
Groundwork for original/detour handling.

meowjesty:
1. Added a proc_macro `hook_fn` to create types based on detour signatures;
2. Changed `try_hook` and `hook` macros to return the original `libc` function (now we only have 1 macro called `replace`);
3. Refactored `socket` module in layer to be more `Result` friendly.


mirrord-macro/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+use proc_macro2::Span;
+use quote::quote;
+use syn::{Ident, ItemFn};
+
+#[proc_macro_attribute]
```
aviramha, line 5: I suggest to add a docstring that describes input -> output for clarity (I really hate proc macro syntax :))

aviramha (approved):
Small nitpicks but wow looks great! well played. 
Regarding the `hook_fn` I was thinking of a possible design to have a derive macro that operates on struct, but we'll discuss it in chat :)

aviramha (changes requested):
scratch that

aviramha (approved):

PR #227, opened 2022-07-30
Fix dns test

### LINKED ISSUE #224: Fix test_remote_dns_enabled_works E2E
@abhirag noticed that in the test we don't run the correct script which is weird that it still succeeds. Need to understand why the test succeeds and fix the test.

infiniteregrets:
We were running the test with `-d true` which is probably the reason it was not erroring out
Closes #224

### DIFF
--- a/tests/src/sanity.rs
+++ b/tests/src/sanity.rs
@@ -638,9 +638,9 @@ mod tests {
         let service = service.await;
         let node_command = vec![
             "node",
-            "node-e2e/remote_dns/test_remote_dns_eneabled_works.mjs",
+            "node-e2e/remote_dns/test_remote_dns_enabled_works.mjs",
         ];
-        let mirrord_args = vec!["-d", "true"];
+        let mirrord_args = vec!["-d"];
         let mut process = run(node_command, &service.pod_name, None, Some(mirrord_args)).await;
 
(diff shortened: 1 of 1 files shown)


infiniteregrets:
@aviramha On a side note while testing in a different branch, I noticed that we are creating temporary directories twice for fileops test. If I remove `get_shared_lib_path` function, the tests get very flaky

aviramha:
> @aviramha On a side note while testing in a different branch, I noticed that we are creating temporary directories twice for fileops test. If I remove `get_shared_lib_path` function, the tests get very flaky

I think it's intended? Why wouldn't we want different temporary directories?

aviramha (approved):

aviramha:
Took the liberty of reverting the random directory name change. Merging. Thanks!

infiniteregrets:
@aviramha are we not already doing that here?
```rs
let temp_dir = tempdir::TempDir::new("test").unwrap();
        let mut mirrord_args = vec![
            "exec",
            "--pod-name",
            pod_name,
            "-c",
            "--extract-path",
            temp_dir.path().to_str().unwrap(),
        ];
```

aviramha:
Woops. You're correct - but this might re-use the same path (atleast from what I've seen) so maybe we should support providing it as a parameter to `run`

PR #237, opened 2022-08-04
reduce e2e flakiness

aviramha:
Add `Subscribed` message sent from Daemon to Layer so Layer can now wait on this message in tests.

### DIFF
--- a/mirrord-protocol/src/tcp.rs
+++ b/mirrord-protocol/src/tcp.rs
@@ -37,4 +37,6 @@ pub enum DaemonTcp {
     NewConnection(NewTcpConnection),
     Data(TcpData),
     Close(TcpClose),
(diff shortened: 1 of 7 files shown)


mirrord-protocol/src/tcp.rs
```diff
@@ -37,4 +37,6 @@ pub enum DaemonTcp {
     NewConnection(NewTcpConnection),
     Data(TcpData),
     Close(TcpClose),
+    /// Used to notify the subscription occured
```
eyalb181, line 40: Since this doesn't serve any actual function in the product, I would also document why it exists

eyalb181 (changes requested):

eyalb181 (approved):

PR #239, opened 2022-08-04
Outgoing traffic

meowjesty:


mirrord-agent/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,228 @@
...
+        trace!("intercept_task -> id {:#?}", connection_id);
+
+        let mut buffer = vec![0; 1500];
+        let (remote_reader, mut remote_writer) = stream.into_split();
+
+        loop {
+            select! {
+                biased;
+
+                readable = remote_reader.readable() => {
```
aviramha, line 78: I think you can use `AsyncReadExt` to avoid the readable, match flow (as it does it for you)

mirrord-layer/src/socket/hooks.rs
```diff
@@ -24,12 +24,44 @@ pub(crate) unsafe extern "C" fn socket_detour(
...
+        debug!("socket_detour -> bypassed");
+        socket_result
+    } else {
+        let (Ok(result) | Err(result)) =
+            socket(socket_result, domain, type_, protocol).map_err(From::from);
+        result
+    }
+}
+
+#[hook_fn]
```
aviramha, line 38: I don't think we need this hook?

meowjesty: Removed, this was just something that I was investigating.

mirrord-agent/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,177 @@
...
+                // [layer] -> [agent]
+                request = request_rx.recv() => {
+                    trace!("interceptor_task -> request {:?}", request);
+
+                    match request {
+                        Some(request) => {
+                            match request {
+                                // [user] -> [layer] -> [agent] -> [layer]
+                                // `user` is asking us to connect to some remote host.
+                                TcpOutgoingRequest::Connect(ConnectRequest { remote_address }) => {
```
aviramha, line 80: I think the Request/Response terminology might be confusing (of the enum, `ConnectRequest` is fine). I'd call it `LayerTcpOutgoing` and `DaemonTcpOutgoing` or smth like that?

mirrord-agent/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,177 @@
...
+                                    {
+                                        Ok(writer) => writer
+                                            .write_all(&bytes)
+                                            .await
+                                            .map_err(ResponseError::from)
+                                            .map(|()| WriteResponse { connection_id }),
+                                        Err(fail) => Err(fail),
+                                    };
+
+                                    let response = TcpOutgoingResponse::Write(write_response);
```
aviramha, line 126: why do you write back write_response? if it fails to write socket will be terminated anyway?

meowjesty: We can get away without it, as you pointed out to me, but as I had already implemented this message flow, why remove it?

mirrord-agent/src/tcp/outgoing.rs
meowjesty: Sure, I wanted to unify terminology for layer -> agent, and agent -> layer messages. As I was using this request/response for the features I've been implementing, so I went ahead and used the same here.

If we're going to change it to be layer/daemon instead of request/response, it's fine for me, but we should change it everywhere as well.

meowjesty: Changed it to be Layer/Daemon, instead of Request/Response.

mirrord-layer/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,308 @@
...
+                            warn!("interceptor_task -> exiting due to local stream closed!");
+                            break;
+                        },
+                        Ok(read_amount) => {
+                            // Sends the message that the user wrote to our interceptor socket to
+                            // be handled on the `agent`, where it'll be forwarded to the remote.
+                            let write = Write { connection_id, bytes: buffer[..read_amount].to_vec() };
+                            let outgoing_data = TcpOutgoing::Write(write);
+
+                            if let Err(fail) = send_hook_message(HookMessage::TcpOutgoing(outgoing_data)).await {
```
aviramha, line 125: Why is this hook message? We're running from the `mirrord-layer` `Runtime` as a task here (AFAIK) - no need to use "bad" synch mechanisms - just pass a sender to this task from the main thread and bridge it.

aviramha: (Hook messages should be used from hook context, not tokio..)

mirrord-layer/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,308 @@
...
+        trace!("handle_hook_message -> message {:?}", message);
+
+        match message {
+            TcpOutgoing::Connect(Connect {
+                remote_address,
+                channel_tx,
+            }) => {
+                trace!("Connect -> remote_address {:#?}", remote_address);
+
+                self.connect_queue.push_back(channel_tx);
```
aviramha, line 182: Starting to get afraid by the way we do this queues instead of some sort of HashMap (there's an odd case for a race condition probably right now). Nothing that should be done right now, just a thought on the first thing we should refactor - some generic API for passing Request/Response between layer/agent that is ID based for avoiding evil

mirrord-layer/src/lib.rs
```diff
@@ -386,33 +364,56 @@ fn enable_hooks(enabled_file_ops: bool, enabled_remote_dns: bool) {
     interceptor.end_transaction();
 }
 
+// TODO: When this is annotated with `hook_guard_fn`, then the outgoing sockets never call it (we
+// just bypass). Everything works, so, should we intervene?
+//
```
aviramha, line 369: As far as I follow (AFAIF?) close will close our "proxy" socket, which will then `layer` will see it in the `interceptor_task` and then will notify the remote agent. sounds good. maybe write this flow?

mirrord-layer/src/lib.rs
```diff
@@ -222,6 +180,20 @@ where
...
+                .await
+                .unwrap(),
+            HookMessage::Exit(message) => {
+                trace!("HookMessage::Exit");
+
+                let request = ClientMessage::ExitRequest;
+
+                self.codec.send(request).await.unwrap();
+
+                message.hook_channel_tx.send(Ok(())).unwrap();
```
aviramha, line 195: break loop for graceful exit here

aviramha: ah we need to send a message when we close.. ermm.. we could call exit on our own (bypassing the hook becuase of our bypass) when loop ends..?

tests/node-e2e/outgoing/test_outgoing_traffic_make_request_after_listen.mjs
```diff
@@ -0,0 +1,84 @@
+import https from "node:https";
```
aviramha, line 1: one fear I have for all tests added - how do you know it's actually hooked? like, if layer wouldn't load this test will still pass. not sure if worth investing, but maybe add a comment for now.

aviramha (changes requested):
Overall looks great - awesome job.
Generally I saw some changes that were peripheral to the feature, so I'd suggest to break those into more PRs next time for easier review (might have missed something in those areas because hard to follow if related to the "big change" or not) 
See inline comments.

mirrord-layer/src/lib.rs
```diff
@@ -146,6 +96,7 @@ fn init() {
     };
 
     let enabled_file_ops = ENABLED_FILE_OPS.get_or_init(|| config.enabled_file_ops);
+    let _ = ENABLED_TCP_OUTGOING.get_or_init(|| config.enabled_tcp_outgoing);
```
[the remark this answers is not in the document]
meowjesty: This looks like the better init function for `OnceLock`, the other being [`set`](https://doc.rust-lang.org/nightly/std/sync/struct.OnceLock.html#method.set) which may block.

mirrord-agent/src/tcp/outgoing.rs
```diff
@@ -0,0 +1,195 @@
...
+                            let daemon_read = read
+                                .map_err(ResponseError::from)
+                                .map(|bytes| DaemonRead { connection_id, bytes: bytes.to_vec() });
+
+                            let daemon_message = DaemonTcpOutgoing::Read(daemon_read);
+                            daemon_tx.send(daemon_message).await?
+                        }
+                        None => {
+                            trace!("interceptor_task -> close connection {:#?}", connection_id);
+                            let daemon_message = DaemonTcpOutgoing::Close(connection_id);
```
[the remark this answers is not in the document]
meowjesty: Manually keeping in sync now, added a comment to improve this in the future.

mirrord-agent/src/sniffer.rs
```diff
@@ -8,7 +8,7 @@ use std::{
 use futures::StreamExt;
 use mirrord_protocol::{
     tcp::{DaemonTcp, NewTcpConnection, TcpClose, TcpData},
-    ConnectionID, Port,
+    ConnectionId, Port,
```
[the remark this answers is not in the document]
meowjesty: Yeah, my bad! I had added a `ConnectionId` type alias as well when working on the feature, as the existing one was using a lower uint range (and I missed it for quite a while ...).

mirrord-layer/src/socket/ops.rs
```diff
@@ -196,8 +239,80 @@ pub(super) fn connect(sockfd: RawFd, remote_address: SocketAddr) -> Result<()> {
...
+            if connect_result == -1 && err_code != libc::EINPROGRESS && err_code != libc::EINTR {
+                error!(
+                    "connect -> Failed call to libc::connect with {:#?} errno is {:#?}",
+                    connect_result,
+                    errno::errno()
+                );
+                return Err(io::Error::from_raw_os_error(connect_result).into());
+            }
+
+            let connected = Connected {
```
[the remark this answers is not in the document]
meowjesty: Added a comment there.

mirrord-layer/src/tcp/outgoing.rs
meowjesty: `TODO` added.

tests/node-e2e/outgoing/test_outgoing_traffic_make_request_after_listen.mjs
meowjesty: Makes sense, I was wondering about a nice way of testing these, but only came up with the simple ways of doing it.

One thing I've been thinking that we could do, is have some sort of commands that we can send back and forth between layer and agent, to be used internally by us (mostly for tests). Like how you guys added a phase check for the pod status, but more general.

mirrord-layer/src/lib.rs
meowjesty: Both ways are pretty valid (imo), so:

1. Should I change it now?
2. Keep as-is is fine?
3. Keep as-is, but add a `TODO`?

aviramha: I'm okay with 3, just add a comment.

mirrord-layer/src/tcp/outgoing.rs
meowjesty: Makes sense, changed it to separate channels.

mirrord-agent/src/main.rs
```diff
@@ -220,17 +235,36 @@ impl ClientConnectionHandler {
...
                         error!("tcp sniffer stopped?");
                         break;
                     }
                 }
+                outgoing_response = self.tcp_outgoing_api.daemon_message() => {
+                    match outgoing_response {
+                        Ok(response) => self.respond(DaemonMessage::TcpOutgoing(response)).await?,
+                        Err(fail) => {
+                            match fail {
+                                AgentError::ReceiverClosed => if self.keep_alive {
```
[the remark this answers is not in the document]
meowjesty: It was the easiest way I thought of to keep agent alive when a single request completes. If we don't check for some sort of `keep_alive` thingy, then it was just `break`ing the loop. If I left this case (of not having any requests left) as `continue`, then we would loop forever (encountered both cases while writing the code).

mirrord-layer/src/lib.rs
meowjesty: Added a flow for closing it on `agent`.

tests/node-e2e/outgoing/test_outgoing_traffic_make_request_after_listen.mjs
aviramha: Don't forget to add todo there so it's clear :)

meowjesty: Added a comment about this in `sanity.rs` before the outgoing traffic tests.

mirrord-agent/src/main.rs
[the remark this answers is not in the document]
meowjesty: Removed the `exit` hook.

eyalb181 (approved):

aviramha (approved):

PR #240, opened 2022-08-05
reduce e2e flakiness - increase timeout time

aviramha:


mirrord-layer/src/pod_api.rs
```diff
@@ -210,15 +228,22 @@ async fn create_ephemeral_container_agent(
...
+                    debug!("container not ready yet");
+                }
+            }
             WatchEvent::Error(s) => {
                 error!("Error watching pod: {:?}", s);
                 break;
             }
             _ => {}
         }
     }
```
infiniteregrets, line 245: Can we return LayerError here? Because if it's an error it just breaks out of the loop

tests/src/sanity.rs
```diff
@@ -531,8 +531,12 @@ mod tests {
         let mut process = application
             .run(&service.pod_name, Some(&service.namespace), agent.flag())
             .await;
-        process.wait_for_line(Duration::from_secs(30), "daemon subscribed");
+        process.wait_for_line(Duration::from_secs(120), "daemon subscribed");
```
infiniteregrets, line 534: How come we have to wait 120 seconds(max) for this message?

mirrord-layer/src/pod_api.rs
aviramha: I thought about it and I think in worst case scenario it won't be able to connect to the agent then it will fail.

tests/src/sanity.rs
aviramha: It's just to reduce flakiness - the macOS machines are a bit slow and we run the tests concurrently so sometimes launching the agent and connecting to it takes time, so..

eyalb181 (approved):

CHANGELOG.md
```diff
@@ -7,8 +7,15 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+### Fixed
+- Ephemeral Containers didn't wait for the right condition, leading to timeouts in many cases.
```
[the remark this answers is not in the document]
aviramha: General

PR #242, opened 2022-08-08
Ephemeral container fileops flakiness

aviramha:
agent: add debug print of args to identify runs


infiniteregrets:
Can you add #[trace] and #[notrace]? https://docs.rs/rstest/latest/rstest/attr.rstest.html#trace-input-arguments

aviramha:
> Can you add #[trace] and #[notrace]? https://docs.rs/rstest/latest/rstest/attr.rstest.html#trace-input-arguments

```
641 |     pub async fn test_remote_env_vars_exclude_works(#[future] service: EchoService) {
    |                                                               ^^^^^^^ `impl futures_util::Future<Output = EchoService>` cannot be formatted using `{:?}` because it doesn't implement `Debug`
    |
    ```

**Comment by infiniteregrets** [comment]
> > Can you add #[trace] and #[notrace]? https://docs.rs/rstest/latest/rstest/attr.rstest.html#trace-input-arguments
> 
> ```
> 641 |     pub async fn test_remote_env_vars_exclude_works(#[future] service: EchoService) {
>     |                                                               ^^^^^^^ `impl futures_util::Future<Output = EchoService>` cannot be formatted using `{:?}` because it doesn't implement `Debug`
>     |
>     ```
> ```

#[notrace] before the arguments that don't implement std::fmt::Debug

File: mirrord-layer/src/pod_api.rs [hunk 1]
```diff
@@ -331,12 +331,19 @@ async fn create_job_pod_agent(
 
     while let Some(status) = stream.try_next().await? {
         match status {
-            WatchEvent::Added(_) => break,
+            WatchEvent::Added(pod) | WatchEvent::Modified(pod) => {
+                debug!("found add = {:?}", pod.clone().status.unwrap().phase);
```
aviramha, line 335: no need to clone, also don't unwrap (propagate)

aviramha: use `&pod` and/or `as_ref()` as needed

aviramha (commented):
Thanks for taking over this @infiniteregrets !
Not sure if it's ready but added some comments :)

mirrord-layer/src/pod_api.rs
```diff
@@ -331,12 +330,23 @@ async fn create_job_pod_agent(
 
     while let Some(status) = stream.try_next().await? {
         match status {
-            WatchEvent::Added(_) => break,
+            WatchEvent::Added(pod) | WatchEvent::Modified(pod) => {
+                let pod_status = &pod
+                    .status
+                    .and_then(|status| status.phase)
+                    .ok_or(LayerError::PodStatusNotFound(pod.metadata.name))?;
```
aviramha, line 337: We shouldn't err on this as this can be other pod theoretically (just do next)

infiniteregrets: Not sure what you mean by do next, how do we handle the Option then?

aviramha: Continue if None, probably can be done using `if let Some(m) else continue`

aviramha: I took care of it (I think?)

eyalb181 (approved):

PR #244, opened 2022-08-14
E2E - Unify node host resolution with WSL support (for better windows development experience)

DmitryDodzin:
With the use of wsl detection remove the requirement to use minikube if there is another Kubernetes api provided (Support Docker for Windows just like Docker for Mac)

for better windows support:
- add matching to cargo version "edition" field to rustfmt.toml 
- remove ./ from ./tests/go-e2e/go-e2e for better windows git support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ node_modules
 .DS_Store
 **/.DS_Store
 target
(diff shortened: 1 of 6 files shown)


eyalb181 (approved):
Looks good!

PR #253, opened 2022-08-16
prevent calling detours from our code using DetourGuard

aviramha:
prevent calling detours from our code using DetourGuard


meowjesty (approved):
👍

PR #255, opened 2022-08-16
Fix DNS issue (ai_addr not living long enough). Try to improve tests.

meowjesty:


tests/node-e2e/remote_dns/test_remote_dns_lookup_google.mjs
```diff
@@ -6,6 +7,12 @@ lookup("google.com")
   .then((resolved) => {
     console.log(">> resolved ", resolved);
 
+    assert(
```
aviramha, line 10: This is an arbitrary test as in your case it would result in `255.127.0.0` but for others might have another invalid IP.
I think the best way to address this kind of test is to resolve from the "caller" (the `sanity.rs`) the "truth" value then check if it's the same? another option would be to send a HTTPs request and see that it succeeds.

aviramha: I took the liberty of doing it as this test became flaky (probably because of use after free) in another PR

aviramha (approved):

PR #256, opened 2022-08-17
e2e: linux print contents of log directory on failure and filename fo…

aviramha:
…r each log printed


eyalb181 (approved):

mirrord-agent/src/sniffer.rs
```diff
@@ -470,6 +468,9 @@ impl TCPConnectionSniffer {
             Some(session) => session,
             None => {
                 if !is_new_connection(tcp_flags) {
+                    if tcp_flags == 24 {
+                        debug!("handle_packet -> wiwiwi {:?}", &tcp_packet);
```
eyalb181, line 472: Is the wiwiwi intentional?

eyalb181 (approved):

mirrord-agent/src/sniffer.rs
aviramha: it's a temp log until we find the issue :D

Issue #220, opened 2022-07-28
Remove unused dependencies from mirrord-layer


abhirag:
`mirrord-layer` has the following unused dependencies -- 
```
mirrord-layer v2.5.0 (/home/parallels/Documents/mirrord_build_time/mirrord/mirrord-layer)
└─── dependencies
     ├─── "clap"
     ├─── "multi-map"
     ├─── "nanoid"
     ├─── "nix"
     └─── "queues"
```
- [ ] Verify the analysis using [`cargo-udeps`](https://github.com/est31/cargo-udeps)
- [ ] Remove the dependencies from [`Cargo.toml`](https://github.com/metalbear-co/mirrord/blob/main/mirrord-layer/Cargo.toml)
- [ ] Make sure that the build succeeds

neimsaci:
Hi I am beginner and it looks like this issue is for me! :D

eyalb181:
@neimsaci All yours! I've assigned you the issue, if you need any help feel free to ask here or on [Discord](https://discord.gg/pSKEdmNZcK). Thanks!

aviramha:
oops. We forgot to close this! Fixed in #243

PR #263, opened 2022-08-18
chore(mirrord-layer): remove nix unused lib

developerfred:
ref #220


eyalb181:
Hey @developerfred, thanks for the contribution! However I think @neimsaci just looked into it and nix isn't actually unused. @neimsaci can you confirm?

eyalb181:
I verified that nix is actually in use, so closing this PR. Sorry about that!

Issue #259, opened 2022-08-17
Support Bash pipe/redirect operators


DmitryDodzin:
There is a problem with bash pipe `|` or redirect `>` operators that perform ops between stdouts and stdins of forked process, needs further investigation on the issue.
Some operations fail some just halt the further execution of the process managed by mirrord

possibly related: #258

infiniteregrets:
Just did some digging:
conceptually I believe, pipe operators call pipe() then dup on the pipe fd with the stdout of one end and otherway around on the receiving end of the pipe. However strace does not show any calls to dup, dup2, dup3, fcntl..

https://gist.github.com/infiniteregrets/1838c511792353116d071ee1bfd17c0a

aviramha:
This seems to work now.

PR #274, opened 2022-08-23
Make detours common for dup, dup2, dup3, & fcntl

### LINKED ISSUE #258: Support dup for file-ops
Support "Duplicating a file descriptor" commands for file-ops and not only for sockets.

Syscalls like `dup`, `dup2`, `dup3` and `fcntl` can duplicate duplicate a file descriptor, if the fd is one that mirrord manages then post duplication the new fd still should be mapped to the remote-fd

infiniteregrets:
Will close #258 and #259


mirrord-layer/src/common.rs
```diff
@@ -29,3 +38,172 @@ pub enum HookMessage {
     File(HookMessageFile),
     GetAddrInfoHook(GetAddrInfoHook),
 }
+
+mod common_ops {
```
meowjesty, line 44: I suggest moving these modules under a separate thing, this `common` module name is already not very helpful. Maybe have a `common_detour` folder/module, and inside have the `hooks` and `ops`, so we access those as `common_detour::hooks` and `common_detour::ops`.

mirrord-layer/src/common.rs
```diff
@@ -29,3 +40,172 @@ pub enum HookMessage {
...
+    use std::{os::unix::prelude::RawFd, sync::OnceLock};
+
+    use libc::c_int;
+
```
meowjesty, line 48: Why do you need this here? Can't you use the already global `ENABLED_FILE_OPS` we have?

meowjesty (commented):
Some initial thoughts on the code.

mirrord-layer/src/common_hooks.rs
```diff
@@ -0,0 +1,154 @@
...
+        trace!("fcntl_detour -> fd {:#?} | cmd {:#?}", fd, cmd);
+
+        let arg = arg.arg::<usize>();
+        let fcntl_result = FN_FCNTL(fd, cmd, arg);
+        let guard = crate::DetourGuard::new();
+        if guard.is_none() {
+            return fcntl_result;
+        }
+
+        if fcntl_result == -1 {
```
eyalb181, line 59: Why not combine it with the above condition? Or even do it before, so you don't needlessly initialize a DetourGuard if fcntl_result is -1 anyway

mirrord-layer/src/common_hooks.rs
```diff
@@ -0,0 +1,154 @@
...
+    use crate::error::LayerError;
+
+    /// https://github.com/metalbear-co/mirrord/issues/184
+    #[hook_fn]
+    pub(super) unsafe extern "C" fn fcntl_detour(fd: c_int, cmd: c_int, mut arg: ...) -> c_int {
+        trace!("fcntl_detour -> fd {:#?} | cmd {:#?}", fd, cmd);
+
+        let arg = arg.arg::<usize>();
+        let fcntl_result = FN_FCNTL(fd, cmd, arg);
+        let guard = crate::DetourGuard::new();
```
eyalb181, line 54: Why is DetourGuard used here but not in e.g. dup?

infiniteregrets:
stale pr, closing this will open a new one once mvp is done

PR #275, opened 2022-08-24
mirrord-layer: Added common Result type, removed dependency of anyhow…

aviramha:
… crate

### DIFF
--- a/mirrord-layer/src/error.rs
+++ b/mirrord-layer/src/error.rs
@@ -116,6 +116,9 @@ pub(crate) enum LayerError {
 
     #[error("mirrord-layer: Null pointer found!")]
     NullPointer,
+
+    #[error("mirrord-layer: Container `{0}` not found in namespace `{1}` pod `{2}`")]
+    ContainerNotFound(String, String, String),
 }
 
 // Cannot have a generic From<T> implementation for this error, so explicitly implemented here.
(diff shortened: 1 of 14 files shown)


DmitryDodzin (approved):
Looking good.
NTH change for the macro If we want to keep it exported

mirrord-layer/src/macros.rs
```diff
@@ -4,11 +4,12 @@ macro_rules! replace {
         let intercept = |interceptor: &mut frida_gum::interceptor::Interceptor,
                          symbol_name,
                          detour: $detour_type|
-         -> Result<$detour_type, LayerError> {
+         -> crate::error::Result<$detour_type> {
```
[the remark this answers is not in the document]
aviramha: Thanks!

mirrord-layer/src/error.rs
```diff
@@ -125,6 +128,8 @@ impl<'a, T> From<std::sync::PoisonError<std::sync::MutexGuard<'a, T>>> for Layer
     }
 }
 
+pub(crate) type Result<T> = std::result::Result<T, LayerError>;
```
eyalb181, line 131: Is it good practice to call it Result? I'm assuming most people going over the code would automatically assume it's std::Result and not realize there's a LayerError hidden there.

aviramha: AFAIK It's more than common. https://doc.rust-lang.org/std/io/type.Result.html for example - many modules have their own `Result` aliases, and so do we.

aviramha: P.S They can know it's not `std::Result` by seeing that only one generic is specified, and not two.

eyalb181 (approved):

PR #276, opened 2022-08-24
mirrord-layer: Split error LayerError Enum into two

### LINKED ISSUE #247: Split mirrord-layer LayerError based on criteria
Right now every `LayerError` variant we add needs to have a matching libc return value, while not all errors are even possible from the hooks, so adding the libc return value is futile.
I think we can solve this in two ways:
1. Have two kinds of `LayerError`s, one that can occur from hooks and one that can occur from the regular runtime.
2. Add a tag for specific `LayerError`s to require adding match, so when we create an error we mark it.

aviramha:
Closes #247


meowjesty (approved):
It's been a long time coming. 👍

PR #291, opened 2022-08-29
metalbear-cloud login

DmitryDodzin:
Add login feature to mirrord-cli

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 
+### Added
+- mirrord-cli add login command for loggin in to metalbear-cloud
+
 ## 2.9.1
 ### Fixed
 - CI - set typescript version at 4.7.4 to fix broken release action
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ## [Unreleased]
 
 ### Added
-- mirrord-cli add login command for loggin in to metalbear-cloud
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+### Added
+- mirrord-cli add login command for loggin in to metalbear-cloud
```
aviramha, line 11: typo `logging`

mirrord-auth/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[error(transparent)]
+    ConfigParseError(#[from] serde_json::Error),
+    #[error(transparent)]
+    #[cfg(feature = "webbrowser")]
+    ConfigRequestError(#[from] reqwest::Error),
+}
+
+type Result<T> = std::result::Result<T, AuthenticationError>;
+
+impl AuthConfig {
```
aviramha, line 48: Using of a raw file is quite "dangerous" - I know that many apps do it, but if it's possible to use a crate such as this https://github.com/hwchen/keyring-rs then let's do it (if not, open an issue to later transition, but would be easier to start with such solution)

aviramha (changes requested):

mirrord-auth/src/lib.rs
DmitryDodzin: the 'username' for the keyring should be the OS username or something from cli input (with default value)?

aviramha: os username sounds good

aviramha: (in the example they use `whoami`)

aviramha (approved):

PR #309, opened 2022-08-31
go_env hook: use a different stack for the hook execution

aviramha:

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -5,6 +5,7 @@
 #![feature(naked_functions)]
 #![feature(result_flattening)]
 #![feature(io_error_uncategorized)]
+#![feature(let_chains)]
 
 use std::{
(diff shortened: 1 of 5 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![feature(naked_functions)]
 #![feature(result_flattening)]
 #![feature(io_error_uncategorized)]
+#![feature(let_chains)]
```
meowjesty, line 8: Do we need to enable this feature? I'm on nightly `rustc 1.65.0-nightly (bc4b39c27 2022-08-29)` and it doesn't warn my anywhere.

meowjesty (approved):
👍

mirrord-layer/src/lib.rs
aviramha: mm when I updated it does that.. let's leave it for now I guess.

PR #313, opened 2022-09-01
Add Resources Config file for sample app.

t4lz:
Add a config file to the sample app to facilitate https://github.com/metalbear-co/mirrord.dev/pull/26


eyalb181 (changes requested):
sample/kubernetes might be more appropriate since conf is pretty general

eyalb181 (approved):

PR #315, opened 2022-09-02
Fix race condition in connection id generation. And some tests.

meowjesty:
Fix tests (dns, outgoing).


tests/node-e2e/outgoing/test_outgoing_traffic_make_request_after_listen.mjs
```diff
@@ -51,36 +51,39 @@ const listen = () => {
...
+    socket.on("close", () => {
       console.log(">> connection from %s closed", remoteAddress);
     });
 
-    conn.on("error", (fail) => {
-      console.error(">> Connection %s error: %s", remoteAddress, err.message);
-      process.exit(-1);
+    socket.on("error", (fail) => {
+      process.stderr.write(
+        `>> failed connectio to ${remoteAddress} with ${err.message}`
```
eyalb181, line 83:
```suggestion
        `>> failed connection to ${remoteAddress} with ${err.message}`
```

tests/node-e2e/outgoing/test_outgoing_traffic_make_request_after_listen.mjs
```diff
@@ -19,22 +19,22 @@ const makeRequest = () => {
     });
 
     response.on("error", (fail) => {
-      console.error(`>> response failed with ${fail}`);
-
-      process.exit(-1);
+      process.stderr.write(`>> response failed with ${fail}`);
```
eyalb181, line 22: When do you use process.stderr.write and when do you use console.log?

meowjesty: Tbh this should all be the same (either write to stdout(err), or console). Changed to be all `console.log` and `console.error`.

eyalb181 (approved):

Issue #321, opened 2022-09-04
Remote DNS doesn't work for node.js


eyalb181:
### Bug Description

@aviramha encountered the following error when enabling remote DNS on a Node.js app:
 
`2022-09-02T16:09:31.397836Z ERROR mirrord_layer::error: Error occured in Layer >> ResponseError(RemoteIO(RemoteIOError { raw_oserror: None, kind: Other }))
Assertion failed: (!"unknown EAI* error code"), function uv__getaddrinfo_translate_error, file ../deps/uv/src/unix/getaddrinfo.c, line 90.
[1]    90277 abort      mirrord exec -s "*" -o -d -p front-end-6585d48b5c-v6q2h -n sock-shop -- node
node dns with uvloop fails`

### Steps to Reproduce

Enable remote DNS on a Node.js app (that does some DNS resolving)

### Your operating system and version

N/A

### Local process

N/A

aviramha:
I found one issue, and that's the DNS being resolved in the agent context instead of the impersonated container context.
Another theoretical issue is that the error is `IO` and not not found for some reason.. might be related to the first issue.

aviramha:
Okay, besides the first theory, the error code didn't propagate but even when it did, macOS expects different codes than Linux so it needs more translation. PR incoming soon.

PR #324, opened 2022-09-04
remote dns doesnt work for nodejs

aviramha:
Closes https://github.com/metalbear-co/mirrord/issues/321

### DIFF
--- a/mirrord-agent/src/dns.rs
+++ b/mirrord-agent/src/dns.rs
@@ -1,4 +1,27 @@
-use mirrord_protocol::GetAddrInfoRequest;
+use std::path::PathBuf;
+
+use mirrord_protocol::{
+    AddrInfoHint, AddrInfoInternal, GetAddrInfoRequest, RemoteResult, ResponseError,
+};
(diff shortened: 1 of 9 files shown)


mirrord-agent/src/dns.rs
```diff
@@ -0,0 +1,90 @@
...
+        let namespace = PathBuf::from("/proc")
+            .join(PathBuf::from(pid.to_string()))
+            .join(PathBuf::from("ns/net"));
+
+        set_namespace(namespace)?;
+    };
+
+    while let Some(DnsRequest { request, tx }) = rx.recv().await {
+        let result = get_addr_info(request);
+        if let Err(result) = tx.send(result) {
```
meowjesty, line 84: Do we want to eat the error here? In most other places when a channel fails to send we return the error, I think.

aviramha: It probably means the caller dropped, it's really an edge case but I think this is okay for now.

eyalb181 (approved):

PR #329, opened 2022-09-05
support read only file access

### LINKED ISSUE #277: Support read-only file access
We should let users configure file access to be read-only, so they don't risk corrupting the remote environment.

aviramha:
closes #277


mirrord-layer/src/lib.rs
```diff
@@ -97,7 +98,10 @@ fn init() {
         HOOK_SENDER = Some(sender);
     };
 
-    let enabled_file_ops = ENABLED_FILE_OPS.get_or_init(|| config.enabled_file_ops);
+    let enabled_file_ops =
+        ENABLED_FILE_OPS.get_or_init(|| (config.enabled_file_ops || config.enabled_file_ro_ops));
+    let _ = ENABLED_FILE_RO_OPS
```
DmitryDodzin, line 103: If you don't set enabled_file_ops to true then the hooks won't be enabled in `enable_hooks` fn

tests/src/sanity.rs
```diff
@@ -662,9 +662,34 @@ mod tests {
 
         let mut args = vec!["--enable-fs"];
 
-        if let Some(ephemeral_flag) = agent.flag() {
```
DmitryDodzin, line 665: no need for ephemeral_flag for this test anymore?

mirrord-layer/src/file/hooks.rs
```diff
@@ -41,7 +41,13 @@ unsafe fn open_logic(raw_path: *const c_char, open_flags: c_int) -> RawFd {
     if IGNORE_FILES.is_match(path.to_str().unwrap_or_default()) || !path.is_absolute() {
         FN_OPEN(raw_path, open_flags)
     } else {
-        let open_options = OpenOptionsInternalExt::from_flags(open_flags);
+        let open_options: OpenOptionsInternal = OpenOptionsInternalExt::from_flags(open_flags);
+        let read_only = ENABLED_FILE_RO_OPS
+            .get()
+            .expect("Should be set during initialization!");
+        if *read_only && !open_options.is_read_only() {
+            return FN_OPEN(raw_path, open_flags);
```
DmitryDodzin, line 49: I think we should somehow warn the user that the process writing to a file on his local machine and not the remote host

mirrord-layer/src/lib.rs
aviramha: but I added ||?

tests/src/sanity.rs
aviramha: we don't test ephemeral on macOS as it's redundant and it always tested job only so redundant

mirrord-cli/src/config.rs
```diff
@@ -39,10 +39,14 @@ pub(super) struct ExecArgs {
     #[clap(short = 'i', long, value_parser)]
     pub agent_image: Option<String>,
 
-    /// Enable file hooking
+    /// Enable file hooking (Both R/W)
```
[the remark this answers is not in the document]
aviramha: can't do it as part of a minor version.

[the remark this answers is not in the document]
aviramha: ( weshould do it as part of the overall UX change)

eyalb181 (approved):

mirrord-layer/src/lib.rs
DmitryDodzin: I mean the local variable of `enabled_file_ops` for line 106, right now its `let _ = ENABLED_FILE_RO_OPS`

aviramha: the get or init sets it

PR #342, opened 2022-09-06
Unset env vars for k8s ops

eyalb181:


mirrord-layer/src/pod_api.rs
```diff
@@ -18,6 +18,32 @@ use crate::{
     error::{LayerError, Result},
 };
 
+struct EnvVarGuard {
+    env_var: String,
+    library: String
+}
+impl EnvVarGuard {
+    fn new() -> Self {
+        let env_var = match std::env::consts::OS
```
aviramha, line 27: this can be `#[cfg]` and also contain more options (macos vs std unix where LD_PRELOAD will work for most)

aviramha (approved):

Issue #344, opened 2022-09-06
CLI crushes because of existing dylib


t4lz:
### Bug Description

The CLI tool crashes, deleting `/tmp/libmirrord_layer.dylib` solves the problem. Once `/tmp/libmirrord_layer.dylib` is valid again, deleting it between runs is not necessary.

### Steps to Reproduce

If `/tmp/libmirrord_layer.dylib` is in a bad state (not sure yet how this happens):
1. Run mirrord from the command line.

### Backtrace

```shell
Exception: EXC_BAD_ACCESS (code=50, address=0x109608000)

Process finished with exit code 9
```

### Relevant Logs

```shell
2022-09-06T12:53:07.313580Z DEBUG mirrord: Extracted library file to "/tmp/libmirrord_layer.dylib"
```

### Your operating system and version

macOS Monterey 12.5.1

### Local process

sample JS app.

### Additional Info

Also reported to happen on Linux.

PR #351, opened 2022-09-06
Add random string prefix to layer file

infiniteregrets:


mirrord-cli/src/main.rs
```diff
@@ -39,11 +40,23 @@ fn extract_library(dest_dir: Option<String>) -> Result<PathBuf> {
     let library_file = env!("MIRRORD_LAYER_FILE");
     let library_path = std::path::Path::new(library_file);
 
-    let file_name = library_path.components().last().unwrap();
+    let extension = library_path
```
eyalb181, line 43: Does this happen every time the CLI runs? Is it deleted at some point or do we just keep creating dylibs?

infiniteregrets: > Does this happen every time the CLI runs?
Not every time, but randomly. 

>  Is it deleted at some point or do we just keep creating dylibs?

Previously, we were overwriting the dylib/so created in tmp, since we added a random string to it now, we are not doing so anymore. not sure what would be the right way to delete the old shared lib

eyalb181: Don't worry about it for now since it's in /tmp anyway

eyalb181 (approved):

infiniteregrets:
@eyalb181 note, i am unable to remove trailing slash from the generated name for the lib

infiniteregrets:
ps - the filename looks like this `gw3tqxzwil-libmirrord_layer.so` after the change

infiniteregrets:
closes #344

PR #357, opened 2022-09-07
Use selected namespace in IntelliJ.

t4lz:
Currently the IntelliJ Plugin does not supply the selected namespace to the layer, but just uses it to fetch pods.
This means that if the user selected a non-default namespace, the mirrord layer will not find the given pod in the default namespace and will panic.


aviramha (approved):
Good job!

PR #364, opened 2022-09-07
E2E: Python fastapi

DmitryDodzin:
Add e2e tests for fastapi python framework with uvicorn as http server


aviramha (approved):
Looks good, just need to fix tests.

PR #365, opened 2022-09-07
add go1.19 support

aviramha:


mirrord-layer/src/go_hooks.rs
```diff
@@ -1,468 +1,659 @@
...
+        frida_gum::Module::find_symbol_by_name(binary, "runtime.buildVersion.str")
+    {
+        // Version str is `go1.xx` - take only last 4 characters.
+        let version = unsafe {
+            std::str::from_utf8_unchecked(std::slice::from_raw_parts(
+                version_symbol.0.add(2) as *const u8,
+                4,
+            ))
+        };
+        let version_parsed: f32 = version.parse().unwrap();
```
DmitryDodzin, line 650: this might crash on non numbered versions of go, example is with Experiment flags enabled

aviramha: mm okay sounds like an edge-enough case to not support atm

DmitryDodzin (approved):

PR #367, opened 2022-09-07
Go out

meowjesty:
Support outgoing traffic for golang.


tests/src/sanity.rs
```diff
@@ -965,4 +965,16 @@ mod tests {
         assert!(res.success());
         process.assert_stderr();
     }
+
+    #[rstest]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    pub async fn test_go_outgoing_traffic_single_request_enabled(#[future] service: EchoService) {
+        let service = service.await;
+        let command = vec!["go-e2e-outgoing/go-e2e-outgoing"];
```
aviramha, line 973: add 18/19 versions (see PR that's about to be merged #365

mirrord-layer/src/socket/ops.rs
```diff
@@ -216,6 +218,59 @@ pub(super) fn connect(sockfd: RawFd, remote_address: SocketAddr) -> Result<()> {
         .get()
         .expect("Should be set during initialization!");
 
+    if let SocketKind::Udp(kind) = user_socket_info.kind {
+        // Prepare this socket to be intercepted.
```
aviramha, line 222: bypass if not port `53`

aviramha (commented):
Looking good! Just 2 comments. see inline.

tests/src/sanity.rs
meowjesty: Added.

mirrord-agent/src/outgoing/udp.rs
```diff
@@ -0,0 +1,276 @@
...
+        .parse()
+        .map_err(RemoteError::from)?;
+
+    Ok(dns_address)
+}
+
+async fn connect(remote_address: SocketAddr) -> Result<UdpSocket, ResponseError> {
+    trace!("connect -> remote_address {:#?}", remote_address);
+
+    let remote_address = if remote_address.port() == 53 {
```
aviramha, line 83: use const you added forthat?

aviramha (changes requested):
nitpick + UDP is always enabled currently - add a flag to enable/disable it (we'll have it enabled by default..)

aviramha (approved):
LGTM! Let's get this merged 🥳

PR #385, opened 2022-09-09
IntelliJ Ext: Add Exception if shared lib is not found

### LINKED ISSUE #345: IntelliJ-ext: prepareSandbox task does not throw an exception if the source files don't exist
We should check if the dylib and so files exist in the working directory or not for this task as we ended up bundling only the .so files in the previous releases and saw no errors

infiniteregrets:
closes #345

### DIFF
--- a/intellij-ext/build.gradle.kts
+++ b/intellij-ext/build.gradle.kts
@@ -89,7 +89,7 @@ tasks {
                 into(pluginName.get())
             }
             // NOTE: comment this line when developing locally without either of shared libs
-            if (System.getenv("CI_BUILD_PLUGIN").toBoolean()) {
+            if (!System.getenv("CI_BUILD_PLUGIN").toBoolean()) {
                 if (!inputs.sourceFiles.files.contains(File(lib))) throw StopExecutionException("Expected library: $lib >> Not Found")
             }
         }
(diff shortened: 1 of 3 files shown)


intellij-ext/build.gradle.kts
```diff
@@ -69,24 +69,29 @@ tasks {
...
-        from(file("$projectDir/libmirrord_layer.so")) {
-            into(pluginName.get())
+        val sharedLibs =
+            mapOf("macos" to "$projectDir/libmirrord_layer.dylib", "linux" to "$projectDir/libmirrord_layer.so")
+        sharedLibs.forEach { (_, lib) ->
+            from(file(lib)) {
+                into(pluginName.get())
+            }
+            // NOTE: comment this line when developing locally without either of shared libs
+            if (!System.getenv("CI_BUILD_PLUGIN").toBoolean()) {
```
eyalb181, line 92: The VS Code extension has the "vscode.ExtensionMode.Development" flag, which then lets us grab the lib from target/debug instead of the cwd. Anything similar in IntelliJ?

infiniteregrets: hmm not sure, the main reason for adding thsi exception was that we shipped a version of mirrord without even loading the dylib, so if in the CI it fails, we know its an issue with dylibs not being put in the plugin dir

eyalb181 (approved):

PR #386, opened 2022-09-11
In go-e2e, exit instead of sending SIGINT.

t4lz:
No CHANGELOG entry, since this is a minor and internal change to the tests, right?


aviramha (approved):

aviramha:
Regarding change log - we try for every change but not critical.

t4lz:
> Regarding change log - we try for every change but not critical.

Got it. Added CHANGELOG entry.

PR #387, opened 2022-09-11
Fix Environment parsing when value contained '='

DmitryDodzin:
If env variable contains '=' in it's value then the GetEnvVarsResponse is completely messed up with the key being the last env variable

Example:
```
PG_CONNECTION=postgres://localhost/?ssl_mode=disable
```
generates response 

```
DaemonMessage::GetEnvVarsResponse {
    ...
    "postgres://localhost/?ssl_mode": "disable",
    ...
}!
```


aviramha:
Looks good! Please add this env to the e2e as well.

aviramha (approved):

Issue #331, opened 2022-09-05
Tcp outgoing fails with "rapid fire" multiple requests (12+).


meowjesty:
### Bug Description

With the `MIRRORD_TCP_OUTGOING` feature enabled, if you try to make multiple requests in quick succession (tested on node), one of the channels appear to close before some request is completed, breaking the whole thing.

### Steps to Reproduce

1. Enable `MIRRORD_TCP_OUTGOING`;
2. Make multiple requests (reproduced with around 12 or more);
3. ???
4. `ECONNRESET` error.

### Backtrace

```shell
>> www.google.com request failed with Error: socket hang up
file:///home/lanedraex/dev/metalbear/mirrord/tests/node-e2e/outgoing/test_outgoing_traffic_multiple_requests.mjs:33
    throw fail;
    ^

Error: socket hang up
    at connResetException (node:internal/errors:692:14)
    at TLSSocket.socketOnEnd (node:_http_client:478:23)
    at TLSSocket.emit (node:events:539:35)
    at endReadableNT (node:internal/streams/readable:1344:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  code: 'ECONNRESET'
}

Node.js v18.0.0
```

### Relevant Logs

```shell
2022-09-05T14:30:53.300557Z ERROR mirrord_layer: Error handling daemon message: SendErrorConnection(SendError([23, 3, 3, 0, 19, 127, 101, 127, 111, 194, 221, 230, 193, 252, 77, 138, 200, 99, 240, 32, 99, 33, 123, 17]))
2022-09-05T14:30:53.300794Z  WARN mirrord_layer::tcp::outgoing: interceptor_task -> exiting due to remote stream closed!
2022-09-05T14:30:53.300808Z  WARN mirrord_layer::tcp::outgoing: interceptor_task -> exiting due to remote stream closed!
2022-09-05T14:30:53.300859Z  WARN mirrord_layer::tcp::outgoing: interceptor_task -> exiting due to remote stream closed!
2022-09-05T14:30:53.300881Z TRACE mirrord_layer::tcp::outgoing: interceptor_task done -> connection_id 11
2022-09-05T14:30:53.300975Z  WARN mirrord_layer::tcp::outgoing: interceptor_task -> exiting due to remote stream closed!
2022-09-05T14:30:53.301020Z TRACE mirrord_layer::tcp::outgoing: interceptor_task done -> connection_id 2
[...]

PR #393, opened 2022-09-11
Bugfix: outgoing traffic with many requests.

t4lz:
- Keep reading open streams until BOTH are done.
- Add meowjesty's many-requests test.
- Fix single-request, outgoing-disabled node test.

https://github.com/metalbear-co/mirrord/issues/331

https://github.com/metalbear-co/mirrord/pull/394 solves this bug by sleeping, as implemented in tcp_mirror.rs already. This PR suggests a different approach with less uncertainty.


aviramha (approved):
Great job!

PR #394, opened 2022-09-11
Bugfix: outgoing traffic with many requests. (like in tcp_mirror.rs).

t4lz:
- Add sleep before exiting after remote_steam closes.
- Add meowjesty's many-requests test.
- Fix single-request, outgoing-disabled node test.

https://github.com/metalbear-co/mirrord/issues/331

This PR solves the bug by the same approach used in [tcp_mirror.rs](https://github.com/metalbear-co/mirrord/blob/main/mirrord-layer/src/tcp_mirror.rs#L71). 

https://github.com/metalbear-co/mirrord/pull/393 uses a different approach without sleeping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ### Fixed
 - Fix Environment parsing error when value contained '='
+- Fix bug in outgoing traffic with multiple requests in quick succession.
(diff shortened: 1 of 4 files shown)


t4lz:
Failed on the new test (succeeded locally). This approach still has uncertainty in it and the bug is still possible if a lot of data is sent or if it's sent slowly. I think the approach https://github.com/metalbear-co/mirrord/pull/393 is preferable if it's correct.

PR #409, opened 2022-09-13
mirrord Config File

### LINKED ISSUE #395: Config file
The CLI has a lot of options, especially if you include env var configuration, a lot of it could be moved to something like a `mirrord.toml` along with a `--config` cli flag for it.

DmitryDodzin:
closes #395


mirrord-cli/src/config/file.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    #[test]
+    fn json() {
+        let input = r#"
+          {
+            "accept_invalid_certificates": false,
+            "agent": {
+              "log_level": "info",
+              "namespace": "default",
+              "image": "",
```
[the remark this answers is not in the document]
DmitryDodzin: yes

mirrord-cli/src/config/file.rs
```diff
@@ -0,0 +1,197 @@
...
+              "image_pull_policy": "",
+              "ttl": 60,
+              "ephemeral": false
+            },
+            "env": true,
+            "fs": "write",
```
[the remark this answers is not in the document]
DmitryDodzin: It's a transparent Enum variant so it's possible to add advanced configuration as another Enum variant with allow/deny lists once they are supported

mirrord-cli/src/config/file.rs
```diff
@@ -0,0 +1,197 @@
...
+            "mode": "mirror",
+            "network": {
```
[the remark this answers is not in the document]
DmitryDodzin: I can wrap it in FlagField so instead of passing udp tcp and dns you can just pass "network": true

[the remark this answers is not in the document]
eyalb181: We'd probably want to be able to configure incoming (mirror/steal) + outgoing, just incoming (mirror/steal) or just outgoing for all protocols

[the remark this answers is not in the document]
tamasfe: I would move `mode` under `network`, as it seems to be network-related to me.

mirrord-layer/src/config.rs
```diff
@@ -1,66 +1,26 @@
-use envconfig::Envconfig;
+pub mod env;
+pub mod file;
+pub mod util;
 
-#[derive(Envconfig, Debug, Clone)]
+#[derive(Debug, Clone)]
```
eyalb181, line 5: All these config structs could be dynamically generated with like a third of the code if Rust let you do that sort of thing, but since it doesn't, what if they were HashMaps themselves? So you would define the schema in a HashMap, then generate HashMaps for instances of the configuration at runtime? Or something similar but more Rusty, if it exists.

mirrord-macro/src/lib.rs
```diff
@@ -183,3 +183,198 @@ pub fn hook_guard_fn(
...
+enum FieldDefinitionMode {
+    Option,
+    UnwrapOption,
+    Nested,
+}
+
+struct FieldDefinition {
+    source_ident: Ident,
+    target_ident: Ident,
+    ty: syn::Type,
```
eyalb181, line 196: Rename to field_type

mirrord-macro/src/lib.rs
```diff
@@ -183,3 +183,198 @@ pub fn hook_guard_fn(
     // function we received as `input`.
     proc_macro::TokenStream::from(output)
 }
+
```
eyalb181, line 186: Could do with some documentation re how to use this

mirrord-macro/src/lib.rs
```diff
@@ -183,3 +183,198 @@ pub fn hook_guard_fn(
...
+    let mapped_fields = match data {
+        syn::Data::Struct(data) => match data.fields {
+            syn::Fields::Named(data) => data
+                .named
+                .into_iter()
+                .filter(|field| {
+                    !field
+                        .attrs
+                        .iter()
+                        .any(|attr| attr.path.is_ident("skip_config"))
```
eyalb181, line 230: Is skip_config used anywhere?

mirrord-config/src/util.rs
```diff
@@ -0,0 +1,99 @@
...
+    T: MirrordConfig + Default,
+{
+    type Generated = T::Generated;
+
+    fn generate_config(self) -> Result<Self::Generated, ConfigError> {
+        self.unwrap_or_default().generate_config()
+    }
+}
+
+pub trait MirrordFlaggedConfig: MirrordConfig + Default {
```
eyalb181, line 23: Does "Flagged" in this context mean "toggleable" or "optional"?

DmitryDodzin: The togglable one, referring to a field wrapped in `FlagField` meaning allow to set as false and not specifying each field as disabled in config file

mirrord-layer/src/lib.rs
```diff
@@ -126,14 +138,13 @@ fn init(config: LayerConfig) {
...
 
-    let enabled_file_ops =
-        ENABLED_FILE_OPS.get_or_init(|| (config.enabled_file_ops || config.enabled_file_ro_ops));
-    let _ = ENABLED_FILE_RO_OPS
-        .get_or_init(|| (config.enabled_file_ro_ops && !config.enabled_file_ops));
-    let _ = ENABLED_TCP_OUTGOING.get_or_init(|| config.enabled_tcp_outgoing);
-    let _ = ENABLED_UDP_OUTGOING.get_or_init(|| config.enabled_udp_outgoing);
+    let enabled_file_ops = ENABLED_FILE_OPS
+        .get_or_init(|| config.feature.fs.is_read() || config.feature.fs.is_write());
+    let _ = ENABLED_FILE_RO_OPS.get_or_init(|| config.feature.fs.is_read());
```
eyalb181, line 143: These all have default values, right? Do you need get_or_init?

eyalb181 (commented):
Made a few comments, will probably have another look tomorrow. I couldn't find the CLI functionality, were you able to get it to work?

mirrord-config/src/config/source.rs
```diff
@@ -0,0 +1,115 @@
+pub trait MirrordConfigSource: Sized {
+    type Result;
```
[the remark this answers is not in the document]
DmitryDodzin: By FromEnv, DefaultValue, Option and blanket tuple implementation, this gives a consistent API and is used mostly by the macro

mirrord-config/src/fs.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl FsField {
+    fn from_env_logic(fs: Option<bool>, ro_fs: Option<bool>) -> Option<Self> {
+        match (fs, ro_fs) {
+            (Some(false), Some(true)) | (None, Some(true)) => Some(FsField::Read),
+            (Some(true), _) => Some(FsField::Write),
+            (Some(false), Some(false)) | (None, Some(false)) | (Some(false), None) => {
+                Some(FsField::Disabled)
+            }
+            (None, None) => None,
```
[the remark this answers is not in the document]
DmitryDodzin: I want the caller of the function to decide what is the value if not provided from the environment, right now can be either the value in the config file or if the entire section marked as `Enabled(false)` then it'll be Disabled

mirrord-layer/src/lib.rs
DmitryDodzin: None of these have default values in mirrord-layer code, this is the first place we initialize them from values in the config

mirrord-config/derive/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+This crate implements derive macro `MirrordConfig`
+
+it addes `config` attribute
+
+for struct:
+- `map_to = Ident` set the name of the struct generated (defaults to `format!("Mapped{}", <struct name>)`)
+
+for fields:
+- `unwrap = true` for generated struct for the field remove the Option and set it to inner type
```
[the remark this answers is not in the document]
DmitryDodzin: Currently it's used only for values that must be provided via file or env, like what previously was `impersonated_pod_name`

mirrord-config/derive/README.md
```diff
@@ -0,0 +1,56 @@
...
+    pub value_3: String,
+    pub value_4: <OtherConfig as MirrordConfig>::Generated,
+}
+
+```
+
+
+### Limitations:
+* values must be `Option<impl FromStr>` or a nested config (value implementing MirrordConfig)
+* `unwrap` and `nested` properties are ignorant of the value theire existance only is required but due to current implementation value must be filled with something so recommned using `true`
```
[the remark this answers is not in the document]
DmitryDodzin: Better solution would be to remove the requirement to add the = in the first place, best would be `config(unwrap, env = "ENV_VARIABLE")` but there is some complication with using the `syn::Meta` as a parsing target for the statements

meowjesty (changes requested):
Minor review (will do a proper one after lunch).

mirrord-config/derive/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+        let named = fields
+            .into_iter()
+            .map(|field| map_field_name_impl(&ident, field))
+            .collect::<Vec<_>>();
+
+        quote! { { #(#named),* } }
+    };
+
+    let output = quote! {
+        #[derive(Clone, Debug)]
```
meowjesty, line 226: Shouldn't we use the original derives? Like if the user had:

```rust
#[derive(Clone, PartialEq, Foo, MirrordConfig)]
struct Foo ...
```

Doesn't this macro kinda eat up the other derives?

meowjesty: Same applies for the doc comments added to the original struct, aren't they lost here?

meowjesty (changes requested):
Alright, another small round of comments.

My main ask would be to add docs, as proc_macro code is kinda hard to parse (ha!), also to the other functions, traits, etc. Docs everywhere.

mirrord-config/derive/src/lib.rs
DmitryDodzin: About the macros not sure how to get the other macros, I printed the macro input and didn't get other derives, also as far as I see derive macros don't touch the original struct so no other derives are overwritten just don't get copied over to Mapped struct

mirrord-config/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+use crate::{
+    agent::AgentField, config::source::MirrordConfigSource, feature::FeatureField, pod::PodField,
+    util::VecOrSingle,
+};
+
+#[derive(MirrordConfig, Deserialize, Default, PartialEq, Eq, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+#[config(map_to = LayerConfig)]
+pub struct LayerFileConfig {
+    #[config(env = "MIRRORD_ACCEPT_INVALID_CERTIFICATES", default = "false")]
```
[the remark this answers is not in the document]
DmitryDodzin: The struct is used with serde to parse the file, if it's not an option you must specify the value and not use the default

mirrord-config/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+
+use crate::{
+    agent::AgentField, config::source::MirrordConfigSource, feature::FeatureField, pod::PodField,
+    util::VecOrSingle,
+};
+
+#[derive(MirrordConfig, Deserialize, Default, PartialEq, Eq, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+#[config(map_to = LayerConfig)]
+pub struct LayerFileConfig {
```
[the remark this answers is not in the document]
DmitryDodzin: This makes me think a new approach is probably needed, instead of making the result struct from the file config one to generate the file config struct from the result struct

mirrord-config/derive/README.md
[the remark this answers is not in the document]
eyalb181: Looks like you can specify a default value in the `serde` attribute, can't the `config` attribute do that when it generates the code?

mirrord-config/derive/src/lib.rs
meowjesty: Oh, thought that `derive` behaved like the other macros, mb.

mirrord-config/derive/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+        Meta::List(list) => {
+            let mut result = Vec::new();
+            for nested in list.nested.iter() {
+                result.push(get_config_flag(nested)?);
+            }
+
+            result
+        }
+        _ => vec![],
+    };
```
meowjesty, line 63:
```suggestion
Meta::List(list) => list.nested.into_iter().flat_map(get_config_flag).collect(),
_ => vec![],
```

ps: requires `get_config_flag` to change as well `(meta: NestedMeta)`.

DmitryDodzin: The problem I have with making it a one liner is how to propagate the Err from `get_config_flag` Result

DmitryDodzin: Huh, apparently collect has an [implementation](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect)

mirrord-config/derive/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    Default(Lit),
+}
+
+fn map_to_ident(source: &Ident, expr: Option<Expr>) -> Ident {
+    let fallback = Ident::new(&format!("Mapped{}", source), Span::call_site());
+
+    if let Some(Expr::Assign(expr)) = expr {
+        if let (Expr::Path(left_path), Expr::Path(right_path)) = (*expr.left, *expr.right) {
+            if left_path.path.is_ident("map_to") {
+                return right_path.path.get_ident().cloned().unwrap_or(fallback);
```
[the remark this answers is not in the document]
DmitryDodzin: Added Diagnostic Errors

mirrord-layer/src/lib.rs
DmitryDodzin: Changed it to `set` function

mirrord-layer/src/lib.rs
```diff
@@ -9,18 +9,19 @@
...
 use common::{GetAddrInfoHook, ResponseChannel};
 use ctor::ctor;
-use envconfig::Envconfig;
 use error::{LayerError, Result};
 use file::OPEN_FILES;
 use frida_gum::{interceptor::Interceptor, Gum};
 use futures::{SinkExt, StreamExt};
 use kube::api::Portforwarder;
 use libc::c_int;
+use mirrord_config::{config::MirrordConfig, util::VecOrSingle, LayerConfig, LayerFileConfig};
```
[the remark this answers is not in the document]
DmitryDodzin: Trait import required to use `generate_config`

mirrord-layer/src/lib.rs
```diff
@@ -84,9 +84,29 @@ fn before_init() {
     if !cfg!(test) {
         let args = std::env::args().collect::<Vec<_>>();
         let given_process = args.first().unwrap().split('/').last().unwrap();
-        let config = LayerConfig::init_from_env().unwrap();
-        if should_load(given_process, &config.skip_processes) {
-            init(config);
+
+        let config = std::env::var("MIRRORD_CONFIG_FILE")
```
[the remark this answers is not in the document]
DmitryDodzin: `unwrap_or_default` creates an empty LayerFileConfig so only env and default values affect the resulting value

mirrord-config/src/network.rs
```diff
@@ -0,0 +1,80 @@
...
+        default_value::DefaultValue, from_env::FromEnv, source::MirrordConfigSource, ConfigError,
+    },
+    incoming::IncomingField,
+    outgoing::OutgoingField,
+    util::{FlagField, MirrordFlaggedConfig},
+};
+
+#[derive(MirrordConfig, Deserialize, Default, PartialEq, Eq, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct NetworkField {
```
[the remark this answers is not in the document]
DmitryDodzin: Changed

eyalb181 (changes requested):

mirrord-config/derive/README.md
```diff
@@ -0,0 +1,54 @@
+# Mirrord Config Derive
+
+This crate implements the derive macro `MirrordConfig`.
+
+`config` attribute:
+
+- container
```
eyalb181, line 7:
```suggestion
- On a struct:
```

mirrord-config/derive/README.md
```diff
@@ -0,0 +1,54 @@
...
+    - `map_to = <Some Ident>` set the name of the new generated struct (defaults to `format!("Mapped{}", <struct name>)`)
+
+- fields
```
eyalb181, line 10:
```suggestion
- On struct fields:
```

mirrord-config/derive/README.md
```diff
@@ -0,0 +1,54 @@
+# Mirrord Config Derive
+
+This crate implements the derive macro `MirrordConfig`.
```
eyalb181, line 3:
```suggestion
This crate implements the derive macro `MirrordConfig`, which introduces the `config` attribute:
```

mirrord-config/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    #[config(env = "MIRRORD_ACCEPT_INVALID_CERTIFICATES", default = "false")]
+    pub accept_invalid_certificates: Option<bool>,
+
+    #[config(env = "MIRRORD_SKIP_PROCESSES")]
+    pub skip_processes: Option<VecOrSingle<String>>,
+
+    #[serde(default)]
```
eyalb181, line 35: Why do you need serde(default) here?

mirrord-config/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    agent::AgentFileConfig, config::source::MirrordConfigSource, feature::FeatureFileConfig,
+    pod::PodFileConfig, util::VecOrSingle,
+};
+
+#[derive(MirrordConfig, Deserialize, Default, PartialEq, Eq, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+#[config(map_to = LayerConfig)]
+pub struct LayerFileConfig {
+    #[config(env = "MIRRORD_ACCEPT_INVALID_CERTIFICATES", default = "false")]
+    pub accept_invalid_certificates: Option<bool>,
```
eyalb181, line 30: Document: // This is the root struct for mirrord-layer's configuration

mirrord-config/derive/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+            }
+
+            Err(expr_span.error("invalid value for config attribute"))
+        }
+        Some(other) => Err(other.span().error("invalid value for config attribute")),
+        None => Ok(fallback),
+    }
+}
+
+/// Parse field attribte to FieldAttr
```
eyalb181, line 42:
```suggestion
/// Parse a flag in the `config` attribute on a struct field to FieldAttr
```

mirrord-config/derive/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+            Ok(FieldAttr::Env(meta.lit))
+        }
+        NestedMeta::Meta(Meta::NameValue(meta)) if meta.path.is_ident("default") => {
+            Ok(FieldAttr::Default(meta.lit))
+        }
+        _ => Err(meta.span().error("unsupported config attribute flag")),
+    }
+}
+
+/// Parse field attribtes to FieldAttr
```
eyalb181, line 57:
```suggestion
/// Parse the `config` attribute on a struct field
```

mirrord-config/src/lib.rs
DmitryDodzin: To not use Option for these fields but to use the Default::default from the underlying struct because it implements `MirrordConfig`

eyalb181 (changes requested):

mirrord-config/derive/src/lib.rs
eyalb181: @DmitryDodzin Is this resolved?

mirrord-config/src/lib.rs
eyalb181: So this is going to be mandatory for any nested field which isn't Toggleable, right? But there's no way to add it as part of the config macro?

DmitryDodzin: For fields that implement `MirrordConfig` and `Default` no need to make them optional, I removed the `Option` and added the `#[serde(default)]` for all nested fields no only the ones without Toggleable

eyalb181 (approved):

PR #370, opened 2022-09-07
Fix python issue - vscode

infiniteregrets:
closes #350

PR #420, opened 2022-09-14
Fix connect

meowjesty:
- Adds a `tokio::time::timeout` to mirrord-agent outgoing `TcpStream::connect`, this fixes go request issue;


mirrord-layer/src/socket/ops.rs
```diff
@@ -210,167 +275,36 @@ pub(super) fn connect(sockfd: RawFd, remote_address: SocketAddr) -> HookResult<(
         .then_some(())
         .ok_or_else(|| HookError::BypassedPort(remote_address.port()))?;
```
eyalb181, line 276: Is this the same as writing`if is_ignored_port(remote_address.port()) {return BypassedPort...}`?

mirrord-layer/src/socket/ops.rs
```diff
@@ -210,167 +275,36 @@ pub(super) fn connect(sockfd: RawFd, remote_address: SocketAddr) -> HookResult<(
...
-                        rawish_remote_address.as_ptr(),
-                        rawish_remote_address.len(),
-                    )
-                };
-
-                if result != 0 {
                     Err(io::Error::last_os_error())?
                 } else {
-                    Ok::<_, HookError>(())
+                    Err(HookError::Bypass)
```
eyalb181, line 301: What does this error mean?

mirrord-layer/src/socket/ops.rs
```diff
@@ -210,167 +275,36 @@ pub(super) fn connect(sockfd: RawFd, remote_address: SocketAddr) -> HookResult<(
...
-            let err_code = errno::errno().0;
-            if connect_result == -1 && err_code != libc::EINPROGRESS && err_code != libc::EINTR {
-                error!(
-                    "connect -> Failed call to libc::connect with {:#?} errno is {:#?}",
-                    connect_result,
-                    errno::errno()
-                );
-                return Err(io::Error::last_os_error())?;
+        SocketKind::Tcp(_) => match user_socket_info.state {
+            SocketState::Initialized if !((is_ignored_port(port)) && (is_ignored_ip(ip))) => {
```
eyalb181, line 283: So we use `is_ignored_port` together with `is_ignored_ip` here, but on its own above, which seems weird? Can you explain it?

.dockerignore
```diff
@@ -1,5 +1,5 @@
-*/target*
-*/.vscode*
-*/.git*
-*/.github*
-*/.devcontainer*
\ No newline at end of file
+**/target*
```
eyalb181, line 1: Why do we need this file in both root and mirrord-agent/?

mirrord-agent/src/outgoing.rs
```diff
@@ -42,6 +43,84 @@ pub(crate) struct TcpOutgoingApi {
...
+    writers: &mut HashMap<ConnectionId, OwnedWriteHalf>,
+    readers: &mut StreamMap<ConnectionId, ReaderStream<OwnedReadHalf>>,
+    daemon_tx: Sender<DaemonTcpOutgoing>,
+) -> Result<(), AgentError> {
+    match layer_message {
+        // [user] -> [layer] -> [agent] -> [layer]
+        // `user` is asking us to connect to some remote host.
+        LayerTcpOutgoing::Connect(LayerConnect { remote_address }) => {
+            let daemon_connect = timeout(
+                Duration::from_millis(1500),
```
eyalb181, line 59: How'd you come up with 1500? Seems like it might be too short?

mirrord-layer/src/socket/ops.rs
meowjesty: This was a close relative of Windows' "Error(Success)".

mirrord-agent/src/outgoing.rs
meowjesty: Increased to 3 seconds. Should I create a more robust mechanism for increasing this timeout?

.dockerignore
meowjesty: Apparently we don't.

mirrord-layer/src/socket/ops.rs
meowjesty: This came in #370. I'll remove the `is_ignored_port` duplicate.

mirrord-agent/src/outgoing.rs
eyalb181: I'm not sure - what are the repercussions here? For example, if the user's application intentionally makes an outgoing request that is long-running, are we going to fail it?

mirrord-layer/src/socket/ops.rs
eyalb181: Would the code I suggested work though? Or do we prefer chaining methods, stylewise?

eyalb181 (approved):

mirrord-agent/src/outgoing.rs
meowjesty: Yeah, it'll fail.

So the question is, should I address this now, or wait until someone opens up an issue?

mirrord-layer/src/socket/ops.rs
meowjesty: Your suggestion works. I'm not a fan of early returns like that, and prefer chaining things.

Should I open a discussion to gather feedback from our pals?

eyalb181: Yeah lets do it

mirrord-agent/src/outgoing.rs
eyalb181: I'm leaning towards option B here, but do you have a solution in mind?

meowjesty: Hmm I could investigate more deeply what's blocking the `connect` call for this specific case in go (while in all other cases it behaves as expected).

mirrord-layer/src/socket/ops.rs
meowjesty: Poll is open, I'll mark this as resolved and wait for feedback (then we can just refactor it later).

mirrord-agent/src/outgoing.rs
eyalb181: I wouldn't worry about it for now, but would add some documentation + a TODO

meowjesty: TODO added.

PR #421, opened 2022-09-15
feat: initial progress reporting

### LINKED ISSUE #389: Provide feedback while the agent is initializing
The CLI currently silently waits for the agent to be up and running, since it can take a while it would be nice to show some minimal progress.

tamasfe:
Initial version of progress reporting via tracing spans, closes #389.

Reporting is done in 4 flavors:

- standard mode: fancy reporting with spinners
- simple mode: simple messages printed
- json: not implmented
- off: the layer is completely disabled

The tasks start when a span is first entered, and end when the span is closed (dropped), the reason for this is async code where a span is repeatedly entered whenever the futures are polled.

In standard mode the same spinner is reused for all nested spans, and the ongoing task is only done when the outermost span is closed.

<details>
<summary>standard mode example</summary>

https://user-images.githubusercontent.com/25967296/190286758-cbd82564-7f60-4fe5-8bc9-4b0d8b9cabd7.mp4

</details>

The agent setup ended up being too quick when I actually got to try it out so I had to add artificial network delays for the vid above...

I also introduced a `mirrord-common` library for stuff like this that are used from both the cli and the layer.

### Todo

- I cannot seem to pass env vars to the layer from the cli, so by default the layer has standard reporting enabled as well.
- We only check if a span is closed, if it's closed, success is printed, even if the task at hand failed. I'm not sure how to handle this, I'd say it's probably not a huge issue as the error is immediately printed, but a false success message is still ugly or probably even confusing.
- Each time I write a tracing layer I need to rediscover how tracing works, so the code is a bit messy and requires a cleanup.


mirrord-common/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+atty = "0.2.14"
+colored = "2.0.0"
+spinoff = "0.5.4"
+tracing = "0.1.36"
```
DmitryDodzin, line 13:
```suggestion
tracing.workspace = true
```

mirrord-common/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+tracing-subscriber = { version = "0.3.15", features = ["env-filter", "registry"] }
```
DmitryDodzin, line 14:
```suggestion
tracing-subscriber = { workspace = true, features = ["env-filter", "registry"] }
```

mirrord-common/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "mirrord-common"
+version = "0.1.0"
+edition = "2021"
+publish = false
```
DmitryDodzin, line 5:
```suggestion
version.workspace = true
authors.workspace = true
description.workspace = true
documentation.workspace = true
readme.workspace = true
homepage.workspace = true
repository.workspace = true
license.workspace = true
keywords.workspace = true
categories.workspace = true
publish.workspace = true
edition.workspace = true
```

mirrord-cli/src/main.rs
```diff
@@ -184,6 +189,8 @@ fn exec(args: &ExecArgs) -> Result<()> {
         std::env::set_var("MIRRORD_UDP_OUTGOING", "false");
     }
 
+    std::env::set_var(MIRRORD_PROGRESS_ENV, "standard");
```
DmitryDodzin, line 192: I would set the `MIRRORD_PROGRESS_ENV` env variable only if not set already

mirrord-common/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "mirrord-common"
```
DmitryDodzin, line 2: I think its better that the package will be only for the progress so changing it to mirrord-progress would be best

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "mirrord-layer",
     "mirrord-cli",
     "mirrord-macro",
+    "mirrord-common",
```
DmitryDodzin, line 10: Because we compile with workspaces need to add the directory to `./mirrord-agent/Dockerfile`

DmitryDodzin (commented):
Looks great overall, I left several comments mostly about consistency with the rest of the repo.

eyalb181:
@tamasfe Hey, wanted to check if you were planning to get back to this one anytime soon? Feel free if you want us to pick it up where you left off.

eyalb181:
We're in no rush to get this merged, so if you're planning to get back to it sometimes in the near future we'll leave it to you. I think we do like the current approach.

tamasfe:
Closing in favor of #537.

Issue #397, opened 2022-09-11
Support mirroring in service mesh proxies


tamasfe:
Currently if a pod is meshed with a [`linkerd-proxy`](https://github.com/linkerd/linkerd2) the traffic will not reach the agent, and the `linkerd-proxy` container cannot be impersonated #396.

tamasfe:
#396 has been fixed, but no traffic is captured still, or at least the traffic I expect. Linkerd uses its own TLS that could interfere, or some other magic. Without knowing mirrord internals, I can't investigate this further for now.

eyalb181:
Update: service mesh proxies should be mostly supported now, with the exception of traffic mirroring (stealing works fine). The reason for this is that we only starts mirroring when a connection is closed and a new connection is opened, and this can take a while to happen on linkerd (keep-alive is 10 seconds, so we'd need it not to receive any requests for that amount of time).

meowjesty:
Should've been fixed by #1887 , see caveats there.

aviramha:
@meowjesty Please create a follow up ticket (mid-packet HTTP detection) and then close this one
Thanks!

meowjesty:
Tracked in #1942

PR #431, opened 2022-09-18
TCP stealing with proxy

DmitryDodzin:
relevant #397


mirrord-agent/src/steal.rs
```diff
@@ -106,6 +105,47 @@ impl Drop for SafeIpTables {
...
+
+    fn redirect_rule(&self, redirected_port: Port, target_port: Port) -> String {
+        let redirect_rule = format!(
+            "-m tcp -p tcp --dport {} -j REDIRECT --to-ports {}",
+            redirected_port, target_port
+        );
+
+        match self {
+            IPTableFormatter::Normal => redirect_rule,
+            IPTableFormatter::Linkerd => format!("-o lo {}", redirect_rule),
```
aviramha, line 144: I think somewhat better implementation would be a trait then func that returns the appropriate struct (by using `dyn Impl`?) but it's a bit a nitpick ofc.
also some UT here would be nice

aviramha (approved):

mirrord-agent/src/steal.rs
aviramha: you can make a trait and accept stuff that `impl` it. maybe @meowjesty have better idea

mirrord-agent/Cargo.toml
```diff
@@ -42,6 +42,7 @@ faccess = "0.2"
 bytes = "1.2"
 regex = "1"
 socket2 = "0.4"
+mockall = "0.11"
```
aviramha, line 45: add this as a dev dependency

aviramha (approved):
LGTM! Nitpick on the cargo.toml.

mirrord-agent/Cargo.toml
DmitryDodzin: I can't because the `mockall::automock` macro exists on the IPTables struct

DmitryDodzin: NVM found how to do it

Issue #368, opened 2022-09-07
VS Code Node debugger crashes


eyalb181:
### Bug Description

When debugging a Node process with mirrord enabled (no matter which features are turned on), I get the attached crash.
<img width="258" alt="Screenshot 2022-09-07 at 22 38 23" src="https://user-images.githubusercontent.com/5508354/188962902-3d3eab0e-3ca7-44fd-93cc-2059c0c203c8.png">

When adding the runtimeExecutable env var as suggested, it works, but then breakpoints are grayed out like so:
<img width="543" alt="Screenshot 2022-09-07 at 22 45 28" src="https://user-images.githubusercontent.com/5508354/188963836-4ecf0e2d-2ea8-4d38-b04d-b58c2aa93852.png">

### Steps to Reproduce

Debug a Node process in VS Code with mirrord enabled.

### Your operating system and version

MacOS M1

### Local process

node

infiniteregrets:
Able to debug the node process, this depends on user's env config from what I can see unrelated to mirrord/extension

eyalb181:
What was the change to the env config that fixed it? Also I thought it worked with mirrord disabled but failed with mirrord enabled?

eyalb181:
This might be happening because we're trying to load into the arm64e compiled debugger. Waiting for the fix @infiniteregrets before testing again.

eyalb181:
Still getting this on 3.0.10.

infiniteregrets:
@eyalb181 does it work normally without mirrord enabled?

eyalb181:
Yup

eyalb181:
Still occurring as of 3.17.0

aviramha:
Updated original post to include workaround for now.

infiniteregrets:
@eyalb181 with sip in place i believe it should work for you.

I just gave it a try, it seems to work!


<img width="1440" alt="Screen Shot 2023-06-15 at 1 51 07 AM" src="https://github.com/metalbear-co/mirrord/assets/65443164/ee7471c3-02f1-46a5-b23c-71f7bf0120af">

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "Launch Program",
            "skipFiles": [
                "<node_internals>/**"
            ],
[...]

Issue #308, opened 2022-08-30
IntelliJ Extension: prevent loading the env into the compiler


infiniteregrets:
After a run on the goland IDE and intellij ultimate with golang plugn:
it seems like the process is being run like so:

```
GOROOT=/opt/homebrew/Cellar/go/1.18.4/libexec #gosetup
GOPATH=/Users/mehularora/go #gosetup
/opt/homebrew/Cellar/go/1.18.4/libexec/bin/go build -o /private/var/folders/sh/wljlm6jj0bd472x2c7c1tcxw0000gn/T/GoLand/___10go_build_main_go -gcflags all=-N -l /Users/mehularora/Documents/go-test/main.go #gosetup
```
which means we are being loaded into the compiler rather than the process after we inject the env as done in #300 

we have the following options here, 
- understand if we can make intellij not load us into the compilation step
- infer that it is the compiler we are loading into at layer level and not load/if its the debugger then also don't load

Issue #298, opened 2022-08-30
IntelliJ Extension - GoLand MacOS doesn't work


aviramha:
Nothing happens, the mirrord button on the toolbar is pressed, I try to launch a debug session and was expecting a dialog but it's stuck there. MacOs 11.6 with Goland 2022.2.2

_Originally posted by @simobgl in https://github.com/metalbear-co/mirrord/issues/280#issuecomment-1230942196_

infiniteregrets:
The above linked patch is for the untimate edition of intellij IDEA, however when starting goland I run into the following exception:
```
Internal error. Please refer to https://jb.gg/ide/critical-startup-errors

java.util.concurrent.CompletionException: com.intellij.diagnostic.PluginException: Fatal error initializing 'com.intellij.ide.ui.LafManager' [Plugin: com.intellij]
    at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
    at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
    at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1739)
    at com.intellij.util.concurrency.BoundedTaskExecutor.doRun(BoundedTaskExecutor.java:216)
    at com.intellij.util.concurrency.BoundedTaskExecutor.access$200(BoundedTaskExecutor.java:27)
    at com.intellij.util.concurrency.BoundedTaskExecutor$1.execute(BoundedTaskExecutor.java:195)
    at com.intellij.util.concurrency.BoundedTaskExecutor$1.run(BoundedTaskExecutor.java:187)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
    at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:668)
    at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:665)
    at java.base/java.security.AccessController.doPrivileged(Native Method)
    at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1.run(Executors.java:665)
    at java.base/java.lang.Thread.run(Thread.java:829)
[...]

PR #434, opened 2022-09-18
Add MIRRORD_SKIP_PROCESSES + Goland support

infiniteregrets:
This pull request will close the following issues:
- #298
- #308 
- #368 

Following changes have been made:
- Introduce a new Environment Variable `MIRRORD_SKIP_PROCESSES` which takes in a semicolon-separated list of processes a user wants to skip i.e. not load mirrord into
- Add goland support, using the above mentioned Environment Variable, load into the go process by skipping delve and debugserver
- It was found that debugserver is arm64e when run with mirrord 
```
mehularora@Mehuls-MacBook-Air mirrord % file /Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver    
/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver: Mach-O universal binary with 3 architectures: [x86_64:Mach-O 64-bit executable x86_64
- Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64
- Mach-O 64-bit executable arm64] [arm64e]
/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver (for architecture x86_64):        Mach-O 64-bit executable x86_64
/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver (for architecture arm64): Mach-O 64-bit executable arm64
/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/A/Resources/debugserver (for architecture arm64e):        Mach-O 64-bit executable arm64e
```
the following error occurred when mirrord was loaded:
```
dyld[62821]: terminating because inserted dylib 'target/aarch64-apple-darwin/release/libmirrord_layer.dylib' could not be loaded: tried: 'target/aarch64-apple-darwin/release/libmirrord_layer.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'arm64e')),
```
to handle this, we split the code into functions that handle the core logic of ctor and a function which decides whether mirrord should be loaded or not. 
NOTE: using an std function in the layer causes a segfault and hence we dont enforce PAC by doing so


mirrord-layer/src/lib.rs
```diff
@@ -133,6 +140,17 @@ fn init() {
     ));
 }
 
+fn should_load() -> bool {
+    let config = LayerConfig::init_from_env().unwrap();
+    if let Some(avoid_list) = config.skip_processes {
+        let args = std::env::args().collect::<Vec<_>>();
+        let process = args.first().unwrap().split('/').last().unwrap();
+        !avoid_list.split(';').any(|x| x == process)
```
eyalb181, line 148: Stylewise, lets avoid non-meaningful variable names, even for iterators. This one could be named e.g. "process_to_avoid".

.github/workflows/release.yaml
```diff
@@ -57,6 +57,10 @@ jobs:
         run: cargo +nightly build --release -p mirrord-layer --target=x86_64-apple-darwin
       - name: build mirrord-layer macOS arm
         run: cargo +nightly build --release -p mirrord-layer --target=aarch64-apple-darwin
+      - name: build mirrord-layer macOS arm64e
+        run: |
+          cp target/aarch64-apple-darwin/release/libmirrord_layer.dylib target/aarch64-apple-darwin/release/libmirrord_layer_arm64e.dylib
+          printf '\x02' | dd of=target/aarch64-apple-darwin/release/libmirrord_layer_arm64e.dylib bs=1 seek=8 count=1 conv=notrunc
```
eyalb181, line 63: Please add some documentation here

intellij-ext/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ repositories {
 dependencies {
     implementation("io.kubernetes:client-java:16.0.0") {
         exclude(group = "org.slf4j", module = "slf4j-api")
+        exclude(group = "org.yaml", module = "snakeyaml")
```
eyalb181, line 29: What's this about?

intellij-ext/src/main/kotlin/com/metalbear/mirrord/GoRunConfig.kt
```diff
@@ -0,0 +1,38 @@
+package com.metalbear.mirrord
+
+import com.goide.execution.GoRunConfigurationBase
+import com.goide.execution.GoRunningState
+import com.goide.execution.extension.GoRunConfigurationExtension
+import com.intellij.execution.configurations.RunnerSettings
+import com.intellij.execution.target.TargetedCommandLineBuilder
+
+class GoRunConfig : GoRunConfigurationExtension() {
```
eyalb181, line 9: Why are these empty function implementations necessary?

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordListener.kt
```diff
@@ -102,8 +130,21 @@ class MirrordListener : ExecutionListener {
...
-    private fun getRunConfigEnv(env: ExecutionEnvironment): LinkedHashMap<String, String> {
-        val envMethod = env.runProfile.javaClass.getMethod("getEnvs")
-        return envMethod.invoke(env.runProfile) as LinkedHashMap<String, String>
+    private fun getRunConfigEnv(env: ExecutionEnvironment): LinkedHashMap<String, String>? {
+        val method = when (env.runProfile::class.simpleName) {
+            "GoApplicationConfiguration" -> return null
+            else -> "getEnvs"
+        }
+        return try {
+            val envMethod = env.runProfile.javaClass.getMethod(method)
```
eyalb181, line 139: So what happens here for Go? Does it always fail because the method is null?

eyalb181 (changes requested):

.github/workflows/release.yaml
```diff
@@ -57,6 +57,10 @@ jobs:
         run: cargo +nightly build --release -p mirrord-layer --target=x86_64-apple-darwin
       - name: build mirrord-layer macOS arm
         run: cargo +nightly build --release -p mirrord-layer --target=aarch64-apple-darwin
```
aviramha, line 59: tbh we can just have arm64e as arm64e can be loaded into arm64 process AFAIK

mirrord-layer/src/lib.rs
```diff
@@ -133,6 +140,17 @@ fn init() {
     ));
 }
 
+fn should_load() -> bool {
```
aviramha, line 143: add a UT for `should_load` and provide all arguments externally

mirrord-layer/src/lib.rs
```diff
@@ -81,6 +81,14 @@ pub(crate) static ENABLED_UDP_OUTGOING: OnceLock<bool> = OnceLock::new();
 
 #[ctor]
 fn init() {
```
aviramha, line 83: rename `init` to `before_init` and rename `init_detour` to `init`.
other names are welcome but `init_detour` is just wrong

intellij-ext/build.gradle.kts
infiniteregrets: This was a gradle warning saying the dependency was vulnerable? Something along the lines of this https://github.com/jeremylong/DependencyCheck/issues/4802 and since we dont use yaml so I excluded it

intellij-ext/src/main/kotlin/com/metalbear/mirrord/GoRunConfig.kt
infiniteregrets: `GoRunConfigurationExtension` is an abstract class, those methods are to be implemented

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordListener.kt
infiniteregrets: `"GoApplicationConfiguration" -> return null` if it is go we just return null from the function, cant find an Option type or anything along the lines. will look if there is a better way

eyalb181: Ah, haven't noticed the `return`. Confusing IMO to return from inside the if clause meant to assign a value to `method`. Why not just start the function with `return if env.runProfile::class.simpleName == 'GoApplicationConfiguration`?

eyalb181 (changes requested):

mirrord-layer/src/lib.rs
```diff
@@ -133,6 +141,14 @@ fn init() {
     ));
 }
 
+fn should_load(given_process: &str, config: &LayerConfig) -> bool {
```
aviramha, line 144: UT

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordListener.kt
```diff
@@ -102,8 +130,21 @@ class MirrordListener : ExecutionListener {
         super.processTerminating(executorId, env, handler)
     }
 
-    private fun getRunConfigEnv(env: ExecutionEnvironment): LinkedHashMap<String, String> {
-        val envMethod = env.runProfile.javaClass.getMethod("getEnvs")
-        return envMethod.invoke(env.runProfile) as LinkedHashMap<String, String>
+    private fun getRunConfigEnv(env: ExecutionEnvironment): LinkedHashMap<String, String>? {
+        return if (env.runProfile::class.simpleName == "GoApplicationConfiguration")
```
eyalb181, line 134:
```suggestion
        if (env.runProfile::class.simpleName == "GoApplicationConfiguration") return null
```
Then the rest of the function, without the else. Lets you avoid needless nesting, which makes functions less readable.

eyalb181 (changes requested):

mirrord-layer/src/lib.rs
eyalb181: @infiniteregrets I was referring to 'x', here

mirrord-layer/src/lib.rs
```diff
@@ -450,3 +466,46 @@ unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
...
 }
+
+#[cfg(test)]
+mod tests {
+
+    use rstest::*;
+
+    use super::*;
+
+    #[rstest]
```
aviramha, line 477: if you're leveraging rstest, most of the values can be parameterized/cases and then have test for not load and case for load..?

aviramha (approved):
LGTM, nitpick on the test.

PR #436, opened 2022-09-19
Bugfix: crashing on connect with UDP sockets.

t4lz:
Currently, when called with `--no-outgoing` and the app calls `connect` on a UDP socket, the hook fails and `connect` returns an error to the app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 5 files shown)


aviramha:
Nice catch!

t4lz:
Turns out https://github.com/metalbear-co/mirrord/pull/420 exists and should also fixes the issue, so as discussed with @meowjesty, I [pushed](https://github.com/metalbear-co/mirrord/pull/420/commits/0da44b88eecbca7f641c87e09569bf9e82086d85) the regression test and the changelog entry to the source branch of that PR, and I am closing this one.

PR #445, opened 2022-09-20
Add UDP outgoing test with separate pods. 

t4lz:
Currently, outgoing tests do not actually verify successful interception and forwarding. The traffic could be sent out directly from the application.
Add a test that actually verifies the existing outgoing UDP functionality (only with bind to non-0 port + call to `connect`):
Set up two services, one of them only reachable from within the cluster.
That internal service runs a server that logs UDP messages.
Impersonate the other service with mirrord, running an application that
sends a UDP message to the internal one. Read the logs from the
internal service to verify the outgoing UDP message arrived.

I left the last commit, which is only renaming one struct, unsquashed, because I think it would be easier to review the other changes separately. Please let me know in case you would rather I also squashed this commit before you start reviewing.


tests/src/sanity.rs
```diff
@@ -1067,35 +1089,77 @@ mod tests {
...
+            &internal_service.name,
         ];
-        let mut process = run(node_command, &service.pod_name, None, None).await;
 
-        // Listen for UDP message from mirrorded app.
-        let mut buf = [0; 27];
-        let amt = socket.recv(&mut buf).unwrap();
-        assert_eq!(amt, 27);
-        assert_eq!(buf, "Can I pass the test please?".as_ref()); // Sure you can.
+        // Meta-test: verify that the application cannot reach the internal service without
```
eyalb181, line 1126: Nice

tests/src/sanity.rs
```diff
@@ -1067,35 +1089,77 @@ mod tests {
...
+            Some(&target_service.namespace),
+            None,
+        )
+        .await;
         let res = process.child.wait().await.unwrap();
         assert!(res.success());
         process.assert_stderr();
+
+        // Verify that the UDP message sent by the application reached the internal service.
+        sleep(Duration::from_secs(1)); // Let the server time to print the log.
```
eyalb181, line 1155: Can we just sample the logs until the message shows up, with a timeout?

t4lz: Thanks for that feedback. Changed it to wait for an item on a log stream :)
CI is still stuck though.

eyalb181 (approved):

PR #448, opened 2022-09-20
Draft: Increase http timeouts in e2e.

t4lz:
Just want to check if this is maybe a cheap fix to the currently impossible CI.


t4lz:
> @t4lz What's the verdict?

It was decided to abandon this approach, disable those tests and implement integration tests instead.
Those tests are disabled on `main` and the first integration test is already running locally and will hopefully be ready for review today.

PR #452, opened 2022-09-21
Support Deployments + replace impersonated pod/container with --target

infiniteregrets:
This PR will add support to impersonate deployments. Deployments can be represented as `deployment/sample_dep/container/abc` or simply `deployment/sample_dep` where the container would default to the first container.

`--pod-name & --impersonated-container...` will be replaced with a generic flag/env variable `--target` which will look like as follows:
`--target=pod/temp` or `--target=pod/temp/container/new`


mirrord-layer/src/config.rs
```diff
@@ -67,3 +71,105 @@ pub struct LayerConfig {
...
+#[derive(Debug, PartialEq)]
+pub struct PodAndContainer(pub String, pub Option<String>);
+
+impl FromStr for PodAndContainer {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let split = s.split('/').collect::<Vec<&str>>();
+        match split.first() {
+            Some(&"pod") if split.len() == 2 => Ok(PodAndContainer(split[1].to_string(), None)),
```
meowjesty, line 167: This part is unclear to me, could you explain or add some comments?

mirrord-layer/src/config.rs
```diff
@@ -67,3 +71,105 @@ pub struct LayerConfig {
...
+            _ => Err(format!("Given target: {:?} is not a deployment.", s)),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct PodAndContainer(pub String, pub Option<String>);
+
+impl FromStr for PodAndContainer {
+    type Err = String;
```
meowjesty, line 162: `Err` should be a proper `Error`, not `String`, for example (feel free to use a better name).

```suggestion
    type Err = ParsePodNameError;
```

mirrord-layer/src/config.rs
```diff
@@ -67,3 +71,105 @@ pub struct LayerConfig {
...
+        let split = s.split('/').collect::<Vec<&str>>();
+        match split.first() {
+            Some(&"deployment") => Ok(Deployment(split[1].to_string())),
+            _ => Err(format!("Given target: {:?} is not a deployment.", s)),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct PodAndContainer(pub String, pub Option<String>);
```
meowjesty, line 159: Hmm I would suggest:

```suggestion
struct Pod {
    name: String,
    container: Option<String>,
}
```

as this struct is kinda being named after its fields anyway.

mirrord-layer/src/config.rs
```diff
@@ -67,3 +71,105 @@ pub struct LayerConfig {
...
+        pods_api: &Api<Pod>,
+        pod_name: &String,
+        container_name: &Option<String>,
+        pod_namespace: &str,
+        client: &Client,
+    ) -> Option<String> {
+        let pod = pods_api.get(pod_name).await.unwrap();
+        let node_name = &pod.spec.unwrap().node_name;
+        let container_statuses = &pod.status.unwrap().container_statuses.unwrap();
+        let container_info = if let Some(container_name) = container_name {
```
meowjesty, line 111: Can't you use `container_name.map(...)` instead of `If let Some` pattern?

mirrord-layer/src/pod_api.rs
```diff
@@ -255,21 +253,21 @@ async fn create_ephemeral_container_agent(
...
-        "targetContainerName": config.impersonated_container_name,
+        "targetContainerName": pod_target.container_name,
         "env": [{"name": "RUST_LOG", "value": config.agent_rust_log}],
         "command": agent_command_line,
     }))?;
     debug!("Requesting ephemeral_containers_subresource");
 
     let mut ephemeral_containers_subresource = pods_api
-        .get_subresource("ephemeralcontainers", &config.impersonated_pod_name)
+        .get_subresource("ephemeralcontainers", &pod_target.pod_name)
```
meowjesty, line 263: The `pod_target` will be consumed in this function, so why not do a de-structuring assignment to reduce the amount of `pod_target.`? Do it like so:

```rust
let PodTarget { pod_name, container_name } = pod_target;
```

You could go one step further and be even fancier by doing this at the function parameter:

```rust
async fn create_ephemeral_container_agent(PodData { pod_name, container_name }: PodData)
```

Up to you right now (and maybe open a discussion around this style to let our pals vote).

mirrord-layer/src/pod_api.rs
```diff
@@ -305,10 +303,10 @@ async fn create_ephemeral_container_agent(
         }
     }
 
-    wait_for_agent_startup(pods_api, &config.impersonated_pod_name, mirrord_agent_name).await?;
+    wait_for_agent_startup(pods_api, &pod_target.pod_name, mirrord_agent_name).await?;
 
     debug!("container is ready");
-    Ok(config.impersonated_pod_name.clone())
+    Ok(pod_target.pod_name.clone())
```
meowjesty, line 309: Why do we `clone` here?

mirrord-layer/src/pod_api.rs
```diff
@@ -417,3 +415,136 @@ async fn create_job_pod_agent(
...
+            ))),
+        }
+    }
+}
+
+impl PodData {
+    pub async fn container_id(&self, pods_api: &Api<Pod>) -> Option<String> {
+        let pod = pods_api.get(&self.pod_name).await.ok()?;
+        let container_statuses = &pod.status?.container_statuses?;
+        // TODO: should the return type be a Result here?
```
meowjesty, line 527: I'm in favor of returning `Result` for functions that can fail, for example on the `pods_api_.get` line, it feels like you're eating up an error?

meowjesty (changes requested):
Looking cleaner overall.

mirrord-layer/src/pod_api.rs
```diff
@@ -49,52 +52,36 @@ struct RuntimeData {
...
-                    )
-                })?
-                .container_id
-        } else {
-            info!("No container name specified, defaulting to first container found");
-            &container_statuses.first().unwrap().container_id
-        };
-
-        let container_info = container_info
+    async fn from_k8s(client: Client, target: &str, pod_namespace: &str) -> Result<Self> {
```
eyalb181, line 55: Looks like pod_namespace should be renamed?

mirrord-layer/src/config.rs
```diff
@@ -14,15 +14,12 @@ pub struct LayerConfig {
     #[envconfig(from = "MIRRORD_AGENT_IMAGE_PULL_POLICY", default = "IfNotPresent")]
     pub image_pull_policy: String,
 
-    #[envconfig(from = "MIRRORD_AGENT_IMPERSONATED_POD_NAME")]
-    pub impersonated_pod_name: String,
+    #[envconfig(from = "MIRRORD_TARGET")]
```
eyalb181, line 17: Is the input format documented anywhere?

mirrord-cli/src/config.rs
```diff
@@ -19,9 +19,9 @@ pub(super) enum Commands {
 
 #[derive(Args, Debug)]
 pub(super) struct ExecArgs {
-    /// Pod name to mirror.
+    /// Target name to mirror.
     #[clap(short, long, value_parser)]
-    pub pod_name: String,
```
eyalb181, line 24: Lets keep the old args available with a deprecation warning

mirrord-layer/src/pod_api.rs
```diff
@@ -136,16 +122,21 @@ pub(crate) async fn create_agent(
...
 
     let pod_name = if ephemeral_container {
-        create_ephemeral_container_agent(&config, agent_image, &pods_api, connection_port).await?
-    } else {
-        let runtime_data = RuntimeData::from_k8s(
-            client.clone(),
-            &impersonated_pod_name,
-            &impersonated_pod_namespace,
-            &impersonated_container_name,
+        let pod_target = target.parse::<PodData>()?;
```
eyalb181, line 125: What happens if target is a deployment?

mirrord-cli/src/config.rs
eyalb181: This will also let us defer the changes in the IDE extensions for now - make sure to test that this change didn't break them.

infiniteregrets: are you saying we keep the previous functionality as well?

eyalb181: The definition here is, we don't want anything using the old parameters to break.

infiniteregrets: using clap, i believe, not sure, we can only provide one primary argument? i think it can be either of --target or --pod-name, still looking into clap

eyalb181: What makes it a primary argument?

infiniteregrets: i mean --pod-name is required, if a user does not specify one then mirrord exec does not work

eyalb181: So this point is resolved with ArgGroup right?

infiniteregrets: yup resolved now

mirrord-layer/src/pod_api.rs
infiniteregrets: @eyalb181 thoughts? do we want to propagate the error from here or just return None if anything bad happens and just say container not found?

mirrord-layer/src/pod_api.rs
```diff
@@ -49,52 +52,37 @@ struct RuntimeData {
...
-        };
-
-        let container_info = container_info
+    async fn from_k8s(client: Client, target: &str, target_namespace: &str) -> Result<Self> {
+        let target = target.parse::<Target>()?;
+
+        let pods_api: Api<Pod> = Api::namespaced(client.clone(), target_namespace);
+        let deployment_api: Api<Deployment> = Api::namespaced(client.clone(), target_namespace);
+
+        // TODO: figure a way to pass the Api in a generic way, don't want to pass two different
```
eyalb181, line 61: Why aren't you passing `client` rather than the APIs, then letting `PodData`/`DeploymentData` initialize the API inside `container_info`?

mirrord-layer/src/pod_api.rs
```diff
@@ -126,32 +113,37 @@ pub(crate) async fn create_agent(
...
 
     let pod_name = if ephemeral_container {
-        create_ephemeral_container_agent(&config, agent_image, &pods_api, connection_port).await?
-    } else {
-        let runtime_data = RuntimeData::from_k8s(
-            client.clone(),
-            &impersonated_pod_name,
-            &impersonated_pod_namespace,
-            &impersonated_container_name,
+        let target = target.parse::<Target>()?;
```
eyalb181, line 129: Why is target only parsed here, but not in the Job branch?

mirrord-layer/src/pod_api.rs
infiniteregrets: thought the same before, that generic thing was a bad design here probably. fixed, thanks!

mirrord-layer/src/lib.rs
```diff
@@ -85,12 +85,36 @@ fn before_init() {
...
         let given_process = args.first().unwrap().split('/').last().unwrap();
         let config = LayerConfig::init_from_env().unwrap();
+        deprecation_check(&config);
         if should_load(given_process, &config.skip_processes) {
             init(config);
         }
     }
 }
 
+fn deprecation_check(config: &LayerConfig) {
```
eyalb181, line 95: Please demarcate with comments exactly which code should be removed on 24.10

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +381,252 @@ async fn create_job_pod_agent(
...
+                .await
+                .ok_or_else(|| LayerError::PodNotFound(self.clone())),
+            Target::Deployment(deployment) => deployment
+                .container_data(&deployment_api, &pod_api)
+                .await
+                .ok_or_else(|| LayerError::DeploymentNotFound(deployment.deployment.clone())),
+        }
+    }
+
+    fn target_apis(&self, client: &Client, namespace: &str) -> Result<(Api<Pod>, Api<Deployment>)> {
```
eyalb181, line 466: Seems unnecessary to split it into a separate method since it's only used in one place inside the struct itself.

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +381,252 @@ async fn create_job_pod_agent(
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum Target {
+    Deployment(DeploymentData),
+    Pod(PodData),
+}
+
+impl Target {
+    pub async fn container_info(&self, client: &Client, namespace: &str) -> Result<Response> {
```
eyalb181, line 452: Why is this called container_info and the other methods called container_data? Seems inconsistent.

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +381,252 @@ async fn create_job_pod_agent(
...
+        let (pod_api, deployment_api) = self.target_apis(client, namespace)?;
+        match self {
+            Target::Pod(pod) => pod
+                .container_data(&pod_api)
```
eyalb181, line 456: Can't `container_data` be abstracted behind a trait?

mirrord-layer/src/pod_api.rs
```diff
@@ -46,56 +49,24 @@ struct RuntimeData {
...
+            })?
+            .parse::<ContainerRuntime>()?;
 
-        Ok(RuntimeData {
-            container_id: container_id.to_string(),
-            container_runtime: container_runtime.to_string(),
-            node_name: node_name.as_ref().unwrap().to_string(),
-            socket_path: socket_path.to_string(),
-        })
+        target.merge(response, container_info)
```
eyalb181, line 69: Not sure what you're going for here. `response_data` and `container_info` both originate in `target`. Why is it the job of `RuntimeData` to merge them? Why doesn't `target` just return RuntimeData?
Taking this a bit further, is `impl RuntimeData` even necessary anymore, or can `from_k8s` be a method of `target`?

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +381,252 @@ async fn create_job_pod_agent(
...
+        }
+    }
+}
+
+impl PodData {
+    pub async fn container_data(&self, pods_api: &Api<Pod>) -> Option<Response> {
+        let pod = pods_api.get(&self.pod_name).await.ok()?;
+        let container_statuses = &pod.clone().status?.container_statuses?;
+        // TODO: should the return type be a Result here?
+        let container_info = if let Some(container_name) = &self.container_name {
```
eyalb181, line 559: So is `container_info` just the container ID?

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +381,252 @@ async fn create_job_pod_agent(
...
         .and_then(|pod| pod.metadata.name.clone())
-        .ok_or(LayerError::PodNotFound(mirrord_agent_job_name))?;
+        .ok_or(LayerError::JobPodNotFound(mirrord_agent_job_name))?;
 
     wait_for_agent_startup(pods_api, &pod_name, "mirrord-agent".to_string()).await?;
     Ok(pod_name)
 }
+
+#[derive(Debug, Clone)]
+pub(crate) struct Response {
```
eyalb181, line 391: It seems weird to have a struct called `Response` here since it's not a communication library. Is there a name that makes more sense?

eyalb181 (changes requested):

mirrord-layer/src/pod_api.rs
infiniteregrets: both use different number of args

mirrord-layer/src/pod_api.rs
infiniteregrets: its runtime + container id

mirrord-layer/src/pod_api.rs
infiniteregrets: `TargetResult`?

mirrord-layer/src/pod_api.rs
infiniteregrets: - i was merging it in target, because I don't have to pass more data in function arguments for the custom errors and can be referred by self
- yup, I don't think we need it. i was just trying to preserve what was done previously

mirrord-layer/src/pod_api.rs
infiniteregrets: agreed, i will change the naming to something more meaningful once everything else is resolved

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +390,342 @@ async fn create_job_pod_agent(
...
+    Pod(PodData),
+}
+
+#[async_trait]
+trait RuntimeDataProvider {
+    async fn runtime_data(&self, client: &Client, namespace: &str) -> Result<RuntimeData>;
+}
+
+impl Target {
+    pub async fn runtime_info(&self, client: &Client, namespace: &str) -> Result<RuntimeData> {
```
eyalb181, line 575: Why's this called runtime_info if it returns runtime_data?

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +390,342 @@ async fn create_job_pod_agent(
...
+        match self {
+            Target::Pod(pod) => pod.runtime_data(client, namespace).await,
```
eyalb181, line 577: Now that they both take the same params, can you hide the function behind a trait?

eyalb181 (changes requested):

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +390,342 @@ async fn create_job_pod_agent(
...
+                    .map(|container| container.name.clone())
+            }?;
+            let container_info = if let Some(container_name) = &self.container_name {
+                container_statuses
+                    .iter()
+                    .find(|&status| &status.name == container_name)?
+                    .container_id
+                    .clone()
+            } else {
+                info!("No container name specified, defaulting to first container found");
```
eyalb181, line 660: If I understand correctly, you've already got container_name at this point (whether it was the one the user specified, or the name of the first container), so this 'else' clause isn't necessary.

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +390,342 @@ async fn create_job_pod_agent(
...
+                container_statuses.first()?.container_id.clone()
+            }?;
+            Some((container_info, container_name))
+        }()
+        .ok_or_else(|| {
```
eyalb181, line 665: Why do we use Option in the block above, if we immediately convert it to Result? Why not use Result to begin with?

mirrord-layer/src/pod_api.rs
```diff
@@ -124,47 +71,76 @@ pub(crate) async fn create_agent(
...
-        agent_namespace
-            .as_ref()
-            .unwrap_or(&impersonated_pod_namespace),
-    );
-
     let agent_image = agent_image.unwrap_or_else(|| {
         concat!("ghcr.io/metalbear-co/mirrord:", env!("CARGO_PKG_VERSION")).to_string()
     });
 
+    // DEPRECATED: - Scheduled for removal on [DATE]
```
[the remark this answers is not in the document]
infiniteregrets: its just a temporary marker, please ignore for now

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +392,345 @@ async fn create_job_pod_agent(
...
+    socket_path: String,
+    container_id: String,
+}
+
+impl FromStr for ContainerRuntime {
+    type Err = LayerError;
+    fn from_str(input: &str) -> Result<Self> {
+        let container_runtime_info = input.split("://").collect::<Vec<&str>>();
+        let (container_runtime, socket_path) = match container_runtime_info.first() {
+            Some(&"docker") => ("docker", "/var/run/docker.sock"),
```
eyalb181, line 714: Why does this need to populate `container_runtime` with the same value that's already in `container_runtime_info.first()`?

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +392,345 @@ async fn create_job_pod_agent(
...
+                    .containers
+                    .first()
+                    .map(|container| container.name.clone())?
+            };
+            let container_info = container_statuses
+                .iter()
+                .find(|&status| status.name == container_name)?
+                .container_id
+                .clone()?;
+            Some((container_info, container_name))
```
eyalb181, line 664: This is confusing - you're populating container_name first and then container_info, why are you returning them in the opposite order?

mirrord-layer/src/pod_api.rs
```diff
@@ -412,8 +392,345 @@ async fn create_job_pod_agent(
...
+            let container_name = if let Some(container_name) = self.container_name.clone() {
+                container_name
+            } else {
+                pod.clone()
+                    .spec?
+                    .containers
+                    .first()
+                    .map(|container| container.name.clone())?
+            };
+            let container_info = container_statuses
```
eyalb181, line 659: `container_info` is called `runtime_and_id` a few rows down, why the inconsistency?
(Also `container_info` isn't a good name - how's it different from `container_data`, which is used for completely different things elsewhere in the code?)

eyalb181 (approved):

.github/workflows/ci.yaml
```diff
@@ -51,6 +51,25 @@ jobs:
       - run: |
           cd mirrord-protocol
           cargo test
+
+  test_mirrord_config:
```
eyalb181, line 55: Is this supposed to be part of this PR?

infiniteregrets: nope this is from dima's pr, probably came along because of the merge resolution?

meowjesty (approved):
🚀

eyalb181 (approved):

PR #454, opened 2022-09-21
Support Go in VS Code, and fix file ops and steal

eyalb181:


t4lz (approved):

.github/workflows/dummy.yaml
```diff
@@ -0,0 +1,59 @@
+# https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks/
+name: CI
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
```
t4lz, line 7: Could this cause a false positive on the CI verification? Like: I do a large change in core files,  but also in one of these files, e.g. `CHANGELOG.md`, both `ci.yaml` and `dummy.yaml` run, `ci.yaml` fails but `dummy.yaml` of course produces successes - what happens then, does GitHub allow me to merge my changes because I have a successful job for each step?
If this problem indeed exists, a solution would be to add `paths-ignore` for all other paths in the repo.

eyalb181: Just tested - both workflows run, but if one fails then it fails the check.

PR #478, opened 2022-09-28
Timeout and user-friendly message on first connection

### LINKED ISSUE #348: Mirrord should exit and display helpful message when receiving close frame.
When mirrord gets the close frame currently it just keeps running uselessly, and even after the agent disconnects.
Instead, when the connection is closed, mirrord should exit and display some helpful information.

This happens for example if you run mirrord against a local cluster on mac with an apple chip, without specifying an agent image (`MIRRORD_AGENT_IMAGE`) built for apple, and an x86 image is used by default.

DmitryDodzin:
closes #348


mirrord-layer/src/lib.rs
```diff
@@ -420,16 +422,26 @@ async fn start_layer_thread(
...
 
-            for (key, value) in remote_env_vars.into_iter() {
-                std::env::set_var(&key, &value);
-                debug_assert_eq!(std::env::var(key), Ok(value));
+                for (key, value) in remote_env_vars.into_iter() {
+                    std::env::set_var(&key, &value);
+                    debug_assert_eq!(std::env::var(key), Ok(value));
+                }
+            } else {
+                gracefull_exit!("unexpected response - expected env vars response {msg:?}");
```
eyalb181, line 435: graceful*

eyalb181 (approved):

PR #481, opened 2022-09-29
New dns resolving + fgets

meowjesty:
- Uses [`trust-dns-resolver`](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/) for our `getaddrinfo` hook;
- Adds `fgets` and `ferror` implementations;
- Remove `tracing` from `*_detour`s;


mirrord-agent/src/dns.rs
```diff
@@ -1,87 +1,76 @@
...
         Self { request, tx }
     }
 }
 
-trait AddrInfoHintExt {
-    fn into_lookup(self) -> dns_lookup::AddrInfoHints;
-}
+#[tracing::instrument(level = "debug")]
+async fn dns_lookup(root_path: &Path, host: String) -> RemoteResult<DnsLookup> {
+    let resolv_conf_path = root_path.join("etc").join("resolv.conf");
```
aviramha, line 30: I suggest to consider creating a resolver once and re-using it for performance, it's a bit of a nitpick but might have value down the road.
While writing this comment I thought about the situation where the files might change, so this needs to be hash-based, so maybe just a comment for later on improvement.

mirrord-agent/src/file.rs
```diff
@@ -198,6 +188,52 @@ impl FileManager {
...
+    ) -> RemoteResult<ReadStringFileResponse> {
+        self.open_files
+            .get_mut(&fd)
+            .ok_or(ResponseError::NotFound(fd))
+            .and_then(|remote_file| {
+                if let RemoteFile::File(file) = remote_file {
+                    let mut reader = BufReader::new(std::io::Read::by_ref(file));
+                    let mut buffer = String::with_capacity(buffer_size);
+                    let read_result = reader
+                        .read_line(&mut buffer)
```
aviramha, line 205: This is a bit ugly (conceptually, not the implementation). I'd document it carefully as *I think* I understand what's going on, but better safe than sorry.
(I understood that the "best" way to read line in Rust is using BufReader, which might advance the file offset further than it'd have been if we just used fgets due to the buffering.. R_R)

mirrord-layer/src/file/hooks.rs
```diff
@@ -136,7 +136,7 @@ pub(super) unsafe extern "C" fn fdopen_detour(fd: RawFd, raw_mode: *const c_char
 /// `open_detour`.
 /// `fd` for a file descriptor with the `O_DIRECTORY` flag.
 #[hook_guard_fn]
-#[tracing::instrument(level = "trace", skip(raw_path))]
+#[tracing::instrument(level = "debug", skip(raw_path))]
```
aviramha, line 139: prefer to remove this all together (tracing directly from hook....), same for above if already touched.

mirrord-layer/src/file/hooks.rs
```diff
@@ -253,6 +255,78 @@ pub(crate) unsafe extern "C" fn fread_detour(
     }
 }
 
+#[hook_guard_fn]
+#[tracing::instrument(level = "trace", skip(out_buffer, file_stream))]
```
aviramha, line 259: no tracing hooks F:

mirrord-layer/src/file/hooks.rs
```diff
@@ -253,6 +255,78 @@ pub(crate) unsafe extern "C" fn fread_detour(
...
+
+        let (Ok(result) | Err(result)) = read_result.map_err(From::from);
+        result
+    } else {
+        FN_FGETS(out_buffer, capacity, file_stream)
+    }
+}
+
+#[hook_guard_fn]
+#[tracing::instrument(level = "trace", skip(file_stream))]
```
aviramha, line 305: same

aviramha (commented):
Looks great! Few comments left inline.

mirrord-layer/src/file/hooks.rs
meowjesty: Removed!

mirrord-agent/src/file.rs
meowjesty: It's less about being the best, and more about `read_line` not being implemented for `File`. Added some comments there, do they make things clearer?

mirrord-agent/src/dns.rs
meowjesty: Added a TODO there. Agree that we should be caching this, hopefully it doesn't hit us too hard for now.

aviramha: Probably not but it's good to have speed ups at hand in case we need :)

mirrord-agent/src/file.rs
aviramha: I think a short paragraph before the implementation is better than inline comments.

meowjesty: Added a proper doc for this. Left some of the inline comments (reduced the text as some was moved to the doc). What do you think?

mirrord-layer/src/file/hooks.rs
```diff
@@ -253,6 +251,79 @@ pub(crate) unsafe extern "C" fn fread_detour(
...
+pub(crate) unsafe extern "C" fn fgets_detour(
+    out_buffer: *mut c_char,
+    capacity: c_int,
+    file_stream: *mut FILE,
+) -> *mut c_char {
+    // Extract the fd from stream and check if it's managed by us, or should be bypassed.
+    let fd = fileno_logic(file_stream);
+
+    // We're only interested in files that are handled by `mirrord-agent`.
+    let remote_fd = OPEN_FILES.lock().unwrap().get(&fd).cloned();
```
aviramha, line 266: wouldn't the lock be held for rest of the operation?

mirrord-layer/src/file/hooks.rs
```diff
@@ -253,6 +251,79 @@ pub(crate) unsafe extern "C" fn fread_detour(
...
+}
+
+/// Used to distinguish between an error or `EOF` (especially relevant for `fgets`).
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn ferror_detour(file_stream: *mut FILE) -> c_int {
+    // Extract the fd from stream and check if it's managed by us, or should be bypassed.
+    let fd = fileno_logic(file_stream);
+
+    // We're only interested in files that are handled by `mirrord-agent`.
+    let remote_fd = OPEN_FILES.lock().unwrap().get(&fd).cloned();
```
aviramha, line 309: same?

mirrord-layer/src/file/hooks.rs
```diff
@@ -313,6 +382,7 @@ pub(crate) unsafe extern "C" fn write_detour(
     buffer: *const c_void,
     count: size_t,
 ) -> ssize_t {
+    // Avoid writing to `std(in|out|err)`.
```
aviramha, line 385: gj

aviramha (approved):
overall looking good, few comments.

PR #489, opened 2022-10-04
Changed errors that aren't fatal and can be part of a happy flow to b…

aviramha:
…e info instead of error (log level change).

### DIFF
--- a/mirrord-layer/src/error.rs
+++ b/mirrord-layer/src/error.rs
@@ -6,7 +6,7 @@ use libc::FILE;
 use mirrord_protocol::{tcp::LayerTcp, ConnectionId, ResponseError};
 use thiserror::Error;
 use tokio::sync::{mpsc::error::SendError, oneshot::error::RecvError};
-use tracing::{error, info, warn};
+use tracing::{error, info};
 
 use super::HookMessage;
 // use crate::pod_api::Target;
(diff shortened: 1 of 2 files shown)


mirrord-layer/src/error.rs
```diff
@@ -176,7 +176,7 @@ impl From<HookError> for i64 {
             | HookError::BypassedType(_)
             | HookError::BypassedDomain(_)
             | HookError::BypassedPort(_) => {
-                warn!("Recoverable issue >> {:#?}", fail)
+                info!("Non-issue libc error >> {:#?}", fail)
```
eyalb181, line 179:
```suggestion
                info!("libc error (doesn't indicate a problem) >> {:#?}", fail)
```

mirrord-layer/src/error.rs
```diff
@@ -195,7 +195,7 @@ impl From<HookError> for i64 {
             | HookError::ResponseError(ResponseError::NotDirectory(_))
             | HookError::ResponseError(ResponseError::Remote(_))
             | HookError::ResponseError(ResponseError::RemoteIO(_)) => {
-                error!("Error occured in Layer >> {:?}", fail)
+                info!("Non-issue libc error >> {:#?}", fail)
```
eyalb181, line 198:
```suggestion
                info!("libc error (doesn't indicate a problem) >> {:#?}", fail)
```

eyalb181 (changes requested):

eyalb181 (approved):

Issue #472, opened 2022-09-24
Add integration tests for traffic mirroring


eyalb181:
Since we've disabled traffic mirroring tests in the e2e suite because they were flakey, we need to cover this functionality in hopefully less flakey integration tests.

PR #491, opened 2022-10-04
First integration test.

t4lz:
https://github.com/metalbear-co/mirrord/issues/472

https://github.com/metalbear-co/mirrord/discussions/450


.github/workflows/ci.yaml
```diff
@@ -168,10 +168,47 @@ jobs:
       - run: sudo apt install -y libpcap-dev cmake
       - run: cargo +nightly build --manifest-path=./Cargo.toml
 
+  integration_tests:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-12, macos-11]
+        target:
+          [x86_64-unknown-linux-gnu, x86_64-apple-darwin, aarch64-apple-darwin]
```
aviramha, line 176: not sure how building aarch64 helps here as we can't run integration tests of it?

.github/workflows/ci.yaml
```diff
@@ -168,10 +168,47 @@ jobs:
...
+          - os: ubuntu-latest
+            target: aarch64-apple-darwin
+          - os: macos-12
+            target: x86_64-unknown-linux-gnu
+          - os: macos-11
+            target: aarch64-apple-darwin
+          - os: macos-11
+            target: x86_64-unknown-linux-gnu
+    runs-on: ${{ matrix.os }}
+    needs: [build_mirrord, layer_integration_test]
```
aviramha, line 189: there's no such test?

.github/workflows/ci.yaml
```diff
@@ -168,10 +168,47 @@ jobs:
...
+          sudo apt-get install -y libpcap-dev cmake
+      - uses: actions/setup-python@v3
+      - run: pip3 install flask
+      - run: cargo +nightly test --test integration
+
   e2e:
     runs-on: ubuntu-latest
     name: e2e
-    needs: [build_mirrord, test_agent_image]
+    needs: [build_mirrord, test_agent_image, layer_integration_test]
```
aviramha, line 211: no `layer_integration_Test`

mirrord-layer/tests/integration.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Tell the layer there is a new incoming connection, then send data "from that connection".
+async fn send_connection_then_data(
+    codec: &mut Framed<TcpStream, DaemonCodec>,
+    message_data: &str,
+    connection_id: u64,
+) {
+    send_new_connection(codec, connection_id).await;
+    send_tcp_data(codec, message_data, connection_id).await;
+}
+
```
aviramha, line 77: Idea for improvement - create a struct that can be created same as `accept_library_connection` then have methods such as above (based on the shared requirement for having codec)

mirrord-layer/tests/integration.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Run python with shell resolving to find the actual executable.
+///
+/// This is to help tests that run python with mirrord work locally on systems with pyenv.
+/// If we run `python3` on a system with pyenv the first executed is not python but bash. On mac
+/// that prevents the layer from loading because of SIP.
+async fn get_python3_executable() -> String{
```
aviramha, line 83: That's actually a great solution that we can use for fixing pyenv with the mirrord cli :)

aviramha (commented):
Great progress!

.github/workflows/ci.yaml
t4lz: Right, thanks

.github/workflows/ci.yaml
aviramha: unless we run emulation (not sure we even can) there are no GitHub runners for Apple Sillicon..

mirrord-layer/tests/integration.rs
t4lz: @aviramha like this?

aviramha: Yup. Please note that in Rust you don't have to type self
```rs
async fn send_new_connection(self: &mut Self) -> u64
```
can be written as
```rs
async fn send_new_connection(&mut self) -> u64
```

mirrord-layer/tests/integration.rs
```diff
@@ -0,0 +1,183 @@
...
+            .unwrap();
+        codec
+    }
+
+    /// Accept the library's connection and verify initial ENV message.
+    /// Handle python's 2 process behaviour.
+    async fn get_initialized_connection(listener: &TcpListener) -> LayerConnection {
+        let mut codec = Self::accept_library_connection(listener).await;
+        let msg = match codec.next().await {
+            Some(option) => option.unwrap(),
```
aviramha, line 72: I'd assert this behavior instead of making it a fallback (i.e expect it to happen or expect it not to happen based on current run parameters.)

t4lz: I made the test require two connections when running with the python server and [it looks like there is no second connection on the CI](https://github.com/metalbear-co/mirrord/actions/runs/3188716994/jobs/5201728879#step:8:35). Can it be that the python on the CI does not fork?
The test does pass locally.

mirrord-layer/tests/integration.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+/// Build the cdylib and return a path to it.
+#[fixture]
+#[once]
+fn dylib() -> PathBuf {
+    println!("Building project...");
+    let before_build = Instant::now();
+    let dylib_path = test_cdylib::build_current_project();
```
t4lz, line 178: Building can take quite some time on the CI (25 seconds in [a typical example](https://github.com/metalbear-co/mirrord/actions/runs/3188761789/jobs/5201818583#step:8:29) and 121 seconds in [this extreme case](https://github.com/metalbear-co/mirrord/actions/runs/3188136393/jobs/5200519924#step:8:600)). Should we maybe try to use an artefact of the `build_mirrord` job (and introduce a `build_mirrord_macos` job)? That would save 1 build (so typically 20~30 seconds) and would allow this test to have a tighter timeout without being flakey.

aviramha: sure

mirrord-layer/tests/integration.rs
t4lz: On the CI there is only one connection also on mac. I'm not sure on what cases there are two connections and on what  cases there's only one. Should we maybe return to the fallback strategy?

aviramha: weird, okay then :octocat:

aviramha (approved):
Great job!

aviramha (approved):

Issue #494, opened 2022-10-05
No warnings on build.


t4lz:
There are some new warnings showing when building the project with cargo.

```
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-agent/Cargo.toml: dependency (trust-dns-resolver) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-agent/Cargo.toml: unused manifest key: dependencies.trust-dns-resolver.worspace
warning: /Users/t4lz/Documents/projects/mirrord/tests/Cargo.toml: unused manifest key: dependencies.mirrord.bin
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-layer/Cargo.toml: dependency (trust-dns-resolver) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-layer/Cargo.toml: unused manifest key: dependencies.trust-dns-resolver.worspace
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-protocol/Cargo.toml: dependency (trust-dns-resolver) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
warning: /Users/t4lz/Documents/projects/mirrord/mirrord-protocol/Cargo.toml: unused manifest key: dependencies.trust-dns-resolver.worspace
```
It would be nice if there were no warnings at all.
The issues they are warning about should be solved, or if they're false alarm they should be silenced.

t4lz:
`good_first_issue`?

t4lz:
Actually I still get
```
warning: /Users/t4lz/Documents/projects/mirrord/tests/Cargo.toml: unused manifest key: dependencies.mirrord.bin
```
That one isn't new, but maybe someone wants to see how to solve or silence it :)

infiniteregrets:
I guess a way to silence it would be to actually use mirrord in the tests and just add allow dead code config on top

PR #495, opened 2022-10-06
Add SIP protection check

### LINKED ISSUE #412: mirrord exec - verify binary executes can be loaded into
Right now we might have a weird UX where we just don't load.
We can easily add simple verifications such as checking if the binary is protected by SIP (is in `/usr/bin` and other paths) and expand the functionality as we go.

infiniteregrets:
closes #412


eyalb181 (approved):

infiniteregrets:
also closes #494

PR #498, opened 2022-10-06
[Merged by Bors] - `Detour<S>` as our `ControlFlow`-like result for `*_detour`s + `pread` hook.

meowjesty:
- Adds `Detour<S>` to be used in our hooks to simplify some of our bypass handling (plenty of hooks used a `Err(bypass)` to be handled by `*_detour` as a sort of "Just call the original `libc` function);
- Adds `pread` hook (used by dotnet);
- Changes `IGNORE_FILES` regex set to ignore some dotnet paths/file types.


aviramha:
This looks like a great improvement! Well done!!

aviramha (approved):
Great job!

aviramha:
bors retry

PR #504, opened 2022-10-07
Add more http mirroring layer integration tests.

t4lz:
https://github.com/metalbear-co/mirrord/issues/472


eyalb181:
General comment: a bit too late in this case, but opt for small PRs - no need to get Flask + Go + Node in one go. Feel free to split Issues too to reflect that.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+### Added
+- Layer integration tests for more apps. Closes
+  [[#472](https://github.com/metalbear-co/mirrord/issues/472)].
```
t4lz, line 12: This PR adds a test with go 1.19.0. Do we also need a test with go 1.18.0? In that case I need to change the changelog entry:
```suggestion
- Layer integration tests for more apps. Addresses but does not close
  [[#472](https://github.com/metalbear-co/mirrord/issues/472)].
```

CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
...
```
t4lz, line 12: This PR adds a test with go 1.19.0. Do we also need one with go 1.18?
In that case I have to change this:
```suggestion
- Layer integration tests for more apps. Addresses but does not close
  [[#472](https://github.com/metalbear-co/mirrord/issues/472)] (go 1.18 still missing).
```

.github/workflows/ci.yaml
```diff
@@ -187,14 +187,26 @@ jobs:
         with:
           key: ubuntu-latest
           cache-on-failure: true
+      - uses: actions/setup-go@v3
```
eyalb181, line 190: Some comments could be good here as to why we install each requirement

.gitignore
```diff
@@ -15,6 +15,8 @@ tests/go-e2e-fileops/go-e2e-fileops
 tests/go-e2e-outgoing/go-e2e-outgoing
 tests/go-e2e-dns/go-e2e-dns
 
+mirrord-layer/tests/apps/app_go/19
```
eyalb181, line 18: This is oddly specific

mirrord-layer/tests/http_mirroring.rs
```diff
@@ -214,26 +273,26 @@ async fn test_mirroring_with_http(
     println!("Application subscribed to port, sending tcp messages.");
 
     layer_connection
-        .send_connection_then_data("GET / HTTP/1.1\r\n\r\n")
+        .send_connection_then_data("GET / HTTP/1.1\r\nHost: localhost\r\n\r\n")
```
eyalb181, line 276: Why was this change necessary?

mirrord-layer/tests/apps/app_node.js
```diff
@@ -0,0 +1,53 @@
...
+
+function exit() {
+  server.close();
+  process.exit();
+}
+
+app.get("/", (req, res) => {
+  console.log("GET: Request completed");
+  res.send("GET");
+  done[0] = true;
```
eyalb181, line 16: Consider using a dict with informative names rather than an array with indexes

mirrord-layer/tests/apps/app_node.js
```diff
@@ -0,0 +1,53 @@
...
+  done[1] = true;
+  if (done[0] && done[2] && done[3])
+    exit();
+});
+
+app.put("/", (req, res) => {
+  console.log("PUT: Request completed");
+  res.send("PUT");
+  done[2] = true;
+  if (done[0] && done[1] && done[3])
```
eyalb181, line 33: Probably a shorter way to do this, e.g. https://stackoverflow.com/questions/53897673/check-if-all-values-in-array-are-true-then-return-a-true-boolean-statement-jav

Besides being shorter, it would let you add more use cases without having to add to this hard coded list.

mirrord-layer/tests/apps/app_node.js
eyalb181: (The comments here apply to the other apps as well)

eyalb181 (changes requested):

mirrord-layer/tests/http_mirroring.rs
t4lz: Without it, one of the apps said the request is invalid.

.gitignore
t4lz: Haha why? It's the binary of the go test app, like all the lines above it.

eyalb181: Ah my mistake

t4lz:
Converted to draft since it turns out [go was still unstable](https://github.com/metalbear-co/mirrord/actions/runs/3218837773/jobs/5263569770). Now tried this new strategy from https://stackoverflow.com/questions/39320025/how-to-stop-http-listenandserve. Not sure about this yet since I have no experience with Go.

mirrord-layer/tests/apps/app_node.js
t4lz: Now using a dict or a string enum in all apps.

mirrord-layer/tests/apps/app_node.js
t4lz: Changed all apps not to used a hardcoded number of methods.

eyalb181 (approved):

PR #516, opened 2022-10-08
Replaced `pcap` dependency with our own `rawsocket` to make cross com…

aviramha:
…piling faster and easier.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,4 @@
 [toolchain]
 channel = "nightly-2022-10-06"
+components = [ "rustfmt", "clippy" ]
+profile = "minimal"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2022-10-06"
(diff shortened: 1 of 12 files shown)


tests/src/sanity.rs
```diff
@@ -606,7 +606,6 @@ mod tests {
         }
     }
 
-    #[ignore] // TODO: create integration test instead.
```
eyalb181, line 609: Why'd you remove it?

rust-toolchain.toml
```diff
@@ -1,2 +1,4 @@
 [toolchain]
-channel = "nightly"
\ No newline at end of file
+channel = "nightly-2022-10-06"
```
eyalb181, line 2: This doesn't seem intentional?

eyalb181 (changes requested):

rust-toolchain.toml
aviramha: it is clippy is broken in newer versions and that's how I pin it globally..

tests/src/sanity.rs
aviramha: want to see that it works since we don't have much coverage here, will re-add once it passes.

eyalb181 (approved):

PR #535, opened 2022-10-10
Disable Go http mirroring test.

t4lz:
The test is unstable e.g. https://github.com/metalbear-co/mirrord/actions/runs/3220168873/jobs/5266839185.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -187,12 +187,13 @@ jobs:
         with:
           key: ubuntu-latest
           cache-on-failure: true
-      - uses: actions/setup-go@v3             # Install Go for http mirroring tests with a Go webserver.
-        with:
-          go-version: "1.19.0"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -187,12 +187,6 @@ jobs:
         with:
           key: ubuntu-latest
           cache-on-failure: true
-      - uses: actions/setup-go@v3             # Install Go for http mirroring tests with a Go webserver.
```
meowjesty, line 190: We're gonna re-add these tests later when they're fixed right? So isn't it better to just comment these lines instead of removing them (and add a TODO)?

meowjesty (approved):
👍

.github/workflows/ci.yaml
t4lz: Changed deletions to comments.

PR #537, opened 2022-10-10
[Merged by Bors] - Progress reporting

tamasfe:
Successor to #421 with the suggestions applied and a slightly different approach used.

This approach uses [`termspin`](https://crates.io/crates/termspin) for hierarchical spinners and does not rely on `tracing` spans but rather explicit progress reporting in code. Additionally it sets no colors, so it should work on Windows without additional rituals.

The environment variable and modes are the same, JSON reporting is still far off.

Integration with `tracing` itself is still possible as a layer, however I would rather not do it due to the following:
- performance implications, the progress reporting has a limited scope within the app and a tracing layer would incur unnecessary overhead even after no reporting is done anymore (e.g. the previous implementation inspected every span that was created)
- it did not play well with async, it was not possible to know when a task was actually done
- messages now can be dynamic if needed, and failures can be trivially reported

<details>
<summary>demo</summary>

https://user-images.githubusercontent.com/25967296/194955184-5a662cc9-e4a6-4c28-a562-956d48156f14.mp4

</details>

I think this is ready for a review/merge sans the changelog.


eyalb181 (approved):

eyalb181:
Lovely, thank you @tamasfe !

Issue #380, opened 2022-09-09
Go crash on signal - linux


aviramha:
### Bug Description

in the CI, go crashes sometimes:

### Steps to Reproduce

Seems to happen on http mirror traffic test in e2e.

### Backtrace

```shell
stderr 2022-09-09T02:16:10.946375451Z 21601: fatal error: unexpected signal during runtime execution

stderr 2022-09-09T02:16:10.948592196Z 21601: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x41d127]

goroutine 20 [running]:

stderr 2022-09-09T02:16:10.950880842Z 21601: runtime.throw({0x80498c?, 0xc000042730?})
	/opt/hostedtoolcache/go/1.19.0/x64/src/runtime/panic.go:1047 +0x5d fp=0xc000042700 sp=0xc0000426d0 pc=0x436fdd
runtime.sigpanic()
	/opt/hostedtoolcache/go/1.19.0/x64/src/runtime/signal_unix.go:819 +0x369 fp=0xc000042750 sp=0xc000042700 pc=0x44ca09
runtime: g 20: unexpected return pc for runtime.gcBgMarkWorker called from 0x5
stack: frame={sp:0xc000042750, fp:0xc0000427e0} stack=[0xc000042000,0xc000042800)
0x000000c000042650:  0x0000000000000000  0x0000000000000000 
0x000000c000042660:  0x0000000000000000  0x0000000000000000 
0x000000c000042670:  0x0000000000000000  0x0000000000000000 
0x000000c000042680:  0x00000000004650ce <runtime.systemstack+0x000000000000002e>  0x000000000043732c <runtime.fatalthrow+0x000000000000006c> 
0x000000c000042690:  0x000000c0000426a0  0x000000c000082b60 
0x000000c0000426a0:  0x0000000000437360 <runtime.fatalthrow.func1+0x0000000000000000>  0x000000c000082b60 
0x000000c0000426b0:  0x0000000000436fdd <runtime.throw+0x000000000000005d>  0x0
stderr 2022-09-09T02:16:10.950896743Z 21601: 00000c0000426d0 
0x000000c0000426c0:  0x000000c0000426f0  0x0000000000436fdd <runtime.throw+0x000000000000005d> 
0x000000c0000426d0:  0x000000c0000426d8  0x0000000000437000 <runtime.throw.func1+0x0000000000000000> 
[...]

PR #539, opened 2022-10-11
Make Go test stable.

t4lz:
Wait for all unique methods to be done once, using a wait group. Synchronize done map with a mutex so that the same http method does not report it's done twice, if e.g. two GETs are sent.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -160,7 +160,7 @@ jobs:
           profile: minimal
           toolchain: nightly
           components: rustfmt
-      - uses: Swatinem/rust-cache@v1
+      - uses: Swatinem/rust-cache@v2
         with:
           key: ubuntu-latest
       - run: cargo +nightly build --manifest-path=./Cargo.toml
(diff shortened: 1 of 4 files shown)


t4lz:
Seems like the Go test is might be finding an actual bug:
https://github.com/metalbear-co/mirrord/actions/runs/3225436965/jobs/5277792764#step:18:103
https://github.com/metalbear-co/mirrord/actions/runs/3225569148/jobs/5278144204#step:38:106
https://github.com/metalbear-co/mirrord/actions/runs/3225672404/jobs/5278304668#step:15:106
https://github.com/metalbear-co/mirrord/actions/runs/3225796801/jobs/5278582331#step:17:106

Could be related to https://github.com/metalbear-co/mirrord/issues/380.

How should we proceed? Here are some options I can think of:
1. Disable the Go test for Linux (I think it only fails there) until we can find the cause of this crash.
2. Disable [the assertion](https://github.com/metalbear-co/mirrord/pull/539/files#diff-3eba238d33bc589a6caf765666d55a9a3d1ced48e4bf2857564c33ef69ab44eeR298) that the word "error" does not appear in the stderr on Go on Linux. - This would reduce the frequency of failures, but the test would still sometimes fail, as the crash [sometimes](https://github.com/metalbear-co/mirrord/actions/runs/3225436965/jobs/5277792764#step:18:431) happens before all requests are handles so the test fails on asserting all the methods were handled, not on no errors in stderr.
3. Try to change the test app to avoid this crash.

t4lz:
Http mirror tests tested in GitHub Actions 100 times on MacOS and 100 times on Linux with no failures: https://github.com/metalbear-co/mirrord/actions/runs/3226531482/jobs/5280161554

meowjesty (approved):
👍

PR #540, opened 2022-10-11
Use shared-key for rust cache on jobs that take long to compile.

t4lz:


t4lz:
I think doing that would lead to the same cache to be used also when the rust version in the environment is different, and since we use `nightly`, the rust version can change between runs without us doing anything. So I think we should just stick to the default key.

Issue #538, opened 2022-10-10
Config file JSON schema


tamasfe:
A JSON schema for `mirrord.toml|yaml` that is submitted to the [schema store](https://schemastore.org) would enable linting/completion in IDEs and other tools.

It could be either hand-written, or preferably generated with something like [schemars](https://crates.io/crates/schemars).

Open tasks:

- [x] documentation of fields
- [x] distribution either via the schema store (or probably the vscode extension)

edit: fields are documented.
edit2: merged to the schema store.

aviramha:
Great idea! we'll prioritize it.

aviramha:
Closed by #543. Thanks @tamasfe !

tamasfe:
The schema exists, but it still needs
- documentation of fields
- distribution either via the schema store (or probably the vscode extension)

aviramha:
@tamasfe reopening - let's use this issue for tracking the required steps that needs to be done.

meowjesty:
Fields documented by #661 

Just need to settle for a file name for the user's `mirrord-config.[json|toml|yaml]`, and then we can close this.

tamasfe:
I'd suggest picking a name that is as short as possible. Right now I use configs such as as `foo.mirrord.toml`, so something like `.*\.mirrord\.(json|toml|(ya?ml))$` could work imo.

I'd also suggest distributing the schema via the vscode extension as well (orthogonal to the schema store).

meowjesty:
Schema is now distributed in the vscode-extension by #714.

PR #543, opened 2022-10-11
feat(config): JSON schema

tamasfe:
I took the liberty of implementing #538 as it took a few minutes only.

I added a test that is ignored by default that is used to generate the schema that can be submitted to schema store afterwards. I usually add the schema generation to my CLIs as well as a command but I don't ever recall actually using them.

Descriptions are generated from doc comments, so they are nonexistent, but completion and validation works:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1673,6 +1673,7 @@ dependencies = [
  "mirrord-config-derive",
  "mirrord-macro",
  "rstest",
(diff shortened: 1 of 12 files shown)


aviramha (commented):
This looks awesome!  Can we incorporate it into the CI?

aviramha (approved):

PR #545, opened 2022-10-11
[Merged by Bors] - Override environment variables from the config

tamasfe:
This PR allows overriding environment variables inside config files.

```toml
[feature.env.override]
FOO = "bar"
```

The above is equivalent to setting `FOO` manually before running mirrord and then specifying `--exclude "FOO"`. The pain points of this approach become apparent when there are many environment variables and one has to keep something like a `.env` file in sync with the `exclude`s passed to mirrord.


aviramha:
Thanks @tamasfe ! 
I wonder if that's not creating additional complexity, maybe the feature should be that mirrord would support loading `.env` file (and those would be the override env?)

tamasfe:
> I wonder if that's not creating additional complexity, maybe the feature should be that mirrord would support loading `.env` file (and those would be the override env?)

The added complexity is currently 14 lines. I'd ideally converge towards single config files, so one could eventually simply just do `mirrord exec -f dev_foo.mirrord.toml` and `mirrord exec -f dev_bar.mirrord.toml`.

I think `.env` support is orthogonal to this feature.

The issue that has just became apparent to me regarding overriding environment is that this currently allows overriding any variable including the ones created by mirrord. It's the last step of init so I'm not sure if it's an issue.

aviramha:
@tamasfe I didn't mean code complexity, rather user/dev experience. 
@eyalb181 wdyt?

tamasfe:
> I didn't mean code complexity, rather user/dev experience.

I can only speak for myself, currently this feature helps me to configure various scenarios during development.

I could also achieve the same thing via `.env` support though, but I feel having all the env configuration (`include/exclude/override`) at one place helps with transparency/maintainability.

eyalb181:
The feature makes sense to me, some notes though:
- It's currently only supported for config files, but should be supported as an env var as well (so we could use it in the IDE extensions)
- There's redundancy between the override and exclude lists. I think a cleaner solution would be to let the exclude list members take a value, e.g. "VAR1; VAR2=FOO; VAR3". Exclude/include could be renamed to local/remote for this to make sense semantically.
What do you think?

tamasfe:
> so we could use it in the IDE extensions

Is there a reason why the config isn't simply passed as a file or even single env var in JSON format? I don't see how env vars are manageable for many potentially complex options.

I guess the `exclude` could be renamed to `set` to combine this feature with the existing one.

E.g.:

```toml
[feature.env.set]
# exclude
FOO = false
# override
BAR = "value"
```

On the CLI we could use something like `--set-env "-FOO" --set-env BAR=value`

Even the current `include` could be merged into this by omitting the value.

eyalb181:
I withdraw my previous comment - it's probably clearer having them in two separate lists than one hybrid list. Re IDE extensions, it's quite likely we'll move to file as the default configuration source as you said, and until then they support exclude + .env (or e.g. launch.json). 
Merging this, thank you @tamasfe!

PR #536, opened 2022-10-10
[Merged by Bors] - Refactor layer integration tests, moving common code to module.

t4lz:

PR #548, opened 2022-10-11
[Merged by Bors] - Print test process output while running in integration tests.

t4lz:
That way, if the test times out, you can still read the mirrord logs.


t4lz:
Branches from https://github.com/metalbear-co/mirrord/pull/536

aviramha (approved):

Issue #531, opened 2022-10-10
exec panic auth error aws


mentos1386:
### Bug Description

Running `mirrord exec` with kubeconfig set to execute aws command to get token, fails with panic.

### Steps to Reproduce

1. Create EKS cluster
2. Run `aws eks update-kubeconfig --name test-cluster --alias test-cluster`
3. Run `mirrord exec ...`

### Backtrace

```shell
thread '<unnamed>' panicked at 'failed to create agent in k8s: mirrord-layer: Kube failed with error `auth error: auth exec command '"aws" "--region" "us-east-1" "eks" "get-token" "--cluster-name" "development-cluster"' failed with status signal: 6 (SIGABRT) (core dumped): Output { status: ExitStatus(unix_wait_status(134)), stdout: "", stderr: "thread '<unnamed>' panicked at 'failed to create agent in k8s: mirrord-layer: Kube failed with error `auth error: auth exec command '\"aws\" \"--region\" \"us-east-1\" \"eks\" \"get-token\" \"--cluster-name\" \"development-cluster\"' failed with status signal: 6 (SIGABRT) (core dumped): Output { status: ExitStatus(unix_wait_status(134)), stdout: \"\", stderr: \"thread '<unnamed>' panicked at 'failed to create agent in k8s: mirrord-layer: Kube failed with error `auth error: auth exec command '\\\"aws\\\" \\\"--region\\\" \\\"us-east-1\\\" \\\"eks\\\" \\\"get-token\\\" \\\"--cluster-name\\\" \\\"development-cluster\\\"' failed with status signal: 11 (SIGSEGV) (core dumped): Output { status: ExitStatus(unix_wait_status(139)), stdout: \\\"\\\", stderr: \\\"\\\\r\\\\nmalloc: unknown:0: assertion botched\\\\r\\\\nmalloc: block on free list clobbered\\\\nAborting...\\\" }`!', mirrord-layer/src/connection.rs:92:18\\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\\nfatal runtime error: failed to initiate panic, error 5\\n\" }`!', mirrord-layer/src/connection.rs:92:18\nnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\nfatal runtime error: failed to initiate panic, error 5\n" }`!', mirrord-layer/src/connection.rs:92:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[...]

PR #552, opened 2022-10-12
[Merged by Bors] - Skip loading mirrord env var

aviramha:
### Changed
- CI: mirrord-layer UT and integration run in same job.
- .devcontainer: Added missing dependencies and also kind for running e2e tests.

### Fixed
- Added another protection [to not execute in child processes from k8s auth](https://github.com/metalbear-co/mirrord/issues/531) by setting an env flag to avoid loading then removing it after executing the api.


eyalb181 (approved):

aviramha:
bors retry

PR #559, opened 2022-10-13
Use KubeConfig namespace for target instead of "default"

### LINKED ISSUE #557: Use kubectl config namespace by default
Right now we use "default" namespace instead of the kubectl config namespace.

infiniteregrets:
closes #557

### DIFF
--- a/mirrord-layer/src/pod_api.rs
+++ b/mirrord-layer/src/pod_api.rs
@@ -64,13 +64,19 @@ pub(crate) async fn create_agent(
         ..
     } = config.clone();
 
-    let client = if accept_invalid_certificates {
-        let mut config = Config::infer().await?;
-        config.accept_invalid_certs = true;
+    let (target_namespace, client) = if accept_invalid_certificates {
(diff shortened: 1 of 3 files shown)


mirrord-layer/src/pod_api.rs
```diff
@@ -64,13 +64,19 @@ pub(crate) async fn create_agent(
...
         warn!("Accepting invalid certificates");
-        Client::try_from(config).map_err(LayerError::KubeError)?
+        (
+            target_namespace.unwrap_or(kube_config.default_namespace.clone()),
+            Client::try_from(kube_config).map_err(LayerError::KubeError)?,
+        )
     } else {
-        Client::try_default().await.map_err(LayerError::KubeError)?
+        (
+            "default".to_string(),
```
aviramha, line 77: why in this case it's default? it should be the same here..

aviramha (changes requested):
Great job - I think I found an implementation issue though.

aviramha:
hey, I'm probably addressing this as part of my refactor PR. Closing this.

PR #565, opened 2022-10-13
[Merged by Bors] - refactor pod api

### LINKED ISSUE #556: Skip service mesh proxy containers
When mirrord runs on a pod in a cluster with a service mesh, that pod would likely have a container for the service mesh proxy. In this case, mirrord should make sure not to impersonate this container but rather the next one after it.

### LINKED ISSUE #557: Use kubectl config namespace by default

aviramha:
### Added
- Skip istio/linkerd-proxy/init container when mirroring a pod without a specific container name. Closes #556 
- Add "linkerd.io/inject": "disabled" annotation to pod created by mirrord to avoid linkerd auto inject.
- mirrord-layer: support `-target deployment/deployment_name/container/container_name` flag to run on a specific container. 

## Changed
- mirrord-layer: refactored `pod_api.rs` to be more maintainble.
- Use kube config namespace by default. Closes #557

### DIFF
--- a/mirrord-config/src/pod.rs
+++ b/mirrord-config/src/pod.rs
@@ -12,7 +12,7 @@ pub struct PodFileConfig {
     #[config(env = "MIRRORD_AGENT_IMPERSONATED_POD_NAME")]
     pub name: Option<String>,
 
-    #[config(env = "MIRRORD_AGENT_IMPERSONATED_POD_NAMESPACE", default = "default")]
+    #[config(env = "MIRRORD_AGENT_IMPERSONATED_POD_NAMESPACE")]
     pub namespace: Option<String>,
 
     #[config(env = "MIRRORD_IMPERSONATED_CONTAINER_NAME")]
@@ -29,9 +29,9 @@ mod tests {
     #[rstest]
     fn default(
         #[values((Some("pod"), Some("pod".to_string())))] name: (Option<&str>, Option<String>),
-        #[values((None, "default"), (Some("namespace"), "namespace"))] namespace: (
+        #[values((None, None), (Some("namespace"), Some("namespace".to_string())))] namespace: (
             Option<&str>,
-            &str,
+            Option<String>,
         ),
         #[values((None, None), (Some("container"), Some("container")))] container: (
             Option<&str>,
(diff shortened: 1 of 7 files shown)


mirrord-config/src/pod.rs
```diff
@@ -29,9 +29,9 @@ mod tests {
     #[rstest]
     fn default(
         #[values((Some("pod"), Some("pod".to_string())))] name: (Option<&str>, Option<String>),
-        #[values((None, "default"), (Some("namespace"), "namespace"))] namespace: (
+        #[values((None, None), (Some("namespace"), Some("namespace".to_string())))] namespace: (
             Option<&str>,
-            &str,
+            Option<String>,
```
DmitryDodzin, line 34: Nitpick but you can use `Option<&str>`, remove the `.to_string()` and change the assertion to `assert_eq!(outgoing.namespace.as_deref(), namespace.1);`

aviramha: lord of nitpicks <3

mirrord-layer/src/pod_api.rs
```diff
@@ -48,106 +48,371 @@ impl Drop for EnvVarGuard {
...
+        Ok(pod_name)
+    }
+
+    pub async fn create_agent(&self, connection_port: u16) -> Result<Portforwarder> {
+        let progress = TaskProgress::new("agent initializing...");
+        let agent_image = agent_image(&self.config);
+        let runtime_data = self.get_runtime_data().await?;
+        let connect_pod_name = if self.config.agent.ephemeral {
+            self.create_ephemeral_container_agent(
+                runtime_data,
```
eyalb181, line 390: Nitpick, but please have the same parameter order for create_ephemeral_container_agent and create_job_pod_agent

DmitryDodzin (approved):
Just small nitpicks

mirrord-layer/src/pod_api.rs
aviramha: fixed

mirrord-config/src/pod.rs
aviramha: leaving it as it's just a test..

PR #566, opened 2022-10-13
[Merged by Bors] - Reuse agent

aviramha:
### Added
- Reuse agent - first process that runs will create the agent and its children will be able to reuse the same one to avoid creating many agents.


mirrord-layer/src/connection.rs
```diff
@@ -71,27 +71,60 @@ where
...
         let k8s_api = KubernetesAPI::new(config).await.unwrap();
-        let mut port_forwarder = k8s_api.create_agent(connection_port).await.unwrap_or_else(|err| match err {
-            LayerError::KubeError(kube::Error::HyperError(err)) => {
-                eprintln!("\nmirrord encountered an error accessing the Kubernetes API. Consider passing --accept-invalid-certificates.\n");
+        let (pod_agent_name, agent_port) = {
+            if let (Some(pod_agent_name), Some(agent_port)) =
+                (&config.connect_agent_name, config.connect_agent_port)
+            {
+                info!(
+                    "Re using existing agent {:?}, port {:?}",
```
t4lz, line 101:
```suggestion
                    "Reusing existing agent {:?}, port {:?}",
```

mirrord-layer/src/connection.rs
```diff
@@ -71,27 +71,60 @@ where
...
+                );
+                (pod_agent_name.to_owned(), agent_port)
+            } else {
+                info!("No existing agent, spawning new one.");
+                let agent_port: u16 = rand::thread_rng().gen_range(30000..=65535);
+                info!("Using port `{agent_port:?}` for communication");
+                let pod_agent_name = match k8s_api.create_agent(agent_port).await {
+                    Ok(pod_name) => pod_name,
+                    Err(err) => error_handle(err),
+                };
```
t4lz, line 112: We're going take that chance with taken ports for now, right?

mirrord-layer/src/connection.rs
```diff
@@ -71,27 +71,60 @@ where
     }
 }
 
+fn error_handle(err: LayerError) -> ! {
```
t4lz, line 74: Nit: maybe `handle_error`?

mirrord-layer/src/connection.rs
```diff
@@ -71,27 +71,60 @@ where
...
             }
-            _ => panic!("failed to create agent in k8s: {}", err),
-        });
-        AgentConnection::Portforwarder(port_forwarder.take_stream(connection_port).unwrap())
+        };
+
+        let mut port_forwarder = match k8s_api.port_forward(&pod_agent_name, agent_port).await {
+            Ok(port_forwarder) => port_forwarder,
+            Err(err) => error_handle(err),
+        };
```
t4lz, line 126: Does this compile:
```suggestion
        let mut port_forwarder = match k8s_api.port_forward(&pod_agent_name, agent_port).await.unwrap_or_else(error_handle);
```

t4lz (approved):
Cool stuff. Does the agent not assume in any point it is managing a single client?

mirrord-config/src/lib.rs
```diff
@@ -44,6 +44,14 @@ pub struct LayerFileConfig {
     #[config(env = "MIRRORD_CONNECT_TCP")]
     pub connect_tcp: Option<String>,
 
+    /// Agent name that already exists that we can connect to.
+    #[config(env = "MIRRORD_CONNECT_AGENT")]
```
eyalb181, line 48: Why is this a configuration value? When would we want users to be able to set it?

mirrord-layer/src/connection.rs
aviramha: always been, but anyway the agent listens on it's own namespace, so it can't really conflict. not the case with ephemeral IIRC though..

aviramha:
> Cool stuff. Does the agent not assume in any point it is managing a single client?

It shouldn't as the whole code is based around multiple clients (which was pretty annoying, but now it's paying it off..)

mirrord-layer/src/connection.rs
```diff
@@ -71,27 +71,60 @@ where
...
 
-                match err.into_cause() {
-                    Some(cause) => panic!("{}", cause),
-                    None => panic!("mirrord got KubeError::HyperError"),
-                }
+                // Set env var for children to re-use.
+                std::env::set_var("MIRRORD_CONNECT_AGENT", &pod_agent_name);
+                std::env::set_var("MIRRORD_CONNECT_PORT", agent_port.to_string());
+                // So children won't show progress as well as it might confuse users
+                std::env::set_var(mirrord_progress::MIRRORD_PROGRESS_ENV, "off");
```
eyalb181, line 118: Would they see the progress as completed even though it's still spawning child processes and connecting them to the agent?

mirrord-config/src/lib.rs
aviramha: It might prove useful for testing and maybe later scenarios. It's not for users though... it's just that the config is the relevant place (like connect_tcp)?

mirrord-layer/src/connection.rs
aviramha: yeah but that doesn't mean that mirrord is still loading - the action of connecting is sub-second in most scenarios and if for example you are running npm or any other cli tool it might run 7 processes which will spam your console.

eyalb181 (approved):

PR #573, opened 2022-10-14
[Merged by Bors] - Fixed Issue #531

aviramha:
### Fixed
Issue [#531](https://github.com/metalbear-co/mirrord/issues/531). We now detect NixOS/Devbox usage and add `sh` to skipped list.

### DIFF
--- a/mirrord-layer/src/lib.rs
+++ b/mirrord-layer/src/lib.rs
@@ -97,7 +97,7 @@ fn nix_devbox_patch(config: &mut LayerConfig) {
         .map(VecOrSingle::to_vec)
         .unwrap_or_default();
 
-    if !current_skip.contains(&"sh".to_string()) && is_nix_or_devbox() {
+    if is_nix_or_devbox() && !current_skip.contains(&"sh".to_string()) {
         current_skip.push("sh".into());
         std::env::set_var("MIRRORD_SKIP_PROCESSES", current_skip.join(";"));
         config.skip_processes = Some(VecOrSingle::Multiple(current_skip));


mirrord-layer/src/lib.rs
```diff
@@ -88,12 +88,43 @@ pub(crate) static ENABLED_UDP_OUTGOING: OnceLock<bool> = OnceLock::new();
...
+
+    if is_nix_or_devbox() {
+        current_skip.push("sh".into());
+        std::env::set_var("MIRRORD_SKIP_PROCESSES", current_skip.join(";"));
+        config.skip_processes = Some(VecOrSingle::Multiple(current_skip));
+    }
+}
+/// Check if NixOS or Devbox by discrimnating env vars.
+fn is_nix_or_devbox() -> bool {
+    if let Ok(res) = std::env::var("IN_NIX_SHELL") && res.as_str() == "1" {
```
meowjesty, line 112: nit: can't this be `if / else if / else` to drop the `return` keyword?

aviramha: ok

mirrord-layer/src/lib.rs
```diff
@@ -88,12 +88,43 @@ pub(crate) static ENABLED_UDP_OUTGOING: OnceLock<bool> = OnceLock::new();
...
+        .skip_processes
+        .clone()
+        .map(VecOrSingle::to_vec)
+        .unwrap_or_default();
+    // already contains, doesn't matter
+    if current_skip.contains(&"sh".to_string()) {
+        return;
+    }
+
+    if is_nix_or_devbox() {
```
meowjesty, line 104: nit-2: to drop the `return`.

```suggestion
    if is_nix_or_devbox() && !current_skip.contains(...) {
```

meowjesty (approved):
Only nits, 👍

PR #580, opened 2022-10-17
Initial work on resizing deployments with replicas > 1

infiniteregrets:
This is just an initial draft:
I want to discuss:
- what are the scenarios where on Dropping/Exiting mirrord should resize the replicas back to its original number, for example when when the connection is dropped, or when the library constructor exits
- how do we go about doing the above when a simple call to exit() happens, do we hook exit()?
- do we install signal handlers for signals like SIGINT
- what kind of patch strategy are we looking to follow: https://docs.rs/kube/latest/kube/api/enum.Patch.html

### DIFF
--- a/mirrord-cli/src/main.rs
+++ b/mirrord-cli/src/main.rs
@@ -100,7 +100,7 @@ fn add_to_preload(path: &str) -> Result<()> {
 
 #[cfg(target_os = "macos")]
 fn sip_check(binary_path: &str) -> Result<()> {
-    let sip_set = RegexSet::new(&[
+    let sip_set = RegexSet::new([
         r"/System/.*",
         r"/bin/.*",
         r"/sbin/.*",
--- a/mirrord-layer/src/connection.rs
+++ b/mirrord-layer/src/connection.rs
@@ -1,5 +1,6 @@
 use std::{
     pin::Pin,
+    sync::OnceLock,
     task::{Context, Poll},
 };
 
@@ -10,6 +11,8 @@ use tokio::net::TcpStream;
 use tracing::log::info;
 
 use crate::{error::LayerError, pod_api::KubernetesAPI};
+
+pub(crate) static K8S_API: OnceLock<KubernetesAPI> = OnceLock::new();
 
 pub(crate) enum AgentConnection<T>
 where
@@ -93,6 +96,10 @@ pub(crate) async fn connect(config: &LayerConfig) -> impl AsyncWrite + AsyncRead
         AgentConnection::TcpStream(stream)
     } else {
         let k8s_api = KubernetesAPI::new(config).await.unwrap();
+        K8S_API
+            .set(k8s_api)
+            .expect("Failed to set K8S_API global variable");
+
         let (pod_agent_name, agent_port) = {
             if let (Some(pod_agent_name), Some(agent_port)) =
                 (&config.connect_agent_name, config.connect_agent_port)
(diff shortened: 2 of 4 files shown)


aviramha:
* When connection is dropped, which should happen on library exit.
* I think we have a graceful shutdown, but might be wrong about it that we need to hook exit to accomplish it (maybe as a prior PR)
* Can you write down the differences? not sure what are the effects.
* I know it's a draft, so just making sure we're aligned that this feature is a flag and not a default.

infiniteregrets:
> * When connection is dropped, which should happen on library exit.
> * I think we have a graceful shutdown, but might be wrong about it that we need to hook exit to accomplish it (maybe as a prior PR)
> * Can you write down the differences? not sure what are the effects.
> * I know it's a draft, so just making sure we're aligned that this feature is a flag and not a default.

> > * I think we have a graceful shutdown, but might be wrong about it that we need to hook exit to accomplish it (maybe as a prior PR)

i thinking hooking exit should be right

> > * Can you write down the differences? not sure what are the effects.

what happens when a user sends an interrupt signal to the process, in that case there is no graceful shutdown
i thinking we should have a mask of signals we can handle and install a handler for them which will result in a graceful exit along with downsizing the replica set


> > * I know it's a draft, so just making sure we're aligned that this feature is a flag and not a default.

yes we are on the same page, this will be a flag

aviramha:
> > * When connection is dropped, which should happen on library exit.
> > * I think we have a graceful shutdown, but might be wrong about it that we need to hook exit to accomplish it (maybe as a prior PR)
> > * Can you write down the differences? not sure what are the effects.
> > * I know it's a draft, so just making sure we're aligned that this feature is a flag and not a default.
> 
> > > * I think we have a graceful shutdown, but might be wrong about it that we need to hook exit to accomplish it (maybe as a prior PR)
> 
> i thinking hooking exit should be right
> 
> > > * Can you write down the differences? not sure what are the effects.
> 
> what happens when a user sends an interrupt signal to the process, in that case there is no graceful shutdown i thinking we should have a mask of signals we can handle and install a handler for them which will result in a graceful exit along with downsizing the replica set
> 
> > > * I know it's a draft, so just making sure we're aligned that this feature is a flag and not a default.
> 
> yes we are on the same page, this will be a flag

interrupt should still call exit..

infiniteregrets:
> > what happens when a user sends an interrupt signal to the process, in that case there is no graceful shutdown i thinking we should have a mask of signals we can handle and install a handler for them which will result in a graceful exit along with downsizing the replica set
> 
> interrupt should still call exit..

sorry, can you elaborate? i was saying that the handler will do a graceful exit.. could be through a common function for exiting

aviramha:
> sorry, can you elaborate? i was saying that the handler will do a graceful exit.. could be through a common function for exiting
AFAIK if a signal handler causes the process to exit, it will call exit hence if we hook exit we'll do it gracefully.

infiniteregrets:
> what kind of patch strategy are we looking to follow: https://docs.rs/kube/latest/kube/api/enum.Patch.html

- do we lose the containers from the pods once we patch? does that not affect any files that existed on a particular container? if we use `Merge` we can merge containers but what happens when we scale back?

aviramha:
> > what kind of patch strategy are we looking to follow: https://docs.rs/kube/latest/kube/api/enum.Patch.html
> 
>     * do we lose the containers from the pods once we patch? does that not affect any files that existed on a particular container? if we use `Merge` we can merge containers but what happens when we scale back?

The strategy doesn't affect what happens in the containers themselves, it's merely how the patch is applied "yaml" wise. What happens in case of conflict, who applies, etc.

PR #589, opened 2022-10-18
Update intelliJ extension

infiniteregrets:


intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordListener.kt
```diff
@@ -66,33 +67,49 @@ class MirrordListener : ExecutionListener {
                         return@invokeLater super.processStartScheduled(executorId, env)
                     }
 
-                    val fileOpsCheckbox = JCheckBox("Enable File Operations")
-                    val remoteDnsCheckbox = JCheckBox("Enable Remote DNS")
-                    val ephemeralContainerCheckBox = JCheckBox("Enable Ephemeral Containers")
+                    val fileOps = JCheckBox("File Operations (default: true)")
```
eyalb181, line 70: Wait is (default: true/false) necessary here?

infiniteregrets: since we dont have toggles, i thought we should make it explicit. we can remove it, lmk

eyalb181: Lets remove it, the default doesn't matter if the checkbox reflects the current value

eyalb181 (approved):

Issue #433, opened 2022-09-18
`-x` has no "*" option


tamasfe:
### Bug Description

`-s '*'` can be used to include all environment variables from a pod but `-x '*'` is ignored. `-s ''` can be used for the same effect as a workaround.

### Steps to Reproduce

1. run `mirrord exec -x '*' ...`
2. assert that env variables from the pod are there

### Your operating system and version

Linux

### Local process

does not matter

eyalb181:
Don't you get the same behavior from omitting these two arguments entirely?

tamasfe:
My observations without looking at the code:

- without flags all env vars are enabled from the pod
- `-s ''`: no env vars from pod
- `-x "FOO"`: env vars from pod without `FOO`

aviramha:
> Don't you get the same behavior from omitting these two arguments entirely?

I assume the desire here is to not have remote environment. The default is to provide same env

eyalb181:
Yup, understood, my mistake.

PR #592, opened 2022-10-18
[Merged by Bors] - Override Env Wildcard Matching

DmitryDodzin:
Better support for wildcard matching to override env variables

Eg. `PREFIX_*` `TEST_*_SUFFIX` `TEST_??_FLAG_*`

closing #433


mirrord-agent/src/env.rs
```diff
@@ -0,0 +1,150 @@
...
+
+struct EnvFilter {
+    include: Vec<WildMatch>,
+    exclude: Vec<WildMatch>,
+}
+
+impl EnvFilter {
+    fn new(select_env_vars: HashSet<String>, filter_env_vars: HashSet<String>) -> Self {
+        let include = if select_env_vars.is_empty() {
+            vec![WildMatch::new("*")]
```
eyalb181, line 19: I think default values should be managed in config (or at least in the layer) rather than here

mirrord-agent/src/env.rs
```diff
@@ -0,0 +1,150 @@
...
+            select_env_vars
+                .iter()
+                .map(|selector| WildMatch::new(selector))
+                .collect()
+        };
+
+        let exclude = {
+            let mut exclude = Vec::new();
+
+            if !select_env_vars.contains("PATH") {
```
eyalb181, line 30: Lets drop this code, for now. I doubt we'll ever have anyone actually use it, and if we do, I'd like to know

eyalb181 (changes requested):

mirrord-agent/src/env.rs
DmitryDodzin: This isn't quite a default value, it's more like the check that was there previously that will pass the variable if `select_env_vars.is_empty()`

DmitryDodzin: If you pass to cli `-x FOOBAR` the filter value in mirrord-layer the values should be `include = None` and `exclude = Some("FOOBAR")` but here I need the `include` to be `*` to allow the exclude (`"FOOBAR" | "HOME" | "PATH"`) filter to work and get all the other env vars

eyalb181 (approved):

PR #594, opened 2022-10-18
[Merged by Bors] - Adds `FileFilter` to improve how mirrord deals with file paths. Closes #530

### LINKED ISSUE #530: Add include/exclude list functionality for fileop paths
We should let users specify which paths to include or exclude from remote file ops.

meowjesty:
- Allows the user to specify paths to exclude/include for file-ops;
- Adds `FileFilter` enum as a more concise way of checking if a file path should be bypassed;
- Adds `is_success` and `is_bypass` helper functions to `Detour` (currently only being used in tests);
- Changes how file ops config work, now it supports `Simple` and `Advanced` modes, where simple is just what we had, and `Advanced` supports include/exclude filters;


mirrord-config/src/lib.rs
```diff
@@ -260,6 +261,7 @@ mod tests {
             feature: FeatureFileConfig {
                 env: ToggleableConfig::Enabled(true),
                 fs: ToggleableConfig::Config(FsConfig::Write),
+                file_select: ToggleableConfig::Enabled(true),
```
eyalb181, line 264: This should be nested under fs

mirrord-layer/src/file/filter.rs
```diff
@@ -0,0 +1,304 @@
...
+            .expect("Failed parsing exclude file regex!")
+            .unwrap_or(default_exclude_regex);
+
+        reduce_to_string(include)
+            .as_deref()
+            .map(Regex::new)
+            .transpose()
+            .expect("Failed parsing include file regex!")
+            .map(Self::Include)
+            .unwrap_or(Self::Exclude(exclude))
```
eyalb181, line 121: When would unwrap fail here?

eyalb181: And/or in the command before it

meowjesty: The above part: if the user specifies no `exclude`, then we `unwrap` as the `default_exclude_regex`.

Below: when the user specifies no `include`, we `unwrap` as the `exclude` (which is guaranteed to exist either as a user specified, or a default value).

mirrord-config/src/fs/advanced.rs
```diff
@@ -0,0 +1,44 @@
...
+        Ok(Self::Generated {
+            mode,
+            include,
+            exclude,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {}
```
DmitryDodzin, line 44: "Do or do not, there is no try"

mirrord-config/src/fs.rs
```diff
@@ -1,109 +1,73 @@
...
-    fn default(#[case] fs: Option<&str>, #[case] ro: Option<&str>, #[case] expect: FsConfig) {
-        with_env_vars(
-            vec![("MIRRORD_FILE_OPS", fs), ("MIRRORD_FILE_RO_OPS", ro)],
-            || {
-                let fs = FsConfig::default().generate_config().unwrap();
-
-                assert_eq!(fs, expect);
-            },
-        );
+impl FsConfig {
```
DmitryDodzin, line 65: I think this should be in advanced.rs

mirrord-config/src/fs.rs
```diff
@@ -1,109 +1,73 @@
...
     config::{from_env::FromEnv, source::MirrordConfigSource, ConfigError, MirrordConfig},
     util::MirrordToggleableConfig,
 };
 
+pub mod advanced;
+pub mod mode;
+
 #[derive(Deserialize, PartialEq, Eq, Clone, Debug)]
 #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
 #[serde(rename_all = "lowercase")]
```
DmitryDodzin, line 14:
```suggestion
#[serde(untagged, rename_all = "lowercase")]
```

DmitryDodzin (changes requested):
`FsUserConfig` should have `untagged` as a serde attribute other than that the config part looks fine (with a couple nitpicks)

mirrord-config/src/lib.rs
meowjesty: Nested! 🪆

meowjesty:
Good afternoon bors-san, can you merge my changes please?

bors r+

PR #595, opened 2022-10-19
Add `exit/_exit` hook

infiniteregrets:
refer: #580


mirrord-layer/src/lib.rs
```diff
@@ -560,6 +562,22 @@ pub(crate) unsafe extern "C" fn close_nocancel_detour(fd: c_int) -> c_int {
     close_detour(fd)
 }
 
+pub(crate) fn exit_logic() {
+    // unimplemented
+}
+
```
aviramha, line 568: Exit logic should be sending a message to the main loop for it to shutdown itself.

eyalb181:
Why is this a separate PR from #580 ?

PR #601, opened 2022-10-19
[Merged by Bors] - IntelliJ version check on enabling, 3 minute cooldown.

t4lz:
IntelliJ version check on enabling instead of on project start. Don't check if last check was less than 3 minutes ago.


intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordEnabler.kt
```diff
@@ -35,10 +47,58 @@ class MirrordEnabler : ToggleAction() {
...
+        if (lastCheckUTCDateTime.isAfter(nowUTC.minusHours(3))) {
+            return // Already checked in the last 3 hours. Don't check again yet.
+        }
+        val nowEpoch = nowUTC.toEpochSecond(ZoneOffset.UTC)
+        pc.setValue(LAST_CHECK_KEY, nowEpoch.toString())
+        val remoteVersion = Version.valueOf(URL(versionCheckEndpoint).readText())
+        val localVersion = Version.valueOf(version)
+        if (localVersion.lessThan(remoteVersion)) {
+            notifier(
+                "Your version of mirrord is outdated, you should update.",
```
meowjesty: Maybe soften this phrase a little bit, something along the lines of:

```suggestion
                "Your version of mirrord is outdated. Do you want to update it now?",
```

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordEnabler.kt
```diff
@@ -35,10 +47,58 @@ class MirrordEnabler : ToggleAction() {
...
+    private val version: String? = PluginManagerCore.getPlugin(pluginId)?.version
+    private val versionCheckEndpoint: String =
+        "https://version.mirrord.dev/get-latest-version?source=3&version=$version"
+    private var versionCheckDisabled
+        get() = PropertiesComponent.getInstance().getBoolean("versionCheckDisabled", false)
+        set(value) {
+            PropertiesComponent.getInstance().setValue("versionCheckDisabled", value)
+        }
+
+    private fun checkVersion(project: Project) {
```
meowjesty, line 68: Could you add some documentation to what this function does? From a quick glance it looks like it does more than just version checking?

t4lz: Added.

PR #603, opened 2022-10-20
Apt get update

IsaacCloos:
Was working on integration tests and ran into this error when trying to rebuild my development container

```
ERROR: failed to solve: executor failed running [/bin/sh -c apt-get install -y clang]: exit code: 100
```
- - -

https://github.com/metalbear-co/mirrord/blob/93af076cfd4b4652101b1751dc2ecb7b98fdcd54/.devcontainer/Dockerfile#L8

was added #552 

after some googling I found [these instructions](https://www.linux.com/news/what-do-when-apt-get-fails/#:~:text=Run%20apt%2Dget%20update%20to,it%20and%20all%20its%20dependencies.) 

# Criticism Welcome ❗ 
I could be missing something obvious 😅


aviramha (approved):
bors r+

aviramha:
Nice catch! Thank you.

PR #609, opened 2022-10-21
Add more env vars to default ignore list

aviramha:


eyalb181 (approved):

mirrord-agent/src/env.rs
```diff
@@ -25,7 +25,15 @@ impl EnvFilter {
         };
 
         let exclude = {
-            let mut exclude = vec![WildMatch::new("PATH"), WildMatch::new("HOME")];
+            let patterns = [
+                "PATH",
+                "HOME",
```
t4lz, line 30: Won't this lead to confusing situations when using the remote filesystem? Maybe this should be included or excluded according to the fs setting?

aviramha: I think the home is mostly used for local stuff anyway.. never saw a user app that uses home, it's probably internal stuff using it.

PR #612, opened 2022-10-21
[Merged by Bors] - Fix errors when setting envars when debugging Kotlin app

jamowei:
Fix double showing of `MirrordKubeDialog` and errors occurring when setting envars when debug a Kotlin app.
image


infiniteregrets (approved):
Thanks for finding this bug @jamowei!

aviramha:
Thanks!!

infiniteregrets:
Just to confirm, did you try this with any other applications besides Kotlin? If so does this happen anywhere else?

PR #614, opened 2022-10-21
[Merged by Bors] - Scrollable list for pods and namespaces for IntelliJ extension

### LINKED ISSUE #610: Add IntelliJ scrollbar for namespaces and pods list
Hi!

I'm trying to use the IntelliJ plugin and running into an issue where the namespaces and pods list is too long for the window size so I cannot select the proper value. I can type in the box so I can select it but scrollbars would be much better!

Thanks!

Screen Shot 2022-10-21 at 10 14 04 AM

infiniteregrets:
closes #610


eyalb181 (approved):

infiniteregrets:
thanks @jbeaudetupgrade for opening the issue!

PR #617, opened 2022-10-21
[Merged by Bors] - Better error message for failed to connect to agent.

meowjesty:
- Improved error message when we fail to connect to agent.


aviramha (approved):
Love it! This is called user empathy :)
bors r+

PR #625, opened 2022-10-25
[Merged by Bors] - Add changelog to intelliJ plugin

### LINKED ISSUE #542: Add changelog to intelliJ extension
Currently, we don't have a changelog in the extension that can be displayed on the website, it would be better if we could just copy the changelog on release from mirrord

infiniteregrets:
- closes #542
- lint changelog and fix `CHANGED` tag to support this
- bump java version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,29 +8,42 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ## [Unreleased]
 
 ### Added
-- Changelog for intelliJ extension
+
+- Add changelog for intelliJ extension, closes [[#542](https://github.com/metalbear-co/mirrord/issues/542)]
 
 ### Changed
-- Added timeout for "waiting for pod to be ready..." in mirrord-layer to prevent unresponsive behavior. See [#579](https://github.com/metalbear-co/mirrord/issues/579)
-- Update intelliJ extension: lint & bump java version to 17
+
+- Added timeout for "waiting for pod to be ready..." in mirrord-layer to prevent unresponsive behavior.
+  See [#579](https://github.com/metalbear-co/mirrord/issues/579)
+- Update intelliJ extension: lint & bump java version to 17.
+
+### Fixed
+
+- Lint Changelog and fix level of a "Changed" tag.
 
 ## 3.2.0
 
 ### Changed
+
 - Extended support for both `-s` and `-x` wildcard matching, now supports `PREFIX_*`, `*_SUFFIX`, ect.
-- Add to env default ignore `JAVA_HOME`,`HOMEPATH`,`CLASSPATH`,`JAVA_EXE` as it's usually runtime that you don't want from remote. Possibly fixes issue discussed on Discord (used complained that they had to use absolute path and not relative).
+- Add to env default ignore `JAVA_HOME`,`HOMEPATH`,`CLASSPATH`,`JAVA_EXE` as it's usually runtime that you don't want
+  from remote. Possibly fixes issue discussed on Discord (used complained that they had to use absolute path and not
+  relative).
 - Add `jvm.cfg` to default bypass for files.
 - Clarify wrong target error message.
 - mirrord-layer: Improve error message in `connection::handle_error`.
 
 ### Fixed
+
 - Don't ignore passed `--pod-namespace` argument, closes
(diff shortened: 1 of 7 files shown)


intellij-ext/gradle.properties
```diff
@@ -8,7 +8,7 @@ pluginVersion = 3.2.0
 
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 # for insight into build numbers and IntelliJ Platform versions.
-pluginSinceBuild = 213
+pluginSinceBuild = 222
```
eyalb181, line 11: Why was this necessary?

CHANGELOG.md
```diff
@@ -7,26 +7,43 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+### Added
```
eyalb181, line 10: Did you use some linter here? Perhaps we should add it to the CI so we don't crash on release?

infiniteregrets: I think that should be a separate PR, it will include ignoring tests when *.md files are changed and a linter

intellij-ext/gradle.properties
infiniteregrets: sorry was just testing something with builds, fixed

CHANGELOG.md
eyalb181: Lets add it to this PR, since if we don't, a typo in the changelog could break our release flow at a very late stage.

infiniteregrets: okay sure

intellij-ext/gradle.properties
```diff
@@ -19,7 +19,7 @@ platformVersion = 2022.2
 platformPlugins = org.jetbrains.plugins.go:222.3345.118, org.intellij.intelliLang
 
 # Java language level used to compile sources and to generate the files for - Java 11 is required since 2020.3
-javaVersion = 11
+javaVersion = 17
```
eyalb181, line 22: Why is this necessary?

intellij-ext/build.gradle.kts
```diff
@@ -112,6 +124,7 @@ tasks {
     }
 
     publishPlugin {
+        dependsOn("patchChangelog")
```
eyalb181, line 127: Why is this needed? Don't we update the CHANGELOG manually?

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordEnabler.kt
```diff
@@ -1,19 +1,15 @@
...
-import com.intellij.openapi.actionSystem.ToggleAction
-import com.intellij.openapi.application.ApplicationManager
-import com.intellij.openapi.project.Project
 import com.github.zafarkhaja.semver.Version
 import com.intellij.ide.plugins.PluginManagerConfigurable
 import com.intellij.ide.plugins.PluginManagerCore
 import com.intellij.ide.util.PropertiesComponent
-import com.intellij.notification.NotificationAction
+import com.intellij.notification.*
+import com.intellij.openapi.actionSystem.AnActionEvent
```
eyalb181, line 8: What happened here?

infiniteregrets: that was intellij's linter

intellij-ext/gradle.properties
infiniteregrets: - The Java configuration specifies sourceCompatibility=11 but IntelliJ Platform 2022.2 requires sourceCompatibility=17.
See: https://jb.gg/intellij-platform-versions

when run `./gradlew buildPlugin`

infiniteregrets: i believe this could be the reason: https://blog.jetbrains.com/platform/2022/08/intellij-project-migrates-to-java-17/
but didn't sound relevant since it says its needed for 2022.3, but here https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html#intellij-platform-based-products-of-recent-ide-versions it says it is needed for 2022.2

intellij-ext/build.gradle.kts
infiniteregrets: looks like this might not be needed here, I picked this from the template we are using but it works without it as well

Issue #390, opened 2022-09-11
Alternative log level env var for `mirrord-layer`


tamasfe:
Currently it's not possible to silence logs from `mirrord-layer` when developing a rust application as it uses the `RUST_LOG` env variable.

Something like `MIRRORD_RUST_LOG` could be introduced that overrides `RUST_LOG` if present.

Alternatively logging to file could be preferred instead of silencing, as these messages are still important, but should not be mixed with the executed app's stderr or stdout.

aviramha:
AFAIK you can use `RUST_LOG=mirrord=warning` then mirrord will print only warnings/errors?

tamasfe:
After some thinking, the current behaviour *feels* wrong but I don't have alternative suggestions yet. `RUST_LOG=info,mirrord=error` or similar could work, but for some reason I think the logs should be separate. I think additional support for logging to a file instead of stderr should be enough.

aviramha:
I thought about it as well - I came to the idea of having a `-v,vv,vvv,vvvv,vvvvv` flags that will enable the logs as most users aren't aware of rust logging facility, maybe as part of it we can separate it from using `RUST_LOG`

PR #641, opened 2022-10-26
[Merged by Bors] - Some Improvements to intelliJ extension

jamowei:
* same default options as CLI
* improved UI layouts
* fixed not reappearing window after pressing cancel-button
* separate tcp and udp outgoing option
* allow only single selection for namespaces and pods
* refactor UI-creating code to a more Kotlin functional style


infiniteregrets:
hi @jamowei, thanks again for improving the extension!

Just a quick note on the namespaces,

They look like this for me:
Screen Shot 2022-10-26 at 11 37 05 AM

even though I have a few more namespaces

intellij-ext/src/main/kotlin/com/metalbear/mirrord/LogLevel.kt
```diff
@@ -0,0 +1,5 @@
+package com.metalbear.mirrord
+
+enum class LogLevel {
+    ERROR, WARN, INFO, DEBUG, TRACE;
+}
```
infiniteregrets, line 5: The reason we kept this as a text field was to allow the user to choose whatever logs they wanna see, forexample, 
RUST_LOG=mirrord=debug, warn

infiniteregrets: @eyalb181 wdyt? should we keep this as a dropdown or a text field

infiniteregrets (commented):
Looks great!

intellij-ext/src/main/kotlin/com/metalbear/mirrord/LogLevel.kt
eyalb181: We're probably going to change how our logs are configured anyway (see #390 ), but for now lets keep it as a dropdown, it looks much better and doesn't limit functionality by much.

infiniteregrets:
> hi @jamowei, thanks again for improving the extension!
> 
> Just a quick note on the namespaces,
> 
> They look like this for me: Screen Shot 2022-10-26 at 11 37 05 AM
> 
> even though I have a few more namespaces

this works okay, on a different run.

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.nio.file.Paths
+
+data class MirrordDefaultConfig(
+    val ldPreloadPath: String = getSharedLibPath("libmirrord_layer.so"),
+    val dylibPath: String = getSharedLibPath("libmirrord_layer.dylib"),
+    val acceptInvalidCertificates: Boolean = true,
+    val skipProcesses: String = "",
+    val fileOps: Boolean = true,
+    val stealTraffic: Boolean = true,
+    val ephemeralContainers: Boolean = true,
```
infiniteregrets, line 13: this should be set to false by default

infiniteregrets (changes requested):
small change

jamowei:
> > hi @jamowei, thanks again for improving the extension!
> > Just a quick note on the namespaces,
> > They look like this for me: Screen Shot 2022-10-26 at 11 37 05 AM
> > even though I have a few more namespaces
> 
> this works okay, on a different run.

I did not change the code for receiving the Kubernetes namespaces ;-)

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
jamowei: Ok, I changed it :)

infiniteregrets (approved):
Thanks again for the improvements!

aviramha:
This is awesome! Thank you so much.

Issue #637, opened 2022-10-26
Use k8s auth once


aviramha:
We have many issues from the scenario where the user has a k8s auth that depends on executing a command (`aws cli`, `gcp`, etc).
We have some workarounds, but it affects use cases where you'd like to run aws cli in context of remote or even normal use cases right now where user mirrors a remote pod which has AWS env, and then it forks - on launch it gets the AWS env affecting the new kube rs client configuration, making it fail to load.

PR #642, opened 2022-10-26
[Merged by Bors] - K8s consistent env variables across forks

DmitryDodzin:
closing #637


mirrord-layer/src/pod_api.rs
```diff
@@ -24,20 +31,80 @@ use crate::{
...
     #[cfg(target_os = "macos")]
     const ENV_VAR: &str = "DYLD_INSERT_LIBRARIES";
+
     fn new() -> Self {
         std::env::set_var(MIRRORD_SKIP_LOAD, "true");
         let library = std::env::var(EnvVarGuard::ENV_VAR).unwrap_or_default();
         std::env::remove_var(EnvVarGuard::ENV_VAR);
-        Self { library }
+
+        let fork_args: HashMap<_, _> = std::env::vars()
```
aviramha, line 54: we should drop this imo, because we control execution now

aviramha (changes requested):
Small correction and I think we're good to go.
P.S I think you forgot to commit the tests itself?

PR #653, opened 2022-10-27
[Merged by Bors] - remove OpenSSL dependency

aviramha:


meowjesty (approved):
👍 

ps: can't reqwest be workspace in sample/rust as well?

aviramha:
> 👍
> 
> ps: can't reqwest be workspace in sample/rust as well?

I was afraid that some folks would want to compile it standalone?
I don't like it being part of the workspace anyway..

PR #659, opened 2022-10-28
[Merged by Bors] - Remove `--pod-name`, `--pod-namespace`, `--impersonated_container_name` in favor of `--target`, `--target-namespace`"

infiniteregrets:


eyalb181 (approved):

mirrord-cli/src/config.rs
```diff
@@ -144,7 +121,7 @@ pub(super) struct ExecArgs {
     pub no_telemetry: bool,
 
     /// Load config from config file
-    #[clap(short = 'f', conflicts_with_all = &["target", "pod-name"], long, value_parser)]
+    #[clap(short = 'f', conflicts_with_all = &["target"], long, value_parser)]
```
DmitryDodzin, line 124: I don't think we need `conflicts_with_all` at all, you can load config file and target, it will work just fine from MirrordConfig's perspective

mirrord-cli/src/config.rs
```diff
@@ -23,42 +23,19 @@ pub(super) enum Commands {
...
 pub(super) struct ExecArgs {
     /// Target name to mirror.    
     /// Target can either be a deployment or a pod.
     /// Valid formats: deployment/name, pod/name, pod/name/container/name
-    #[clap(short, long, value_parser)]
+    #[clap(short = 't', long, value_parser)]
     pub target: Option<String>,
 
     /// Namespace of the pod to mirror. Defaults to "default".
     #[clap(long, value_parser)]
```
DmitryDodzin, line 36:
```suggestion
    #[clap(short = 'n', long, value_parser)]
```

infiniteregrets: why "n"? --target, or -t would make sense, wouldn't it?

eyalb181: +1

DmitryDodzin (approved):

Issue #644, opened 2022-10-26
`ERROR: transport error 202: connect failed: Unknown error: 111` when debugging a JVM-app


jamowei:
### Bug Description

I get the following error when I set the `MIRRORD_TCP_OUTGOING` to `true` on macOS an and a JVM-based service:
```
2022-10-26T18:13:26.498142Z  INFO ThreadId(01) mirrord_layer::go_env: replace -> hooking "runtime.goenvs_unix"
2022-10-26T18:13:26.688625Z  INFO ThreadId(15) mirrord_layer::socket::hooks: 14
2022-10-26T18:13:26.718586Z  INFO ThreadId(15) mirrord_layer::error: libc error (doesn't indicate a problem) >> ResponseError(
    RemoteIO(
        RemoteIOError {
            raw_os_error: Some(
                111,
            ),
            kind: ConnectionRefused,
        },
    ),
)
ERROR: transport error 202: connect failed: Unknown error: 111
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [./src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:735]

Process finished with exit code 2
```

Here are my env vars which I'm using:
```
DYLD_INSERT_LIBRARIES=/Users/jamowei/Library/Application Support/JetBrains/IntelliJIdea2022.2/plugins/mirrord/libmirrord_layer.dylib
LD_PRELOAD=/Users/jamowei/Library/Application Support/JetBrains/IntelliJIdea2022.2/plugins/mirrord/libmirrord_layer.so
MIRRORD_AGENT_RUST_LOG=DEBUG
RUST_LOG=DEBUG
MIRRORD_ACCEPT_INVALID_CERTIFICATES=true
MIRRORD_EPHEMERAL_CONTAINER=false
MIRRORD_SKIP_PROCESSES=
MIRRORD_OVERRIDE_ENV_VARS_INCLUDE=*
MIRRORD_IMPERSONATED_TARGET=pod/my-service-xxx
MIRRORD_TARGET_NAMESPACE=mynamespace
MIRRORD_FILE_OPS=false
MIRRORD_AGENT_TCP_STEAL_TRAFFIC=false
MIRRORD_REMOTE_DNS=true
MIRRORD_TCP_OUTGOING=true
MIRRORD_UDP_OUTGOING=false
```

[...]

PR #660, opened 2022-10-28
[Merged by Bors] - Env var to ignore ports used by debugger for intelliJ

infiniteregrets:
refer #644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,6 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 ### Fixed
 
 - Env var to ignore ports used by a debugger for intelliJ/VSCode, refer [#644](https://github.com/metalbear-co/mirrord/issues/644)
-- Ignore debugger ports only when ip is localhost, fixed a bug which was reintroduced in refactors - [#370](https://github.com/metalbear-co/mirrord/pull/370)
 
 ## 3.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ### Fixed
 
-- Env var to ignore ports used by a debugger for intelliJ, refer [#644](https://github.com/metalbear-co/mirrord/issues/644)
+- Env var to ignore ports used by a debugger for intelliJ/VSCode, refer [#644](https://github.com/metalbear-co/mirrord/issues/644)
+- Ignore debugger ports only when ip is localhost, fixed a bug which was reintroduced in refactors - [#370](https://github.com/metalbear-co/mirrord/pull/370)
 
 ## 3.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 - mirrord-config: Changed the way [`fs`](mirrord-config/src/fs.rs) works: now it supports 2 modes `Simple` and `Advanced`, where `Simple` is similar to the old behavior (enables read-only, read-write, or disable file ops), and `Advanced` allows the user to specify include and exclude (regexes) filters for [`FileFilter`](mirrord-layer/src/file/filter.rs).
 - Lint `README` and update it for `--target` flag.
 
+### Fixed
+
+- Env var to ignore ports used by a debugger for intelliJ, refer [#644](https://github.com/metalbear-co/mirrord/issues/644)
+
 ## 3.4.0
 
 ### Added
--- a/intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
+++ b/intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
@@ -19,6 +19,7 @@ data class MirrordDefaultConfig(
     val rustLog: LogLevel = LogLevel.INFO,
     val overrideEnvVarsExclude: String = "",
     val overrideEnvVarsInclude: String = "*",
+    val ignorePorts: String = "",
 )
 
 private fun getSharedLibPath(libName: String): String {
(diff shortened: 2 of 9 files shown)


mirrord-layer/src/lib.rs
```diff
@@ -117,6 +117,23 @@ fn is_nix_or_devbox() -> bool {
     }
 }
 
+/// Prevent mirrord from connecting to ports used by the intelliJ debugger
+pub(crate) fn intellij_debug_patch(port: Port) -> bool {
+    if let Ok(ports) = std::env::var("INTELLIJ_DEBUG_PATCH") {
+        // port range can be specified as "1234-1236" or just "1234"
+        let ports: Vec<u16> = ports.split('-').map(|p| p.parse().unwrap()).collect();
```
meowjesty, line 124: Change `unwrap` here to `expect`, so you can give a nicer user error, like "Tried to parse a value that is not a number" or something along those lines.

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
```diff
@@ -19,6 +19,7 @@ data class MirrordDefaultConfig(
     val rustLog: LogLevel = LogLevel.INFO,
     val overrideEnvVarsExclude: String = "",
     val overrideEnvVarsInclude: String = "*",
+    val ignorePorts: String = "",
```
meowjesty, line 22: Considering how lacking the documentation is around this, will the user even know which ports to ignore? Maybe we should do a default range for intellij?

jamowei:
Nice work! I would recommend using a more general name for the new env var instead of `INTELLIJ_IGNORE_PORTS`. Because in my opinion, this is not a problem with IntelliJ itself and also for other use-cases a valid option to choose. My suggestion is something like `MIRRORD_SKIP_PORTS` like `MIRRORD_SKIP_PROCESSES`.

intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordDefaultConfig.kt
infiniteregrets: sure, but it is kind of random. we can keep it from 45K-65K wdyt?

meowjesty:
@jamowei 

 > MIRRORD_SKIP_PORTS

Yeah we plan on adding something like this in the (near) future, just gotta figure out a good API for this first (the user should also be able to specify ip:port combination, port range + some individual specific ports, maybe even have some default skips based on the debugger - like the intellij plugin checking the port automatically).

That's why we're (initially) committing for a small intellij fix.

meowjesty (approved):
👍

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 - mirrord-config: Changed the way [`fs`](mirrord-config/src/fs.rs) works: now it supports 2 modes `Simple` and `Advanced`, where `Simple` is similar to the old behavior (enables read-only, read-write, or disable file ops), and `Advanced` allows the user to specify include and exclude (regexes) filters for [`FileFilter`](mirrord-layer/src/file/filter.rs).
 - Lint `README` and update it for `--target` flag.
 
+### Fixed
+
+- Env var to ignore ports used by a debugger for intelliJ, refer [#644](https://github.com/metalbear-co/mirrord/issues/644)
+
```
meowjesty, line 25: Add a notice about the regression you fixed as well with `is_ignored_ip`.

mirrord-layer/src/socket.rs
```diff
@@ -129,6 +129,12 @@ const fn is_ignored_port(port: Port) -> bool {
     port == 0 || (port > 50000 && port < 60000)
 }
 
+/// Ignore localhost Ipv4 and Ipv6 addresses, used along with `is_ignored_port` for debuggers
+#[inline]
+fn is_ignored_ip(ip: IpAddr) -> bool {
```
aviramha, line 134: why not mix it with the `port_debug_patch`? those ips aren't generally ignored and might confuse reader

aviramha (changes requested):
Almost there.
Shouldn't be a change to the IntelliJ extension to use the new flag?

mirrord-layer/src/lib.rs
```diff
@@ -117,6 +118,30 @@ fn is_nix_or_devbox() -> bool {
     }
 }
 
+/// Prevent mirrord from connecting to ports used by the intelliJ debugger
+pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
+    if let Ok(ports) = std::env::var("DEBUGGER_IGNORE_PORTS_PATCH") {
+        let (ip, port) = (addr.ip(), addr.port());
+        let ignored_ip =
```
aviramha, line 125: nit pick - can just return false here in case of not ignored ip.

mirrord-layer/src/lib.rs
```diff
@@ -117,6 +118,30 @@ fn is_nix_or_devbox() -> bool {
     }
 }
 
+/// Prevent mirrord from connecting to ports used by the intelliJ debugger
+pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
+    if let Ok(ports) = std::env::var("DEBUGGER_IGNORE_PORTS_PATCH") {
```
eyalb181, line 123: Doesn't Rust have some method that would let you unwrap_or_return(false)?

mirrord-layer/src/socket.rs
```diff
@@ -125,8 +125,10 @@ pub struct UserSocket {
 }
 
 #[inline]
-const fn is_ignored_port(port: Port) -> bool {
-    port == 0 || (port > 50000 && port < 60000)
+fn is_ignored_port(addr: SocketAddr) -> bool {
+    let (ip, port) = (addr.ip(), addr.port());
+    let ignored_ip = ip == IpAddr::V4(Ipv4Addr::LOCALHOST) || ip == IpAddr::V6(Ipv6Addr::LOCALHOST);
+    port == 0 || ignored_ip && (port > 50000 && port < 60000)
```
eyalb181, line 131: Aren't these ports the VSCode debugger ports, and thus should be passed from the VS Code extension?

aviramha (approved):
bors r+

PR #662, opened 2022-10-29
Add bump homebrew formula job to release CI

infiniteregrets:


.github/workflows/release.yaml
```diff
@@ -289,3 +289,17 @@ jobs:
 
       - name: Update latest tag
         uses: EndBug/latest-tag@latest
+
+  bump_homebrew:
```
aviramha, line 293: add if verifying it's not a manual run.

.github/workflows/release.yaml
```diff
@@ -289,3 +289,17 @@ jobs:
...
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
+        with:
+          formula-name: mirrord
+          formula-path: mirrord.rb
+          homebrew-tap: metalbear-co/homebrew-mirrord
+          download-url: https://github.com/metalbear-co/mirrord/releases/download/${{ github.ref_name }}/mirrord_mac_universal.zip
+          commit-message: Bump mirrord to {{version}}
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
```
aviramha, line 305: did we add this?

.github/workflows/release.yaml
```diff
@@ -289,3 +289,17 @@ jobs:
...
+  bump_homebrew:
+    needs: release_gh
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
+        with:
+          formula-name: mirrord
+          formula-path: mirrord.rb
+          homebrew-tap: metalbear-co/homebrew-mirrord
+          download-url: https://github.com/metalbear-co/mirrord/releases/download/${{ github.ref_name }}/mirrord_mac_universal.zip
```
aviramha, line 302: can we add support for linux as well?
see this reference for a formula btw

infiniteregrets: sure, we can but,
- how many users even use homebrew on linux? I don't there is a good enough user base to add the linux formula now, rather just open an issue for adding it to linux
- besides this we'd probably have to package a tar zip containing both linux and macos binaries and then'd we'd unwrap them in the installation phase

.github/workflows/release.yaml
infiniteregrets: nope, we need to add this

.github/workflows/release.yaml
aviramha: sorry I didn't post reference:
https://github.com/goreleaser/homebrew-tap/blob/master/Formula/goreleaser.rb
It looks very simple and valuable. The main benefit is not having to use curl | sh which is intimidating to some users (feedback I received)

PR #679, opened 2022-11-02
[Merged by Bors] - IntelliJ plugin: use the default namespace if namespaces cant be accessed

### LINKED ISSUE #678: intellij-ext: mirrord doesn't work when the service account doesn't have permissions to list namespaces
### Bug Description

Hello! I have limited access to a work cluster and one of the permissions I'm lacking is the ability to list namespaces (https://github.com/metalbear-co/mirrord/blob/main/intellij-ext/src/main/kotlin/com/metalbear/mirrord/MirrordListener.kt#L42).

I'm wondering if you can add some type of handling to either get the current namespace only, or allow user input to specify the namespace.

Thanks!

### Steps to Reproduce

1. Create a service account that doesn't have LIST permissions to namespaces.
2. Set your current context to use that service account.
3. Use IntelliJ and enable the mirrord plugin.
4. Run your application.

### Backtrace

_No response_

### Relevant Logs

```shell
When you run your application, you will get an error notification saying: "Error occurred while fetching namespaces from Kubernetes context"
```

macOS Monterey 12.6

### Local process

IntelliJ + Mirrord Extension

### Local process version

_No response_

### Additional Info

I have access to list the pods from the current namespace like so `kubectl get pods --namespace=test-namespace` just not able to do `kubectl get namespaces`.

infiniteregrets:
closes #678

### DIFF
--- a/intellij-ext/src/main/kotlin/com/metalbear/mirrord/KubeDataProvider.kt
+++ b/intellij-ext/src/main/kotlin/com/metalbear/mirrord/KubeDataProvider.kt
@@ -1,20 +1,21 @@
 package com.metalbear.mirrord
+
 import io.kubernetes.client.openapi.apis.CoreV1Api
 import io.kubernetes.client.util.Config
 import io.kubernetes.client.util.Namespaces
 
 class KubeDataProvider : CoreV1Api(Config.defaultClient()) {
-    fun getNameSpacedPods(namespace: String?): List<String> {
+    fun getNameSpacedPods(namespace: String): List<String> {
         val pods = listNamespacedPod(namespace, null, null, null, null, null, null, null, null, null, null)
-        return pods.items.map{ it.metadata!!.name!! }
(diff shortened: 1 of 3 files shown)


intellij-ext/src/main/kotlin/com/metalbear/mirrord/KubeDataProvider.kt
```diff
@@ -1,15 +1,20 @@
 package com.metalbear.mirrord
 import io.kubernetes.client.openapi.apis.CoreV1Api
 import io.kubernetes.client.util.Config
+import io.kubernetes.client.util.Namespaces
 
 class KubeDataProvider : CoreV1Api(Config.defaultClient()) {
-    fun getNameSpacedPods(namespace: String): List<String> {
+    fun getNameSpacedPods(namespace: String?): List<String> {
```
eyalb181, line 7: When is this ever null?

eyalb181 (approved):

PR #696, opened 2022-11-06
[Merged by Bors] - Change error message to info when connection is closed by agent

### LINKED ISSUE #693: Layer exits on bad connection state
Layer exits on the following error:
` 2022-11-06T12:55:18.475810Z ERROR ThreadId(10) mirrord_layer: Error handling daemon message: SendErrorConnection(SendError([23, 3, 3, 12, 88, 0, 0, 0, 0, 0, 0, 0, 1, 119, 99, 145, 233, 176, 18, 222, 251, 141, 119, 184, 89, 17, 151, 52, 0, 100, 92, 40, 86, 223, 109, 220, 69, 30, 42, 35, 254, 70, 114, 66,`
This could happen because local/remote connection state mismatch on mirroring, and we should be tolerant for it and also not print it as an error, maybe a warning explaining the situation - on mirroring, state mismatch between local/remote?

eyalb181:
Closes #693

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -396,6 +396,10 @@ async fn thread_loop(
                     Some(Ok(message)) => {
                         if let Err(err) = layer.handle_daemon_message(
                             message).await {
+                            if let LayerError::SendErrorConnection(_) = err {
+                                info!("Connection closed by agent");
+                                continue;
+                            }
                             error!("Error handling daemon message: {:?}", err);
                             break;
                         }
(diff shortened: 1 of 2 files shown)


aviramha (approved):

mirrord/layer/src/lib.rs
```diff
@@ -396,6 +396,10 @@ async fn thread_loop(
                     Some(Ok(message)) => {
                         if let Err(err) = layer.handle_daemon_message(
                             message).await {
+                            if let LayerError::SendErrorConnection(_) = err {
+                                info!("Connection closed by agent");
+                                continue;
```
infiniteregrets, line 401: perhaps we can write why was the connection closed or what could be a possibility

infiniteregrets (commented):
👍

aviramha (approved):

mirrord/layer/src/lib.rs
eyalb181: It's not an error so I think we can leave it as is

PR #699, opened 2022-11-06
[Merged by Bors] - Config macro refactor

### LINKED ISSUE #664: Reverse MirrordConfig macro struct implementation
Right now `MirrordConfig`'s macro is placed on the file representation of the config and not the code version.

Better solution would be to generate the file config representation with macro
```rust
#[derive(Debug, Clone, MirrordConfig)]
#[cfg_attr(test, config(derive = PartialEq))]
#[cfg_attr(test, config(derive = Eq))]
pub struct TestConfig {
    #[config(from = "MIRRORD_TEST")]
    pub test: String,
}
```
will generate
```rust
#[derive(Debug, serde::Deserialize, PartialEq, Eq)]
pub struct FileTestConfig {
    pub test: Option<String>,
}
```

though not sure how to allow the functionality of simplifying collapsible structs in config ("Simple" / "Advanced" typed scenario)

DmitryDodzin:
closes #664


mirrord/config/derive/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+        let fields = match data {
+            Data::Struct(DataStruct { fields, .. }) => match fields {
+                Fields::Named(FieldsNamed { named, .. }) => named
+                    .into_iter()
+                    .map(ConfigField::try_from)
+                    .collect::<Result<_, _>>()?,
+                _ => return Err(fields.span().error("Unnamed Structs are not supported")),
+            },
+            _ => return Err(source.span().error("Enums and Unions are not supported")),
+        };
```
meowjesty, line 46: If we only care about 1 variant, think we should probably use `if let else` instead of `match`?

```suggestion
        let fields = if let Data::Struct(DataStruct { fields: Fields::Named(FieldsNamed { named, .. }), .. }) = data {
            Ok(named
                .into_iter()
                .map(ConfigField::try_from)
                .collect::<Result<_, _>>()?)
        } else {
            Err(source.span().error("Enums, Unions, and Unnamed Structs are not supported"))
        }?;
```

mirrord/config/derive/src/config/field.rs
```diff
@@ -0,0 +1,123 @@
...
+    flags: ConfigFlags,
+}
+
+impl ConfigField {
+    fn is_option(field: &Field) -> Option<Type> {
+        let seg = if let Type::Path(ty) = &field.ty {
+            ty.path.segments.first()?
+        } else {
+            return None;
+        };
```
meowjesty, line 22: I would like us to keep being "expression based" as much as possible:

```suggestion
        let seg = if let Type::Path(ty) = &field.ty {
            ty.path.segments.first()
        } else {
            None
        }?;
```

mirrord/config/derive/src/config/field.rs
```diff
@@ -0,0 +1,123 @@
...
+            return None;
+        }
+
+        match &seg.arguments {
+            PathArguments::AngleBracketed(generics) => match generics.args.first() {
+                Some(GenericArgument::Type(ty)) => Some(ty.clone()),
+                _ => None,
+            },
+            _ => None,
+        }
```
meowjesty, line 34: Again, my preference is for expressions, so I would suggest something like:

```suggestion
(seq.ident == "Option").then(|| {
    if let PathArguments::AngleBracketed(Some(GenericArgument::Type(ty))) = &seq.arguments {
        Some(ty.clone())
    } else {
        None
    }
})       
```

mirrord/config/derive/src/config/field.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    pub fn definition(&self) -> impl ToTokens {
```
meowjesty, line 37: Add some rust docs here, ideally with an example of what this function will generate.

mirrord/config/derive/src/config/field.rs
```diff
@@ -0,0 +1,123 @@
...
+            quote! { #ty }
+        };
+
+        quote! {
+            #(#docs)*
+            #vis #ident: Option<#target>
+        }
+    }
+
+    pub fn implmentation(&self, parent: &Ident) -> impl ToTokens {
```
meowjesty, line 69: Same request for documentation here.

mirrord/config/derive/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+            source,
+            ident,
+            fields,
+            flags,
+        })
+    }
+}
+
+impl ToTokens for ConfigStruct {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
```
meowjesty, line 64: Document the end result with an example.

mirrord/config/src/env.rs
```diff
@@ -58,10 +57,10 @@ use crate::{
 /// [feature.env]
 /// exclude = "USER;SECRET"
 /// ```
-#[derive(MirrordConfig, Default, Deserialize, PartialEq, Eq, Clone, Debug, JsonSchema)]
-#[serde(deny_unknown_fields)]
-#[config(map_to = EnvConfig)]
-pub struct EnvFileConfig {
+#[derive(MirrordConfig, Clone, Debug)]
+#[config(map_to = "EnvFileConfig", derive = "JsonSchema")]
```
meowjesty, line 61: This one is more of a question rather than an actual suggestion:

Do you prefer having these arguments be `Lit` and not `Ident`? I'm ok with either (slight bias towards `Ident`, but again, it's ok as-is).

mirrord/config/src/env.rs
```diff
@@ -79,16 +78,15 @@ pub struct EnvFileConfig {
     ///
     /// For example, if the remote pod has an environment variable `REGION=1`, but this is an
     /// undesirable value, it's possible to use `overrides` to set `REGION=2` (locally) instead.
-    #[serde(rename = "override")]
-    pub overrides: Option<HashMap<String, String>>,
+    pub r#override: Option<HashMap<String, String>>,
```
meowjesty, line 81: How about creating a `#[rename("override")]` attribute for the macro so we can avoid this?

mirrord/config/src/config.rs
```diff
@@ -30,3 +30,7 @@ where
         self.unwrap_or_default().generate_config()
     }
 }
+
+pub trait FromMirrordConfig {
```
meowjesty, line 34: We're planning on turning on "deny no docs" sooner rather than later, so please add docs for this.

meowjesty (changes requested):
Initial review, full of nitpicks 😅 .

Also, requesting rust docs for everything (we want to turn on "deny no docs" - or whatever the lint is called).

mirrord/config/src/env.rs
DmitryDodzin: I wanted them to be Ident but it doesn't work, If I don't specify it as a Lit it simply doesn't appear in result of `parse_meta`

meowjesty: Yeah, that's probably why many projects tend to go with `Lit`. Again this is fine, will mark as resolved.

meowjesty (approved):
👍

PR #700, opened 2022-11-06
[Merged by Bors] - intelliJ: use custom delve on macos

### LINKED ISSUE #695: use custom delve in IntelliJ
Following https://github.com/metalbear-co/mirrord/pull/694 we need to do the same for IntelliJ.
In case of macOS + Go, we edit the delve command line to use our delve instead of the one provided by homebrew/other means.

infiniteregrets:
closes #695

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -260,10 +260,20 @@ jobs:
         build_binaries_macos,
         release_docker_image,
       ]
-    runs-on: ubuntu-latest
+    runs-on: macos-12
     steps:
       - uses: actions/checkout@v2
         id: version
+      - name: build patched Delve
+        uses: actions/setup-go@v3
+        with:
+          go-version: "1.19.0"
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -260,10 +260,20 @@ jobs:
         build_binaries_macos,
         release_docker_image,
       ]
-    runs-on: ubuntu-latest
+    runs-on: macos-12
```
eyalb181, line 263: Does the entire intellij release step run on macos now? Was it tested?

aviramha: wait.. no need to do it. run it on ubuntu, just wget the version like the other job does.

aviramha (changes requested):

.github/workflows/release.yaml
infiniteregrets: oh right, just saw the vscode fix was already done

aviramha (approved):
LGTM

Issue #649, opened 2022-10-27
SIP binaries support on macOS


aviramha:
Currently we don't load into sip protected binaries on macOS. We have a way to fix it, need to implement.
If you face a use case where mirrord failed to load on macOS, please upvote/comment on this ticket with details so we can prioritize it accordingly.

t4lz:
Status update: 
https://github.com/metalbear-co/mirrord/pull/706
Makes it possible to run a sip protected binary with mirrord, also via a shebang chain.
However, we don't yet bypass SIP for binaries that are executed by the executed binary. So things like bash scripts, where usually almost each command is starting some SIP protected binaries won't work on Mac as expected yet. Working now on supporting this.

Small note: we should now put a hash of the original file in the path of the patched file, so that if you write a bash script with a shebang, run it with mirrord, we patch it and save it, then you change your script and run it with mirrord again, we now patch the new version, and not use the old patched version.

t4lz:
Additional note: we should probably mark all the generated files as excluded from filesystem hooks.
Maybe even walk the shebang chain also on Linux to mark all to be ignored regardless of SIP?

It seems for example that when running a script with this shebang `#!/bin/bash` bash `open`s the file of the script for reading (and does not find it if FS is not disabled).

Are there scenarios where a user would want some scripts/commands to be read from the pod?

aviramha:
I agree but the filter should probably be extension of file and that should be general enough.

aviramha:
> Small note: we should now put a hash of the original file in the path of the patched file, so that if you write a bash script with a shebang, run it with mirrord, we patch it and save it, then you change your script and run it with mirrord again, we now patch the new version, and not use the old patched version.

Agree. Create an issue on that, we'll probably wait a bit with that as that's pretty edgy, but ofc we want to fix it.

t4lz:
Created. https://github.com/metalbear-co/mirrord/issues/717
Should be very easy though, if it's the 1 line change I think it is, I would include it in the PR for this issue, if that's ok.

aviramha:
> Created. #717 Should be very easy though, if it's the 1 line change I think it is, I would include it in the PR for this issue, if that's ok.

Wouldn't bother right now. It might seem simple, but testing/actual running might become complicated.

t4lz:
[...]

PR #706, opened 2022-11-08
[Merged by Bors] - Bypass SIP on MacOS

t4lz:
Part of https://github.com/metalbear-co/mirrord/issues/649.

On MacOS, test if the file that the user specified to run is protected by SIP, or is a script with a shebang that leads to such a binary (also via multiple shebang scripts). If it is, create an unprotected version of the binary (and the shebang scripts that lead to it) and run that instead.

This does not yet bypass SIP on binaries that the executed application executes.

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -8,10 +8,9 @@ use std::{
 use anyhow::{anyhow, Context, Result};
 use clap::Parser;
 use config::*;
-use exec::execvp;
-
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use errno;
+use exec::execvp;
 use mirrord_auth::AuthConfig;
 use mirrord_progress::TaskProgress;
 #[cfg(target_os = "macos")]
@@ -223,8 +222,10 @@ fn exec(args: &ExecArgs) -> Result<()> {
     let err = execvp(binary, binary_args);
     error!("Couldn't execute {:?}", err);
     #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
-    if let exec::Error::Errno(errno::Errno(86)) = err { // "Bad CPU type in executable"
-        if did_sip_patch { // We did a SIP patch.
+    if let exec::Error::Errno(errno::Errno(86)) = err {
+        // "Bad CPU type in executable"
+        if did_sip_patch {
+            // We did a SIP patch.
             error!(
                 "The file you are trying to run, {}, is either SIP protected or a script with a
                 shebang that leads to a SIP protected binary. In order to bypass SIP protection,
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -11,12 +11,12 @@ use config::*;
 use exec::execvp;
 use mirrord_auth::AuthConfig;
 use mirrord_progress::TaskProgress;
+#[cfg(target_os = "macos")]
+use mirrord_sip::sip_patch;
 use rand::distributions::{Alphanumeric, DistString};
 use semver::Version;
 use tracing::{debug, error, info, warn};
 use tracing_subscriber::{fmt, prelude::*, registry, EnvFilter};
-#[cfg(target_os = "macos")]
-use {regex::RegexSet, which::which};
 
 mod config;
 
@@ -98,33 +98,6 @@ fn add_to_preload(path: &str) -> Result<()> {
     }
 }
 
-#[cfg(target_os = "macos")]
-fn sip_check(binary_path: &str) -> Result<()> {
-    let sip_set = RegexSet::new([
-        r"/System/.*",
-        r"/bin/.*",
-        r"/sbin/.*",
-        r"/usr/.*",
-        r"/var/.*",
-        r"/Applications/.*",
-    ])?;
-    let complete_path = which(binary_path)?;
-
-    let sliced_path = complete_path.to_str().ok_or_else(|| {
-        anyhow!(
-            "Failed to convert path to a string slice: {}",
-            binary_path.to_string()
-        )
-    })?;
-
-    if sip_set.is_match(sliced_path) {
-        println!("[WARNING]: Provided binary: {:?} is located in a SIP directory. mirrord might fail to load into it.
-        >> for more info visit https://support.apple.com/en-us/HT204899", binary_path);
-    }
-
-    Ok(())
-}
-
 fn exec(args: &ExecArgs) -> Result<()> {
     if !args.no_telemetry {
         prompt_outdated_version();
@@ -135,7 +108,7 @@ fn exec(args: &ExecArgs) -> Result<()> {
     );
 
     #[cfg(target_os = "macos")]
-    sip_check(&args.binary)?;
+    sip_patch(&mut args.binary)?;
 
     if !(args.no_tcp_outgoing || args.no_udp_outgoing) && args.no_remote_dns {
         warn!("TCP/UDP outgoing enabled without remote DNS might cause issues when local machine has IPv6 enabled but remote cluster doesn't")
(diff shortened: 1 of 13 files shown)


mirrord/sip/src/whitespace.rs
```diff
@@ -0,0 +1,67 @@
+// source: https://github.com/dtolnay/syn/blob/master/src/whitespace.rs
```
aviramha, line 1: include its license

mirrord/cli/src/main.rs
```diff
@@ -236,11 +208,39 @@ fn exec(args: &ExecArgs) -> Result<()> {
     let library_path = extract_library(args.extract_path.clone())?;
     add_to_preload(library_path.to_str().unwrap()).unwrap();
 
+    #[cfg(target_os = "macos")]
```
aviramha, line 211: Thinking out loud - maybe we want to check if rosetta is available before doing the patch?
Also the desired behavior might be just a warning and then "not loading" to the said binary?
wdyt?

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,407 @@
+#[cfg(target_os = "macos")]
+mod codesign;
+#[cfg(target_os = "macos")]
+mod error;
+#[cfg(target_os = "macos")]
+mod whitespace;
+
+#[cfg(target_os = "macos")]
+mod main {
```
aviramha, line 9: @meowjesty when I originally created this layout, I thought maybe there's a better solution than all this `#[cfg(..)]` maybe you have a suggestion.

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /// Creates a new file at `patched_path` which has the same contents as `original_path` except
+    /// for the shebang which is `new_shebang`.
+    fn patch_script<P: AsRef<Path>, K: AsRef<Path>>(
+        original_path: P,
+        patched_path: K,
+        new_shebang: &str,
+    ) -> Result<()> {
+        if let Some(original_shebang) = read_shebang_from_file(&original_path)? {
```
aviramha, line 113: I think a better pattern here would be `.map`? not sure.

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+                path, patched_path_string
+            );
+            return Ok(patched_path_string);
+        }
+
+        std::fs::create_dir_all(output.parent().ok_or_else(|| {
+            SipError::UnlikelyError("Failed to get parent directory".to_string())
+        })?)?;
+
+        return match shebang_target {
```
aviramha, line 251: no need return?

aviramha (changes requested):
Great progress

mirrord/sip/src/whitespace.rs
t4lz: Ok, like commit their license file to our repo?
Also apparently we can choose between MIT and Apache 2.0,  so MIT, right?

mirrord/cli/src/main.rs
t4lz: > just a warning and then "not loading" to the said binary

Do you mean letting the application run without mirrord?

mirrord/sip/src/whitespace.rs
aviramha: not as a file, just as a header to this file.

mirrord/cli/src/main.rs
aviramha: yeah, on your next feature that might be the situation when exec and it might not affect the work"iness"

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    /// Patches a binary to disable SIP.
+    /// Right now it extracts x64 binary from fat/MachO binary and patches it.
+    fn patch_binary<P: AsRef<Path>, K: AsRef<Path>>(path: P, output: K) -> Result<()> {
+        let data = std::fs::read(path.as_ref())?;
+        let binary_info = BinaryInfo::from_object_bytes(&data)?;
+
+        let x64_binary = &data[binary_info.offset..binary_info.offset + binary_info.size];
+        std::fs::write(output.as_ref(), x64_binary)?;
+        std::fs::set_permissions(output.as_ref(), Permissions::from_mode(0o755))?;
```
eyalb181, line 102: Is this line the actual patch? If so please add comment

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+                    std::fs::metadata(original_path)?.permissions(),
+                )?;
+                Ok(())
+            })
+            .ok_or(UnlikelyError("Can't read shebang anymore.".to_string()))?
+    }
+
+    const SF_RESTRICTED: u32 = 0x00080000; // entitlement required for writing, from stat.h (macos)
+
+    /// Extract shebang from file contexts.
```
eyalb181, line 133: Contexts? Or contents?

mirrord/sip/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    enum SipStatus {
+        /// The binary that ends up being executed is SIP protected.
+        /// The Option is `Some(SipStatus)` when this file is not a SIP binary but a file with a
+        /// shebang which points to a SIP binary (possibly via a chain of files with shebangs).
+        SomeSIP(PathBuf, Option<Box<SipStatus>>),
+        /// The binary that ends up being executed is not SIP protected.
+        NoSIP,
+    }
+
+    /// Determine status recursively, keep seen_paths, and return an error if there is a cycle.
```
eyalb181, line 183:
```suggestion
    /// Determine status recursively, keep seen_paths, and return an error if there is a cyclical reference.
```

mirrord/sip/src/lib.rs
t4lz: Contents, thanks!

mirrord/sip/src/lib.rs
t4lz: This line just adds permissions to the new file so that it is executable. Added a comment, changed the function doc, and also changed this line to give the new file the same permissions as the original binary instead of 755. Is that ok @aviramha? (asking because you wrote that function and maybe there's a reason to keep it as 755)

mirrord/sip/src/whitespace.rs
t4lz: Added.

mirrord/cli/src/main.rs
t4lz: Yeah on an `exec` hook it might make sense to keep running even if the patch failed, but here it's the main binary the user is trying to run, so if we can't load into there we're just not running at all. I think just running the binary without loading into it is not helpful for the user, as they can also do that without mirrord, and it's possible that they don't notice the warning and think mirrord is running and just don't understand why it's not working as they expect.

mirrord/sip/src/lib.rs
t4lz: LIke this?

mirrord/cli/src/main.rs
aviramha: yes though there might be situations where we are wrong - we detected SIP but it ain't really protected.

t4lz: I think the current flow is good:
Patch if SIP -->  run --> if failed because of rosetta show explanation

But should I add rosetta detection and change to:
Patch if SIP and x86/has rosetta --> run (even if there's SIP but we didn't patch for lack of rosetta).
? 

I don't know how reliably we can detect rosetta.

One other possible flow: 
Patch if SIP --> run --> if failed to run patched, emit error, then try to run original binary.
In this flow, like in the second one, we would probably sometimes be running just the plain binary without the layer being loaded, which I think we should not. But we would not have false negatives for rosetta because we would just do the patch and try to run it.

All in all I think the first flow is the best one, but let me know which one you like best.

aviramha: okay you got me. Let's see how users react.

aviramha (approved):

PR #707, opened 2022-11-08
[Merged by Bors] - Remove unwraps from mirrord-layer initialization.

meowjesty:
- Removes `unwrap`s from mirrord-layer initialization.

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -15,6 +15,7 @@ extern crate alloc;
 use std::{
     collections::{HashSet, VecDeque},
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    panic,
     path::PathBuf,
     sync::{LazyLock, OnceLock},
 };
@@ -172,16 +173,7 @@ fn mirrord_layer_entry_point() {
     // If we try to use `#[cfg(not(test))]`, it gives a bunch of unused warnings, unless you specify
     // a profile, for example `cargo check --profile=dev`.
     if !cfg!(test) {
-        if let Err(fail) = layer_pre_initialization() {
-            eprintln!(
-                r"
-            mirrord-layer: Encountered unrecoverable error during initialization!
-
-            >> Detailed error: {fail:#?}
-            "
-            );
-            std::process::exit(-1)
-        }
+        let _ = panic::catch_unwind(layer_pre_initialization);
     }
 }
 
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -143,8 +143,8 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
 }
 
 /// Loads mirrord configuration and applies [`nix_devbox_patch`] patches.
-fn layer_pre_initialization(args: Vec<String>) -> Result<(), LayerError> {
-    // let args = std::env::args().collect::<Vec<_>>();
+fn layer_pre_initialization() -> Result<(), LayerError> {
+    let args = std::env::args().collect::<Vec<_>>();
 
     println!("pre_initialization args {args:#?}");
 
@@ -176,8 +176,14 @@ fn mirrord_layer_entry_point() {
     // If we try to use `#[cfg(not(test))]`, it gives a bunch of unused warnings, unless you specify
     // a profile, for example `cargo check --profile=dev`.
     if !cfg!(test) {
-        let args = std::env::args().collect::<Vec<_>>();
-        let _ = panic::catch_unwind(|| layer_pre_initialization(args));
+        let _ = panic::catch_unwind(|| {
+            if let Err(fail) = layer_pre_initialization() {
+                match fail {
+                    LayerError::NoProcessFound => (),
+                    _ => std::process::exit(-1),
+                }
+            }
+        });
     }
 }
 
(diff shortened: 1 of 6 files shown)


mirrord/layer/src/lib.rs
```diff
@@ -141,36 +141,45 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
...
-                }
-            }
-            Err(err) => {
-                panic!("Failed to load config: {}", err);
-            }
-        }
+fn mirrord_layer_entry_point() {
+    if let Err(fail) = layer_pre_initialization() {
+        // TODO(alex) [mid] 2022-11-08: Improve these.
+        eprintln!("Failed initialization with {fail:#?}");
```
infiniteregrets, line 174: I would refrain from using std functions, because iirc it was segfaulting arm64e binaries. not sure if it still happens, but we'll have to test this thoroughly that it works with arm64e

mirrord/layer/src/lib.rs
```diff
@@ -141,36 +141,45 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
...
+
+    if should_load(given_process, skip_processes) {
+        layer_start(config);
+    }
+
+    Ok(())
+}
+
+/// The one true start of mirrord-layer.
+#[cfg(not(test))]
```
infiniteregrets, line 169: the only reason i used !cfg(test) was because it gave a bunch of unused code warnings, and you can also check with vscode it highlights everything yellow as unused code

infiniteregrets (commented):
👍

mirrord/layer/src/lib.rs
meowjesty: Oh I was wondering why it was like that. Guess this was fixed, as I'm not getting any unused warnings, plus vscode is highlighting everything correctly now.

It's kinda crappy because it's macro inside macro, but I'm reducing this function to be a one-liner anyway (or very close to it).

mirrord/layer/src/lib.rs
meowjesty: Noted, I'll leave only the panic.

mirrord/layer/src/lib.rs
infiniteregrets: sorry it is still there, https://github.com/metalbear-co/mirrord/actions/runs/3422857852/jobs/5700779282#step:15:153

we can either find a different solution to it or just keep using cfg!

mirrord/layer/src/lib.rs
```diff
@@ -141,36 +141,50 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
...
+fn mirrord_layer_entry_point() {
+    if let Err(fail) = layer_pre_initialization() {
+        eprintln!(
+            r"
+            mirrord-layer: Encountered unrecoverable error during initialization!
+
+            >> Detailed error: {fail:#?}
+            "
+        );
+        std::process::exit(-1)
```
infiniteregrets, line 180: dont think we can use eprint here again, ill have to check with an arm64e binary

infiniteregrets (commented):
👍

mirrord/layer/src/lib.rs
meowjesty: Back to `cfg!`!

mirrord/layer/src/lib.rs
meowjesty: No more prints!

mirrord/layer/src/lib.rs
```diff
@@ -141,36 +141,50 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
     }
 }
 
+/// Loads mirrord configuration and applies [`nix_devbox_patch`] patches.
+fn layer_pre_initialization() -> Result<(), LayerError> {
+    let args = std::env::args().collect::<Vec<_>>();
+    let given_process = args
```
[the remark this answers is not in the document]
meowjesty: Now `NoProcessFound` just ends up as `()`.

mirrord/layer/src/lib.rs
```diff
@@ -141,36 +142,53 @@ pub(crate) fn port_debug_patch(addr: SocketAddr) -> bool {
...
+
+    let mut config = std::env::var("MIRRORD_CONFIG_FILE")
+        .map_err(LayerError::from)
+        .map(PathBuf::from)
+        .and_then(|path| Ok(LayerFileConfig::from_path(&path)?))
+        .unwrap_or_default()
+        .generate_config()?;
+
+    nix_devbox_patch(&mut config);
+    let skip_processes = config.skip_processes.clone().map(VecOrSingle::to_vec);
```
infiniteregrets, line 162: maybe we can remove the nixos patch in this PR, see https://github.com/metalbear-co/mirrord/issues/711

infiniteregrets (approved):
LGTM

mirrord/layer/src/lib.rs
meowjesty: How do I test this?

aviramha: I opened it after I saw this PR and thought that we can remove it (irrelevant to this PR though, we fixed the root cause in another one)
I don't think it should be fixed as part of it. Also it's less prioritized

aviramha (approved):
Excellent.
bors r+

PR #709, opened 2022-11-08
[Merged by Bors] - Include missing job in ci-success.

t4lz:


meowjesty (approved):
👍 

p.s. do we add stuff like this to the changelog?

t4lz:
> p.s. do we add stuff like this to the changelog?

I think this raises the question: who is the changelog for?
If it's for users and people outside of the core team, I think that minor internal changes to the code or to the CI should not be included.
Changelog entries are also included in release notes and in discussions in github, and I think that if we include such details there those notes would be worthless.
If we want a good mechanism of tracking all the changes to our code, I think that meaningful commit messages is the right way.

t4lz:
> meaningful commit messages

which with `bors` means meaningful PR comments actually.

PR #716, opened 2022-11-10
[Merged by Bors] - Change bypass log to trace, add /home/ to exclude list.

t4lz:
Two changes that slipped away from #680.


eyalb181 (changes requested):

eyalb181 (approved):

aviramha (approved):
bors r+

PR #719, opened 2022-11-11
[Merged by Bors] - layer/agent: add control of network interface to use

aviramha:


t4lz (approved):
Looks good!

Question: Why not a CLI argument also?
General thought: looks like ephemeral container and job pod creation could share some code (but it should not hold back this PR).

t4lz:
Ok looks like the tests need to be adapted and the CHANGELOG formatted.

aviramha:
> Question: Why not a CLI argument also? 
Thanks! 
I don't see how it's relevant for most people so I prefer less arguments.

> General thought: looks like ephemeral container and job pod creation could share some code (but it should not hold back this PR).
Agree

t4lz (changes requested):
Regenerate schema please!
https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/lib.rs#L18

t4lz (approved):
Doesn't have to be in the config schema for now.

PR #722, opened 2022-11-12
[Merged by Bors] - Change extension to use mirrord.config, and use VS Code telemetry config

eyalb181:


vscode-ext/src/extension.ts
```diff
@@ -1,17 +1,41 @@
...
+        "env": true
+    }
+}
+`;
+
+const VS_CODE_DIR = function() {
+	let folders = vscode.workspace.workspaceFolders;
+	if (folders === undefined) {
+		throw new Error('No workspace folder found');
+	}
```
infiniteregrets, line 35: instead of throwing an exception, we should catch and spawn a notification

vscode-ext/src/extension.ts
```diff
@@ -1,17 +1,41 @@
...
+	return vscode.Uri.joinPath(folders[0].uri, '.vscode');
+}();
```
infiniteregrets, line 37: not sure what is going on here, could you please add a comment?

vscode-ext/src/extension.ts
```diff
@@ -24,90 +48,33 @@ function getK8sApi(): CoreV1Api {
...
+	let vsCodeDir = VS_CODE_DIR;
+	let fileUri = vscode.Uri.joinPath(vsCodeDir, 'mirrord.+(toml|json|y?(a)ml)');
+	let files = glob.sync(fileUri.fsPath);
+	if (files.length !== 0) {
+		let file = files[0];
+		return file;
+	}
+	else {
+		return '';
+	}
```
infiniteregrets, line 61: how about something like `file = (files.length !== 0) ?  files[0] : '';`

vscode-ext/src/extension.ts
```diff
@@ -24,90 +48,33 @@ function getK8sApi(): CoreV1Api {
...
-				} else if (setting.startsWith('Change namespace for impersonated pod')) {
-					globalContext.workspaceState.update('impersonatedPodNamespace', namespaceName);
-				}
-			});
-		}
-	});
+async function configFilePath() {
+	let vsCodeDir = VS_CODE_DIR;
+	let fileUri = vscode.Uri.joinPath(vsCodeDir, 'mirrord.+(toml|json|y?(a)ml)');
+	let files = glob.sync(fileUri.fsPath);
```
infiniteregrets, line 54: looks like we are trying to look for the config.. can we add a short comment here

vscode-ext/src/extension.ts
```diff
@@ -24,90 +48,33 @@ function getK8sApi(): CoreV1Api {
...
-					globalContext.workspaceState.update('agentNamespace', namespaceName);
-				} else if (setting.startsWith('Change namespace for impersonated pod')) {
-					globalContext.workspaceState.update('impersonatedPodNamespace', namespaceName);
-				}
-			});
-		}
-	});
+async function configFilePath() {
+	let vsCodeDir = VS_CODE_DIR;
+	let fileUri = vscode.Uri.joinPath(vsCodeDir, 'mirrord.+(toml|json|y?(a)ml)');
```
infiniteregrets, line 53: How about something like `^mirrord.*(toml|json|ya?ml)$`

- we can replace + with * to match zero or more, its the same as + in this case, but would make more sense
- `y?(a)ml` accepst `aml`
- we should use `$` in the end because otherwise it'll accept something like `mirrord.tomlshsjs` too

vscode-ext/src/extension.ts
```diff
@@ -24,90 +48,33 @@ function getK8sApi(): CoreV1Api {
...
 
+async function openConfig() {
+	let path = await configFilePath();
+	if (!path) {
+		path = vscode.Uri.joinPath(VS_CODE_DIR, 'mirrord.json');
+		await vscode.workspace.fs.writeFile(path, Buffer.from(DEFAULT_CONFIG));
+	}
+	vscode.workspace.openTextDocument(path).then(doc => {
+		vscode.window.showTextDocument(doc);
+	});
```
infiniteregrets, line 72: a .catch() would be nice

vscode-ext/src/extension.ts
```diff
@@ -139,6 +106,22 @@ async function checkVersion(version: string) {
...
 }
 
+async function parseNamespace() {
+	let filePath: string = await configFilePath();
+	let parsed;
+	if (filePath) {
+		const file = await (await vscode.workspace.fs.readFile(vscode.Uri.parse(filePath))).toString();
+		if (filePath.endsWith('json')) {
+			parsed = JSON.parse(file);
+		} else if (filePath.endsWith('yaml') || filePath.endsWith('yml')) {
```
infiniteregrets, line 116: is it possible to assign the return value of an if statement to a variable in typescript?

vscode-ext/src/extension.ts
```diff
@@ -139,6 +106,22 @@ async function checkVersion(version: string) {
...
+			parsed = YAML.parse(file);
+		} else if (filePath.endsWith('toml')) {
+			parsed = TOML.parse(file);
+		}
+	}
+	return parsed?.['target']?.['namespace'] || 'default';
```
infiniteregrets, line 122: can we handle exception if there is a parse error and let the user know?

infiniteregrets (commented):
Great work!

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 - `MIRRORD_AGENT_NETWORK_INTERFACE` environment variable/file config to let user control which network interface to use. Workaround for [#670](https://github.com/metalbear-co/mirrord/issues/670).
 
+### Changed
+
+- VS Code Extension: change extension to use a mirrord-config file for configuration
```
aviramha, line 16: I think this is major (breaking) no?

vscode-ext/src/extension.ts
```diff
@@ -1,17 +1,41 @@
...
+
+const DEFAULT_CONFIG = `
+{
+    "accept_invalid_certificates": false,
+    "feature": {
+        "network": {
+            "incoming": "mirror",
+            "outgoing": true
+        },
+        "fs": "write",
```
aviramha, line 25: read only

aviramha (changes requested):

vscode-ext/src/extension.ts
eyalb181: And then what? It's non-recoverable

vscode-ext/src/extension.ts
eyalb181: Actually not sure why we had the + before the file extension, copied it from @meowjesty 's PR. Fixed (also changed to accept e.g. foo.mirrord.yaml)

eyalb181: Note that it's glob, not regex. However, fixed to support both x.mirrord.ext and mirrord.ext

vscode-ext/src/extension.ts
eyalb181: When would it be called?

vscode-ext/src/extension.ts
eyalb181: Only with ternary operators

CHANGELOG.md
eyalb181: Keeping this a minor since it only affects the VSCode extension

vscode-ext/src/extension.ts
```diff
@@ -24,90 +49,28 @@ function getK8sApi(): CoreV1Api {
...
-				if (setting.startsWith('Change namespace for mirrord agent')) {
-					globalContext.workspaceState.update('agentNamespace', namespaceName);
-				} else if (setting.startsWith('Change namespace for impersonated pod')) {
-					globalContext.workspaceState.update('impersonatedPodNamespace', namespaceName);
-				}
-			});
-		}
-	});
+// Get the file path to the user's mirrord-config file, if it exists. 
+async function configFilePath() {
```
aviramha, line 53: We want it to be in a `.vscode` dir? I think it's less useful and would imagine it to be placed in our directory/root directory so it can be used across different IDEs/CLIs

vscode-ext/src/extension.ts
infiniteregrets: even though its not, we can let the user know what happened

vscode-ext/src/extension.ts
infiniteregrets:
```
vscode.workspace.openTextDocument(path).then(doc => {
		vscode.window.showTextDocument(doc);
	}).catch(err: Any => {
// foo
});
```
just meant that, similar to my comment before maybe we can handle any exceptions and let the user know?

vscode-ext/src/extension.ts
```diff
@@ -24,90 +49,28 @@ function getK8sApi(): CoreV1Api {
...
+	let vsCodeDir = MIRRORD_DIR;
```
infiniteregrets, line 54: maybe just a style thing, feel free to ignore. if MIRRORD_DIR is a const function, why not use it directly instead of assignment?

vscode-ext/src/extension.ts
eyalb181: VS Code does that for you when you throw an exception

aviramha (approved):
LGTM

vscode-ext/src/extension.ts
eyalb181: A. Let the user know what, though? If we don't know possible causes for this exception, then what can we write that would be more helpful than what's being written automatically?
B. Since we can't come up with any actual scenario where this would throw an exception, then I see no reason to clutter the codebase with a catch clause that would never be called. I prefer to take the risk that there's something I haven't thought about (this reasoning doesn't apply to the mirrord-agent/layer by the way).

infiniteregrets (commented):
On a side not, we need to update the readme of the vscode extension and add a better visual of it works now, can we do that here?

vscode-ext/src/extension.ts
infiniteregrets: okay, sounds good

vscode-ext/src/extension.ts
eyalb181: Totally. By the way, it's a const variable, not a const function - note that I call the function during assignment.

eyalb181:
> On a side not, we need to update the readme of the vscode extension and add a better visual of it works now, can we do that here?

Good call, I'll do it here

PR #723, opened 2022-11-13
[Merged by Bors] - CI: update homebrew formula on release

infiniteregrets:
don't know how accurately this would work, but this was just a rough idea how I thought we could update the formula. feel free to drop any suggestions

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -373,6 +373,7 @@ jobs:
   release_homebrew:
     needs: release_gh
     runs-on: ubuntu-latest
+    if: github.event_name != 'workflow_dispatch'
     steps:
       - uses: actions/checkout@v3
       - name: Get release version and hashes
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -369,3 +369,36 @@ jobs:
...
         uses: EndBug/latest-tag@latest
+
+  release_homebrew:
+    needs: release_gh
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get release version and hashes
+        run: |
+          echo "::set-output name=version::$(grep -m 1 version Cargo.toml | cut -d' ' -f3 | tr -d '\"' | cut -d'-' -f1)"
```
eyalb181, line 380: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Now seems like a good time to migrate to the env vars, before set-output becomes more prevalent

eyalb181 (changes requested):

eyalb181 (approved):

infiniteregrets:
hoping nothing is broken
bors r+

PR #727, opened 2022-11-14
[Merged by Bors] - Split KubernetesAPI to seperate package

DmitryDodzin:


mirrord/kube/src/error.rs
```diff
@@ -0,0 +1,48 @@
+use thiserror::Error;
+
+pub type Result<T, E = KubeApiError> = std::result::Result<T, E>;
+
+#[derive(Debug, Error)]
+pub enum KubeApiError {
+    #[error("mirrord-layer: Kube failed with error `{0}`!")]
```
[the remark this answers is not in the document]
DmitryDodzin: I'm thinking of renaming it from `KubeApi` to something else like `AgentApi` or something of this sort because it is not only for the kubernetes but connection too

mirrord/kube/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[dependencies]
+mirrord-config = { path = "../config"}
+mirrord-progress = { path = "../progress" }
+mirrord-protocol = { path = "../protocol" }
+
+actix-codec.workspace = true
+async-trait = "0.1"
+futures.workspace = true
+k8s-openapi.workspace = true
```
aviramha, line 28: I don't think we still need kube in the workspace?

mirrord/kube/src/api.rs
```diff
@@ -0,0 +1,114 @@
...
+        Api::default_namespaced(client.clone())
+    }
+}
+
+pub(crate) fn wrap_raw_connection(
+    stream: impl AsyncRead + AsyncWrite + Unpin + Send + 'static,
+) -> Result<(mpsc::Sender<ClientMessage>, mpsc::Receiver<DaemonMessage>)> {
+    let mut codec = actix_codec::Framed::new(stream, ClientCodec::new());
+
+    let (in_tx, mut in_rx) = mpsc::channel(1000);
```
aviramha, line 39: make const

mirrord/kube/src/api.rs
```diff
@@ -0,0 +1,114 @@
...
+    let (out_tx, out_rx) = mpsc::channel(1000);
+
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
+                Some(msg) = in_rx.recv() => {
```
aviramha, line 45: I think it's worth covering the specific case of None from each side to know who disconnected first.

aviramha (commented):
Looking good! See inline comments

mirrord/kube/Cargo.toml
DmitryDodzin: e2e tests also use kube and k8s-openapi so we can keep them but ill remove k8s-openapi from mirrord-layer

aviramha: put it in the specific cargo.tomls then

mirrord/kube/src/api.rs
```diff
@@ -0,0 +1,129 @@
...
+                msg = in_rx.recv() => {
+                    match msg {
+                        Some(msg) => {
+                            if let Err(fail) = codec.send(msg).await {
+                                error!("Error sending client message: {:#?}", fail);
+                                break;
+                            }
+                        }
+                        None => {
+                            error!("initiated disconnect from agent");
```
aviramha, line 56: not error

mirrord/kube/src/api.rs
```diff
@@ -0,0 +1,129 @@
...
+                msg = in_rx.recv() => {
+                    match msg {
+                        Some(msg) => {
+                            if let Err(fail) = codec.send(msg).await {
+                                error!("Error sending client message: {:#?}", fail);
+                                break;
+                            }
+                        }
+                        None => {
+                            warn!("initiated disconnect from agent");
```
aviramha, line 56: sorry shouldn't be so vague - this is info/trace as that's bound to happen when process ends ;)

aviramha (approved):

PR #733, opened 2022-11-15
[Merged by Bors] - Make connection errors nicer.

t4lz:
- Only recommend `--accept-invalid-certificates` if not already set.
  - Exit gracefully instead of `panic`king: this means also the user's
    application is terminated, not just mirrord's layer.
  - Wrap `graceful_exit` code in a block so that it can be used as a
    whole `match` arm, add (unreachable) `panic` at the end, so that it
    can be used where the return type is `!`.

### DIFF
--- a/mirrord/layer/src/macros.rs
+++ b/mirrord/layer/src/macros.rs
@@ -91,17 +91,18 @@ macro_rules! hook_symbol {
 
 #[macro_export]
 macro_rules! graceful_exit {
-    ($($arg:tt)+) => {
+    ($($arg:tt)+) => {{
         eprintln!($($arg)+);
         graceful_exit!()
-    };
-    () => {
+    }};
+    () => {{
         nix::sys::signal::kill(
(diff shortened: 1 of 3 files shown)


mirrord/layer/src/macros.rs
```diff
@@ -91,17 +91,18 @@ macro_rules! hook_symbol {
 
 #[macro_export]
 macro_rules! graceful_exit {
-    ($($arg:tt)+) => {
+    ($($arg:tt)+) => {{
```
eyalb181, line 94: Missing something, why the double parentheses?

t4lz: So that a call to that macro can be used as its own arm on `match`. Without it we would have to wrap calls to it in curly brackets, because without them we would be putting multiple statements directly in a `match` arm which is not allowed.

eyalb181 (approved):

PR #741, opened 2022-11-16
[Merged by Bors] - Randomize library identifier at compile time to avoid file duplication

eyalb181:


mirrord/cli/src/main.rs
```diff
@@ -48,36 +48,25 @@ use mac::temp_dir;
...
+    let file_name = format!("{}-libmirrord_layer.{extension}", const_random!(u64));
     let file_path = match dest_dir {
         Some(dest_dir) => std::path::Path::new(&dest_dir).join(file_name),
         None => temp_dir().as_path().join(file_name),
     };
-    let mut file = File::create(&file_path)
-        .with_context(|| format!("Path \"{}\" creation failed", file_path.display()))?;
-    let bytes = include_bytes!(env!("MIRRORD_LAYER_FILE"));
-    file.write_all(bytes).unwrap();
+    if !file_path.exists() {
```
t4lz, line 62: Should we maybe, only when creating a new layer file because it does not yet exist in this version, also delete the last one? That way we avoid a filling up the user's drive at a later point, but they still get a quick run without copying files most of the time - when they are rerunning the same mirrord binary.

aviramha: I think it's hard to solve and also not very beneficial - how many updates does a user do?

aviramha (approved):
lgtm

mirrord/cli/src/main.rs
eyalb181: Why's it hard to solve? Don't we just remove all *-libmirrord_layer.{extension} from the temp_dir?

eyalb181: Nevermind merging

PR #743, opened 2022-11-16
[Merged by Bors] - Hook `execve` and patch SIP before executing.

t4lz:
Part of #649.

- Add the temp dir to excluded in the cli.
- If SIP patch fails (on something that is not `FileNotFound`) warn, but keep running.
  - FileNotFound would be very common, as apparently `!#/usr/bin/env bash` tries to execute a bunch of non-existing `bash` binaries, and that is a very common shebang.
- Add integration test with a bash script, to verify the layer can now hook SIP binaries on mac.
  - Add the useful stuff in the new test as public functions in `common`.
- Use inner attributes (`#![cfg(...)]`) and target specific dependencies in order to compile the SIP code conditionally.
- Add `unwrap_or` to `Detour`.
- Make a public function in order to reuse rawish string code from the file hooks.


mirrord/layer/src/detour.rs
```diff
@@ -201,6 +205,38 @@ impl<S> Detour<S> {
...
+                trace!("Detour failed with: {:#?}, continuing despite of error.", e);
+                default
+            }
+        }
+    }
+}
+
+/// Allows you to call `.collect()` on an iterator of Detours and get a Detour of a Vec.
+impl<S> FromIterator<Detour<S>> for Detour<Vec<S>> {
+    fn from_iter<T: IntoIterator<Item = Detour<S>>>(iter: T) -> Detour<Vec<S>> {
```
meowjesty, line 233: Maybe we could implement this with [`Iterator::scan`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan) like Rust's [`Result::from_iter`](https://doc.rust-lang.org/src/core/result.rs.html#2068) is saying in a comment?

mirrord/layer/tests/apps/env_bash_cat.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# With this shebang what happens when this script is executed normally:
+# 1. env is executed.
+# 2. env executes bash (calls execve)
+# 3. bash executes (calls execve) the binaries that are called in the script. Here: cat.
+
+# With mirrord's SIP bypassing mechanism:
+# 1. We create a SIP-free version of env and a version of this script that has the SIP-free env in its shebang.
+# 2. We call the that version of the script instead of the original.
```
meowjesty, line 9:
```suggestion
# 2. We call that version of the script instead of the original.
```

mirrord/sip/src/lib.rs
```diff
@@ -287,15 +304,55 @@ mod main {
...
+        // Important: do also if the binary has no SIP, so that a temp directory is set, added to
+        // ignored files, and saved in an environment variable, so that if the binary execs SIP
+        // binaries the patched versions are created in a directory that is already added to the
+        // excluded files. (It's currently not possible to add excluded paths when the layer is
+        // already running, and the first call to this function is from the cli.)
+        let tmp_dir = get_tmp_dir()?;
+
         if let SipStatus::SomeSIP(path, shebang_target) = get_sip_status(&binary_path)? {
-            return patch_some_sip(&path, shebang_target);
+            debug!("Using temp dir: {:?} for sip patches", &tmp_dir);
```
meowjesty, line 351:
```suggestion
            trace!("Using temp dir: {:?} for sip patches", &tmp_dir);
```

Let's try leaving everything useful only to us as `trace` when it comes time to merge.

meowjesty (changes requested):
Nits and picks

mirrord/sip/src/lib.rs
t4lz: Done. What's `debug` for then?

mirrord/layer/src/detour.rs
t4lz: Sounds good, but actually this is not used anywhere yet, I implemented it for something that didn't make it to the final version. I thought if I already have the code I can leave it there in case it's useful for someone in the future. So how about if it's not correct as it is I just delete it and we implement it correctly if we need it in the future?

mirrord/sip/src/lib.rs
meowjesty: Hmm I would leave `debug` for when we're in the middle of developing something. I'm trying to use it like that, then putting everything back to `trace` once I'm done.

Just an idea though, we have not really discussed this type of stuff (only that we should avoid leaking higher tracing levels to the user with information they don't need).

mirrord/layer/src/detour.rs
t4lz: deleted.

aviramha:
nit: you can remove the workaround you added for python tests on macOS (where you resolve the interpreter). those should work now ;)

t4lz:
> nit: you can remove the workaround you added for python tests on macOS (where you resolve the interpreter). those should work now ;)

Requires calling `sip_patch` in those tests explicitly too, is that fine?

aviramha:
> > nit: you can remove the workaround you added for python tests on macOS (where you resolve the interpreter). those should work now ;)
> 
> Requires calling `sip_patch` in those tests explicitly too, is that fine?

Ah those don't use the cli. maybe it's better it uses it?
Leave it for another issue.

mirrord/layer/src/lib.rs
```diff
@@ -151,6 +155,22 @@ fn layer_pre_initialization() -> Result<(), LayerError> {
...
+    // to the file filter after it was constructed).
+    #[cfg(target_os = "macos")]
+    let _tmp_dir = get_tmp_dir()
+        .inspect_err(|err| {
+            trace!(
+                "Getting temp dir failed with {:?} in layer_pre_initialization",
+                err
+            )
+        })
+        .unwrap_or_default();
```
t4lz, line 172: Added this part now, sorry @meowjesty  for adding stuff after the first review. But I think it's a small change that brings some value.

meowjesty (approved):
🐻 👍

PR #751, opened 2022-11-17
[Merged by Bors] - fix: don't ignore fd < 2 on writes now that we use real 'fd' backend

### LINKED ISSUE #750: stdin/err/out fds behavior
We have a patch on `write` hook that checks if fd < 2 then runs it locally, though the check shouldn't be required since we use `shm_open` to allocate a real free fd, so if fd `0` is used, and it's managed by us, that should be okay..

aviramha:
closes #750


DmitryDodzin:
Won't this cause an infinite loop with trace?

aviramha:
> Won't this cause an infinite loop with trace?

Nice catch. Fixed

aviramha:
bors cancel

PR #752, opened 2022-11-17
[Merged by Bors] - Don't unwrap status.code

t4lz:
On Unix, `code()` will return `None` if the process was terminated by a signal, so get the raw wait status which has the wait status code if there is no exit status code.


aviramha (approved):
bors r+

PR #755, opened 2022-11-17
Drop fd

aviramha:
To whom it might concern, when reviewing this we need to make sure that I didn't make fds that aren't really owned into ownedfds, and didn't take ownership of fds that I shouldn't ;)


meowjesty (approved):
👍 

Hopefully I'm not missing a borrow turn owned (or vice-versa).

PR #758, opened 2022-11-17
[Merged by Bors] - fix: possible cases where we don't close fds correctly

aviramha:


mirrord/layer/src/lib.rs
```diff
@@ -606,21 +606,18 @@ pub(crate) unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
...
+    let res = FN_CLOSE(fd);
+    if SOCKETS.lock().unwrap().remove(&fd).is_none() && *enabled_file_ops
         && let Some(remote_fd) = OPEN_FILES.lock().unwrap().remove(&fd) {
         let close_file_result = file::ops::close(remote_fd);
 
-        close_file_result
+        let _ = close_file_result
             .map_err(|fail| {
                 error!("Failed closing file with {fail:#?}");
                 -1
```
t4lz, line 617: This could be deleted and the map simplified a bit (no block), right?

mirrord/layer/src/lib.rs
```diff
@@ -606,21 +606,18 @@ pub(crate) unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
         .get()
         .expect("Should be set during initialization!");
 
-    if SOCKETS.lock().unwrap().remove(&fd).is_some() {
-        FN_CLOSE(fd)
-    } else if *enabled_file_ops
+    let res = FN_CLOSE(fd);
```
t4lz, line 609: Why do we want to call close also if the `fd` is remote and does not exist locally?

mirrord/layer/src/lib.rs
```diff
@@ -606,21 +606,18 @@ pub(crate) unsafe extern "C" fn close_detour(fd: c_int) -> c_int {
...
+    if SOCKETS.lock().unwrap().remove(&fd).is_none() && *enabled_file_ops
         && let Some(remote_fd) = OPEN_FILES.lock().unwrap().remove(&fd) {
         let close_file_result = file::ops::close(remote_fd);
 
-        close_file_result
+        let _ = close_file_result
```
t4lz, line 614: Why shouldn't we pass the remote close result to the local process anymore?

t4lz (commented):
Could you please explain the reasons for these changes?

mirrord/layer/src/lib.rs
aviramha: Because each remote file is backed by a real fd locally and we don't want to leak it.

mirrord/layer/src/lib.rs
aviramha: It's not an easy decision, best would be to return 0 if both are okay or return the invalid one, but what happens when both fail? Anyhow close is usually used as a best effort so I don't see too much value digging it

mirrord/layer/src/lib.rs
t4lz: Oh so until now we did, right?

t4lz (changes requested):
Looks like e2e tests aren't passing.

mirrord/layer/src/lib.rs
aviramha: Thanks

t4lz (approved):

PR #759, opened 2022-11-18
[Merged by Bors] - Intercept argv and remove mirrord's temp dir

t4lz:
When an executable script `my_script` with a shebang `#!/usr/bin/env some_interpreter` is executed, it resolves to 
```bash
/usr/bin/env some_interpreter my_script
```
So currently if we patch it and save a patch version in `mirrord-tmp/`, running `/mirrord-tmp/my_script` resolves to 
```bash
/usr/bin/env some_interpreter /mirrord-temp/my_script
```
so what then happens when we run `my_script` with the mirrord CLI (if `some_interpreter` is a binary executable that does not have SIP):
1. We read `my_script` and its shebang.
2. We create a version of `my_script` in `/mirrrord-tmp/path/to/my_script` with the shebang `#!/mirrord-tmp/usr/bin/env some_interpreter`, and create a SIP-free version of `env` in `/mirrord-tmp/usr/bin/env`.
3. We execute `/mirrrord-tmp/path/to/my_script` which resolves to `/mirrord-tmp/usr/bin/env my_interpreter /mirrrord-tmp/path/to/my_script`
4. patched `env` is executed with loaded layer.
5. `env` calls `execve` with path `my_interpreter` and argv: `["my_interpreter", "/mirrrord-tmp/path/to/my_script"]`.

This is a problem when running `node` as the interpreter (`#!/usr/bin/env node`), since node determines the location of modules relatively to the executed file, which it receives as an argument (here argv[1]). In our example node would search for modules relative to `/mirrrord-tmp/path/to/my_script`, instead of relative to `/path/to/my_script`.
This PR suggests a mitigation that problem: 
In the `execve` hook: go over the arguments in `argv` and strip the path to our temp directory from them if it's there.

That way node reads the original script, not the patched one, and can find the required modules.

### DIFF
--- a/mirrord/layer/src/exec.rs
+++ b/mirrord/layer/src/exec.rs
@@ -1,16 +1,22 @@
 #![cfg(target_os = "macos")]
 
-use std::ffi::{CStr, CString};
+use std::{
+    env,
+    ffi::{CStr, CString},
+};
 
 use frida_gum::interceptor::Interceptor;
 use libc::{c_char, c_int};
 use mirrord_layer_macro::hook_guard_fn;
-use mirrord_sip::{sip_patch, SipError};
-use tracing::{trace, warn};
+use mirrord_sip::{sip_patch, SipError, TMP_DIR_ENV_VAR_NAME};
+use null_terminated::Nul;
+use tracing::{debug, log::error, warn};
 
 use crate::{
     detour::{
-        Bypass::{ExecOnNonExistingFile, NoSipDetected},
+        Bypass::{
+            ExecOnNonExistingFile, NoSipDetected, NoTempDirInArgv, TempDirEnvVarNotSet, TooManyArgs,
+        },
         Detour,
         Detour::{Bypass, Error, Success},
     },
@@ -18,6 +24,8 @@ use crate::{
     file::ops::str_from_rawish,
     replace,
 };
+
+static MAX_ARGC: usize = 16;
 
 pub(crate) unsafe fn enable_execve_hook(interceptor: &mut Interceptor) {
     let _ = replace!(interceptor, "execve", execve_detour, FnExecve, FN_EXECVE);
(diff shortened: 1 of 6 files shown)


t4lz:
With this change, the user needs to make sure the script they are running is excluded from FS hooks.
Before this change node would read the patched version of the script, which was in a directory that the sip patch code added to excluded directories. So if you would put your script in an included path, you could still run it on mac because of this side effect of the sip patch. Now if you try to run a script that is in a path that is included in FS ops, it will fail both on linux and mac. This seems like correct behaviour to me. To ensure good user experience I suggest we add the path of the executable (binary or script) the user is running to the excluded files at the start of the run (in a separate PR).

t4lz:
No changelog entry because the bug was not yet released.

mirrord/layer/src/exec.rs
```diff
@@ -19,6 +25,8 @@ use crate::{
     replace,
 };
 
+static MAX_ARGC: usize = 16;
```
aviramha, line 28: why 16?

aviramha:
Need to add it to `cli`'s exec as well.

mirrord/layer/src/exec.rs
t4lz: Just seemed high enough to me. When there are more args we bypass the argv interception.
Do you think it should be greater? How much?

aviramha: 254 and I think it needs to be a const not static

t4lz: cool, thanks!

t4lz:
> Need to add it to `cli`'s exec as well.

Now using the original executable as `argv[0]`, but I wouldn't change the other args to `execvp` as they are supplied by the user. Anyway I think it does not affect the actual issue we're dealing with.

The theoretical remaining questions is what happens in an imaginary situation where node itself is SIP (I think it's usually not) and we run a script at `/path/to/script.js` that starts with:
```
#!/bin/node
```

We would patch node at `/mirrord-temp/bin/node` and create `/mirrord-temp/path/to/script.js` with:
```
#!/mirrord-temp/bin/node
```

Then call `execvp` with:
`path`: `"/mirrord-temp/path/to/script.js"`
`argv`: `["/path/to/script.js"]`

Which would result in the execution of `/mirrord-temp/bin/node` with `/mirrord-temp/path/to/script.js` (even though we have `/path/to/script.js` in `argv[0]`). If there is an extra call to `execve` there, it's before we have any hooks.
Node would then read and interpret `/mirrord-temp/path/to/script.js` and would not find modules that are `required` with a relative path.
I don't know how important that remaining problem is, but if it is important, a possible solution I can think of is:
Instead of running the patched script and letting shebang resolving occur naturally, we parse the shebang and execute 
`interpreter script` directly. In the example from before that would mean call `execvp` with:
`path`: `"/mirrord-temp/bin/node"` (patched binary interpreter)
`argv`: `["/bin/node", "/path/to/script.js"]` (original script)

I'm not sure what happens with arguments if we take this approach. Like is everything guaranteed to go as we would hope if we run `/usr/bin/env node script.js arg-a arg-b arg-c` and more generally `/usr/bin/env interpreter script arg-a arg-b arg-c` or `/my/interpreter script arg-a arg-b arg-c`

aviramha (approved):
bors r+

Issue #756, opened 2022-11-17
mirrord crashes when running `bash -i`


eyalb181:
### Bug Description

The following error is displayed (full logs attached under Additional Info):
`malloc: unknown:0: assertion botched
malloc: block on free list clobbered
Aborting..../.workspace/runtime/src/shell.sh: line 6: 383940 Aborted                 (core dumped)`

It's not a memory thing though.

### Steps to Reproduce

Run `bash -i` with mirrord

### Your operating system and version

TBD (some Linux)

### Local process

bash

### Additional Info

[mirrord-bash.log.txt](https://github.com/metalbear-co/mirrord/files/10034026/mirrord-bash.log.txt)

aviramha:
It doesn't happen when running just bash, without `-i`. I'm not sure what that even means in this context.
It seems to happen 80% of the time, attaching relevant logs.
https://gist.github.com/aviramha/9d7a003b1c177eab3a0ba058c14eb9fa

aviramha:
I investigated deeper, and found out that it seems fixing this would require quite big change (it seems like a ctor/tokio/parking_lot integration issue, where ctor shouldn't be used to create threads.. and changing this behavior might cause new problems).
I tend to believe that this use case is very niche, especially that it doesn't happen with regular bash + without specifying `-i`. @mentos1386 please let us know how important this to you.

aviramha:
I think that #763 solves it/reduces the odds of the crash, yay? :)

aviramha:
We just released 3.11. Please let us know if that fixed it for you.

PR #763, opened 2022-11-19
[Merged by Bors] - reduce tokio features

aviramha:
Seems that it also solves #756 (since it drops `parking_lot`)
I think tradeoff is okay. (performance impact (5x but on a spot that doesn't really matter..))
also binary size should improve


infiniteregrets (approved):
LGTM

PR #768, opened 2022-11-21
[Merged by Bors] - Operator API

DmitryDodzin:
Operator Protocol and Client that implements AgentManagment and can be used instead of KubernetesAPI to connect to mirrord-agent


eyalb181 (changes requested):

mirrord/operator/src/client.rs
```diff
@@ -0,0 +1,130 @@
...
+use tokio::{
+    net::{TcpStream, ToSocketAddrs},
+    sync::mpsc,
+};
+
+use crate::protocol::{AgentInitialize, OperatorCodec, OperatorRequest, OperatorResponse};
+
+pub struct OperatorApi<T: ToSocketAddrs> {
+    addr: T,
+    agent: AgentConfig,
```
eyalb181, line 18: Adding to my previous comment - seems the only reason the operator API is even aware of the Agent is the ability to configure the Agent from the layer. This isn't required though. Agents should be configured at the operator level, not on a per-agent basis through the layer.

mirrord/config/src/lib.rs
```diff
@@ -116,6 +116,9 @@ pub struct LayerConfig {
     /// Controls mirrord features, see [`feature::FeatureFileConfig`].
     #[config(nested)]
     pub feature: FeatureConfig,
+
+    #[config(env = "MIRRORD_OPERATOR")]
+    pub operator: Option<String>,
```
aviramha, line 121: I think we should check on start if operator exist and just use that instead of adding a flag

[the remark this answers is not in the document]
DmitryDodzin: Right now its address to the operator, with the user being required to port-forward for now it's gonna be the port that is port-forwarded

mirrord/layer/src/connection.rs
```diff
@@ -59,6 +60,18 @@ pub(crate) async fn connect(
             .create_connection(stream)
             .await
             .unwrap_or_else(|err| handle_error(err, config))
+    } else if let Some(addr) = &config.operator {
+        let operator_api = OperatorApi::new(addr, config.target.clone());
+
+        let stream = operator_api
```
eyalb181, line 66: Why are create_agent and create_connection two separate functions?

mirrord/operator/src/protocol.rs
```diff
@@ -0,0 +1,158 @@
...
+use bytes::{Buf, BufMut, BytesMut};
+use mirrord_config::target::TargetConfig;
+use mirrord_protocol::{ClientMessage, DaemonMessage};
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct AgentInitialize {
+    target: TargetConfig,
+}
+
+impl AgentInitialize {
```
eyalb181, line 18: Why are there still agent references in the protocol?

mirrord/config/src/lib.rs
DmitryDodzin: Adding it in different PR this one already too big

t4lz (approved):
Looks good. Left a couple of optional comments.

mirrord/progress/src/lib.rs
```diff
@@ -106,6 +106,16 @@ pub trait Progress: Sized {
     fn set_done(&mut self, msg: Option<&str>, failure: bool);
 }
 
+pub struct NoProgress;
```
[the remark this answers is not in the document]
DmitryDodzin: Technically no mostly using it in operator, I wanted to implement progress on `()` but then decided to make it a custom type to prevent any future misunderstandings

mirrord/operator/src/protocol.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    pub fn target(&self) -> &TargetConfig {
+        &self.target
+    }
+}
+
+impl Encode for Handshake {
+    fn encode<E: bincode::enc::Encoder>(&self, encoder: &mut E) -> Result<(), EncodeError> {
+        let target = serde_json::to_vec(&self.target).expect("target not json serializable");
```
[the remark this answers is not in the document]
DmitryDodzin: It's a bit strange, I can map it into EncodeError:::Other with a string as value, no way to propagate a custom error like FromStr or similar traits

mirrord-schema.json
```diff
@@ -56,6 +56,12 @@
         }
       ]
     },
+    "operator": {
```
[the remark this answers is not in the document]
DmitryDodzin: Namespace will be added in #798

aviramha (approved):

eyalb181 (approved):

PR #771, opened 2022-11-21
Filter incoming http (1) traffic.

meowjesty:


mirrord/agent/src/sniffer.rs
```diff
@@ -293,21 +299,26 @@ impl TCPConnectionSniffer {
...
 
-    #[tracing::instrument(level = "trace")]
-    pub async fn new(
+    /// Creates and prepares a new [`TcpConnectionSniffer`] that uses BPF filters to capture network
+    /// packets.
+    ///
+    /// The capture uses a network interface specified by the user, if there is none, then it tries
+    /// to find a proper one by starting a connection. If this fails, we use "eth0" as a last
+    /// resort.
+    #[tracing::instrument(level = "info")]
```
aviramha, line 316: trace

mirrord/agent/src/sniffer.rs
```diff
@@ -330,16 +341,7 @@ impl TCPConnectionSniffer {
...
-        receiver: Receiver<SnifferCommand>,
-        pid: Option<u64>,
-        network_interface: Option<String>,
-        cancel_token: CancellationToken,
-    ) -> Result<(), AgentError> {
-        let sniffer = Self::new(receiver, pid, network_interface).await?;
-        sniffer.run(cancel_token).await
-    }
-
+    #[tracing::instrument(level = "info", skip(self, sender))]
```
aviramha, line 344: trace

mirrord/agent/src/sniffer.rs
```diff
@@ -441,27 +447,25 @@ impl TCPConnectionSniffer {
...
-                warn!("failed to send message to client {}", client_id);
+                warn!("Failed sending message to client {}", client_id);
                 let _ = self.handle_client_closed(*client_id);
                 err
             })?;
         }
         Ok(())
     }
 
+    #[tracing::instrument(level = "info", skip(self), fields(packet = %String::from_utf8_lossy(&eth_packet)))]
```
aviramha, line 467: trace

mirrord/agent/src/steal.rs
```diff
@@ -242,30 +250,37 @@ impl StealWorker {
...
+
         Ok(())
     }
 
+    #[tracing::instrument(level = "debug", skip(self))]
     pub async fn handle_client_message(&mut self, message: LayerTcpSteal) -> Result<()> {
         use LayerTcpSteal::*;
         match message {
-            PortSubscribe(port) => {
+            PortSubscribe((port, traffic_filter)) => {
```
aviramha, line 261: I'm thinking if it needs to be two messages or whole new one as we'll need to update it "on the fly" in operator mode.
Okay I think it's fine for now - if port already subscribed it will just update the filter.

aviramha: (no need to add it, just thinking out loud for future, maybe add a comment that that's how it will work)

mirrord/agent/src/steal.rs
```diff
@@ -278,20 +293,42 @@ impl StealWorker {
...
+                //
+                // Now I just need to parse this as an "HTTP" request and then check for the filter?
+                //
+                // 1. Check if `bytes > n`, where `n` is some small number for a small HTTP packet.
+                // 2. Try to parse HTTP message.
+                //
+                // Read on HTTP to get a better understanding of whats going on.
+                //
+                // ADD(alex) [high] 2022-11-16: Start dealing with HTTP/1.1 first.
+                let data_debug = String::from_utf8_lossy(&data.bytes);
```
aviramha, line 307: not sure if intentional but wouldn't convert it..

mirrord/config/src/network.rs
```diff
@@ -47,6 +47,23 @@ pub struct NetworkConfig {
     /// Resolve DNS via the remote pod.
     #[config(env = "MIRRORD_REMOTE_DNS", default = "true")]
     pub dns: bool,
+
+    // TODO(alex) [mid] 2022-11-17: Improve these docs.
```
aviramha, line 51: question - would we be able to generate mutual exclusive regexes to detect if more than 1 filter includes hte other? (for operator)

aviramha: seems we can do regex -> dfa -> reduce

aviramha (changes requested):
Great progress, but it's a bit too naive at the moment.

1. It assumes packet == request, which isn't the case.
2. It writes the data as is to the local connection, I was thinking we should wrap it and make a new message that defines a HTTP1 Request, then send it to the client to invoke it.
3. What happens on skip? Right now it looks like the data is just dropped? We need to be a transparent proxy when it's a request we don't care about.

I think the best implementation is to leverage hyper server, forward the request to it then handle the request using it's framework.
To be more precise - check packet, see if it's HTTP 1.1, if it is, forward it to a hyper server to handle. else bypass.

Now thinking about it, because of the feature being conditional, maybe we should have it as a different message.

aviramha:
Another thought -  I see many welcome quality of life improvements here. It's a bit distracting so it will be valuable to sent those as separate PR.

meowjesty:
Closing this in favor of #797 as this one had a bunch of unrelated changes to the feature.

PR #782, opened 2022-11-23
[Merged by Bors] - clippy macOS

aviramha:


.github/workflows/ci.yaml
```diff
@@ -68,8 +68,26 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       # Needed for agent
       - run: cargo fmt --all -- --check
+      - run: rustup target add --toolchain nightly aarch64-apple-darwin
```
t4lz, line 71: Does this have an affect? `--all-targets` is not needed for that?

.github/workflows/ci.yaml
```diff
@@ -68,8 +68,26 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: rustfmt, clippy
+          target: aarch64-apple-darwin
+      - uses: Swatinem/rust-cache@v1
+      # only mirrord, mirrord-sip contain macOS specific, so run clippy on that.
```
t4lz, line 87: Would it be notably slower to run it on all the code? That way we wouldn't have to remember to add stuff to here if we add MacOS-specific code in the future.

mirrord/cli/src/main.rs
```diff
@@ -9,8 +9,6 @@ use anyhow::{anyhow, Context, Result};
 use clap::Parser;
 use config::*;
 use const_random::const_random;
-#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
-use errno;
```
t4lz, line 13: How does that work if we use `errno` there: https://github.com/metalbear-co/mirrord/blob/a76adadaca201b4e9408e85df2ad8bea10728987/mirrord/cli/src/main.rs#L234?

t4lz (approved):

mirrord/cli/src/main.rs
aviramha: `errno:Errno` is "fully qualified" name. In Rust `use errno` usually has no particular meaning

.github/workflows/ci.yaml
aviramha: Problem is not all code compiles for macOS (agent)..

.github/workflows/ci.yaml
aviramha: oops, my bad.

.github/workflows/ci.yaml
aviramha: (It can't compile because of dependencies)

PR #785, opened 2022-11-23
[Merged by Bors] - File filter change

aviramha:
It shouldn't be a breaking change, so need to make sure of that.


mirrord/layer/Cargo.toml
```diff
@@ -41,7 +41,8 @@ tokio-util.workspace = true
 thiserror.workspace = true
 trust-dns-resolver.workspace = true
 rand = "0.8"
-fancy-regex = { version = "0.10" }
+fancy-regex = "0.10"
+regex = "1"
```
meowjesty, line 45: `regex` doesn't support some useful regex operations. One that comes to mind and triggered me changing this was "Skip /etc/*, except if it's /etc/foo.txt".

meowjesty: But if we're ok with losing functionality, then `regex` is fine.

mirrord/layer/src/file/filter.rs
```diff
@@ -85,103 +97,154 @@ pub(crate) static FILE_FILTER: OnceLock<FileFilter> = OnceLock::new();
...
-            FileFilter::Exclude(exclude) if !exclude.is_match(text).unwrap() => Detour::Success(()),
-            _ => Detour::Bypass(op()),
+        match &self.old_filter {
+            OldFilter::Include(include) if include.is_match(text) => {}
+            OldFilter::Exclude(exclude) if !exclude.is_match(text) => {}
+            OldFilter::Nothing => {}
+            _ => return Detour::Bypass(op()),
+        }
+
+        if some_regex_match!(&self.read_write, text) {
```
meowjesty, line 214: What do you think of something like this (to drop the macro):

```suggestion
        if self.read_write.map(|regex| regex.is_match(text)).unwrap_or_default()
```

mirrord/config/src/fs/advanced.rs
```diff
@@ -14,29 +14,36 @@ use crate::{
...
-/// - `MIRRORD_FILE_OPS` and `MIRRORD_FILE_RO_OPS`;
-/// - `MIRRORD_FILE_FILTER_INCLUDE` and `MIRRORD_FILE_FILTER_EXCLUDE`;
-///
+/// What is the default behavior for file operations:
+/// 1. read - Read from the remote file system (default)
+/// 2. write - Read/Write from the remote file system.
+/// 3. local - Read from the local file system.
+/// 4. disabled - Disable file operations.
+/// 
+/// Besides the default behavior, user can specify behavior for specific regex patterns. Case insensitive.
```
meowjesty, line 23: Why be case insensitive?

meowjesty (changes requested):
Love that the docs are pretty spot on.

Possible breaking change due to drop in functionality from going back to `regex` crate.

As a not fan of `if`s and `else`s, maybe the whole thing in `continue_or_bypass_with` could be a chain of `map.or_else`s, but it's fine as-is right now.

mirrord/layer/Cargo.toml
aviramha: doesn't `|` cover it?

aviramha: nvm

aviramha: I went with the regex because of regexset making way more sense.

mirrord/config/src/fs/advanced.rs
meowjesty: If we're letting the user pick their own regexes, then I think the least surprising thing would be for us to do exactly what they said (if they want case insensitive then they specify as `/etc/[lL]ala`).

I say this because I can imagine this being annoying if i wanted to differentiate between both, and saw that the tool doesn't.

mirrord/layer/Cargo.toml
meowjesty: The easiest solution to the example I gave involves using "lookaheads", which `regex` doesn't support.

mirrord/layer/src/file/filter.rs
aviramha: Applied it as a macro (as it's still very repetitive)

meowjesty: I would go for a closure instead of a macro there, as it is a small block of code.

The closure could take both the regex and `text` as references, and is clearer than a macro for this kind of stuff, imo.

Your approach is fine though, your pick.

mirrord/cli/src/config.rs
```diff
@@ -26,6 +26,14 @@ pub(super) enum Commands {
         .multiple(true)
         .args(&["target", "config-file"]),
 ))]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+enum FsMode {
+    ReadWrite,
+    Read,
+    Local,
```
eyalb181, line 33: What's the difference between Local and Disabled?

mirrord/config/src/fs.rs
```diff
@@ -35,25 +35,16 @@ pub mod mode;
...
-/// # mirrord-config.toml
-///
-/// [feature.fs]
-/// mode = "write"
-/// include = "^.*\.baz$"
+/// [fs]
+/// mode = read
+/// read_write = ["/var/run"]
+/// read_only = ["/etc"]
+/// local_only = ["/lib"]
```
eyalb181, line 47:
```suggestion
/// local = ["/lib"]
```

mirrord/config/src/fs/advanced.rs
```diff
@@ -14,29 +14,36 @@ use crate::{
...
+/// Besides the default behavior, user can specify behavior for specific regex patterns. Case
+/// insensitive. 1. read_write - List of patterns that should be read/write remotely.
+/// 2. read_only - List of patterns that should be read only remotely.
+/// 3. local_only - List of patterns that should be read locally.
 ///
-/// ```yaml
-/// # mirrord-config.yaml
+/// The logic for choosing the behavior is as follows:
+/// 1. Check if one of the patterns match the file path, do the corresponding action. There's no
+/// specified order    if two lists match the same path, we will use the first one (and we do not
```
eyalb181, line 30: Would write that "Specifying the same path in two lists is unsupported and can lead to undefined behaviour." Would also log a warning when it happens.

mirrord/config/src/fs/advanced.rs
```diff
@@ -14,29 +14,36 @@ use crate::{
...
 /// ```yaml
 /// # mirrord-config.yaml
 ///
 /// [fs]
-/// mode = write
-/// include = "^.*\.baz$"
+/// mode = read
+/// read_write = ["/var/run"]
+/// read_only = ["/etc"]
+/// local_only = ["/lib"]
```
eyalb181, line 46:
```suggestion
/// local = ["/lib"]
```

mirrord/config/src/fs/mode.rs
```diff
@@ -38,6 +36,7 @@ pub enum FsModeConfig {
     #[default]
```
eyalb181, line 36: Weirdly inconsistent, I feel like I'm missing something. Shouldn't it be ReadOnly, ReadWrite, Local, and Disabled?

mirrord/config/src/fs/advanced.rs
aviramha: We can't know that unless we process all Regexes then see if more than one match which feels like an overkill.

mirrord/cli/src/config.rs
aviramha: Local means you can still provide overrides while disabled means nothing is done using mirrord in terms of fs.

mirrord/config/src/fs/mode.rs
aviramha: I missed the enum here with Disabled (as it wasn't before)
I don't want to change the field name itself though as it will be a breaking change.

mirrord/cli/src/config.rs
eyalb181: So local without added overrides is the same as disabled? Seems redundant

mirrord/config/src/fs/advanced.rs
eyalb181: Okay but the log comment still stands

mirrord/cli/src/config.rs
aviramha: There are also defaults that are made by us (for example, aws cli cache) that'd read remotely even if user didn't include anything.
Also, what do you suggest as an alternative?

eyalb181: So Disabled == Local without the default overrides? Best option I can think of is to have a separate parameter for "default_overrides". Otherwise it's really unclear what the difference between the two is, plus "Disabled" feels more like the "default" option.

aviramha: Disabled means that any overrides don't work, i.e no file hooks..

eyalb181 (approved):

mirrord/cli/src/config.rs
```diff
@@ -59,6 +67,10 @@ pub(super) struct ExecArgs {
     #[arg(long = "rw")]
     pub enable_rw_fs: bool,
 
+    /// Default file system behavior: disabled, read, write, local
+    #[clap(value_parser)]
+    pub fs_mode: Option<FsMode>,
```
DmitryDodzin, line 72: Is there a use of it in cli/main.rs? I don't see any env being set

mirrord/cli/src/config.rs
```diff
@@ -28,6 +28,14 @@ pub(super) enum Commands {
         .multiple(true)
         .args(&["target", "config_file"]),
 ))]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+enum FsMode {
```
DmitryDodzin, line 32: This needs mapping to FsModeConfig or something like this?

DmitryDodzin (changes requested):
cli needs a bit of a change

meowjesty (approved):
🐻

DmitryDodzin (approved):

PR #787, opened 2022-11-24
[Merged by Bors] - Source env improvement2

aviramha:


mirrord/config/src/config.rs
```diff
@@ -15,6 +15,12 @@ pub enum ConfigError {
     #[error("value for {1:?} not provided in {0:?} (env override {2:?})")]
     ValueNotProvided(&'static str, &'static str, Option<&'static str>),
 
+    #[error("value {0:?} for {1:?} is invalid.")]
+    InvalidValue(String, &'static str),
+
+    #[error("default value {0:?}")]
+    InvalidDefaultValue(&'static str),
```
meowjesty, line 22: How can we have an invalid default value? Is this an user provided default?

meowjesty: Can't we move this to the macro crate and have it checked at compile time? So we get a nice error if we put a bonkers value as the default somewhere?

meowjesty (changes requested):
Only 1 thing sticks out to me, but I'm not the config master.

Overall looks good.

aviramha:
I noticed that in case of `nested` config all the other attributes don't matter (you can't env it, nor default.) so I do early return to make code "work"

aviramha:
I have no idea what's going on with the incoming config where the default doesn't default ;_;

mirrord/config/derive/src/config/field.rs
```diff
@@ -173,10 +170,23 @@ impl ConfigField {
...
-        quote! { #ident: (#(#impls),*) #(#layers)* .source_value()#unwrapper }
+        let impls = impls
+            .into_iter()
+            .reduce(|acc, impl_| quote! { #acc.or(#impl_) });
+
+        let layers = layers
+            .into_iter()
+            .reduce(|acc, layer| quote! { #acc.or(#layer) });
+
+        quote! { #ident: #impls #layers .source_value().transpose()?#unwrapper }
```
DmitryDodzin, line 189: layers already contains the call to `.layer(Fn())` will not work with `.or(Fn())`

```suggestion
        quote! { #ident: #impls (#(#impls),*) .source_value().transpose()?#unwrapper }
```

mirrord/config/src/config/source.rs
```diff
@@ -104,19 +58,15 @@ mod tests {
...
     #[case(None, 10)]
     #[case(Some("13"), 13)]
-    fn basic(#[case] env: Option<&str>, #[case] expect: i32) {
-        with_env_vars(vec![("TEST_VALUE", env)], || {
-            let val = (FromEnv::new("TEST_VALUE"), None, DefaultValue::new("10"));
-
-            assert_eq!(val.source_value(), Some(expect));
+    fn basic(#[case] env: Option<&str>, #[case] outcome: i32) {
+        with_env_vars(vec![("TEST_VALUE", env), ("FALLBACK", Some("10"))], || {
+            let val = FromEnv::<i32>::new("TEST_VALUE").or(FromEnv::new("FALLBACK"));
```
DmitryDodzin, line 68:
```suggestion
            let val = FromEnv::<i32>::new("TEST_VALUE").or(None).or(FromEnv::new("FALLBACK"));
```

mirrord/config/derive/src/config/field.rs
```diff
@@ -137,21 +137,18 @@ impl ConfigField {
             ..
         } = &self;
 
+        // Rest of flow is irrelevant for nested config.
+        if flags.nested {
+            return quote! { #ident: self.#ident.unwrap_or_default().generate_config()? };
```
DmitryDodzin, line 142: Breaking here for nested types will prevent us from using `.layer` on nested types in the future (if we want to keep the layer design for augmenting config or just messaging the user with the configs)

mirrord/config/src/network.rs
```diff
@@ -37,41 +35,31 @@ pub struct NetworkConfig {
...
 impl MirrordToggleableConfig for NetworkFileConfig {
     fn disabled_config() -> Result<Self::Generated, ConfigError> {
+        let incoming = FromEnv::new("MIRRORD_AGENT_TCP_STEAL_TRAFFIC")
+            .source_value()
+            .transpose()?
+            .unwrap_or(IncomingConfig::Mirror);
+        let dns = FromEnv::new("MIRRORD_REMOTE_DNS")
+            .source_value()
+            .transpose()?
+            .unwrap_or(true);
```
DmitryDodzin, line 59:
```suggestion
            .unwrap_or(false);
```

aviramha: but default should be true

mirrord/config/derive/src/config/field.rs
aviramha: how can a nested type need flag though? it's probably its children that will need it anyway?

mirrord/config/derive/src/config/field.rs
aviramha: double impls? wat?

mirrord/config/derive/src/config/field.rs
DmitryDodzin: I was thinking if we want some tracing layer or something similar

DmitryDodzin: Not important just nitpick

mirrord/config/src/network.rs
DmitryDodzin: it's the disabled_config

mirrord/config/derive/src/config/field.rs
DmitryDodzin: ops it should be layers

DmitryDodzin (approved):

meowjesty (approved):
👍

Issue #712, opened 2022-11-09
Add the ability to pause the remote pod


aviramha:
Today, mirrord runs side-by-side with the existing pod/s. Many users want to "replace" the remote pod.
I think the best, nicest, approach wouldn't be to really replace the replicaset with our agent, but still run it side by side, making the agent send `SIGSTOP` to the pod's process, then `SIGCONT` when agent finishes running.
This way we're just pausing the activity of the original pod.

Small hint-  I think we should have a loop that sends SIGSTOP every few seconds, as there might be some OS stuff that will send `SIGCONT` without us knowing/wanting.

Suggestion - flag would be `--pause` `-p`
- We need to test it on multiple runtimes, so this should be covered in e2e, not just integration testing.
- Manually test few frameworks (Python, Java, Go) to see how they behave with this method.
- Check a scenario where the remote pod has health checks - I assume the local service would answer for those so k8s won't mark it as not ready / dead

t4lz:
Should we use [docker's `pause` command](https://docs.docker.com/engine/reference/commandline/pause/#examples) if docker is the container's runtime?

And maybe if it's not docker we can still use [cgroup freezer](https://www.kernel.org/doc/Documentation/cgroup-v1/freezer-subsystem.txt) which is apparently what docker uses to pause?

aviramha:
I wouldn't bother with `docker` pause as it's not the common runtime.
`cgroup freezer` sounds cool but also overkill for now if we already went with the sigstop..

t4lz:
Saw your reply too late. But now we have a neat implementation using the docker/containerd API to pause the container instead of doing it ourselves :)

t4lz:
Is adding this feature to the IDEs part of this issue or should I open separate issues for that?
(PRs would be separate either way I guess)

aviramha:
> Is adding this feature to the IDEs part of this issue or should I open separate issues for that? (PRs would be separate either way I guess)

No, this feature should be possible to enable from the config file - VS Code already uses it and IntelliJ is underway.

t4lz:
Stable status tracked in #809, support for ephemeral tracked in #795.

bcouetil:
That exactly what is preventing me from using Telepresence 2.

Many thanks 🙏🎉

aviramha:
@bcouetil Thanks for letting us know! It'll make us prioritize stabilizing this feature :) #809 
Let us know if you have any more feature requests/feedback.

PR #791, opened 2022-11-28
[Merged by Bors] - Pause

t4lz:
https://github.com/metalbear-co/mirrord/issues/712

Add the `--pause` flag which pauses the target container while there are clients connected to the agent.
When the last client disconnects the container resumes.

Currently the pause behaviour is determined in the creation of the agent. So if clients from a different run want to reuse the same agent they will just get the existing pause behaviour of the agent.

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -177,7 +177,7 @@
           ]
         },
         "pause": {
-          "description": "Controls target pause feature.\n\nWith this feature enabled, the deployed app is paused while clients are connected to the agent.",
+          "description": "Controls target pause feature.\n\nWith this feature enabled, the remote container is paused while clients are connected to the agent.",
           "type": [
             "boolean",
             "null"
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -177,7 +177,7 @@
           ]
         },
         "pause": {
-          "description": "Controls target pause feature.\n\nWith this feature enabled, the remote container is paused while clients are connected to the agent.",
+          "description": "Controls target pause feature. Unstable.\n\nWith this feature enabled, the remote container is paused while clients are connected to the agent.",
           "type": [
             "boolean",
             "null"
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -176,6 +176,13 @@
             "null"
           ]
         },
+        "pause": {
+          "description": "Controls target pause feature.\n\nWith this feature enabled, the deployed app is paused while clients are connected to the agent.",
+          "type": [
+            "boolean",
+            "null"
+          ]
+        },
         "startup_timeout": {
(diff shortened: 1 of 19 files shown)


tests/node-app.yaml
```diff
@@ -0,0 +1,44 @@
...
+  template:
+    metadata:
+      labels:
+        app: node-http
+    spec:
+      containers:
+        - name: node-http
+          image: ghcr.io/metalbear-co/mirrord-node:latest
+          ports:
+            - containerPort: 80
```
aviramha, line 21: health checks

mirrord/agent/src/main.rs
```diff
@@ -51,27 +52,87 @@ mod util;
...
+/// Keeps track of connected clients.
+/// If pausing target, also pauses and unpauses when number of clients changes from or to 0.
 #[derive(Debug)]
 struct State {
-    pub clients: HashSet<ClientID>,
+    clients: HashSet<ClientID>,
     index_allocator: IndexAllocator<ClientID>,
+    /// Was the pause argument passed? If true, will pause the container when no clients are
+    /// connected.
+    should_pause: bool,
```
aviramha, line 63: we should issue a warning that pausing without stealing might kill the original pod because of live checks.

tests/node-app.yaml
t4lz: Yeah I'm still working on the tests, and I didn't push most of what I already have for tests. I created a deployment with liveness probes in the `test-images` repo, and used it to test manually, but didn't push it yet because I don't yet know if I'll use it in the tests.

mirrord-schema.json
```diff
@@ -176,6 +176,13 @@
             "null"
           ]
         },
+        "pause": {
+          "description": "Controls target pause feature.\n\nWith this feature enabled, the deployed app is paused while clients are connected to the agent.",
```
aviramha, line 180: "the deployed app" -> "the remote container"

mirrord/config/src/agent.rs
```diff
@@ -67,6 +67,13 @@ pub struct AgentConfig {
     /// and if that fails it uses eth0.
     #[config(env = "MIRRORD_AGENT_NETWORK_INTERFACE")]
     pub network_interface: Option<String>,
+
+    /// Controls target pause feature.
+    ///
+    /// With this feature enabled, the deployed app is paused while clients are connected to the
+    /// agent.
```
aviramha, line 74: Add to the doc that it's unstable.

aviramha (changes requested):
Few changes. Overall great implementation!

mirrord/config/src/agent.rs
t4lz: Done.

aviramha (approved):
LGTM!
bors r+

Issue #290, opened 2022-08-29
Support filtering of incoming traffic by header


eyalb181:
In order to support concurrent usage of mirrord on the same service in e.g. a staging cluster, we need to let users filter requests routed to them by header.
For example, when running mirrord-layer with FILTER_INCLUDE="mirrord-user: eyal", only requests with the specified header would be routed by the agent to the layer. Extrapolate to FILTER_EXCLUDE. 

Architecture [here](https://miro.com/app/board/uXjVP8O_H0Q=/).

Subtasks:

- [x] Change existing stealer to be global per agent, rather than multiple stealers per agent (one per subscribed layer)
- [x] Develop the Filter component - this component will know to receive raw HTTP traffic and a collection of filters, and determine whether the traffic answers one or more of the filter conditions. If it answers more than one filter conditions, just use the first one.
- [x] Change the Stealer component to use the Filter component.
- [ ] Implement subscription with filters by configuration in the layer.

meowjesty:
Partially implemented in #843, we're missing HTTP/2.

meowjesty:
Closing this in favor of #922

PR #797, opened 2022-11-28
HTTP/1 traffic filter.

meowjesty:
- Adds a new type `HttpFilter` that allows the user to capture HTTP/1 requests if they match the configured regexes;
- Adds an HTTP/1 proxy to check if a request should be captured or pass through;

See #290 for more info.


mirrord/layer/src/tcp_steal.rs
```diff
@@ -102,8 +109,14 @@ impl TcpHandler for TcpStealHandler {
             .ok_or(LayerError::ListenAlreadyExists)?;
 
         tx.send(ClientMessage::TcpSteal(LayerTcpSteal::PortSubscribe(port)))
```
aviramha, line 111: Thus should be if else case where if no filter steal all and if yes filter steal with filter

meowjesty:
Closing in favor of #843

PR #799, opened 2022-11-29
[Merged by Bors] - First e2e test for pause: http logger

t4lz:
Add a test that deploys two services, 1 that logs strings from http requests to stdout and 1 that sends requests.
Then we pause the requester with mirrord and verify it pauses.


t4lz:
This is still not working. I'm having some difficulties with the kube logs.

t4lz:
The images that this test uses are at https://github.com/metalbear-co/test-images/pull/12.

PR #804, opened 2022-11-30
[Merged by Bors] - Added /tmp/ to be excluded from file ops by default.

### LINKED ISSUE #800: Using asdf as a version manager fails to find a version of NodeJS/NPM
### Bug Description

When using asdf as my version manager for a bunch of languages, I cannot mirror any apps into a local k8s cluster.

I get the same message as if I had forgotten to set a local or global version for `node`. I'm sure this is somehow related to running in a different context than if I were running locally... I'm just confused. Running the same command without `mirrord exec` does work.

### Steps to Reproduce

1. `mirrord exec npm run dev --target pod/vs-deployment-<hash>`

### Backtrace

```shell
⠇ mirrord cli starting
  ✓ ready to launch process
    ✓ layer extracted
    ✓ agent pod created
    ✓ pod is ready
No version is set for command npm
Consider adding one of the following versions in your config file at /home/jim/Sites/ww-guest-site/.tool-versions```
```

### Relevant Logs

_No response_

Linux kernel 6.0.9

### Local process

`npm run dev` (which runs `astro dev`)

### Local process version

node v18.8.0

### Additional Info

_No response_

aviramha:
Closes #800


mirrord/layer/src/lib.rs
```diff
@@ -85,8 +85,7 @@ static GUM: LazyLock<Gum> = LazyLock::new(|| unsafe { Gum::obtain() });
 
 pub(crate) static mut HOOK_SENDER: Option<Sender<HookMessage>> = None;
 
-pub(crate) static ENABLED_FILE_OPS: OnceLock<bool> = OnceLock::new();
-pub(crate) static ENABLED_FILE_RO_OPS: OnceLock<bool> = OnceLock::new();
+pub(crate) static FILE_MODE: OnceLock<FsConfig> = OnceLock::new();
```
meowjesty, line 88: Better name, but considering that there is a "mode" for opening files, maybe we could have it be less ambiguous, like `FILE_OPS_MODE`?

meowjesty (approved):
👍

mirrord/layer/src/lib.rs
aviramha: I like that it's the same name as the argument/env etc.. Leaving asi s

PR #813, opened 2022-12-04
Ignore path + hook issues

aviramha:
- Fix hooking wrong libc functions
- Add more paths to the default ignore list (`/snap` and `*/.asdf/*`)

### DIFF
--- a/mirrord/layer/src/socket/hooks.rs
+++ b/mirrord/layer/src/socket/hooks.rs
@@ -277,30 +277,57 @@ unsafe extern "C" fn freeaddrinfo_detour(addrinfo: *mut libc::addrinfo) {
     }
 }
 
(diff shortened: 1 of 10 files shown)


mirrord/layer/src/socket/hooks.rs
```diff
@@ -277,78 +277,118 @@ unsafe extern "C" fn freeaddrinfo_detour(addrinfo: *mut libc::addrinfo) {
...
     );
 
     let _ = replace!(
         interceptor,
         "getsockname",
         getsockname_detour,
         FnGetsockname,
-        FN_GETSOCKNAME
+        FN_GETSOCKNAME,
+        Some("libc-2.31.so")
```
DmitryDodzin, line 339: Is this should be hardcoded?

aviramha: oops, nice catch.

DmitryDodzin (approved):

DmitryDodzin (approved):

PR #815, opened 2022-12-05
[Merged by Bors] - Make tcp stealer "global" in the agent.

meowjesty:
- Refactors `steal_worker` into 2 parts: `TcpConnectionStealer` and `TcpStealerApi` (similar to how `sniffer` does things);
- `TcpConnectionStealer` holds all client connections;
- `TcpStealerApi` handles the communication between `agent` and the `TcpConnectionStealer` task;


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 - IntelliJ Plugin: downgrade Java to version 11.
 - Disable progress in mirrord-layer - can cause issues with forks and generally confusing now
   that agent is created by cli (and soon to be created by IDE plugin via cli).
+- mirrord-agent: Stealer feature changed from working per connection to now starting with
```
aviramha, line 15: I'd add a tl;dr that this is a background change that shouldn't have effect on users.

mirrord/agent/src/steal/connection.rs
```diff
@@ -0,0 +1,354 @@
...
+            set_namespace(namespace).unwrap();
+        }
+
+        Ok(Self {
+            port_subscriptions: Subscriptions::new(),
+            command_rx,
+            clients: HashMap::with_capacity(8),
+            index_allocator: IndexAllocator::new(),
+            stealer: TcpListener::bind((Ipv4Addr::UNSPECIFIED, 0)).await?,
+            iptables: SafeIpTables::new(iptables::new(false).unwrap())?,
```
aviramha, line 80: nitpick -if we could initialize iptables only when needed (if there's active steal) and remove when there are no steals, that'd be great (in terms of clean) - don't put much effort but if it's like 30-40min work then it's worth it

aviramha (changes requested):
Great work on the documentation and code.
I wish we had already generics to solve some of the "boiler plate" that is being done here, but that's less important for now.
Main blocker is the fact that many connections can "steal" same port. We need to error on that and change the design a bit so it can't happen (because that's weird and doesn't really work)

mirrord/agent/src/steal/connection.rs
t4lz: https://github.com/meowjesty/mirrord/pull/5 Like this?

mirrord/agent/src/main.rs
```diff
@@ -365,6 +378,12 @@ async fn start_agent() -> Result<()> {
             .and_then(|sniffer| sniffer.start(sniffer_cancellation_token)),
     );
 
+    let stealer_cancellation_token = cancellation_token.clone();
```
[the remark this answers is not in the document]
meowjesty: Most of this duplication could be solved with some generics, like we could have some `TcpThingy<T>` where `T` are `struct Stealer` and `struct Sniffer` with code that is specific for each (could be even better than a trait, imo), while `struct TcpThingy<T>` has the parts that are common for both.

mirrord/config/src/lib.rs
```diff
@@ -42,7 +42,7 @@ use crate::{
 /// skip_processes = ["ide-debugger", "ide-service"] # we don't want mirrord to hook into these
 ///
 /// [agent]
-/// log_level = "debug"
+/// log_level = "trace"
```
meowjesty, line 45: This change was a search/replace mistake, oops.

mirrord/agent/src/main.rs
```diff
@@ -336,6 +330,21 @@ impl ClientConnectionHandler {
             LayerTcp::PortUnsubscribe(port) => self.tcp_sniffer_api.port_unsubscribe(port).await,
         }
     }
+
+    async fn handle_client_steal(&mut self, message: LayerTcpSteal) -> Result<()> {
```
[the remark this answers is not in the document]
t4lz: Ok, I'm working on extracting a common interface for the API and make this code more concise. About moving the translation of the layer message to an API operation to the API - I think maybe the authors of this code (stealer is copied from sniffer) wanted to keep a separation between the layer and the API, and not have the API "layer aware".
Personally I think both options are fine, but before moving this logic to the API I want to check again: is it ok to make the api aware of the `LayerTcp` messages from `mirrord_protocol`? it is already using `DaemonTcp` and `TcpData` from `mirrord_protocol`, but not `LayerTcp` .

mirrord/config/src/lib.rs
t4lz:
```suggestion
/// log_level = "debug"
```

mirrord/agent/src/main.rs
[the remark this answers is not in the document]
t4lz: Moved client message handling into the APIs.

mirrord/agent/src/main.rs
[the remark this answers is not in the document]
t4lz: It was decided to leave the generics for now, marking as resolved.

eyalb181 (approved):

aviramha (approved):

mirrord/agent/src/steal/connection.rs
```diff
@@ -86,6 +91,7 @@ impl TcpConnectionStealer {
             iptables: None, // Initialize on first subscription.
             write_streams: HashMap::with_capacity(8),
             read_streams: StreamMap::with_capacity(8),
+            connection_client: HashMap::with_capacity(8),
```
eyalb181, line 94: Is this supposed to be `connection_clients`**?

mirrord/agent/src/steal/connection.rs
```diff
@@ -347,10 +369,35 @@ impl TcpConnectionStealer {
...
+    fn connection_unsubscribe(&mut self, connection_id: ConnectionId) {
+        if let Some(client_id) = self.remove_connection(connection_id) {
+            // Remove the connection from the set of the connections that belong to its client.
+            self.client_connections
+                .entry(client_id)
+                .and_modify(|connections| {
+                    connections.remove(&connection_id);
+                });
+            // If we removed the last connection of this client, remove client from map.
+            if self
```
eyalb181, line 393: Can't you remove the connection within the `and_modify` block above? If not, maybe just check that the connections are empty within `and_modify` and toggle some flag, then use that to remove the client, avoiding getting the client connection three times?

mirrord/agent/src/steal/connection.rs
t4lz: Should I change it to that? Called it `connection_client` because it gives you the single client of a connection, unlike `client_connections` which gives you the set of connections of a client.

mirrord/agent/src/steal/connection.rs
t4lz: Can't remove it inside `and_modify`, because `self.client_connections.remove` borrows `self` as mutable, and `self` is already borrowed in the `entry(..).and_modify(..)` call.

Added the flag, doing one less hashmap lookup, thanks! :)

mirrord/agent/src/steal/connection.rs
eyalb181: If it's a data structure containing many clients, each associates with a connection, then yeah, it should be `connection_clients` (and the other one should be `client_connectionses`, but alas).

t4lz: done

eyalb181 (approved):

PR #827, opened 2022-12-08
[Merged by Bors] - Asdf fix at last

aviramha:


mirrord/layer/src/lib.rs
```diff
@@ -567,6 +567,24 @@ fn enable_hooks(enabled_file_ops: bool, enabled_remote_dns: bool) {
     }
 }
 
+/// Shared code for closing fd in our data structures
+/// Callers should call their respective close before calling this.
+pub(crate) fn close_logic(fd: c_int) {
```
DmitryDodzin, line 572: maybe rename to post_close() because unlike other _logic functions are the ones that call original libc functions

DmitryDodzin (approved):
Small Nitpick

PR #832, opened 2022-12-08
[Merged by Bors] - Add lstat

aviramha:
draft


mirrord/agent/src/file.rs
```diff
@@ -159,6 +160,14 @@ impl FileManager {
                 let access_result = self.access(pathname.into(), mode);
                 Ok(FileResponse::Access(access_result))
             }
+            FileRequest::Lstat(LstatRequest { path }) => {
+                let pathname = path
```
eyalb181, line 164: Anyway to avoid this duplication with Access? Nest the match arms?

mirrord/layer/src/file/hooks.rs
```diff
@@ -393,6 +394,54 @@ unsafe fn fileno_logic(file_stream: *mut FILE) -> c_int {
...
+}
+
+/// Hook for `libc::lstat`.
+#[hook_guard_fn]
+unsafe extern "C" fn lstat_detour(raw_path: *const c_char, out_stat: *mut stat) -> c_int {
+    let path = (!raw_path.is_null()).then(|| CStr::from_ptr(raw_path));
+
+    let (Ok(result) | Err(result)) = lstat(path)
+        .map(|res| {
+            let res = res.metadata;
```
eyalb181, line 416: Why's all this in the hook and not in the `lstat` method?

aviramha: following how other hooks work - the `ops` methods are rusty and safe, and here we assign to a `*mut T` (raw pointer) values from the rust struct, which is unsafe so is done here.

mirrord/agent/src/file.rs
aviramha: it will create another level of nesting..
it's also copy-paste of the stuff above so tbh we need more macros.

eyalb181: Do we think nesting is somehow worse than having duplicate code? And what's the stuff above?

aviramha: Having match inside match to share 2 lines of code seems like bad choice, especially as two structs have nothing in common aside from that 2-line-long-prolog.
This whole place needs to be revamped with generics +macros so I don't think that's worth doing now

aviramha:
Found out that I need to cover all *stat functions for this to be valuable, marking as draft until I finish.

aviramha:
Done, need review :)

mirrord/agent/src/file.rs
```diff
@@ -457,4 +466,62 @@ impl FileManager {
...
+        pathname: Option<PathBuf>,
+        fd: Option<usize>,
+        follow_symlink: bool,
+    ) -> RemoteResult<XstatResponse> {
+        let path = match (pathname, fd) {
+            // lstat/stat or fstatat with fdcwd
+            (Some(pathname), None) => pathname,
+            // fstatat
+            (Some(pathname), Some(fd)) => {
+                if let RemoteFile::Directory(path) = self
```
eyalb181, line 482:
```suggestion
                if let RemoteFile::Directory(parent_path) = self
```

mirrord/agent/src/file.rs
```diff
@@ -457,4 +466,62 @@ impl FileManager {
...
+                if let RemoteFile::File(file) = self
+                    .open_files
+                    .get(&fd)
+                    .ok_or(ResponseError::NotFound(fd))?
+                {
+                    return Ok(XstatResponse {
+                        metadata: file.metadata()?.into(),
+                    });
+                } else {
+                    return Err(ResponseError::NotDirectory(fd));
```
eyalb181, line 503: Shouldn't this be `NotFile`?

mirrord/agent/src/file.rs
```diff
@@ -457,4 +466,62 @@ impl FileManager {
...
+                }
+            }
+            // invalid
+            _ => {
+                return Err(ResponseError::RemoteIO(
+                    std::io::ErrorKind::Unsupported.into(),
+                ))
+            }
+        };
+        let pathname = path
```
eyalb181: Is there some difference in meaning between `path` and `pathname`? Why are you ping-ponging between them throughout the function?

mirrord/layer/src/file/ops.rs
```diff
@@ -418,3 +422,64 @@ pub(crate) fn access(rawish_path: Option<&CStr>, mode: u8) -> Detour<c_int> {
...
+/// that.
+/// rawish_path is Option<Option<&CStr>> because we need to differentiate between null pointer
+/// and non existing argument (For error handling)
+#[tracing::instrument(level = "trace")]
+pub(crate) fn xstat(
+    rawish_path: Option<Option<&CStr>>,
+    fd: Option<RawFd>,
+    follow_symlink: bool,
+) -> Detour<XstatResponse> {
+    // Can't use map because we need to propogate captured error
```
eyalb181, line 437:
```suggestion
    // Can't use map because we need to propagate captured error
```

mirrord/layer/src/file/ops.rs
```diff
@@ -418,3 +422,64 @@ pub(crate) fn access(rawish_path: Option<&CStr>, mode: u8) -> Detour<c_int> {
...
+    let (path, fd) = match (rawish_path, fd) {
+        // fstatat
+        (Some(path), Some(fd)) => {
+            let path = path_from_rawish(path)?;
+            let fd = {
+                if fd == AT_FDCWD {
+                    if path.is_relative() {
+                        // Calls with non absolute paths are sent to libc::open.
```
eyalb181, line 445: i.e. by libc::fstatat?

mirrord/agent/src/file.rs
aviramha: depends on where I copy-pasted it from :p

aviramha:
Fixed all comments @eyalb181 - nice catches, thank you.

eyalb181 (approved):

infiniteregrets (commented):
overall looks good, for the failing tests:
```
 file_ops::file_ops::test_file_ops::agent_1_Agent__Ephemeral::ops_1_FileOps__Python    file_ops::file_ops::test_file_ops::agent_2_Agent__Job::ops_1_FileOps__Python
```
they assert whether the file exists locally using .exists() or something, so i believe we should replace the assertion with something else

aviramha:
> overall looks good, for the failing tests:
> 
> ```
>  file_ops::file_ops::test_file_ops::agent_1_Agent__Ephemeral::ops_1_FileOps__Python    file_ops::file_ops::test_file_ops::agent_2_Agent__Job::ops_1_FileOps__Python
> ```
> 
> they assert whether the file exists locally using .exists() or something, so i believe we should replace the assertion with something else

mirrord should "proxy" the exist to the remote, if I'm not wrong then we should make it work by hooking missing fstat/stat calls that are used for that.

PR #835, opened 2022-12-09
[Merged by Bors] - Start e2e tests in parallel to integration tests.

t4lz:
Currently the e2e tests depend on the integration tests, so that they are only run if integration tests are successful. Sometimes (but not always) the integration tests finish after the other dependencies of the e2e tests, so by removing the dependency the e2e tests would start earlier in those cases and the whole github action would end earlier.


aviramha (approved):
Nice catch!
bors r+

PR #836, opened 2022-12-11
[Merged by Bors] - vscode: use mirrord ls

infiniteregrets:
wip


aviramha:
Lets include in this change to only include the cli, and use extract to create the so.

vscode-ext/src/extension.ts
```diff
@@ -177,52 +185,36 @@ class ConfigurationProvider implements vscode.DebugConfigurationProvider {
...
-				return;
-			}
-			let k8sApi = getK8sApi();
-			// Get pods from kubectl and let user select one to mirror
-			let pods: { response: any, body: V1PodList } = await k8sApi.listNamespacedPod(podNamespace);
-			let podNames = pods.body.items.map((pod) => pod.metadata!.name!);
-			let impersonatedPodName = '';
-			await vscode.window.showQuickPick(podNames, { placeHolder: 'Select pod to mirror' }).then(async podName => {
+		// If target wasn't specified in the config file, let user choose pod from dropdown			
+		if (!await isTargetInFile()) {
```
eyalb181, line 207: Why'd you remove the namespace logic? If the user specified a namespace in mirrord.config, we want only targets from that namespace to be displayed.

aviramha: tbh I think better approach would be to just load the config file (so extension don't need to parse it)
but that might be an overkill right now

eyalb181: Load it in the CLI? And then `mirrord ls` would be filtered without any additional args?

aviramha: or you can specify to override when using the cli, but yes

eyalb181: Agreed that it's a better approach but overkill so lets stick with current implementation for now

aviramha: just making sure - right now the default is default which shouldn't be as it'd override the kubeconfig's default
if nothing in target file, don't set the flag.

vscode-ext/src/extension.ts
```diff
@@ -177,52 +216,38 @@ class ConfigurationProvider implements vscode.DebugConfigurationProvider {
...
+			let targetNamespace = await parseNamespace();
+			let targets = targetNamespace != undefined ? await getMirrordTargets(cliPath, targetNamespace) : await getMirrordTargets(cliPath);
```
aviramha, line 241: why not `let targets = getMirrordTargets(cliPath, targetNamespace)` as `targetNamespace` will be undefined anyway?

PR #838, opened 2022-12-12
[Merged by Bors] - adjust filters and add /Applications

aviramha:


eyalb181 (approved):

mirrord/layer/src/file/filter.rs
```diff
@@ -82,8 +82,9 @@ fn generate_local_set() -> RegexSet {
         // asdf
         r".*/\.tool-versions$",
         // macOS
-        r"^/Users/.*$",
-        r"^/Library/.*$",
+        r"^/Users",
+        r"^/Library",
+        r"^/Applications",
```
t4lz, line 87: We could be more precise:
```suggestion
        r"^/Users(/.*)?$",
        r"^/Library(/.*)?$",
        r"^/Applications(/.*)?$",
```

aviramha: What cases would it solve though?

t4lz: It would prevent matching paths that have those directories as prefixes, e.g. `/Users.db` or `/LibraryFiles/whatever` probably not very important.

aviramha: 💀

PR #865, opened 2022-12-28
[Merged by Bors] - Custom mirrord-agent container engine socket support

DmitryDodzin:
Support custom socket paths for container runtimes like for k3s `/var/run/k3s/containerd/containerd.sock`


aviramha (approved):
bors r+

PR #884, opened 2023-01-02
[Merged by Bors] - refactored detours to use Result::inner

aviramha:


mirrord/layer/src/detour.rs
```diff
@@ -240,3 +240,39 @@ impl<T> OptionExt for Option<T> {
...
+    fn inner(self) -> Self::Res;
+}
+
+impl<T> ResultExt for Result<T, HookError>
+where
+    T: From<HookError>,
+{
+    type Res = T;
+
+    fn inner(self) -> T {
```
meowjesty, line 272: I would suggest naming this something like `c_result`, or maybe `c_code`, `as_c_result`, `as_hook_result`, `as_hook_code`. Some name that better represents that we're converting `Result -> int`.

mirrord/layer/src/detour.rs
```diff
@@ -223,6 +217,35 @@ impl<S> Detour<S> {
...
 }
 
+impl<S> Detour<S>
+where
+    S: From<HookError>,
+{
+    /// Helper function for returning a detour return value from a hook.
+    /// Success -> Return the contained value.
+    /// Bypass -> call the bypass and return its value.
+    /// Error ->> Convert to libc value and return it.
```
meowjesty, line 227:
```suggestion
    /// Helper function for returning a detour return value from a hook.
    ///
    /// - `Success` -> Return the contained value.
    /// - `Bypass` -> Call the bypass and return its value.
    /// - `Error` -> Convert to libc value and return it.
```

mirrord/layer/src/detour.rs
```diff
@@ -223,6 +217,35 @@ impl<S> Detour<S> {
...
+            Detour::Success(s) => s,
+            Detour::Bypass(b) => op(b),
+            Detour::Error(e) => e.into(),
+        }
+    }
+
+    /// Helper function for returning a detour return value from a hook.
+    /// Success -> Return the contained value.
+    /// Bypass -> Return provided value.
+    /// Error ->> Convert to libc value and return it.
```
meowjesty, line 239:
```suggestion
    /// Helper function for returning a detour return value from a hook.
    ///
    /// `Success` -> Return the contained value.
    /// `Bypass` -> Return provided value.
    /// `Error` -> Convert to libc value and return it.
```

meowjesty (changes requested):
Nicer, commenting some fixes to the rust docs.

meowjesty (approved):
🐻 👍

PR #888, opened 2023-01-02
[Merged by Bors] - Better Docker Compatibility

DmitryDodzin:
mount /var from host and reconfigure docker socket to /var/run/docker.sock for better compatibility


aviramha (commented):
bors r+

aviramha (approved):

PR #904, opened 2023-01-05
[Merged by Bors] - partial ls support

infiniteregrets:


mirrord/layer/src/file/ops.rs
```diff
@@ -217,6 +219,80 @@ pub(crate) fn fdopen(fd: RawFd, rawish_mode: Option<&CStr>) -> Detour<*mut FILE>
...
+    // usize == ptr size
+    // we don't return a pointer to an address that contains DIR
+
+    let open_files = OPEN_FILES.lock()?;
+    let remote_fd = open_files.get(&fd).ok_or(Bypass::LocalFdNotFound(fd))?;
+
+    let (dir_channel_tx, dir_channel_rx) = oneshot::channel();
+
+    let open_dir_request = OpenDir {
+        remote_fd: *remote_fd,
```
aviramha, line 234: Don't use same fd.. let the remote generate fd like we do for everything else.

aviramha (changes requested):

mirrord/agent/src/file.rs
```diff
@@ -22,14 +26,15 @@ use crate::{error::Result, util::IndexAllocator};
 
 #[derive(Debug)]
 pub enum RemoteFile {
-    File(File),
+    File(File, PathBuf),
```
aviramha, line 29: Pretty sure it's not needed as you can't really call `opendir/fdopendir` on a file

mirrord/layer/src/file/ops.rs
infiniteregrets: i am not using the same fd here, i am just letting the agent know which path it should refer to

mirrord/agent/src/file.rs
infiniteregrets: its open() then fdopendir(), the fd received by open is what we use to refer to the directory, that's how I understand it, please let me know if you have any other suggestions as to what this should be

aviramha: But if you call open on a directory you would get the directory case and not the file as it will be invalid other wise

mirrord/layer/src/file/ops.rs
```diff
@@ -217,6 +219,80 @@ pub(crate) fn fdopen(fd: RawFd, rawish_mode: Option<&CStr>) -> Detour<*mut FILE>
...
+        dir_channel_tx,
+    };
+
+    blocking_send_file_message(HookMessageFile::OpenDir(open_dir_request))?;
+
+    let _ = dir_channel_rx.blocking_recv()??;
+
+    let fake_local_dir_name = CString::new(remote_fd.to_string())?;
+    let local_dir_fd = unsafe { create_local_fake_file(fake_local_dir_name, *remote_fd) }?;
```
aviramha, line 243: That's the fd of the open, not of the stream being reused instead of letting agent generate another one..

mirrord/layer/src/file/ops.rs
aviramha: Didn't put it in correct location but point stands

aviramha (changes requested):

mirrord/agent/src/file.rs
infiniteregrets: oh yeah my bad, we check on the metadata. ill fix this thanks!

mirrord/layer/src/file/ops.rs
infiniteregrets: Ah i see what you mean, ill try and fix it. thanks for clarifying

mirrord/agent/src/file.rs
```diff
@@ -516,4 +531,65 @@ impl FileManager {
...
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub(crate) fn open_dir(&mut self, fd: u64) -> RemoteResult<OpenDirResponse> {
+        trace!("FileManager::open_dir -> fd {:#?}", fd,);
+        let path = match self
+            .open_files
+            .get(&fd)
+            .ok_or(ResponseError::NotFound(fd))?
+        {
+            RemoteFile::Directory(path) => path,
+            _ => panic!("Opened path is a file."),
```
aviramha, line 544: Panic is bad, this is a "valid" case where we return an error that translates to the appropriate libc error (`ENOTDIR`)

mirrord/agent/src/file.rs
```diff
@@ -516,4 +531,65 @@ impl FileManager {
         })
         .map_err(ResponseError::from)
     }
+
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub(crate) fn open_dir(&mut self, fd: u64) -> RemoteResult<OpenDirResponse> {
```
aviramha, line 536: `fdopen_dir` since we'll prolly soon add `opendir` which will work by path

mirrord/agent/src/file.rs
```diff
@@ -516,4 +531,65 @@ impl FileManager {
...
+                libc::S_IFCHR => libc::DT_CHR,
+                libc::S_IFIFO => libc::DT_FIFO,
+                libc::S_IFSOCK => libc::DT_SOCK,
+                _ => libc::DT_UNKNOWN,
+            };
+
+            let internal_entry = DirEntryInternal {
+                inode: entry.ino(),
+                position: offset as u64,
+                name: entry.file_name().into_string().unwrap().into_bytes(),
```
aviramha, line 583: Pass it as String internally then convert it in layer to the proper encoding. (also unwrap is evil, return error.)

mirrord/layer/src/file/ops.rs
```diff
@@ -217,6 +219,98 @@ pub(crate) fn fdopen(fd: RawFd, rawish_mode: Option<&CStr>) -> Detour<*mut FILE>
...
+    if let Some(direntry) = direntry {
+        let direntry_name = direntry.name.as_slice();
+
+        #[cfg(target_os = "linux")]
+        let mut entry_name: [i8; 256] = [0; 256];
+
+        #[cfg(target_os = "macos")]
+        let mut entry_name: [i8; 1024] = [0; 1024];
+
+        let casted_name: Vec<i8> = direntry_name.iter().map(|c| *c as i8).collect();
```
aviramha, line 279: Cast the entry name to `Vec<u8>` then write it to the `entry.d_name` using `std::ptr::copy`
See opt 2 here (not exact, but similar)
https://dev.to/kgrech/7-ways-to-pass-a-string-between-rust-and-c-4ieb

mirrord/layer/src/file.rs
```diff
@@ -47,9 +50,17 @@ struct RemoteFile {
...
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct DirStream {
+    direntry: Box<DirEntryInternal>,
+}
+
 pub(crate) static OPEN_FILES: LazyLock<Mutex<HashMap<LocalFd, RemoteFd>>> =
     LazyLock::new(|| Mutex::new(HashMap::with_capacity(4)));
 
+pub(crate) static OPEN_DIRS: LazyLock<Mutex<HashMap<LocalFd, RemoteFd>>> =
```
aviramha, line 61: `LocalFd` -> `DirStreamFd`
`type DirStreamFd = usize`

mirrord/layer/src/file/ops.rs
```diff
@@ -217,6 +219,98 @@ pub(crate) fn fdopen(fd: RawFd, rawish_mode: Option<&CStr>) -> Detour<*mut FILE>
...
+
+// fetches the current entry in the directory stream created by `fdopendir`
+#[tracing::instrument(level = "trace")]
+pub(crate) fn readdir_r(
+    dirp: *mut DIR,
+    entry: *mut dirent,
+    result: *mut *mut dirent,
+) -> Detour<c_int> {
+    trace!("readdir_r -> {dirp:#?} {entry:#?} {result:#?}");
+    let remote_fd = OPEN_DIRS.lock()?.get(&(dirp as i32)).cloned().unwrap();
```
aviramha, line 257: see above comment to have the key as usize to avoid cast.

aviramha (changes requested):
Getting close.
Need to add integration test on macOS for the use case.

mirrord/agent/src/file.rs
```diff
@@ -168,6 +173,16 @@ impl FileManager {
                 let xstat_result = self.xstat(path, fd, follow_symlink);
                 Ok(FileResponse::Xstat(xstat_result))
             }
+
+            FileRequest::FdOpenDir(FdOpenDirRequest { remote_fd }) => {
+                let open_dir_result = self.fdopen_dir(remote_fd);
+                Ok(FileResponse::OpenDir(open_dir_result))
```
t4lz, line 179: Not affected by this PR, but would it be good style to move out the `Ok` to have a single `Ok` around the whole `match` instead of wrapping the return value of every single arm with `Ok`? - If yes, we could take this opportunity to do this little change.

mirrord/agent/src/file.rs
```diff
@@ -516,4 +531,64 @@ impl FileManager {
...
+            .open_files
+            .get(&fd)
+            .ok_or(ResponseError::NotFound(fd))?
+        {
+            RemoteFile::Directory(path) => Ok(path),
+            _ => ResponseError::NotDirectory(fd),
+        }?;
+
+        let fd = self.index_allocator.next_index().ok_or_else(|| {
+            ResponseError::AllocationFailure("FileManager::open_relative".to_string())
```
infiniteregrets, line 547: Typo here, shouldn't be open_relative. Sorry about that

mirrord/agent/src/file.rs
aviramha: Done

mirrord/layer/src/file.rs
```diff
@@ -218,6 +232,11 @@ impl FileHandler {
                 debug!("DaemonMessage::XstatResponse {:#?}!", xstat);
                 pop_send(&mut self.xstat_queue, xstat)
             }
+            ReadDir(read_dir) => {
+                debug!("DaemonMessage::ReadDirResponse {:#?}!", read_dir);
```
t4lz, line 236: Is this intentionally still in `debug` or should this be changed to `trace`:
```suggestion
                trace!("DaemonMessage::ReadDirResponse {:#?}!", read_dir);
```
personally I have nothing against leaving interesting logs as `debug`, as long as they are not spammy and it's intentional.

mirrord/layer/src/file/ops.rs
```diff
@@ -217,6 +219,91 @@ pub(crate) fn fdopen(fd: RawFd, rawish_mode: Option<&CStr>) -> Detour<*mut FILE>
...
+    let open_dir_request = OpenDir {
+        remote_fd: *remote_file_fd,
+        dir_channel_tx,
+    };
+
+    blocking_send_file_message(HookMessageFile::OpenDir(open_dir_request))?;
+
+    let OpenDirResponse { fd } = dir_channel_rx.blocking_recv()??;
+
+    let fake_local_dir_name = CString::new(fd.to_string())?;
```
t4lz, line 242: Maybe add some comments here about why there's a fake dir name, how it will be used and by who?

aviramha: This is a common pattern in code, we can improve it later if we want (I'd change the signature of fake file to just use remote fd and generate the fake name there as all code repeats it)

mirrord/layer/src/file/ops.rs
```diff
@@ -249,7 +331,9 @@ pub(crate) fn openat(
         let fake_local_file_name = CString::new(remote_fd.to_string())?;
         let local_file_fd = unsafe { create_local_fake_file(fake_local_file_name, remote_fd) }?;
 
-        OPEN_FILES.lock()?.insert(local_file_fd, remote_fd);
+        OPEN_FILES
```
eyalb181, line 334: Why'd you have to add Arc here? Consider documenting

eyalb181 (approved):

aviramha (approved):
bors r+

Issue #912, opened 2023-01-06
Document dependency on PACKET_IGNORE_OUTGOING


footplus:
First, thanks for developing this software :)

I had one issue when trying it: I was experimenting with RockyLinux 8 and minikube, and did get stuck when trying to mirror one local process. The error message was very generic (the agent sent a close frame with None error).

I did enable trace logs in the agent, and got a better message explaining there was some Protocol Not Supported error 92 returnef by the OS to the agent.

Digging into the agent wasn’t very easy (i had to build a custom image for stracing the agent to find where the 92 error came from).

It turns out that it was some kernel feature for packet capture that wasn’t present until kernel 4.20 i believe  (PACKET_IGNORE_OUTGOING, introduced by [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=fa788d986a3aac5069378ed04697bd06f83d3488) - RockyLinux 8 default kernel is 4.18.

After upgrading to kernel-lt from elrepo, mirrord started successfully.

I would therefore suggest:

- documenting the kernel requirements (or other requirements, like supported CRI, etc)
- better trace the error to the problematic call in agent logs
- if possible, surface the error in mirrord CLI.

Thanks !

aviramha:
Thanks for the feedback!
I'm sorry for the troubles you've gone through, but happy that you managed so well alone, well played 😎 
Your suggestions are well put, some were already some-what planned but aren't as easy.
I think we'll start with documenting CRI/Kernel in mirrord.dev/docs, and hopefully can get to the other suggestions as well.
Leaving this issue open until we get all of the suggestions done.

aviramha:
Docs updated here:
https://github.com/metalbear-co/mirrord.dev/pull/51

aviramha:
Closed with #913 , #914 and https://github.com/metalbear-co/mirrord.dev/pull/51

Thanks!

PR #913, opened 2023-01-06
[Merged by Bors] - Agent send error to layer on error

aviramha:
Agent now sends error encountered back to layer for better UX when bad times happen.

Part of #912


meowjesty (approved):
👍

PR #914, opened 2023-01-06
[Merged by Bors] - improve error for packet ignore outgoing

aviramha:
agent: Return better error message when failing to use `PACKET_IGNORE_OUTGOING` flag.

Part of #912

### DIFF
--- a/mirrord/agent/src/error.rs
+++ b/mirrord/agent/src/error.rs
@@ -106,7 +106,8 @@ pub enum AgentError {
     #[error("An internal invariant of the agent was violated, this should not happen.")]
     AgentInvariantViolated,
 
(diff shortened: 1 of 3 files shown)


mirrord/agent/src/error.rs
```diff
@@ -105,6 +105,9 @@ pub enum AgentError {
 
     #[error("An internal invariant of the agent was violated, this should not happen.")]
     AgentInvariantViolated,
+
+    #[error("Failed to set socket flag PACKET_IGNORE_OUTGOING, this might be due to kernel version before 4.20")]
```
meowjesty, line 109: Will this display the "raw" error also?

aviramha: No, good catch.

meowjesty (approved):
🐻 👍

PR #915, opened 2023-01-08
[Merged by Bors] - Operator k8s API

DmitryDodzin:
Support Operator Kubernetes API Extension


mirrord/layer/src/connection.rs
```diff
@@ -52,13 +56,8 @@ pub(crate) async fn connect(
...
             .await
             .unwrap_or_else(|err| handle_error(err, config))
-    } else if config.operator.enabled && let Some((operator_api, operator_ref)) =
-        OperatorApiDiscover::discover_operator(config, &progress).await
-    {
-        operator_api
-            .create_connection(operator_ref)
-            .await
-            .unwrap_or_else(|err| handle_error(err, config))
+    } else if config.operator && let Ok((operator_api, operator_ref)) = OperatorApi::discover(&config, &progress).await {
```
eyalb181, line 59: Lets add documentation to each of the three options here (pre-specified address, operator, new agent)

mirrord/config/src/lib.rs
```diff
@@ -119,8 +118,9 @@ pub struct LayerConfig {
     #[config(nested)]
     pub feature: FeatureConfig,
 
-    #[config(nested, toggleable)]
-    pub operator: OperatorConfig,
+    /// Allow to lookup if operator is installed on cluster and use it
+    #[config(env = "MIRRORD_OPERATOR_ENABLE", default = true)]
```
eyalb181, line 122: I'm not sure we want to allow non-operator connections in clusters with operators in them. Might be worth discussing in the channel.

mirrord/layer/src/connection.rs
```diff
@@ -52,13 +56,8 @@ pub(crate) async fn connect(
...
+        operator_api.create_connection(operator_ref).await.unwrap_or_else(|err| handle_operator_error(err))
```
eyalb181, line 60: Why doesn't Discover already create the connection? Why does it return `operator_api` and `operator_ref`, only for `operator_ref` to be sent right back to `operator_api`?

mirrord/operator/src/client.rs
```diff
@@ -1,184 +1,184 @@
...
-        Some((api, connection))
+        Ok(OperatorApi { client, target })
     }
 }
 
 #[async_trait]
-impl AgentManagment for OperatorApiDiscover {
-    type AgentRef = String;
-    type Err = KubeApiError;
+impl AgentManagment for OperatorApi {
```
eyalb181, line 59: Do we have to implement AgentManagement for OperatorApi? It doesn't actually manage agents.

eyalb181 (changes requested):

eyalb181 (approved):

Issue #926, opened 2023-01-10
Stealer stops if http filter gets a TCP stream with little data.


t4lz:
### Bug Description

ReversibleStream fails if the data is shorter then the header, so the stealer stops if the http filter gets a tcp stream with little data.

### Steps to Reproduce

Run mirrord with the http-filter, send 5 bytes over a bare TCP connection to the HTTP port.

### Relevant Logs

```shell
2023-01-10T15:48:08.979751Z ERROR ThreadId(36) start:incoming_connection:new: mirrord_agent::steal::http::filter: Something went wrong in http filter IO(
    Custom {
        kind: UnexpectedEof,
        error: "early eof",
    },
) cancellation_token=CancellationToken { is_cancelled: false } stream=PollEvented { io: Some(TcpStream { addr: 10.1.14.249:37039, peer: 192.168.65.3:41280, fd: 131 }) } address=192.168.65.3:41280 local=Ok(10.1.14.249:37039) peer=Ok(192.168.65.3:41280)
2023-01-10T15:48:08.979792Z TRACE ThreadId(36) start:incoming_connection:new: mirrord_agent::steal::http::filter: exit cancellation_token=CancellationToken { is_cancelled: false } stream=PollEvented { io: Some(TcpStream { addr: 10.1.14.249:37039, peer: 192.168.65.3:41280, fd: 131 }) } address=192.168.65.3:41280 local=Ok(10.1.14.249:37039) peer=Ok(192.168.65.3:41280)
2023-01-10T15:48:08.979802Z TRACE ThreadId(36) start:incoming_connection: mirrord_agent::steal::connection: exit cancellation_token=CancellationToken { is_cancelled: false } stream=PollEvented { io: Some(TcpStream { addr: 10.1.14.249:37039, peer: 192.168.65.3:41280, fd: 131 }) } address=192.168.65.3:41280
2023-01-10T15:48:08.979966Z TRACE ThreadId(18) start: mirrord_agent: enter cancellation_token=CancellationToken { is_cancelled: false }
2023-01-10T15:48:08.980019Z TRACE ThreadId(18) start:recv: mirrord_agent::steal::api: enter cancellation_token=CancellationToken { is_cancelled: false }
2023-01-10T15:48:08.980030Z TRACE ThreadId(18) start:recv: mirrord_agent::steal::api: exit cancellation_token=CancellationToken { is_cancelled: false }
[...]

PR #936, opened 2023-01-12
[Merged by Bors] - Deal with short streams (open and closed) in the HTTP filter.

t4lz:
https://github.com/metalbear-co/mirrord/issues/926

Make reversible stream deal with short streams.

- Add timeout for reading first bytes, so that it does not hang if
 the client sends too few bytes and keeps the connection open.
- Change from `read_exact` to `read_buf` loop so that we don't get an
 error and lose bytes if the client closes the connection after
 too few bytes.
   
Move the stream-peeking into the filter task and simplify the filter task.

Since creating a reversible stream can wait for up to 10 seconds,
move that into the separate filter task.
This also has the effect of not the stealer not being able to
fail and stop because of an error in dealing with one HTTP
connection.
And since there's no more any logic in the HttpFilterBuilder,
just convert it to a function.

Add test: complete passthrough on short stream.

### DIFF
--- a/mirrord/agent/src/steal/http/reversible_stream.rs
+++ b/mirrord/agent/src/steal/http/reversible_stream.rs
@@ -28,7 +28,7 @@ pub(crate) struct ReversibleStream<const HEADER_SIZE: usize> {
     /// We could be writing less then `HEADER_SIZE` bytes into `header`, in two cases:
     /// 1. Got a timeout before reading as `HEADER_SIZE` bytes.
     /// 2. Read an EOF after less than `HEADER_SIZE` bytes.
-    /// So we need to always now how many bytes we actually have.
+    /// So we need to always know how many bytes we actually have.
     header_len: usize,
 
     /// How many bytes out of the `header` were already read by the owner of self.
--- a/mirrord/agent/src/steal/http/reversible_stream.rs
+++ b/mirrord/agent/src/steal/http/reversible_stream.rs
@@ -2,12 +2,15 @@ use pin_project::pin_project;
 use tokio::{
     io::{AsyncRead, AsyncReadExt, AsyncWrite},
     net::TcpStream,
+    time,
+    time::{Duration, Instant},
 };
+use tracing::{trace, warn};
 
-use super::error::HttpTrafficError;
+use crate::steal::http::error::HttpTrafficError;
 
-/// Wraps a [`TcpStream`] to allow a sort of _peek_ functionality, even though we're actually going
-/// to be calling [`TcpStream::read_exact`] on it.
+/// Wraps a [`TcpStream`] to allow a sort of _peek_ functionality, by reading the first bytes, but
+/// then keeping them for later reads.
 ///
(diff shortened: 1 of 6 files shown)


t4lz:
You might want to review this one commit at a time, as they are split logically and have meaningful messages.

mirrord/agent/src/steal/http/reversible_stream.rs
```diff
@@ -19,25 +22,79 @@ use super::error::HttpTrafficError;
 pub(crate) struct ReversibleStream<const HEADER_SIZE: usize> {
     #[pin]
     stream: TcpStream,
+
     header: [u8; HEADER_SIZE],
+
+    /// We could be writing less then `HEADER_SIZE` bytes into `header`, in two cases:
+    /// 1. Got a timeout before reading as `HEADER_SIZE` bytes.
+    /// 2. Read an EOF after less than `HEADER_SIZE` bytes.
+    /// So we need to always now how many bytes we actually have.
```
eyalb181, line 31:
```suggestion
    /// So we need to always know how many bytes we actually have.
```

mirrord/agent/src/steal/http/reversible_stream.rs
```diff
@@ -19,25 +22,79 @@ use super::error::HttpTrafficError;
...
+    header_len: usize,
+
+    /// How many bytes out of the `header` were already read by the owner of self.
```
eyalb181, line 34: Not really clear

mirrord/agent/src/steal/http/reversible_stream.rs
```diff
@@ -19,25 +22,79 @@ use super::error::HttpTrafficError;
...
+    /// Build a Reversible stream from a TcpStream, move on if not done within given timeout.
+    /// Return an Error if there was an error while reading from the TCP Stream.
+    pub(crate) async fn read_header(
+        stream: TcpStream,
+        timeout: Duration,
+    ) -> Result<Self, HttpTrafficError> {
         let mut this = Self {
             stream,
             header: [0; HEADER_SIZE],
+            header_len: 0,
```
eyalb181, line 48: Add derive(Default) so we don't have to pass all these zeroes at initialization?

mirrord/agent/src/steal/http/reversible_stream.rs
t4lz: What about now https://github.com/metalbear-co/mirrord/pull/936/commits/103974dae475f1be2e4e40a2f419a2a1458c8c8e?

eyalb181: Great thanks

mirrord/agent/src/steal/http/reversible_stream.rs
t4lz: I can't derive `Default` for `ReversibleStream`, because arrays and `TcpStream`s don't implement `Default`.

eyalb181 (approved):

PR #938, opened 2023-01-12
[Merged by Bors] - fix getaddrinfo crash

aviramha:


meowjesty (approved):
🐻 👍

mirrord/layer/src/socket/hooks.rs
```diff
@@ -213,21 +227,41 @@ unsafe extern "C" fn getaddrinfo_detour(
...
+    MANAGED_ADDRINFO
+        .lock()
+        .map(|mut managed_addrinfo| {
+            managed_addrinfo
+                .remove(&(addrinfo as usize))
+                .then(|| {
+                    // Iterate over `addrinfo` linked list dropping it.
+                    let mut current = addrinfo;
+                    while !current.is_null() {
+                        managed_addrinfo.remove(&(current as usize));
```
meowjesty, line 246: Don't we already remove this `current` above in `.remove(&(addrinfo as usize))`?

Think you should move this to the end (near the `drop` calls)?

meowjesty (changes requested):
Think we have a double remove (harmless, but could be avoided)?

meowjesty (approved):
🐻 👍

PR #958, opened 2023-01-17
getdents64 hook

t4lz:
Add a detour for the `getdents64` Linux syscall to support `os.ReadDir` on Go.
Part of https://github.com/metalbear-co/mirrord/issues/120.


mirrord/agent/src/file.rs
```diff
@@ -2,13 +2,16 @@ use std::{
...
+    io,
     io::{prelude::*, BufReader, SeekFrom},
     iter::Enumerate,
-    os::unix::prelude::{DirEntryExt, FileExt, MetadataExt},
+    os::unix::prelude::FileExt,
     path::{Path, PathBuf},
 };
 
 use faccess::{AccessMode, PathExt};
+#[cfg(target_os = "linux")]
```
aviramha, line 13: currently agent doesn't compile for macOS, so not sure how helpful is adding those stuff.

aviramha (commented):
Small comment, looking great so far.

eyalb181 (approved):
Well done!

PR #963, opened 2023-01-18
[Merged by Bors] - remove --no-fs and --rw from cli args.

t4lz:
They were already deprecated and their functionality was removed, but the arguments were not deleted from the cli and so they still appear in the help message.


aviramha (approved):
Great catch
bors r+

PR #964, opened 2023-01-18
User Friendly Operator Errors

DmitryDodzin:


mirrord/layer/src/lib.rs
```diff
@@ -409,6 +409,19 @@ impl Layer {
                 .ok_or(LayerError::SendErrorGetAddrInfoResponse)?
                 .send(get_addr_info.0)
                 .map_err(|_| LayerError::SendErrorGetAddrInfoResponse),
+            DaemonMessage::Operator(operator_message) => {
```
aviramha, line 412: use miette?

PR #975, opened 2023-01-23
Operator Status

DmitryDodzin:

Please make sure you added a CHANGELOG entry.

### DIFF
--- a/mirrord/cli/src/operator.rs
+++ b/mirrord/cli/src/operator.rs
@@ -1,9 +1,14 @@
 use std::{fs::File, path::PathBuf};
 
+use kube::Api;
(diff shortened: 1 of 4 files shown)


mirrord/cli/src/operator.rs
```diff
@@ -58,6 +64,91 @@ async fn operator_setup(
...
+    status_progress
+        .subtask(&format!(
+            "Operator default namespace: {}",
+            mirrord_status.spec.default_namespace
+        ))
+        .done();
+
+    let license_progress = status_progress.subtask("Operator License");
+
+    license_progress
```
eyalb181, line 113: Are you using the progress reporter to print out the output of the function? It's confusing (+ can be disabled by the user), why not use a normal print?

eyalb181 (approved):

PR #978, opened 2023-01-23
[Merged by Bors] - Process Level Guard for IPTables

### LINKED ISSUE #955: mirrord-agent can leave leftover iptables records
### Bug Description

Not sure when this happens, but either way we need to add a more robust cleanup mechanism:
When an agent is started, it forks:
* The forked process does the actual agent logic
* The parent process waits for the forked process to exit, and when it does, ensures that it hasn't left any records in iptables

### Steps to Reproduce

N/A

### Backtrace

_No response_

### Relevant Logs

_No response_

N/A

### Local process

N/A

### Local process version

_No response_

### Additional Info

_No response_

DmitryDodzin:
closes #955


mirrord/agent/src/main.rs
```diff
@@ -464,6 +468,69 @@ async fn start_agent() -> Result<()> {
...
+}
+
+fn spawn_child_agent() -> Result<()> {
+    use std::process::{Command, Stdio};
+
+    let command_args = std::env::args().collect::<Vec<_>>();
+
+    let mut child_agent = Command::new(&command_args[0])
+        .args(&command_args[1..])
+        .stdin(Stdio::inherit())
```
aviramha, line 492: inherit is default for all so you can remove the three lines

mirrord/agent/src/main.rs
```diff
@@ -464,6 +468,69 @@ async fn start_agent() -> Result<()> {
...
+
+async fn clear_iptable_chain(chain_name: String) -> Result<()> {
+    let ipt = iptables::new(false).unwrap();
+    let formatter = IPTableFormatter::detect(&ipt)?;
+
+    SafeIpTables::remove_chain(&ipt, &formatter, &chain_name)
+}
+
+fn spawn_child_agent() -> Result<()> {
+    use std::process::{Command, Stdio};
```
aviramha, line 486: I think I would prefer a fork instead of using command?

mirrord/agent/src/main.rs
```diff
@@ -464,6 +468,69 @@ async fn start_agent() -> Result<()> {
...
+        .stderr(Stdio::inherit())
+        .spawn()?;
+
+    let _ = child_agent.wait();
+
+    Ok(())
+}
+
+async fn start_iptable_guard() -> Result<()> {
+    debug!("start_iptable_guard -> Initializing iptable-guard.");
```
aviramha, line 503: imo the chain should be created in the child and the parent only cleans (doesn't create anything)
this will guard us from cases where the guard itself gets killed unexpectedly during creation

aviramha (approved):

aviramha:
lgtm

PR #979, opened 2023-01-23
sendmsg hook

infiniteregrets:
This pr is a work in progress right now. It will add support for the sendmsg syscall.


mirrord/agent/src/outgoing/udp.rs
```diff
@@ -265,3 +265,34 @@ impl UdpOutgoingApi {
...
+        }
+    }
+
+    pub(crate) async fn handle_message(
+        &self,
+        request: SendRecvRequest,
+    ) -> Result<Option<SendRecvResponse>> {
+        match request {
+            SendRecvRequest::SendMsg(message) => {
+                let socket = UdpSocket::bind(message.addr.clone()).await?;
```
aviramha, line 286: the address in the message is of destination, we can't bind it (binding is for local)

mirrord/agent/src/main.rs
```diff
@@ -152,6 +157,7 @@ struct ClientConnectionHandler {
     tcp_outgoing_api: TcpOutgoingApi,
     udp_outgoing_api: UdpOutgoingApi,
     dns_sender: Sender<DnsRequest>,
+    send_recv_manager: SendRecvManager,
```
aviramha, line 160: why is it SendRecvManager?

infiniteregrets: since it will manage both send/recv, i thought it would make sense to have a common interface for this? would you rather just have a Sender?

mirrord/agent/src/outgoing/udp.rs
infiniteregrets: right it should be localhost

aviramha: you don't need to bind..

aviramha:
```rs
This function will create a new UDP socket and attempt to bind it to the addr provided.

Binding with a port number of 0 will request that the OS assigns a port to this listener. The port allocated can be queried via the local_addr method.
```

mirrord/agent/src/main.rs
aviramha: if it's recv then it would go through our normal udp socket flow

PR #980, opened 2023-01-23
[Merged by Bors] - Change HOOK_SENDER to be an OnceLock.

meowjesty:
Small refactor to bring `HOOK_SENDER` global to be consistent with the other lazily initiated globals, by changing it from `Option<T>` to `OnceLock<T>`.

Has the added benefit of removing a couple of `unsafe` blocks.


eyalb181 (approved):

meowjesty:
Should be a quick one bors!

bors r+

meowjesty:
Why you do this to me bors?

bors r+

Issue #987, opened 2023-01-26
No hooks (probably failing to load into a new process) when running node from homebrew. 


t4lz:
### Bug Description

https://github.com/metalbear-co/nodejs-example/issues/3

### Steps to Reproduce

1. `brew install node`
2. follow the steps from https://mirrord.dev/docs/overview/real-world-example/.
3. Requests are not mirrored :(

### Your operating system and version

macOS 13.1

### Local process

yarn

### Local process version

1.22.19

### Additional Info

node 19.5.0, installed via homebrew.

(did not reproduce before installing with homebrew).

PR #988, opened 2023-01-27
[Merged by Bors] - Hook `posix_spawn` with same logic as `execve`.

t4lz:
Fixes https://github.com/metalbear-co/mirrord/issues/987.

The problem was that this newer node version does not use libc's `execve` to spawn a new process but libuv's `uv_spawn`, which calls `posix_spawn`. This means that our `execve` was not being called and so we could not patch SIP binaries, and we would therefore not load into the new process.
So now we also hook `posix_spawn`.

This PR also adds a regression test. The test was verified to be meaningful by checking that it indeed fails without the new hook: https://github.com/t4lz/mirrord/actions/runs/4028503663/jobs/6925471767#step:19:283.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -414,6 +414,10 @@ jobs:
       - run: pip3 install uvicorn[standard] # For http mirroring test with FastAPI.
       - run: cargo build -p mirrord-layer # Build layer lib. The tests load it into the apps.
       - run: cargo test -p mirrord-layer
+      - uses: actions/setup-node@v3 # version 19 spawns processes with `posix_spawn`, so test that also.
+        with:
+          node-version: 19
+      - run: cargo test -p mirrord-layer --test spawn
 
   e2e:
     runs-on: ubuntu-latest
--- a/mirrord/layer/src/exec.rs
+++ b/mirrord/layer/src/exec.rs
@@ -2,10 +2,10 @@
 
 use std::{
     env,
-    ffi::{CStr, CString},
+    ffi::{c_void, CStr, CString},
 };
 
-use libc::{c_char, c_int};
+use libc::{c_char, c_int, pid_t};
 use mirrord_layer_macro::hook_guard_fn;
 use mirrord_sip::{sip_patch, SipError, TMP_DIR_ENV_VAR_NAME};
 use null_terminated::Nul;
@@ -19,7 +19,8 @@ use crate::{
         Detour,
         Detour::{Bypass, Error, Success},
     },
-    error::{HookError, HookError::Null},
+    error::HookError,
+    exec::patched_args::PatchedSpawnArgs,
     file::ops::str_from_rawish,
     hooks::HookManager,
     replace,
@@ -29,6 +30,13 @@ const MAX_ARGC: usize = 254;
 
 pub(crate) unsafe fn enable_execve_hook(hook_manager: &mut HookManager) {
     replace!(hook_manager, "execve", execve_detour, FnExecve, FN_EXECVE);
+    replace!(
+        hook_manager,
+        "posix_spawn",
+        posix_spawn_detour,
+        FnPosix_spawn,
+        FN_POSIX_SPAWN
+    );
 }
 
 /// Check if the file that is to be executed has SIP and patch it if it does.
@@ -86,6 +94,7 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
                 return Bypass(TooManyArgs);
             }
             let arg_str = raw_to_str(arg)?;
+            trace!("execve arg: {arg_str}");
             ptr_vec.push(
                     arg_str
                         .strip_prefix(&tmp_dir)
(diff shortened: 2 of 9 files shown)


mirrord/layer/src/exec.rs
```diff
@@ -112,6 +121,89 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
...
+    use std::ffi::CString;
+
+    use libc::c_char;
+
+    /// The arguments for `execve` or `posix_spawn` after patching the for side-stepping SIP.
+    /// Since the arguments for those functions are pointers, if we indeed modify the arguments, we
+    /// create new objects, so we have to keep them alive, so we put them in this struct, and the
+    /// detour code has to hold on to this struct until after the call to the original libc
+    /// functions.
+    pub struct PatchedSpawnArgs {
```
aviramha, line 134: I think `PatchedArgs` is okay since the context of this file is `exec`/`spawn`

mirrord/layer/src/exec.rs
```diff
@@ -125,31 +217,26 @@ pub(crate) unsafe extern "C" fn execve_detour(
...
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn posix_spawn_detour(
+    pid: *const pid_t,
+    path: *const c_char,
+    file_actions: *const c_void,
+    attrp: *const c_void,
+    argv: *const *const c_char,
+    envp: *const *const c_char,
+) -> c_int {
+    let patched_args = patch_sip_for_new_process(path, argv);
```
aviramha, line 233: not sure about design/usage here - anyone who calls `patch_sip_for_new_proecss` would get the struct, and would use the methods to get fields.
I don't see a reason to have the caller have to call the external interface, since both detours have the same flow - they require path + argv, so I suggest either returning a tuple or change the fields of the struct to be `path, argv` and make those public.

mirrord/layer/src/exec.rs
```diff
@@ -125,31 +217,26 @@ pub(crate) unsafe extern "C" fn execve_detour(
...
-    let final_argv = intercept_tmp_dir(argv_arr)
-        .map(|new_vec| {
-            ptr_vec = new_vec; // Make sure vector still lives when we pass the pointer to execve.
-            ptr_vec.as_ptr()
-        })
-        .unwrap_or(argv);
-
-    FN_EXECVE(final_path, final_argv, envp)
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn posix_spawn_detour(
```
aviramha, line 225: add todo for spawnp?

.github/workflows/ci.yaml
```diff
@@ -393,6 +393,10 @@ jobs:
       - run: pip3 install uvicorn[standard] # For http mirroring test with FastAPI.
       - run: cargo build -p mirrord-layer # Build layer lib. The tests load it into the apps.
       - run: cargo test -p mirrord-layer
+      - uses: actions/setup-node@v3 # version 19 spawns processes with `posix_spawn`, so test that also.
```
eyalb181, line 396: So is the bug the result of our not supporting node 19? Because I did try to run locally with node 19 and didn't run into any problems.

aviramha (commented):
great work, few design/implementation change suggesitons.

.github/workflows/ci.yaml
t4lz: On my machine the bug reproduced when I ran `brew install node` which installed node 19 (alongside a non-homebrew installation I have on my system) and went away whenever I ran `brew uninstall node`. Can it be that even when you installed node 19 yarn used a different version? Or maybe somehow you have a different binary for node 19 than me?
Could you please run again with node 19 and send me the logs?
And what do you get when you run `nm /opt/homebrew/bin/node | grep -e spawn -e exec | grep " U "` (with the path to wherever node 19 is on your system).

mirrord/layer/src/exec.rs
t4lz: done.

mirrord/layer/src/exec.rs
t4lz: Added. If we're lucky `posix_spawnp` calls `posix_spawn` the same way `execvp` calls `execve`.

.github/workflows/ci.yaml
eyalb181: `node -v` returns v19.5.0, and `which node` returns /opt/homebrew/bin/node
The command you asked me to run returned `                 U _uv_spawn`
Logs are attached:
[logs.log](https://github.com/metalbear-co/mirrord/files/10535412/logs.log)

t4lz: It seems that on your machine, when node spawns `/bin/sh -c node .`, (at least that is what it does not my system), mirrord is loaded into that without hooking and patching SIP. I don't know why this is possible, since `/bin/sh` should probably be SI-Protected. If you'd compile from this branch and run again, we'd have a little more information, since there are a couple of new logs here that say what executables mirrord loads into, with what args etc.

eyalb181: See attached.
[logs2.log](https://github.com/metalbear-co/mirrord/files/10537006/logs2.log)

mirrord/layer/src/exec.rs
aviramha: @meowjesty halp

mirrord/layer/src/exec.rs
```diff
@@ -112,6 +121,89 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
     Bypass(TempDirEnvVarNotSet) // Temp dir was not set. Cannot intercept. Should not happen.
 }
 
+mod patched_args {
```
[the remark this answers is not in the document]
aviramha: if it's not pub then it's hidden? you can also do pub(crate)/pub(super)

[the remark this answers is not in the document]
t4lz: Not unless the struct is defined in a separate module. [If you remove that module and define the struct directly there - the "private" fields of that struct can be accessed freely in that file](https://github.com/metalbear-co/mirrord/commit/ebf0c38af089e1f2919b7d5ec244b32ccaad967c#diff-21f26db6a9c9c6adc7e947255b7e19a148bf19f1087505c70c64040bb5a15959R214) ([this compiles](https://github.com/t4lz/mirrord/actions/runs/4045720549/jobs/6957585934)).

mirrord/layer/src/exec.rs
meowjesty: I'm not in tune with the SIP handling yet, so why do we need to keep these pointers alive like that anyway? I mean, these pointers are coming from C, right? So we're not really keeping them alive when we hold them (we don't get to decide when they're `free`d,).

meowjesty: I'm not getting why a tuple like [this](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=d40fc021c5726d11c6727ab9e0cba981) doesn't work?

t4lz: > why do we need to keep these pointers alive like that anyway? I mean, these pointers are coming from C, right?

If we do a SIP patch, we don't use the original pointer, but create a new string (for path) or array of strings (for argv).
Example:

1. Application calls `execve` with path `/bin/bash`.
2. mirrord's `execve_detour` is executed:
  i. We check `/bin/bash` and determine it is SI-Protected.
  ii. We create  a patched binary for `/bin/bash` at `/tmp/mirrord-bin/bin/bash`.
  iii. We call the original `execve` with a pointer to the path of the patched binary. That is a String we created. If we drop that string, we're giving `execve` an invalid pointer as far as I can tell.

t4lz: > I'm not getting why a tuple like [this](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=d40fc021c5726d11c6727ab9e0cba981) doesn't work?

So we could return a 4-tuple with path pointer, optional new path string, argv pointer and optional argv vector.
This would have the same hazard as mentioned above, where it would allow destructing the tuple with `..`, thus dropping the actual objects and keeping two nice dangling pointers. Even if we wanted to ignore that problem and hope it will be ok, I do not see how returning a 4-tuple with 2 fields you must not use but also not drop is preferable over returning a struct and accessing the relevant values by a getter function.

.github/workflows/ci.yaml
t4lz: Just to make sure, could you please run `ls -lO /bin/sh`? (That's a capital o)

mirrord/layer/src/exec.rs
meowjesty: Alright, think I get it now.

This is kinda whats going on right now:

```rust
{
    // patched_path lifetime starts here
    let P { original_path, patched_path, .. } = patch_sip_for_new_process(path, argv);
    FN_EXECVE(patched_path.or(original_path), ...)

    // patched_path dropped
    // original_path dangling
}
```

1. You need `patched_path` to live longer?
2. Are we responsible for freeing `original_path`?

t4lz: > ```rust
>     // original_path dangling
> ```

No, that's not what's happening.

The dangling pointer scenario would only happen if we change the code to return a tuple / make the members public:

```Rust
    let (
        path_ptr, // (maybe) a raw pointer to another item of this tuple.
        argv_ptr,
        .. // path_ptr points to an object that is dropped here.
    ) = patch_sip_for_new_process(original_path_ptr, original_argv_ptr);

    FN_EXECVE(path_ptr, argv_ptr, envp_ptr); // Calling libc with the dangling pointer path_ptr.
```

t4lz: > * You need `patched_path` to live longer?

We need to pass `FN_EXECVE` raw pointers to objects we create in the detour. We get the pointers from the `patch_sip_for_new_process`. As far as I can tell, in order for those pointers to be valid, also the pointed objects need to be moved to the caller. The discussion here is about what's the best interface for that.

t4lz: > 2\. Are we responsible for freeing `original_path`?

no

t4lz: I could keep lifetime managed references to the private pointers and make them public, which would make it possible to access the pointers without calling a function, but should also prevent the struct from being destructed.

.github/workflows/ci.yaml
t4lz: We investigated this and it turns out the reason the bug did not reproduce with node 19 for Eyal is that SIP was disabled on his machine.

mirrord/layer/src/exec.rs
t4lz: it was decided to merge this as is and release asap and then redo it for the next release.

aviramha (approved):
bors r+

Issue #994, opened 2023-01-31
go20 fileops flakiness


aviramha:
### Bug Description

the CI fails too many times with the following
```
stderr 2023-01-31T14:24:11.588117950Z 14737: 2023-01-31T14:24:11.588036Z TRACE ThreadId(03) handle_daemon_message:pop_send: mirrord_layer::file: enter daemon_message=File(Write(Ok(WriteFileResponse { written_amount: 445 }))) value=Ok(WriteFileResponse { written_amount: 445 })

stderr 2023-01-31T14:24:11.588309849Z 14737: 2023-01-31T14:24:11.588115Z TRACE ThreadId(01) mirrord_layer::go_hooks: c_abi_syscall6_handler: syscall=3 param1=15 param2=0 param3=0 param4=0 param5=0 param6=0
2023-01-31T14:24:11.588170Z TRACE ThreadId(01) mirrord_layer: Closing fd 15

deleting "http-echo-9vpgfsa"
deleting "http-echo-9vpgfsa"
thread 'file_ops::file_ops::test_file_ops::agent_1_Agent__Ephemeral::ops_4_FileOps__Go20' panicked at 'Timeout 240s expired', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rstest-0.16.0/src/timeout.rs:27:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```	

### Steps to Reproduce

run e2e ci

### Your operating system and version

Linux

### Local process

go 20 fileops

aviramha:
tbh I'm not sure we should have that many variants in the e2e - both of job/ephemeral and applications but I do want us to understand why it happens on Go20 specifically as there might be an underlying issue.
I'd split this issue into two:
1. figure out why this happens only on it
2. reduce the overlapping tests; (re)move fileops tests that overlap with integration, leave in the e2e for file ops only agent/ephemeral with one type of application

t4lz:
Deadlock cause was fixed in #1009, Go Fileops tests were converted to integration tests in #1019. Tracking the conversion of the other tests for other languages in #1032.

PR #996, opened 2023-01-31
[Merged by Bors] - comment out fileops::go20 until we fix it

aviramha:


aviramha:
Should uncomment it on #994

eyalb181 (approved):

Issue #917, opened 2023-01-09
Improve documentation of mirrord/layer


meowjesty:
We should strive to get the whole of mirrord documented, and passing a `[deny(missing_docs)]` check.

Bellow is a list of the main modules and their documentation progress.

- [ ] Add documentation to mirrord-layer so that `#![deny(clippy::missing_docs_in_private_items)]` passes:
  - [x] `common`;
  - [x] `connection`;
  - [x] `detour`;
  - [x] `dns`;
  - [ ] `error`;
  - [ ] `exec`;
  - [ ] `file`;
  - [ ] `hooks`;
  - [x] `macros`;
  - [ ] `outgoing`;
  - [ ] `socket` (needs better module docs);
  - [ ] `tcp`;
  - [ ] `tcp_mirror`;
  - [ ] `tcp_steal`;
  - [ ] `tracing_util`;
  - [ ] `go_hooks`

PR #998, opened 2023-01-31
[Merged by Bors] - Update CI to build and deploy rust docs to gh-pages.

meowjesty:
- Issue: #917 

Adds a workflow to push rust docs to the `gh-pages` branch.

The docs are built and checked without deploying during the normal workflow, and are built and deployed when pushed to the `main` branch.


aviramha (approved):
bors r+

Issue #956, opened 2023-01-17
Rust binary from `cargo run` still running locally after killing mirrord with ctrl+c and thus mirrord agent does not get removed on k8s


timonbimon:
### Bug Description

I noticed that when I run mirrord with cargo watch and then hit ctrl+c to kill the process, the mirrord agent doesn't disappear. after some digging I found that it's because the underlying binary (in my case from our own crate epibator) is still running on my mac. once killed it also shuts down the mirrord agent.

```
❯ mirrord exec -a dev-`whoami` -n dev-`whoami` -t deployment/backend-core --fs-mode local --steal cargo -- watch -x 'run --bin epibator'
⠼ mirrord cli starting
New mirrord version available: 3.19.1. To update, run: `curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash`.
⠹ mirrord cli starting
  ✓ ready to launch process
    ✓ layer extracted
    ✓ agent pod created
    ✓ pod is ready
[Running 'cargo run --bin epibator']
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/epibator`
{"timestamp":"2023-01-16T23:25:29.129597Z","level":"INFO","fields":{"message":"Running migrations for tenant testing"},"target":"epibator::db_connection","threadName":"main"}
{"timestamp":"2023-01-16T23:25:29.425890Z","level":"INFO","fields":{"message":"Running migrations for tenant luminovo"},"target":"epibator::db_connection","threadName":"main"}
{"timestamp":"2023-01-16T23:25:29.805984Z","level":"INFO","fields":{"message":"Starting server at 0.0.0.0:4999"},"target":"epibator","threadName":"main"}
^C
```

When running `cargo -- watch -x 'run --bin epibator'` without mirrord and hitting ctrl+c it does correctly kill the process from the epibator binary as well.

### Steps to Reproduce

Since you guys also use Rust, the first thing to try would probably be with any Rust binary of your own to see if you can re-produce it.

### Your operating system and version

macOS Moneterey, 12.5.1

### Local process

/Users/timon/.cargo/bin/cargo: Mach-O 64-bit executable arm64

### Additional Info

mirrord 3.18.2

eyalb181:
Hey @timonbimon, thanks for reporting this! We've managed to reproduce this, will let you know when we release a solution.

DmitryDodzin:
[...]

PR #1000, opened 2023-02-01
[Merged by Bors] - Add sip exec workaround for skipped processes

DmitryDodzin:
relevant for #956 with `--skip-processes cargo-watch`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ### Changed
 
-- mirrord-layer: partialy load mirrord on certian processes that spawn other processes to allow sip patch on the spawned process to prevent breaking mirrord-layer load if parent process is specified in `--skip-processes`.
+- mirrord-layer: partialy load mirrord on certian processes that spawn other processes to allow sip patch on the spawned process.
+  This to prevent breaking mirrord-layer load if parent process is specified in `--skip-processes`.
 
 ## 3.21.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ### Changed
 
-- mirrord-layer: partialy load mirrord on certian processes that spawn other processes to allow sip patch on the spawned process.
-  This to prevent breaking mirrord-layer load if parent process is specified in `--skip-processes`.
+- mirrord-layer: Partialy load mirrord on certian processes that spawn other processes to allow sip patch on the spawned process.
+  This to prevent breaking mirrord-layer load if parent process is specified in `--skip-processes`.  (MacOS only)
 
 ## 3.21.0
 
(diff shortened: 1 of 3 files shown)


mirrord/layer/src/load.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+pub fn load_type(given_process: &str, config: LayerConfig) -> LoadType {
+    let skip_processes = config.skip_processes.clone().map(VecOrSingle::to_vec);
+
+    if should_load(given_process, skip_processes) {
+        LoadType::Full(Box::new(config))
+    } else {
+        #[cfg(target_os = "macos")]
+        if sip::is_sip_exec(given_process) {
```
aviramha, line 29: I don't agree with the terminology - I think this should be `is_sip_only`

mirrord/layer/src/load.rs
```diff
@@ -0,0 +1,77 @@
+use mirrord_config::{util::VecOrSingle, LayerConfig};
+
+#[cfg(target_os = "macos")]
+mod sip {
```
aviramha, line 4: Add docs explaining this (why we do that)

mirrord/layer/src/load.rs
```diff
@@ -0,0 +1,77 @@
...
+    use std::{collections::HashSet, sync::LazyLock};
+
+    static SIP_EXEC_PROCESSES: LazyLock<HashSet<&str>> =
+        LazyLock::new(|| HashSet::from(["sh", "bash", "cargo", "rustc", "cargo-watch"]));
```
aviramha, line 8: I don't agree on the sh/bash - it might be beneficial for users to load into those.

aviramha (changes requested):

mirrord/layer/src/load.rs
DmitryDodzin: it's only if specified in `--skip-processes sh`

aviramha: ah, okay need better docs 🖌️

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## [Unreleased]
 
+### Changed
+
+- mirrord-layer: partialy load mirrord on certian processes that spawn other processes to allow sip patch on the spawned process.
```
aviramha, line 12: add (macOS only)

mirrord/layer/src/load.rs
```diff
@@ -0,0 +1,84 @@
+use mirrord_config::{util::VecOrSingle, LayerConfig};
+
+/// For processes that spawn other processes whe should also patch sip on the spawned instances,
```
aviramha, line 3: we*

mirrord/layer/src/load.rs
```diff
@@ -0,0 +1,84 @@
...
+/// curretly limiting to list from `SIP_ONLY_PROCESSES`
+#[cfg(target_os = "macos")]
```
aviramha, line 5: Add a comment explaining that this list is being hit only when it's also in the `skip processes` list

aviramha (approved):
gj!
bors r+

Issue #999, opened 2023-02-01
Redo Exec/Spawn detour


t4lz:
Eliminate the `PatchedArgs` struct and return a tuple instead.

PR #1001, opened 2023-02-01
[Merged by Bors] - Redo spawn detour logic.

t4lz:
#999

### DIFF
--- a/mirrord/layer/src/exec.rs
+++ b/mirrord/layer/src/exec.rs
@@ -5,7 +5,7 @@ use std::{
     ffi::{c_void, CStr, CString},
     marker::PhantomData,
     mem::MaybeUninit,
-    ptr::null,
+    ptr,
 };
 
 use itertools::Itertools;
@@ -89,15 +89,24 @@ struct StringPtr<'a> {
 }
 
 impl Argv {
-    fn null_vec<'a>(&'a self) -> Vec<StringPtr<'a>> {
-        self
+    fn null_string_ptr() -> StringPtr<'static> {
+        StringPtr {
+            ptr: ptr::null(),
+            ..Default::default()
+        }
+    }
+
+    fn null_vec(&self) -> Vec<StringPtr> {
+        let mut vec = self
             .0
             .iter()
             .map(|cstr| StringPtr {
                 ptr: cstr.as_ptr(),
-                _phantom: PhantomData::default(),
+                ..Default::default()
             })
-            .collect()
+            .collect();
+        vec.push(Self::null_string_ptr());
+        vec
     }
 }
 
@@ -172,7 +181,6 @@ unsafe fn patch_sip_for_new_process(
     Success((path_c_string, argv_vec))
 }
 
-
 /// Hook for `libc::execve`.
 ///
 /// Patch file if it is SIPed, used new path if patched.
@@ -188,10 +196,7 @@ pub(crate) unsafe extern "C" fn execve_detour(
 ) -> c_int {
     match patch_sip_for_new_process(path, argv) {
         Success((new_path, new_argv)) => {
-            let mut new_argv = new_argv.null_vec();
-            new_argv.push(StringPtr{ptr: std::ptr::null(), _phantom: PhantomData::default()});
-            // Here we assume that in the memory - a buffer of `CStrings` is exactly the same
-            // as a buffer of the respective `.as_ptr()`s of those `CString`s.
+            let new_argv = new_argv.null_vec();
             let res = FN_EXECVE(
                 new_path.as_ptr(),
                 new_argv.as_ptr() as *const *const c_char,
@@ -216,19 +221,15 @@ pub(crate) unsafe extern "C" fn posix_spawn_detour(
 ) -> c_int {
     match patch_sip_for_new_process(path, argv) {
         Success((new_path, new_argv)) => {
-            let mut new_argv = new_argv.null_vec();
-            new_argv.push(StringPtr{ptr: std::ptr::null(), _phantom: PhantomData::default()});
+            let new_argv = new_argv.null_vec();
             let res = FN_POSIX_SPAWN(
                 pid,
                 new_path.as_ptr(),
                 file_actions,
                 attrp,
-                // Here we assume that in the memory - a buffer of `CStrings` is exactly the same
-                // as a buffer of the respective `.as_ptr()`s of those `CString`s.
                 new_argv.as_ptr() as *const *const c_char,
                 envp,
             );
-
 
             res
         }
--- a/mirrord/layer/src/exec.rs
+++ b/mirrord/layer/src/exec.rs
@@ -132,10 +132,7 @@ unsafe fn patch_sip_for_new_process(
 
     // Continue even if there were errors - just run without patching.
     let path_c_string = patch_if_sip(path_str)
-        .and_then(|w| {
-            // TODO: is there a shorter way to convert a Result into a Detour in this situation?
-            CString::new(w).map_or_else(|err| Error(Null(err)), Success)
-        })
+        .and_then(|w| Success(CString::new(w)?))
         .unwrap_or(
             CString::new(path_str.to_string()).unwrap(),
             // unwrap(): path_str was created from CString so it won't have any nulls.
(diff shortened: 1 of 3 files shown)


mirrord/layer/src/exec.rs
```diff
@@ -96,108 +95,66 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
...
     let exe_path = env::current_exe()
         .map(|path| path.to_string_lossy().to_string())
         .unwrap_or_default();
     trace!("Executable {} called execve/posix_spawn", exe_path);
 
-    // Do unsafe part of path conversion here.
-    let rawish_path = (!path.is_null()).then(|| CStr::from_ptr(path));
+    // TODO: should I change this function to return a result, and if exec/spawn are called with a
+    //       null pointer we could just pass it on instead of panicking?
+    let path_str = raw_to_str(&path).expect("Exec/Spawn hook got null pointer for path.");
```
t4lz, line 131: What do you think? That would mean having (duplicate) handling code in the `..._detour` functions.

How would you handle an error here?

mirrord/layer/src/exec.rs
```diff
@@ -96,108 +95,66 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
...
     // Continue even if there were errors - just run without patching.
-    let path_c_string = patch_if_sip(rawish_path)
-        .success()
-        .map(CString::new)
-        .transpose()
-        .unwrap_or_default();
+    let path_c_string = patch_if_sip(path_str)
+        .and_then(|w| {
+            // TODO: is there a shorter way to convert a Result into a Detour in this situation?
+            CString::new(w).map_or_else(|err| Error(Null(err)), |c_string| Success(c_string))
```
t4lz, line 137: Is there already a good way to convert a result into a detour? If not, should I create one?

mirrord/layer/src/exec.rs
```diff
@@ -96,108 +95,66 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
...
     let argv_arr = Nul::new_unchecked(argv);
 
-    // If we intercept args, we create a new array.
-    // execve takes a null terminated array of char pointers.
-    // ptr_vec will own the vector that will be passed to execve as an array.
-    let argv_vec = intercept_tmp_dir(argv_arr).success();
+    // TODO: if we want this function to return a Vec, we have to somehow return a vec also if there
+    //      is some problem in intercept_tmp_dir, up until now we just returned the original pointer
+    //      and let libc handle the error. Now we return an empty vector. Is this ok?
+    let argv_vec = intercept_tmp_dir(argv_arr).unwrap_or_default();
```
t4lz, line 149: What do you think? Should we just panic instead of returning an empty vector?

I liked the current solution of just passing on the original pointer if we had a problem, but I don't think we can do that now.

mirrord/layer/src/exec.rs
```diff
@@ -96,108 +95,66 @@ fn intercept_tmp_dir(argv_arr: &Nul<*const c_char>) -> Detour<Vec<*const c_char>
...
+    (path_c_string, argv_vec)
+}
 
-    PatchedArgs::new(path, path_c_string, argv, argv_vec)
+// TODO: get rid of this.
+fn c_string_vec_to_null_terminated_pointer_vec(vec: &Vec<CString>) -> Vec<*const c_char> {
+    let mut res = vec.iter().map(|c_string| c_string.as_ptr()).collect_vec();
+    res.push(null());
+    res
 }
```
t4lz, line 158: This solution sucks - we iterate the new vector of CStrings another time to create a vector of pointers with a null at its end. Any ideas of how not to do that?

t4lz (commented):
I need a little help here.

mirrord/layer/src/exec.rs
meowjesty: The `?` operator should work, as long as `E: Into<HookError>`.

mirrord/layer/src/exec.rs
meowjesty: Hmm can't this function return a `Detour`, and on null pointer here, you return it as a `Bypass`?

t4lz: Sure. However this means we have to have (potentially duplicate) handling code in the callers of this function. Is that ok?

meowjesty: Do you mean that 

```rust
let (path, argv) = patch_sip_for_new_process(path, argv);
let null_terminated_argv = c_string_vec_to_null_terminated_pointer_vec(&argv);
FN_EXECVE(path.as_ptr(), null_terminated_argv.as_ptr(), envp)
```

has to be something like

```rust
match patch_sip_for_new_process(path, argv) {
    Ok(...) => {
        let null_terminated_argv = c_string_vec_to_null_terminated_pointer_vec(&argv);
        FN_EXECVE(path.as_ptr(), null_terminated_argv.as_ptr(), envp)
    }
    Bypass(...) => {
        let null_terminated_argv = c_string_vec_to_null_terminated_pointer_vec(&argv);
        FN_EXECVE(path.as_ptr(), null_terminated_argv.as_ptr(), envp)
    }
}

```

?

mirrord/layer/src/exec.rs
meowjesty: Why can't we return the original pointer?

If this failed/returns a default value, can't you use `Bypass`?

mirrord/layer/src/exec.rs
t4lz: Yeah, something like that.

mirrord/layer/src/exec.rs
meowjesty: Well do we need to hold a `Vec<CString>`? Can't you convert to `CString` for the inspection, and back to `*const c_char`?

mirrord/layer/src/exec.rs
t4lz: We can't return the original pointer from this function if we want this function to return a `Vec` because that would be a type mismatch.

And yeah, we could return a `Detour` from this function, as we're discussing above.

mirrord/layer/src/exec.rs
meowjesty: If all 3 variants of `Detour` in `patch_sip_for_new_process` return the same type of value, you can do:

```rust
let Success(null_terminated_argv) | Bypass(null_terminated_argv) | Error(null_terminated_argv) = ...
FN_EXECVE(path.as_ptr(), null_terminated_argv.as_ptr(), envp)
```

Maybe you can reduce this even more, with some clever `map`-style usage, but I think this is good enough?

t4lz: I don't think they would return the same type. I mean, if they would, we wouldn't need to return a `Detour` at all, we could just return that type.

mirrord/layer/src/exec.rs
t4lz: If we don't hold the `CString` anywhere and let its binding go out of scope, I would assume the `CString` would be dropped and the raw pointer we took from it is no longer valid. Am I wrong about that?

meowjesty: Nope, you're right.

meowjesty: Maybe `Vec<MaybeUninit<CString>>`?

mirrord/layer/src/exec.rs
meowjesty: Well, if they're all the same function call just extract it into a shared function. We kinda have this for a few detours.

mirrord/layer/src/exec.rs
aviramha: I don't think `MaybeUininit` is right since `Uninit` means any value

mirrord/layer/src/exec.rs
```diff
@@ -72,15 +72,48 @@ fn raw_to_str(raw_str: &*const c_char) -> Detour<&str> {
...
+
+/// This must be memory-same as just a `*const c_char`.
+#[repr(C)]
+struct StringPtr<'a> {
+    ptr: *const c_char,
+    _phantom: PhantomData<&'a ()>,
+}
+
+impl Argv {
+    fn null_string_ptr() -> StringPtr<'static> {
```
meowjesty, line 86: Missing docs.

meowjesty (approved):
`StringPtr` is a nice solution.

👍  just missing docs for a couple of things.

PR #1003, opened 2023-02-02
live share

aviramha:

Please make sure you added a CHANGELOG entry.

### DIFF
--- a/mirrord/layer/src/exec.rs
+++ b/mirrord/layer/src/exec.rs
@@ -132,10 +132,7 @@ unsafe fn patch_sip_for_new_process(
 
     // Continue even if there were errors - just run without patching.
     let path_c_string = patch_if_sip(path_str)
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/exec.rs
```diff
@@ -213,8 +186,22 @@ pub(crate) unsafe extern "C" fn execve_detour(
     argv: *const *const c_char,
     envp: *const *const c_char,
 ) -> c_int {
-    let patched_args = patch_sip_for_new_process(path, argv);
-    FN_EXECVE(patched_args.get_path(), patched_args.get_argv(), envp)
+    match patch_sip_for_new_process(path, argv) {
+        Success((new_path, new_argv)) => {
+            let mut new_argv = new_argv.null_vec();
```
t4lz, line 191: Don't the pointers outlive the strings right here?

Issue #997, opened 2023-01-31
Fix E2E Flakiness with test_complete_passthrough


aviramha:
The issue there is race with agent removal/spawn of new one - we right now use sleeps. The best mechanism would be to wait for the agent pod to get removed instead of sleeping. There's another possible issue in which we don't connect in time (then we need to increase timeout).

meowjesty:
(hopefully) Fixed by #1004

PR #1004, opened 2023-02-02
[Merged by Bors] - Remove passthrough meta tests.

meowjesty:
- Issue: #997 

Removes the start of a new process from the couple of stealer pass-through tests (`test_complete_passthrough` and `test_websocket_upgrade`).

With this we no longer check that the stolen traffic reaches the user's process (at least not in the same tests).


aviramha (approved):
bors r+

PR #1009, opened 2023-02-03
[Merged by Bors] - Fix rare file close deadlock.

t4lz:
The deadlock has only been observed in the E2E fileops test with Go-1.20 on GitHub Actions.

This PR hopefully eliminate this deadlock, but also contains some fixes of problematic lockings that did not cause this deadlock.

Part of #994.

### DIFF
--- a/mirrord/layer/src/file/ops.rs
+++ b/mirrord/layer/src/file/ops.rs
@@ -30,7 +30,9 @@ impl RemoteFile {
 
 impl Drop for RemoteFile {
     fn drop(&mut self) {
-        trace!("dropping RemoteFile {self:?}");
+        // TODO: put a warning about logging from here.
+        // TODO: delete commented out log.
+        // trace!("dropping RemoteFile {self:?}");
         let closing_file = Close { fd: self.fd };
 
         if let Err(err) = blocking_send_file_message(FileOperation::Close(closing_file)) {
--- a/mirrord/layer/src/file/ops.rs
+++ b/mirrord/layer/src/file/ops.rs
@@ -242,8 +242,11 @@ pub(crate) fn fdopendir(fd: RawFd) -> Detour<usize> {
     // usize == ptr size
     // we don't return a pointer to an address that contains DIR
 
-    let mut open_files = OPEN_FILES.lock()?;
-    let remote_file_fd = open_files.get(&fd).ok_or(Bypass::LocalFdNotFound(fd))?.fd;
+    let remote_file_fd = OPEN_FILES
+        .lock()?
+        .get(&fd)
+        .ok_or(Bypass::LocalFdNotFound(fd))?
+        .fd;
 
     let (dir_channel_tx, dir_channel_rx) = oneshot::channel();
 
@@ -263,7 +266,7 @@ pub(crate) fn fdopendir(fd: RawFd) -> Detour<usize> {
         .insert(local_dir_fd as usize, remote_dir_fd);
 
     // According to docs, when using fdopendir, the fd is now managed by OS - i.e closed
-    open_files.remove(&fd);
+    OPEN_FILES.lock()?.remove(&fd);
 
     Detour::Success(local_dir_fd as usize)
 }
--- a/mirrord/layer/src/file/ops.rs
+++ b/mirrord/layer/src/file/ops.rs
@@ -30,13 +30,18 @@ impl RemoteFile {
 
 impl Drop for RemoteFile {
     fn drop(&mut self) {
-        // TODO: put a warning about logging from here.
-        // TODO: delete commented out log.
-        // trace!("dropping RemoteFile {self:?}");
+        // Warning: Don't log from here. This is called when self is removed from OPEN_FILES, so
+        // during the whole execution of this function, OPEN_FILES is locked.
+        // When emitting logs, sometimes a file `write` operation is required, in order for the
+        // operation to complete. The write operation is hooked and at some point tries to lock
+        // `OPEN_FILES`, which means the thread deadlocks with itself (we call
+        // `OPEN_FILES.lock()?.remove()` and then while still locked, `OPEN_FILES.lock()` again)
         let closing_file = Close { fd: self.fd };
 
         if let Err(err) = blocking_send_file_message(FileOperation::Close(closing_file)) {
-            warn!("Failed to send close file {self:?} message: {err:?}");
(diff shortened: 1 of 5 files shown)


mirrord/layer/src/file/ops.rs
```diff
@@ -30,11 +30,18 @@ impl RemoteFile {
...
+        // during the whole execution of this function, OPEN_FILES is locked.
+        // When emitting logs, sometimes a file `write` operation is required, in order for the
+        // operation to complete. The write operation is hooked and at some point tries to lock
+        // `OPEN_FILES`, which means the thread deadlocks with itself (we call
+        // `OPEN_FILES.lock()?.remove()` and then while still locked, `OPEN_FILES.lock()` again)
         let closing_file = Close { fd: self.fd };
 
         if let Err(err) = blocking_send_file_message(FileOperation::Close(closing_file)) {
-            warn!("Failed to send close file {self:?} message: {err:?}");
+            println!(
```
t4lz, line 42: Is printing instead of emitting a warning a good enough solution?
I'm not sure that printing is safer than logging, but when I fixed the problematic `trace` above with a `println` the deadlock did not happen (for many tests).
Other alternatives would be not to show any message at all, or just to risk it with the warning log, because the deadlock does not happen every time, and this situation (sending message to main layer thread fails) should already be very rare. Another alternative would be to `expect`, because if we can't send messages to the main thread maybe we should exit anyways.

mirrord/layer/src/file/ops.rs
```diff
@@ -30,11 +30,18 @@ impl RemoteFile {
 
 impl Drop for RemoteFile {
     fn drop(&mut self) {
-        trace!("dropping RemoteFile {self:?}");
```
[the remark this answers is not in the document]
t4lz: Does anyone know why doesn't `hook_guard_fn` prevent `write` from being hooked during the execution of the file close hook?

mirrord/layer/src/file/ops.rs
```diff
@@ -242,8 +249,11 @@ pub(crate) fn fdopendir(fd: RawFd) -> Detour<usize> {
     // usize == ptr size
     // we don't return a pointer to an address that contains DIR
 
-    let mut open_files = OPEN_FILES.lock()?;
-    let remote_file_fd = open_files.get(&fd).ok_or(Bypass::LocalFdNotFound(fd))?.fd;
+    let remote_file_fd = OPEN_FILES
+        .lock()?
+        .get(&fd)
+        .ok_or(Bypass::LocalFdNotFound(fd))?
+        .fd;
```
t4lz, line 256: Freeing the lock after this line.
This is ok, right? It's not necessary for the code to hold `OPEN_FILES` locked until the end of this function, is it?

mirrord/layer/src/lib.rs
```diff
@@ -788,11 +788,10 @@ pub(crate) fn close_layer_fd(fd: c_int) {
...
     // Remove from sockets, or if not a socket, remove from files if file mode active
-    SOCKETS.lock().unwrap().remove(&fd).is_none().then(|| {
-        if file_mode_active {
-            OPEN_FILES.lock().unwrap().remove(&fd);
-        }
-    });
+    if SOCKETS.lock().unwrap().remove(&fd).is_none() && file_mode_active {
+        // SOCKETS lock dropped, not holding it while locking OPEN_FILES.
+        OPEN_FILES.lock().unwrap().remove(&fd);
+    }
```
t4lz, line 794: By using just a normal `if` instead of `.then(...)` we free the `SOCKETS` lock before locking `OPEN_FILES`.

This was not the cause of this deadlock.

mirrord/layer/src/lib.rs
```diff
@@ -832,8 +831,9 @@ pub(crate) unsafe extern "C" fn close_nocancel_detour(fd: c_int) -> c_int {
 #[cfg(target_os = "linux")]
 #[hook_guard_fn]
 pub(crate) unsafe extern "C" fn uv_fs_close(a: usize, b: usize, fd: c_int, c: usize) -> c_int {
+    let res = FN_UV_FS_CLOSE(a, b, fd, c);
     close_layer_fd(fd);
-    FN_UV_FS_CLOSE(a, b, fd, c)
+    res
```
t4lz, line 836: The documentation of `close_layer_fd` says it should be called after the original close.

mirrord/layer/src/file/ops.rs
```diff
@@ -263,7 +273,7 @@ pub(crate) fn fdopendir(fd: RawFd) -> Detour<usize> {
         .insert(local_dir_fd as usize, remote_dir_fd);
 
     // According to docs, when using fdopendir, the fd is now managed by OS - i.e closed
-    open_files.remove(&fd);
+    OPEN_FILES.lock()?.remove(&fd);
```
t4lz, line 276: Lock (and free) here again, instead of holding locked all the way from the top down here.

t4lz (commented):
Some explanations of the changes.

mirrord/layer/src/file/ops.rs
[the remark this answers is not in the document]
aviramha: possibly other thread

mirrord/layer/src/file/ops.rs
aviramha: agree with the latter

aviramha (approved):

mirrord/layer/src/file/ops.rs
t4lz: Changed to `expect` https://github.com/metalbear-co/mirrord/pull/1009/commits/c74c2151dbf1a8f643c3594188a6384b84b7946c.

mirrord/layer/src/lib.rs
aviramha: just remembered - the reason I did this this way is because calling `FN_UV_FS_CLOSE` causes it to not return to us ever afaik (due to event loop fun)

t4lz: Ok, so I'm reverting this change and adding a comment. Is `close_layer_fd` fine with being called before in this case?

aviramha: yeah, it's overall okay to do so it's just being "tidy" (freeing resources in the reverse order of being acquired, though there's no real constraint)

Issue #1029, opened 2023-02-08
flush existing connections on steal


aviramha:
### Bug Description

on calico k3s we found out that steal doesn't work until established connections are reset, probably due to different behavior in that setup.
possible solution is to have conntrack binary in our image and call `conntrack -F` on each stolen port after adding the rule.

**this should be controlled via env var, at least for the first version, then decide if to make it enabled by default **

### Steps to Reproduce

calico + k3s

### Your operating system and version

n/a

### Local process

n/a

aviramha:
This fixed their issue. Need to decide if we want is by default.

eyalb181:
@meowjesty Discussed this with @aviramha , can you please change it to default to `true`?

PR #1035, opened 2023-02-09
flush after redirect with conntrack

meowjesty:
- Issue: #1029

Adds [`conntrack --flush`](https://manpages.debian.org/testing/conntrack/conntrack.8.en.html) to the agent whenever stealer redirect rule is added.

The feature is gated behind the `flush_connections: bool` config, and the `MIRRORD_AGENT_STEALER_FLUSH_CONNECTIONS` agent env var.

### DIFF
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -449,7 +449,7 @@ impl TcpConnectionStealer {
 
         if first_subscriber && let Ok(port) = steal_port {
             self.iptables()?
-                .add_stealer_iptables_rules(port, self.stealer.local_addr()?.port())?;
+                .add_stealer_iptables_rules(port, self.stealer.local_addr()?.port()).await?;
         }
 
         self.send_message_to_single_client(&client_id, DaemonTcp::SubscribeResult(steal_port))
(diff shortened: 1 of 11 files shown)


meowjesty:
Added an env var for this feature, and you can enable it in mirrord-config with the `flush_connections: true` option.

Have tested it locally (not yet in k3s), and it doesn't seem to be breaking anything.

mirrord/config/src/agent.rs
```diff
@@ -73,6 +73,16 @@ pub struct AgentConfig {
     /// the agent.
     #[config(env = "MIRRORD_PAUSE", default = false, unstable)]
     pub pause: bool,
+
+    /// Temporary fix for issue [#1029]().
```
aviramha, line 77: Start with the description and have the note in the end

```
Flushes existing connections when starting to steal, might fix issues where connections aren't stolen (due to being already established)
Currently used as temproary fix for issue [#1029]()

File: mirrord/agent/src/steal/connection.rs [hunk 2]
```diff
@@ -389,7 +389,15 @@ impl TcpConnectionStealer {
 
     /// Initialize iptables member, which creates an iptables chain for our rules.
     fn init_iptables(&mut self) -> Result<()> {
-        self.iptables = Some(SafeIpTables::new(iptables::new(false).unwrap())?);
+        let flush_connections = std::env::var("MIRRORD_AGENT_STEALER_FLUSH_CONNECTIONS")
```
aviramha, line 392: I had a wrong assumption that we have a file with all env settings for agent in one place and wanted you to add it there but now i see there's no such file so we should have a todo on that because having env spread around the code is evil :)

aviramha (approved):
Great job! We can merge it.

PR #1038, opened 2023-02-09
mirrord-layer: Changed default to read AWS credentials from remote pod.

aviramha:


DmitryDodzin (approved):

mirrord/layer/src/file/filter.rs
```diff
@@ -100,6 +100,7 @@ fn generate_remote_ro_set() -> RegexSet {
         // AWS cli cache
         // "file not exist" for identity caches (AWS)
         r".aws/cli/cache/.+\.json$",
+        r".aws/credentials$",
```
DmitryDodzin, line 103: Nitpick but maybe add `.aws/config` too for default region

aviramha: nice one!

aviramha: nice one!

PR #1040, opened 2023-02-09
IntelliJ-ext: if multiple configs are present, provide a dialog to ch…

### LINKED ISSUE #1030: Selectable mirrord config in ide
In a monorepo where there are many services, we should add a dropdown to select used mirrord config file.
Maybe scan the whole project for mirrord config files based on pattern and display that?

TODO:
- [x] metalbear-co/mirrord#1411 
- [x] metalbear-co/mirrord-vscode#4

infiniteregrets:
closes #1030


intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,38 @@
...
+            })
+        }
+        return actionGroup
+    }
+
+    override fun update(e: AnActionEvent) {
+        val project = e.project ?: throw Error("required to have configured project to use mirrord config.")
+        configPaths = MirrordConfigAPI.searchConfigPaths(project) as ArrayList<String>
+        if (configPaths.size > 1) {
+            configPaths.add(0, "Default Configuration")
```
aviramha, line 33: If default file is used, just hide the config selection.

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,38 @@
...
+                    e.presentation.text = selectedConfig
+                }
+            })
+        }
+        return actionGroup
+    }
+
+    override fun update(e: AnActionEvent) {
+        val project = e.project ?: throw Error("required to have configured project to use mirrord config.")
+        configPaths = MirrordConfigAPI.searchConfigPaths(project) as ArrayList<String>
```
aviramha, line 31: this isn't good; update can be called every half second which means it might add a lot of overhead to call this.
I think the way to go is to register a BulkFileListener that updates the available files and then use it's value.
Please notice that `BulkFileListener` isn't project level so need to filter out files not in the project.

aviramha (changes requested):

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
infiniteregrets: okay so, should we initialize the configs in the constructor and then listen for events, so when a file that is a config file i.e. mirrord.json is moved anywhere/copied/deleted we update our list?

aviramha: yes

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    override fun update(e: AnActionEvent) {
+        e.presentation.apply {
+            project = project ?: e.project
+            if (shouldUpdatePaths) {
+                updatePaths()
+                shouldUpdatePaths = false
+            }
+            isVisible = when (configPaths.size) {
```
aviramha, line 52: `isVisible = configPaths.size > 1`

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.intellij.openapi.actionSystem.ex.ComboBoxAction
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.newvfs.BulkFileListener
+import com.intellij.openapi.vfs.newvfs.events.VFileEvent
+import javax.swing.JComponent
+
+class MirrordConfigDropDown : ComboBoxAction() {
+    companion object {
+        private var project: Project? = null
+        private val configPaths by lazy { arrayListOf<String>() }
```
aviramha, line 13: overall change I think that `configPaths` should be `Option` (or Kotlin alternative?) then that's our flag for initialization

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,65 @@
...
+        fun updatePaths() {
+            project?.takeIf { it.isInitialized }?.let { prj ->
+                val paths = MirrordConfigAPI.searchConfigPaths(prj)
+                configPaths.clear()
+                configPaths.addAll(paths)
+                if (configPaths.isNotEmpty()) {
+                    if (!paths.contains(project?.basePath + "/.mirrord/mirrord.json")) {
+                        configPaths.add(0, "Default Configuration")
+                    }
+                    selectedConfig = "Select Configuration"
```
aviramha, line 26: I don't think we want to have `Select configuration` option rather just have Default Configuration as the default selected, or the first one that is found if not the default.

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
aviramha: also if you do it that way it can be just an accessor instead of a function

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -0,0 +1,65 @@
...
+                else -> true
+            }
+            text = selectedConfig
+        }
+    }
+}
+
+class MirrordConfigFileRefresher : BulkFileListener {
+    override fun before(events: MutableList<out VFileEvent>) {
+        MirrordConfigDropDown.updatePaths()
```
aviramha, line 63: I think we should do filtering here since this might cause very weird behavior

aviramha (changes requested):

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
infiniteregrets: what weird behavior it would cause? it calls updatePaths() which calls the search function doing the filtering

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
infiniteregrets: i think for options in kotlin we need to depend on Arrow? and i believe null values in kotlin have the same meaning as the option type

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
infiniteregrets: oops sorry, this was really silly not sure why i changed that

Issue #1037, opened 2023-02-09
Debugging service that uses tls crashes


Shanik1:
### Bug Description

Mirrord agent version: 3.22.0
Golang version: 2022.3.2

Dubbging an https server pod (TLS) that has the following security context:
```yaml
securityContext:                                                                                                                                                                                                                                                         
│       allowPrivilegeEscalation: false                                                                                                                                                                                                                                        
│       capabilities:                                                                                                                                                                                                                                                          
│         add:                                                                                                                                                                                                                                                                 
│         - NET_BIND_SERVICE                                                                                                                                                                                                                                                   
│         drop:                                                                                                                                                                                                                                                                
│         - ALL                                                                                                                                                                                                                                                                
│       readOnlyRootFilesystem: false                                                                                                                                                                                                                                          
│       runAsUser: 0
```

Traceback in details

Failed when we are trying to create a new certificate in the code


### Steps to Reproduce

?

### Backtrace

```shell
2023/02/09 11:32:16 Getting signed certificate
SIGILL: illegal instruction
PC=0x7ff80f0d54d9 m=14 sigcode=1
instruction bytes: 0xf 0xb 0x48 0x8d 0x5 0xe9 0xeb 0x20 0x0 0x48 0x89 0x5 0xf 0x95 0x66 0x41

goroutine 0 [idle]:
crypto/x509/internal/macos.syscall(0xc0010ee370?, 0x100d3ab?, 0x31e3360?, 0xc00000fe60?, 0xc0010ee3f8?, 0x0?, 0x0?)
        /usr/local/opt/go/libexec/src/runtime/sys_darwin.go:99 +0x58 fp=0xc0010ee338 sp=0xc0010ee2d8 pc=0x106d2d8
crypto/x509/internal/macos.CFRelease(0x0)
        /usr/local/opt/go/libexec/src/crypto/x509/internal/macos/corefoundation.go:152 +0x3d fp=0xc0010ee388 sp=0xc0010ee338 pc=0x136507d
crypto/x509/internal/macos.ReleaseCFArray(0x600000c10480)
        /usr/local/opt/go/libexec/src/crypto/x509/internal/macos/corefoundation.go:204 +0x55 fp=0xc0010ee3c0 sp=0xc0010ee388 pc=0x1365355
crypto/x509.(*Certificate).systemVerify.func2()
[...]

PR #1048, opened 2023-02-11
Added env guard to be used in cli + extension

aviramha:
to prevent (self) misconfigurations (our kube settings being used from remote).


aviramha:
Might be related to #1037

eyalb181 (approved):

PR #1050, opened 2023-02-11
getsockname now returns the pod's local address

aviramha:


mirrord/layer/src/outgoing.rs
```diff
@@ -16,12 +16,15 @@ pub(crate) mod udp;
 
 /// Wrapper type for the (layer) socket address that intercepts the user's socket messages.
 #[derive(Debug)]
-pub(crate) struct MirrorAddress(pub(crate) SocketAddr);
+pub(crate) struct RemoteConnectResult {
```
DmitryDodzin, line 19: I would rename it to `RemoteConnection` or `RemoteAddresses` because suffixing it with Result makes me think it is a type alias for a specific `Result`

mirrord/layer/tests/http_mirroring.rs
```diff
@@ -79,6 +79,7 @@ async fn test_mirroring_with_http(
                     mirrord_protocol::outgoing::DaemonConnect {
                         connection_id: 0,
                         remote_address: std::net::SocketAddr::from(([10, 253, 155, 219], 58162)),
+                        local_address: std::net::SocketAddr::from(([10, 253, 155, 218], 58161)),
```
DmitryDodzin, line 82: I think local_address should be 1.1.1.1 to be more inline with what is returned in NewTcpConnection or is this irrelevant with this test?

mirrord/layer/tests/apps/app_flask.py
```diff
@@ -6,7 +6,7 @@
 from os import getpid, kill
 from signal import SIGTERM
 
-from flask import Flask
+from flask import Flask, request
```
DmitryDodzin, line 9:
```suggestion
from flask import Flask
```

DmitryDodzin (approved):
Only small nitpicks

mirrord/layer/tests/http_mirroring.rs
aviramha: irrelevant (this is udp)

mirrord/layer/src/outgoing.rs
aviramha: great!

mirrord/layer/tests/apps/app_flask.py
aviramha: shouldn't have committed this to begin with, nice catch.

Issue #1041, opened 2023-02-09
Failure when impersonating a service in an Akka cluster


eyalb181:
### Bug Description

When impersonating a pod in an Akka cluster, the local process makes the initial gossip request in order to join the cluster. This communication fails, so the service doesn't join the cluster and everything else fails as well.

### Steps to Reproduce

Impersonate a pod in an Akka cluster

### Relevant Logs

```shell
"Cluster Node [akka.tcp://ClusterSystem@10.0.0.26:2551] - Couldn't join seed nodes after [6] attempts, will try again. seed-nodes=[akka.tcp://ClusterSystem@akka-cluster1:2551, akka.tcp://ClusterSystem@akka-cluster2:2551, akka.tcp://ClusterSystem@akka-cluster3:2551, akka.tcp://ClusterSystem@akka-cluster4:2551]"


Association with remote system [akka.tcp://ClusterSystem@10.99.126.148:2551] has failed, address is now gated for [5000] ms. Reason: [Association failed with [akka.tcp://ClusterSystem@10.99.126.148:2551]] Caused by: [java.net.SocketException: Unknown error: 111]"
```

### Your operating system and version

MacOS

### Local process

Scala

aviramha:
This happens via JetBrains so might be that we just don't load since we don't support sbt?

p.s setting up Akka in k8s
https://github.com/akka/akka-sample-cluster-kubernetes-scala

aviramha:
Managed to reproduce it. Setup cluster:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: appka
  name: appka
  namespace: appka-1
spec:
  replicas: 3
  selector:
    matchLabels:
      app: appka
  template:
    metadata:
      labels:
        app: appka
        actorSystemName: appka
    spec:
      containers:
      - name: appka
        image: docker.io/aviramhassan/akka-sample-cluster-kubernetes:latest
        imagePullPolicy: Always
        readinessProbe:
          httpGet:
            path: /ready
            port: management
          periodSeconds: 10
          failureThreshold: 3
          initialDelaySeconds: 10
        livenessProbe:
          httpGet:
            path: "/alive"
            port: management
          periodSeconds: 10
          failureThreshold: 5
          initialDelaySeconds: 20
        ports:
        # akka-management and bootstrap
        - name: management
          containerPort: 8558
          protocol: TCP
        - name: http
          containerPort: 8080
          protocol: TCP
        env:
[...]

PR #1057, opened 2023-02-12
mirrod-layer: gethostname detour

DmitryDodzin:
gethostname detour that uses the `/host/hostname` file on the remote to get the hostname

relevant for #1041

### DIFF
--- a/mirrord/layer/src/detour.rs
+++ b/mirrord/layer/src/detour.rs
@@ -366,11 +366,11 @@ impl<T> OnceLockExt<T> for OnceLock<T> {
         F: FnOnce() -> Detour<T>,
     {
         if let Some(value) = self.get() {
-            return Detour::Success(value);
+            Detour::Success(value)
+        } else {
+            let value = f()?;
+
+            Detour::Success(self.get_or_init(|| value))
         }
-
-        let value = f()?;
-
-        Detour::Success(self.get_or_init(|| value))
     }
 }
--- a/mirrord/layer/src/detour.rs
+++ b/mirrord/layer/src/detour.rs
@@ -298,6 +298,17 @@ impl<S> Detour<S> {
             _ => default,
         }
     }
+    /// Return the contained `Success` wrapped in `Some` and swallows `Bypass` or `Error` in
+    /// exchange for `None`
+    ///
+    /// To be used in hooks that are deemed non-essential, and the run should continue even if they
+    /// fail.
+    pub(crate) fn ok(self) -> Option<S> {
+        match self {
+            Detour::Success(s) => Some(s),
+            _ => None,
+        }
+    }
 }
 
 impl<S> Detour<S>
--- a/mirrord/layer/src/detour.rs
+++ b/mirrord/layer/src/detour.rs
@@ -354,7 +354,9 @@ impl<T> OptionExt for Option<T> {
     }
 }
 
+/// Extends [`OnceLock`] with a helper function to initialize it with a [`Detour`].
 pub(crate) trait OnceLockExt<T> {
+    /// Initializes a [`OnceLock`] with a [`Detour`] (similar to [`OnceLock::get_or_try_init`]).
     fn get_or_detour_init<F>(&self, f: F) -> Detour<&T>
     where
         F: FnOnce() -> Detour<T>;
--- a/mirrord/layer/src/detour.rs
+++ b/mirrord/layer/src/detour.rs
@@ -9,7 +9,7 @@ use core::{
     convert,
     ops::{FromResidual, Residual, Try},
 };
-use std::{cell::RefCell, ops::Deref, os::unix::prelude::*, path::PathBuf};
+use std::{cell::RefCell, ops::Deref, os::unix::prelude::*, path::PathBuf, sync::OnceLock};
 
 use crate::error::HookError;
 
@@ -85,7 +85,7 @@ impl Drop for DetourGuard {
 /// to simplify calls to the original functions as `FN_ORIGINAL()`, instead of
 /// `FN_ORIGINAL.get().unwrap()`.
 #[derive(Debug)]
-pub(crate) struct HookFn<T>(std::sync::OnceLock<T>);
+pub(crate) struct HookFn<T>(OnceLock<T>);
 
 impl<T> Deref for HookFn<T> {
     type Target = T;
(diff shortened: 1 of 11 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -631,3 +634,47 @@ pub(super) fn getaddrinfo(
...
+            .bytes
+            .into_iter()
+            .take(hostname_file.read_amount as usize)
+            .collect::<Vec<_>>(),
+    )?)
+}
+
+/// Resolve the fake local address to the real local address.
+#[tracing::instrument(level = "trace", skip(raw_name, name_length))]
+pub(super) fn gethostname(raw_name: *mut c_char, name_length: usize) -> Detour<i32> {
```
aviramha, line 664: we prefer `ops` logic to be safe, so I suggest that `gethostname` returns `Detour<String>` then the caller does the actual assignment

aviramha (changes requested):
Design comment

mirrord/layer/src/socket/ops.rs
```diff
@@ -631,3 +634,47 @@ pub(super) fn getaddrinfo(
...
+    let host = match HOSTNAME.get() {
```
aviramha, line 665: why not `let host = HOSTNAME.get_or_init(|| remote_read..)`

mirrord/layer/src/socket/ops.rs
```diff
@@ -631,3 +634,42 @@ pub(super) fn getaddrinfo(
...
+            .collect::<Vec<_>>(),
+    )
+    .map(Detour::Success)?
+}
+
+/// Resolve the fake local address to the real local address.
+#[tracing::instrument(level = "trace")]
+pub(super) fn gethostname() -> Detour<&'static CString> {
+    let hostname = HOSTNAME.get_or_init(|| {
+        remote_hostname_string().ok().or_else(|| {
```
meowjesty, line 667: Are we "hook guarded" here? Because in theory you could just do `std::fs::read_to_string("etc/hostname")` here.

meowjesty: Also, if you're using the `else` case just to log, then you could use `inspect_err` before the `ok` conversion.

meowjesty (changes requested):
Do we need the `remote_hostname_string` function?

mirrord/layer/src/detour.rs
```diff
@@ -353,3 +353,24 @@ impl<T> OptionExt for Option<T> {
...
+    where
+        F: FnOnce() -> Detour<T>;
+}
+
+impl<T> OnceLockExt<T> for OnceLock<T> {
+    fn get_or_detour_init<F>(&self, f: F) -> Detour<&T>
+    where
+        F: FnOnce() -> Detour<T>,
+    {
+        if let Some(value) = self.get() {
```
meowjesty, line 368: Chains?

Or at the very least I would suggest removing `return` and using an `else` clause.

mirrord/layer/src/socket/ops.rs
aviramha: that's cool idea

mirrord/layer/src/detour.rs
DmitryDodzin: Actually this is the only part left from the implementation of `get_or_try_init` but sure Ill change it to an else to prevent complexity with ?

mirrord/layer/src/socket/hooks.rs
```diff
@@ -72,6 +73,27 @@ pub(crate) unsafe extern "C" fn getsockname_detour(
...
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn gethostname_detour(
+    raw_name: *mut c_char,
+    name_length: usize,
+) -> c_int {
+    gethostname()
+        .map(|host| {
+            let host_len = host.as_bytes().len();
+
+            if host_len <= name_length {
```
aviramha, line 85: the implementaiton should be always copy but return error in case of size issue

mirrord/layer/src/detour.rs
```diff
@@ -353,3 +353,24 @@ impl<T> OptionExt for Option<T> {
         }
     }
 }
+
+pub(crate) trait OnceLockExt<T> {
```
meowjesty, line 357:
```rust
/// Extends [`OnceLock`] with a helper function to initialize it with a [`Detour`].
```

mirrord/layer/src/detour.rs
```diff
@@ -353,3 +353,24 @@ impl<T> OptionExt for Option<T> {
...
+    fn get_or_detour_init<F>(&self, f: F) -> Detour<&T>
```
meowjesty, line 358:
```rust
/// Initializes a [`OnceLock`] with a [`Detour`] (similar to [`OnceLock::get_or_try_init`]).
```

mirrord/layer/src/socket/hooks.rs
```diff
@@ -72,6 +73,28 @@ pub(crate) unsafe extern "C" fn getsockname_detour(
         .unwrap_or_bypass_with(|_| FN_GETSOCKNAME(sockfd, address, address_len))
 }
 
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn gethostname_detour(
```
meowjesty, line 77: Warning: missing docs.

mirrord/layer/src/socket/ops.rs
```diff
@@ -5,28 +5,31 @@ use std::{
...
     error::HookError,
-    file::OPEN_FILES,
+    file::{self, OPEN_FILES},
     outgoing::{tcp::TcpOutgoing, udp::UdpOutgoing, Connect, RemoteConnection},
     port_debug_patch,
     tcp::{Listen, TcpIncoming},
     ENABLED_TCP_OUTGOING, ENABLED_UDP_OUTGOING,
 };
 
+pub(crate) static HOSTNAME: OnceLock<CString> = OnceLock::new();
```
meowjesty, line 31: Warning: missing docs.

```rust
/// Hostname initialized from the agent with [`gethostname`].
```

mirrord/layer/src/socket/ops.rs
```diff
@@ -631,3 +634,34 @@ pub(super) fn getaddrinfo(
 
     Detour::Success(result)
 }
+
+fn remote_hostname_string() -> Detour<CString> {
```
meowjesty, line 638: Warning: missing docs.

```rust
/// Retrieves the `hostname` from the agent's `/etc/hostname` to be used by [`gethostname`]
```

meowjesty (changes requested):
Feel free to improve on my doc suggestions.

meowjesty (approved):
:bear: :+1:

PR #1058, opened 2023-02-12
jetbrains: better targets UX

### LINKED ISSUE #1055: Change focus from search to target selection using tab
When entering text to search, implement tab to change focus back and forth.

aviramha:
Closes #1055


t4lz (approved):
Weird that we have to write this handler and it doesn't just work like that by default.

Really awesome that we're supporting mouse-less navigation.

PR #1059, opened 2023-02-13
internal proxy - remove k8s stuff from layer

aviramha:
## Motivation

We have issues with layer using env from remote since the process it is injected to has the remote env, so we had `EnvGuard` and other fixes but now we also see crashes when doing k8s stuff from the layer level.

## Solution

CLI now spawns an `internal proxy` that relays connections from layers to the remote pod.
The basic flow is:
1. cli creates pod
2. spawns internal proxy with information about created pod
3. exec the original process (via ext or alone) that connects to agent using internal proxy

**note** currently there's still 1:1 connections


mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+use crate::error::{InternalProxyError, Result};
+
+const FIRST_CONNECTION_TIMEOUT: u64 = 5;
+
+fn print_port(listener: &TcpListener) -> Result<()> {
+    let port = listener
+        .local_addr()
+        .map_err(InternalProxyError::LocalPortError)?
+        .port();
+    println!("{port:?}\n");
```
DmitryDodzin, line 37:
```suggestion
    println!("{port}\n");
```
I think it's better to Display rather debug the u16

DmitryDodzin (approved):
Small Nitpic

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
+//! Internal proxy is accepting connection from local layers and forward it to agent
+//! while having 1:1 relationship - each layer connection is another agent connection.
+//! This might be changed later on.
+//! The main advantage of this design is that we remove kube logic from the layer itself,
+//! thus eliminating bugs that happen due to mix of remote env vars in our code
+//! (previously was solved using envguard which wasn't good enough)
+//! The proxy will either directly connect to an existing agent (currently only used for tests),
+//! or let the [`OperatorApi`] handle the connection.
```
meowjesty, line 8:
```suggestion
//! Internal proxy is accepting connection from local layers and forward it to agent
//! while having 1:1 relationship - each layer connection is another agent connection.
//!
//! This might be changed later on.
//!
//! The main advantage of this design is that we remove kube logic from the layer itself,
//! thus eliminating bugs that happen due to mix of remote env vars in our code
//! (previously was solved using envguard which wasn't good enough)
//!
//! The proxy will either directly connect to an existing agent (currently only used for tests),
//! or let the [`OperatorApi`] handle the connection.
```

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+    net::{TcpListener, TcpStream},
+    select,
+    sync::mpsc::{self},
+    task::JoinSet,
+    time::timeout,
+};
+
+use crate::error::{InternalProxyError, Result};
+
+const FIRST_CONNECTION_TIMEOUT: u64 = 5;
```
meowjesty, line 30: Warning: missing docs.

meowjesty: Maybe document something related to the [first connection](https://github.com/metalbear-co/mirrord/pull/1059/files#diff-53987291a539547efc68f94456805ec79cec10fdb79bcc7f4a660abcbd4ba2fbR447) notice in the agent?

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+
+fn print_port(listener: &TcpListener) -> Result<()> {
```
meowjesty, line 32: Warning: missing docs.

meowjesty: Not obvious to me why we need this? Do we rely on `stdout` to know the port?

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+        .local_addr()
+        .map_err(InternalProxyError::LocalPortError)?
+        .port();
+    println!("{port}\n");
+    Ok(())
+}
+
+/// Supposed to run as an async detached task, proxying the connection.
+/// We parse the protocol so we might add some logic here in the future?
+async fn handle_connection(
```
meowjesty, line 43: Style convention question:

What do you think of enforcing a name like `x_task` for every function that we have running as a detached task? Currently I think most of what we have are either `handle` or `handler`.

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+                layer_connection.send(agent_message).await.expect("failed to send agent message to layer");
+            }
+            else => {
+                break;
+            }
+        }
+    }
+}
+
+pub(crate) async fn proxy() -> Result<()> {
```
meowjesty, line 64: Warning: missing docs.

mirrord/cli/src/internal_proxy.rs
meowjesty: Oh now I see the comment on printing this to the user.

So something along the lines of:

```rust
/// Displays the port we've bound to the user.
```

mirrord/cli/src/internal_proxy.rs
```diff
@@ -0,0 +1,128 @@
...
+                if active_connections.is_empty() {
+                    break;
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+async fn connect(
```
meowjesty, line 108: Warning: missing docs.

meowjesty (changes requested):
Some initial comments, mostly about docs.

mirrord/layer/src/connection.rs
```diff
@@ -1,130 +1,91 @@
...
-                )
-                .await
-                .unwrap_or(Err(KubeApiError::AgentReadyTimeout))
-                .unwrap_or_else(|err| handle_error(err, config));
-                info!("Created new agent {pod_agent_name}, port {agent_port}");
-
-                // Set env var for children to re-use.
-                std::env::set_var("MIRRORD_CONNECT_AGENT", &pod_agent_name);
-                std::env::set_var("MIRRORD_CONNECT_PORT", agent_port.to_string());
+pub(crate) async fn connect(addr: SocketAddr) -> (Sender<ClientMessage>, Receiver<DaemonMessage>) {
```
meowjesty, line 29: I would suggest changing this to`connect_to_proxy`, as for 1 second there I thought this was the `connect` hook :sweat_smile:

mirrord/layer/src/connection.rs
```diff
@@ -1,130 +1,91 @@
...
+        Err(e) => {
+            error!("Couldn't connect to internal proxy: {e:?}");
+            graceful_exit!("{FAIL_STILL_STUCK:?}");
+        }
+    };
+    wrap_raw_connection(stream)
+}
 
-                (pod_agent_name, agent_port)
+fn wrap_raw_connection(
```
meowjesty, line 40: Warning: missing docs.

Add docs saying that this creates the task that handles the messaging between layer/agent.

mirrord/layer/src/connection.rs
```diff
@@ -1,130 +1,91 @@
...
+    stream: TcpStream,
+) -> (mpsc::Sender<ClientMessage>, mpsc::Receiver<DaemonMessage>) {
+    let mut codec = actix_codec::Framed::new(stream, ClientCodec::new());
+
+    let (in_tx, mut in_rx) = mpsc::channel(CONNECTION_CHANNEL_SIZE);
+    let (out_tx, out_rx) = mpsc::channel(CONNECTION_CHANNEL_SIZE);
```
meowjesty, line 46: Are `in_tx = ClientMessage` and `out_tx = DaemonMessage`?

```suggestion
    let (client_in_tx, mut client_in_rx) = mpsc::channel(CONNECTION_CHANNEL_SIZE);
    let (daemon_out_tx, daemon_out_rx) = mpsc::channel(CONNECTION_CHANNEL_SIZE);
```

mirrord/kube/src/api.rs
```diff
@@ -14,15 +14,10 @@ use tracing::{error, info};
...
-#[cfg(feature = "env_guard")]
-mod env_guard;
-#[cfg(feature = "env_guard")]
-pub use env_guard::MIRRORD_GUARDED_ENVS;
-
 pub mod kubernetes;
 mod runtime;
 
-static CONNECTION_CHANNEL_SIZE: usize = 1000;
+const CONNECTION_CHANNEL_SIZE: usize = 1000;
```
meowjesty, line 20: Warning: missing docs.

```suggestion
/// Size of the channel used in [`wrap_raw_connection`] for [`ClientMessage`]s.
const CONNECTION_CHANNEL_SIZE: usize = 1000;
```

meowjesty (changes requested):
More nits.

mirrord/kube/src/api.rs
```diff
@@ -36,9 +31,9 @@ where
     }
 }
 
-pub(crate) fn wrap_raw_connection(
+pub fn wrap_raw_connection(
```
meowjesty, line 34: Warning: missing docs.

Even though it was missing before, would  you mind adding docs to this?

meowjesty (changes requested):
Last missing docs warning I can see.

meowjesty (approved):
:bear: :+1: 

Approving this (just gotta fix the busy loop).

mirrord/cli/src/internal_proxy.rs
```diff
@@ -103,7 +122,7 @@ pub(crate) async fn proxy() -> Result<()> {
                 let agent_connection = connect(&config).await?;
                 active_connections.spawn(connection_task(stream, agent_connection));
             },
-            _ = active_connections.join_next() => {},
+            Some(_) = active_connections.join_next(), if !active_connections.is_empty() => {},
             _ = tokio::time::sleep(Duration::from_secs(1)) => {
```
DmitryDodzin, line 125: Im not sure how this will behave because the timeout will always resolve so anything that is None for active_connections and Err for listener is swollen and I think

DmitryDodzin (approved):

mirrord/agent/src/main.rs
```diff
@@ -390,67 +444,71 @@ async fn start_agent() -> Result<()> {
...
-                let client_id = client?;
-                state.remove_client(client_id).await?;
-            },
-            _ = tokio::time::sleep(std::time::Duration::from_secs(args.communication_timeout.into())) => {
-                if state.no_clients_left() {
-                    trace!("Main thread timeout, no clients connected.");
-                    break;
+            // Logic here is if we have already accepted a client, then we drop immediately when all clients disconnect.
+            // if not, we wait for the communication timeout to expire before dropping. (to have more time for the first
+            // client to connect)
```
t4lz, line 499: > we wait for the communication timeout to expire before dropping

Where is that?
> (to have more time for the first client to connect)

Didn't the first client already connect (above the loop) if we reached here?

aviramha: outdated comment, removed

mirrord/cli/src/internal_proxy.rs
```diff
@@ -146,6 +146,29 @@ pub(crate) async fn proxy() -> Result<()> {
...
+
+/// Sends a ping the connection and expects a pong.
+async fn ping(
+    sender: &mut mpsc::Sender<ClientMessage>,
+    receiver: &mut mpsc::Receiver<DaemonMessage>,
+) -> Result<(), InternalProxyError> {
+    sender.send(ClientMessage::Ping).await?;
+    match receiver.recv().await {
+        Some(DaemonMessage::Pong) => Ok(()),
+        _ => Err(InternalProxyError::AgentClosedConnection),
```
DmitryDodzin, line 168: Maybe should be like CliError::InvalidMessage to be more indicative that it didn't get desired pong back

DmitryDodzin (approved):
Error type nitpick =]

PR #1062, opened 2023-02-14
filter adjustments - macOS tmp and fix tmp regex

aviramha:


DmitryDodzin (approved):

mirrord/layer/src/file/filter.rs
```diff
@@ -58,7 +58,9 @@ fn generate_local_set() -> RegexSet {
         r"^/dev/.*$",
         r"^/opt/.*$",
         r"^/home/.*$",
-        r"^/tmp.*$",
+        r"^/tmp/.*$",
+        // macOS' temp
+        r"^/private/var/folders/.*$",
```
t4lz, line 63: Should already be covered here
https://github.com/metalbear-co/mirrord/pull/1062/files#diff-7ad0a3830b8167045e2ea263c517a878b1355dae56b278f9e9c56130e0dc683bR88

But I guess it doesn't hurt to have this specific path regardless of the `TMPDIR`...

t4lz: Did you experience a case where that directory was not covered?

aviramha: Yes, with the IntelliJ issue - I think that tmpdir can be different between processes

PR #1068, opened 2023-02-16
Filter out agent pods from mirrord ls

### LINKED ISSUE #1045: Exclude mirrord-agent pods from target selection dialog in IDE extensions
I don't think any non-mirrord-developer user is ever going to want to impersonate a mirrord-agent pod, so it would increase usability if we don't include those pods in the list of pods to choose from:
image

I don't know if it's worth the extra effort - but it would be nice if we could exclude those pods based on more than just the name. We could e.g. use the image (without the version).

Saw this on JetBrains (PyCharm). As part of this issue it should be checked if the agent pods also appear in vscode, and if they do, open an equivalent issue for vs-code.

infiniteregrets:
closes #1045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 - Fix cache does not work on test-agent workflow. See [#251](https://github.com/metalbear-co/mirrord/issues/251).
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


aviramha (approved):
great job
@DmitryDodzin I think we need the same in the operator too?

DmitryDodzin:
> great job @DmitryDodzin I think we need the same in the operator too?

Agree, right now you can do it manually the target handler supports label selector

Issue #1061, opened 2023-02-14
[ vscode extension ] nodejs script fails to resolve domains when executed via `npx ts-node`


gshaibi:
### Bug Description

I'm trying to debug a typescript application by launching `npx ts-node`.
When using the mirrord vscode extension, the process fails to resolve cluster domains.
Using the CLI with the same npx command works.

In addition, the "select target" dialog on vscode appears twice (I choose the target, and then it requests to choose it again) on that case.

Both bugs don't happen when launching go applications.

versions:
VSCode extension and CLI - 3.24.0
VSCode - 1.75.1

Thanks!

### Steps to Reproduce

1. Enable mirrord vscode extension
2. Install `ts-node`
3. On vscode, debug a typescript file using `npx ts-node`
4. Try to resolve cluster domains on that process

### Relevant Logs

```shell
trace logs:

❯ RUST_LOG=mirrord=trace mirrord exec --target-namespace runai-backend --target deployment/runai-backend-deployment -- npx ts-node js/backend/backendmain.ts
⠹ mirrord cli starting
⠸ mirrord cli starting
  ⠋ preparing to launch process
2023-02-14T09:14:17.882144Z TRACE mirrord_sip::main: Using temp dir: "/var/folders/v1/ys1f7fsj6c13ymy8b9mq8p3w0000gn/T/mirrord-bin" for sip patches
2023-02-14T09:14:17.882436Z TRACE mirrord_sip::main: "/Users/guyshaibi/.nvm/versions/node/v16.19.0/bin/npx" is a script with a shebang that leads to a SIP protected binary.
2023-02-14T09:14:17.882468Z TRACE mirrord_sip::main: Shebang points to: "/usr/bin/env". Patching it recursively and making a version of "/Users/guyshaibi/.nvm/versions/node/v16.19.0/bin/npx" with an altered shebang
⠼ mirrord cli starting
  ⠙ preparing to launch process
  ✓ layer extracted
⠦ mirrord cli starting
⠋ mirrord cli starting
⠙ mirrord cli starting
  ⠇ preparing to launch process
  ✓ layer extracted
⠹ mirrord cli starting
  ✓ ready to launch process
  ✓ layer extracted
  ✓ agent pod created
  ✓ pod is ready
2023-02-14T09:14:19.902406Z  INFO ThreadId(01) mirrord_layer: Initializing mirrord-layer!
[...]

Issue #747, opened 2022-11-17
SIP bypassing for main application also on IDE


t4lz:
We Bypass SIP on two occasions: executing the user's application, and when the application executes other executables.
The first one is currently only done in the CLI, so if a user tries to run a SIP application (Maybe a bash script, if there is IDE support for that?) in the IDE with mirrord, the layer would not be loaded into the application, and the application would run alone.

The second kind is handled also on IDEs, so if the user's application is not SIP, but executes SIP binaries, we do handle that (by hooking `execve`), assuming the executable's path is ignored by the file hooks (otherwise we would try to read the file for SIP handling, then we would hook the file open, try to open it in the cloud, not find it, return "File not found", the SIP detour would fail, and the executable would be executed without the layer loading into it).

If you would like to have support for executing SIP binaries with mirrord in an IDE, please let us know and we will prioritise it :)

t4lz:
Whenever there are developments in this issue, **the part about SIP in the [FAQ](https://github.com/metalbear-co/mirrord.dev/blob/main/content/docs/overview/faq.md) should be updated accordingly.**

t4lz:
In vscode we modify the debug configuration on launch in order to inject the layer. Unfortunately there is no universal "executable" configuration field that specifies what will be the command that is run. The value of the `program` field is just an argument to the executable, when running programs written in interpreted languages (`config.program` might be `main.py`, but the executable that has to run is the python interpreter).
I think we could sidestep SIP by replacing the interpreter, and debugging would still work, but I am not sure of it.
If we want to replace the interpreter the same way we set the environment, by modifying the debug configuration, we would have to have custom code for each interpreted language, and we would only support sidestepping SIP for those debug types. This is because the interpreter config field is different for each debug type.
For example, with the built-in debugger type `node`, [`config.runtimeExecutable` sets the interpreter](https://code.visualstudio.com/docs/nodejs/nodejs-debugging), but [for `python` its `config.python`](https://code.visualstudio.com/docs/python/debugging#_python).
This approach might lead to a situation where users decide to use some new interpreted language on Mac with a dedicated vscode debugger extension, and we would have to quickly add a couple of lines to our extension's code in order to support that debug type. The change itself would not be difficult, we would just have to find out the name of the field that is used to specify the interpreter with that debugger, and add
```typescript
if (config.type === "thisnewdebugger") {
    let patched_interpreter_path = sip_patch(config.interpreter); // simplified code
[...]

PR #1069, opened 2023-02-16
Add SIP sidestepping to the vs-code extension, with support from the cli extension command.

t4lz:
Part of #747, solves #1061.

Since we do this by changing the executable field in the launch configuration and that field is different in each configuration type we need to add specify the configuration type -> configuration field name for each configuration type we want to support, currently supporting node and python.

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -264,6 +264,9 @@ pub(super) struct ExtensionExecArgs {
     /// Specify target
     #[arg(short = 't')]
     pub target: Option<String>,
+    /// User executable - the executable the layer is going to be injected to.
+    #[arg(short = 'e')]
+    pub executable: Option<String>,
 }
 
 #[derive(Args, Debug)]
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -37,7 +37,7 @@ pub(crate) struct MirrordExecution {
     #[serde(skip)]
     child: Child,
 
-    /// The path to the patched binary, if patched.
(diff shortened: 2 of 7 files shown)


mirrord/cli/src/config.rs
```diff
@@ -264,6 +264,9 @@ pub(super) struct ExtensionExecArgs {
     /// Specify target
     #[arg(short = 't')]
     pub target: Option<String>,
+    /// User executable - the executable the layer is going to be injected to.
+    #[arg(short = 'e')]
```
eyalb181, line 268: Do we want to expose this to users? If not maybe use an env var instead?

t4lz: This is an argument to the extension command, so none of this should be exposed to the users. So, as far as I can see this argument is not different from the other arguments in this config in this regard.

mirrord/cli/src/execution.rs
```diff
@@ -31,13 +33,19 @@ const INJECTION_ENV_VAR: &str = "DYLD_INSERT_LIBRARIES";
 #[derive(Debug, Serialize)]
 pub(crate) struct MirrordExecution {
     pub environment: HashMap<String, String>,
+
     #[serde(skip)]
     child: Child,
+
+    /// The path to the patched binary, if patched.
+    pub patched_path: Option<String>,
```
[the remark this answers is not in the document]
t4lz: Added SIP to the documentation. I am already out, so I won't get to renaming the variable today (also has to be renamed in the parsing in vscode).

eyalb181 (approved):

Issue #1044, opened 2023-02-09
Failure after a few incoming requests when using linkerd


eyalb181:
### Bug Description

In some cases, when running with linkerd (with `incoming_mode` set to `steal`), mirrord fails after a few incoming requests. Resetting the process fixes it (until the next batch of requests).

### Steps to Reproduce

Impersonate a pod in a cluster running linkerd.

### Your operating system and version

MacOS

### Local process

N/A

PR #1070, opened 2023-02-16
More consistent behavior between Mesh and Non Mesh stealing

DmitryDodzin:
This changes when tcp stealing on a meshed service, requests from localhost to localhost like healthprobes will skip mirrord 
and will be routed to the underlying service like when stealing from a non meshed service

Relevant for #1044

### DIFF
--- a/mirrord/agent/src/main.rs
+++ b/mirrord/agent/src/main.rs
@@ -555,14 +555,14 @@ async fn start_iptable_guard() -> Result<()> {
 
     let result = spawn_child_agent();
 
-    run_thread_in_namespace(
+    let _ = run_thread_in_namespace(
         clear_iptable_chain(chain_name),
         "clear iptables".to_owned(),
         pid,
         "net",
     )
     .join()
-    .map_err(|_| AgentError::JoinTask)??;
+    .map_err(|_| AgentError::JoinTask)?;
 
     result
 }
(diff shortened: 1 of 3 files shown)


mirrord/agent/src/main.rs
```diff
@@ -555,14 +555,14 @@ async fn start_iptable_guard() -> Result<()> {
 
     let result = spawn_child_agent();
 
-    run_thread_in_namespace(
+    let _ = run_thread_in_namespace(
```
meowjesty, line 558: Aren't we eating up the error now with this?

mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -220,24 +226,38 @@ where
...
 impl IPTableFormatter {
     const MESH_OUTPUTS: [&'static str; 2] = ["-j PROXY_INIT_OUTPUT", "-j ISTIO_OUTPUT"];
+    const MESH_NAMES: [&'static str; 2] = ["PROXY_INIT_OUTPUT", "ISTIO_OUTPUT"];
 
     #[tracing::instrument(level = "trace", skip_all)]
     pub(crate) fn detect<IPT: IPTables>(ipt: &IPT) -> Result<Self> {
         let output = ipt.list_rules("OUTPUT")?;
 
-        if output.iter().any(|rule| {
+        if let Some(mesh_ipt_chain) = output.iter().find_map(|rule| {
```
meowjesty, line 243: Probably good to add a comment here mentioning that we need this for the health probes.

meowjesty (changes requested):
Just requesting some docs, and I think you're eating up an error.

mirrord/agent/src/main.rs
aviramha: It's by design since it can fail

aviramha (approved):
LGTM (besides meow comments)

meowjesty (approved):
:bear: :+1:

Issue #951, opened 2023-01-16
Don't fail on SIP when an executable/script does not exist.


t4lz:
When the SIP check does not find a file that should be executed, it fails.
In practice, this error usually happens when the user is trying to execute a file that does not exist, so it might be confusing to show a SIP error.

I suggest we emit a warning, but just return from the SIP check as if the file is not SIP. If the SIP check could not find the file, it's most likely because it's not there, and then the execution would fail on a less confusing spot.

The tradeoff is that if there's a bug in the SIP's logic and the SIP check could not find a file that is actually there and is SIP - then the SIP check would say "no SIP", and then the program would run but mirrord would fail to load to it. Hopefully in that case the warning mentioned above would help us find such a bug and fix it.

PR #1074, opened 2023-02-17
Don't exit on SIP-check fail, show error.

t4lz:
#951. Thought this would help me with another PR. It did not, but at least we can close this issue.


aviramha:
Overall comment since I'm not sure - if the flow returns an error, then the log should be in the error definition in enum instead of in the code path itself, then caller can choose whether to ignore it (and show the warning) or exit with the error.

aviramha (approved):
LGTM!

Issue #1063, opened 2023-02-14
Layer/Agent can't handle large PUT request


aviramha:
User runs steal on a rest service written in Go + macOS that receives a big PUT request.
The request gets stuck and rest of the stuff gets timeout (kafka, etc).

Possible issues:
1. maybe we have some blocking logic so it can't answer health check / do anything else while accepting?
2. maybe it's writing/receiving a lot of data causing a busy loop not giving time for other threads to run?
3. ??

aviramha:
Solved via #1092

PR #1092, opened 2023-02-20
Handle the case where the local app closes a stolen TCP connection.

t4lz:
Potentially related to #1063.


t4lz:
Oops, commited too much. Reseting.

eyalb181 (approved):

PR #1098, opened 2023-02-21
Pause Warn Only

DmitryDodzin:
Only warn if pause was unsuccessful


eyalb181:
@DmitryDodzin @aviramha I think we can close this right?

PR #1104, opened 2023-02-22
make sniffer optional

aviramha:


mirrord/agent/src/main.rs
```diff
@@ -285,12 +290,20 @@ impl ClientConnectionHandler {
                         break;
                     }
                 },
-                message = self.tcp_sniffer_api.recv(), if sniffer_available => {
+                // use to prevent closing when sniffer is not available
+                // while avoiding possible busy loop (infinite iteration of None)
+                message = async {
+                    if let Some(ref mut sniffer_api) = self.tcp_sniffer_api {
+                        sniffer_api.recv().await
+                    } else {
```
t4lz, line 298: And does the compiler force us to have this else clause? Is it impossible to have just the then clause?

mirrord/agent/src/main.rs
```diff
@@ -285,12 +290,20 @@ impl ClientConnectionHandler {
...
+                        unreachable!()
```
t4lz, line 299: Is this unreachable even though the async expression is evaluated (but not polled) even when the precondition does not hold?

DmitryDodzin (approved):

mirrord/agent/src/main.rs
```diff
@@ -285,12 +290,20 @@ impl ClientConnectionHandler {
                         break;
                     }
                 },
-                message = self.tcp_sniffer_api.recv(), if sniffer_available => {
+                // use to prevent closing when sniffer is not available
```
DmitryDodzin, line 293: Should update comment

mirrord/agent/src/main.rs
aviramha: I think it's good since it's an assertion

mirrord/agent/src/main.rs
aviramha: I copied it from other codebase tbh, but pretty sure this should work since sniffer api shouldn't become unavailable during the run, only at load so this should be okay

mirrord/agent/src/main.rs
```diff
@@ -234,7 +234,16 @@ impl ClientConnectionHandler {
...
         let tcp_sniffer_api =
-            TcpSnifferApi::new(id, sniffer_command_sender, tcp_receiver, tcp_sender).await?;
+            TcpSnifferApi::new(id, sniffer_command_sender, tcp_receiver, tcp_sender)
+                .await
+                .map_err(|e| {
+                    warn!(
+                        "Failed to create TcpSnifferApi: {}, this could be due to kernel version.",
+                        e
+                    );
+                    e
```
t4lz, line 244: Nitpick: I think you could use `inspect_err` here instead of returning the same error.

aviramha: I don't like adding more dependencies on nightly

PR #1115, opened 2023-02-23
Fix error in vscode ext and update to 3.28.2

eyalb181:


vscode-ext/src/extension.ts
```diff
@@ -78,6 +78,7 @@ class MirrordAPI {
 				throw new Error("Unsupported platform: " + process.platform + " " + process.arch);
 			} else {
 				console.log("Running in WSL.");
```
t4lz, line 80: This is not from this PR, but shouldn't this be 
```suggestion
				console.log("Running in Windows.");
```
?

t4lz (approved):

PR #1125, opened 2023-02-26
Meshed stealing remove accidental outgoing capture

DmitryDodzin:
relevant for #1041

### DIFF
--- a/mirrord/agent/src/steal/ip_tables.rs
+++ b/mirrord/agent/src/steal/ip_tables.rs
@@ -392,7 +392,7 @@ mod tests {
         mock.expect_insert_rule()
             .with(
                 str::starts_with("MIRRORD_REDIRECT_"),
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -257,7 +257,7 @@ impl IPTableFormatter {
                 .list_rules(mesh_ipt_chain)?
                 .iter()
                 .find_map(|rule| UID_LOOKUP_REGEX.find(rule).ok().flatten())
-                .map(|m| m.as_str().to_owned());
+                .map(|m| format!("-o lo {}", m.as_str()));
 
             Ok(IPTableFormatter::Mesh(
                 filter.unwrap_or_else(|| "-o lo".to_owned()),
```
aviramha, line 263: add a warning log incase `filter` is None as it might be another issue we face and it is "hidden" right now

aviramha (approved):

PR #1130, opened 2023-02-27
Removed test from function names

### LINKED ISSUE #1065: Remove the word "test" from all tests.
All of our tests are called `test_...`. Seems redundant.
Since tests are marked with `#[rstest]`, we don't need to mark them with the word "test" at the beginning of their name.

I suggest we just call them as what they are testing, so if a test is testing that a feature like http mirroring works, then we just call it `http_mirroring`, and if it verifies some invariant, then use the invariant as the name, e.g. `tmp_files_are_bypassed`.

What do you think?

gautamprikshit1:
Fixes #1065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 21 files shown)


t4lz (approved):
Nice, thank you for your contribution!

PR #1132, opened 2023-02-28
Link license and changelog in vscode dir.

t4lz:
Also: remove the line copying those files in the release action.

I mostly want this because right now when you build the extension locally you need to copy those files to the directory, and then you have them as untracked files in git. So if we reject this PR we can maybe just add those paths to .gitignore.

Not linking the mirrord binary because you would want to use a different target path in different situations, so we still need to copy over the mirrord binary. So I added that to `.gitignore`.


aviramha (approved):
Cool :)

PR #1137, opened 2023-03-01
Block ipv6 binds to prevent C# errors

### LINKED ISSUE #1121: C# Support
Simple example given here https://github.com/aeb-dev/mirrord-test doesn't work with mirrord.
```
The AddressFamily InterNetwork is not valid for the System.Net.IPEndPoint end point, use InterNetworkV6 instead. (Parameter socketAddress)
```

DmitryDodzin:
closes #1121


aviramha:
CHANGELOG :)

aviramha (changes requested):

mirrord/layer/src/error.rs
```diff
@@ -193,6 +196,9 @@ impl From<HookError> for i64 {
             HookError::IO(ref e) if (is_ignored_code(e.raw_os_error())) => {
                 info!("libc error (doesn't indicate a problem) >> {:#?}", fail)
             }
+            HookError::SocketUnsuportedIpv6 => {
+                info!("{}", fail)
```
aviramha, line 200: `info!("{fail}")`

aviramha (approved):
nitpick but can merge :)

Issue #1054, opened 2023-02-12
fix dup with getsockname/getpeername/accept


aviramha:
When application calls `bind` -> `listen` -> `dup` -> `accept` (on dup'd fd) we wouldn't override the remote/local address due to `SocketInformation` being passed by `fd` and the listen fd we hold isn't the same as dup'd.
The solution is probably to add an identifier to `UserSocket` and use that for passing `SocketInformation`.

meowjesty:
Should [this](https://gist.github.com/meowjesty/db509a8bbfa8fe54e648b625b95bd7fc) code trigger this behavior? 

I'm only seeing the expected output from it, so maybe I'm getting the order of operations wrong?

aviramha:
> Should [this](https://gist.github.com/meowjesty/db509a8bbfa8fe54e648b625b95bd7fc) code trigger this behavior?
> 
> I'm only seeing the expected output from it, so maybe I'm getting the order of operations wrong?

That looks correct, but to be on the safe side, change the `app_flask.py` as described in the issue.

meowjesty:
[Output](https://gist.github.com/meowjesty/c1e61ef672f415dd63fa7d15c66aa194) with mirrord.

The client for this whole thing is just a `TcpStream` connecting to `localhost:30030`.

I've tried it with the changes to `app_flask.py` and the assertion fails, so I was trying to try and reproduce it with a more direct approach.

aviramha:
You need to connect from elsewhere, since in this case both remote and local is 127.0.0.1. Run it on a remote cluster :)

meowjesty:
#1149 should help a bit with this issue, but it's not the full solution, as we're dependent on #1123

eyalb181:
This was fixed by @t4lz

PR #1139, opened 2023-03-01
Investigating possible id solution (added an id to UserSocket).

meowjesty:
- Issue #1054 

very wip


aviramha:
The issue is probably that you're asserting for the IP from UDP and not from the HTTP (TCP) session..
In regards to the actual change - I think we can just "allocate" an `u64` and put it in socketinformation then just use that for `.get` no? not sure why we need the broad change

meowjesty:
Closed in favor of #1149

PR #1143, opened 2023-03-02
No error logs on failed SIP.

t4lz:
This was an error log so that if SIP detection fails the user is informed, so that they don't think everything is going fine but actually mirrord might not be loaded into their new process.

However this produced false-positives, because e.g. `env` works by trying to execute the given name from all paths in `PATH` one by one and just failing until one of them works. So SIP detection failed on the non existing paths, and did not exit mirrord, but did emit an error log.

This PR just makes those logs `trace`. Alternatively we could match the error type and emit an error log only if it's not `FileNotFound`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 
+### Fixed
+
+- Unnecessary error logs when running a script that uses `env` in its shebang.
+
 ## 3.30.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


aviramha:
Nice catch. I think `info` is more suited though

t4lz:
> Nice catch. I think `info` is more suited though

Even on `FileNotFound` errors or should I make a distinction? There could be many of them in sequence as `env` iterates over the paths in `PATH`.

aviramha:
> > I think `info` is more suited though
> 
> Even on `FileNotFound` errors or should I make a distinction? There could be many of them in sequence as `env` iterates over the paths in `PATH`.

Fair point

aviramha:
Fair point

PR #1153, opened 2023-03-05
layer,cli: add option to specify binaries to patch

aviramha:


mirrord/sip/src/lib.rs
```diff
@@ -206,13 +210,20 @@ mod main {
             return Err(CyclicShebangs(canonical_path.to_string_lossy().to_string()));
         }
         seen_paths.insert(canonical_path);
+
+        // Patch binary if it is in the list of binaries to patch.
+        // See `ends_with` docs for understanding better when it returns true.
+        if patch_binaries.iter().any(|x| complete_path.ends_with(x)) {
```
DmitryDodzin, line 216: Nitpick but maybe it shouldn't be just ends_with but rather `complete_path.file_name() == Some(x) || complete_path == x`

DmitryDodzin (approved):

mirrord/sip/src/lib.rs
aviramha: I think that `ends_with` provides better use experience

PR #1154, opened 2023-03-06
config: add ignore_localhost to outgoing

aviramha:


DmitryDodzin (approved):

mirrord/config/src/incoming.rs
```diff
@@ -118,7 +120,7 @@ impl MirrordToggleableConfig for IncomingFileConfig {
     }
 }
 
-#[derive(Deserialize, Clone, Debug, JsonSchema)]
+#[derive(MirrordConfig, Deserialize, Clone, Debug, JsonSchema)]
```
DmitryDodzin, line 123: If you are adding `MirrordConfig` on this then need to change IncomingFileConfig to accept the output of the macro and call generate_config on it inside `MirrordConfig` impl for `IncomingFileConfig`

DmitryDodzin: Also Should add 
```rust
#[config(env = "MIRRORD_AGENT_TCP_STEAL_TRAFFIC", default)]
pub mode: Option<IncomingMode>
```

DmitryDodzin (changes requested):

DmitryDodzin (approved):

PR #1162, opened 2023-03-06
possible socket leak

aviramha:


meowjesty (approved):
Oohh, nice catch  :baseball:

PR #1181, opened 2023-03-12
fix environment regression by mixing the two approaches together.

### LINKED ISSUE #1180: Environment variables regression
We are using [Bank-Vaults mutating webhook](https://banzaicloud.com/docs/bank-vaults/mutating-webhook/) to inject env variables, and this setup works with mirrord 3.29.0 but not with 3.30.0 - the target is an Alpine-based container, and I can see them in `/proc/1/environ`. Perhaps a combination of the two approaches is needed?

_Originally posted by @artem-collectai in https://github.com/metalbear-co/mirrord/issues/1135#issuecomment-1464072811_

aviramha:
Fixes #1180

### DIFF
--- a/mirrord/agent/src/env.rs
+++ b/mirrord/agent/src/env.rs
@@ -52,11 +52,11 @@ impl EnvFilter {
 }
 
 /// Translate ToIter<AsRef<str>> of "K=V" to HashMap.
-pub(crate) fn parse_raw_env<'a, S: AsRef<str>, T: IntoIterator<Item = S>>(
+pub(crate) fn parse_raw_env<'a, S: AsRef<str> + 'a + ?Sized, T: IntoIterator<Item = &'a S>>(
     raw: T,
 ) -> HashMap<String, String> {
     raw.into_iter()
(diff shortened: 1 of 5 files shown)


DmitryDodzin (approved):

mirrord/agent/src/env.rs
```diff
@@ -46,23 +51,46 @@ impl EnvFilter {
...
+    Ok(parse_raw_env(raw_env_vars.split_terminator(char::from(0))))
+}
+
 /// Helper function that loads the process' environment variables, and selects only those that were
 /// requested from `mirrord-layer` (ignores vars specified in `filter_env_vars`).
 ///
 /// NOTE: can remove `RemoteResult` when we break protocol compatibility.
-pub(crate) fn select_env_vars(
+#[tracing::instrument(level = "trace", skip(full_env))]
+pub(crate) async fn select_env_vars(
```
DmitryDodzin, line 88: No need async I think

aviramha: nice catch! thanks

PR #1183, opened 2023-03-12
Mesh support for --skip-inbound-ports

DmitryDodzin:
relevant for #1041


aviramha:
High level comment - I think that the code is very complex unneedingly. Initially we set the `nat` as default table since we didn't need to support different tables, but now that it has changed, I think we need to think better what's the proper API. Right now the logic is generic at places that shouldn't be generic and "static" in places where it should be generic.
For example, I'd stop having the `nat` table as a default and just change the API to receive `table` then each caller specifies explicitly what table it edits. 

Another possible design can be having a `PrerouteManager` and `NatManager` each one implements same trait then an enum that calls both if mesh, or only one if not mesh or something similar.. Just an idea but right now this is too complex imo

DmitryDodzin:
```rust
trait RewriteChain {
    /// Add current rewrite chain to relevant entrypoint in table
    fn add_rewrite(&self) -> Result<()>;
    /// Add current rewrite chain to relevant entrypoint in table
    fn remove_rewrite(&self) -> Result<()>;

    /// Create port redirection
    fn add_redirect(&self, redirected_port: Port, target_port: Port) -> Result<()>;
    /// Remove port redirection
    fn remove_redirect(&self, redirected_port: Port, target_port: Port) -> Result<()>;
}
```
So something like this trait and implemented on `PrerouteRewrite` and `MeshRewrite` or `LinkerdRewrite`/`IstioRewrite` with the appropriate changes in IPTables trait

mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -205,101 +200,188 @@ where
...
+                    .list_rules(Self::MESH_OUTPUT_NAMES[mesh_index])?
+                    .iter()
+                    .find_map(|rule| UID_LOOKUP_REGEX.find(rule).ok().flatten())
+                    .map(|m| format!("-o lo {}", m.as_str())).or_else(|| {
+                        warn!("Couldn't find --uid-owner of meshed chain {:?} falling back on \"-o lo\" rule", Self::MESH_OUTPUT_NAMES[mesh_index]);
+
+                        Some("-o lo".to_owned())
+                    });
+
+            let skiped_ports = ipt
```
eyalb181, line 254:
```suggestion
            let skipped_ports = ipt
```

mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -45,258 +75,117 @@ impl IPTables for iptables::IPTables {
...
+            Redirects::Mesh(MeshRedirect::create(Arc::new(ipt), vendor)?)
+        } else {
+            Redirects::Standard(PreroutingRedirect::create(Arc::new(ipt))?)
+        };
 
-        ipt.add_rule(formatter.entrypoint(), &format!("-j {}", &chain_name))?;
+        let redirect = if flush_connections {
+            Redirects::FlushConnections(FlushConnections::new(Box::new(redirect)))
+        } else {
+            redirect
```
eyalb181, line 134: A bit weird, can you replace this with the if clause above? i.e. `if let Some(vendor) = MeshVendor::detect(&ipt)?...`

mirrord/agent/src/steal/ip_tables/chain.rs
```diff
@@ -0,0 +1,92 @@
...
+        let chain_size = AtomicI32::from((existing_rules - 1) as i32);
+
+        Ok(IPTableChain {
+            inner,
+            chain,
+            chain_size,
+        })
+    }
+
+    pub fn chain_name(&self) -> &str {
```
eyalb181, line 54: Should be named the same as `chain` (like, either both should be named `chain` or both should be named `chain_name`

eyalb181 (commented):
Nice! Some tiny nits

mirrord/agent/src/steal/ip_tables.rs
DmitryDodzin: I changed it a bit to not have the else clause, also experimented with .with_flush_connections() but I'm not sure about it being needed

eyalb181 (approved):

Issue #947, opened 2023-01-16
Handle file streams differently


aviramha:
Right now we do weird stuff with file streams, we should handle those similar to DirStreams - an object that is identified by usize (pointer) and points to a remote object (RemoteFile).
For example, on calling `fclose` we call `fclose` even if it's a stream managed by us, meaning this would most probably crash

This has not been tested yet, but we assume that running even simple flows with filestreams (e.g. by running mirrord on a node app) would break.

Functions to hook/fix

- [ ] ferror - right now returns last os error but not os error of the specific stream meaning we're probably returning something invalid.
- [ ] feof - need to hook
- [ ] clearerr - need to hook
- [ ] fileno - need to fix implementation
- [ ] fdopen - check mode matches existing open file
- [ ] fclose - need to fix

My proposed idea would be that most of the logic is layer-only, meaning we need some additional abstraction for fds similar to what libc provides:
```rs
#[derive(Default)]
pub(crate) struct FileStream {
    /// Raw Fd to use for operations (should be one managed by mirrord)
    pub(crate) fd: RawFd,
    /// Last error for `ferror` implementation
    pub(crate) last_error: c_int,
    /// is EOF for `feof` implementation
    pub(crate) is_eof: bool,
    /// for `flockfile`/`stuff_unlocked` implementation - skipped for now, leaving this
    /// in case we need to implement it later on.
    // pub(crate) lock: parking_lot::Mutex<()>,
}
```

aviramha:
We needed to comment out flask http tests because those started failing with `Fatal error: glibc detected an invalid stdio handle` after most probably an update to werkzeug.

meowjesty:
Looks like the culprit was [`__fsetlocking`] that would de-reference the `* FILE`, which we never filled with anything (as we were just passing it as a pointer address), you can see it in [this](https://android.googlesource.com/platform/bionic/+/master/libc/stdio/stdio_ext.cpp#85) implementation of it.

[...]

PR #1198, opened 2023-03-20
Fix `FILE` stream operations (by removing them)

meowjesty:
- Issue: #947 

## Removal of file stream operations

We initially implemented hooks for functions that deal with `* FILE` thinking that they would be needed, but as these functions just end up calling their fd counterparts, it's safe to completely remove these hooks from `mirrord-layer`.

## Introduce `CheckedInto` trait

Some hooks that deal with pointers were converting these into Rust types in each function (mainly `* char` to `CString`). This trait and it's implementation should help to simplify the process of converting these, and helps make sure that you can't forget to check for `null`.


mirrord/layer/src/file/ops.rs
```diff
@@ -379,26 +295,13 @@ fn remote_read(remote_fd: u64, read_amount: u64) -> Detour<ReadFileResponse> {
...
-}
-
-#[tracing::instrument(level = "trace")]
-pub(crate) fn fgets(local_fd: RawFd, buffer_size: usize) -> Detour<ReadFileResponse> {
-    // We're only interested in files that are paired with mirrord-agent.
-    let remote_fd = get_remote_fd(local_fd)?;
+    let read = file_channel_rx.blocking_recv()??;
 
-    let (file_channel_tx, file_channel_rx) = oneshot::channel();
+    let d = String::from_utf8_lossy(&read.bytes[..read.read_amount as usize]);
```
eyalb181, line 300: d?

eyalb181 (approved):

PR #1212, opened 2023-03-22
fix wsl config with IntelliJ

### LINKED ISSUE #1210: JetBrains WSL doesn't work with config file
log:
```
creating command line and executing C:\Windows\system32\wsl.exe [--distribution, Ubuntu-22.04, --exec, /bin/sh, -c, /mnt/c/Users/U/AppData/Local/JetBrains/Toolbox/apps/IDEA-U/ch-0/223.8836.41.plugins/mirrord/bin/linux/x86-64/mirrord ls -o json -f 'E:\marosa\workspace\plain-report-service\.mirrord\mirrord.json']
```

as can be seen, the issue is that we don't convert the path to a WSL one.

aviramha:
Fix #1210


intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordApi.kt
```diff
@@ -49,10 +49,14 @@ object MirrordApi {
     fun listPods(configFile: String?, project: Project?, wslDistribution: WSLDistribution?): List<String> {
         MirrordLogger.logger.debug("listing pods")
         var commandLine = GeneralCommandLine(cliPath(wslDistribution), "ls", "-o", "json")
-        configFile?.let {
+        configFile?.let {config ->
             MirrordLogger.logger.debug("adding configFile to command line")
             commandLine.addParameter("-f")
-            commandLine.addParameter(it)
+            wslDistribution?.let {
+                commandLine.addParameter(it.getWslPath(config) ?: config)
```
eyalb181, line 56: Are both options valid?

aviramha: no but I think it's a worthy fallback - my concern that path is already `WSL'd` for some reason and that's why it didn't work, hence fallback

aviramha: added clarification

eyalb181 (approved):

PR #1219, opened 2023-03-23
Fallible Operator Discover

### LINKED ISSUE #1218: make finding operator fallible
In some cases, where trying to connect to operator fails because it's not really there, or lack of permissions to even check that cause an error mirrord fails and forces the user to mention to not use operator.
Instead of this behavior, we should just issue a warning and continue without using the operator.

DmitryDodzin:
closes #1218

### DIFF
--- a/changelog.d/1218.fixed.md
+++ b/changelog.d/1218.fixed.md
@@ -1,2 +1,2 @@
-mirrord-cli: update cli so failute to connect to operator will try operatorless flow instead of exiting
+mirrord-cli: update cli so failing to use operator will fallback to no-operator mode.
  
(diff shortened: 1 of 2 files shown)


changelog.d/1218.fixed.md
```diff
@@ -0,0 +1,2 @@
+mirrord-cli: update cli so failute to connect to operator will try operatorless flow instead of exiting
```
eyalb181, line 1:
```suggestion
mirrord-cli: update cli so failure to connect to operator will try operatorless flow instead of exiting
```

aviramha, line 1: update cli so failing to use operator will fallback to no-operator mode.

eyalb181 (approved):

Issue #1202, opened 2023-03-21
gethostname leak


aviramha:
`gethostname` detour leaks fd because it calls `open` then `read` then `close_layer_fd` but none calls `close` on the local fd itself.
doesn't have real effect, but we should fix it.

meowjesty:
We handle this file fully remote now as per #1226

PR #1226, opened 2023-03-27
Open hostname file remotely.

meowjesty:
- Issue: #1202 

PR #1213 added `RemoteFile` methods that operate directly on the agent (no local fd is created), but missed calling these in `gethostname`.


mirrord/layer/src/socket/ops.rs
```diff
@@ -711,23 +713,24 @@ pub(super) fn getaddrinfo(
...
             read: true,
             ..Default::default()
         },
     )?;
 
-    let hostname_file = file::ops::read(hostname_fd, 256)?;
+    let ReadFileResponse { bytes, read_amount } = file::ops::RemoteFile::remote_read(fd, 256)?;
 
-    close_layer_fd(hostname_fd);
+    let _ = file::ops::RemoteFile::remote_close(fd).inspect_err(|fail| {
```
eyalb181, line 726: Don't you still need to close the layer fd?

meowjesty: We should be opening the file now only in the remote, no local fd is created, and we don't insert the file into `OPEN_FILES` either, so there is nothing local to close.

eyalb181 (approved):

mirrord/layer/src/socket/ops.rs
eyalb181: Gotcha

mirrord/layer/tests/common/mod.rs
```diff
@@ -183,20 +183,26 @@ impl LayerConnection {
...
-                // triggered by the app.
-                // So accept the next connection which will be the one by the library that was
-                // loaded to the python process that actually runs the application.
-                codec = Self::accept_library_connection(listener).await;
-                codec.next().await.unwrap().unwrap()
-            }
-        };
-        assert_eq!(msg, ClientMessage::Tcp(LayerTcp::PortSubscribe(app_port)));
+
+        if codec.next().await.is_none() {
```
t4lz, line 187: Here you are dropping the next message of the codec, even if it's `Some`. What are the current changes in this file for?
When checking out this file from main into this branch, only the one test that is actually affected by the change in gethostname fails, instead of multiple tests that use this function.

For that test you would need to change this function to optionally receive and answer a `FileRequest(Open(OpenFileRequest { path: "/etc/hostname" ... }))` before the `PortSubscribe`, or maybe instead of making this function also accept that case, create a `get_initialized_connection_with_hostname_and_port` function that is like this one but always expects a hostname `FileRequest` - up to you.

The integration tests work by opening a TCP port for the layer to connect to and then answering as an agent would, simulating the agent for the layer. This means that the integration tests have to anticipate and answer each message that the layer is going to send in the scenario they create. Because of the `gethostname` change, in one of the tests (on mac it's only one, maybe it's different on Linux), the layer sends a `FileRequest` before it sends a `PortSubscribe`, so in that case the integration test has to read that message and then send a response.

Let me know if that explanation wasn't clear, we can talk about it over voice :)

mirrord/layer/src/outgoing/udp.rs
```diff
@@ -140,7 +140,7 @@ impl UdpOutgoingHandler {
                             break;
                         },
                         Ok((read_amount, from)) => {
-                            trace!("interceptor_task -> Received data from {:#?}", from);
+                            info!("interceptor_task -> Received data {:?} from user socket {:#?}", &recv_from_buffer[..read_amount], from);
```
aviramha, line 143: revert

mirrord/layer/src/outgoing/udp.rs
```diff
@@ -158,6 +158,7 @@ impl UdpOutgoingHandler {
                 bytes = remote_stream.next() => {
                     match bytes {
                         Some(bytes) => {
+                            info!("interceptor_task -> Received data {:?} from remote socket", bytes);
```
aviramha, line 161: revert

aviramha (changes requested):

changelog.d/1202-leak.fixed.md
aviramha, line 1: fix filename (`1202.fixed.md`)

meowjesty: fixed?

mirrord/layer/src/outgoing/udp.rs
```diff
@@ -140,7 +140,7 @@ impl UdpOutgoingHandler {
                             break;
                         },
                         Ok((read_amount, from)) => {
-                            trace!("interceptor_task -> Received data from {:#?}", from);
+                            trace!("interceptor_task -> Received data {:?} from user socket {:#?}", &recv_from_buffer[..read_amount], from);
```
aviramha, line 143: I wouldn't put data in logs.. ever

meowjesty: Oh snap!

t4lz:
wait, it looks like the test is still not stable. removing from queue, ok?

aviramha:
> wait, it looks like the test is still not stable. removing from queue, ok?

Crap, good catch. I assumed it's our usual flake ;(

PR #1229, opened 2023-03-28
Tonic api Draft

DmitryDodzin:


mirrord/agent/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM --platform=$BUILDPLATFORM ghcr.io/metalbear-co/ci-agent-build:latest as bui
 ARG TARGETARCH
 WORKDIR /app
 
+RUN apt-get update && apt-get install -y protobuf-compiler
```
aviramha, line 5: leaving this as a reminder to move it to the base image

mirrord/protocol/proto/api.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package mirrord.api;
+
+message Empty {}
+
+message BincodeMessage {
+  bytes encoded = 1;
+}
```
t4lz, line 9: I somehow expected that the protocol messages would be defined as protocol buffer messages. If all we're sending are already unstructured bytes (as far as protocol buffers are concerned), why do we use protocol buffers at all?

aviramha: It's a good question and should be written somewhere in the draft for better discussion.
We wanted to have a way to convert functionality piece by piece, without needing to rewrite the whole agent. The solution @DmitryDodzin came up with is that the new protocol will contain the old protocol so we can start moving parts to the new protocol as we go.

PR #1259, opened 2023-04-04
Fixed console logger mirrord crashe

aviramha:


meowjesty (approved):
:bear: :+1: 

PS. I think one of the tests used `tokio-tugstenite` (websockets tests)? Do we have it as a dev-dependency still?

PR #1262, opened 2023-04-05
TESTING.md improvement

Razz4780:
* Moved from Docker Desktop to Minikube, the former was flaky
* Added missing dependency for Python example apps
* Fixed command to run integration tests


changelog.d/+testing-guide.internal.md
```diff
@@ -0,0 +1 @@
+Improve testing guide.
```
eyalb181, line 1: This actually isn't internal - should be relevant to external contributors as well

t4lz: I think the internal/external border should pass between mirrord developers and mirrord users, and not between the core team and the rest of the world. So I would also put it in internal.

eyalb181: Fair enough

eyalb181 (approved):

PR #1263, opened 2023-04-06
Use default for index allocator

Razz4780:
I noticed a `TODO` comment for the `Default` implementation on `IndexAllocator`, the fix was simple.

New trait bound for the implementation is as general as the definition of the struct itself allows. This enables to create an unusable instance of the struct, but that's the convention followed by Rust's std (see `HashMap` for example).


meowjesty (approved):
:bear: :+1:

PR #1268, opened 2023-04-10
Implemented `Default` for `Subscriptions` struct

Razz4780:
According to `todo` left by @meowjesty 

Removed `Subscriptions::new` method and implemented `Default` trait.


meowjesty (approved):
:bear: :+1:

PR #1269, opened 2023-04-10
Error message displayed in VS Code when there is no selectable target

Razz4780:
Solves #1266 
When there is no target specified in the mirrord configuration file and there is no available target in the current namespace, an error prompt is shown to the user and debugging session is cancelled.
See in action [here](https://user-images.githubusercontent.com/34063647/230900208-bf32030b-94f0-4f6a-a559-de6e5627c73a.webm)


eyalb181 (approved):
Nice!

PR #1281, opened 2023-04-12
Fix pause test

t4lz:
The log_stream provided by kube-rs just returns a stream of chunks of bytes, and up until last week those chunks were one line per chunk, and then this changed and it's not clear why. In this PR we convert this stream into `AsyncRead` and use the `.lines()` method to buffer the bytes and read the logs line-by-line.

This new version was tested to work multiple time on the CI https://github.com/t4lz/mirrord/actions/runs/4677010752.

### DIFF
--- a/changelog.d/1261.internal.md
+++ b/changelog.d/1261.internal.md
@@ -1,0 +1,1 @@
+Fix pause E2E test.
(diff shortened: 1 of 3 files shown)


DmitryDodzin (approved):
Small Nitpick

PR #1282, opened 2023-04-12
Add imagePullSecrets configuration

eyalb181:


mirrord/config/src/agent.rs
```diff
@@ -36,6 +38,11 @@ pub struct AgentConfig {
     #[config(env = "MIRRORD_AGENT_IMAGE_PULL_POLICY", default = "IfNotPresent")]
     pub image_pull_policy: String,
 
+    /// List of secrets the agent pod has access to.
+    ///
+    /// Takes an array of hash with the format {name: <secret-name>. Read more [here](https://kubernetes.io/docs/concepts/containers/images/).
```
meowjesty, line 43:
```suggestion
    /// Takes an array of hash with the format `{ name: <secret-name> }`. Read more [here](https://kubernetes.io/docs/concepts/containers/images/).
```

mirrord/kube/src/api/container.rs
```diff
@@ -211,10 +208,11 @@ impl ContainerApi for JobContainer {
                                 }
                             }
                         ],
+                        "imagePullSecrets": agent.image_pull_secrets,
                         "containers": [
                             {
                                 "name": "mirrord-agent",
-                                "image": agent_image,
+                                "image": Self::agent_image(agent),
```
meowjesty, line 215: Can't you keep the `agent_image` we had outside, and just call `agent_image.clone()` to clone the string value here?

mirrord/kube/src/api/container.rs
```diff
@@ -334,11 +331,9 @@ impl ContainerApi for EphemeralContainer {
             agent_command_line.push(timeout.to_string());
         }
 
-        let flush_connections = agent.flush_connections.to_string();
-
         let ephemeral_container: KubeEphemeralContainer = serde_json::from_value(json!({
             "name": mirrord_agent_name,
-            "image": agent_image,
+            "image": Self::agent_image(agent),
```
meowjesty, line 336: Same `clone` here.

meowjesty (changes requested):
Suggestions.

mirrord/kube/src/api/container.rs
eyalb181: Why do I need agent_image outside though?

meowjesty: You don't, I find it a bit more readable (but that's debatable).

I'll leave it up to you what looks better.

meowjesty (approved):
:bear: :+1:

mirrord/kube/src/api/container.rs
eyalb181: I think we should initialize variables as close as possible to where they're actually used. It's easier to follow when I read through the function (rather than seeing the variable being initialized, going through a bunch of other unrelated code, then seeing how it's used).

Issue #1155, opened 2023-03-06
SIP macOS - add fallback to arm64


aviramha:
Currently we extract x64 binary but sometimes arm64 is available so we can avoid emulation. In this case we should extract the arm64 binary and patch it.

aviramha:
> Almost done, just one concern, @aviramha: Isn't it possible that there are going to be fat binaries with arm64 binaries included in them, on macs with intel chips? In that case, if we prefer the arm64 binaries over the x64 binaries and use them, the patched arm64 binary is not going to be able to run on the intel x64 mac, is it?
> 
> Should we choose the binary according to the system's architecture?

That's correct.. That should be probably done in compile time, no?

PR #1283, opened 2023-04-12
Use arm64 if there in SIP patching

t4lz:
#1155 

Up until now we only used x86_64 binaries for SIP-patched binaries.
This means that if a binary did not have a x64 binary in it, we couldn't patch it.
Now, we can also patch it if there is an arm64 binary (when running on apple chips). Also, if both x86_64 and arm64 are present, we use arm64 (when running on apple chips), and avoid running over rosetta.

### DIFF
--- a/mirrord/sip/src/error.rs
+++ b/mirrord/sip/src/error.rs
@@ -13,8 +13,10 @@ pub enum SipError {
     #[error("Can't patch file format `{0}`")]
     UnsupportedFileFormat(String),
 
-    #[error("No x64 architecture in file")]
-    NoX64Arch,
+    #[error(
+        "No supported architecture in file (x86_64 on intel chips, x86_64 or arm64 on apple chips)"
+    )]
+    NoSupportedArchitecture,
 
     #[error("ObjectParse failed with `{0}`")]
     ObjectParse(#[from] object::Error),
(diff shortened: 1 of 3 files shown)


mirrord/sip/src/error.rs
```diff
@@ -13,8 +13,8 @@ pub enum SipError {
     #[error("Can't patch file format `{0}`")]
     UnsupportedFileFormat(String),
 
-    #[error("No x64 architecture in file")]
-    NoX64Arch,
+    #[error("No x64 or arm64 architecture in file")]
+    NoX64OrArm64,
```
aviramha, line 17: `NoSupportedArchitecture`

mirrord/sip/src/lib.rs
```diff
@@ -44,47 +57,81 @@ mod main {
...
                                 "MachO 32 file parsing failed".to_string(),
                             )
                         })?;
-                    if header.cputype(Endianness::default()) == CPU_TYPE_X86_64 {
-                        Ok(Self::new(0, bytes.len()))
-                    } else {
-                        Err(SipError::NoX64Arch)
-                    }
+                    Self::from_thin_mach_o(header.cputype(Endianness::default()), bytes)
                 }
```
t4lz, line 95: @aviramha do you agree that we can remove this arm?

aviramha (approved):
Great job!

eyalb181 (approved):

PR #1286, opened 2023-04-12
Improve e2e cleanup

### LINKED ISSUE #1256: Add a constant label to E2E test kubernetes resources for easy cleanup
Add some constant label to all the resources that are created by the E2E tests (Deployments, Services, Namespaces), so that they can be easily deleted.

See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

Then document in `TESTING.md` how to quickly delete all the test resources.

Razz4780:
Fixes #1256

* Added a new environment variable `MIRRORD_E2E_FORCE_CLEANUP` - when set, E2E tests will delete their resources even after failure
* Added a new constant label `MIRRORD_E2E_TEST_RESOURCE=true` to all E2E test resources for easy cleanup
* Updated testing guides
* Made some style improvements and added docs around resource creation and deletion

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -83,6 +83,14 @@ And then in order to use that binary in the tests, run the tests like this:
 MIRRORD_TESTS_USE_BINARY=../target/universal-apple-darwin/debug/mirrord cargo test -p tests
 ```
 
+### Cleanup
+
+The Kubernetes resources created by an E2E test are automatically deleted when the test exists successfully. However, failed tests by default don't delete their resources to allow debugging. You can force resource deletion by setting a `MIRRORD_E2E_FORCE_CLEANUP` variable to any value.
+
+```bash
+MIRRORD_E2E_FORCE_CLEANUP=y cargo test --package tests
+```
+
 ## Integration Tests
 
 The layer's integration tests test the hooks and their logic without actually using a Kubernetes cluster and spawning
--- a/TESTING.md
+++ b/TESTING.md
@@ -91,6 +91,13 @@ The Kubernetes resources created by an E2E test are automatically deleted when t
 MIRRORD_E2E_FORCE_CLEANUP=y cargo test --package tests
 ```
 
+All test resources share a common label `MIRRORD_E2E_TEST_RESOURCE=true`. To delete them, simply run:
+
+```bash
+kubectl delete namespaces,deployments,services -l MIRRORD_E2E_TEST_RESOURCE=true
+```
+
+
 ## Integration Tests
 
 The layer's integration tests test the hooks and their logic without actually using a Kubernetes cluster and spawning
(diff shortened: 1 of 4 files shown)


TESTING.md
```diff
@@ -83,6 +83,21 @@ And then in order to use that binary in the tests, run the tests like this:
 MIRRORD_TESTS_USE_BINARY=../target/universal-apple-darwin/debug/mirrord cargo test -p tests
 ```
 
+### Cleanup
+
+The Kubernetes resources created by an E2E test are automatically deleted when the test exists successfully. However, failed tests by default don't delete their resources to allow debugging. You can force resource deletion by setting a `MIRRORD_E2E_FORCE_CLEANUP` variable to any value.
```
eyalb181, line 88:
```suggestion
The Kubernetes resources created by the E2E tests are automatically deleted when the test exits successfully. However, failed tests by default don't delete their resources to allow debugging. You can force resource deletion even for failed tests by setting the `MIRRORD_E2E_FORCE_CLEANUP` variable to any value.
```

tests/src/lib.rs
```diff
@@ -39,22 +39,30 @@ mod utils {
...
     pub const CONTAINER_NAME: &str = "test";
 
+    /// Name of the environment variable used to control cleanup after failed tests.
+    /// By default, resources from failed tests are preserved for debugging.
+    /// However, if this variable is set, resources will always be deleted.
+    pub const FORCE_CLEANUP_ENV_NAME: &'static str = "MIRRORD_E2E_FORCE_CLEANUP";
+
+    /// All Kubernetes resources created for testing purposes share this label.
+    pub const TEST_RESOURCE_LABEL: (&'static str, &'static str) =
+        ("MIRRORD_E2E_TEST_RESOURCE", "true");
```
eyalb181, line 55: Kubernetes labels conventionally use kebab case (so mirrord-e2e-test-resource)

eyalb181 (changes requested):

eyalb181 (approved):

tests/src/lib.rs
```diff
@@ -471,62 +492,67 @@ mod utils {
...
+    /// * `randomize_name` - whether a random suffix should be added to the end of the resource
+    ///   names
     #[fixture]
     pub async fn service(
         #[default("default")] namespace: &str,
         #[default("NodePort")] service_type: &str,
         #[default("ghcr.io/metalbear-co/mirrord-pytest:latest")] image: &str,
         #[default("http-echo")] service_name: &str,
         #[default(true)] randomize_name: bool,
-        #[default(false)] delete_after_fail: bool,
```
t4lz, line 484: I would leave this there and then OR it with the env var. Some deployments do work as long as they exist, so I would rather not have them keep running on fail by default.

Alternatively, we could make deleting everything the default, and only keep the resources when an env var is set. If you want to have it that way, please consult the team before making the change.

Razz4780: The team seems to like the second idea, I changed the flow

t4lz (approved):
Just the changelog category.

PR #1292, opened 2023-04-13
Integration tests for outgoing UDP/TCP

### LINKED ISSUE #1051: Add integration test for outgoing (UDP/TCP)
Create connection, send data back and forth, test getsockanme + getpeername.

Razz4780:
Fixes #1051 

Udp test fails for now, remote address returned from `recv_from` is some localhost port.

### DIFF
--- a/mirrord/layer/tests/outgoing.rs
+++ b/mirrord/layer/tests/outgoing.rs
@@ -1,4 +1,5 @@
 #![feature(assert_matches)]
+#![cfg(target_os = "linux")]
 
 use std::{net::SocketAddr, path::PathBuf, time::Duration};
 
--- a/mirrord/layer/tests/outgoing.rs
+++ b/mirrord/layer/tests/outgoing.rs
@@ -1,5 +1,4 @@
 #![feature(assert_matches)]
-#![cfg(target_os = "linux")]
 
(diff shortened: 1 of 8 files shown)


mirrord/layer/tests/outgoing.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let peers = RUST_OUTGOING_PEERS
+        .split(',')
+        .map(|s| s.parse::<SocketAddr>().unwrap())
+        .collect::<Vec<_>>();
+
+    for peer in peers {
+        let msg = conn.try_next().await.unwrap().unwrap();
+        let ClientMessage::UdpOutgoing(LayerUdpOutgoing::Connect(LayerConnect { remote_address: SocketAddress::Ip(addr) })) = msg else {
+            panic!("Invalid message received from layer: {msg:?}");
```
eyalb181, line 48: Why are you panicking rather than asserting (here and elsewhere)?

Razz4780: The `assert_matches` macro we have does not allow for binding values. For example, in line 62 I bind the `bytes` field, which is embedded deeply into the message.

Issue #1295, opened 2023-04-13
Ignoring a port for steal/mirror


aviramha:
Right now there's no way to exclude ports from mirroring/stealing. This is problematic in case where there are debug ports or stuff you want to still be able to connect locally.
I suggest to use the existing `port_mapping` configuration and add support for `null` as the "remote" port to mark it as ignored.
For example:
```json
{
    "feature": {
        "network": {
            "incoming": {
                "port_mapping": [[9222,null]],
            }
        }
    }
}
```

aviramha:
Other option is to add a new configuration. Might make more sense in terms of code and also UX..

PR #1296, opened 2023-04-13
Add `ignore_ports` to `incoming` configuration

aviramha:
so you can have ports that only listen locally (mirrord will not steal/mirror those ports). Solves #1295

### DIFF
--- a/changelog.d/1295.added.md
+++ b/changelog.d/1295.added.md
@@ -1,3 +1,2 @@
 Add `ignore_ports` to `incoming` configuration so you can have ports that only listen
 locally (mirrord will not steal/mirror those ports).
-Solves [#1295](https://github.com/metalbear-co/mirrord/issues/1295).
--- a/changelog.d/1295.added.md
+++ b/changelog.d/1295.added.md
@@ -1,0 +1,3 @@
+Add `ignore_ports` to `incoming` configuration so you can have ports that only listen
+locally (mirrord will not steal/mirror those ports).
+Solves [#1295](https://github.com/metalbear-co/mirrord/issues/1295).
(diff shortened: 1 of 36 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -139,7 +139,13 @@ pub(super) fn bind(
         return Detour::Bypass(Bypass::IgnoreLocalhost(requested_port));
     }
 
-    if is_ignored_port(&requested_address) || port_debug_patch(&requested_address) {
+    if is_ignored_port(&requested_address)
+        || port_debug_patch(&requested_address)
+        || INCOMING_IGNORE_PORTS
+            .get()
+            .expect("`INCOMING_IGNORE_PORTS` not set. Please report a bug")
+            .contains(&requested_port)
```
t4lz, line 147: I think it would make sense to have this inside `is_ignored_port`, is that possible?

changelog.d/1295.added.md
```diff
@@ -0,0 +1,3 @@
+Add `ignore_ports` to `incoming` configuration so you can have ports that only listen
+locally (mirrord will not steal/mirror those ports).
+Solves [#1295](https://github.com/metalbear-co/mirrord/issues/1295).
```
t4lz, line 3:
```suggestion
```
Let towncrier take care of that :)

t4lz (changes requested):
Two little comments in the code, and also, how about a test for this?

mirrord/layer/src/socket/ops.rs
aviramha: It was also my initial thought but `is_ignored_port` is used for outbound connections as well.

aviramha:
Added tests.

t4lz (approved):
Nice! Some changes still needed (small comments, and I guess a couple of `None`s are still missing for the new function parameter), but approving as these are trivial changes, so that you can already merge when resolved.

PR #1298, opened 2023-04-14
more clippy coverage, fix its warnings and fix armv7 agent build

aviramha:


tests/src/lib.rs
```diff
@@ -95,6 +95,7 @@ mod utils {
     }
 
     #[derive(Debug)]
+    #[allow(dead_code)]
     pub enum FileOps {
         Python,
         Rust,
```
t4lz, line 101: I think the thing here is that `Python` and `Rust` are only used on linux, so alternatively to allowing dead code, we could also conditionally compile them only on linux. Just an idea, up to you.

t4lz (commented):
Just saw there were a lot of changes here, so just flushing my review as comments.

PR #1311, opened 2023-04-17
Improve target selection in IntelliJ extension

### LINKED ISSUE #1267: Suggest different namespace if there are no targets (IntelliJ)

Razz4780:
Fixes #1267

I changed the error notification for the case when no target is available. Now the message matches the one we show in the VS Code extension and suggests changing the namespace or kubeconfig.
By the way, I noticed that a failure of `mirrord ls` was not handled and resulted with a null pointer exception. Added an error prompt and a log of stderr.
[See in action here](https://user-images.githubusercontent.com/34063647/232569283-d7895a63-fa56-4491-b320-5843f5548f71.webm)

### DIFF
--- a/changelog.d/1267.added.md
+++ b/changelog.d/1267.added.md
@@ -1,0 +1,1 @@
+A descriptive message is now presented in the IntelliJ extension when no target is available. Listing targets failure is now handled and an error notification is presented.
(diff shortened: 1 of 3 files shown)


t4lz (approved):
Nice.

PR #1319, opened 2023-04-18
Add userextras

DmitryDodzin:
Add `impersonate` permission on `userextras/accesskeyid`, `userextras/arn`, `userextras/canonicalarn` and `userextras/sessionname` resources to operator setup.


aviramha (approved):
Great! Thank you

PR #1333, opened 2023-04-19
Indexing slicing lint

### LINKED ISSUE #1049: Add indexing slicing clippy lint
https://rust-lang.github.io/rust-clippy/master/#indexing_slicing

Need to do it in operator and here.

Razz4780:
Resolves #1049


changelog.d/1049.add-indexing-slicing-lint.internal.md
aviramha, line 1: `1049.internal.md`

mirrord/agent/src/steal/http/filter.rs
```diff
@@ -101,7 +101,7 @@ pub(super) async fn filter_task(
         Ok(mut reversible_stream) => {
             match HttpVersion::new(
                 reversible_stream.get_header(),
-                &H2_PREFACE[..MINIMAL_HEADER_SIZE],
+                H2_PREFACE.get(..MINIMAL_HEADER_SIZE).unwrap(),
```
aviramha, line 104: shouldn't that be okay to slice since it's const + const? if not, maybe we need to const something?

mirrord/layer/src/file/hooks.rs
```diff
@@ -322,7 +325,9 @@ pub(crate) unsafe extern "C" fn pread_detour(
             //
             // Callers can check for EOF by using `ferror`.
             if read_amount > 0 {
-                let bytes_slice = &bytes[0..fixed_read as usize];
+                let bytes_slice = bytes
+                    .get(..fixed_read as usize)
+                    .expect("read_amount exceeds bytes length in ReadFileResponse");
```
aviramha, line 330: I think it's better we propagate errors? It's probably pretty fatal but we would get nicer error message I believe?

aviramha: Let's chat about on Discord

mirrord/agent/src/steal/http/filter.rs
aviramha: maybe use https://docs.rs/const-str/latest/const_str/macro.to_char_array.html ?
(not very important though)

mirrord/agent/src/main.rs
```diff
@@ -7,6 +7,7 @@
 #![feature(tcp_quickack)]
 #![feature(async_fn_in_trait)]
 #![allow(incomplete_features)]
+#![warn(clippy::indexing_slicing)]
```
eyalb181, line 10: Isn't there a more centralized option for enabling this, like the configuration toml?

Razz4780: As of now there's no way to enable lints globally via configuration files. I could just add this lint to clippy arguments in the CI, but I bet it would be annoying

eyalb181 (approved):

PR #1339, opened 2023-04-20
Add support for older IntelliJ versions

aviramha:


intellij-ext/modules/products/pycharm/build.gradle.kts
```diff
@@ -17,8 +17,8 @@ tasks {
 }
 
 intellij {
-    version.set(properties("platformVersion"))
-    plugins.set(listOf("PythonCore:223.7571.4"))
+    version.set("2022.1.4")
```
eyalb181, line 20: Does this mean we don't support older versions in Pycharm specifically?

aviramha:
I have a dead end because the `runConfiguration` extension for NodeJS was changed leaving it with same name and we can't limit the dependency to a specific nodejs version so we have to either support pre-2023 or post-2023 :|

Issue #1302, opened 2023-04-15
Error thrown when using HTTP filter


eyalb181:
### Bug Description

A user ran mirrord with an HTTP filter, in both the CLI and Goland, and got this error:
```
502 Bad Gateway
mirrord tried to forward the request to the local application and got hyper::Error(Canceled, "connection was not ready")
```

### Steps to Reproduce

N/A

### Your operating system and version

N/A

### Local process

N/A

### Additional Info

https://discord.com/channels/933706914808889356/1096092286821089461

aviramha:
Fixed with #1340 according to user on Discord.

PR #1340, opened 2023-04-20
Magic fix for "connection was not ready".

meowjesty:
- Issue: #1302

1. Applies a magic fix that should help with hyper's "connection was not ready" error.
2. Adds some more logs for HTTP stealer in layer.


mirrord/layer/src/tcp_steal/http.rs
```diff
@@ -84,6 +85,7 @@ pub(super) trait HttpV: Sized {
 
     /// Sends the [`HttpRequest`] through `Self::sender`, converting the response into a
     /// [`HttpResponse`].
+    #[tracing::instrument(level = "trace", skip(self))]
```
aviramha, line 88: skip `HttpRequest` or it's without data?

mirrord/layer/src/tcp_steal/http.rs
```diff
@@ -101,6 +103,8 @@ pub(super) trait HttpV: Sized {
 
         // Retry once if the connection was closed.
         if let Err(HttpForwarderError::ConnectionClosedTooSoon(request)) = response {
+            trace!("Request {request:#?} connection was closed too soon, retrying once!");
```
aviramha, line 106: same as above?

aviramha (approved):
LGTM besides few nitpicks!

mirrord/layer/src/tcp_steal/http.rs
meowjesty: Yeah, the `body` is logged as just `bytes (length): 100` for example.

Issue #574, opened 2022-10-14
Agent-only (DNS+network) mode


tamasfe:
It's a bit of a niche, but there are occasions when I'd just like to access some service or services within the cluster with no container to impersonate.

Usually `k port-forward` is sufficient, however when domain names get involved (e.g. with multiple kafka brokers), it is just not enough.

It is currently possible to achieve this with mirrord, however one always has to find a pod to impersonate, even though the pod itself is not relevant.

I'd imagine something like this:

```sh
mirrord dns -n foo-namespace curl http://internal-service:1234
```

TODO:
- [x] Add the feature to mirrord.
- [x] #1351
- [x] https://github.com/metalbear-co/mirrord.dev/issues/72
- [x] #1375
- [x] #1376
- [x] https://github.com/metalbear-co/metalbear.co/issues/50

t4lz:
> ```shell
> mirrord dns -n foo-namespace curl http://internal-service:1234
> ```

Here we're still executing some program, so I would say that `exec` should also be used for that new use-case. Maybe that mode could be enabled by a special `--target`? Maybe something like `--target cluster`? I think this functionality would also go well with the feature suggested in https://github.com/metalbear-co/mirrord/issues/567.

eyalb181:
How about just not specifying a target?

t4lz:
> How about just not specifying a target?

Yeah that would be the nicest for users that want to use this feature, but I thought using an explicit target might be nicer for users that actually do want to use a target but forgot to specify it. Another possible alternative is to have a `--no-target` flag.

tamasfe:
> How about just not specifying a target?

This would feel natural to me, just typing `mirrord exec foo` would cause `foo` to "run inside" the cluster. I think omitting a target and a message at the beginning should be sufficient.

t4lz:
My plan is to implement it in such a way that the agent is completely independent (could also run on a cluster that has no pods at all), but before I continue implementing it this way, I wanted to ask if anyone thinks we should instead choose an arbitrary target pod and use that pod if the user does not specify one. This second approach would mean less changes to the code, but would mean we do need an existing target pod.

t4lz:
Ok, feedback from discord is that the agent should indeed be completely independent. So the next question is: we should not support an ephemeral agent and independent agent mode together, right? When users choose ephemeral agents, they must specify a target, yeah?

aviramha:
[...]

PR #1343, opened 2023-04-21
Support targetless agents.

t4lz:
#574.

When not specifying any target, a targetless agent is spawned, which is independent of any other pod.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -544,7 +544,10 @@ jobs:
         with:
           container-runtime: ${{matrix.container-runtime}}
       - run: minikube image load /tmp/test.tar
-      - name: cargo test containerd
+      # By running the test of the targetless agent first, we prove it works on an empty cluster without any pods.
+      - name: Run targetless E2E test.
+        run: cargo test -p tests targetless
+      - name: Run all E2E test
         run: cargo test -p tests
       - name: Collect logs
         if: ${{ failure() }}
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -190,8 +190,8 @@ impl MirrordExecution {
         })
     }
 
-    /// Some features don't make a lot of sense together with targetless agents, so if needed,
-    /// warn the user, if possible, disable unnecessary features.
+    /// Some features are incompatible with targetless agents, return error if they are set and
+    /// there is no target.
     ///
     /// # Errors
     ///
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -17,7 +17,7 @@ use tracing::trace;
 
 use crate::{
     connection::{create_and_connect, AgentConnectInfo},
-    error::CliError,
+    error::{CliError, CliError::IncompatibleWithTargetless},
     extract::extract_library,
     Result,
 };
@@ -52,6 +52,7 @@ impl MirrordExecution {
     where
         P: Progress + Send + Sync,
     {
+        Self::check_config_for_targetless_agent(config)?;
         let lib_path = extract_library(None, progress, true)?;
         let mut env_vars = HashMap::new();
         let (connect_info, mut connection) = create_and_connect(config, progress).await?;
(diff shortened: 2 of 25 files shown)


mirrord/cli/src/execution.rs
```diff
@@ -189,6 +194,40 @@ impl MirrordExecution {
...
+    ///
+    /// [`IncompatibleWithTargetless`] if both targetless and ephemeral agent.
+    fn check_config_for_targetless_agent(config: &mut LayerConfig) -> Result<()> {
+        if matches!(config.target.path, Target::Targetless) {
+            if config.feature.fs.is_active() {
+                // active by default, so we don't want to fail on this and force users to disable,
+                // just disable it for them.
+                warn!("Disabling file system operations because no target was specified and the agent is not going to read the filesystem of any existing pod.");
+                config.feature.fs = FsConfig {
+                    mode: FsModeConfig::Local,
```
aviramha, line 210: I think this needs to be `LocalWithOverrides` since some files are needed (`/etc/hostname`/`/etc/hosts`)

aviramha: I'm thinking maybe that's an overkill to have this override :\
If we want to leave it, I'd remove the warning and just mention in docs it's supported only with `LocalWithOverrides` unlike the different target modes which you can specify.

t4lz: ok, removed the fs check.

eyalb181 (approved):
A few nitpicks for your consideration, but approved! Well done.

mirrord/kube/src/api.rs
```diff
@@ -30,6 +30,28 @@ where
     }
 }
 
+/// Get a vector of namespaces from an optional namespace. If the given namespace is Some, then
+/// fetch its Namespace object, and return a vector only with that.
+/// If the namespace is Some - return all namespaces.
```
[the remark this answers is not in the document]
t4lz: Yes, thanks!

mirrord-schema.json
```diff
@@ -668,21 +668,18 @@
               "$ref": "#/definitions/PodTarget"
             }
           ]
+        },
+        {
+          "description": "Run without a target. An independent agent pod will be spawned in an arbitrary node.",
```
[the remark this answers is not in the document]
t4lz: Good question, I guess it would not have any effect on functionality, right? But maybe some users would like to choose the node for cloud organization reasons, idk. We could also add it in a separate PR, or open an issue and prioritize it first, to deliver this feature asap. wdyt?

mirrord/kube/src/api/container.rs
```diff
@@ -213,7 +217,9 @@ impl ContainerApi for JobContainer {
                                 "imagePullPolicy": agent.image_pull_policy,
```
aviramha, line 217: we don't need securityContext as well, also volume mounts not needed..

mirrord/kube/src/api/container.rs
```diff
@@ -189,7 +194,6 @@ impl ContainerApi for JobContainer {
 
                     "spec": {
                         "hostPID": true,
```
aviramha, line 196: no need if targetless

mirrord-schema.json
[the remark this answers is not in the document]
aviramha: I don't need we need it unless someone asks for it.

mirrord/cli/src/execution.rs
```diff
@@ -189,6 +190,31 @@ impl MirrordExecution {
         })
     }
 
+    /// Some features don't make a lot of sense together with targetless agents, so if needed,
+    /// warn the user, if possible, disable unnecessary features.
```
Razz4780, line 194: This new logic doesn't actually warn the user nor disable features, it just stops the execution with an error. The whole command is aborted, I think

mirrord/layer/src/lib.rs
```diff
@@ -426,6 +430,10 @@ fn layer_start(config: LayerConfig) {
         .set(config.feature.network.incoming.ignore_localhost)
         .expect("Setting INCOMING_IGNORE_LOCALHOST singleton");
 
+    TARGETLESS
+        .set(config.target.is_none())
+        .expect("Setting INCOMING_IGNORE_LOCALHOST singleton");
```
Razz4780, line 435:
```suggestion
        .expect("Setting TARGETLESS singleton");
```

.github/workflows/ci.yaml
```diff
@@ -532,7 +532,10 @@ jobs:
         with:
           container-runtime: ${{matrix.container-runtime}}
       - run: minikube image load /tmp/test.tar
-      - name: cargo test containerd
+      # By running the test of the targetless agent first, we prove it works on an empty cluster without any pods.
+      - name: Run targetless E2E test.
+        run: cargo test -p tests targetless
+      - name: Run all E2E test
```
Razz4780, line 538: `targetless` will run twice :<

t4lz: Yeah, once on an empty cluster, and once on a cluster with pods. This is good. The test should not take a long time either.

PR #1345, opened 2023-04-23
Add operator message

eyalb181:


mirrord/cli/src/connection.rs
```diff
@@ -71,6 +71,12 @@ where
             AgentConnection { sender, receiver },
         ))
     } else {
+        if matches!(config.target.path, Some(mirrord_config::target::Target::Deployment{..})) {
+            // progress.subtask("text").done_with("text");
+            println!("When targeting multi-pod deployments, mirrord impersonates the first pod in the deployment.\n \
```
aviramha, line 76: I would use `eprintln` since using `println` might interfere with the real app.

eyalb181: Done

aviramha (approved):

PR #1349, opened 2023-04-24
Last selected target prioritized in VS Code

### LINKED ISSUE #1348: VSCode

Razz4780:
Resolves #1348

The first entry in target selection list is now the last used target. The rest of the list is sorted.

### DIFF
--- a/vscode-ext/src/extension.ts
+++ b/vscode-ext/src/extension.ts
@@ -40,6 +40,9 @@ const MIRRORD_DIR = function () {
 const versionCheckEndpoint = 'https://version.mirrord.dev/get-latest-version';
 const versionCheckInterval = 1000 * 60 * 3;
 
+/// Key used to store the last selected target in the global state.
+const LAST_TARGET_KEY = "mirrord-last-target";
+
 let buttons: { toggle: vscode.StatusBarItem, settings: vscode.StatusBarItem };
 let globalContext: vscode.ExtensionContext;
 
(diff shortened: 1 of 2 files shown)


vscode-ext/src/extension.ts
```diff
@@ -153,7 +157,23 @@ class MirrordAPI {
 			throw new Error("error occured listing targets: " + stderr);
 		}
 
-		return JSON.parse(stdout);
+		const targets: string[] = JSON.parse(stdout);
+		const lastTarget = globalContext.globalState.get(LAST_TARGET_KEY);
```
aviramha, line 161: I would first check if there's workspace state then use global state.

vscode-ext/src/extension.ts
```diff
@@ -153,7 +157,23 @@ class MirrordAPI {
...
+		targets.sort((t1, t2) => {
```
aviramha, line 162: I would just call `sort` then 
```js
let pop = targets.pop(lastTarget)
if (pop) {
targets.unshift(pop)
}
```

aviramha (approved):

aviramha (changes requested):
Missclick

vscode-ext/src/extension.ts
Razz4780: Thanks, indeed looks better

vscode-ext/src/extension.ts
Razz4780: Added :heavy_check_mark:

aviramha (approved):
GJ!

PR #1361, opened 2023-04-27
Add warning on version mismatch.

eyalb181:


changelog.d/warn-on-version-mismatch.added.md
t4lz, line 1: I think towncrier wants a `+` prefix for the filename if it's not an issue number.

Cargo.toml
```diff
@@ -60,6 +60,8 @@ libc = "0.2"
 bimap = "0.6.2"
 socket2 = { git = "https://github.com/t4lz/socket2", branch = "unix-sockets", features = ["all"]}
 which = "4"
+semver = "1"
```
t4lz, line 63: The one crate that would never make breaking changes on a minor version 🙃

t4lz (approved):

CHANGELOG.md
aviramha, line 1: ?

aviramha (approved):

PR #1362, opened 2023-04-27
Detect Python debugger PyCharm

### LINKED ISSUE #1020: PyCharm debugger can't attach
### Bug Description

When using PyCharm with debugger, PyCharm can't connect to the debugee's listening port.
`/home/user/PycharmProjects/pythonProject/venv/bin/python /opt/pycharm-professional/plugins/python/helpers/pydev/pydevd.py --module --multiprocess --qt-support=auto --client 127.0.0.1 --port 36141 --file flask run --port=5000 `

error: 
`Could not connect to 127.0.0.1: 36141`

### Steps to Reproduce

pycharm run debugger with flask

### Backtrace

_No response_

### Relevant Logs

_No response_

Linux

### Local process

Python

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Fixes #1020 

Added new `DebuggerType` variant for PyDevD used by PyCharm.

Also fixed `:runIde` command for PyCharm Community and Ultimate

### DIFF
--- a/mirrord/layer/src/debugger_ports.rs
+++ b/mirrord/layer/src/debugger_ports.rs
@@ -22,9 +22,33 @@ pub const MIRRORD_DETECT_DEBUGGER_PORT_ENV: &str = "MIRRORD_DETECT_DEBUGGER_PORT
 pub const MIRRORD_IGNORE_DEBUGGER_PORTS_ENV: &str = "MIRRORD_IGNORE_DEBUGGER_PORTS";
 
 /// Type of debugger which is used to run the user's processes.
+/// Determines the way we parse the command line arguments the debugger's port.
+/// Logic of processing the arguments is based on examples taken from the IDEs.
 #[derive(Debug, Clone, Copy)]
 pub enum DebuggerType {
+    /// An implementation of the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/) for Python 3.
+    /// Used in VS Code.
+    ///
+    /// Command used to invoke this debugger looked like
+    /// `/path/to/python /path/to/vscode/extensions/debugpy --connect 127.0.0.1:57141
+    /// --configure-qt none --adapter-access-token
+    /// c2d745556a5a571d09dbf9c14af2898b3d6c174597d6b7198d9d30c105d5ab24 /path/to/script.py`
+    ///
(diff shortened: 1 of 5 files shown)


mirrord/layer/src/debugger_ports.rs
```diff
@@ -33,34 +57,47 @@ impl FromStr for DebuggerType {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
```
aviramha, line 58: looking at this PR, I think we shouldn't have a variant for each `DebuggerType` since the detection and port extracting logic is quite similar and repeated..
A good question would be why not have the enum's impl as a standalone function that just returns port to ignore? holding the variant doens't help us anyway?

aviramha: to be more concrete - I am afraid that the detection here is too "lite" while just trying to extract the port is way more full

Razz4780: > to be more concrete - I am afraid that the detection here is too "lite" while just trying to extract the port is way more full

I'm not sure if I follow. When we use this feature we know exactly which debugger is used.

aviramha: I am saying we shouldn't split the logic into "detection" and "extraction" - just try to extract and if it fails fallback to env port and if that is missing/fails don't ignore anything.

aviramha: instead of 
```rs
let debugger_type = "pydevd".parse()
let ignore_port = debugger_type.get_port()
```
do
```rs
let ignore_port = get_debugger_ignore_port()

fn get_debugge_ignore_port() -> Option<u16> {    
  if python {
        if pydevd {}
        else if otherthingy {}
        else { envport}
    } else { envport} 
}

Razz4780: Hmm, this detection thing was added having a specific case in mind - when the debuggers are invoked by the IDEs and the users have no way to select the port themselves. If the user has control over the port selection, I'd rather leave it up to them to explicitly ignore it in the config (or not, if that is their use case). But if you think this would improve UX, I can change it

Razz4780: Resolving after a call :heavy_check_mark:

aviramha (approved):

aviramha:
Great work!

PR #1363, opened 2023-04-27
Operator Session ID

DmitryDodzin:
Add x-session-id to operator request

### DIFF
--- a/mirrord/cli/src/operator.rs
+++ b/mirrord/cli/src/operator.rs
@@ -126,11 +126,12 @@ Operator License
 
     let mut sessions = Table::new();
 
-    sessions.add_row(row!["Target", "User", "Session Duration"]);
+    sessions.add_row(row!["Session ID", "Target", "User", "Session Duration"]);
 
     if let Some(status) = mirrord_status.status {
         for session in &status.sessions {
             sessions.add_row(row![
+                session.id.clone().unwrap_or_default(),
                 &session.target,
                 &session.user,
                 humantime::format_duration(Duration::from_secs(session.duration_secs)),
--- a/mirrord/cli/src/operator.rs
+++ b/mirrord/cli/src/operator.rs
@@ -131,7 +131,7 @@ Operator License
     if let Some(status) = mirrord_status.status {
         for session in &status.sessions {
             sessions.add_row(row![
-                session.id.clone().unwrap_or_default(),
+                session.id.as_deref().unwrap_or(""),
                 &session.target,
                 &session.user,
                 humantime::format_duration(Duration::from_secs(session.duration_secs)),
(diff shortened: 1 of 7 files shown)


mirrord/operator/src/client.rs
```diff
@@ -69,7 +71,17 @@ impl OperatorApi {
         let target_api: Api<TargetCrd> =
             get_k8s_resource_api(&client, target_config.namespace.as_deref());
 
+        let session_id = std::env::var(MIRRORD_OPERATOR_LAYER_SESSION)
```
Razz4780, line 74: Coud you add docs to the public `discover` method above, saying that it now uses env? It's always surprising, when function's behavior depends on something more than its arguments

mirrord/cli/src/operator.rs
```diff
@@ -126,11 +126,12 @@ Operator License
 
     let mut sessions = Table::new();
 
-    sessions.add_row(row!["Target", "User", "Session Duration"]);
+    sessions.add_row(row!["Session ID", "Target", "User", "Session Duration"]);
 
     if let Some(status) = mirrord_status.status {
         for session in &status.sessions {
             sessions.add_row(row![
+                session.id.clone().unwrap_or_default(),
```
Razz4780, line 134:
```suggestion
                session.id.as_ref().unwrap_or(""),
```
Nitpick

DmitryDodzin: Need to use `as_deref` because otherwise `unwrap_or` expects `&String` instead of `&str` =]

mirrord/operator/src/client.rs
```diff
@@ -17,6 +17,7 @@ use tracing::error;
 use crate::crd::{MirrordOperatorCrd, TargetCrd, OPERATOR_STATUS_NAME};
 
 static CONNECTION_CHANNEL_SIZE: usize = 1000;
+static MIRRORD_OPERATOR_LAYER_SESSION: &str = "MIRRORD_OPERATOR_LAYER_SESSION";
```
eyalb181, line 20: This is the ID right? So MIRRORD_OPERATOR_SESSION_ID?

mirrord/operator/src/client.rs
```diff
@@ -137,10 +138,23 @@ impl OperatorApi {
...
+    ///
+    /// Note: Uses `MIRRORD_OPERATOR_LAYER_SESSION` to have a persistant session_id across child
+    /// processes
     async fn connect_target(
         &self,
         target: TargetCrd,
     ) -> Result<(mpsc::Sender<ClientMessage>, mpsc::Receiver<DaemonMessage>)> {
+        let session_id: u64 = std::env::var(MIRRORD_OPERATOR_LAYER_SESSION)
+            .ok()
+            .and_then(|val| val.parse().ok())
```
eyalb181, line 151: Shouldn't we differentiate between parse failing (invalid) and no ID exists (valid)?

eyalb181 (changes requested):

eyalb181 (approved):

Issue #1123, opened 2023-02-26
Better IP bind/listen flow


aviramha:
while working on #1121 I found out that on my setup the listen fails because it binds each IP separately, leading to `ListenAlreadyExists` error and killing the layer.
I am sending a PR that ignores this error, but we need in `bind` call to check if address is not already bound to preserve normal behavior (return error in that case instead of ignoring it)

Edit:
We should hook all the functions that help resolve the local stack (ip, subnet, ports) so app want to listen on "localhost" it will then resolve to the real remote ip and will be able to bind on that ip, so `getsockname` would return also the remote ip even before accepting any connection.

gautamprikshit1:
Can you provide a little more insight on this one. Thanks

aviramha:
> Can you provide a little more insight on this one. Thanks

Sure. It's a bit complex issue so feel free to ask more questions.

mirrord hooks calls to bind/listen, so instead of actually binding/listening the given ports we randomize a port then pass the data to the agent so it can know what connections to steal/mirror for us.
Due to the IP being different between local/remote, we currently just ignore cases where the same port is bound twice. We ignore it because to preserve "proper" behavior we need to check if IP:PORT combination is bound, and then error in that case instead of returning success. Right now we don't preserve the bounded IP, only port so in order to support apps that bound each IP.

After writing this, I think maybe we should consider the IP specified in bind to be passed to the remote agent, but for it to work we need to also make sure that the local applications is able to obtain the remote's IPs to bind those.

meowjesty:
While investigating #1054 we've found out that `bind` has to be more "correct" than what we have right now.

When running the [flask test](https://github.com/metalbear-co/mirrord/blob/main/mirrord/layer/tests/apps/app_flask.py), the server address stays as some `127.0.0.1` (initial address), even though internally the requests are being handled in `1.1.1.1`.

- Some logs to give a bit more information:

```sh
# first call to `getsockname`
stderr 2023-03-02T20:04:08.775908614Z 62732: 2023-03-02T20:04:08.775887Z DEBUG ThreadId(01) getsockname: mirrord_layer::socket::ops: getsockname -> local_address 127.0.0.1:33839 sockfd=10

(...)

# correct address
stderr 2023-03-02T20:04:08.787741559Z 62732: 2023-03-02T20:04:08.787718Z DEBUG ThreadId(01) getsockname: mirrord_layer::socket::ops: getsockname -> local_address 1.1.1.1:80 sockfd=11

[...]

PR #1371, opened 2023-05-01
Prevent double bind on same address.

meowjesty:
- Issue: #1123

1. Solves the issue where we allowed binding the same `{address:port}` twice;
2. Adds a new CI run (build issue1123 integration test);

Doesn't close the issue yet though, as we still need to "fake" the ip as the impersonated pod ip, which involves some changes to `getsockname`, `bind`, and probably `listen`.


mirrord/layer/src/socket/ops.rs
```diff
@@ -165,6 +165,21 @@ pub(super) fn bind(
...
 
+    // Check if the user's requested address isn't already in use, even though it's not actually
+    // bound, as we bind to a different address, but if we don't check for this then we're
+    // changing normal socket behavior (see issue #1123).
+    if SOCKETS.iter().any(|socket| match &socket.state {
+        SocketState::Initialized => false,
+        SocketState::Bound(bound) | SocketState::Listening(bound) => {
+            bound.requested_address == requested_address
+        }
+        SocketState::Connected(connected) => SocketAddr::try_from(connected.local_address.clone())
```
aviramha, line 176: This shouldn't return bound AFAIK. Once connection is made a new socket is created and bound/listening socket state stays the same and can also be closed letting other binding/listening happen

meowjesty: You mean for the `Socket::Connected` state only?

mirrord/layer/tests/common/mod.rs
```diff
@@ -124,6 +124,7 @@ impl TestProcess {
     }
 
     pub fn assert_stdout_contains(&self, string: &str) {
+        println!("{:#?}", self.stdout.lock().unwrap().as_str());
```
Razz4780, line 127: Is this needed? Will clutter the output for all tests

mirrord/layer/tests/issue1054.rs
```diff
@@ -0,0 +1,45 @@
...
+    println!("Application subscribed to port, sending tcp messages.");
+
+    let new_connection_id = layer_connection
+        .send_new_connection(application.get_app_port())
+        .await;
+    layer_connection.send_close(new_connection_id).await;
+
+    test_process.wait().await;
+    test_process.assert_stdout_contains("test issue 1054: START");
+    test_process.assert_stdout_contains("test issue 1054: SUCCESS");
```
Razz4780, line 44: Nitpick: we could just check the return code

mirrord/layer/tests/issue1123.rs
```diff
@@ -0,0 +1,37 @@
...
+            ],
+            None,
+        )
+        .await;
+
+    println!("Application subscribed to port, sending tcp messages.");
+
+    test_process.wait().await;
+    test_process.assert_stdout_contains("test issue 1123: START");
+    test_process.assert_stdout_contains("test issue 1123: SUCCESS");
```
Razz4780, line 36: Nitpick: we could just check the return code

Razz4780 (changes requested):

Razz4780 (approved):
Looks good!

PR #1372, opened 2023-05-01
Advanced sessions

DmitryDodzin:


mirrord/operator/src/client/certificate.rs
```diff
@@ -0,0 +1,205 @@
+use std::{fmt, ops::Deref, path::PathBuf, sync::LazyLock};
```
aviramha, line 1: I tend to believe this belongs in the `auth` crate?

mirrord/operator/src/client.rs
```diff
@@ -14,7 +14,11 @@ use tracing::error;
 
 use crate::crd::TargetCrd;
 
+#[cfg(feature = "client-certificate")]
+pub mod certificate;
+
 static CONNECTION_CHANNEL_SIZE: usize = 1000;
+static MIRRORD_OPERATOR_LAYER_SESSION: &str = "MIRRORD_OPERATOR_LAYER_SESSION";
```
eyalb181, line 21:
```suggestion
static MIRRORD_OPERATOR_SESSION_ID: &str = "MIRRORD_OPERATOR_SESSION_ID";
```

mirrord/operator/src/client/certificate.rs
```diff
@@ -0,0 +1,205 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct ClientCredentials {
+    root: SerdeCertificate,
+
+    pub client_certificate: ClientCertificate,
+}
+
+impl ClientCredentials {
+    pub async fn accuire(client: Client) -> Result<Self> {
```
eyalb181, line 168:
```suggestion
    pub async fn acquire(client: Client) -> Result<Self> {
```

mirrord/operator/src/client/certificate.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use base64::{engine::general_purpose, Engine as _};
+use kube::{api::PostParams, Api, Client};
+use rcgen::{Certificate, CertificateParams, KeyPair};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use thiserror::Error;
+use tokio::fs;
+use tracing::warn;
+
+static CREDENTIALS_DIR_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
```
eyalb181, line 11: Use this?
https://doc.rust-lang.org/std/env/fn.home_dir.html

mirrord/operator/src/client/certificate.rs
```diff
@@ -0,0 +1,205 @@
...
+            .or_else(|_| std::env::var("HOMEPATH"))
+            .unwrap_or_else(|_| "~".to_owned()),
+    )
+    .join(".mirrord")
+});
+
+static CREDENTIALS_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
+    std::env::var("MIRRORD_AUTHENTICATION")
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| CREDENTIALS_DIR_PATH.join("credentials"))
```
eyalb181, line 23: Why isn't "credentials" joined to the path in the previous declaration?

mirrord/operator/src/client/certificate.rs
```diff
@@ -0,0 +1,205 @@
...
+    Rcgen(#[from] rcgen::RcgenError),
+    #[error(transparent)]
+    Kube(#[from] kube::Error),
+    #[error(transparent)]
+    SerdeJson(#[from] serde_json::Error),
+}
+
+type Result<T, E = CertificateError> = std::result::Result<T, E>;
+
+fn serialize_base64_der<S: Serializer>(der: &Vec<u8>, ser: S) -> Result<S::Ok, S::Error> {
```
eyalb181, line 44: What is der?

eyalb181 (changes requested):

mirrord/auth/src/credential_store.rs
```diff
@@ -0,0 +1,85 @@
...
+        if !credentials.is_ready() {
+            let common_name = self
+                .common_name
+                .clone()
+                .or_else(|| gethostname::gethostname().into_string().ok())
+                .unwrap_or_default();
+
+            credentials
+                .get_client_certificate::<R>(client.clone(), &common_name)
+                .await?;
```
t4lz, line 80: The way I understood the requirements, when a user works with different clusters that have different licenses, `mirrord` should identify the correct license automatically by the CN.
In this code, the used certificate is determined by the `active` configuration field, not by which operator mirrord is connecting to.
Is automatic license choosing a requirement, or not @aviramha  / @eyalb181?

Either way - with this code, I think that if a user uses a mirrord operator with one license, and then tries to use a second mirrord operator with a different license, mirrord would try to authenticate with an invalid certificate unless they manually edit the credential store file and set a new `active`. Does that fulfil the requirements? Should users be manually editing that file?
In my opinion it would be better to identify credentials in the store by the operator they are valid for (maybe by the *common name* of the license of that operator?) instead of by the `active` value at the time of their creation. This makes more sense together with automatic credential choosing.

t4lz (commented):
The questions about the credential store are the most important.

Also, I think you could add some docs, I commented where I was missing them the most. Let me know if you disagree.

Impressive stuff!

mirrord/auth/src/credential_store.rs
```diff
@@ -0,0 +1,85 @@
...
+        R: Resource + Clone + Debug,
+        R: for<'de> Deserialize<'de>,
+        R::DynamicType: Default,
+    {
+        let credentials = match self.client_credentials.entry(self.active.clone()) {
+            Entry::Vacant(entry) => entry.insert(Credentials::init()?),
+            Entry::Occupied(entry) => entry.into_mut(),
+        };
+
+        if !credentials.is_ready() {
```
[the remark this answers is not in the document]
DmitryDodzin: I want the `is_ready` to fail if certificate is out of date, so a new one should be created in it's place

mirrord/auth/src/credential_store.rs
```diff
@@ -0,0 +1,85 @@
...
+        let buffer = fs::read(&*CREDENTIALS_PATH)
+            .await
+            .map_err(CertificateStoreError::from)?;
+
+        serde_yaml::from_slice(&buffer)
+            .map_err(CertificateStoreError::from)
+            .map_err(AuthenticationError::from)
+    }
+
+    pub async fn save(&self) -> Result<()> {
```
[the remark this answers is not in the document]
DmitryDodzin: Iv'e replaced how the interaction with FS works so edge cases should be avoided

mirrord/auth/src/credential_store.rs
t4lz: Ok, so when we introduce multiple-license support we'll have kind of a backward compatible logic of - if there is no credentials entry for this operator - try "default", and only if that fails - request a new one? Is that the plan, @DmitryDodzin?

mirrord/operator/src/crd.rs
```diff
@@ -83,3 +82,11 @@ pub struct Session {
     pub user: String,
     pub target: String,
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct LicenseInfoOwned {
+    pub name: String,
+    pub organization: String,
+    pub expire_at: NaiveDate,
+    pub fingerprint: Option<String>,
```
t4lz, line 91: Why is this optional?

t4lz (approved):

PR #1373, opened 2023-05-03
Remember last selected target in IntelliJ

### LINKED ISSUE #1347: Prioritize last used target in target selection in IntelliJ

infiniteregrets:
closes #1347

### DIFF
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordExecDialog.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordExecDialog.kt
@@ -18,7 +18,15 @@ object MirrordExecDialog {
     private const val targetLabel = "Select Target"
 
     fun selectTargetDialog(targets: List<String>): String? {
-        val jbTargets = targets.sorted().asJBList()
+        var targets = targets.sorted().toMutableList()
+        MirrordSettingsState.instance.mirrordState.lastChosenTarget?.let {
+            val idx = targets.indexOf(it)
+            if (idx != -1) {
+                targets.removeAt(idx)
+                targets.add(0, it)
+            }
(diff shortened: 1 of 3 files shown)


intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordExecDialog.kt
```diff
@@ -16,9 +16,19 @@ import javax.swing.event.DocumentListener
 object MirrordExecDialog {
     private const val dialogHeading: String = "mirrord"
     private const val targetLabel = "Select Target"
+    private var lastSelectedTarget: String? = null
```
aviramha, line 19: Why do we need to store it here and in the settings too?

aviramha (changes requested):

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordExecDialog.kt
infiniteregrets: trying to keep the behavior same as what was done here https://github.com/metalbear-co/mirrord/pull/1349/files we keep it for both global and workspace and I'm trying the same by mapping it persistent storage for global

infiniteregrets: you mean in case of intellij? the one stored in the object variable will be lost when you reopen the workspace

aviramha: The difference between the object and the setting is just that one is persistent, so why not always use the persistent one?

aviramha (approved):

PR #1378, opened 2023-05-03
Display mirrord loading progress in the status bar

### LINKED ISSUE #1337: Not getting the "mirrord started!" notification on IntelliJ based IDEs
### Bug Description

When running programs with mirrord on JetBrains products, the last notification I see is mirrord preparing to launch, and it stays there, and the "mirrord started!" notification does not appear.
image

In some cases it's only the "extracting layer" notification.

### Steps to Reproduce

Run the python-e2e app on pycharm or the sample/node app on IntelliJ.

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS 13.3

### Local process

python3

### Local process version

_No response_

### Additional Info

_No response_

infiniteregrets:
closes #1337

the above-mentioned issue mentions "mirrord started!" is never displayed, this happened because of the underlying implementation of the Signleton notifier being used as it did not display the notification if a bubble had not expired. To fix this we use the ProgressManager API, however we still don't usually see "mirrord started!" because it happens too fast.

some considerations while making this change:

- 
```
  /**
   * Runs the given process synchronously in calling thread, associating this thread with the specified progress indicator.
   * This means that it'll be returned by {@link ProgressManager#getProgressIndicator()} inside the {@code process},
   * and {@link ProgressManager#checkCanceled()} will throw a {@link ProcessCanceledException} if the progress indicator is canceled.
   *
   * @param progress an indicator to use, {@code null} means reuse current progress
   */
  public abstract void runProcess(@NotNull Runnable process, @Nullable ProgressIndicator progress) throws ProcessCanceledException;
```

this works the same as what has been implemented in the PR except we associate it to a particular progress indicator which is `ProgressManager.getInstance().progressIndicator` and we can use it directly

- 
```
  /**
   * Runs a specified {@code task} in either background/foreground thread and shows a progress dialog.
   *
   * @param task task to run (either {@link Task.Modal} or {@link Task.Backgroundable}).
   */
  public abstract void run(@NotNull Task task);
```

this and most other methods run on dialog boxes and not on status indicators, i haven't been able to find a way to run a `Task` with a custom indicator

- 
```
public abstract void runProcessWithProgressAsynchronously(@NotNull Task.Backgroundable task, @NotNull ProgressIndicator progressIndicator);
```
errors out and recommends to use: 
```
  /**
   * @param progress an indicator to use, {@code null} means reuse current progress
   */
  public abstract void executeProcessUnderProgress(@NotNull Runnable process, @Nullable ProgressIndicator progress) throws ProcessCanceledException;
```
, also not desirable IMO since its run in a background thread

-- 
Should the buffer loop be run as a background task or a foreground task using `Task`, `Runnable` constructs? or are we okay with running it in sync as implemented right now while directly accessing the ProgressManager

### DIFF
--- a/changelog.d/1337.changed.md
+++ b/changelog.d/1337.changed.md
@@ -1,0 +1,1 @@
+mirrord loading progress is displayed in the staus indicator on IntelliJ, replacing the singleton notifier
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordApi.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordApi.kt
@@ -4,7 +4,11 @@ import com.google.gson.Gson
 import com.intellij.execution.configurations.GeneralCommandLine
 import com.intellij.execution.wsl.WSLCommandLineOptions
 import com.intellij.execution.wsl.WSLDistribution
+import com.intellij.openapi.progress.ProgressIndicator
+import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.progress.Task
 import com.intellij.openapi.project.Project
+import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit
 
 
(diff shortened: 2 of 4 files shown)


eyalb181 (approved):
Nice!

aviramha:
This PR lead to null exception on the New UI. I am reverting this and I believe we should stick to the notification thingy and just fix it.

PR #1379, opened 2023-05-03
Improve mirrord-config docs to be more user friendly.

meowjesty:
- Issue: 

Changes the `config` crate docs to be more digestible by users, with the intent of turning these into the documentation for the user's `config.json` file on mirrord.dev.

- Adds _json-ish_ type examples;
- Adds config examples in json.

(wip)

Feedback welcome.


mirrord/config/src/agent.rs
```diff
@@ -5,23 +5,75 @@ use schemars::JsonSchema;
 
 use crate::config::source::MirrordConfigSource;
 
+/// # agent
+///
 /// Configuration for the mirrord-agent pod that is spawned in the Kubernetes cluster.
+///
+/// ## Types
+///
+/// ```json
```
eyalb181, line 14: These examples are all going to be really difficult to keep up to date manually

meowjesty: Agree, but I'm not sure of a good way to do this automatically.

Like, the json documentation thingies always put those crappy `anyOf (anyOf (anyOf null undefined Thing String anyOf (null Thing))))` that is borderline useless.

Maybe we could write our own tool to generate examples/type definitions for us?

meowjesty: The more I look at the generated markdown, the more I wonder if we even need these type definitions at all.

I can't recall a single tool that I have used that gave me all of these types for something. And I don't think users care about `foo: null | String | Whatever`. 

As a user myself, I want a docs page to have examples of the most common stuff (chances are my config will fit in one of these cases), and tell me what the properties do, so I can change what I need.

What do you think?

An example of this would be [docker config/daemon](https://docs.docker.com/config/daemon/).

PR #1393, opened 2023-05-08
Targetless intellij

t4lz:
#1375 

Allow null targets in the intelliJ plugin, notify about targetless runs.
To run targetless the user can choose the targetless target (or no target).
If the user presses `cancel` or closes the dialog, we cancel the loading of mirrord.


infiniteregrets (approved):
LGTM

PR #1394, opened 2023-05-08
Targetless on vscode

t4lz:
https://github.com/metalbear-co/mirrord/issues/1376

Add targetless target to dialog, notify about targetless runs.
To run targetless the user can choose the targetless target (or press escape).

### DIFF
--- a/vscode-ext/src/extension.ts
+++ b/vscode-ext/src/extension.ts
@@ -433,11 +433,12 @@ class ConfigurationProvider implements vscode.DebugConfigurationProvider {
 			if (targetName) {
 				if (targetName != TARGETLESS_TARGET_NAME) {
 					target = targetName;
-				} else {
-					vscode.window.showInformationMessage("mirrord running targetless");
 				}
 				globalContext.globalState.update(LAST_TARGET_KEY, targetName);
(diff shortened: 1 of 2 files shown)


vscode-ext/src/extension.ts
```diff
@@ -420,21 +422,24 @@ class ConfigurationProvider implements vscode.DebugConfigurationProvider {
...
 					"No mirrord target available in the configured namespace. " +
-					"Set a different target namespace or kubeconfig in the mirrord configuration file.",
+					"You can run targetless, or set a different target namespace or kubeconfig in the mirrord configuration file.",
 				);
-
-				return undefined;
 			}
-
+			targets.push(TARGETLESS_TARGET_NAME);
 			let targetName = await vscode.window.showQuickPick(targets, { placeHolder: 'Select a target path to mirror' });
```
infiniteregrets, line 432: Why not check here as well just like intellij that if the target list is empty and notify the user?

infiniteregrets (approved):
so pressing escape/anywhere else on the screen means target is null we run in target less mode? but on intellij cancel == normal run? and empty ok == targetless? sounds like a limitation of vscode

upto you though, pressing escape from target selection for me means I don't want anything to do with the extension

Issue #1397, opened 2023-05-10
mirrord fails to load to `go run .` on macOS


aviramha:
### Bug Description

```
> mirrord exec -c -n httpbin  -t pod/httpbin-0 go -- run .
⠧ mirrord cli starting
  ✓ ready to launch process
  ✓ layer extracted
  ✓ No Operator Detected
  ✓ agent pod created
  ✓ pod is ready
Starting server at port 8080
Error:
  × Internal proxy error: FirstConnectionTimeout
  ╰─▶ Couldn't accept connection before timeout
```
This happens because mirrord fails to load to Go, after debugging it I found out that the go binary is signed, and can't be loaded without patching. 
After adding it to `sip_binaries` it manages to load, but then fails on `fatal: bad g in signal handler`
debugging it further I found out the issue was actually `"local": ["/tmp", "/var/private", "/private"]` being accessed remotely.

### Steps to Reproduce

1.
```sh
go mod init example.com/app
```
2. Create app.go
```go
package main

import (
    "fmt"
    "net/http"
)

func main() {
    serverOne := &http.Server{
        Addr: ":8080",
        Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
            fmt.Printf("Req: %s %s\n", r.Host, r.URL.Path)
        }),
    }

    fmt.Printf("Starting server at port 8080\n")
    serverOne.ListenAndServe()
}
```
3. run using mirrord `mirrord exec -c -n httpbin  -t pod/httpbin-0 go -- run .`

### Your operating system and version

macOS

### Local process

go

### Local process version

go version go1.20 darwin/arm64

PR #1398, opened 2023-05-10
Fixed `go run` failing because of reading remote files 

aviramha:
tweaked locals overrides to have `/private` `/tmp` and removes `.*$` from the pattern to include directory listing as well

Partially solves #1397

### DIFF
--- a/mirrord/layer/src/file/filter.rs
+++ b/mirrord/layer/src/file/filter.rs
@@ -58,9 +58,10 @@ fn generate_local_set() -> RegexSet {
         r"^/dev/.*$",
         r"^/opt/.*$",
         r"^/home/.*$",
-        r"^/tmp/.*$",
-        r"^/private/.*$",
-        r"^/var/folders/.*$",
+        // don't use `/.*$` so lstat on the directory will also be local
+        r"^/private",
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/file/filter.rs
```diff
@@ -58,9 +58,10 @@ fn generate_local_set() -> RegexSet {
         r"^/dev/.*$",
         r"^/opt/.*$",
         r"^/home/.*$",
-        r"^/tmp/.*$",
-        r"^/private/.*$",
-        r"^/var/folders/.*$",
+        // don't use `/.*$` so lstat on the directory will also be local
+        r"^/private",
+        r"^/var/folders",
+        r"^/tmp",
```
t4lz, line 64:
```suggestion
        r"^/private(/|$)",
        r"^/var/folders(/|$)",
        r"^/tmp(/|$)",
```
If we want to match `/private`, `/private/` and `/private/foo`, but not `/private_key`

aviramha: gee I'm bad at regex, thanks!

t4lz (approved):

PR #1405, opened 2023-05-11
update frida

aviramha:


t4lz (commented):
Results in `zsh: illegal hardware instruction` when running `mirrord exec` on macOS on M2.

Issue #1403, opened 2023-05-11
Strip mirrord's bin directory from file operations


t4lz:
When we hook a file operation on a path that is inside our temp directory - access the path without our directory instead, locally.

aviramha:
when I found it happens I was wondering maybe it's solvable by hooking the function used to determine the location of the binary if there is one?

PR #1406, opened 2023-05-11
Strip mirrord's temp dir from file ops operations, read locally

t4lz:
#1403

If the user's application tries to read files from our temp directory, strip our directory from the path, and read the stripped path locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ num-traits = "0.2"
 regex = "1"
 miette = "5"
 fancy-regex = { version = "0.10" }
+lazy_static = "1"
 
 # TODO(alex): Change when https://github.com/hyperium/http-body/issues/88 is solved.
 hyper = { git = "https://github.com/meowjesty/hyper", branch = "issue/922/http2", features = ["full"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,6 @@ num-traits = "0.2"
 regex = "1"
 miette = "5"
 fancy-regex = { version = "0.10" }
-lazy_static = "1"
 
 # TODO(alex): Change when https://github.com/hyperium/http-body/issues/88 is solved.
 hyper = { git = "https://github.com/meowjesty/hyper", branch = "issue/922/http2", features = ["full"] }
--- a/mirrord/auth/Cargo.toml
+++ b/mirrord/auth/Cargo.toml
@@ -20,7 +20,7 @@ default = []
 webbrowser = ["dep:rand", "dep:reqwest", "dep:webbrowser"]
 
 [dependencies]
-lazy_static = {workspace = true}
+lazy_static.workspace = true
 rand = { version = "0.8", optional = true }
 reqwest = { workspace = true, features=["json", "rustls-tls"], default-features = false, optional = true }
 serde.workspace = true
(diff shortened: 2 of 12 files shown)


mirrord/layer/src/common.rs
```diff
@@ -16,6 +18,18 @@ use crate::{
     HOOK_SENDER,
 };
 
+lazy_static! {
```
aviramha, line 21: put it in `sip` crate so we'll also have it only on macOS

aviramha (commented):
Overall comment - I think the implementation can be that you just edit the original string, as it will always be less characters. It will make the implementation easier.

t4lz:
> just edit the original string

What do you mean? This is already how it's done.

aviramha:
> > just edit the original string
> 
> What do you mean? This is already how it's done.

Aren't you creating a new string then replacing the ptr?

aviramha:
> > just edit the original string
> 
> What do you mean? This is already how it's done.

Oops. Missed that

aviramha:
You should check if the path is of the binary itself, then we should leave it as is.
I wouldn't add a bypass for it, but just change it as part of checked into without changing the behavior - maybe do a bypass if it's that path though

aviramha:
I agree we should bypass on this case, just not a new case of bypass

t4lz:
> I agree we should bypass on this case, just not a new case of bypass

Oh, I see. But I use this bypass variant to pass back the new pointer. How should I do it then?

aviramha:
> > I agree we should bypass on this case, just not a new case of bypass
> 
> Oh, I see. But I use this bypass variant to pass back the new pointer. How should I do it then?

In the checked into?

t4lz:
> In the checked into?

I am returning a `Bypass` with a the new `*const c_char` from `checked_into` if original path is in our dir. How should I do that without introducing a new variant of `Bypass`?

aviramha:
> > In the checked into?
> 
> I am returning a `Bypass` with a the new `*const c_char` from `checked_into` if the original path is in our dir. How should I do that without introducing a new variant of `Bypass`?

I thought `checked_into` requires `self` and not `&self` (because Into does, my bad).

mirrord/layer/src/common.rs
```diff
@@ -16,6 +20,12 @@ use crate::{
     HOOK_SENDER,
 };
 
+#[cfg(target_os = "macos")]
+lazy_static! {
+    /// Path of current executable, None if fetching failed.
+    pub static ref CURRENT_EXE: Option<String> = std::env::current_exe().ok().map(|path_buf| path_buf.to_string_lossy().to_string());
+}
```
infiniteregrets, line 27: Can we not use https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html here

mirrord/auth/Cargo.toml
```diff
@@ -20,7 +20,7 @@ default = []
 webbrowser = ["dep:rand", "dep:reqwest", "dep:webbrowser"]
 
 [dependencies]
-lazy_static = "1"
+lazy_static.workspace = true
 rand = { version = "0.8", optional = true }
```
infiniteregrets, line 24: same

Cargo.toml
```diff
@@ -51,6 +51,7 @@ num-traits = "0.2"
 regex = "1"
 miette = "5"
 fancy-regex = { version = "0.10" }
+lazy_static = "1"
```
infiniteregrets, line 55: no need of these

infiniteregrets (changes requested):

PR #1395, opened 2023-05-08
Config dropdown for intelliJ

infiniteregrets:
closes #1411 

note: a bit unstable rn, wip

PR #1413, opened 2023-05-11
Prevent a possible race condition on config dropdown

infiniteregrets:
Problem: query the index while it is being built is bad

Solution:
If there is an update available, spawn a task that queries the index while blocking new tasks from being started so there is only one task in case there is an update available

### DIFF
--- a/changelog.d/+towncrier.fixed.md
+++ b/changelog.d/+towncrier.fixed.md
@@ -1,0 +1,1 @@
+Prevent a possible race condition
--- a/changelog.d/+towncrier.fixed.md
+++ b/changelog.d/+towncrier.fixed.md
@@ -1,1 +1,0 @@
-Prevent a possible race condition with indexes on intelliJ extension
--- a/changelog.d/+towncrier.fixed.md
+++ b/changelog.d/+towncrier.fixed.md
@@ -1,1 +1,1 @@
-Prevent a possible race condition
+Prevent a possible race condition with indexes on intelliJ extension
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
@@ -17,7 +17,6 @@ import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.util.indexing.*
 import com.intellij.util.io.EnumeratorStringDescriptor
 import com.intellij.util.io.KeyDescriptor
-import com.intellij.util.io.exists
 import java.nio.file.Path
 import java.util.*
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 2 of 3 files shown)


changelog.d/+towncrier.fixed.md
t4lz, line 1: This fixes unreleased code, so I don't think it should be in the changelog (which is all about what's new in each released version). I think you should just make a small change to the entry file you made in the first PR, so that towncrier is happy, but not add a new entry.

t4lz (approved):

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -56,20 +56,18 @@ class MirrordConfigDropDown : ComboBoxAction() {
...
             if (!::configFiles.isInitialized) {
-                val basePath = project.basePath ?: throw Error("couldn't resolve project path")
-                configFiles = FileBasedIndex.getInstance().getAllKeys(MirrordConfigIndex.key, project).filter {
-                    it.startsWith(basePath)
-                }.toHashSet()
+                // there is a possibility of a race condition in case index is accessed while being built
+                // so all index queries need to be done in smart mode
+                runInSmartModeOrDirectly({
+                    initializeConfigs(project)
+                }, project)
```
t4lz, line 64: Nitpick: I think it would be prettier to have `runInSmartModeOrDirectly` take a function that takes a project parameter, so that you can call it like this: `runInSmartModeOrDirectly(initializeConfigs, project)`
(and then in runInSmartModeOrDirectly call `action(project)` instead of `action()`).

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -111,16 +111,46 @@ class MirrordConfigDropDown : ComboBoxAction() {
...
+                            if (project.isDisposed) {
+                                throw ProcessCanceledException()
+                            }
+                            if (dumbService.isDumb) {
+                                return@compute false
+                            }
+                            try {
+                                indicator.text = "mirrord: updating config files"
+                                query(project)
+                                blockQueries.set(false)
```
t4lz, line 135: This synchronisation mechanism is incorrect. Possible scenario:
1. In background task 1 (after file system change): `query(project)` (`updateConfigFiles(project)`) completes.
2. New filesystem change, so in updating thread: `queryIndexInSmartMode` is called, `blockQueries` is still true, so not creating new task.
3. In background task 1: `blockQueries.set(false)` (too late).
Now as long as there is no new file system change, `queryIndexInSmartMode` (`updateConfigFiles`) is not called, and the last change is not visible for the user.

t4lz (changes requested):
In general: if there is no issue, can you please explain the problem the PR solves and how it solves it?

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
t4lz: > but the update flag is still set to false

When the update flag is set to false, the update will not happen in the next call to `update`.

> not even need atomics
About the atomics already in `main`, see https://github.com/metalbear-co/mirrord/pull/1395#discussion_r1190325799. About the atomics added in this PR, I don't know, please explain why this PR is needed.

infiniteregrets:
Merging with the theoretical bug, will be fixed in a separate PR

PR #1419, opened 2023-05-16
Fix the lost update bug with config dropdown for intelliJ

### LINKED ISSUE #1420: Config dropdown for intelliJ can lose updates due to incorrect synchronization
### Bug Description

Some updates can be lost due to queries being blocked when the filesystem change happens

### Steps to Reproduce

This was seen theoretically, however, upon testing a few times I actually ended up with a lost update!

### Backtrace

_No response_

### Relevant Logs

_No response_

macos

### Local process

intellij IDE run with configs

### Local process version

_No response_

### Additional Info

_No response_

infiniteregrets:
closes #1420

An example condition where an update can be lost was found in the following case:

```
query(project)
                            new fs update ==> [updateConfigs =  true]
                                                                                                     update() {
                                                                                                                        update is true, queries are blocked
                                                                                                                        updateConfigs.getAndSet(false) // this condition sets                                                                                                                                                                                                                                                 
                                                                                                                        updateConfigs to false from true and spawns a new 
                                                                                                                        task 
                                                                                                                        but queries are still blocked ===> update is lost!
                                                                                                         }
blockQueries.set(false)
```                                  

Solution:

Querying the index and setting the flag should be done together in a critical section such that no such updates are lost, the task and the background thread share a flag and all reads and writes for the flag are synchronized in a block with a separate lock object and not the class

### DIFF
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
@@ -153,8 +153,9 @@ class MirrordConfigDropDown : ComboBoxAction() {
         }.queue()
     }
 
-    // this function checks the `blockQueries` flag in a synchronized block
-    // and executes the update accordingly
+    // this function checks the `blockQueries` flag and the updateConfigs flag and blocks queries
+    // by laying down concurrent instructions in a manner that everytime updateConfigs is set
+    // a query will follow
     private fun blockAndQueryIndex(project: Project) {
         if (updateConfigs && !blockQueries) {
             blockQueries = true
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
@@ -17,7 +17,6 @@ import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.util.indexing.*
 import com.intellij.util.io.EnumeratorStringDescriptor
 import com.intellij.util.io.KeyDescriptor
-import com.intellij.util.io.exists
 import java.nio.file.Path
 import java.util.*
 import java.util.concurrent.atomic.AtomicBoolean
@@ -119,6 +118,8 @@ class MirrordConfigDropDown : ComboBoxAction() {
             object : Task.Backgroundable(project, "mirrord") {
                 override fun run(indicator: ProgressIndicator) {
                     val dumbService = DumbService.getInstance(project)
+                    // refer to `DumbService`, there is no "guarantee" still, but we stay in a loop
+                    // todo: should probably look into using the message bus to listen for indexing to finish
                     while (true) {
                         dumbService.waitForSmartMode()
                         val success = ReadAction.compute<Boolean, RuntimeException> {
@@ -199,7 +200,7 @@ class MirrordConfigIndex : ScalarIndexExtension<String>() {
 
     override fun getInputFilter(): FileBasedIndex.InputFilter {
         return FileBasedIndex.InputFilter {
-            it.isInLocalFileSystem && !it.isDirectory && it.path.endsWith("mirrord.json") && Path.of(it.path).exists()
+            it.isInLocalFileSystem && !it.isDirectory && it.path.endsWith("mirrord.json")
         }
     }
 
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
@@ -50,44 +50,24 @@ class MirrordConfigDropDown : ComboBoxAction() {
         }
     }
 
-    private fun initializeConfigs(project: Project) {
-        val basePath = project.basePath ?: throw Error("couldn't resolve project path")
-        configFiles = FileBasedIndex.getInstance().getAllKeys(MirrordConfigIndex.key, project).filter {
-            it.startsWith(basePath)
-        }.toHashSet()
-    }
-
     // mostly what is viewed by the user is done here, goal is to have minimal logic in the function
     // because it is called every half a second
     override fun update(e: AnActionEvent) {
         e.project?.let { project ->
             // this check ensures that we don't query the index when it is being built
             // querying the index during the startup can give us 0
-            if (DumbService.isDumb(project)) {
-                e.presentation.isVisible = false
-                return
-            }
-
             if (!::configFiles.isInitialized) {
                 // there is a possibility of a race condition in case index is accessed while being built
                 // so all index queries need to be done in smart mode
-                if (DumbService.isDumb(project)) {
-                    DumbService.getInstance(project).runReadActionInSmartMode {
-                        initializeConfigs(project)
-                    }
-                } else {
+                runInSmartModeOrDirectly({
                     initializeConfigs(project)
-                }
+                }, project)
             }
 
             if (updateConfigs.getAndSet(false)) {
-                if (DumbService.isDumb(project)) {
-                    DumbService.getInstance(project).runReadActionInSmartMode {
-                        updateConfigFiles(project)
-                    }
-                } else {
+                runInSmartModeOrDirectly({
                     updateConfigFiles(project)
-                }
+                }, project)
             }
 
             val files = configFiles
@@ -105,6 +85,21 @@ class MirrordConfigDropDown : ComboBoxAction() {
 
             e.presentation.text = getReadablePath(chosenFile!!, project)
             e.presentation.isVisible = true
+        }
+    }
+
+    private fun initializeConfigs(project: Project) {
+        val basePath = project.basePath ?: throw Error("couldn't resolve project path")
+        configFiles = FileBasedIndex.getInstance().getAllKeys(MirrordConfigIndex.key, project).filter {
+            it.startsWith(basePath)
+        }.toHashSet()
+    }
+
+    private fun runInSmartModeOrDirectly(action: () -> Unit, project: Project) {
+        if (DumbService.isDumb(project)) {
+            DumbService.getInstance(project).runReadActionInSmartMode(action)
+        } else {
+            action()
         }
     }
 
--- a/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
+++ b/intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
@@ -113,7 +113,7 @@ class MirrordConfigDropDown : ComboBoxAction() {
 
     // this function spawns a background task to query the index, not blocking the current thread
     // it maintains a single task to do so, using the `blockQueries` flag
-    private fun queryIndexInSmartMode(project: Project, query: (project: Project) -> Unit) {
+    private fun queryIndexInSmartMode(project: Project, query: (Project) -> Unit) {
         if (!blockQueries.getAndSet(true)) {
             object : Task.Backgroundable(project, "mirrord") {
                 override fun run(indicator: ProgressIndicator) {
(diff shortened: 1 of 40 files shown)


t4lz (commented):
I think this probably solve the lost update bug, but instead of missing an update - in the case of a second filesystem change while an earlier update is ongoing, now the thread that creates background tasks is blocked on a previous background task. As this case is probably rare and querying the file index probably does not take that long maybe this is good enough. But how about this mechanism:

1. Set a normal boolean `updateConfigs` to `true` when a filesystem change is detected.
2. In the periodically called `update` function: 
    ```kotlin
    if (updateConfigs && !queryRunning) {
        // assuming `update` cannot run multiple times in parallel.
        //  If it can, make queryRunning atomic, and use `getAndSet` in the condition.
        queryRunning = true
        // create background task
    }
    ```
 3. In the background task:
   i) before making the update, set `updateConfigs` to `false`.
   ii) after making the update, before exiting from the background task, set `queryRunning` to `false`.
 
 #### No filesystem change is lost:
- Every time there is a filesystem change, `updateConfigs` is set to `true`.
- On the next call to `update` after `updateConfigs` is set to `true`:
  - If `!queryRunning`, then a task will be created, because `updateConfigs && !queryRunning`.
    - The created task unconditionally processes the update, and so the update is not lost.
  - If `queryRunning`
    - if the task sets `updateConfigs` to `false` after the file watcher had set it to `true`, then it also does the query afterwards (because the query happens after setting to `false`) and so the update is not lost because the already running tasks processes it.
    - if `updateConfigs` was set to `true` only after the task has set it to `false`, then `updateConfigs` remains `true` until a new task starts (because it is only set to `false` inside tasks). Before the running task exits it sets `queryRunning` to `false`, and the first time `update` is called after this happens, the conditions will apply to starting a new task, in which the index will be queried, and the config files updated.

This covers all possible cases, and in none of them we remain with an old state for more then until the the first `update` is called after an already running query finishes.

 #### The file index is not queried in parallel:
- The query only happens in a background task.
- After the first task, no new task is created until the task before it sets `queryRunning` to `false`.
- This happens after querying the index has completed, so a new query can only start after the last query has completed.

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -65,28 +66,26 @@ class MirrordConfigDropDown : ComboBoxAction() {
...
-            if (chosenFile !in files) {
-                chosenFile = files.first()
+                e.presentation.text = chosenFile?.let { getReadablePath(it, project) }
+            } else {
+                chosenFile = configFiles.firstOrNull()
             }
-
-            e.presentation.text = getReadablePath(chosenFile!!, project)
-            e.presentation.isVisible = true
+            e.presentation.isVisible = configFiles.size > 1
```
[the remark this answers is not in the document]
infiniteregrets: @eyalb181 and I discussed this, and we chose to keep a single file as the default. However, we can open a discussion/issue about this, separate from this PR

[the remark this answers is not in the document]
eyalb181: I wouldn't worry about it, since clicking the settings button opens the file.

t4lz, line 88: No need to do all this if `blockAndQueryIndex` was not called, right? Why not have 1 if for both the not-initialized-yet case and the `updateConfig` case, and in it call `blockAndQueryIndex` and then do all of this?

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -114,39 +113,67 @@ class MirrordConfigDropDown : ComboBoxAction() {
...
+                shouldQuery = blockQueries
+            }
+        }
+        // shouldQuery gives us the locked value of blockQueries without
+        // causing a deadlock because also queryIndexInSmartMode "synchronizes" querying and setting
+        // the flag
+        if (shouldQuery) {
+            queryIndexInSmartMode(project) {
+                updateConfigFiles(project)
+            }
```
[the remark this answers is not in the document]
infiniteregrets: more like a style thing but this is correct: `queryIndexInSmartMode(project , ::updateConfigFiles)`

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
infiniteregrets: initialization needs to set `e.presentation.isVisible = false`

infiniteregrets:
> I think this probably solve the lost update bug, but instead of missing an update - in the case of a second filesystem change while an earlier update is ongoing, now the thread that creates background tasks is blocked on a previous background task. As this case is probably rare and querying the file index probably does not take that long maybe this is good enough. But how about this mechanism:
> 
> 1. Set a normal boolean `updateConfigs` to `true` when a filesystem change is detected.
> 2. In the periodically called `update` function:
>    ```kotlin
>    if (updateConfigs && !queryRunning) {
>        // assuming `update` cannot run multiple times in parallel.
>        //  If it can, make queryRunning atomic, and use `getAndSet` in the condition.
>        queryRunning = true
>        // create background task
>    }
>    ```
> 3. In the background task:
>    i) before making the update, set `updateConfigs` to `false`.
>    ii) after making the update, before exiting from the background task, set `queryRunning` to `false`.
> 
> #### No filesystem change is lost:
> * Every time there is a filesystem change, `updateConfigs` is set to `true`.
> * On the next call to `update` after `updateConfigs` is set to `true`:
>   
>   * If `!queryRunning`, then a task will be created, because `updateConfigs && !queryRunning`.
>     
>     * The created task unconditionally processes the update, and so the update is not lost.
>   * If `queryRunning`
>     
>     * if the task sets `updateConfigs` to `false` after the file watcher had set it to `true`, then it also does the query afterwards (because the query happens after setting to `false`) and so the update is not lost because the already running tasks processes it.
>     * if `updateConfigs` was set to `true` only after the task has set it to `false`, then `updateConfigs` remains `true` until a new task starts (because it is only set to `false` inside tasks). Before the running task exits it sets `queryRunning` to `false`, and the first time `update` is called after this happens, the conditions will apply to starting a new task, in which the index will be queried, and the config files updated.
> 
> This covers all possible cases, and in none of them we remain with an old state for more then until the the first `update` is called after an already running query finishes.
> 
> #### The file index is not queried in parallel:
> * The query only happens in a background task.
> * After the first task, no new task is created until the task before it sets `queryRunning` to `false`.
> * This happens after querying the index has completed, so a new query can only start after the last query has completed.

Okay, from what I have understood this seems to have the same problem we are trying to solve here. Let's just move it to a separate discussion
Screen Shot 2023-05-16 at 10 28 54 AM

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
t4lz: > initialization needs to set `e.presentation.isVisible = false`

why?

infiniteregrets: why show the config dropdown when there are no configs?

t4lz:
> Okay, from what I have understood this seems to have the same problem we are trying to solve here. Let's just move it to a separate discussion Screen Shot 2023-05-16 at 10 28 54 AM

The scenario in your diagram is valid, there is no problem there. The scenario does not contain a file system change that is not followed by a an update (querying the index and updating the hashmap) in a later point in time.

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
t4lz: How do you know there are no configs? The hash map is not initialized yet, but then you call the function that updates them, so now they may or may not be values in the HashMap, and you can just have the same logic as after an update - if there are (more then 1) configs - show the menu, otherwise - dont.

infiniteregrets: you get the configFiles not initialized exception, would need further synchronization

t4lz: Where do you get that exception?
After calling the function that updates them they are initialized, no?

infiniteregrets: if you dont return and try to access the configFiles

infiniteregrets: remember, the update is a task we don't know if the hashmap has been initialized

infiniteregrets:
> > Okay, from what I have understood this seems to have the same problem we are trying to solve here. Let's just move it to a separate discussion Screen Shot 2023-05-16 at 10 28 54 AM
> 
> The scenario in your diagram is valid, there is no problem there. The scenario does not contain a file system change that is not followed by a an update (querying the index and updating the hashmap) in a later point in time.

okay, that's right. I am still going over all possible scenarios will get back to you

infiniteregrets:
> > Okay, from what I have understood this seems to have the same problem we are trying to solve here. Let's just move it to a separate discussion Screen Shot 2023-05-16 at 10 28 54 AM
> 
> The scenario in your diagram is valid, there is no problem there. The scenario does not contain a file system change that is not followed by an update (querying the index and updating the hashmap) in a later point in time.

Alrighty per my testing, all the instructions based on your suggestion will be laid out similarly to the solution I proposed and I believe there is no need for atomics/locks anymore so great suggestion on that, thanks!

infiniteregrets:
@talz before you review again, I can see one downside of the new approach. Consider the following case:

here updateConfigs is set to true before the query and is not set to false, we spawn an extra task here. However, the cost of locks is more than having an extra update as it is amortized for our case, so we stick to the new approach

Screen Shot 2023-05-16 at 3 13 04 PM

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -125,42 +120,53 @@ class MirrordConfigDropDown : ComboBoxAction() {
...
-                                indicator.text = "mirrord: updating config files"
-                                query(project)
-                                blockQueries.set(false)
-                            } catch (e: IndexNotReadyException) {
-                                return@compute false
-                            }
-                            true
+        object : Task.Backgroundable(project, "mirrord") {
+            override fun run(indicator: ProgressIndicator) {
+                updateConfigs = false
```
t4lz, line 125: nitpick: I think it would be good to unset this flag as close as possible to the querying of the index, but up to you.

t4lz (approved):
Looks good.

It might be that the two flags have to be `volatile`, please check that before merging, and make them volatile if necessary.

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordConfigDropDown.kt
```diff
@@ -120,9 +122,9 @@ class MirrordConfigDropDown : ComboBoxAction() {
     // this function spawns a background task to query the index, not blocking the current thread
     // it maintains a single task to do so, using the `blockQueries` flag
     private fun queryIndexInSmartMode(project: Project, query: (Project) -> Unit) {
+        updateConfigs = false
         object : Task.Backgroundable(project, "mirrord") {
             override fun run(indicator: ProgressIndicator) {
-                updateConfigs = false
```
t4lz: Hey @infiniteregrets, why did you move the unsetting of `updateConfigs` out of the background task?

infiniteregrets: sorry, sounds like a mistake on my end, i pushed the wrong diff. ill fix this

PR #1424, opened 2023-05-17
don't error on operator check fail

aviramha:


DmitryDodzin (approved):
Just the lint err

mirrord/cli/src/connection.rs
```diff
@@ -48,9 +48,9 @@ where
             None
         }
         Err(err) => {
-            sub_progress.fail_with("Unable to connect to Operator");
+            sub_progress.done_with("Unable to check if operator exists, probably due to RBAC");
 
-            warn!("{err}");
+            trace!("{err}");
```
Razz4780, line 53: Shouldn't this still be `warn`? Most users probably won't see it if it's `trace`

Razz4780 (approved):

mirrord/cli/src/connection.rs
aviramha: It can happen to any user just trying use the open source so prefer not to show it.

PR #1432, opened 2023-05-19
Fix broken target config.

t4lz:
In the unreleased targetless PR I broke the target config because I misunderstood when the `Simple` and `Advanced` variants are generated. This PR fixes that.


changelog.d/574.added.md
```diff
@@ -1 +1 @@
-Support for targetless execution: when not specifying any target for the agent, mirrord now spins up an independent agent. This can be useful e.g. if you are just interested in getting the cluster's dns resolution and outgoing connectivity but don't want any pod's incoming traffic or FS. 
+Support for targetless execution: when not specifying any target for the agent, mirrord now spins up an independent agent. This can be useful e.g. if you are just interested in getting the cluster's DNS resolution and outgoing connectivity but don't want any pod's incoming traffic or FS.
```
meowjesty, line 1: Wrong changelog?

meowjesty (changes requested):
Think this is the feature added changelog?

changelog.d/574.added.md
t4lz: As this was broken and fixed without being released in between, it should not be in the changelog. However I think towncrier requires a changelog change, so I just made a change to an existing entry.

meowjesty: Oh, ok.

meowjesty (approved):
:+1:

PR #1443, opened 2023-05-23
Change mirrord ls to show only pods that are in running state 

### LINKED ISSUE #1436: Show only targets in "Running" phase
If you have pods in other state than running, `mirrord ls` shows those as valid targets. We should ignore those targets.

aviramha:
Closes #1436 
(not crashing,starting,etc)


infiniteregrets (approved):
🏖️

PR #1444, opened 2023-05-23
Add E2E tests for vscode extension

### LINKED ISSUE #201: Add sanity tests for the VS Code extension
We don't test the VS Code extension at the moment, so changes there are risky. Please add a suite of sanity tests and have them run in the CI.

infiniteregrets:
closes #201
This PR makes the following changes:
- removes unwanted dependencies
- lint job for typescript
- e2e for vscode


vscode-ext/src/extension.ts
```diff
@@ -165,13 +153,11 @@ class MirrordAPI {
...
 
-		const [stdout, stderr] = await this.exec(args);
+		let [stdout, stderr] = await this.exec(args);
 
 		if (stderr) {
-			const match = stderr.match(/Error: (.*)/)?.[1];
-			const notificationError = match ? JSON.parse(match)["message"] : stderr;
 			mirrordFailure(stderr);
-			throw new Error(`mirrord failed to 'ls' targets\n${notificationError}`);
+			throw new Error("error occured listing targets: " + stderr);
```
eyalb181, line 160:
```suggestion
			throw new Error("error occurred listing targets: " + stderr);
```

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    before(async function () {
+        console.log("podToSelect: " + podToSelect);
+        console.log("kubeService: " + kubeService);
+        expect(podToSelect).to.not.be.undefined;
+        expect(kubeService).to.not.be.undefined;
+        browser = VSBrowser.instance;
+        // need to bring the flask app in open editors
+        await browser.openResources(testWorkspace, join(testWorkspace, fileName));
+        await sleep(10000);
```
eyalb181, line 42: Why all these sleeps? Seems very prone to breakage, doesn't the testing library provide a way to poll for changes?

eyalb181 (changes requested):

vscode-ext/src/tests/e2e.ts
infiniteregrets: I think the pattern I have already used in the code `this.browser.wait` should eliminate a lot of sleeps

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (await debugToolbar?.isDisplayed()) {
+            await debugToolbar.stop();
+        }
+    });
+
+    it("enable mirrord", async function () {
+        const statusBar = new StatusBar();
+        await browser.driver.wait(async () => {
+            const enableButton = await statusBar.getItem("Enable mirrord");
+            if (enableButton !== undefined) {
```
eyalb181, line 56: In which cases would this be undefined?

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+                return true;
+            }
+        }, DEFAULT_CONDITION_TIMEOUT, "mirrord `enable` button not found -- timed out");
+
+        await browser.driver.wait(async () => {
+            const enableButton = await statusBar.getItem("Disable mirrord");
+            if (enableButton !== undefined) {
+                return true;
+            }
+        }, DEFAULT_CONDITION_TIMEOUT, "Mirrord not enabled -- timed out");
```
eyalb181, line 67:
```suggestion
        }, DEFAULT_CONDITION_TIMEOUT, "mirrord `disable` button not found -- timed out");
```

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+// We wait for 10 seconds to make sure that we are in listening state
+async function sendTrafficToPod() {
+    await new Promise(resolve => setTimeout(resolve, 10000));
+    const response = await get(kubeService!!);
+    expect(response.status).to.equal(200);
+    expect(response.data).to.equal("OK - GET: Request completed\n");
+}
+
+// opens and sets a breakpoint in the given file
+async function setBreakPoint(fileName: string, breakPoint: number, browser: VSBrowser, timeout: number = 10000) {
```
eyalb181, line 121: Why doesn't this just use the default timeout?

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+    }, timeout, "editor tab title not found -- timed out");
+
+    const textEditor = new TextEditor();
+    const result = await textEditor.toggleBreakpoint(breakPoint);
+    expect(result).to.be.true;
+}
+
+// starts debugging the current file with the provided configuration
+// debugging starts from the "Run and Debug" button in the activity bar
+async function startDebugging(configurationFile: string) {
```
eyalb181, line 140:
```suggestion
async function startDebugging(configurationFile: string = "Python: Current File") {
```

And remove the const

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+        const inputBox = await InputBox.create();
+        // assertion that podToSelect is not undefined is done in "before" block   
+        await browser.driver.wait(async () => {
+            return await inputBox.isDisplayed();
+        }, DEFAULT_CONDITION_TIMEOUT, "quickPick not found -- timed out");
+        await inputBox.selectQuickPick(podToSelect!);
+    });
+
+    it("wait for breakpoint to be hit", async function () {
+        debugToolbar = await DebugToolbar.create(20000);
```
eyalb181, line 99: Is this the timeout? Why is it different from the default?

vscode-ext/src/tests/e2e.ts
```diff
@@ -0,0 +1,146 @@
...
+        }, 20000, "debug toolbar not found -- timed out");
+
+        await Promise.all([debugToolbar.waitForBreakPoint(), sendTrafficToPod()]);
+    });
+});
+
+// This promise is run in parallel to the promise waiting for the breakpoint to be hit
+// We wait for 10 seconds to make sure that we are in listening state
+async function sendTrafficToPod() {
+    await new Promise(resolve => setTimeout(resolve, 10000));
```
eyalb181, line 114: How is this different from a sleep?

eyalb181 (changes requested):
Looks good overall, made some comments

vscode-ext/src/tests/e2e.ts
infiniteregrets: its undefined when it has not loaded

vscode-ext/src/tests/e2e.ts
infiniteregrets: needs more than 10 seconds

vscode-ext/src/tests/e2e.ts
infiniteregrets: I cant find a way to make sure that we are listening on the breakpoint, so thought we can sleep for 10 secs and then send the request, wdyt?

vscode-ext/src/tests/e2e.ts
eyalb181: Is that feasible? If so, shouldn't there be an else clause? If not, shouldn't it just be enableButton?.click();?

eyalb181 (changes requested):

vscode-ext/src/tests/e2e.ts
infiniteregrets: it automatically means the condition is not true if we don't return true

infiniteregrets: wait requires a boolean

eyalb181 (approved):

PR #1447, opened 2023-05-23
Change fs mode to be local with overrides when targetless is used

aviramha:


mirrord/layer/src/lib.rs
```diff
@@ -438,7 +438,13 @@ fn layer_start(config: LayerConfig) {
...
         .expect("Setting INCOMING_IGNORE_PORTS failed");
 
-    FILE_FILTER.get_or_init(|| FileFilter::new(config.feature.fs.clone()));
+    FILE_FILTER.get_or_init(|| {
+        let mut fs_config = config.feature.fs.clone();
+        if config.target.is_none() {
+            fs_config.mode = FsModeConfig::LocalWithOverrides;
+        }
+        FileFilter::new(fs_config)
+    });
```
t4lz, line 447: Does this override what the user has in the configuration file?

aviramha: in terms of mode, yes. but I can't see why it would want anything but `LocalWithOVerrides` since the fs access would be to the agent file system which has almost nothing..

t4lz: What about `local` mode?

t4lz: What if a user's app creates files on paths that don't exist on their machine, so they are running in `Write` mode? Is this far fetched? 😅

aviramha: What's the user case though?

t4lz: I don't know... Maybe even just not wanting a local version of a file to be overridden by an application, or an application creating huge logs and the user not wanting the application to create huge files on their local machine.
I'm asking "How can we be sure users don't need any other mode?" rather than "Why would users use a different mode?" because I think we can't think in advance about all use cases, so if it's valid - we should try to support it, and if we don't support something, at the best case scenario we should be able to justify why it's never valid.

But if you think overriding the mode will improve UX for most users most of the time, that's a valid argument, and I don't want to block the merge 😄

Issue #69, opened 2022-05-22
sockets fd isn't removed on close


aviramha:
### Bug Description

We currently don't hook close, meaning that if a program runs close on a socket we manage we don't delete it from our managed fds.

### Steps to Reproduce

N/A

### Your operating system and version

N/A

### Your Node version (`node --version`)

N/A

aviramha:
closed via #72

PR #1449, opened 2023-05-23
Tool to extract documentation into md.

meowjesty:
- Issue: [#69](https://github.com/metalbear-co/mirrord.dev/issues/69)

1. Introduce tool (`medschool`) to auto-generate documentation;
2. Update mirrord/config docs to match what's in mirrord.dev;
3. Update mirrord-schema with these new docs;

P.S. as the docs are just a transfer from what we have in mirrord.dev, it might be easier to review by looking at a diff of the auto-generated docs and the previously available docs there.


eyalb181 (approved):

medschool/src/main.rs
```diff
@@ -0,0 +1,550 @@
...
+                        }
+                    }
+                    _ => None,
+                })
+        })
+        .collect::<BTreeSet<_>>();
+
+    // The types that have been rebuild after following each field "reference".
+    let mut new_types = Vec::with_capacity(8);
+
```
infiniteregrets, line 396: why is the capacity hardcoded to 8

medschool/src/main.rs
```diff
@@ -0,0 +1,550 @@
...
+                            })
+                        } else {
+                            None
+                        }
+                    }
+                    _ => None,
+                })
+        })
+        .collect::<BTreeSet<_>>();
+
```
infiniteregrets, line 393: can you please document the reason why you use btreeset/btreemap over hashmap

medschool/src/main.rs
meowjesty: Oh snap, some of these were just random values.

medschool/src/main.rs
meowjesty: Yup, can you spot more places where docs are lacking?

Will open a PR with whatever needs tidying up.

infiniteregrets: sure will let you know

meowjesty: Added a brief doc for this in the newest PR.

PR #1462, opened 2023-05-25
fixed IntelliJ failing silently when error happened on listing pods

aviramha:


intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordApi.kt
```diff
@@ -104,7 +104,6 @@ object MirrordApi {
                 val error = gson.fromJson(trimmedError, Error::class.java)
                 MirrordNotifier.errorNotification(error.message, project)
                 MirrordNotifier.notify(error.help, NotificationType.INFORMATION, project)
-                return null
```
Razz4780, line 107: The only difference I observe is that the IDE shows an additional error notification (`IDE error occurred`). When `ls` fails we already show one error, one warn and one info notification :sweat_smile: 
Also, imo `IDE error occurred` doesn't look good. Looks like we had a bug or something

t4lz (approved):

intellij-ext/modules/core/src/main/kotlin/com/metalbear/mirrord/MirrordApi.kt
aviramha: wdyt now?

Issue #1461, opened 2023-05-25
namespace setting broken since targetless


aviramha:
```
2023-05-25 13:43:41,182 [  63378]   FINE - mirrord - mirrord ls failed: Error:   × Configuration file parsing failed - TargetNamespaceWithoutTarget
  ╰─▶ A target namespace was specified, but no target was specified. If you
      want to set the
              namespace in which the agent will be created, please set the
      agent namespace, not the
              target namespace. That value can be set with agent.namespace in
      the configuration file,
              the -a argument of the CLI, or the MIRRORD_AGENT_NAMESPACE
      environment variable.
      
              If you are not trying to run targetless, please specify a target
      instead.
  help: Configuration file parsing failed. Inspect your config file or
        arguments provided.
        
        - If you're still stuck:
        
        >> Please open a new bug report at https://github.com/metalbear-co/
        mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help
        in #mirrord-help
        
        >> Or email us at hi@metalbear.co
```

mirrord.json:
```json

{
    "operator": false,
    "accept_invalid_certificates": false,
    "target": {
        "namespace": "kube-system"
    },
    "feature": {
        "network": {
            "incoming": "steal",
            "outgoing": true
        },
        "fs": "read",
        "env": true
    }
}
```

aviramha:
Fixed in #1467 and released as 3.44.1

PR #1467, opened 2023-05-25
Make the path field of the TargetConfig optional.

t4lz:
`mirrord ls` can run with just a target namespace, without a target path, so since it shares a configuration with `mirrord exec`, the path has to be optional.  #1461.


meowjesty (approved):
:bear: :+1:

Issue #1353, opened 2023-04-26
Next-protocol tracking


aviramha:
This issue is for tracking the "next" protocol for mirrord-agent/operator/layer/proxy.

We want to make the transition as easy as possible with least breaking between versions.

The list is WIP and follow up on https://github.com/metalbear-co/operator/issues/110

- [x] #1354
- [ ] #1356
- [ ] Add IDs to messages to match request/response
- [ ] Parallel handling (blocked on IDs)
- [x] #1355
- [x] Add CI for protocol breaking warning to bump versions when needed (backward/forward compatible)

PR #1482, opened 2023-05-29
Fix nodejs failing randomly on outgoing requests

### LINKED ISSUE #1470: Node HTTP outgoing request fails randomly
I believe this is a fd leak / leak issue since it happens randomly. We also had an e2e test of many outgoing http requests that was flaky so it might be relevant.

Happens on Linux container on Mac M1 with node.

infiniteregrets:
closes #1470 

It was found that when remote closes, we terminated our loop resulting in ECONNRESET. we fix this by shutting down the local socket as well


aviramha (approved):
Great!

mirrord/layer/src/outgoing/tcp.rs
```diff
@@ -381,22 +381,28 @@ impl TcpOutgoingHandler {
...
+                            .ok_or(LayerError::NoConnectionId(connection_id))?;
+
+                        sender.send(bytes).await.unwrap_or_else(|_| {
+                            warn!(
                         "Got new data from agent after application closed socket. connection_id: \
                     connection_id: {connection_id}"
                     );
-                });
+                        });
+                    }
```
aviramha, line 401: Add a comment describing the best solution:
We should either:
1. add `DaemonTcpOutgoing::Error(errno)` and then pass it to the `interceptor_task` to break.
2. change `DaemonTcpOutgoing::Close` to `Close(Option<errno>)` and pass it to the `interceptr_task` to know when to shutdown and when to break loop (sounds uglier though)

aviramha: and that we don't do that yet because that'd require breaking the protocol, so we're waiting for #1353

mirrord/layer/tests/outgoing.rs
```diff
@@ -100,6 +100,13 @@ async fn outgoing_tcp(dylib_path: &PathBuf) {
         .collect::<Vec<_>>();
 
     for peer in peers {
+        #[cfg(target_os = "linux")]
```
aviramha, line 103: that's not the solution.. we need to figure out why this happens

aviramha (changes requested):

mirrord/layer/tests/outgoing.rs
infiniteregrets: yeah just looking into it

Issue #1453, opened 2023-05-24
e2e flakiness on passthrough test


aviramha:
https://github.com/metalbear-co/mirrord/actions/runs/5065827059/jobs/9095418740

```
failures:
    traffic::steal::steal::complete_passthrough::agent_1_Agent__Job::application_1_Application__PythonFastApiHTTP::tcp_data_1___THISISNOTHTTP_n__
    traffic::steal::steal::complete_passthrough::agent_1_Agent__Job::application_1_Application__PythonFastApiHTTP::tcp_data_2___short_n__
```

aviramha:
Still happens
```
thread 'traffic::steal::steal::complete_passthrough::agent_1_Agent
__Job::application_1_Application__PythonFastApiHTTP::tcp_data_2___short_n__' panicked at 'Timeout 60s expired', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rstest-0.17.0/src/timeout.rs:42:21
```
https://github.com/metalbear-co/mirrord/actions/runs/5117873780

attached logs
[logs_8244.zip](https://github.com/metalbear-co/mirrord/files/11596638/logs_8244.zip)

aviramha:
Pretty sure doesn't happen anymore.

PR #1483, opened 2023-05-29
Unflaking e2e tests

Razz4780:
Solves #1453

I looked into the tests and the code around, couldn't find any exact reason for the tests to behave badly. Also, tests are failing only if too many of them are being run at the same time. I assume it is simply too much load for minikube. Even when the runtime is given 32 cores, running all 71 tests at once causes some to fail.

With concurrency now limited to 5 tests at the same time, the suite passes without problems.


aviramha (approved):
Good job - I really hope this will solve the flakes!

PR #1485, opened 2023-05-30
Fix possible race with `IndexAllocator`

aviramha:
Provide buffer for `IndexAllocator` to avoid re-use of indices too fast

### DIFF
--- a/mirrord/agent/src/util.rs
+++ b/mirrord/agent/src/util.rs
@@ -90,6 +90,8 @@ where
 /// It keeps tracked of the indices that were freed, and reuses them when possible.
 /// You can provide a buffer of indices using the second generic `N` to avoid re-using indices
 /// too fast.
+/// TODO: Consider just having a simple IndexAllocator that keeps track of last index and increases by 1 as
+/// it'd be hard to exhaust u64. (which is usually T)
 #[derive(Debug)]
 pub struct IndexAllocator<T, const N: usize>
 where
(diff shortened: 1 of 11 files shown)


mirrord/agent/src/util.rs
```diff
@@ -86,22 +86,26 @@ where
     }
 }
 
+/// `IndexAllocator` is a helper struct that allocates unique indices for a given type.
+/// It keeps tracked of the indices that were freed, and reuses them when possible.
+/// You can provide a buffer of indices using the second generic `N` to avoid re-using indices
```
meowjesty, line 91: The solution is a bit weird, and if feels like it's kinda just pushing the potential problem forward. As I think most users don't reach the `100`  indexes that are being used.

What do we need to fully solve this? Have the allocator be aware of multiple threads? Hold the allocator with a `Mutex`? Or maybe we drop the idea of freeing and just keep increasing the index?

I'm ok with approving this if we add a `TODO` with possible better solutions.

meowjesty (changes requested):
A bit weird, so I suggest adding a `TODO`.

mirrord/agent/src/util.rs
aviramha: The problem isn't multiple threads but case when layer doesn't know of re-use by agent then weird stuff can happen.

meowjesty: Oh, ok.

PR #1486, opened 2023-05-30
Layer dns errors

### LINKED ISSUE #1481: Translate dns errors between macOS and Linux
The definition of `EAI_*` errors is different between Linux and macOS, leading to the following error happening when resolving fails:
```
Assertion failed: (!"unknown EAI_* error code"), function uv__getaddrinfo_translate_error, file ../deps/uv/src/unix/getaddrinfo.c, line 90.
```

This is Linux:

```
pub const EAI_BADFLAGS: ::c_int = -1;
pub const EAI_NONAME: ::c_int = -2;
pub const EAI_AGAIN: ::c_int = -3;
pub const EAI_FAIL: ::c_int = -4;
pub const EAI_NODATA: ::c_int = -5;
pub const EAI_FAMILY: ::c_int = -6;
pub const EAI_SOCKTYPE: ::c_int = -7;
pub const EAI_SERVICE: ::c_int = -8;
pub const EAI_MEMORY: ::c_int = -10;
pub const EAI_SYSTEM: ::c_int = -11;
pub const EAI_OVERFLOW: ::c_int = -12;
```

macOS:

```
pub const EAI_AGAIN: ::c_int = 2;
pub const EAI_BADFLAGS: ::c_int = 3;
pub const EAI_FAIL: ::c_int = 4;
pub const EAI_FAMILY: ::c_int = 5;
pub const EAI_MEMORY: ::c_int = 6;
pub const EAI_NODATA: ::c_int = 7;
pub const EAI_NONAME: ::c_int = 8;
pub const EAI_SERVICE: ::c_int = 9;
pub const EAI_SOCKTYPE: ::c_int = 10;
pub const EAI_SYSTEM: ::c_int = 11;
pub const EAI_OVERFLOW: ::c_int = 14;
```

I suggest to have an enum that maps the errors to our internal type, which can be converted from/to Linux and to macOS (so agent converts the Linux error, sends it over the wire, then layer converts it to Linux/macOS based on need)

infiniteregrets:
closes #1481

### DIFF
--- a/mirrord/layer/src/error.rs
+++ b/mirrord/layer/src/error.rs
@@ -265,6 +265,32 @@ impl From<HookError> for i64 {
                 },
                 ResponseError::DnsLookup(dns_fail) => match dns_fail.kind {
                     mirrord_protocol::ResolveErrorKindInternal::Timeout => libc::EAI_AGAIN,
+                    mirrord_protocol::ResolveErrorKindInternal::NoConnections => libc::EAI_AGAIN,
+                    mirrord_protocol::ResolveErrorKindInternal::Message(msg) => {
+                        error!("DNS lookup failed with message: {}", msg);
+                        libc::EAI_FAIL
+                    }
+                    mirrord_protocol::ResolveErrorKindInternal::Io(err) => {
+                        #[cfg(target_os = "macos")]
+                        {
+                             match err {
+                                -1 => libc::EAI_BADFLAGS,
+                                -2 => libc::EAI_NONAME,
+                                -3 => libc::EAI_AGAIN,
+                                -4 => libc::EAI_FAIL,
+                                -5 => libc::EAI_NODATA,
+                                -6  => libc::EAI_FAMILY,
+                                -7 => libc::EAI_SOCKTYPE,
+                                -8 => libc::EAI_SERVICE,                                
+                                -10 => libc::EAI_MEMORY,
+                                -11 => libc::EAI_SYSTEM,
+                                -12 => libc::EAI_OVERFLOW,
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/error.rs
```diff
@@ -264,7 +265,32 @@ impl From<HookError> for i64 {
                     _ => libc::EINVAL,
                 },
                 ResponseError::DnsLookup(dns_fail) => match dns_fail.kind {
-                    mirrord_protocol::ResolveErrorKindInternal::Timeout => libc::EAI_AGAIN,
+                    ResolveErrorKindInternal::Timeout => libc::EAI_AGAIN,
+                    ResolveErrorKindInternal::NoConnections => libc::EAI_AGAIN,
+                    ResolveErrorKindInternal::Message(msg) => {
+                        error!("DNS lookup failed with message: {}", msg);
```
aviramha, line 271: Setting libc errno isn't the right way to do it. `getaddrinfo` returns the `EAI_*` error as return value directly, not via `errno`.

aviramha: Also, thinking about it - we probably can be lossy until we change the protocol and add it as todo for when we break the protocol.

infiniteregrets: i assumed we would need to set the errno properly as well since we already do:
```
ResponseError::DnsLookup(dns_fail) => match dns_fail.kind {
                    mirrord_protocol::ResolveErrorKindInternal::Timeout => libc::EAI_AGAIN,
                    _ => libc::EAI_FAIL,
                },
```

PR #1490, opened 2023-05-31
Fix outgoing tcp not working when server ends connection fast

aviramha:


mirrord/layer/src/outgoing/tcp.rs
```diff
@@ -375,23 +382,45 @@ impl TcpOutgoingHandler {
...
+                        //   loop
+                        // we don't know who failed so we'll iterate the list then remove closed
+                        // sockets.
+                        debug!("Read error: {:?}", e);
+                        warn!("Read error: {:?}", e);
+                        self.mirrors = self
+                            .mirrors
+                            .drain()
+                            .filter(|(_k, v)| !v.is_closed())
+                            .collect();
```
Razz4780, line 421: I'd use `retain` here

PR #1491, opened 2023-06-01
Reduce Ping interval from 60s down to 30s

DmitryDodzin:


aviramha (approved):

changelog.d/+ping-interval.internal.md
```diff
@@ -0,0 +1 @@
+Change Ping interval from 60s down to 30s.
```
aviramha, line 1: Change it to be `changed` instead of `internal` please

PR #1496, opened 2023-06-03
Integration test for dns resolution

infiniteregrets:
integration test for #1481


changelog.d/+integration-dns.internal.md
```diff
@@ -0,0 +1 @@
+Add integration test for DNS resolution
```
aviramha, line 1: This doesn't solve what we wanted to have this for - checking `getaddrinfo` return value. I suggest implementing the test in rust, and asserting in the rust app that domain `X` would return result and domain `Y` would return the appropriate error. 
Having said that, I do see a value in the case shown here (where it's resolved over UDP using `socket/connect/send/recv`) but you'd need to add assertions to the app and also I believe that because the dns queries are being generated "on the fly", each would have different query id, so a constant response wouldn't work.

infiniteregrets: sure, will do, makes sense. was trying to make something basic work in node and then go ahead with implementing other scenarios, ill remove the node test for now and move to rust

mirrord/layer/tests/apps/dns_resolve/src/main.rs
```diff
@@ -0,0 +1,47 @@
+use std::{ffi::CString, ptr, str};
+
+fn get_ip_addresses(node: &str) -> (Vec<String>, i32) {
+    let node = CString::new(node).unwrap();
+    let service = CString::new("http").unwrap();
```
aviramha, line 5: can just be null, not sure what this adds to the test

mirrord/layer/tests/apps/dns_resolve/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let mut hints: libc::addrinfo = unsafe { std::mem::zeroed() };
+    hints.ai_family = libc::AF_UNSPEC;
```
aviramha, line 8: not sure why you do it, passing empty hints is ok?

aviramha (changes requested):

mirrord/layer/tests/apps/dns_resolve/src/main.rs
infiniteregrets: I am not sure if converting to a readable form like so is needed:
 ```rs
let ipv4_addr = std::net::Ipv4Addr::from(ip);
let addr_str = ipv4_addr.to_string();

addresses.push(addr_str);
```
I wanted to make sure that the addresses we get are not empty, so maybe I should just push `(*sockaddr_in).sin_addr.s_addr` without converting it to a readable form? 

Are we concerned about asserting the response is not empty or just the return code?

File: mirrord/layer/tests/apps/dns_resolve/src/main.rs [hunk 4]
```diff
@@ -0,0 +1,45 @@
+use std::{ffi::CString, ptr, str};
+
+fn get_ip_addresses(node: &str) -> (Vec<String>, i32) {
+    let node = CString::new(node).unwrap();
+    let service = ptr::null_mut::<libc::c_char>();
+
+    let hints: libc::addrinfo = unsafe { std::mem::zeroed() };
+
+    let mut res: *mut libc::addrinfo = ptr::null_mut();
```
aviramha, line 9: btw you can probably lookup use `.to_socket_addrs` instead of doing libc stuff (I checked, they use `getaddrinfo` underneath..)

aviramha (changes requested):

mirrord/layer/tests/apps/dns_resolve/src/main.rs
infiniteregrets: oh ): yes I saw on stackoverflow when i was doing recv_from, i should use that

infiniteregrets: i keep getting `thread 'main' panicked at 'Error: invalid socket address', mirrord/layer/tests/apps/dns_resolve/src/main.rs:7:19
stack backtrace:`

rust app:

```rs
use std::net::{ToSocketAddrs, SocketAddr};

fn get_ip_addresses(node: &str) -> (Vec<String>, i32) {
    print!("Looking up IP addresses for {}... ", node   );
    let addresses: Vec<SocketAddr> = match node.to_socket_addrs() {
        Ok(addrs) => addrs.collect(),
        Err(e) => panic!("Error: {}", e),
    };

    let addresses_str = addresses
        .iter()
        .map(|addr| addr.ip().to_string())
        .collect::<Vec<String>>();

    (addresses_str, 0)
}

fn test_getaddrinfo() {
    let domain_x = "example.com";
    let domain_y = "nonexistent-domain";

    let (addresses_x, result_x) = get_ip_addresses(domain_x);
    assert!(!addresses_x.is_empty(), "Expected addresses for domain X");
    assert_eq!(result_x, 0, "Expected success for domain X");

    let (addresses_y, result_y) = get_ip_addresses(domain_y);
    assert!(addresses_y.is_empty(), "Expected no addresses for domain Y");
    assert_eq!(result_y, libc::EAI_FAIL, "Expected failure for domain Y");
}

fn main() {
    test_getaddrinfo();
}
```

the discussion i mentioned:
https://stackoverflow.com/questions/68042772/how-to-do-a-dns-lookup-without-any-crates

> That will just resolve an IP address. But DNS is so much more powerful - there are TXT,MX,PTR, CNAME,etc, all kinds of different record types which you cannot access with this approach

i am not 100% sure about how DNS queries work at a lower level, but i am assuming its not enough and we still probably need getaddrinfo?

aviramha: We don't care about it. Also, getaddrinfo is A records afaik..

aviramha (approved):
Great!

Issue #1287, opened 2023-04-12
Reconsidering the way background tasks are handled


Razz4780:
Right now mirrord agent uses 3 global background tasks (dns, sniffer and stealer) and 2 per-client background tasks (tcp and udp). All tasks are run in their own tokio runtimes constructed with `Builder::new_multi_thread`, which spawns `nproc` threads. Main function is also run in a multi-threaded runtime. That means that agent uses `(4 + clients) * nproc` threads. Maybe we could consider using single-threaded runtimes?

eyalb181:
Seems premature at this point - not something our users will really feel the effect of

Razz4780:
> Seems premature at this point - not something our users will really feel the effect of?

Just something I noticed when working with the agent. The change would be very simple - using [different](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.new_current_thread) tokio runtime flavor.

Razz4780:
Closed with #1497

PR #1497, opened 2023-06-04
Use one thread for namespaced runtimes

aviramha:
Not sure if it matters a lot, but if tests pass might worth merging it.


eyalb181 (approved):

infiniteregrets:
quick question, what is the performance benefit here? any benchmarks for using the current thread vs multithreaded?

aviramha:
> quick question, what is the performance benefit here? any benchmarks for using the current thread vs multithreaded?

No benchmark, threads have overhead and should be avoided in async if not needed..

changelog.d/+internal-single-thread-agent.internal.md
t4lz, line 1: This can link to #1287.

aviramha: Thanks!

aviramha:
Closes #1287

PR #1505, opened 2023-06-05
Detecting Rider debugger port

Razz4780:
Debugging in Rider has the same problem as debugging Python in VS Code/IntelliJ. Layer now detects the debugger's port in the command line args.


aviramha (approved):
Nice catch

PR #1506, opened 2023-06-05
Fix outgoing sockets crash/error

aviramha:


mirrord/layer/src/outgoing/tcp.rs
```diff
@@ -376,22 +376,30 @@ impl TcpOutgoingHandler {
...
+                        sender.send(bytes).await.unwrap_or_else(|_| {
+                            warn!(
+                                "Got new data from agent after application closed socket. \
+                            connection_id: {connection_id}"
+                            );
+                        });
+                    }
+                    // We need connection id with these errors to actually do something
+                    // pending protocol change
+                    Err(e) => {
```
DmitryDodzin, line 398: Maybe trace debug!/warn! error just to see that it happened

DmitryDodzin (approved):
Nitpick

PR #1508, opened 2023-06-05
Integration test for recv_from

infiniteregrets:


mirrord/layer/tests/apps/recv_from/src/main.rs
```diff
@@ -0,0 +1,33 @@
+use std::{mem::MaybeUninit, net::ToSocketAddrs};
+
+use socket2::{Domain, SockAddr, Socket, Type};
+
+fn main() {
+    let google_dns_server = "8.8.8.8:53";
```
aviramha, line 6: I think you can just use `parse`

mirrord/layer/tests/apps/recv_from/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+    socket
+        .connect(&google_dns_server_addr)
+        .expect("Failed to connect to socket");
+
+    let query = "example.com. IN A\n";
+
+    socket.send(query.as_bytes()).expect("Failed to send query");
+
+    let mut response = [MaybeUninit::<u8>::uninit(); 1024];
+    let (_, source_address) = socket
```
aviramha, line 25: I would make the agent return data instead of an empty message to assert that the data is valid.

mirrord/layer/tests/apps/recv_from/src/main.rs
```diff
@@ -0,0 +1,33 @@
+use std::{mem::MaybeUninit, net::ToSocketAddrs};
+
+use socket2::{Domain, SockAddr, Socket, Type};
+
+fn main() {
+    let google_dns_server = "8.8.8.8:53";
+    let google_dns_server_addr: SockAddr = google_dns_server
```
aviramha, line 7: I'm not sure why the test is "DNS" related - for all we care this could be any ip/port so I think it might be confusing for other people, thinking it covers DNS-related stuff while it's just UDP flow.

aviramha (changes requested):

mirrord/layer/tests/apps/recv_from/src/main.rs
infiniteregrets: i understand it can be confusing, but the goal on my end was to describe a simple scenario of `connect` -> `send data` -> `recv_from` which can be accomplished with DNS query, perhaps i can document it for more clarity, but do you have a simpler flow in mind?

aviramha: do the same, but don't use dns names/ports? in any case the flow is integration test so it doesn't have to be protocol-valid.
i.e
````rs
s.connect("1.3.4.7:9111".parse())
s.send("bbbb".to_bytes())
let (_, addr) = s.recv(&mut buffer)
assert_eq!(buffer, b"aaaa")
assert_eq(addr, "1.3.4.7:9111".parse())
```

infiniteregrets: > doesn't have to be protocol-valid

i was assuming it should also run locally, thanks for the clarification

aviramha: not sure why it wouldn't run locally, or you mean outside of the integration test?

infiniteregrets: Yes outside of the test, as a standalone app

aviramha: not needed

mirrord/layer/tests/apps/recv_from/src/main.rs
infiniteregrets: there is no FromStr impl for SockAddr, cant use parse ):

aviramha:
```
let address: SocketAddr = "1.2.3.4".parse().unwrap();
socket.connect(&address.into())?;
```

infiniteregrets: thanks!

PR #1509, opened 2023-06-05
Skip `gcc` - debugging Go with VSCode extension

Razz4780:
When investigating user's issues in Go + VSCode setup I noticed that layer gets loaded into a `gcc` process spawned with `gcc -### -x c -c -`. Probably not what we want


aviramha (approved):

aviramha:
Thanks!

PR #1515, opened 2023-06-07
Avoid unnecessary sudo - for installing mirrord in a Dockerfile

### LINKED ISSUE #1514: Installation script should not require `sudo` to be installed if running as root
When adding `RUN curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash` as step in an ubuntu based Dockerfile, it fails if sudo is not installed in an earlier step:

```
 => ERROR [10/11] RUN curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash                                                                                                                                                                                                3.4s 
------                                                                                                                                                                                                                                                                                                                     
 > [10/11] RUN curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash:                                                                                                                                                                                                           
#19 1.234 Installing version 3.45.0                                                                                                                                                                                                                                                                                        
#19 3.375 installation target directory is write protected, run as root to override                                                                                                                                                                                                                                        
#19 3.375 main: line 32: sudo: command not found                                                                                                                                                                                                                                                                           
------
executor failed running [/bin/sh -c curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash]: exit code: 127
```

It would be nicer not to require users to add step for installing sudo before installing mirrord.

First idea for a solution: first try the command without sudo, only try with sudo if it fails. I think that the `RUN` steps in a Dockerfile run as root, so we should not need sudo there.

t4lz:
Closes #1514

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -29,7 +29,8 @@ copy() {
       mv /tmp/mirrord "$HOME/.local/bin/mirrord"
   else
       echo "installation target directory is write protected, run as root to override"
-      sudo mv /tmp/mirrord /usr/local/bin/mirrord
+      # Try without sudo first, run with sudo only if mv failed without it.
+      mv /tmp/mirrord /usr/local/bin/mirrord || sudo mv /tmp/mirrord /usr/local/bin/mirrord
   fi
 }
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -28,9 +28,11 @@ copy() {
       fi
       mv /tmp/mirrord "$HOME/.local/bin/mirrord"
   else
-      echo "installation target directory is write protected, run as root to override"
       # Try without sudo first, run with sudo only if mv failed without it.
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -29,7 +29,8 @@ copy() {
       mv /tmp/mirrord "$HOME/.local/bin/mirrord"
   else
       echo "installation target directory is write protected, run as root to override"
-      sudo mv /tmp/mirrord /usr/local/bin/mirrord
+      # Try without sudo first, run with sudo only if mv failed without it.
+      mv /tmp/mirrord /usr/local/bin/mirrord || sudo mv /tmp/mirrord /usr/local/bin/mirrord
```
aviramha, line 33: maybe change the above message and print it only when using sudo?

aviramha (approved):
Nice!

PR #1517, opened 2023-06-07
Bogus vscode error

### LINKED ISSUE #1504: VSCode go debugging ends with an error
### Bug Description

When debugging Golang code with VSCode on macOS, after everything works and you finished working, when you click "Stop" the VSCode shows an error "mirrord failed to start" and there's "Connection error: ERROR: read ECONNRESET"

### Steps to Reproduce

not sure, but the above might work.

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS

### Local process

go binary

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Fixes #1504 

An error notification was shown every time the extension was run (on every `close` from the child process stderr). Added a check for stderr being empty.

Also improved error notifications a bit.

### DIFF
--- a/vscode-ext/src/extension.ts
+++ b/vscode-ext/src/extension.ts
@@ -155,7 +155,7 @@ class MirrordAPI {
 					return reject(error["message"]);
 				}
 
-				if (code !== null && code !== 0) {
+				if (code) {
 					return reject(`process exited with error code: ${code}`);
 				}
 
@@ -251,7 +251,7 @@ class MirrordAPI {
 						return reject(error["message"]);
 					}
 
-					if (code !== null && code !== 0) {
+					if (code) {
 						return reject(`process exited with error code: ${code}`);
 					}
 


aviramha:
The stderr can have warnings or logs if logging enabled, so not sure if this condition is smart.

Razz4780:
> The stderr can have warnings or logs if logging enabled, so not sure if this condition is smart.

Hmm, that's right. We use the same check for `mirrord ls`, so I assumed it would be correct. Looks like we have to ignore stderr completely if it doesn't match the error pattern?

aviramha:
> > The stderr can have warnings or logs if logging enabled, so not sure if this condition is smart.
> 
> Hmm, that's right. We use the same check for `mirrord ls`, so I assumed it would be correct. Looks like we have to ignore stderr completely if it doesn't match the error pattern?

I guess - I thought maybe checking error code but even if mirrord was okay the running process might return error exit code.

Razz4780:
Made some improvements to the extension code:
1. Found a bug where an error notification was not shown if two notifications were fired quickly. Fixed it by including error message in the general failed/get help notification.
2. Removed use of `exec` from the code. `exec` takes the command as a string, so we had to escape whitespace
3. Added inspecting error codes and signals

infiniteregrets (approved):
looks good overall! small nit

infiniteregrets:
The e2e is not running, the fix is in #1528, ill move it to a separate pr

PR #1519, opened 2023-06-07
Fix vscode-e2e flake

infiniteregrets:
closes #1518


vscode-ext/src/tests/e2e.ts
```diff
@@ -78,7 +78,7 @@ describe("mirrord sample flow test", function () {
         await setBreakPoint(fileName, browser, defaultTimeout);
         await startDebugging();
 
-        const inputBox = await InputBox.create();
+        const inputBox = await InputBox.create(10000);
```
meowjesty, line 81: What's going on here?

infiniteregrets: we needed to extend the timeout here to 10000 since their default timeout was way too less and lead to some flakes

meowjesty (approved):
:bear: :+1:

infiniteregrets:
i am not sure whats going on, i opened a pr locally and same error!!

Issue #1458, opened 2023-05-25
DNS Resolve fails Spring Boot macOS


aviramha:
### Bug Description

User complains that their Spring Boot service fails to resolve DNS.

### Steps to Reproduce

Use this repository, have a Redis server in the cluster with service name `redis-man`
https://github.com/aviramha/issue-1458

### Relevant Logs

```shell
2023-05-25 04:04:03.602 ERROR [progression-svc,,] 89301 --- [           main] i.n.r.d.DnsServerAddressStreamProviders  : Unable to load io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider, fallback to system defaults. This may result in incorrect DNS resolutions on MacOS. Check whether you have a dependency on 'io.netty:netty-resolver-dns-native-macos'. Use DEBUG level to see the full stack: java.lang.UnsatisfiedLinkError: failed to load the required native library
```

```
Caused by: io.lettuce.core.RedisConnectionException: Unable to connect to keydb.island-milestone-infra.svc/<unresolved>:6379
        at io.lettuce.core.RedisConnectionException.create(RedisConnectionException.java:78)
        at io.lettuce.core.RedisConnectionException.create(RedisConnectionException.java:56)
        at io.lettuce.core.AbstractRedisClient.getConnection(AbstractRedisClient.java:330)
        at io.lettuce.core.RedisClient.connect(RedisClient.java:216)
        at org.springframework.data.redis.connection.lettuce.StandaloneConnectionProvider.lambda$getConnection$1(StandaloneConnectionProvider.java:115)
        at java.base/java.util.Optional.orElseGet(Optional.java:364)
        at org.springframework.data.redis.connection.lettuce.StandaloneConnectionProvider.getConnection(StandaloneConnectionProvider.java:115)
        at org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory$ExceptionTranslatingConnectionProvider.getConnection(LettuceConnectionFactory.java:1595)
        ... 86 more
Caused by: io.netty.resolver.dns.DnsResolveContext$SearchDomainUnknownHostException: Failed to resolve 'keydb.island-milestone-infra.svc' and search domain query for configured domains failed as well: [router]
````
```


### Your operating system and version

macOS

### Local process

Spring Boot

### Local process version

_No response_

### Additional Info

_No response_

clrxbl:
[...]

PR #1520, opened 2023-06-07
Hook `sendto` and `recvfrom` to resolve DNS.

meowjesty:
- Issue: #1458 

1. Adds hooks for libc `send_to` and `recv_from`;
2. Removes `build_intellij` from CI;

`send_to` hook creates an UDP connection, taking advantage of how we handle outgoing UDP socket stuff. It sends a message to the connected interceptor socket, and when the user calls `recv_from`, they should get the correct messages back.

This has only been implemented to the specific flow of how [`netty`](https://netty.io/) resolves DNS:

1. `send_to` with remote address being a DNS resolver (let's call it `A:53`);
3. `recv_from` receives data from `A:53`;
4. `recv_from` receives rest of the data from `A:53`;

Thus, from this flow, we can assume that our connection + interceptor mechanism should work.


mirrord/layer/tests/apps/issue-1458/src/main/java/issue/Issue1458.java
```diff
@@ -0,0 +1,48 @@
+/*
```
aviramha, line 1: FYI, my 2 cents - as much as I am happy to have some Java in our integration - if this becomes too hard to maintain/get working we can use Rust/Python/C (any language that lets us reproduce the case for the integration test)

meowjesty: Alright. When I come back to this, will check for some github action (think mehul mentioned it), as the sample code is very self-contained (only dependency is netty + maven).

mirrord/layer/src/socket/ops.rs
```diff
@@ -765,13 +774,96 @@ pub(super) fn gethostname() -> Detour<&'static CString> {
...
+)]
+pub(super) fn send_to(
+    sockfd: RawFd,
+    raw_message: *const c_void,
+    message_length: usize,
+    flags: i32,
+    raw_destination: *const sockaddr,
+    destination_length: socklen_t,
+) -> Detour<isize> {
+    let destination = SockAddr::try_from_raw(raw_destination, destination_length)?;
```
aviramha, line 838: we should add a test when we change addresses inbetween send_to with recv_from from both to see that it's working as well

aviramha: if it doesn't work maybe we should make this implementation available only for dns (port 53) to avoid introducing bugs in the apps

meowjesty: Alright, now that the basic integration test is working, I'll try testing this out.

Should this patch be part of `MIRRORD_UDP_OUTGOING` or `MIRRORD_DNS` config? I can see an user wanting dns resolution, but disabling udp, so I'm not sure where to gate this.

meowjesty: Quick test shows that we end up blocking on the second `sendto`:

```sh
sent 10 to 127.0.0.1:8888
2023-06-13T21:01:18.363645Z TRACE ThreadId(12) interceptor_task: mirrord_layer::outgoing::udp: interceptor_task -> Received data from user socket 127.0.0.1:41947 connection_id=0 mirror_socket=PollEvented { io: Some(UdpSocket { addr: 127.0.0.1:52083, fd: 13 }) }

(logs showing that stuff is working ...)

sent 10 to 127.0.0.1:9999
2023-06-13T21:01:18.408825Z TRACE ThreadId(19) interceptor_task: mirrord_layer::outgoing::udp: interceptor_task -> Received data from user socket 127.0.0.1:41947 connection_id=1 mirror_socket=PollEvented { io: Some(UdpSocket { addr: 127.0.0.1:32931, fd: 14 }) }

2023-06-13T21:01:18.408869Z TRACE ThreadId(01) recv_from: mirrord_layer::socket::ops: new sockfd=10 buffer_length=20 flags=0

(we never receive anything, `recv_from` just blocks)

2023-06-13T21:06:48.472229Z TRACE ThreadId(01) recv_from: mirrord_layer::socket::ops: new sockfd=10 buffer_length=20 flags=0
2023-06-13T21:06:48.472275Z TRACE ThreadId(19) next: mirrord_layer::tcp_steal: new
2023-06-13T21:06:48.472287Z TRACE ThreadId(19) next: mirrord_layer::tcp_steal: close time.busy=1.98µs time.idle=10.4µs
2023-06-13T21:06:48.472306Z TRACE ThreadId(19) mirrord_layer::connection: Sending client message: UdpOutgoing(Write(LayerWrite { connection_id: 1, bytes (length): 10 }))

(we stay here forever)
```

mirrord/layer/tests/apps/issue-1458/src/main/java/issue/Issue1458.java
meowjesty: Dropped java in favor of a simpler rust test.

mirrord/layer/src/socket/ops.rs
meowjesty: Wow, what a wild ride this became:

1. `fill_address`:

To properly `fill_address` of a `recvfrom` address, we have to check if this address belongs to us, due to how we change a bound address in `bind`.

2. port 0:

Now that we have port 0 enabled to `bind` for UDP, when the user calls `getsockname` they should get a valid address + port combo, `bind(127.0.0.1:0)` has to become address `127.0.0.1:{not 0}` (this is what happens without mirrord).

3. `sendto` and `recvfrom`:

These functions now have to check if the address they're sending/receiving is an actual address, that we can send packets on.

mirrord/layer/src/socket.rs
```diff
@@ -95,7 +95,7 @@ impl SocketInformation {
 
 #[derive(Debug)]
 pub struct Connected {
-    /// Remote address we're connected to.
+    /// Remote address we're (supposedly) connected to.
```
eyalb181, line 98: Yeah these parentheses are gonna need an explanation

meowjesty: Better?

```rust
    /// The address requested by the user that we're "connected" to.
    ///
    /// Whenever the user calls [`libc::getpeername`], this is the address we return to them.
    ///
    /// For the _outgoing_ feature, we actually connect to the `layer_address` interceptor socket,
    /// but use this address in the [`libc::recvfrom`] handling of [`fill_address`].

```

eyalb181: Yes! Thanks.

eyalb181 (approved):

Issue #1105, opened 2023-02-22
Support unix sockets


aviramha:
Support making outgoing connections from to unix sockets on the remote pod.

Requested by customer who's application reads from the containerd socket.

aviramha:
Adding more information, as I initially started implementing it and have some thoughts:

1. This should be part of the "outgoing" configuration, a field named `unix` (feel free to suggest other) that accepts path patterns to connect remotely.
2. This should leverage the TCP outgoing handler, as it is pretty similar to it in terms of everything (connect, send/recv).
3. Above point is debatable, but I don't really think that splitting and complicating the code makes sense.

t4lz:
> 2\. This should leverage the TCP outgoing handler

Does this mean that for now we are only going to support TCP over unix sockets, not UDP?

aviramha:
It's possible to have UDP over unix sockets?

PR #1521, opened 2023-06-07
Update socket2 dependency

t4lz:
In https://github.com/rust-lang/socket2/pull/403, we added some code to `socket2`, for https://github.com/metalbear-co/mirrord/issues/1105. Up until now the dependency was on my `socket2` fork, but a version that includes those changes was released already, so we can depend on the new version of the official crate again.


meowjesty (approved):
🐻 👍

Issue #1512, opened 2023-06-07
Steal based on HTTP Path Filter


aviramha:
Currently we support filtering based on HTTP headers. Some of the users asked to support stealing based on the path accessed.

- [x] merging #1550  
- [x] adding to operator

aviramha:
We probably need to deprecate the old filter config, since it's "explicitly" http header. Having said that, would we want the filters to be `or` / `and` / or mutual exclusive? like if user specifies a http header filter and path header, should it be matching both or only one or should we just error for now?

eyalb181:
Personally I don't see a use case for either, and would rather error out until a user asks for it than support a use case we don't understand.

PR #1525, opened 2023-06-08
HTTP Path Filtering partial

aviramha:
Partial work for #1512 - would love to merge it already to avoid huge PR :)

### DIFF
--- a/mirrord/agent/src/steal/http/filter.rs
+++ b/mirrord/agent/src/steal/http/filter.rs
@@ -63,7 +63,7 @@ impl HttpFilter {
     }
 }
 
-/// Struct to hold cache/hold/reuse the attributes that a [`Request`] can be matched upon
+/// Struct to cache/hold/reuse the attributes that a [`Request`] can be matched upon
 pub(crate) struct RequestMatchCandidate<'a> {
     /// The original request
     request: &'a Request<Incoming>,
--- a/mirrord/agent/src/steal/http/filter.rs
+++ b/mirrord/agent/src/steal/http/filter.rs
@@ -15,16 +15,12 @@
 use std::{future::Future, net::SocketAddr, sync::Arc, time::Duration};
 
 use dashmap::DashMap;
+use enum_dispatch::enum_dispatch;
 use fancy_regex::Regex;
-use hyper::rt::Executor;
+use hyper::{body::Incoming, rt::Executor, Request};
 use mirrord_protocol::ConnectionId;
 use tokio::{io::copy_bidirectional, net::TcpStream, sync::mpsc::Sender};
 use tracing::{error, trace};
-use enum_dispatch::enum_dispatch;
-use hyper::{
-    body::Incoming,
-    Request
-};
 
 use super::{
     error::HttpTrafficError, v1::HttpV1, v2::HttpV2, DefaultReversibleStream, HttpV, HttpVersion,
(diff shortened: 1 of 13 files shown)


mirrord/agent/src/steal/http/filter.rs
```diff
@@ -35,6 +37,108 @@ const H2_PREFACE: &[u8; 14] = b"PRI * HTTP/2.0";
...
+    pub fn new_header_filter(header: Regex) -> Self {
+        Self::Header(HttpHeaderFilter(header))
+    }
+    #[allow(dead_code)]
+    pub fn new_path_filter(path: Regex) -> Self {
+        Self::Path(HttpPathFilter(path))
+    }
+}
+
+/// Struct to hold cache/hold/reuse the attributes that a [`Request`] can be matched upon
```
meowjesty, line 66:
```suggestion
/// Struct to cache/hold/reuse the attributes that a [`Request`] can be matched upon
```

One "hold" too much?

mirrord/agent/src/steal/http/filter.rs
```diff
@@ -35,6 +37,108 @@ const H2_PREFACE: &[u8; 14] = b"PRI * HTTP/2.0";
...
+                    error!(
+                        "Error while matching header: {header:?}, {:?}, {err:?}",
+                        self.0
+                    );
+                    return false;
+                }
+            }
+        }
+        false
+    }
```
meowjesty, line 129:
```suggestion
request.headers().find_map(|header| self.0.is_match(header).unwrap_or_default()).unwrap_or_default()
```

Something along those lines? (sorry, I can't help myself not suggesting iterators)

mirrord/agent/src/steal/http/filter.rs
```diff
@@ -35,6 +37,108 @@ const H2_PREFACE: &[u8; 14] = b"PRI * HTTP/2.0";
...
+        Self::Path(HttpPathFilter(path))
+    }
+}
+
+/// Struct to hold cache/hold/reuse the attributes that a [`Request`] can be matched upon
+pub(crate) struct RequestMatchCandidate<'a> {
+    /// The original request
+    request: &'a Request<Incoming>,
+    /// Headers in `k: v` format.
+    headers: OnceCell<Vec<String>>,
```
meowjesty, line 71: Not understanding why we're using `OnceCell` here?

Can't we just initialize an empty `Vec`?

mirrord/agent/src/steal/http/hyper_handler.rs
```diff
@@ -76,31 +79,19 @@ pub(super) struct RawHyperConnection {
...
-                        header
-                            .as_ref()
-                            .expect("The header value has to be convertible to `String`!"),
-                    )
-                    .expect("Something went wrong in the regex matcher!")
-                    .then_some(*filter.key())
-            })
-        })
+    let mut request = request.into();
+    filters.iter().find_map(|entry| {
```
meowjesty, line 88: sweet, sweet iterators

meowjesty (changes requested):
I assume the config for this is coming in another PR?

My first time seeing the `protocol_break` macro.

Only big thing that I'm not getting is the `OnceCell<Vec>>`, otherwise looks :+1:

mirrord/agent/src/steal/http/filter.rs
aviramha: The idea is to fill this lazily, only when needed. If the only filters that exist are path based, we wouldn't need to even initialize it.

mirrord/agent/src/steal/http/filter.rs
aviramha: I think with error handling the loop looks better

mirrord/agent/src/steal/http/filter.rs
meowjesty: hmm I get it. I'm not a fan of lazy initialization, but think this is too minor to require arguing against.

I can see that adds more meaning, just looks a bit weird (at least initially).

meowjesty (approved):
🐻 👍

PR #1527, opened 2023-06-09
support kubectl port-forward

### LINKED ISSUE #1479: request through `kubectl port forward` doesn't get stolen
When stealing pod requests, if one uses `kubectl port forward deploy/deploymentname 8080` those requests still get answered by the remote.

infiniteregrets:
closes #1479


mirrord/agent/src/steal/ip_tables/redirect.rs
```diff
@@ -98,6 +98,83 @@ where
     }
 }
 
+pub struct StandardRedirect<IPT: IPTables> {
```
DmitryDodzin, line 101: I would move this struct to `standard.rs`

mirrord/agent/src/steal/ip_tables/redirect.rs
```diff
@@ -98,6 +98,83 @@ where
...
+    preroute: PreroutingRedirect<IPT>,
```
DmitryDodzin, line 102: I see some usage of preroute's `managed` field and this can cause problems with the unloading nature of the ip table guards, I would recommend using a separate `IPTableChain` for this Redirect
```suggestion
    preroute: PreroutingRedirect<IPT>,
    managed: IPTableChain<IPT>,
```

mirrord/agent/src/steal/ip_tables/redirect.rs
```diff
@@ -98,6 +98,83 @@ where
...
+#[async_trait]
+impl<IPT> Redirect for StandardRedirect<IPT>
+where
+    IPT: IPTables + Send + Sync,
+{
+    async fn mount_entrypoint(&self) -> Result<()> {
+        self.preroute.managed.inner().add_rule(
+            PreroutingRedirect::<IPT>::ENTRYPOINT,
+            &format!("-j {}", self.preroute.managed.chain_name()),
+        )?;
```
DmitryDodzin, line 135:
```suggestion
        self.preroute.mount_entrypoint()?;
```

mirrord/agent/src/steal/ip_tables/redirect.rs
```diff
@@ -98,6 +98,83 @@ where
...
+
+        Ok(())
+    }
+
+    async fn unmount_entrypoint(&self) -> Result<()> {
+        self.preroute.managed.inner().remove_rule(
+            PreroutingRedirect::<IPT>::ENTRYPOINT,
+            &format!("-j {}", self.preroute.managed.chain_name()),
+        )?;
```
DmitryDodzin, line 149:
```suggestion
        self.preroute.unmount_entrypoint()?;
```

mirrord/agent/src/steal/ip_tables/redirect.rs
```diff
@@ -98,6 +98,83 @@ where
...
+            Self::ENTRYPOINT,
+            &format!("-j {}", self.preroute.managed.chain_name()),
+        )?;
+
+        Ok(())
+    }
+
+    async fn add_redirect(&self, redirected_port: Port, target_port: Port) -> Result<()> {
+        let redirect_rule =
+            format!("-m tcp -p tcp --dport {redirected_port} -j REDIRECT --to-ports {target_port}");
```
DmitryDodzin, line 161: I fear just reusing the same chain can be problematic, I would check that a outgoing request to another service listening on the same port as your's does, (I fear the rules would need to be a bit different like adding a filter on gid or something simmilar)

DmitryDodzin (changes requested):
Looks Great but I have some concerns with using the same chain for both `PREROUTING` & `OUTPUT`

aviramha:
before merging this, we have to test this on mesh to see we haven't broken it as we don't have mesh in the CI.

DmitryDodzin (approved):
Nitpick =]

PR #1528, opened 2023-06-09
refactor vscode-ext to be more extensible

infiniteregrets:


.github/workflows/ci.yaml
```diff
@@ -228,7 +228,7 @@ jobs:
   test_agent_image:
     runs-on: ubuntu-latest
     needs: changed_files
-    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
+    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' || needs.changed_files.outputs.vscode_changed == 'true'}}
```
eyalb181, line 231: Why is this necessary?

vscode-ext/src/util.ts
```diff
@@ -0,0 +1,20 @@
+import * as vscode from 'vscode';
+
+/// Key used to store the last selected target in the persistent state.
+export const LAST_TARGET_KEY = "mirrord-last-target";
+
+/// Option in the target selector that represents no target.
+export const TARGETLESS_TARGET_NAME = "No Target (\"targetless\")";
+
+// Display error message with help
+export function mirrordFailure(err: any) {
```
eyalb181, line 10: I'd merge this entire file into api.ts, it's only used there.

vscode-ext/src/debugger.ts
```diff
@@ -0,0 +1,126 @@
...
+			return config;
+		}
+
+		// For some reason resolveDebugConfiguration runs twice for Node projects. __parentId is populated.
+		if (config.__parentId || config.env?.["__MIRRORD_EXT_INJECTED"] === 'true') {
+			return config;
+		}
+
+		// do not send telemetry for CI runs
+		if (vscode.env.isTelemetryEnabled && !CI_BUILD_PLUGIN) {
```
eyalb181, line 50: The checkVersion logic is split between this file and versionCheck.ts. Can this entire `if` be moved into versionCheck.ts?

eyalb181 (changes requested):
Looks good in general, made some comments

.github/workflows/ci.yaml
infiniteregrets: thanks for the review! the `vscode_e2e` job depends on `test_agent_image` and `build_mirrord` and hence they need to run as well to make `vscode_e2e` run

vscode-ext/src/util.ts
infiniteregrets: its also used by the extension.ts file, however i believe it makes sense to keep it with the API, we don't really have any "utility functions" as such anyway, fixed!

.github/workflows/ci.yaml
eyalb181: I understand that, but why is the vscode job dependent on the agent tests running (as opposed to just building the agent image)?

vscode-ext/src/api.ts
```diff
@@ -0,0 +1,267 @@
+import * as vscode from 'vscode';
+import * as path from 'node:path';
+import * as fs from 'node:fs';
+import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+import { globalContext } from './extension';
+const util = require('node:util');
+const exec = util.promisify(require('node:child_process').exec);
```
Razz4780, line 7: These two should no longer be needed, right?

Razz4780 (changes requested):

vscode-ext/src/api.ts
infiniteregrets: yes you are right, thanks!

Razz4780 (approved):

eyalb181 (approved):

PR #1532, opened 2023-06-10
Frida bump and other dependencies

aviramha:


eyalb181 (approved):

mirrord/layer/src/error.rs
```diff
@@ -338,3 +338,9 @@ impl From<HookError> for *mut c_char {
         ptr::null_mut()
     }
 }
+
+impl From<frida_gum::Error> for LayerError {
```
DmitryDodzin, line 342: Why manual impl? isn't this the same as `#[from]`

aviramha: `#[from]` requires `impl std::error::Error` and frida-rust was changed to be no-stdlib by default, so its errors don't provide that.

PR #1548, opened 2023-06-12
fixed using dotnet debugger on VSCode

### LINKED ISSUE #1529: unable to debug C# application with vscode
configuration:

```json
{
    "accept_invalid_certificates": false,
    "feature": {
        "network": {
            "incoming": "steal",
            "outgoing": true
        },
    },
    "target": {
          "path": "pod/x",
          "namespace": "dev"
     },
}
```
user gets this error:
```
The target process exited without raising a CoreCLR started event. Ensure that the target process is configured to use .NET Core. This may be expected if the target process did not run on .NET Core.
```

OS: macos

aviramha:
fixed using dotnet debugger on VSCode closes #1529


mirrord/layer/src/lib.rs
```diff
@@ -304,6 +304,21 @@ fn layer_pre_initialization() -> Result<(), LayerError> {
...
+        if let Ok(path) = std::env::current_exe() {
+            if let Ok(Some(binary)) =
+                mirrord_sip::sip_patch(path.to_str().unwrap(), &patch_binaries)
+            {
+                let err = exec::execvp(binary, std::env::args());
+                error!("Couldn't execute {:?}", err);
+                return Err(LayerError::ExecFailed(err));
+            }
+        }
+    }
```
Razz4780, line 320: Is this the right place for this logic? Seems oddly specific, haven't we encountered any similiar case in the past?

mirrord/layer/src/load.rs
```diff
@@ -31,6 +32,11 @@ pub enum LoadType {
 pub fn load_type(given_process: &str, config: LayerConfig) -> LoadType {
     let skip_processes = config.skip_processes.clone().map(VecOrSingle::to_vec);
 
+    #[cfg(target_os = "macos")]
+    {
+        if given_process.ends_with("dotnet") {}
```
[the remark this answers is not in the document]
aviramha: oops

mirrord/layer/src/lib.rs
aviramha: not that I know of, I thought of adding a new enum variant for the `LoadType` but seems overkillatm

Razz4780: imho new variant would be more clear, keeping the logic in one place

aviramha: but it's kinda same sport (2 lines above)

Razz4780: not pushing it, it's fine with me

eyalb181 (approved):

PR #1550, opened 2023-06-13
HTTP Path Filter

aviramha:
related to #1512


mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -12,6 +12,7 @@ use crate::{
 // TODO(alex) [high] 2023-05-18: Wrong example (incomplete) here and in the oficial docs.
 /// Filter configuration for the HTTP traffic stealer feature.
 ///
+/// DEPRECATED - USE http_filter instead, unless using old operator/agent version (pre 3.46.0)
```
DmitryDodzin, line 15: I'd add `#[config(deprecated)]` to `filter` and `ports` to show warning in runtime

mirrord/config/src/lib.rs
```diff
@@ -19,12 +19,26 @@ use std::path::Path;
...
+                                           not be handled. The target container running the deployed application is paused,
+                                           and responses from the local application are dropped.
+
+                                           Attention: if network based liveness/readiness probes are defined for the
+                                           target, they will fail under this configuration.
+
+                                           To have the local application handle incoming requests you can run again with
+                                           `--steal`. To have the deployed application handle requests, run again without
+                                           specifying `--pause`.
+    ";
```
t4lz, line 40: I think I did something weird there with the indentation. Can we fix it now?
Maybe like this:
```suggestion
    "--pause specified without --steal: Incoming requests to the application will
    not be handled. The target container running the deployed application is paused,
    and responses from the local application are dropped.

    Attention: if network based liveness/readiness probes are defined for the
    target, they will fail under this configuration.

    To have the local application handle incoming requests you can run again with
    `--steal`. To have the deployed application handle requests, run again without
    specifying `--pause`.
    ";
```

mirrord/config/src/feature/network/incoming/http_filter.rs
aviramha: I tried to do it to the root object, (incomingconfig) but that didn't work, good suggestion

mirrord/protocol/src/tcp.rs
```diff
@@ -113,7 +113,7 @@ pub enum StealType {
     /// PROTOCOL
     FilteredHttp(Port, Filter),
     /// Steal HTTP traffic matching a given filter - supporting more than once kind of filter
-    FilteredHttpMany(Port, HTTPFilter),
+    FilteredHttpEx(Port, HttpFilter),
```
t4lz, line 116: I don't get the name 😬

mirrord/cli/src/main.rs
```diff
@@ -162,15 +149,7 @@ async fn exec(args: &ExecArgs, progress: &TaskProgress) -> Result<()> {
...
-    if config.pause {
-        if config.agent.ephemeral {
-            error!("Pausing is not yet supported together with an ephemeral agent container.");
-            panic!("Mutually exclusive arguments `--pause` and `--ephemeral-container` passed together.");
-        }
-        if !config.feature.network.incoming.is_steal() {
-            warn!("{PAUSE_WITHOUT_STEAL_WARNING}");
-        }
-    }
+    config.verify()?;
```
t4lz, line 152: What do you say about moving this to `MirrordExecution::start`, so that verification is also done when running from an IDE? It could be called [where the targetless config checks are done now](https://github.com/metalbear-co/mirrord/blob/79061d1e755c0f7da86e1fc1da9f222df8440b74/mirrord/cli/src/execution.rs#L60), and those checks could be moved into `verify()`.

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -51,6 +54,54 @@ pub struct HttpHeaderFilterConfig {
...
+/// Allows the user to set a filter (regex) for the HTTP headers, so that the stealer traffic
+/// feature only captures HTTP requests that match the specified filter, forwarding unmatched
+/// requests to their original destinations.
+///
+/// Only does something when [`feature.network.incoming.mode`](#feature-network-incoming-mode) is
+/// set as `"steal"`, ignored otherwise.
+///
+/// ```json
+/// {
+///   "filter": "host: api\..+",
```
t4lz, line 68: `"filter"` doesn't seem right.

mirrord/layer/src/tcp_steal/http.rs
```diff
@@ -195,3 +198,52 @@ where
...
+    /// Header or Path, new HTTP filter.
+    Path(String),
+}
+
+pub(crate) struct HttpFilterSettings {
+    /// The HTTP filter to use.
+    pub(crate) filter: LayerHttpFilter,
+    /// Ports to filter HTTP on
+    pub(crate) ports: HashSet<Port>,
+}
```
t4lz, line 218: Why not use the http filter config?

mirrord/layer/src/tcp_steal.rs
```diff
@@ -230,8 +228,20 @@ impl TcpHandler for TcpStealHandler {
             return Ok(());
         }
 
-        let steal_type = if self.http_ports.contains(&original_port) && let Some(filter_str) = self.http_filter.take() {
-            FilteredHttp(request_port, Filter::new(filter_str)?)
+        let steal_type = if self.http_filter_settings.ports.contains(&original_port) {
+            match self.http_filter_settings.filter {
+                LayerHttpFilter::None => All(request_port),
+                LayerHttpFilter::HeaderDeprecated(ref header) => {
+                    FilteredHttp(request_port, Filter::new(header.clone())?)
```
t4lz, line 235: We could also use the new type if the user specified the config in the old way, no?

mirrord/layer/src/tcp_steal.rs
```diff
@@ -230,8 +228,20 @@ impl TcpHandler for TcpStealHandler {
             return Ok(());
         }
 
-        let steal_type = if self.http_ports.contains(&original_port) && let Some(filter_str) = self.http_filter.take() {
-            FilteredHttp(request_port, Filter::new(filter_str)?)
+        let steal_type = if self.http_filter_settings.ports.contains(&original_port) {
+            match self.http_filter_settings.filter {
+                LayerHttpFilter::None => All(request_port),
```
t4lz, line 233: Thought about existing code, not this PR: It's a bit weird our config allows http filter ports to be specified without a filter. Maybe we can add a todo to change the config such that it's not possible on the type system level, so that we don't have to handle those useless situations.

t4lz (changes requested):
Looks good, had a couple of minor comments/questions.

Also: did you forget to push a test for the path filter? There is a config for that test but no test.

aviramha:
> Looks good, had a couple of minor comments/questions.
> 
> Also: did you forget to push a test for the path filter? There is a config for that test but no test.

Thanks! No, I just held it to see that the normal flow works first.

mirrord/layer/src/tcp_steal.rs
aviramha: We could, but I want to let the users have the ability to configure if old protocol or new essentially.

mirrord/protocol/src/tcp.rs
aviramha: too much time on windows but this means `Extended`

mirrord/layer/src/tcp_steal.rs
aviramha: not sure where I should add this todo?

t4lz:
> > Looks good, had a couple of minor comments/questions.
> > Also: did you forget to push a test for the path filter? There is a config for that test but no test.
> 
> Thanks! No, I just held it to see that the normal flow works first.

I don't get it 😅
Why not add the test now? Why add the new feature without a test?

aviramha:
> I don't get it 😅 Why not add the test now? Why add the new feature without a test?

Hell no! I meant I want to see that the CI passes after changes, then introduce the test (as part of the same PR) - just probably won't be the focus of review so asked to start now :)

tests/src/traffic/steal.rs
```diff
@@ -185,6 +186,60 @@ mod steal {
...
+        let mut flags = Vec::new();
+        if let Some(flag) = agent.flag() {
+            flags.extend(flag)
+        }
+
+        let mut client = application
+            .run(
+                &service.target,
+                Some(&service.namespace),
+                Some(flags),
```
t4lz, line 215:
```suggestion
        let mut client = application
            .run(
                &service.target,
                Some(&service.namespace),
                agent.flag(),
```

tests/src/traffic/steal.rs
```diff
@@ -154,14 +154,15 @@ mod steal {
...
-        }
 
         let mut config_path = config_dir.clone();
         config_path.push("http_filter_header.json");
 
+        let mut flags = Vec::new();
+        if let Some(flag) = agent.flag() {
+            flags.extend(flag)
+        }
+
```
t4lz, line 165:
```suggestion
```

Also here: just pass `agent.flag()` to `run`

tests/src/traffic/steal.rs
```diff
@@ -185,6 +186,60 @@ mod steal {
         application.assert(&client);
     }
 
+    #[rstest]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[timeout(Duration::from_secs(120))]
+    async fn filter_with_single_client_and_only_matching_requests_by_path(
```
t4lz, line 192: This test is called `...only_matching_requests...` but contains a `GET` request that is unmatched.
I don't know if sending that `GET` request somehow verifies that the request was not stolen.

I suggest changing the test's name, removing the `GET` request, and adding a `DELETE` request that doesn't get stolen before sending the one that does. That way we actually verify that an unmatched request is not stolen - because if it were, the local app would exit. There is already a comment for that in line 228, that I think is currently wrong.

t4lz (changes requested):

tests/src/traffic/steal.rs
aviramha: We verify the response

t4lz: Nice, so we can just change the test name and remove the wrong comment.

aviramha: Yup, done!

t4lz (approved):

PR #1557, opened 2023-06-13
Handle user app closing a socket

### LINKED ISSUE #1530: Handle socket close
I think we don't update the layer's TcpHandlers and the agent if the app closes a socket.

We could maybe send a hook message from `close...`, handle it on `TcpHandler` and tell the agent to stop listening.

t4lz:
closes #1530.

When `close()` is called, if it's a socket managed by mirrord - notify the TcpHandler (mirror/steal) via `HookMessage`, which removes the port from the listen set, and sends a `PortUnsubscribe` to the agent.

The agent side's handling was already implemented.

### DIFF
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -518,6 +518,9 @@ impl TcpConnectionStealer {
         };
         if port_unsubscribed {
             // No remaining subscribers on this port.
+
+            // If there are ongoing connections, this will interrupt them. This is a known issue
+            // https://github.com/metalbear-co/mirrord/issues/1575
             self.iptables()?
                 .remove_redirect(port, self.stealer.local_addr()?.port())
                 .await?;
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -118,11 +118,11 @@ use tracing::{error, info, trace, warn};
 use tracing_subscriber::{fmt::format::FmtSpan, prelude::*};
 
 use crate::{
-    common::HookMessage,
+    common::{blocking_send_hook_message, HookMessage},
     debugger_ports::DebuggerPorts,
     file::{filter::FILE_FILTER, FileHandler},
     load::LoadType,
-    socket::CONNECTION_QUEUE,
+    socket::{SocketKind, CONNECTION_QUEUE},
 };
 
 mod common;
@@ -934,6 +934,20 @@ pub(crate) fn close_layer_fd(fd: c_int) {
     // Remove from sockets, also removing the `ConnectionQueue` associated with the socket.
     if let Some((_, socket)) = SOCKETS.remove(&fd) {
         CONNECTION_QUEUE.remove(socket.id);
+
+        // Closed file is a socket, so if it's already bound to a port - notify agent to stop
+        // mirroring/stealing that port.
+        if let Some(port) = socket.get_port() {
+            match socket.kind {
+                SocketKind::Tcp(_) => {
+                    // TODO: can we do anything with this result?
+                    let _ =
+                        blocking_send_hook_message(HookMessage::Tcp(tcp::TcpIncoming::Close(port)));
+                }
+                // TODO: I think we don't need to notify the agent about a udp socket close, right?
+                SocketKind::Udp(_) => {}
+            }
+        }
     } else if file_mode_active {
         OPEN_FILES.remove(&fd);
     }
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -940,18 +940,21 @@ pub(crate) fn close_layer_fd(fd: c_int) {
         if let Some(port) = socket.get_port() {
             match socket.kind {
                 SocketKind::Tcp(_) => {
-                    let SocketState::Connected(_) = socket.state else {
-                        // if the socket state is connected, then closing its fd is closing the one 
-                        // connection, not the whole listening socket. We will no about this close
-                        // from the mirrord layer's side of the socket and will inform the agent
-                        // about the closed connection from there.
-                        // TODO: can we do anything with this result?
-                        let _ =
-                            blocking_send_hook_message(HookMessage::Tcp(tcp::TcpIncoming::Close(port)));
-                        return;
-                    };
+                    match socket.state {
+                        // if the socket state is connected, then closing its fd is closing the
+                        // one connection, not the whole listening socket. We will know about this
+                        // close from the mirrord layer's side of the socket and will inform the
+                        // agent about the closed connection from there.
+                        SocketState::Connected(_) => {}
+                        _ => {
+                            // TODO: can we do anything with this result?
+                            let _ = blocking_send_hook_message(HookMessage::Tcp(
+                                tcp::TcpIncoming::Close(port),
+                            ));
+                        }
+                    }
                 }
-                // TODO: I think we don't need to notify the agent about a udp socket close, right?
+                // We don't do incoming UDP, so no need to notify anyone about this.
                 SocketKind::Udp(_) => {}
             }
         }
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -122,7 +122,7 @@ use crate::{
     debugger_ports::DebuggerPorts,
     file::{filter::FILE_FILTER, FileHandler},
     load::LoadType,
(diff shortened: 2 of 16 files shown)


mirrord/agent/src/steal/connection.rs
```diff
@@ -505,6 +505,15 @@ impl TcpConnectionStealer {
         client_id: ClientId,
         port: Port,
     ) -> Result<(), AgentError> {
+        // Remove all existing connections of the unsubscribing client on that port.
```
aviramha, line 508: when you stop listening (close the listening socket) it doesn't close all the established connections..

t4lz: Hey, sorry, just converted this PR into a draft to remove this part 😆 

Removing the connections will of course happen when the layer sees the connection is closed and send a `ConnectionUnsubscribe` to the agent. Don't know what I was thinking when I added these lines.

They also remove all the client's connections, no matter what port. 🤦‍♂️

mirrord/layer/src/lib.rs
```diff
@@ -935,6 +935,33 @@ pub(crate) fn close_layer_fd(fd: c_int) {
...
+                            let _ = blocking_send_hook_message(HookMessage::Tcp(
+                                tcp::TcpIncoming::Close(port),
+                            ));
+                        }
+                    }
+                }
+                // We don't do incoming UDP, so no need to notify anyone about this.
+                SocketKind::Udp(_) => {}
+            }
+        }
```
meowjesty, line 964: Add this whole thing as a method for `UserSocket`, and here just call:

```suggestion
        socket.close();
```

mirrord/layer/src/tcp.rs
```diff
@@ -82,13 +85,41 @@ pub(crate) trait TcpHandler {
...
+    fn check_connection_handling_result(res: Result<(), LayerError>) -> Result<(), LayerError> {
+        if let Err(LayerError::NewConnectionAfterSocketClose(connection_id)) = res {
+            // This can happen and is valid:
+            // 1. User app closes socket.
+            // 2. mirrord sends `PortUnsubscribe` to agent.
+            // 3. Agent sniffs new incoming connection and sends it to the layer.
+            // 4. Agent receives `PortUnsubscribe`.
+            // Step 2 could even happen after 4.
+            if Self::IS_STEAL {
+                warn!(
```
meowjesty, line 98: I'm not quite sure what our `warn` policy is, should we do it even though this is valid and returns `Ok`?

Does it impact the user?

wdyt @aviramha

mirrord/layer/src/tcp.rs
```diff
@@ -82,13 +85,41 @@ pub(crate) trait TcpHandler {
...
     /// Returns true to let caller know to keep running
     #[tracing::instrument(level = "trace", skip(self))]
     async fn handle_daemon_message(&mut self, message: DaemonTcp) -> Result<(), LayerError> {
         let handled = match message {
-            DaemonTcp::NewConnection(tcp_connection) => {
-                self.handle_new_connection(tcp_connection).await
-            }
+            DaemonTcp::NewConnection(tcp_connection) => Self::check_connection_handling_result(
+                self.handle_new_connection(tcp_connection).await,
+            ),
```
meowjesty, line 122:
```suggestion
            DaemonTcp::NewConnection(tcp_connection) => self.handle_new_connection(tcp_connection).await.and_then(Self::check_connection_handling_result),
```

Looks nicer? Less nice? Think they're equivalent right?

mirrord/layer/src/tcp_steal.rs
```diff
@@ -111,6 +111,8 @@ pub struct TcpStealHandler {
 }
 
 impl TcpHandler for TcpStealHandler {
+    const IS_STEAL: bool = true;
```
meowjesty, line 114: Mom always told me that stealing is wrong!

mirrord/layer/src/tcp.rs
```diff
@@ -69,6 +71,7 @@ impl From<&Listen> for SocketAddr {
 }
 
 pub(crate) trait TcpHandler {
+    const IS_STEAL: bool;
```
meowjesty, line 74: We don't hold `TcpHandler` as `Box<dyn TcpHandler>` anywhere right?

This could be turned in a const generic parameter on `TcpHandler` (I think).

Not worth going into this rabbit hole though, just something to dream about.

tests/python-e2e/close_socket_keep_connection.py
```diff
@@ -0,0 +1,39 @@
...
+
+    initial_socket.close()
+    print("Closed socket.", file=stderr)
+
+    data = connected_socket.recv(16, socket.MSG_WAITALL)
+    connected_socket.sendall(data.upper())
+
+    connected_socket.close()
+
+    # The (Rust) test code informs this application via stdin when it can exit
```
meowjesty, line 34: We could also try to interact with the closed socket, and assert that it crashes.

But I think your approach is more well-behaved :nerd_face: .

meowjesty (changes requested):
Mostly small stuff.

I think this PR is really _close_ to being done.

tests/python-e2e/close_socket_keep_connection.py
t4lz: I don't think the application would crash in that case.

mirrord/layer/src/lib.rs
t4lz: Done here https://github.com/metalbear-co/mirrord/pull/1557/commits/e5e6226e86c6fd8d5438197de07c9356c194288f. Also changed `get_port` to `get_bound_port` (return `None` if socket state is `Connected`), so that we don't match `socket.state` twice.

mirrord/layer/src/tcp.rs
t4lz: I shouldn't have written this situation is _valid_ (removed this wording in https://github.com/metalbear-co/mirrord/pull/1557/commits/a076294bd0de3fef497ebe94015ae866a9d31523).

In some circumstances it might be a hiccup for the user - without mirrord, that new connection would have been routed to the pod, and accepted if the remote app is still listening. With mirrord - that connection is not handled by either process (local or remote), and will be reset once the agent gets the `PortUnsubscribe` message, even if a new connection a second later is accepted by the remote app.

With a lot of effort we could partially handle this situation: we could introduce a new message to the protocol that tells the agent, that this happens and also messages for sending back already forwarded bytes, then the agent could try to start a connection with the remote app and be a proxy. I think it's not worth it, but I think we should give the user some indication that we swallowed that connection.

mirrord/layer/src/tcp.rs
t4lz: Here:
```
Self::check_connection_handling_result(
	                self.handle_new_connection(tcp_connection).await,
	            )
```
`check_connection_handling_result` is called on the `Result`  of `handle_new_connection`. With `and_then` it would only be called on an `Ok`.

mirrord/layer/src/tcp.rs
t4lz: Like this? https://github.com/metalbear-co/mirrord/pull/1557/commits/7983ee5a17ba7b4911b246c619541e62ed4543b4

mirrord/layer/src/tcp_steal.rs
t4lz: What did she say about mirroring?

mirrord/layer/src/tcp.rs
meowjesty: Alright, I think this behavior is weird enough that I would also vote for a `warn` there.

mirrord/layer/src/tcp.rs
meowjesty: Yup. Thought this would require more changes, but this trait is not very "shared" I guess.

`const`fy-all the things!

tests/python-e2e/close_socket_keep_connection.py
meowjesty: Oh, you can always make the app segfault, you just have to believe tal!

mirrord/layer/src/tcp.rs
meowjesty: Oh! Yup, ... errmm ... hmm huh... it was a ... huh ... trick question! Yes, a trick question! And you have passed, congratulations ....

(sorry, my brain created a sort of `chain` method for `Result` I guess)

meowjesty (approved):
:bear: :+1:

PR #1561, opened 2023-06-14
add listen ports feature

### LINKED ISSUE #1554: Support local traffic in addition to traffic forwarded from the cluster
Hey folks, I really enjoy working with mirrord so far, its speed and flexibility is impressive so huge kudos! 👏 

One feature I find compelling is the local application instance's ability to receive both traffic from the cluster, as it is today using either `mirror` or `steal`, and also locally generated traffic (i.e. calling the application from a gRPC client).

For example, I’m developing an app that exposes a gRPC server on port 50051; traffic generated inside the remote cluster on this port it will be forwarded to my localhost, but also, I’d like to be able to send requests from my localhost to port 50051 and have them received in the locally running instance.

Right now I work around this by running the gRPC client in targetless mode which works like a charm, but if there’s a way to spare the extra step it would be even more awesome.

Conceptually, a new configuration can help here, which will allow enabling such capability, along with port mapping for cases where users will want to use different ports locally, something like-

```
"network": {
      "incoming": "steal",
      "outgoing": true,
      "local": {
        "enabled": true,
        "port_mapping": [ "8080:80", "5000:50051" ]
      }
},
...
```

Thanks!

aviramha:
Closes #1554

### DIFF
--- a/mirrord/config/src/feature/network/incoming.rs
+++ b/mirrord/config/src/feature/network/incoming.rs
@@ -56,9 +56,10 @@ use http_filter::*;
 ///           "filter": "host: api\..+",
 ///           "ports": [80, 8080]
 ///         },
-///         "port_mapping": [[ 7777: 8888 ]],
+///         "port_mapping": [[ 7777, 8888 ]],
 ///         "ignore_localhost": false,
 ///         "ignore_ports": [9999, 10000]
+///         "listen_ports": [[80, 8111]]
 ///       }
 ///     }
 ///   }
@@ -69,7 +70,7 @@ use http_filter::*;
 #[serde(untagged, rename_all = "lowercase")]
 pub enum IncomingFileConfig {
     Simple(Option<IncomingMode>),
-    Advanced(IncomingAdvancedFileConfig),
+    Advanced(Box<IncomingAdvancedFileConfig>),
 }
 
 impl Default for IncomingFileConfig {
@@ -120,6 +121,10 @@ impl MirrordConfig for IncomingFileConfig {
                     .layer(|layer| Unstable::new("IncomingFileConfig", "ignore_localhost", layer))
                     .source_value()
                     .transpose()?
+                    .unwrap_or_default(),
+                listen_ports: advanced
+                    .listen_ports
+                    .map(|m| m.into_iter().collect())
                     .unwrap_or_default(),
             },
         };
@@ -187,6 +192,22 @@ pub struct IncomingAdvancedFileConfig {
     /// Ports to ignore when mirroring/stealing traffic. Useful if you want specific ports to be
     /// used locally only.
     pub ignore_ports: Option<Vec<u16>>,
+
+    /// ### listen_ports
+    ///
+    /// Mapping for local ports to actually used local ports.
+    /// When application listens on a port while steal/mirror is active
+    /// we fallback to random ports to avoid port conflicts.
+    /// Using this configuration will always use the specified port.
+    /// If this configuration doesn't exist, mirrord will try to listen on the original port
+    /// and if it fails it will assign a random port
+    ///
+    /// This is useful when you want to access ports exposed by your service locally
+    /// For example, if you have a service that listens on port `80` and you want to access it,
+    /// you probably can't listen on `80` without sudo, so you can use `[[80, 4480]]`
+    /// then access it on `4480` while getting traffic from remote `80`.
+    /// The value of `port_mapping` doesn't affect this.
+    pub listen_ports: Option<Vec<(u16, u16)>>,
 }
 
 /// Controls the incoming TCP traffic feature.
@@ -230,6 +251,7 @@ pub struct IncomingAdvancedFileConfig {
 ///         "port_mapping": [[ 7777, 8888 ]],
 ///         "ignore_localhost": false,
 ///         "ignore_ports": [9999, 10000]
+///         "listen_ports": [[80, 8111]]
 ///       }
 ///     }
 ///   }
@@ -267,6 +289,22 @@ pub struct IncomingConfig {
 
     /// #### feature.network.incoming.filter {#feature-network-incoming-http-filter}
     pub http_filter: HttpFilterConfig,
+
+    /// #### feature.network.incoming.listen_ports {#feature-network-incoming-listen_ports}
+    ///
+    /// Mapping for local ports to actually used local ports.
+    /// When application listens on a port while steal/mirror is active
+    /// we fallback to random ports to avoid port conflicts.
+    /// Using this configuration will always use the specified port.
+    /// If this configuration doesn't exist, mirrord will try to listen on the original port
+    /// and if it fails it will assign a random port
+    ///
+    /// This is useful when you want to access ports exposed by your service locally
+    /// For example, if you have a service that listens on port `80` and you want to access it,
+    /// you probably can't listen on `80` without sudo, so you can use `[[80, 4480]]`
+    /// then access it on `4480` while getting traffic from remote `80`.
+    /// The value of `port_mapping` doesn't affect this.
+    pub listen_ports: BiMap<u16, u16>,
 }
 
 impl IncomingConfig {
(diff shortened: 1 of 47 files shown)


mirrord/config/src/feature/network/incoming.rs
```diff
@@ -56,9 +56,10 @@ use http_filter::*;
 ///           "filter": "host: api\..+",
 ///           "ports": [80, 8080]
 ///         },
-///         "port_mapping": [[ 7777: 8888 ]],
+///         "port_mapping": [[ 7777, 8888 ]],
```
meowjesty, line 59: Thank you :sweat_smile:

mirrord/config/src/feature/network/incoming.rs
```diff
@@ -267,6 +289,22 @@ pub struct IncomingConfig {
...
+    /// Using this configuration will always use the specified port.
+    /// If this configuration doesn't exist, mirrord will try to listen on the original port
+    /// and if it fails it will assign a random port
+    ///
+    /// This is useful when you want to access ports exposed by your service locally
+    /// For example, if you have a service that listens on port `80` and you want to access it,
+    /// you probably can't listen on `80` without sudo, so you can use `[[80, 4480]]`
+    /// then access it on `4480` while getting traffic from remote `80`.
+    /// The value of `port_mapping` doesn't affect this.
+    pub listen_ports: BiMap<u16, u16>,
```
meowjesty, line 307: With `BiMap` the user can't have an invalid config such as `[[10, 1234], [10, 1234]]`, right?

I'm assuming this holds true for the other `listen_ports`?

mirrord/layer/src/socket/ops.rs
```diff
@@ -191,15 +177,68 @@ pub(super) fn bind(
...
+        unsafe { FN_BIND(sockfd, unbound_address.as_ptr(), unbound_address.len()) };
     if bind_result != 0 {
-        error!(
-            "listen -> Failed `bind` sockfd {:#?} to address {:#?} with errno {:#?}!",
-            sockfd,
-            unbound_address,
-            errno::errno()
-        );
-        return Err(io::Error::last_os_error())?;
+        // If we didn't have `listen_ports` used here, just assign a random address.
```
meowjesty, line 208: You could extract this whole thing here into a shared `call_raw_bind(listen_port)`, and have a more _rusty_ block of code.

Like the code before would be something like:

```rust
let bound = do_the_bind(listen_port).or_else(|e| { 
    warn!("..."); // as you don't like `inspect_err` much
    do_the_bind(0)
})?;
```

mirrord/layer/tests/apps/listen_ports/src/main.rs
```diff
@@ -0,0 +1,20 @@
+#![warn(clippy::indexing_slicing)]
```
meowjesty, line 1: Hmm, for some reason I was under the impression that we were not using full cargo projects for integration tests (at least when avoidable due to it having no dependencies).

meowjesty (changes requested):
Overall looks good.

Thank you for fixing some of the examples.

mirrord/layer/tests/apps/listen_ports/src/main.rs
aviramha: welp, I think it's worth having just for IDE support

mirrord/config/src/feature/network/incoming.rs
aviramha: yeah

mirrord/layer/tests/apps/listen_ports/src/main.rs
meowjesty: Yup, makes sense. I was writing my test samples in a separate project thinking this was the way :crying_cat_face:

mirrord/layer/src/socket/ops.rs
```diff
@@ -191,15 +201,26 @@ pub(super) fn bind(
...
+    if let Some(port) = listen_port {
+        bind_port(sockfd, socket.domain, port)?;
+    } else {
+        bind_port(sockfd, socket.domain, requested_address.port()).or_else(|e| {
+            warn!(
+                "bind -> first `bind` failed on port {listen_port:?} with {e:?}, trying to bind to a random port"
+            );
+            bind_port(sockfd, socket.domain, 0)
+        })?
     }
```
meowjesty, line 224:
```suggestion
    if let Some(port) = listen_port {
        bind_port(sockfd, socket.domain, port)
    } else {
        bind_port(sockfd, socket.domain, requested_address.port()).or_else(|e| {
            warn!(
                "bind -> first `bind` failed on port {listen_port:?} with {e:?}, trying to bind to a random port"
            );
            bind_port(sockfd, socket.domain, 0)
        })
    }?;
```

I think this is valid Rust?

meowjesty (approved):
One final nit to rule them all!

Good luck on the tests

:bear: :+1:

mirrord/layer/src/socket/ops.rs
```diff
@@ -108,6 +108,30 @@ pub(super) fn socket(domain: c_int, type_: c_int, protocol: c_int) -> Detour<Raw
...
+    }?;
+
+    trace!("bind -> {address:#?}");
+
+    let bind_result = unsafe { FN_BIND(sockfd, address.as_ptr(), address.len()) };
+    if bind_result != 0 {
+        return Detour::Error(io::Error::last_os_error().into());
+    }
+
+    Detour::Success(())
```
meowjesty, line 132:
```suggestion
    if bind_result != 0 {
        Detour::Error(io::Error::last_os_error().into())
    } else {
        Detour::Success(())
    }
```

Considering that this is the last bit of the function, wdyt?

mirrord/layer/src/socket/ops.rs
```diff
@@ -191,16 +201,25 @@ pub(super) fn bind(
...
+        .get()
+        .expect("`LISTEN_PORTS` not set. Please report a bug")
+        .get_by_left(&requested_address.port())
+        .cloned();
+
+    // Listen port was specified
+    if let Some(port) = listen_port {
+        bind_port(sockfd, socket.domain, port)
+    } else {
+        bind_port(sockfd, socket.domain, requested_address.port()).or_else(| e | {
```
meowjesty, line 218: `| e |` Is this a (a) visual github error, or is (b) rustfmt not formatting this properly?

I think it's (b), as I think rustfmt would also put the `.or_else` in a new line, weird.

mirrord/layer/src/socket/ops.rs
```diff
@@ -226,7 +245,10 @@ pub(super) fn bind(
 
     SOCKETS.insert(sockfd, socket);
 
-    Detour::Success(bind_result)
+    // node reads errno to check if bind was successful and doesn't care about the return value
+    // (???)
+    errno::set_errno(errno::Errno(0));
```
t4lz, line 250: Nitpick: would be nicer to get the errno before we call `bind_port` and restore the original errno.

aviramha: afaik `bind_port` doesn't reset the errno and that's why we got there in the first place

t4lz: I'm not sure I understand what you're saying, but bind's supposed to only change the errno if there is an error, right?
Because now we also change it (to 0) if everything's fine, which is different than bind's behaviour, no?

If for example a user application decides to do:

1. set errno to 0.
2. do first operation that might fail and set errno.
3. bind port.
4. check errno to see if 2 or 3 failed.

In 3 we would set errno to 0 and erase the errno from 2 and the user application would lose that information.

I don't think that's a big problem, still more of a nitpick, but I want to make sure we understand each other 😃

aviramha: I know that, but Node checks errno even if success code is 0 so we have to reset it. I also mentioned in the comment explaining the above...

t4lz: Yeah, but we can reset it to the value it had before we did our detour.

t4lz: get errno before we do anything that can change errno, then restore that value when we're done.

t4lz: if it was 0 before, it's 0 after we're done.

aviramha: mm, ok not bad idea but we merged it already G:

PR #1564, opened 2023-06-15
Changed back required IntelliJ version to 222+ from 223+

aviramha:


Razz4780:
I ran `verifyPluginConfiguration` and it was successful. A bit surprising, given that I had to bump the `idea-version` before because of the same task.
I also ran `runPluginVerifier` and it was successful. Also surprising, given that I had to disable it before.
I don't know what to think

Razz4780:
```
> Task :verifyPluginConfiguration
[gradle-intellij-plugin :verifyPluginConfiguration] The following plugin configuration issues were found:
- The 'since-build' property is lower than the target IntelliJ Platform major version: 222 < 223.
```
The task still passes though, don't know what to think

Razz4780 (approved):

Issue #1457, opened 2023-05-24
nightly-2023-05-24 doesn't compile


meowjesty:
### Bug Description

Something changed between `nightly-2023-03-29` (our current version) and `nightly-2023-05-23` that breaks some of our const trait implementations.

If you try to `cargo check` with a newer version (the latest working I've tested was `nightly-04-15`, but I'm not sure if this is the last working one or something more recent also works) you'll get a `#[const_trait]` error.

### Steps to Reproduce

1. Update `rust-toolchain.toml` to `nightly-2023-05-23`;
2. Run `cargo check`;
3. sadness;

### Backtrace

```shell
error: const `impl` for trait `Default` which is not marked with `#[const_trait]`
   --> mirrord/layer/src/detour.rs:114:15
    |
114 | impl<T> const Default for HookFn<T> {
    |               ^^^^^^^
    |
    = note: marking a trait with `#[const_trait]` ensures all default method bodies are `const`
    = note: adding a non-const method body in the future would be a breaking change
```

### Your operating system and version

linux

### Local process

NA

### Additional Info

P.S. You'll probably get an error in the `protocol/file.rs` and not on the `layer` crate. In `protocol` the issue is easy to solve, we could just drop the const trait implementation, but in `layer` it requires bigger changes, due to `HookFn` expecting const stuff.

meowjesty:
[This](https://github.com/rust-lang/rust/pull/110393) broke our `Default` const implementations.

Workaround would be to have our own `Default`-like trait and implement it (and call the right function in the macros for `HookFn`).

aviramha:
mm the default breakage makes me desiring of `?const` idea

PR #1569, opened 2023-06-16
Bump rust to nightly that still supports const std traits.

meowjesty:
- Issue: #1457 

Doesn't update to the latest nightly, just a small bump to the last nightly that still had `#[const_trait]` in std traits.


aviramha (approved):

aviramha:
Thanks!

Issue #1570, opened 2023-06-17
Python doesn't work with mirrord with default config on Apple CPUs


t4lz:
### Bug Description

Python just does nothing with mirrord on a mac with M2.

We get different results when connecting to the mirrord console, in that case the process exits with code 132 (Illegal instruction).

it does work when specifying fs-mode `local-with-overrides`.

### Steps to Reproduce

`mirrord exec -t deploy/py-serv-deployment python3`

### Your operating system and version

macOS 13.4

### Local process

file $(which python3) /Users/tal/.pyenv/shims/python3: Bourne-Again shell script text executable, ASCII text

t4lz:
#1571 prevents this issue from reproducing, but it's not clear that there is no underlying issue that still exists, so we might want to leave this open.

eyalb181:
Can you close this issue (since Python does run on Apple CPUs now) and open a new one that's more descriptive of the actual problem?

t4lz:
Closing for now. If a bug still exists, we'll open an issue when we see it.

PR #1571, opened 2023-06-17
make python run on apple.

t4lz:
Prevents #1570 from happening and enables using mirrord for running python on mac, but does not solve the underlying problem with Rust/Tokio/Frida/our SIP sidestepping.

### DIFF
--- a/changelog.d/1570.changed.md
+++ b/changelog.d/1570.changed.md
@@ -1,0 +1,1 @@
+The path `/opt` itself is read locally by default (up until now paths inside that directory were read locally by default, but not the directory itself).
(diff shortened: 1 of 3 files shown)


eyalb181 (approved):

eyalb181:
@t4lz Can you elaborate re the underlying problem with Rust/Tokio/Frida/our SIP sidestepping?

t4lz:
> @t4lz Can you elaborate re the underlying problem with Rust/Tokio/Frida/our SIP sidestepping?

These are my suspects because of the `132` exit code - "illegal instruction" which indicates a problem with the binary. Rust, Frida, and our SIP code all edit binary code that ends up running.

I suspected tokio because I think the point where we get stuck is on receiving from a oneshot channel, but I now think it's not tokio's fault, the result just does not arrive because the thread that is supposed to produce it dies or something like that.

I'm still debugging to pin down exactly where things go wrong.

Issue #1582, opened 2023-06-23
pycharm debugger hangs debugging gunicorn app.


meowjesty:
### Bug Description

User was trying to debug their [gunicorn](https://gunicorn.org/) app using Pycharm with mirrord, it seems to hang.

You can reproduce this even with just the example from the gunicorn link above.

If you click `Run` in Pycharm, everything works, as `Run` doesn't use `pydev`, but clicking `Debug` does and hangs.

### Steps to Reproduce

1. Prepare a gunicorn app (`pip install gunicorn`);
2. Configure Pycharm to run a sample gunicorn app with:
  - Parameters: `--bind 0.0.0.0:80 -w 4 {app_file}:{main_function}`, e.g `--bind 0.0.0.0:80 -w 4 app:app`;
  - Environment variables: `PYTHONUNBUFFERED=1;RUST_LOG=mirrord=debug,warn`
3. Enable mirrord;
4. Use the mirrord config:

```json
{
    "accept_invalid_certificates": true,
    "feature": {
       "env": true,
       "fs": "read",
       "network": {
           "incoming": {
               "mode": "steal",
               "ignore_localhost": false
           },
           "outgoing": {
               "tcp": true,
               "ignore_localhost": false
           },
           "dns": true
       },
       "capture_error_trace": false
    }

}
```

5. Click on the `Debug` icon in Pycharm;
6. Send a request with `curl http://{service}:{port}`;
7. No response / hangs;

### Relevant Logs

- Logs from `Debug` run:

```sh
pydevd.py --multiprocess --qt-support=auto --client 127.0.0.1 --port 41231 --file /home/meowjesty/dev/metalbear/samples/issue-gunicorn-user/bin/gunicorn --bind 0.0.0.0:80 -w 4 app:app 
2023-06-23T14:39:51.211244Z  INFO ThreadId(01) mirrord_layer: Initializing mirrord-layer!
Connected to pydev debugger (build 231.9161.41)
[2023-06-23 11:39:51 -0300] [24874] [INFO] Starting gunicorn 20.1.0
2023-06-23T14:39:51.665696Z  WARN ThreadId(01) bind_port: mirrord_layer::socket::ops: new sockfd=13 domain=2 port=80
2023-06-23T14:39:51.665734Z  WARN ThreadId(01) bind_port: mirrord_layer::socket::ops: return=Error(IO(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })) sockfd=13 domain=2 port=80
[...]

PR #1587, opened 2023-06-26
Add usr and home to the default file filter.

meowjesty:
- Issue: #1582 

Adds `/usr` and `/home` to `generate_local_set` default file filter.

### DIFF
--- a/mirrord/layer/src/file/filter.rs
+++ b/mirrord/layer/src/file/filter.rs
@@ -96,6 +96,9 @@ fn generate_local_set() -> RegexSet {
         r"^/System",
         #[cfg(target_os = "macos")]
         r"^/var/run/com.apple",
+        // pycharm + gunicorn
+        r"^/usr$",
+        r"^/home$",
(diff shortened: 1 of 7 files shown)


mirrord/layer/src/file/filter.rs
```diff
@@ -96,6 +96,9 @@ fn generate_local_set() -> RegexSet {
         r"^/System",
         #[cfg(target_os = "macos")]
         r"^/var/run/com.apple",
+        // pycharm + gunicorn
+        r"^/usr$",
```
aviramha, line 100: I think it should be like other similar thingies:
```
        "^/usr(/|$)",
 ```

aviramha: and modify the existing `/usr` above then it fits both

aviramha (changes requested):

aviramha (approved):

aviramha:
Great, Thank you.

PR #1590, opened 2023-06-28
Fix some text

eyalb181:


mirrord/cli/src/config.rs
```diff
@@ -218,19 +218,19 @@ pub(super) enum OperatorCommand {
         #[arg(long)]
         accept_tos: bool,
 
-        /// License key to be stored in mirrord-operator-license secret
+        /// A mirrord for Teams license key
         #[arg(long)]
         license_key: Option<String>,
 
-        /// License key file path to be stored in mirrord-operator-license secret
+        /// Path to a file containing a mirrord for Teams license key
```
aviramha, line 225: tbh it's a certificate file now, so license certificate

eyalb181: Isn't it the same as the `license_key`? I'd rather have them both misnaming it as key, rather than having each of them calling it something else and confusing users.

aviramha: license_key is deprecated, I think @DmitryDodzin didn't remove it to allow backward compatbility.

aviramha (approved):

Issue #702, opened 2022-11-07
Support filtering IPs for outgoing traffic


eyalb181:
We should let users enable outgoing traffic, but then specify a list of IPs for which outgoing traffic would remain local.

## Motivation
We want to let users mix and match resources to be used, locally and remote.
* For example, if the process sends traffic to localhost, they can configure it to use the local machine's localhost rather than the remote pod's.
* Filtering should be based on protocol, port, host, ip and subnet.

## Implementation

`outgoing` config should support the field `remote` and `local`, each overriding the default behavior of the general setting.
The value for these fields will be a list of items with the following format:
`[PROTOCOL://]IP/[subnet]:PORT`
`HOST:PORT`
`:PORT`
`HOST`

For example:
```toml
remote = ["udp://1.1.1.0/24:1337", "1.1.5.0/24", "tcp://google.com:53", "google.com", ":53"]
local = ["localhost"]
```
will lead to the following behavior:

1. Any TCP connection to 1.1.1.0/24 subnet with destination port 1337 will go from the remote
2. Any connection to 1.1.5.0/24 subnet will go from the remote
3. Connection to google.com resolved DNS will go from the remote if TCP (need dynamic management here probably).
4. Connection to any IP with any protocol with port 53 will go from remote
5. Any connection to localhost or it's resolved IPs will be local.

## Details

This issue should solve the following use cases:

1. DNS Resolving
6. TCP connections
7. UDP

### Warnings

This feature has some sharp edges that might hurt, those edges come to mind:
* Go/Custom DNS resolving (using UDP instead of just libc's getaddrinfo etc) - harder to cover, need to do some sort of MITM for the resolving. Can be deferred for a follow up issue.
* State management - for example if we resolve `google.com` we would want to passthrough connections to it but we need to maintain some state dns <-> sockets that doesn't exist currently -> can be deferred to another issue

errietta:
This would be great to have. We have an external API we use, and we have to allow list specific IPs. So we've allowlisted our k8s cluster. I want to be able to call that API, but the code will also write to the db, and I want to write to my local db not the remote one

mentos1386:
My usecase is https://www.prisma.io/ which starts an "engine" that listens on a random port  (it seems to me). Which makes it impossible to use, as the "in cluster pod" might not be listening on exactly the same port 😦

aviramha:
I have implemented https://github.com/metalbear-co/mirrord/pull/1154 until we get this done fully.
[...]

PR #1607, opened 2023-07-03
Outgoing traffic filter.

meowjesty:
- Issue: #702 

Adds the outgoing traffic filter feature, behind the config `feature.network.outgoing.[remote|local]`.

Supports configs as specified in #702 , with the caveat that named traffic is only partially implemented in this PR. There is a parser, and we go as far as putting the unresolved addresses in the layer, but it won't properly resolve anything when `connect_remote` is called.


mirrord/config/src/feature/network/outgoing.rs
t4lz, line 1: The implementation is consistent with the specification from the issue, so it would have been better if I had commented this on the issue when it was created, but:
I think we should reconsider whether we really want to allow specifying both the remote and the local lists together. If I understand correctly, the only thing the remote list is useful for, is to override filters from the local list (e.g. specifying a whole subnet in `local`,  and then specifying a single IP from that subnet in `remote` to make every IP from that subnet except that one be connected to locally). That seems like pretty complicated semantics. I think that a simpler configuration field would do, but if we really do want to have these semantics, we should probably at least not name that field `remote`, but something that reveals its true meaning, like `except_remote` or `force_remote`, or something like that.

I think these semantics would be simpler:
Allow specifying either `local` or `remote` (not both). If `local` is specified, everything happens remotely except if it's on that list. If `remote` is specified, everything happens locally except if it's on that list.

If we allow both - specifying `remote` rarely makes sense, as everything except for what's in `local` is remote anyways. Most of the values passed in the `remote` list in the configuration examples in this PR seem redundant to me.

Am I misunderstanding the semantics of `remote`?

t4lz (commented):
I'm still in early stages of this review, but I'm flushing the comments because the discussion about the configuration is crucial for the rest of the review.

mirrord/config/src/feature/network/outgoing.rs
t4lz: @eyalb181 / @aviramha, anyone wants to weigh in on this?

mirrord/layer/tests/outgoing.rs
```diff
@@ -137,3 +145,26 @@ async fn outgoing_tcp(dylib_path: &PathBuf) {
...
+    config_dir: &PathBuf,
+) {
+    outgoing_tcp_logic(with_config, dylib_path, config_dir).await;
+}
+
+#[rstest]
+#[tokio::test]
+#[timeout(Duration::from_secs(10))]
+#[should_panic]
+async fn outgoing_tcp_local_broken(
```
t4lz, line 164: why "broken"?

t4lz (commented):
Very nice code. Thanks for the documentation and tests <3
My main comments are about possibly simplifying the parser code a bit.

mirrord/config/src/feature/network/outgoing.rs
eyalb181: Thanks for pointing this out, I missed it when I wrote the issue. Agreed of course, meaning possible values for outgoing would be `true/false/local=[]/remote=[]`

Sorry @meowjesty for the rework

t4lz: Thanks, @eyalb181. So should we deprecate the `udp` and `tcp` fields, as they are a bit redundant with the filter?

mirrord/layer/tests/outgoing.rs
t4lz: Oh, I see. Sure, you could listen for a connection in the test's code. But I'm not saying you have to, the test seems fine to me.
Either way please add documentation explaining what this test verifies and how.

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -83,6 +144,232 @@ impl MirrordToggleableConfig for OutgoingFileConfig {
...
+    /// Valid protocol types.
+    pub protocol: ProtocolFilter,
+
+    /// Address|name|subnet we're going to filter.
+    pub address: OutgoingAddress,
+}
+
+/// <!--${internal}-->
+/// It's dangerous to go alone!
+/// Take [this](https://github.com/rust-bakery/nom/blob/main/doc/choosing_a_combinator.md).
```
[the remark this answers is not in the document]
meowjesty: :crossed_swords:

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -83,6 +144,232 @@ impl MirrordToggleableConfig for OutgoingFileConfig {
...
+        let (rest, protocol) = protocol(input)?;
+        let (rest, address) = address(rest)?;
+        let (rest, subnet) = subnet(rest)?;
+        let (rest, port) = port(rest)?;
+
+        // Stringify and convert to proper types.
+        let protocol = str::from_utf8(protocol)?.parse()?;
+        let address = str::from_utf8(&address)?;
+        let subnet = subnet.map(String::from_utf8).transpose()?;
+        let port = str::from_utf8(&port)?.parse::<u16>()?;
```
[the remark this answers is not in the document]
meowjesty: I've tried, but some combinator functions wanted to deal with `&[u8]`, and hated `&str`.

Don't really recall which one it was (I think the `is_***` family of functions), or even if we're still using them, but as `&[u8]` worked better I ended up switching to it.

Considering that we're checking that the characters are valid on each parser function, we could use `from_utf8_unchecked` (which is just a `transmute`) and not lose any performance with somewhat redundant checks.

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -83,6 +144,232 @@ impl MirrordToggleableConfig for OutgoingFileConfig {
...
+    ///
+    /// 1. `name.with.dots`;
+    /// 2. `1.2.3.4.5.6`;
+    /// 3. `[dad:1337:fa57::0]`
+    ///
+    /// Where 1 and 2 are handled by `dotted_address`.
+    ///
+    /// The parser is not interested in only eating correct values here for hostnames, ip addresses,
+    /// etc., it just tries to get a good enough string that could be parsed by
+    /// [`SocketAddr::parse`], or [`IpNet::parse`].
```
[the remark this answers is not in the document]
meowjesty: Removed the doc links, but they still appear in the PR for me?

mirrord/config/src/feature/network/outgoing.rs
eyalb181: Not sure I follow, why are they redundant? Something like {local=[], udp=false} could make sense (at least as much sense as anything else with `udp=false`).

meowjesty: 1. simple outgoing enabled:

```json
{
  "outgoing": true
}
```

2. simple outgoing disabled:

```json
{
  "outgoing": false
}
```

3. outgoing with local filter (everything is remote, except what's in `local`):

```json
{
  "outgoing": 
    {
      "tcp": true,
      "udp": true,
      "local": ["localhost"]
    }
}
```

4. outgoing with remote filter (everything is local, except what's in `remote`):

```json
{
  "outgoing": 
    {
      "tcp": true,
      "udp": true,
      "remote": ["localhost"]
    }
}
```

5. Should this config even be allowed?

```json
{
  "outgoing": 
    {
      "tcp": false,
      "udp": false,
      "remote": ["localhost"]
    }
}
```

meowjesty: Forgot 6.

6. both filters specified, config error?

```json
{
  "outgoing": 
    {
      "tcp": true,
      "udp": true,
      "remote": ["localhost"],
      "local": ["foo"]
    }
}
```

mirrord/layer/tests/outgoing.rs
meowjesty: Added docs, are they clear enough though?

mirrord/config/src/feature/network/outgoing.rs
t4lz: > Should this config even be allowed?

This is why I think we can deprecate `outgoing.tcp` and `outgoing.udp`. And in the meantime they could be applied with an `AND` when specified. So we would connect remotely if the filter says so AND the relevant `tcp`/`udp` is `true` or isn't specified.

mirrord/config/src/feature/network/outgoing.rs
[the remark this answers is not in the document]
t4lz: I think the conversions from and to strings add some unnecessary noise to the code, and while adding unsafe code where we can prove it's safe is ok, not having to do the operation at all is even better. Does this work: https://github.com/meowjesty/mirrord/pull/11? it was a bit too much for a suggestion so I opened a PR from my fork to your fork 😄

t4lz (changes requested):

mirrord/config/src/feature/network/outgoing.rs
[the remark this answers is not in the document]
meowjesty: Yup, I started out with `&str`, but while messing around with `nom` found some functions that had to take char as byte. Looking into it again, they are the `is_digit` and some of the other `is_x`, which if we had to use those, we could just convert the char to a byte there.

Thanks for changing it for me, merged your PR.

t4lz (changes requested):
Some parts need to be updated to the semantics - most importantly the remote/local connection check.

t4lz (changes requested):
Some parts need to be updated to the semantics - most importantly the remote/local connection check.

mirrord/layer/src/socket.rs
```diff
@@ -254,6 +266,130 @@ impl UserSocket {
...
+                    })
+                })
+                .unwrap_or_default(),
+            AddressFilter::Subnet((subnet, port)) => {
+                subnet.contains(&address.ip()) && (port == 0 || port == address.port())
+            }
+        };
+
+        self.remote.iter().filter(filter_protocol).any(any_address)
+            || !self.local.iter().filter(filter_protocol).any(any_address)
```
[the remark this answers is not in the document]
meowjesty: Please check if I changed it to the right thing.

mirrord/layer/src/socket.rs
```diff
@@ -254,6 +266,135 @@ impl UserSocket {
...
+/// [`OutgoingSelector::new`].
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub(crate) struct OutgoingSelector {
+    /// If the address from `connect` matches this, then we send the connection through the
+    /// remote pod.
+    remote: HashSet<OutgoingFilter>,
+
+    /// If the address from `connect` matches this, then we send the connection from the local app.
+    local: HashSet<OutgoingFilter>,
+}
```
t4lz, line 279: This definition does not allow for a distinction between "not specified" and "specified to be empty".

When specifying `"local": []`everything should always be remote, but I think right now I think it would make everything be local.

The selector could have just one `HashSet` and a `bool`, or an an enum with two variants with each only 1 `HashSet`.

t4lz (changes requested):

mirrord/layer/src/socket.rs
meowjesty: Went ahead and converted it to an `enum`.

Now the semantics should be:

1. user specifies nothing: `Unfiltered`;
2. user specifies remote: `Remote([...])`;
3. user specifies local: `Local([...])`;
4. user tries to specify both: ??? crash, and it goes against the schema I think;

meowjesty: Now to update the docs and see if this actually works.

t4lz: Looks good, is there a validation that the user does not pass an empty list in the configuration though?

mirrord/layer/src/socket.rs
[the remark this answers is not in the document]
meowjesty: New way of setting the filter, this is no longer relevant.

mirrord/layer/src/socket.rs
meowjesty: There is a validation during initialization, where we set up our globals and stuff, but can we do it when parsing the config itself?

t4lz: Caller-crate-side validation is ok for now if it's easier. Can you share a link to where the non-empty-list validation happens?

meowjesty: [layer/socket/OutgoingSelector::try_from](https://github.com/metalbear-co/mirrord/pull/1607/files#diff-bf9c5f9340f82cf6357f9ea7b2d2a735776603ed1ead3daa2cd95db99c024650R324).

And in the layer lib initialization it crashes when we convert the filters [here](https://github.com/metalbear-co/mirrord/pull/1607/files#diff-478993a1def0c54eb220a1d4dcdc0ebd61c73c3d44b05ba2d89c4fef509996e0R425).

meowjesty:
```sh
    ✓ agent pod created
    ✓ pod is ready
thread '<unnamed>' panicked at 'Failed setting up outgoing traffic filter!: MissingConfigValue("Outgoing traffic filter cannot be empty!")', mirrord/layer/src/lib.rs:426:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c609da59d9fc05b1c7dc879d79700ccd8140b5fc/library/std/src/panicking.rs:578:5
```

t4lz: Nice, thanks! Definitely enough validation for now.

mirrord/layer/tests/configs/outgoing_filter.json
```diff
@@ -0,0 +1,11 @@
+{
+    "feature": {
+        "network": {
+            "outgoing": {
+                "filter": {
+                    "remote": ["tcp://1.1.1.0/24:1111", "2.2.2.2:2222"]
```
t4lz, line 6: Why does the test pass if in the list of peers there is also `3.3.3.3:3333`, but it's not in the filter?

meowjesty: Makes no sense, I'm investigating.

Running it as an app, mirrord behaves as expected, where if it's not in the `remote` list, it doesn't go through the agent.

mirrord/layer/tests/outgoing.rs
```diff
@@ -88,12 +88,20 @@ async fn outgoing_udp(dylib_path: &PathBuf) {
...
+    let config = with_config
+        .and_then(|config| {
+            let mut config_path = config_dir.clone();
+            config_path.push(config);
+            Some(config_path)
+        })
+        .unwrap_or_default();
+
+    let config = config.to_str().unwrap();
+    let config = (!config.is_empty()).then_some(config);
```
t4lz, line 101: Isn't this the same?
```suggestion
    let config = with_config
        .map(|config| {
            let mut config_path = config_dir.clone();
            config_path.push(config);
            config_path
        })
        .map(|path_buf| path_buf.to_str().unwrap());
```

t4lz: updated this suggestion.

mirrord/layer/tests/outgoing.rs
t4lz, line 1: This file is set to conditionally compile only on linux for some reason. If when removing that attribute the other tests already pass on macOS, please just remove that attributes and let all the tests run on macOS as well. If the existing tests don't pass, that's out of the scope of this PR, but please make sure at least the new tests are executed also on macOS (and pass).

t4lz (changes requested):

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -1,17 +1,75 @@
...
+#[derive(Deserialize, PartialEq, Eq, Clone, Debug, JsonSchema)]
+#[serde(untagged, rename_all = "lowercase")]
+pub enum OutgoingFilterConfig {
+    /// Traffic that matches what's specified here will go through the remote pod, everything else
+    /// will go through local.
+    Remote(VecOrSingle<String>),
+
+    /// Traffic that matches what's specified here will go through the local app, everything else
+    /// will go through the remote pod.
+    Local(VecOrSingle<String>),
```
t4lz, line 63: This config uses `VecOrSingle`, which I think expects a semicolon-delimited string (`"stuff;more stuff"`), but in all of the documentation, example and tests, you use lists (`["stuff", "more stuff"]`). That probably doesn't work, right? Maybe check what the parsed configuration ends up being with those configs?
If as I suspect `VecOrSingle` does not parse lists, then either change this to something that does (probably just `Vec`?) or change the docs, examples and tests - whatever's quicker to do.

mirrord/layer/tests/outgoing.rs
meowjesty: Removed this `#[cfg]`, now let's see if the test works (with the fixed config).

meowjesty: On macos:

```sh
thread 'outgoing_tcp::with_config_2_Some___outgoing_filter_json___' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', mirrord/layer/tests/common/mod.rs:124:14
```

On linux it's passing (at least)!

t4lz: > On macos:
> 
> ```shell
> thread 'outgoing_tcp::with_config_2_Some___outgoing_filter_json___' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', mirrord/layer/tests/common/mod.rs:124:14
> ```
> 
> On linux it's passing (at least)!

Just need to build the test app on macOS too.

mirrord/config/src/feature/network/outgoing.rs
meowjesty: Had to remove the `untagged` attribute, now the schema recognizes it as `filter: anyOf(object, object)` (something like that), which is what we wanted, where the `object`s are `remote` or `local`.

The list works with `,`, and I think it also supports custom separators (when you want to concat into string).

mirrord/layer/tests/outgoing.rs
meowjesty: Both passing!

I'm adding another config to the test that expects a `panic`, where we don't put the `3333` address, so it should go local and hang.

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -49,6 +55,61 @@ pub struct OutgoingConfig {
     #[config(unstable, default = false)]
     pub ignore_localhost: bool,
 
+    /// #### feature.network.outgoing.remote {#feature.network.outgoing.remote}
+    ///
+    /// List of addresses/ports/subnets that should be sent through the remote pod.
```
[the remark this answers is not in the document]
meowjesty: This is not relevant anymore, now you can only set up either, not both.

The text saying that "only the ..." is now in the samples of the config.

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -49,6 +55,61 @@ pub struct OutgoingConfig {
...
+    /// }
+    /// ```
+    /// 
+    /// Valid values follow this pattern: `[protocol]://[name|address|subnet/mask]:[port]`.
+    #[config(default)]
+    pub remote: HashSet<String>,
+
+    /// #### feature.network.outgoing.local {#feature.network.outgoing.local}
+    ///
+    /// List of addresses/ports/subnets that should be sent through the local app.
```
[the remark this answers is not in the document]
meowjesty: Also not relevant anymore?

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -1,17 +1,75 @@
...
+/// List of addresses/ports/subnets that should be sent through either the remote pod or local app,
+/// depending how you set this up with either `remote` or `local`.
+///
+/// You may use this option to specify when outgoing traffic is sent from the remote pod (which
+/// is the default behavior when you enable outgoing traffic), or from the local app (default when
+/// you have outgoing traffic disabled).
+///
+/// Takes a list of values, such as:
+///
+/// - Only UDP traffic on subnet `1.1.1.0/24` on port 1337 will go throuh the remote pod.
```
t4lz, line 24:
```suggestion
/// - Only UDP traffic on subnet `1.1.1.0/24` on port 1337 will go through the remote pod.
```

t4lz (approved):
Congratulations, you did it!

Issue #1615, opened 2023-07-04
Thin out and document list of capabilities needed for the agent to function properly


eyalb181:
Instead of having the agent run as a privileged pod, we need to find the minimal list of capabilities it needs to run and then use that.

PR #1616, opened 2023-07-04
Run agent as a non-privileged container

Razz4780:
Solves #1615 

I found that `CAP_NET_ADMIN`, `CAP_SYS_PTRACE` and `CAP_SYS_ADMIN` is a minimal set allowing the agent to function. E2E tests pass and removing any of these capabilities causes them to fail.

* `CAP_NET_ADMIN` - modifying ip tables
* `CAP_SYS_PTRACE` - reading `/proc/<target>/environ`
* `CAP_SYS_ADMIN` - entering target's network namespace

Sadly, `CAP_SYS_ADMIN` is a very broad capability. AFAIK it is used by kernel developers too often. Quote from "Note to developers": "It can plausibly be called the new root"

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,9 @@ You can read more about it [here](https://mirrord.dev/docs/overview/introduction
 ### Additional capabilities
 
 Container run inside the pod launched by mirrord requires additional [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html):
-* `CAP_NET_ADMIN` - for modifying routing tables
-* `CAP_SYS_PTRACE` - for reading target pod environment
-* `CAP_SYS_ADMIN` - for joining target pod network namespace
+- `CAP_NET_ADMIN` - for modifying routing tables
+- `CAP_SYS_PTRACE` - for reading target pod environment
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -118,11 +119,18 @@ mirrord exec node app.js --target pod/my-pod
...
 
-When you select a pod to impersonate, mirrord launches a privileged pod on the same node as the pod you selected.
+When you select a pod to impersonate, mirrord launches a pod on the same node as the pod you selected.
 The new pod is then used to connect your local process and the impersonated pod: it mirrors incoming traffic from the pod to your process,
 routes outgoing traffic from your process through the pod, and does the same for file reads, file writes, and environment variables.
 You can read more about it [here](https://mirrord.dev/docs/overview/introduction/).
 
+### Additional capabilities
+
+Container run inside the pod launched by mirrord requires additional [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html):
```
aviramha, line 129: Great, this needs to be documented in mirrord.dev as well, with reference to mirrord for Teams (i.e if you don't want to have users allowed to create such pods, the operator can take care of it etc)

aviramha (approved):

PR #1625, opened 2023-07-06
Have the same fs-mode everywhere.

t4lz:
We override the fs mode on targetless to be `localwithoverrides` regardless of user config.
However, it was only being done for the `FILE_FILTER` global. This PR changes it to be consistent everywhere (also in `FILE_MODE` and in the config object passed to the layer task).


meowjesty (approved):
:bear: :+1:

PR #1637, opened 2023-07-10
Double slash in uri path is fixed

aohoy:

### DIFF
--- a/mirrord/agent/src/runtime/crio.rs
+++ b/mirrord/agent/src/runtime/crio.rs
@@ -47,7 +47,7 @@ impl CriOContainer {
                 Request::builder()
                     .method("GET")
                     .header("Host", "localhost")
-                    .uri(format!("http://localhost/{}", path))
+                    .uri(format!("http://localhost/{}", path.to_string().trim_start_matches('/')))
                     .body(Empty::<Bytes>::new())?,
             )
             .await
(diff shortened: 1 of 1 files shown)


mirrord/agent/src/runtime/crio.rs
```diff
@@ -47,7 +47,7 @@ impl CriOContainer {
                 Request::builder()
                     .method("GET")
                     .header("Host", "localhost")
-                    .uri(format!("http://localhost/{}", path))
+                    .uri(format!("http://localhost/{}", path.to_string().trim_start_matches('/')))
```
DmitryDodzin, line 50: The `api_get` is a private function so since you have changed the paths it can get I wouldn't do a second removal of `/`, and in this scenario `to_string` is completely unnecessary because the resulting `String` is being derefed back down to str for `trim_start_matches`.

Changing to `format!("http://localhost{path}")` and keeping the paths `/pause/{}` and `/unpause/{}` should also be fine because because `api_get` will probably remain private for it's existence

DmitryDodzin (changes requested):
Thank you very much =]

I've left one comment mostly to keep the the code as simple as possible

mirrord/agent/src/runtime/crio.rs
aohoy: Fixed

DmitryDodzin (approved):

aviramha:
Added to MQ. Thank you!

PR #1639, opened 2023-07-10
E2E: Operator setup with --dry-run

DmitryDodzin:
Add test to check the validity the yaml from `mirrord operator setup`


aviramha:
I think this would be a redundant test when we add also setup of operator then using it?

eyalb181:
> I think this would be a redundant test when we add also setup of operator then using it?

Yeah but it would be a good place to start from when we do the full tests

eyalb181 (approved):

PR #1645, opened 2023-07-11
Remove Operator state PVC

DmitryDodzin:


changelog.d/+remove-operator-pvc.internal.md
```diff
@@ -0,0 +1 @@
+Remove operator pvc from setup & dau/mau counters from `mirrord operator status`
```
aviramha, line 1: I think the part after the `&` can be removed :p

aviramha (approved):

Issue #1560, opened 2023-06-14
Alert user when running on OpenShift


eyalb181:
When we detect that mirrord is being run on OpenShift, warn the user that issues might occur and direct them to the documentation (see metalbear-co/mirrord.dev#87). Make sure the alerts show in the IDE as well.

infiniteregrets:
blocked by #1664

aviramha:
Not blocked anymore.

PR #1654, opened 2023-07-12
Add warning when using openshift

infiniteregrets:
closes https://github.com/metalbear-co/mirrord/issues/1560


mirrord/cli/src/connection.rs
```diff
@@ -83,6 +83,8 @@ where
         let k8s_api = KubernetesAPI::create(config)
             .await
             .map_err(CliError::KubernetesApiFailed)?;
+
+        k8s_api.detect_openshift().await?;
```
aviramha, line 87: We shouldn't exit on this error - just log in info level that we failed trying to detect.

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -56,6 +56,18 @@ impl KubernetesAPI {
...
         }
     }
+
+    pub async fn detect_openshift(&self) -> Result<()> {
+        Discovery::new(self.client.clone())
+            .run()
+            .await
+            .map(|discovery| {
+                if discovery.has_group("route.openshift.io") {
+                    eprintln!("mirrord is running on openshift, due to default PSP of openshift, mirrord may not be able to create the agent. Please refer to https://mirrord.dev/docs/overview/faq/");
```
aviramha, line 66: use the task api?

mirrord/cli/src/connection.rs
aviramha: we should also do it only if we don't use operator.

aviramha (changes requested):

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -57,6 +57,23 @@ impl KubernetesAPI {
...
+    where
+        P: Progress + Send + Sync,
+    {
+        let mut sub_progress = progress.subtask("checking for openshift");
+        let _= Discovery::new(self.client.clone())
+            .run()
+            .await
+            .map(|api_discovery| {
+                if api_discovery.has_group("route.openshift.io") {
+                    sub_progress.success(Some("mirrord is running on openshift, due to default PSP of openshift, mirrord may not be able to create the agent. Please refer to https://mirrord.dev/docs/overview/faq/"));
```
meowjesty, line 71: Is this a `success`? It kinda reads like a warning?

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -57,6 +57,23 @@ impl KubernetesAPI {
...
+                } else {
+                    sub_progress.success(None);
+                }
+        });
```
meowjesty, line 75:
```suggestion
        if let Some(api_discovery) = Discovery::new(self.client.clone()).run().await && api_discovery.has_group("route.openshift.io")  {
                 sub_progress.success(Some("mirrord is running on openshift, due to default PSP of openshift, mirrord may not be able to create the agent. Please refer to https://mirrord.dev/docs/overview/faq/"));
        };
```

Isn't this just an `if let` as the `None` case ends up being just a noop?

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -57,6 +57,23 @@ impl KubernetesAPI {
             target,
         }
     }
+
+    pub async fn detect_openshift<P>(&self, progress: &P) -> ()
```
meowjesty, line 61: Do we need to return the `()`?

Have not looked into the progress stuff, so my bad if this is a requirement.

meowjesty (changes requested):
Just some questions.

mirrord/kube/src/api/kubernetes.rs
infiniteregrets: Oh my bad, i didnt push the updated branch

mirrord/kube/src/api/kubernetes.rs
infiniteregrets: It returns an Option, I am wondering if we shouldnt eat the error when its not found?

eyalb181 (changes requested):

eyalb181 (approved):

mirrord/cli/src/connection.rs
```diff
@@ -91,6 +91,8 @@ where
             .await
             .map_err(CliError::KubernetesApiFailed)?;
 
+        k8s_api.detect_openshift(progress).await?;
```
aviramha, line 94: Create a new task for detecting openshift (progress)

1. When openshift is detected, use the warning api to leave a message about the possible issues
2. When openshift isn't detected, end the progress with "openshift not detected"
3. When you fail detecting, (i.e `detect_openshift` returns an err) use warning API to let user know check wasn't done "couldn't determine if OpenShift"

aviramha (changes requested):

aviramha (approved):

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -57,6 +57,20 @@ impl KubernetesAPI {
...
+    pub async fn detect_openshift<P>(&self, progress: &P) -> Result<()>
+    where
+        P: Progress + Send + Sync,
+    {
+        if Discovery::new(self.client.clone())
+            .run()
+            .await?
+            .has_group("route.openshift.io")
+        {
+            progress.warning("mirrord is running on openshift, due to default PSP of openshift, mirrord may not be able to create the agent. Please refer to https://mirrord.dev/docs/overview/faq/#can-i-use-mirrord-with-openshift");
```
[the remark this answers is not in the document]
eyalb181: This seems to have been reverted @infiniteregrets

meowjesty (approved):
:bear: :+1: 

But there are some clippy warnings though.

PR #1660, opened 2023-07-13
Add `readinessProbe` and `livenessProbe` to Operator Deployment

DmitryDodzin:


mirrord/operator/src/setup.rs
```diff
@@ -314,6 +314,18 @@ impl OperatorDeployment {
             });
         }
 
+        let health_probe = Probe {
```
aviramha, line 317: you can just use https probe..

DmitryDodzin: There is no https probe, grpc and http only

mirrord/operator/src/setup.rs
```diff
@@ -314,6 +314,17 @@ impl OperatorDeployment {
...
+            http_get: Some(HTTPGetAction {
+                path: Some("/heath".to_owned()),
```
aviramha, line 319: `health`

aviramha (approved):

PR #1667, opened 2023-07-14
Reduce amount of API calls to use operator

aviramha:


mirrord/operator/src/client.rs
```diff
@@ -13,15 +13,16 @@ use mirrord_kube::{
...
 use tokio::sync::mpsc::{self, Receiver, Sender};
 use tokio_tungstenite::tungstenite::{Error as TungsteniteError, Message};
-use tracing::{error, trace, warn};
+use tracing::error;
 
 use crate::crd::{MirrordOperatorCrd, TargetCrd, OPERATOR_STATUS_NAME};
 
 static CONNECTION_CHANNEL_SIZE: usize = 1000;
-static MIRRORD_OPERATOR_SESSION_ID: &str = "MIRRORD_OPERATOR_SESSION_ID";
+static MIRRORD_OPERATOR_SESSION: &str = "MIRRORD_OPERATOR_SESSION";
```
t4lz, line 25: nitpick - could be `const`, right? Up to you.

mirrord/operator/src/client.rs
```diff
@@ -57,11 +60,52 @@ pub struct OperatorApi {
...
+            session_id: rand::random::<u64>().to_string(),
+            target,
+            fingerprint,
+        }
+    }
+
+    /// Returns environment variable holding the information
+    pub fn env_key() -> &'static str {
+        MIRRORD_OPERATOR_SESSION
+    }
```
t4lz, line 83: if you make the var a const, you can also make this function `const`.

mirrord/operator/src/client.rs
```diff
@@ -88,35 +132,27 @@ impl OperatorApi {
...
-                id
-            })
+    /// Connect to session using operator and session information
+    pub async fn connect(
+        config: &LayerConfig,
+        session_information: &OperatorSessionInformation,
+    ) -> Result<(mpsc::Sender<ClientMessage>, mpsc::Receiver<DaemonMessage>)> {
+        let operator_api = OperatorApi::new(config).await?;
+        let (sender, receiver) = operator_api.connect_target(&session_information).await?;
+        Ok((sender, receiver))
```
t4lz, line 155:
```suggestion
        operator_api.connect_target(&session_information).await
```

t4lz (approved):

mirrord/operator/src/client.rs
```diff
@@ -45,29 +46,77 @@ pub enum OperatorApiError {
...
+                    .map_err(|e| OperatorApiError::InvalidOperatorSession(val, e))
+            })
+            .transpose()
+    }
+}
+
 impl OperatorApi {
-    pub async fn discover<P>(
+    /// Creates a new operator session, setting the session information in environment variables.
+    pub async fn create_session<P>(
```
DmitryDodzin, line 106: I'm not sure renaming it good because it can maybe create a session if operator exists otherwise it will return none, discover was fine IMO

mirrord/operator/src/client.rs
```diff
@@ -45,29 +46,77 @@ pub enum OperatorApiError {
...
+        Self {
+            session_id: rand::random::<u64>().to_string(),
+            target,
+            fingerprint,
+            operator_features,
+        }
+    }
+
+    /// Returns environment variable holding the information
+    pub const fn env_key() -> &'static str {
```
DmitryDodzin, line 88: I need to remember to use these more instead of importing consts everywhere 👍

DmitryDodzin (approved):
Nitpick =]

mirrord/operator/src/client.rs
aviramha: well we can split the API into three, but I think that attempting to create session if operator exists is ok

PR #1675, opened 2023-07-17
Fixed dynamic pause with operator not working

aviramha:
- moved pause request to be from internal proxy
- Fixed case where proxy can timeout since it holds a stale connection. Added heartbeat to the connection handling

### DIFF
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -145,7 +145,7 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
     // Create a main connection, that will be held until proxy is closed.
     // This will guarantee agent staying alive and will enable us to
     // make the agent close on last connection close immediately (will help in tests)
-    let (main_connection, operator) = connect_and_ping(&config).await?;
+    let (mut main_connection, operator) = connect_and_ping(&config).await?;
     if config.pause {
         tokio::time::timeout(
             Duration::from_secs(config.agent.communication_timeout.unwrap_or(30).into()),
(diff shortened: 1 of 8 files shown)


mirrord/cli/src/internal_proxy.rs
```diff
@@ -114,7 +145,19 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
     // Create a main connection, that will be held until proxy is closed.
     // This will guarantee agent staying alive and will enable us to
     // make the agent close on last connection close immediately (will help in tests)
-    let (_main_connection, operator) = connect_and_ping(&config).await?;
+    let (main_connection, operator) = connect_and_ping(&config).await?;
```
DmitryDodzin: I fear after some time `main_connection` dies because there are no messages sent via it (maybe we should add a timer that sends pings while the proxy is still alive)

mirrord/kube/src/api.rs
```diff
@@ -81,6 +84,8 @@ pub fn wrap_raw_connection(
                     }
                 }
                 daemon_message = codec.next() => {
+                    timeout_check.reset();
```
DmitryDodzin, line 87: Shouldn't this be only on DaemonMessage::Pong not on all requests (in case agent got stuck in some loop and is only sending nocensene and doesn't actually respond to Ping with Pong)

aviramha: I don't think we need such validation - the ping/pong is to make sure connection is alive, not for making sure it makes senes

DmitryDodzin (approved):

PR #1684, opened 2023-07-18
Fixed ping mechanism not working

DmitryDodzin:
Partial revert of #1675

### DIFF
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -25,10 +25,11 @@ use mirrord_protocol::{pause::DaemonPauseTarget, ClientMessage, DaemonCodec, Dae
 use tokio::{
     net::{TcpListener, TcpStream},
     select,
-    sync::mpsc::{self},
-    task::JoinSet,
+    sync::mpsc,
+    task::{JoinHandle, JoinSet},
     time::timeout,
 };
+use tokio_util::sync::CancellationToken;
 use tracing::{error, info, log::trace};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


mirrord/cli/src/internal_proxy.rs
```diff
@@ -159,6 +160,9 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
         })??;
     }
 
+    let (main_connection_cancalation_token, main_connection_task_join) =
```
Razz4780, line 163:
```suggestion
    let (main_connection_cancellation_token, main_connection_task_join) =
```
Plus more like this below :p

mirrord/cli/src/internal_proxy.rs
```diff
@@ -204,6 +211,11 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
         )
         .await;
     }
+
+    if let Ok(Err(err)) = main_connection_task_join.await {
```
Razz4780, line 215: What if the task panics?

Razz4780 (commented):
Why do we need extra pings in the internal proxy? If the pinging mechanism is correct in the layer, we should not need it. Or am I wrong

DmitryDodzin:
> Why do we need extra pings in the internal proxy? If the pinging mechanism is correct in the layer, we should not need it. Or am I wrong

Extra pings because this connection is left dangling as to keep the agent alive for the duration of the internal proxy execution so operator hard closes this connection due to inactivity

mirrord/cli/src/internal_proxy.rs
```diff
@@ -233,6 +249,43 @@ async fn ping(
     }
 }
 
+fn create_ping_loop(
+    mut connection: (mpsc::Sender<ClientMessage>, mpsc::Receiver<DaemonMessage>),
+) -> (
+    CancellationToken,
+    JoinHandle<Result<(), InternalProxyError>>,
+) {
+    let cancalation_token = CancellationToken::new();
```
Razz4780, line 258: Some `cancalation`s left :p

Razz4780 (changes requested):

Razz4780 (approved):

PR #1693, opened 2023-07-20
Taints and Tolerations

### LINKED ISSUE #1692: Allow tolerations to be configurable
First discussed [here](https://discord.com/channels/933706914808889356/950020257156833340/threads/1131518307120447509)

Our cluster uses `kyverno` which provides an admission controller that rejects the agent pod with the following message:

`Pods may not use tolerations which schedule on control plane nodes`

The node taint is a simple control plane taint:

```
  taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
```

Could you allow the agent tolerations to be configurable, or at least toggleable, so that users can customise appropriately?

Thanks!

DmitryDodzin:
closes #1692


aviramha:
This won't solve the problem because the default toleration exists. We should probably make it as part of the default toleration unless user specifies their own. (So if they specify an explicit empty list, that means no tolerations, which is what user wants)

mirrord-schema.json
```diff
@@ -287,6 +287,17 @@
           "format": "uint64",
           "minimum": 0.0
         },
+        "tolerations": {
+          "title": "agent.tolerations {#agent-tolerations}",
+          "description": "Add extra pod tolerations. (not with ephemeral agents)",
```
aviramha, line 292: Set tolerations (overrides default tolerate everything, set to an empty array to have no tolerations at all)

aviramha (approved):

PR #1700, opened 2023-07-20
Fixed flakes caused by stdout/stderr not being flushed

aviramha:
Fixed flakes caused by stdout/stderr not being flushed after process is done


eyalb181 (approved):

mirrord/cli/src/main.rs
```diff
@@ -404,12 +404,17 @@ async fn main() -> miette::Result<()> {
...
+    match commands {
+        Commands::ListTargets(_) | Commands::ExtensionExec(_) => {
+            mirrord_progress::init_from_env(ProgressMode::Json);
+            let _ = miette::set_hook(Box::new(|_| Box::new(JSONReportHandler::new())));
+            true
+        }
+        Commands::InternalProxy(_) => {
+            // don't have logs since it breaks when we close stdout/err
+            true
+        }
```
infiniteregrets, line 416: not sure if I understand what you mean here

mirrord/layer/tests/fileops.rs
```diff
@@ -629,13 +629,11 @@ async fn read_go(
...
-            Pid::from_raw(process.id().unwrap() as pid_t),
-            Signal::SIGTERM,
-        )
-        .unwrap()
-    });
+    signal::kill(
+        Pid::from_raw(test_process.child.id().unwrap() as pid_t),
+        Signal::SIGTERM,
+    )
+    .unwrap();
```
infiniteregrets, line 636: is this just a better way to write or is it supposed to make a change the way we kill a process?

tests/src/pause.rs
```diff
@@ -1,5 +1,7 @@
 #[cfg(test)]
 
+/// The pause tests use static names, so they can't be run at the same time.
+/// We use `serial_test` to run them one after the other.
 mod pause {
```
infiniteregrets, line 5: static names as in? why cant they run in parallel?

mirrord/layer/tests/fileops.rs
aviramha: I changed the underlying struct, it's supposed to do the same.

mirrord/layer/tests/sip.rs
```diff
@@ -35,7 +37,27 @@ async fn tmp_dir_read_locally(dylib_path: &Path) {
     let mut layer_connection = LayerConnection::get_initialized_connection(&listener).await;
     println!("Application subscribed to port, sending tcp messages.");
 
-    layer_connection.expect_gethostname(1).await;
+    layer_connection
+        .expect_file_open_for_reading("/etc/hostname", 1)
+        .await;
```
infiniteregrets, line 42: so we dont need the `expect_gethostname` anymore? what was the issue on this one?

aviramha: Sometimes we don't get the close message because the process exits before we manage to send the message.

infiniteregrets: got it

infiniteregrets (approved):
lgtm, just not sure why pause tests cant run in parallel

aviramha:
> lgtm, just not sure why pause tests cant run in parallel

Did you read the new comment?

infiniteregrets:
> > lgtm, just not sure why pause tests cant run in parallel
> 
> Did you read the new comment?

oh i see you updated it, thanks that clears it

PR #1708, opened 2023-07-24
Add option to skip build tools [defaults: true]

### LINKED ISSUE #1478: `go run .` flaky crash - fatal: bad g in signal handler
on macOS m1, Golang 1.20.4 (not signed! - signed won't work because of #1477):

run
`mirrord exec --steal -t pod/pod -- bash`

then inside the bash (happens with playground app for example)
`go run .`

results sometimes in 
`fatal: bad g in signal handler`

DmitryDodzin:
closes #1478

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -94,6 +94,14 @@
         }
       ]
     },
+    "skip_build_tools": {
+      "title": "skip_build_tools {#root-skip_build_tools}",
+      "description": "Automatically skip mouting mirrord on build tools",
+      "type": [
+        "boolean",
+        "null"
+      ]
+    },
     "skip_processes": {
       "title": "skip_processes {#root-skip_processes}",
       "description": "Allows mirrord to skip unwanted processes.\n\nUseful when process A spawns process B, and the user wants mirrord to operate only on process B. Accepts a single value, or multiple values separated by `;`.\n\n```json { \"skip_processes\": \"bash;node\" } ```",
--- a/mirrord/layer/src/load.rs
+++ b/mirrord/layer/src/load.rs
@@ -1,6 +1,29 @@
 use mirrord_config::{util::VecOrSingle, LayerConfig};
 use tracing::trace;
 
+mod dev_tools {
+    use std::{collections::HashSet, sync::LazyLock};
+
+    static DEV_TOOLS_PROCESSES: LazyLock<HashSet<&str>> = LazyLock::new(|| {
+        HashSet::from([
+            "cc",
+            "ld",
+            "go",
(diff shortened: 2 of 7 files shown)


mirrord-schema.json
```diff
@@ -94,6 +94,14 @@
         }
       ]
     },
+    "skip_build_tools": {
+      "title": "skip_build_tools {#root-skip_build_tools}",
+      "description": "Automatically skip mouting mirrord on build tools",
```
eyalb181, line 99:
```suggestion
      "description": "Allows mirrord to skip build tools. Useful when running command lines that build and run the application in a single command.",
```

eyalb181: Also can you document the default list?

mirrord/layer/src/load.rs
```diff
@@ -1,18 +1,44 @@
...
+            "math",
+            "cargo",
+            "hpack",
+            "rustc",
+            "compile",
+            "cargo-watch",
+        ])
+    });
+
+    pub fn is_dev_tool(given_process: &str) -> bool {
```
eyalb181, line 23: Why don't you just use BUILD_TOOL_PROCESSES.contains directly? Also why is it called is_dev_tool and not is_build_tool?

eyalb181 (changes requested):

mirrord/layer/src/load.rs
DmitryDodzin: The only one I didn't change I figured dev-tools is a bit incorrect

mirrord/layer/src/macros.rs
```diff
@@ -124,7 +124,7 @@ macro_rules! graceful_exit {
             nix::sys::signal::Signal::SIGTERM,
         )
         .expect("unable to graceful exit");
-        panic!()
+        std::process::exit(0)
```
Razz4780, line 127: This is called when we can't kill the current process. Shouldn't we return a non-zero exit code?

mirrord/layer/src/load.rs
```diff
@@ -1,18 +1,40 @@
...
+        "air",
+        "asm",
+        "math",
+        "cargo",
+        "hpack",
+        "rustc",
+        "compile",
+        "cargo-watch",
+    ])
+});
```
Razz4780, line 20: Some other tools we skip in the VS Code extension: `[dlv, debugserver, asm, cgo, link, git, gcc, as, collect2, cc1]`
Maybe it's worth it to include these as well

eyalb181 (approved):

Razz4780 (approved):

Issue #1716, opened 2023-07-24
Internal proxy resets connection from layer when running shell with mirrord and piping input


t4lz:
### Bug Description

When running a shell (was observed with both sh and zsh) that reads some commands from stdin with mirrord the internal proxy sometimes resets an incoming connection from a layer and an error is displayed. 

### Steps to Reproduce

```
echo "nvm use" | MIRRORD_PROGRESS_MODE=off mirrord exec -f .mirrord/mirrord.json zsh -- -is
```

where `.mirrord/mirrord.json` is:
```
{
    "accept_invalid_certificates": true,
    "target": {
        "path": {
            "deployment": "dotenv"
        }
    },
    "feature": {
        "network": {
            "incoming": "mirror",
            "outgoing": true
        },
        "fs": {
            "mode": "localwithoverrides",
            "read_only": ".*/.env"
        },
        "env": true
    }
}
```

Also reproduced with just `echo "nvm use" | MIRRORD_PROGRESS_MODE=off mirrord exec zsh -- -is`, but less often.

### Relevant Logs

```
2023-07-24T20:16:49.964783Z ERROR ThreadId(09) mirrord_layer::connection: Couldn't connect to internal proxy: Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" }, 127.0.0.1:51068                                                                                      
```
### Your operating system and version

macOS 13.4

### Local process

zsh

### Additional Info

The error never reproduced with `trace` level logs.

Running zsh with that stdin results in **hundreds** of `execve`s, each of them results in the initialization of a new layer, each of those layers connects to the internal proxy (most of them soon disconnect as their process is replaced by the next executed process).

So one thing I'm looking into is whether the internal proxy is somehow overwhelmed by that crazy chain of process execusions.

aviramha:
Now getting connectionrefused from internal proxy

t4lz:
#1724 did not fix this. It seems like the intproxy does not die, but sends a RST to one request, but then keeps accepting other requests, after some connections close, so we think maybe it reaches some limit of connections (and running a shell with mirrord triggers it because it executes hundreds of processes). It seems like setting `ulimit -n 1024` makes the issue happen less often, but maybe it was random.
There is also https://github.com/rust-lang/rust/issues/55614

[...]

PR #1724, opened 2023-07-26
Send internal proxy child process stderr to /dev/null

t4lz:
Might help with #1716.


mirrord/cli/src/internal_proxy.rs
```diff
@@ -47,6 +46,8 @@ fn print_port(listener: &TcpListener) -> Result<()> {
         .map_err(InternalProxyError::LocalPortError)?
         .port();
     println!("{port}\n");
+    // "leak" the `Gag` in order to redirect stdout to /dev/null for the whole run.
+    std::mem::forget(Gag::stdout());
```
aviramha, line 50: Gag still holds stdout since it's meant for temporary redirection, which means we don't close stdout which means it might make caller process still wait for it's stdout to finish.

aviramha: I would copy daemonize code that does the dup - afaik if you use dup2 it closes the original fd

> If the file descriptor newfd was previously open, it is closed
       before being reused; the close is performed silently (i.e., any
       errors during the close are not reported by dup2()).

So you actually do `dup2(dev_null_fd, stdout)`

aviramha: Also I would still do it for stdin and stderr even though the caller does `/dev/null` just to be on the safe side.

mirrord/cli/src/internal_proxy.rs
```diff
@@ -47,6 +52,9 @@ fn print_port(listener: &TcpListener) -> Result<()> {
         .map_err(InternalProxyError::LocalPortError)?
         .port();
     println!("{port}\n");
+    unsafe {
+        redirect_stdout_to_dev_null();
```
aviramha, line 56: I would still do it for stdin/err also.

t4lz: done

mirrord/cli/src/internal_proxy.rs
```diff
@@ -138,6 +146,11 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
     // This makes the process not to receive signals from the "mirrord" process or it's parent
     // terminal fixes some side effects such as https://github.com/metalbear-co/mirrord/issues/1232
     nix::unistd::setsid().map_err(InternalProxyError::SetSidError)?;
```
aviramha, line 148: imo move all the "daemonizing" stuff to be done after printing the port for two reasons:
1. Have it at one place
2. Have it only after the proxy had launched successfully (so we might get errors out)

Besides that, I think we can remove the patch I added for not tracing on internal_proxy since now we can send traces it will just go nowhere (then when we use console or other thingy it'd also work)

aviramha:
Great job!

PR #1726, opened 2023-07-27
Reusable workflow for e2e tests

infiniteregrets:


.github/workflows/reusable-e2e.yaml
```diff
@@ -0,0 +1,191 @@
...
+        type: boolean
+
+jobs:
+  operator_e2e:    
+    if: ${{inputs.operator}}
+    runs-on: ubuntu-latest
+    name: e2e
+    env:
+      MIRRORD_AGENT_RUST_LOG: "warn,mirrord=debug"
+      MIRRORD_OPERATOR_IMAGE: "test"
```
DmitryDodzin, line 17: Wouldn't this conflict with agent image tag?

.github/workflows/reusable-e2e.yaml
```diff
@@ -0,0 +1,191 @@
...
+          path: /tmp
+      - name: start minikube
+        uses: medyagh/setup-minikube@master
+        with:
+          container-runtime: containerd
+      - run: minikube image load /tmp/test-operator.tar
+      - run: minikube image list
+      - name: setup operator
+        run: |
+          target/debug/mirrord operator setup --accept-tos --license-path ./tests/test-license.pem | kubectl apply -f -
```
DmitryDodzin, line 45: Need to patch the deployment to use a different agent image (by default it will use by crate semver) via MIRRORD_AGENT_IMAGE on the deployment like 

Something like
```bash
kubectl -n mirrord patch deployment mirrord-operator -p '{"spec":{"template":{"spec":{"containers":[{"name":"mirrord-operator","env": [{"name":"MIRRORD_AGENT_IMAGE","value":"test"}]}]}}}}'
```

.github/workflows/reusable-e2e.yaml
```diff
@@ -0,0 +1,191 @@
...
+        run: cargo test -p tests targetless
+      - name: Run all E2E test
+        run: cargo test -p tests -- --test-threads=6
+      - name: Collect logs
+        if: ${{ failure() }}
+        run: |
+          kubectl describe pods
+          docker exec minikube find /var/log/pods -print -exec cat {} \;
+
+  mirrord_e2e:
```
DmitryDodzin, line 122: Maybe it's better to split this entire step into 2 steps one for mirrord build and another for e2e and leave the creation of the cluster for the "reuser" (on second step assume you have a cluster running already)

DmitryDodzin (changes requested):
I think need to split the actions to build and test and maybe don't add the actions/checkout at all and just rely on being in the right tag already (just for simplicity)

infiniteregrets:
we will be using a simple script instead

PR #1743, opened 2023-08-01
Add support for selecting Kubeconfig context 

### LINKED ISSUE #1735: Provide a `--context` option for the `exec` command
In addition to the "sticky" `config use-context` command, `kubectl` provides a `--context` option to allow run-time overrides. This is especially useful in multi-cluster environments to enable quick, temporary switching between clusters. It is also invaluable for use in scripting.

An equivalent option in `mirrord exec` would be equally helpful and provide a nice symmetry with `kubectl`. Thank you!

aviramha:
Closes #1735

### DIFF
--- a/mirrord/config/src/lib.rs
+++ b/mirrord/config/src/lib.rs
@@ -287,6 +287,19 @@ pub struct LayerConfig {
     /// [For more information](https://github.com/metalbear-co/mirrord/blob/main/TELEMETRY.md)
     #[config(env = "MIRRORD_TELEMETRY", default = true)]
     pub telemetry: bool,
+
+    /// ## kube_context {#root-kube_context}
+    ///
+    /// Kube context to use from the kubeconfig file.
+    /// Will use current context if not specified.
+    ///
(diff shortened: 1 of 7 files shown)


mirrord/config/src/lib.rs
```diff
@@ -287,6 +287,19 @@ pub struct LayerConfig {
     /// [For more information](https://github.com/metalbear-co/mirrord/blob/main/TELEMETRY.md)
     #[config(env = "MIRRORD_TELEMETRY", default = true)]
     pub telemetry: bool,
+
+    /// ## kube_context {#root-kube_context}
+    ///
+    /// Kube context to use from the kubeconfig file.
+    /// Will use current context if not specified.
+    ///
+    /// ```json
```
meowjesty, line 296: Also update the global example that contains all the config.

meowjesty (changes requested):
Missing the `kube_context` in the global example.

eyalb181 (approved):

Issue #1745, opened 2023-08-02
Python debug fork error


aviramha:
on macOS 
When debugging in PyCharm, with gunicorn the worker crashes with the following message:
```
objc[45306]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[45306]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

aviramha:
Can be mitigated by `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`

aviramha:
This was found out while investigating #1740
For some reason it happens for user only when debugging two services, maybe internal proxy load causes a bit of delay triggering the race?

aviramha:
Interesting literature - https://bugs.python.org/issue33725

aviramha:
I think that the issue is that the app is single threaded, then mirrord makes it multi-threaded and then it causes it to trigger it because if the original app was multi threaded and did those calls that would've been problematic.
I think it's okay if we just add in mirrord to add this env always on macOS. The drawbacks is that in case there's a real issue/race possible, it won't trigger and crash (leading to undefined behavior)

PR #1746, opened 2023-08-02
Attempt to fix fork issue on macOS by avoiding access to CF on fork

aviramha:
Attempt on fixing #1745 (really not sure it would fix it, since it's racy)

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -293,7 +293,7 @@ pub(crate) fn is_debugger_port(addr: &SocketAddr) -> bool {
         .contains(addr)
 }
 
-/// Loads mirrord configuration and applies [`nix_devbox_patch`] patches.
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/lib.rs
```diff
@@ -272,35 +273,17 @@ pub(crate) static DEBUGGER_IGNORED_PORTS: OnceLock<DebuggerPorts> = OnceLock::ne
...
-fn is_nix_or_devbox() -> bool {
-    if let Ok(res) = std::env::var("IN_NIX_SHELL") && res.as_str() == "1" {
-        true
-    }
-    else if let Ok(res) = std::env::var("DEVBOX_SHELL_ENABLED") && res.as_str() == "1" {
-        true
-    } else {
-        false
-    }
-}
```
t4lz, line 303: Why can this be removed now?

t4lz (approved):

PR #1757, opened 2023-08-03
Try to fix 1716 - internal proxy under load

### LINKED ISSUE #1716: Internal proxy resets connection from layer when running shell with mirrord and piping input

aviramha:
Tries to fix #1716 

Try to resolve an issue where internal proxy is under heavy load since bash scripts does a lot of fork/exec by:
1. Increasing internal proxy's listen backlog (might not help on macOS)
2. Change internal proxy to create the upstream (agent) connection in a different task, allowing it to keep accepting.

I also tried
Not blocking the process we load into until we get the connection - that happens in the background. - That lead to e2e/integration failing, and we don't really need it.

### DIFF
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -31,10 +31,7 @@ use tokio::{
     time::timeout,
 };
 use tokio_util::sync::CancellationToken;
-use tracing::{
-    error, info,
-    log::{debug, trace},
-};
+use tracing::{error, info, log::trace};
 
 use crate::{
     config::InternalProxyArgs,
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -191,6 +191,10 @@ fn create_listen_socket() -> Result<TcpListener, InternalProxyError> {
         .listen(1024)
         .map_err(InternalProxyError::ListenError)?;
 
+    socket
+        .set_nonblocking(true)
+        .map_err(InternalProxyError::ListenError)?;
+
     // socket2 -> std -> tokio
     TcpListener::from_std(socket.into()).map_err(InternalProxyError::ListenError)
 }
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -58,7 +58,7 @@ unsafe fn detach_io() -> Result<()> {
         // best effort
         let _ = std::io::stdout().lock().flush();
     }
-    for fd in [libc::STDIN_FILENO, libc::STDOUT_FILENO] {
+    for fd in [libc::STDIN_FILENO, libc::STDOUT_FILENO, libc::STDERR_FILENO] {
         redirect_fd_to_dev_null(fd);
     }
     Ok(())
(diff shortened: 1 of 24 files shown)


aviramha:
> Looks good. Let's also make the layer -> proxy connection lazy. I'll open a separate issue for that.

I tried making it semi-lazy (have it connect in the background) and it failed the tests, okay with having a backlog issue for that.

mirrord/cli/src/internal_proxy.rs
```diff
@@ -205,8 +230,13 @@ pub(crate) async fn proxy(args: InternalProxyArgs) -> Result<()> {
             res = listener.accept() => {
                 match res {
                     Ok((stream, _)) => {
-                        let (agent_connection, _) = connect_and_ping(&config).await?;
-                        active_connections.spawn(connection_task(stream, agent_connection));
+                        let config = config.clone();
+                        active_connections.spawn(async move {
+                            match connect_and_ping(&config).await {
+                                Ok((agent_connection, _)) => connection_task(stream, agent_connection).await,
+                                Err(err) => error!("connectino to agent failed {err:#?}")
```
t4lz, line 237:
```suggestion
                                Err(err) => error!("connection to agent failed {err:#?}")
```

t4lz:
The pause test also fails locally, when run alone. So it seems this introduces (or uncovers) a real issue with pause or with its test. Working to find the issue.

aviramha:
The issue it introduced was me forgetting to set `nonblocking` on the self-created socket before converting it into tokio's :(

eyalb181 (approved):

PR #1742, opened 2023-08-01
Framed Http Request/Response

DmitryDodzin:
closes #1731

PR #1760, opened 2023-08-03
Protocol Features

DmitryDodzin:
Add features to protocol to allow limiting for backward/forwards compatibility

required for #1742

### DIFF
--- a/mirrord/agent/src/cli.rs
+++ b/mirrord/agent/src/cli.rs
@@ -4,6 +4,7 @@ use clap::{
     error::{Error, ErrorKind},
     Parser,
 };
+use mirrord_protocol::features::Features;
 
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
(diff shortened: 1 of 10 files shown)


mirrord/agent/src/cli.rs
```diff
@@ -36,6 +37,9 @@ pub struct Args {
     #[arg(short = 'p', long, default_value_t = false)]
     pub pause: bool,
 
+    #[arg(long, default_value_t = Features::all())]
+    pub protocol_features: Features,
```
aviramha, line 41: Why we would want users to specify that?
Also, why wouldn't it just be v1/v2?

DmitryDodzin: It's optional to clamp the agent to a certain version so we can clamp it from the operator if needed

aviramha: but why is it features and not just a version?

mirrord/protocol/src/features.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl FromStr for Features {
+    type Err = bitflags::parser::ParseError;
+
+    fn from_str(flags: &str) -> Result<Self, Self::Err> {
+        Ok(Self(flags.parse()?))
+    }
+}
+
+thread_local!(
```
aviramha, line 28: This is ugly, what can we do about it?
I'd prefer it to be a OnceLock/LazyCell - no need for thread local..

PR #1762, opened 2023-08-04
Fix detecting skipped processes

### LINKED ISSUE #1725: `go` stopped working in VScode devcontainer after a few hours of running mirrord.
### Bug Description

User reported:

After working for several hours, stealing traffic from a remote container, debugging locally a GO service (Intel Mac, VScode devcontainer), suddendly hit this:

```sh
Build Error: go build -o /workspaces/the-x/__debug_bin2795526562 -gcflags all=-N -l ./main.go
# command-line-arguments
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: cannot find /lib64/ld-linux-x86-64.so.2: No such file or directory
collect2: error: ld returned 1 exit status (exit status 1)
 ```

Only rebuilding the devcontainer from scratch, worked...

Intel Mac, Go is 1.20, I'm using the vscode ext. so I assume mirrord version is fetched from latest?

This happened only once. The only possible way to recover was rebuilding the devcontainer.

It was bunch of trial & errors, running in debug, stopping, re-running, stopping again untill I hit this.

[Discord thread](https://discord.com/channels/933706914808889356/1133862304837030028)

### Steps to Reproduce

1. Run `go` project with mirrord in `steal` mode a bunch of times;
2. After a few hours of this flow, `go` stopped working;

### Backtrace

_No response_

### Relevant Logs

_No response_

Intel Mac, Go is 1.20

### Local process

go build -o /workspaces/the-x/__debug_bin2795526562 -gcflags all=-N -l ./main.go

### Local process version

_No response_

### Additional Info

_No response_

### PR DESCRIPTION
**Pull request by Razz4780**
Closes #1725 

Using mirrord with a go project was failing even without the devcontainer, because the layer was injecting into common build tools like `gcc` or `ld`. This is a regression introduced recently.

Layer now inspects both path to the executable name and the process arguments when determining load type.

### DIFF
--- a/mirrord/layer/tests/skipping_process.rs
+++ b/mirrord/layer/tests/skipping_process.rs
@@ -80,7 +80,7 @@ async fn dont_skip(dylib_path: &PathBuf) {
         .await;
 
     let mut conn = LayerConnection::get_initialized_connection(&listener).await;
-    conn.expect_file_open_for_reading("/etc/hostname", 5).await;
+    conn.expect_file_open_for_reading("/etc/resolv.conf", 5).await;
 
     test_process.wait_assert_success().await;
 }
--- a/mirrord/layer/tests/skipping_process.rs
+++ b/mirrord/layer/tests/skipping_process.rs
@@ -1,0 +1,86 @@
+#![feature(assert_matches)]
+#![warn(clippy::indexing_slicing)]
+
+use std::{env::temp_dir, os::unix::fs, path::PathBuf, time::Duration};
+
+use rand::prelude::*;
+use rstest::rstest;
+
+mod common;
+pub use common::*;
+
+/// Creates a new [`Application`], which is a symlink to the existing executable.
+/// The link is created in the temporary directory and its name is randomized.
+async fn symlink_app(app: &Application) -> Application {
+    let path = temp_dir().join(format!(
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by aviramha** [comment]
Can you add some integration tests so we won't have this regression again?

**Comment by Razz4780** [comment]
> Can you add some integration tests so we won't have this regression again?

Ok ok

**Comment by aviramha** [comment]
I think the failure now on macOS is just due to macOS having it's own magic on executable name so your symlink doesn't work..?
I wouldn't delve too much into it, can disable that test on macOS saying it doesn't work but doesn't matter really

File: mirrord/layer/tests/skipping_process.rs [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+        .get_test_process_and_listener(dylib_path, vec![("MIRRORD_SKIP_PROCESSES", &ignore)], None)
+        .await;
+
+    test_process.wait_assert_success().await;
+}
+
+#[rstest]
+#[tokio::test]
+#[timeout(Duration::from_secs(20))]
+async fn skip_based_on_invokation_name(dylib_path: &PathBuf) {
```
aviramha, line 51: invocation*

aviramha (approved):

PR #1769, opened 2023-08-07
Change operator / cli version mismatch warning

aviramha:
Change operator / cli version mismatch to show only when mirrord is older than operator

### DIFF
--- a/mirrord/operator/src/client.rs
+++ b/mirrord/operator/src/client.rs
@@ -126,18 +126,14 @@ impl OperatorApi {
             // This is printed multiple times when the local process forks. Can be solved by e.g.
             // propagating an env var, don't think it's worth the extra complexity though
             let mirrord_version = Version::parse(env!("CARGO_PKG_VERSION")).unwrap();
-            if operator_version != mirrord_version {
+            if operator_version > mirrord_version {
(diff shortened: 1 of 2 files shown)


t4lz (approved):

mirrord/operator/src/client.rs
```diff
@@ -126,18 +126,14 @@ impl OperatorApi {
...
-                            "Consider updating your mirrord plugin/CLI to match the operator version.",
-                        ),
-                    );
-                } else {
-                    progress.subtask("comparing versions").print_message(MessageKind::Warning, Some("Consider either updating your operator version to match your mirrord plugin/CLI version, or downgrading your mirrord plugin/CLI."));
-                }
+                progress.subtask("comparing versions").print_message(
+                    MessageKind::Warning,
+                    Some(
+                        "Consider updating your mirrord plugin/CLI to match the operator version.",
```
t4lz, line 134: Maybe we should display the operator version here, so that users update exactly to that version and not newer (if they have the possibility, with homebrew maybe).

aviramha: I think the main use case should be always upgrade to latest

PR #1777, opened 2023-08-08
Add missing hook for `read$NOCANCEL`

aviramha:
Add missing hook for `read$NOCANCEL`, fixes reading remote files in some scenarios.

### DIFF
--- a/mirrord/layer/src/file/hooks.rs
+++ b/mirrord/layer/src/file/hooks.rs
@@ -57,7 +57,7 @@ fn update_ptr_from_bypass(ptr: *const c_char, bypass: Bypass) -> *const c_char {
 
 /// Implementation of open_detour, used in open_detour and openat_detour
 /// We ignore mode in case we don't bypass the call.
-#[tracing::instrument(level = "trace")]
+#[tracing::instrument(level = "trace", ret)]
 unsafe fn open_logic(raw_path: *const c_char, open_flags: c_int, mode: c_int) -> Detour<RawFd> {
     let path = raw_path.checked_into();
     let open_options = OpenOptionsInternalExt::from_flags(open_flags);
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/file/hooks.rs
```diff
@@ -336,6 +336,31 @@ pub(crate) unsafe extern "C" fn _read_nocancel_detour(
         .unwrap_or_bypass_with(|_| FN__READ_NOCANCEL(fd, out_buffer, count))
 }
 
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn read_nocancel_detour(
```
t4lz, line 340: Is this the exact same code as `_read_nocancel_detour`? Do we actually still need `_read$NOCANCEL`? Does it really exist? If we do - can't we use the same detour function for both hooks?

t4lz (approved):

mirrord/layer/src/file/hooks.rs
aviramha: Not sure about the first question and we can't share because one might not resolve so the other would fail

t4lz: > one might not resolve so the other would fail

What does this mean? How does a detour function _resolve_?
My suggestion is to replace both symbols (with and without a preceding `_`) with the same detour function. Meaning jump from both libc functions to the same mirrord function. Can you explain what would fail there?

aviramha: In casw of a bypass, it would try to call the original. If the original is the other function and it doesn't exist it would fail

t4lz: Ahhh, I see, makes sense. Thanks!

Issue #1776, opened 2023-08-08
Mongodb C# client fails to connect


infiniteregrets:
### Bug Description

mongodb fails to connect as it makes a raw dns query using sendmsg and we dont have a hook for it in place, solution is to use our flow for send_to by extracting fields from the msghdr

### Steps to Reproduce

to reproduce run:
- dotnet new console
- dotnet add package MongoDB.Driver

program:

```

using MongoDB.Driver;
using MongoDB.Bson;

const string connectionUri = "mongodb+srv://mehula:bar@cluster0.qewwzlp.mongodb.net/?retryWrites=true&w=majority";

var settings = MongoClientSettings.FromConnectionString(connectionUri);

// Set the ServerApi field of the settings object to Stable API version 1
settings.ServerApi = new ServerApi(ServerApiVersion.V1);

// Create a new client and connect to the server
var client = new MongoClient(settings);

// Send a ping to confirm a successful connection
try {
  var result = client.GetDatabase("admin").RunCommand<BsonDocument>(new BsonDocument("ping", 1));
  Console.WriteLine("Pinged your deployment. You successfully connected to MongoDB!");
} catch (Exception ex) {
  Console.WriteLine(ex);
}
```

### Backtrace

```shell
DnsClient.DnsResponseException: Query 59662 => cluster0.qewwzlp.mongodb.net IN TXT on 10.96.0.10:53 timed out or is a transient error.
 ---> System.Net.Sockets.SocketException (60): Operation timed out
   at System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
   at DnsClient.DnsUdpMessageHandler.Query(IPEndPoint server, DnsRequestMessage request, TimeSpan timeout)
   at DnsClient.LookupClient.ResolveQuery(IReadOnlyList`1 servers, DnsQuerySettings settings, DnsMessageHandler handler, DnsRequestMessage request, LookupClientAudit audit)
   --- End of inner exception stack trace ---
   at DnsClient.LookupClient.ResolveQuery(IReadOnlyList`1 servers, DnsQuerySettings settings, DnsMessageHandler handler, DnsRequestMessage request, LookupClientAudit audit)
   at DnsClient.LookupClient.QueryInternal(DnsQuestion question, DnsQuerySettings queryOptions, IReadOnlyCollection`1 servers)
   at DnsClient.LookupClient.Query(DnsQuestion question)
[...]

PR #1778, opened 2023-08-08
Adds sendmsg and recvmsg hooks mainly for DNS resolving (#1776).

meowjesty:
- Issue: #1776 

The C# mongo client tries to resolve DNS with [`sendmsg`](https://man.freebsd.org/cgi/man.cgi?query=sendmsg&apropos=0&sektion=2&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html) and [`recvmsg`](https://man.freebsd.org/cgi/man.cgi?query=recvmsg&apropos=0&sektion=2&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html) which we were not hooking.

Now both are hooked, and work similar PR #1520 , where they only handle traffic on port `53` (it's tailor made for DNS resolution).

If the destination is on port `53`, we create an interceptor socket, and connect to it, otherwise we just perform a fake connection (change mirrord socket state, not actual socket).


mirrord/layer/src/socket/hooks.rs
```diff
@@ -370,6 +371,128 @@ pub(super) unsafe extern "C" fn send_to_detour(
...
     }
 }
+
+#[hook_guard_fn]
+pub(super) unsafe extern "C" fn recvmsg_detour(
+    sockfd: i32,
+    message_header: *mut libc::msghdr,
+    flags: c_int,
+) -> ssize_t {
+    // Returns `-1`, but calling the original function sets whatever `errno` for us.
```
aviramha, line 381: I think you can just call recvmsg on the socket, then just modify the msg headers.. no?

mirrord/layer/src/socket/hooks.rs
```diff
@@ -370,6 +371,128 @@ pub(super) unsafe extern "C" fn send_to_detour(
...
+                .reduce(|acc, _| {
+                    let block = data_blocks.byte_add(offset);
+                    offset += (*data_blocks).iov_len;
+
+                    acc + (*block).iov_len
+                })
+                .unwrap_or_default();
+            // TODO(alex) [mid] 2023-08-08: Do we need this `total_length` check though? Real use
+            // example will tell us the answer.
+            let layout = Layout::array::<u8>(total_length).unwrap();
```
aviramha, line 401: You can use more high level API for interacting with the cmsg.
https://docs.rs/nix/0.26.2/nix/sys/socket/enum.ControlMessageOwned.html

aviramha (changes requested):

mirrord/layer/src/socket/hooks.rs
meowjesty: Nice, had no idea about this.

mirrord/layer/src/socket/hooks.rs
```diff
@@ -370,6 +371,62 @@ pub(super) unsafe extern "C" fn send_to_detour(
...
+    flags: c_int,
+) -> ssize_t {
+    debug!("recvmsg_detour -> fd {sockfd:#?}");
+    let recvmsg_result = FN_RECVMSG(sockfd, message_header, flags);
+
+    if recvmsg_result == -1 && errno::errno() != errno::Errno(libc::EAGAIN) {
+        recvmsg_result
+    } else {
+        if errno::errno() == errno::Errno(libc::EAGAIN) {
+            debug!("we hit EAGAIN");
```
aviramha, line 391: trace or remove?

mirrord/layer/src/socket/ops.rs
```diff
@@ -431,7 +433,7 @@ fn connect_to_local_address(
 /// that will be handled by `(Tcp|Udp)OutgoingHandler`, starting the request interception procedure.
 ///
 /// 3. `sockt.state` is `Bound`: part of the tcp mirror feature.
-#[tracing::instrument(level = "trace", ret, skip(raw_address))]
```
aviramha, line 434: trace?

mirrord/layer/src/socket/ops.rs
```diff
@@ -699,7 +701,7 @@ pub(super) fn dup<const SWITCH_MAP: bool>(fd: c_int, dup_fd: i32) -> Result<(),
 ///
 /// `-layer` sends a request to `-agent` asking for the `-agent`'s list of `addrinfo`s (remote call
 /// for the equivalent of this function).
-#[tracing::instrument(level = "trace", ret)]
```
aviramha, line 702: trace?

aviramha (approved):
Nice, small nitpicks

mirrord/layer/src/socket/hooks.rs
meowjesty: For now, it seems like we're safe to ignore this stuff.

mirrord/layer/src/socket/hooks.rs
meowjesty: trace all the things!

PR #1780, opened 2023-08-09
Print agent version in the initial message

Razz4780:
Needed for
* https://github.com/metalbear-co/mirrord-vscode/issues/50
* https://github.com/metalbear-co/mirrord-intellij/issues/79

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1073,9 +1073,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.81"
+version = "1.0.82"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


aviramha:
I think this wrongly affects the operator as well? @DmitryDodzin wdyt?

DmitryDodzin:
I think in general it means you can't use older cli with newer images, I wouldn't touch the agent-ready message, maybe just add another message beforehand if we want to keep the log-parsing code

Razz4780:
> I think in general it means you can't use older cli with newer images, I wouldn't touch the agent-ready message, maybe just add another message beforehand if we want to keep the log-parsing code

I believe this change is backwards-compatible. Previous CLI version used `line.contains("agent ready")`. Adding a suffix should not be a problem

DmitryDodzin (approved):

PR #1783, opened 2023-08-12
incoming config supports off

aviramha:


mirrord/config/src/feature/network/incoming.rs
```diff
@@ -385,6 +386,12 @@ pub enum IncomingMode {
     /// data on a port is HTTP (in a best-effort kind of way, not guaranteed to be HTTP), and
     /// steals the traffic on the port if it is HTTP;
     Steal,
+
+    /// <!--${internal}-->
```
eyalb181, line 390: Don't we want this to be available to users?

mirrord/config/src/feature/network/incoming.rs
```diff
@@ -397,10 +404,11 @@ impl FromStr for IncomingMode {
     fn from_str(val: &str) -> Result<Self, Self::Err> {
         match val.parse::<bool>() {
             Ok(true) => Ok(Self::Steal),
-            Ok(false) => Ok(Self::Mirror),
+            Ok(false) => Ok(Self::Off),
```
eyalb181, line 407: When would this be used? Couldn't it possibly break existing configurations?

mirrord/layer/src/lib.rs
```diff
@@ -221,20 +221,13 @@ pub(crate) static REMOTE_UNIX_STREAMS: OnceLock<Option<RegexSet>> = OnceLock::ne
...
 pub(crate) static OUTGOING_IGNORE_LOCALHOST: OnceLock<bool> = OnceLock::new();
 
-/// Tells us if the user enabled wants to ignore listening on localhost in [`IncomingConfig`].
-///
-/// ## Usage
-///
-/// When true, localhost connections will stay local - wont mirror or steal.
-pub(crate) static INCOMING_IGNORE_LOCALHOST: OnceLock<bool> = OnceLock::new();
+/// Incoming config [`IncomingConfig`].
+pub(crate) static INCOMING_CONFIG: OnceLock<IncomingConfig> = OnceLock::new();
```
eyalb181, line 225: Why'd you remove ignore_localhost/ports?

mirrord/config/src/feature/network/incoming.rs
aviramha: See the other modes, they all have it so I assumed that's on purpose.

mirrord/config/src/feature/network/incoming.rs
aviramha: If someone had `"incoming": false` it'll change the behavior but I believe it makes more sense.

mirrord/layer/src/lib.rs
aviramha: Instead of having a configuration variable for each I just have one variable (`INCOMING_CONFIG`) that has all of those fields.

mirrord/config/src/feature/network/incoming.rs
eyalb181: What if they don't have a config file, or they have a config file without the `incoming` section?

aviramha: It still mirrors (it's considered disabled config, which still defaults to mirror)

eyalb181 (approved):

PR #1784, opened 2023-08-13
Add check to prevent OutOfPods error on agent job

### LINKED ISSUE #1782: Add a check for outofpods before creating agent
There are situations where we spawn an agent and the target pod is running on a node that reached the limit of running pods.
We can check this in advance and show a better error.

DmitryDodzin:
closes #1782


mirrord/kube/src/api/runtime.rs
```diff
@@ -101,6 +107,52 @@ impl RuntimeData {
...
+            .and_then(|status| status.allocatable)
+            .ok_or_else(|| KubeApiError::NodeBadAllocatable(self.node_name.clone()))?;
+
+        let allowed: u32 = allocatable
+            .get("pods")
+            .and_then(|Quantity(quantity)| quantity.parse().ok())
+            .ok_or_else(|| KubeApiError::NodeBadAllocatable(self.node_name.clone()))?;
+
+        let mut pod_count = 0;
+        let mut list_params = ListParams {
```
aviramha, line 128: `kubectl describe node N` returns the count of pods, so I assume it is available also in the get?

mirrord/kube/src/api/runtime.rs
```diff
@@ -101,6 +107,52 @@ impl RuntimeData {
...
+            field_selector: Some(format!("spec.nodeName={}", self.node_name)),
+            ..Default::default()
+        };
+
+        loop {
+            let pods_on_node = pod_api.list(&list_params).await?;
```
aviramha, line 134: I believe this might fail and in this case should just issue a warning and continue.

mirrord/kube/src/api/runtime.rs
```diff
@@ -101,6 +107,52 @@ impl RuntimeData {
             container_name,
         })
     }
+
+    pub async fn check_node(&self, client: &kube::Client) -> Result<()> {
```
aviramha, line 111: going with my idea that this should be optional (can fail due to permissions and would be ok) maybe make it `Result<bool>` for whether check was even done?

mirrord/kube/src/error.rs
```diff
@@ -72,4 +72,10 @@ pub enum KubeApiError {
 
     #[error("mirrord-layer: Failed to wait for agent startup")]
     AgentReadyMessageMissing,
+
+    #[error("Unable to fetch node limits for {0:?}")]
+    NodeBadAllocatable(String),
+
+    #[error("Node {0:?} is too full ({1}/{2}) pods")]
+    NodePodsFull(String, u32, u32),
```
aviramha, line 80: `NodePodLimitExceeded` ?

mirrord/kube/src/api/runtime.rs
DmitryDodzin: Updated with a custom type

mirrord/config/src/agent.rs
```diff
@@ -214,6 +214,14 @@ pub struct AgentConfig {
     /// Set to an empty array to have no tolerations at all
     pub tolerations: Option<Vec<Toleration>>,
 
+    /// ### agent.check_out_of_pods {#agent-check_out_of_pods}
+    ///
+    /// Determine if to check whether there is room for agent job in target node.
+    /// Can be disabled if the check takes too long and you are sure there is enough resources on
+    /// each node
```
aviramha, line 221: Add a note that it's relevant only if ephemeral isn't used.

aviramha (approved):

mirrord/config/src/agent.rs
```diff
@@ -216,7 +216,9 @@ pub struct AgentConfig {
 
     /// ### agent.check_out_of_pods {#agent-check_out_of_pods}
     ///
-    /// Determine if to check whether there is room for agent job in target node.
+    /// Determine if to check whether there is room for agent job in target node.  (not with
```
aviramha, line 219: I'd rephrase it to
`Not applicable when using ephemeral containers feature`

PR #1786, opened 2023-08-15
Add support for Linux Go on arm

### LINKED ISSUE #563: Support for linux/arm64
I would like to run `mirrord` in a Docker container.
For Apple silicon, this requires arm64 support.

aviramha:
Closes #563

### DIFF
--- a/mirrord/layer/src/go/linux_aarch64.rs
+++ b/mirrord/layer/src/go/linux_aarch64.rs
@@ -1,0 +1,302 @@
+use std::arch::asm;
+
+use errno::errno;
+use tracing::trace;
+
+use crate::{
+    close_detour, file::hooks::*, hooks::HookManager, macros::hook_symbol, socket::hooks::*,
+    FILE_MODE,
+};
+
+
+/// [Naked function] hook for Syscall6
+#[naked]
+unsafe extern "C" fn go_syscall6_detour() {
+    asm!(
+        "nop",
+        options(noreturn)
+    );
+}
+
+/// [Naked function] hook for Syscall
+#[naked]
+unsafe extern "C" fn go_syscall_detour() {
+    asm!(
+        "nop",
+        options(noreturn)
+    );
+}
+
+/// [Naked function] maps to gasave_systemstack_switch, called by asmcgocall.abi0
+#[no_mangle]
+#[naked]
+unsafe extern "C" fn gosave_systemstack_switch() {
+    asm!(
+        "nop",
+        options(noreturn)
+    );
+}
+
+/// [Naked function] maps to runtime.abort.abi0, called by `gosave_systemstack_switch`
+#[no_mangle]
+#[naked]
+unsafe extern "C" fn go_runtime_abort() {
+    asm!("nop", options(noreturn));
+}
+
+/// Syscall & Rawsyscall handler - supports upto 4 params, used for socket,
+/// bind, listen, and accept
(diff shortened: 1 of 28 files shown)


mirrord/layer/src/hooks.rs
```diff
@@ -101,12 +101,18 @@ impl<'a> HookManager<'a> {
     ) -> Result<NativePointer> {
         // This can't fail
         let module = self.modules.first().unwrap().clone();
+        // for symbol in Module::enumerate_symbols(&module) {
+        //     trace!("found symbol {}", symbol.name);
+        // }
+
```
t4lz, line 107:
```suggestion
```

mirrord/layer/src/hooks.rs
```diff
@@ -101,12 +101,18 @@ impl<'a> HookManager<'a> {
...
         self.hook_symbol(&module, symbol, detour)
     }
 
     /// Resolve symbol in main module
-    #[cfg(target_os = "linux")]
-    #[cfg(target_arch = "x86_64")]
+    #[cfg(all(
+        target_os = "linux",
+        any(target_arch = "x86_64", target_arch = "aarch64")
```
t4lz, line 114: Do we need this arch selection (also in all other locations)?

mirrord/layer/src/lib.rs
```diff
@@ -448,6 +460,21 @@ fn layer_start(mut config: LayerConfig) {
...
+
+    // Disable all features that require the agent
+    if trace_only {
+        config.feature.fs.mode = FsModeConfig::Local;
+        config.feature.network.dns = false;
+        config.feature.network.incoming.mode = IncomingMode::Off;
+        config.feature.network.outgoing.tcp = false;
+        config.feature.network.outgoing.udp = false;
+    }
+
```
t4lz, line 477: We can consider extracting this into a function (`fn set_trace_only(mut config: LayerConfig) -> bool`), in order to start making the `layer_start` function a bit shorter.

mirrord/layer/src/go/linux_aarch64.rs
```diff
@@ -0,0 +1,303 @@
...
+        "1:",
+        "ret",
+        options(noreturn)
+    );
+}
+
+/// Syscall & Syscall6 handler - supports upto 6 params, mainly used for
+/// accept4 Note: Depending on success/failure Syscall may or may not call this handler
+#[no_mangle]
+unsafe extern "C" fn c_abi_syscall6_handler(
```
t4lz, line 37: If there's a way to share this code between the architectures instead of duplicating it that would be nice.

t4lz (changes requested):
Awesome. Only required change is deleting the couple of lines of commented out code - everything else is up to you.

mirrord/layer/src/lib.rs
aviramha: I'd like to make it a real "feature" later on, so maybe do it then?

mirrord/layer/src/hooks.rs
aviramha: well, in the case someone tries to compile to mips or something we didn't implement.. doesn't hurt

t4lz (approved):

mirrord/layer/src/go/linux_aarch64.rs
aviramha: ah right, I wanted to do it but forgot.

PR #1796, opened 2023-08-15
Move e2e setup to a bash script

infiniteregrets:
We have to move these to a script such that this file can also be used for the operator to complete the setup.

- NVM for node
- rustup for cargo/rust
- GVM for golang


.github/workflows/ci.yaml
```diff
@@ -458,48 +458,9 @@ jobs:
...
-      - run: |
-          cd tests/rust-e2e-fileops
-          cargo build
-      - run: |
-          cd tests/rust-unix-socket-client
-          cargo build
-      - run: |
-          cd tests/rust-bypassed-unix-socket
-          cargo build
+          chmod +x ./scripts/setup_e2e.sh
```
DmitryDodzin, line 462: I think you can do it with `git update-index --chmod=+x` beforehand on the file so it will be saved as an executable

DmitryDodzin (approved):
Nitpick but I feel like integration tests can also benefit from this change 👍

PR #1800, opened 2023-08-16
Out of pods operator fixes

DmitryDodzin:
Small changes relevant to operator for #1782.


mirrord/protocol/src/codec.rs
```diff
@@ -80,6 +82,9 @@ pub enum FileRequest {
     GetDEnts64(GetDEnts64Request),
 }
 
+pub static CLIENT_READY_FOR_LOGS: LazyLock<VersionReq> =
```
eyalb181, line 85: Can you document this?

eyalb181 (changes requested):

PR #1801, opened 2023-08-16
Fix Go Arm on Linux crash

aviramha:
I managed to find a cleaner way, and that's by using internal Go functions to call our functions ;)

### DIFF
--- a/mirrord/layer/tests/apps/fileops/go/main.go
+++ b/mirrord/layer/tests/apps/fileops/go/main.go
@@ -6,5 +6,11 @@ import (
 )
 
 func main() {
-	syscall.Splice(0, nil,2,nil,4,5)
+	tempFile := "/tmp/test_file.txt"
+	syscall.Open(tempFile, syscall.O_CREAT|syscall.O_WRONLY, 0644)
+	var stat syscall.Stat_t
+	err := syscall.Stat(tempFile, &stat)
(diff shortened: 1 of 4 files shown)


mirrord/layer/tests/apps/fileops/go/main.go
```diff
@@ -6,11 +6,5 @@ import (
...
 
 func main() {
-	tempFile := "/tmp/test_file.txt"
-	syscall.Open(tempFile, syscall.O_CREAT|syscall.O_WRONLY, 0644)
-	var stat syscall.Stat_t
-	err := syscall.Stat(tempFile, &stat)
-	if err != nil {
-		panic(err)
-	}
+	syscall.Splice(0, nil,2,nil,4,5)
```
meowjesty, line 9:
```suggestion
	syscall.Splice(0, nil, 2, nil, 4, 5)
```

Is this the same test? This looks like a file ops, but the amazing go documentation for it is:

```go
// THIS FILE IS GENERATED BY THE COMMAND AT THE TOP; DO NOT EDIT
```

Add some docs so we can tell what this is from a glance.

meowjesty (changes requested):
Just docs stuff, the asm looks understandable (especially with the comments, thanks).

eyalb181 (approved):

PR #1808, opened 2023-08-17
Add option to run agent container as privileged

aviramha:


infiniteregrets (approved):
Looks good, but im not sure if setting capabilities and setting privileged conflicts?

aviramha:
> Looks good, but im not sure if setting capabilities and setting privileged conflicts?

Why would it conflict?

infiniteregrets:
> > Looks good, but im not sure if setting capabilities and setting privileged conflicts?
> 
> Why would it conflict?

Ah I guess I thought both options would be mutually exclusive in some way, either you can pick and choose or just get all. I see thats not the case

PR #1810, opened 2023-08-17
`ResponseError::DnsLookup` for errors during DNS lookup

### LINKED ISSUE #1809: Return the correct error for `dns_lookup` agent -> layer
```
async fn dns_lookup(root_path: &Path, host: String) -> RemoteResult<DnsLookup> {
    let resolv_conf_path = root_path.join("etc").join("resolv.conf");
    let hosts_path = root_path.join("etc").join("hosts");

    let resolv_conf = fs::read(resolv_conf_path)?;
    let hosts_file = File::open(hosts_path)?;
```

If the file is not found, the error is a filesystem error and we should return the correct "dns" related error here

infiniteregrets:
closes #1809 

Just converted other errors to be DnsLookup errors so we always send ResponseError::DnsLookup as an error

### DIFF
--- a/mirrord/agent/src/dns.rs
+++ b/mirrord/agent/src/dns.rs
@@ -78,9 +78,9 @@ pub async fn dns_worker(mut rx: Receiver<DnsRequest>, pid: Option<u64>) -> Resul
                     kind: ResolveErrorKindInternal::Unknown,
                 }),
             });
-            if let Err(result) = tx.send(GetAddrInfoResponse(result)) {
-                error!("couldn't send result to caller {result:?}");
-            }
+        if let Err(result) = tx.send(GetAddrInfoResponse(result)) {
+            error!("couldn't send result to caller {result:?}");
+        }
     }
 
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/dns.rs
```diff
@@ -67,8 +67,18 @@ pub async fn dns_worker(mut rx: Receiver<DnsRequest>, pid: Option<u64>) -> Resul
     while let Some(DnsRequest { request, tx }) = rx.recv().await {
         trace!("dns_worker -> request {:#?}", request);
 
-        let result = dns_lookup(root_path.as_path(), request.node);
-        if let Err(result) = tx.send(GetAddrInfoResponse(result.await)) {
+        let result = dns_lookup(root_path.as_path(), request.node)
+            .await
+            .inspect_err(|err| {
```
aviramha, line 72: You can drop the `inspect_err` and do the logging in the map

aviramha (approved):

PR #1816, opened 2023-08-20
Operator Analytics Collection

### LINKED ISSUE #1805: Add more information to analytics
When user is using operator, add the `license_hash` of the operator and the user's hash to the analytics.

DmitryDodzin:
closes #1805

### DIFF
--- a/mirrord/analytics/src/lib.rs
+++ b/mirrord/analytics/src/lib.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use base64::{engine::general_purpose, Engine as _};
 use serde::{Deserialize, Serialize};
 use tracing::info;
 
@@ -14,6 +15,7 @@ pub enum AnalyticValue {
     Bool(bool),
     Number(u32),
     Nested(Analytics),
+    Hash(AnalyticsHash),
 }
 
 /// Struct to store analytics data.
@@ -66,6 +68,22 @@ impl Analytics {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AnalyticsHash(#[serde(with = "serde_base64")] Vec<u8>);
+
+impl AnalyticsHash {
+    pub fn from_digest(val: u64) -> Self {
+        AnalyticsHash(val.to_be_bytes().to_vec())
+    }
+
+    pub fn from_base64(val: &str) -> Option<Self> {
+        general_purpose::STANDARD_NO_PAD
+            .decode(val)
+            .map(AnalyticsHash)
+            .ok()
+    }
+}
+
 /// Structs that collect analytics about themselves should implement this trait
 pub trait CollectAnalytics {
     /// Write analytics data to the given `Analytics` struct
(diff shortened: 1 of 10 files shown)


mirrord/analytics/src/lib.rs
```diff
@@ -66,6 +67,31 @@ impl Analytics {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AnalyticsHash(#[serde(with = "serde_base64")] Vec<u8>);
```
meowjesty, line 71: I have no idea why this has to be base64, so maybe add docs explaining?

mirrord/analytics/src/lib.rs
```diff
@@ -66,6 +67,31 @@ impl Analytics {
...
+
+impl AnalyticsHash {
+    pub fn from_digest(val: u64) -> Self {
```
meowjesty, line 74: Why `to_be_bytes` (docs)?

mirrord/analytics/src/lib.rs
```diff
@@ -66,6 +67,31 @@ impl Analytics {
...
+        AnalyticsHash(val.to_be_bytes().to_vec())
+    }
+
+    pub fn from_base64(val: &str) -> Option<Self> {
```
meowjesty, line 78:
```suggestion
    /// Converts from a base64 `&str`.
    pub fn from_base64(val: &str) -> Option<Self> {
```

But, isn't this a `TryFrom<&str>` implementation? Why as a separate method?

mirrord/auth/src/credential_store.rs
```diff
@@ -133,19 +131,20 @@ impl CredentialStoreSync {
...
 
-    /// Get client certificate while keeping an exclusive lock on `CREDENTIALS_PATH` (and releasing
-    /// it regrading of result)
-    pub async fn get_client_certificate<R>(
+    /// Actual implementation for exclusive lock on `CREDENTIALS_PATH`.
+    pub async fn with_exclusive_lock<R, C, V>(
         client: &Client,
         credential_name: String,
-    ) -> Result<Vec<u8>>
+        callback: C,
```
Razz4780, line 141: Nitpick - since we don't know what the callback is, we should probably be prepared for a panic. Right now, if the callback panics, we don't unlock the file

mirrord/operator/src/client.rs
```diff
@@ -145,7 +150,22 @@ impl OperatorApi {
...
 
+            let client_certificate =
+                if let Some(credential_name) = status.spec.license.fingerprint.as_ref() {
+                    CredentialStoreSync::with_exclusive_lock::<MirrordOperatorCrd, _, _>(
+                        &operator_api.client,
+                        credential_name.to_string(),
+                        |credentails| Ok(credentails.as_ref().clone()),
+                    )
+                    .await
+                    .map_err(|err| debug!("CredentialStore error: {err}"))
```
meowjesty, line 161: Is `debug` the level we want here (asking as I don't actually know)?

meowjesty (approved):
The usual "add docs for stuff" spiel

mirrord/analytics/src/lib.rs
DmitryDodzin: `Vec<u8>` serializes really badly to json so base64

mirrord/operator/src/client.rs
DmitryDodzin: I don't think this is something the user should care about. Mostly for us

mirrord/auth/src/credential_store.rs
DmitryDodzin: Realised I don't need it anymore =]

PR #1824, opened 2023-08-21
Add configuration templating

### LINKED ISSUE #1817: Configuration templating
Right now our configuration supports constant values, but some users asked for templating features, so they could have a config file that is mutated by the specific run.
I think maybe we should start with adding ability to use environment variables, so for example have
```json
{
  "target": {
    "path": "{DEPLOYMENT_NAME}",
    "namespace": "{NAMESPACE_NAME}"
  }
}
```
then user can set `DEPLOYMENT_NAME` and fetch it from env.

aviramha:
We now support templating in our configuration files using the Tera engine.
```
{
    "target": {
        "path": "{{ get_env(name="TARGET", default="pod/fallback") }}"
    }
}
```

Closes #1817


mirrord/config/src/lib.rs
```diff
@@ -415,13 +433,14 @@ impl LayerFileConfig {
...
-        let path = path.as_ref();
-        let file = std::fs::read(path)?;
-
-        match path.extension().and_then(|os_val| os_val.to_str()) {
-            Some("json") => Ok(serde_json::from_slice::<Self>(&file[..])?),
-            Some("toml") => Ok(toml::from_str::<Self>(&String::from_utf8_lossy(&file))?),
-            Some("yaml") => Ok(serde_yaml::from_slice::<Self>(&file[..])?),
+        let mut template_engine = Tera::default();
+        template_engine.add_template_file(path.as_ref(), Some("main"))?;
+        let rendered = template_engine.render("main", &tera::Context::new())?;
```
DmitryDodzin, line 438: Maybe add some values like `mirrord-version` & `protocol-version` values to context (`mirrord-version` can be useful if you have a proxy ghcr.io)

Razz4780:
One important issue - we parse config files in the extensions. Tera is based on Jinja2/Django templates, but it's not fully compatible: `...does not aim to be 100% compatible with them`. We need to ensure the configs are valid jsons/yamls/tomls

meowjesty:
> other approach would be to add another command to cli, which would read the config and transform it into plain json

We've added something along those lines, currently only for the target config.

Razz4780 (commented):
Looks ok, but maybe it's a good opportunity to add some config parsing tests?

Razz4780 (approved):

PR #1825, opened 2023-08-21
`not-found` file filter

### LINKED ISSUE #1694: AWS/cloud configurations issues
User tries to use AWS api from mirrord session and it uses their local aws config instead (because the sdk tries to find it).
This happened to me also with kubeconfig when I debugged the operator. I think maybe we should consider unsetting ENV or overriding files by default to allow better experience when dealing with such cases.
Need to research and think of a nice solution - for example unsetting envs or disallowing certain files.

Razz4780:
Closes #1694 

I passed on resolving the `~`, as it's already handled by regexes. Open to suggestions.

~~Also, there is a conflict between our default `remote-readonly` and `not-found` filters. The default `remote-readonly` filter matches paths `\.aws/cli/cache/.+\.json$`, `\.aws/credentials$` and `\.aws/config$`. Not sure how to deal with this. An ideal solution would be to have a list of `(pattern, logic)` entries and stop at the first one matching, but this is not backwards-compatible~~

### DIFF
--- a/mirrord/config/src/feature/fs.rs
+++ b/mirrord/config/src/feature/fs.rs
@@ -87,9 +87,7 @@ impl MirrordConfig for FsUserConfig {
                 local: FromEnv::new("MIRRORD_FILE_LOCAL_PATTERN")
                     .source_value()
                     .transpose()?,
-                not_found: FromEnv::new("MIRRORD_FILE_NOT_FOUND_PATTERN")
-                    .source_value()
-                    .transpose()?,
+                not_found: None,
             },
             FsUserConfig::Advanced(advanced) => advanced.generate_config()?,
         };
@@ -110,16 +108,13 @@ impl MirrordToggleableConfig for FsUserConfig {
         let local = FromEnv::new("MIRRORD_FILE_LOCAL_PATTERN")
             .source_value()
             .transpose()?;
-        let not_found = FromEnv::new("MIRRORD_FILE_NOT_FOUND_PATTERN")
-            .source_value()
-            .transpose()?;
 
         Ok(FsConfig {
             mode,
             read_write,
             read_only,
             local,
-            not_found,
+            not_found: None,
         })
     }
 }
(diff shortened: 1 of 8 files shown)


mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -83,6 +86,12 @@ pub struct FsConfig {
     /// Specify file path patterns that if matched will be opened locally.
     #[config(env = "MIRRORD_FILE_LOCAL_PATTERN")]
     pub local: Option<VecOrSingle<String>>,
+
+    /// ### feature.fs.not_found {#feature-fs-not_found}
+    ///
+    /// Specify file path patterns that if matched will be treated as non-existent.
+    #[config(env = "MIRRORD_FILE_NOT_FOUND_PATTERN")]
```
aviramha, line 93: I don't think we need an env for it

mirrord/layer/src/file/filter.rs
```diff
@@ -112,9 +104,9 @@ fn generate_remote_ro_set() -> RegexSet {
     let patterns = [
         // AWS cli cache
         // "file not exist" for identity caches (AWS)
-        r".aws/cli/cache/.+\.json$",
-        r".aws/credentials$",
-        r".aws/config$",
+        r"\.aws/cli/cache/.+\.json$",
```
aviramha, line 107: the aws here was actually a patch to return not found by making it doing it remote, so you can remove it :)

mirrord/config/src/feature/fs.rs
```diff
@@ -87,6 +87,9 @@ impl MirrordConfig for FsUserConfig {
                 local: FromEnv::new("MIRRORD_FILE_LOCAL_PATTERN")
                     .source_value()
                     .transpose()?,
+                not_found: FromEnv::new("MIRRORD_FILE_NOT_FOUND_PATTERN")
```
aviramha, line 90: remove

aviramha (approved):
Great job!

PR #1833, opened 2023-08-22
Add pause feature for ephemeral container

### LINKED ISSUE #1358: Support pause with ephemeral containers

aviramha:
Added support for pause feature on ephemeral, requires agent to be privileged.
Closes #1358

### Summary by CodeRabbit

- New Feature: Added support for pausing ephemeral containers. This feature requires the agent to have privileged access.
- Refactor: Introduced modules for handling cgroups and namespaces, improving code organization and maintainability.
- Test: Added new tests to verify the behavior of logging and unpausing a container when the agent exits early due to an error.
- Documentation: Updated documentation to clarify that the `pause` feature requires the agent configuration to be set to privileged when running with the ephemeral agent option.
- Bug Fix: Modified the standard error output redirection of a command to the parent process instead of discarding it, improving error visibility.


mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn freeze_path(&self) -> &'static str;
+    fn freeze_command(&self) -> &'static str;
+    fn unfreeze_command(&self) -> &'static str;
+}
+
+#[derive(Debug)]
+pub(crate) struct CgroupV1 {}
+
+impl CgroupFreeze for CgroupV1 {
+    fn type_name(&self) -> &'static str {
```
meowjesty, line 61:
```suggestion
    const fn type_name(&self) -> &'static str {
```

Does this work?

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,117 @@
...
+    error::Result,
+    namespace::{set_namespace, NamespaceType},
+};
+
+const CGROUP_MOUNT_PATH: &str = "/mirrord_cgroup";
+
+/// Trait for objects that can be paused
+#[enum_dispatch]
+pub(crate) trait CgroupFreeze {
+    fn pause(&self) -> Result<()> {
```
meowjesty, line 20: The `pause` method is doing too much, and thus violating the Single-Responsibility Principle, established by the 1949 Genoa Convention. Consider breaking it into 20 different npm modules to avoid breaking international laws.

meowjesty (commented):
Some nits.

mirrord/agent/src/namespace.rs
```diff
@@ -0,0 +1,40 @@
...
+
+/// Non exhaustive namespace type enum. Add as needed
+#[derive(Debug)]
+pub(crate) enum NamespaceType {
+    Net,
+    Cgroup,
+}
+
+impl NamespaceType {
+    fn path_from_pid(&self, pid: u64) -> String {
```
meowjesty, line 15:
```suggestion
    #[tracing::instrument(level = "trace", ret)]
    fn path_from_pid(&self, pid: u64) -> String {
```

meowjesty (approved):
Other than missing docs + tracing, looks fine.

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut open_options = OpenOptions::new();
+        let mut file = open_options.write(true).open(self.freeze_path())?;
+        file.write_all(self.unfreeze_command().as_bytes())?;
+        Ok(())
+    }
+
+    fn type_name(&self) -> &'static str;
+    fn freeze_path(&self) -> &'static str;
+    fn freeze_command(&self) -> &'static str;
+    fn unfreeze_command(&self) -> &'static str;
```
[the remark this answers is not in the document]
aviramha: It does support but enum dispatch doesn't

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,165 @@
...
+        // Enter the namespace, we might be already in it but it doesn't really matter
+        // Note: Entering the cgroup namespace **doesn't** set put our process in the cgroup :phew:
+        set_namespace(1, NamespaceType::Cgroup)?;
+        // Check if our cgroup is mounted, if not, mount it.
+        let cgroup_path = Path::new(CGROUP_MOUNT_PATH);
+        if !cgroup_path.exists() {
+            trace!("mounting cgroup");
+            std::fs::create_dir(cgroup_path)?;
+            mount(
+                None::<&str>,
```
meowjesty, line 124: lol

meowjesty (approved):
:rabbit: :rabbit2: :gun:

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,256 @@
...
+            let line = line?;
+            let mut line_iter = line.split(':');
+            // we don't care about the number, prefer the string for comparison
+            line_iter.next().ok_or(AgentError::PauseFailedCgroup(
+                "malformed ID cgroup v1 file".to_string(),
+            ))?;
+            let cgroup_type = line_iter.next().ok_or_else(|| {
+                AgentError::PauseFailedCgroup("malformed cgroup type cgroup v1 file".to_string())
+            })?;
+            if cgroup_type == "freezer" {
```
meowjesty, line 51:
```suggestion
let cgroup_type = line.split(':').find(|value| value == "freezer").ok_or_else(...)?;
```

Can't this be reduced to something like this?

mirrord/agent/src/runtime.rs
```diff
@@ -288,11 +287,26 @@ impl ContainerRuntime for ContainerdContainer {
...
+    /// When running on ephemeral, root pid is always 1 and env is the current process' env. (we
+    /// copy it from the k8s spec)
+    async fn get_info(&self) -> Result<ContainerInfo> {
+        Ok(ContainerInfo::new(1, std::env::vars().collect()))
+    }
+
+    /// Pause requires root privileges, so if it fails on permission we send a message.
+    async fn pause(&self) -> Result<()> {
+        get_cgroup()
+            .and_then(|x| x.pause())
```
meowjesty, line 303:
```suggestion
            .and_then(CgroupFreeze::pause)
```

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,256 @@
...
+    error::{AgentError, Result},
+    namespace::{set_namespace, NamespaceType},
+};
+
+const CGROUP_MOUNT_PATH: &str = "/mirrord_cgroup";
+const CGROUP_SUBGROUP_MOUNT_PATH: &str = "/mirrord_cgroup/subgroup";
+const CGROUPV2_PROCS_FILE: &str = "cgroup.procs";
+
+/// Trait for objects that can be paused
+#[enum_dispatch]
```
meowjesty, line 23: Why use `enum_dispatch`? Are we going to implement this for more than 2 cgroup types?

Can't we just have a normal enum 

```rust
enum Cgroup {
    V1(CgroupV1),
    V1(CgroupV2),
}
```

Implement the differences in the inner structs, e.g.:

```rust
impl CgroupV1 {
    fn new(...) -> Result<Self> { ... }
}
```

and the parts that are the same in the enum impl:

```rust
impl Cgroup {
    fn freeze(&mut self) -> Result<()> {
        match self {
            V1(v1) => v1.freeze(),
            V2(v2) => v2.freeze(),
        }
    }
}
```

and if you really want, you can have a trait for the inner structs to implement:

```rust
trait Freezer {
    fn freeze(&self);
    fn unfreeze(&self);
}

impl Freezer for CgroupV1 { ... }
impl Freezer for CgroupV2 { ... }
```

Then we can use tokio IO stuff and have the file operations be `async`.

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,256 @@
...
+                }
+            }
+            v
+        })
+        .unwrap_or_default()
+}
+
+/// Write given pids to given path's "cgroup.procs" file
+#[tracing::instrument(level = "trace", ret)]
+fn move_pids_to_cgroupv2(cgroup_path: &Path, pids: Vec<u64>) -> Result<()> {
```
meowjesty, line 143: `cgroup_path` is being passed in some functions, why not put it into the `CgroupV2` struct, then turn these free functions into methods taking `&self`?

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,256 @@
...
+        std::fs::write(cgroup_path.join(CGROUPV2_PROCS_FILE), format!("{pid}")).map_err(|_| {
+            AgentError::PauseFailedCgroup("write pid to subgroup failed".to_string())
+        })?;
+    }
+    Ok(())
+}
+
+/// (Un)Freeze the given cgroup
+#[tracing::instrument(level = "trace", ret)]
+fn freeze_cgroupv2(cgroup_path: &Path, on: bool) -> Result<()> {
```
meowjesty, line 154: Same suggestion as the other "put cgroup_path in the struct" as above.

meowjesty (changes requested):
Have some suggestions around code organization.

Don't see the need for `enum_dispatch` here, as we're only ever going to handle 2 types of cgroup, right?

mirrord/agent/src/cgroup.rs
aviramha: I prefer a more explicit parsing so if the format isn't what we expect we error, and we know beter what's going on..

mirrord/agent/src/cgroup.rs
aviramha: It's being called twice with different path, so not really helpful

mirrord/agent/src/runtime.rs
aviramha:
```
type mismatch in function arguments
expected function signature `fn(cgroup::Cgroup) -> _`
   found function signature `for<'a> fn(&'a _) -> _`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B2%5D?2#file%3A%2F%2F%2FUsers%2Faviramhassan%2FCode%2Fmirrord%2Fmirrord%2Fagent%2Fsrc%2Fruntime.rs)
cgroup.rs(26, 5): found signature defined here
runtime.rs(309, 14): required by a bound introduced by this call
result.rs(1338, 27): required by a bound in `std::result::Result::<T, E>::and_then`
cgroup.rs(26, 16): do not borrow the argument
```

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+use tracing::{trace, warn};
+
+use crate::{
+    error::{AgentError, Result},
+    namespace::{set_namespace, NamespaceType},
+};
+
+const CGROUP_MOUNT_PATH: &str = "/mirrord_cgroup";
+const CGROUP_SUBGROUP_MOUNT_PATH: &str = "/mirrord_cgroup/subgroup";
+const CGROUPV2_PROCS_FILE: &str = "cgroup.procs";
```
meowjesty, line 19: Are associated constants still unstable? If so can leave it like this, otherwise I would suggest putting the v1/v2 specific consts in the impl blocks:

```rust
impl CgroupV1 {
    const CGROUP_V1_FREEZE_PATH: ....;
}
```

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+                });
+            }
+        }
+        Err(AgentError::PauseFailedCgroup(
+            "no freezer cgroup found".to_string(),
+        ))
+    }
+}
+
+const CGROUP_V1_FREEZE_PATH: &str = "freezer.state";
```
meowjesty, line 61: Same associated const suggestion.

mirrord/agent/src/cgroup.rs
meowjesty: Ouch, but this kinda adds a lot of noise, because we don't really care about all these in-between steps (seeing that they all pretty much return the same error), the goal here is to either get the path with "freeze" or fail.

The extra steps feels likes debug code, like we're doing a manual `inspect` trying to find what's going wrong.

meowjesty (commented):
Suggestions that I'm not sure are in stable Rust yet, so leaving them as comments.

mirrord/agent/src/runtime.rs
```diff
@@ -288,11 +287,30 @@ impl ContainerRuntime for ContainerdContainer {
...
+    }
 
-    setns(fd, nix::sched::CloneFlags::CLONE_NEWNET)?;
-    Ok(())
+    async fn unpause(&self) -> Result<()> {
+        let cgroup = Cgroup::new().await?;
+        cgroup
+            .unpause()
+            .await
+            .map_err(|e| AgentError::PauseFailedCgroup(e.to_string()))
```
meowjesty, line 314: Use `?`.

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// Path to the the process' cgroup - usually /kubepods/besteffort/ID
+    cgroup_path: PathBuf,
+}
+
+impl CgroupV1 {
+    #[tracing::instrument(level = "trace", ret)]
+    pub(crate) async fn new() -> Result<Self> {
+        let file = File::open("/proc/1/cgroup")
+            .await
+            .map_err(|_| AgentError::PauseFailedCgroup("opening proc cgroup failed".to_string()))?;
```
meowjesty, line 35: We're using this `AgentError::PauseFailedCgroup` with `map_err` a lot, due to this error being a kind of "catch all" and having the context be a string.

One of the reasons for using `thiserror` instead of `anyhow` is to have stronger error types and avoid this "string as context".

Maybe this error should in the category of `PauseCgroup::OpeningProc`.

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+        let mut reader = BufReader::new(file).lines();
+        while let Some(line) = reader.next_line().await? {
+            let mut line_iter = line.split(':');
+            // we don't care about the number, prefer the string for comparison
+            line_iter.next().ok_or(AgentError::PauseFailedCgroup(
```
meowjesty, line 40: Stronger type for the error, something like `PauseCgroup::MalformedIdV1`

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+                "malformed ID cgroup v1 file".to_string(),
+            ))?;
+            let cgroup_type = line_iter.next().ok_or_else(|| {
+                AgentError::PauseFailedCgroup("malformed cgroup type cgroup v1 file".to_string())
```
meowjesty, line 44: `PauseCgroup::MalformedTypeV1`

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+            })?;
+            if cgroup_type == "freezer" {
+                let cgroup_path = line_iter.next().ok_or_else(|| {
+                    AgentError::PauseFailedCgroup("malformed path cgroup v1 file".to_string())
```
meowjesty, line 48: `PauseCgroup::MalformedPathV1`

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+                })?;
+                return Ok(Self {
+                    // strip / since joining "/a" and "/b" results in "/b"
+                    cgroup_path: Path::new(CGROUP_MOUNT_PATH).join(PathBuf::from(
+                        cgroup_path.strip_prefix('/').unwrap_or(cgroup_path),
+                    )),
+                });
+            }
+        }
+        Err(AgentError::PauseFailedCgroup(
```
meowjesty, line 58: `PauseCgroup::NoFreezer`

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+/// then find the cgroup path, join it and write freeze to it.
+impl CgroupV1 {
+    #[tracing::instrument(level = "trace", ret, skip(self))]
+    async fn pause(&self) -> Result<()> {
+        // Check if our cgroup is mounted, if not, mount it.
+        let cgroup_path = Path::new(CGROUP_MOUNT_PATH);
+        if !try_exists(cgroup_path).await? {
+            trace!("mounting cgroup");
+            create_dir(cgroup_path).await.map_err(|_| {
+                AgentError::PauseFailedCgroup("create dir failed cgroupv1".to_string())
```
meowjesty, line 70: `PauseCgroup::CreateDirV1` and drop `map_err` in favor of `?`.

mirrord/agent/src/cgroup.rs
```diff
@@ -0,0 +1,262 @@
...
+            })?;
+            mount(
+                Some("cgroup"),
+                cgroup_path,
+                Some("cgroup"),
+                MsFlags::MS_NOSUID | MsFlags::MS_NOEXEC | MsFlags::MS_NODEV,
+                Some("freezer"),
+            )
+            .map_err(|_| AgentError::PauseFailedCgroup("mount failed cgroupv1".to_string()))?;
```
meowjesty, line 79: `PauseCgroup::MountV1`, no `map_err`?

meowjesty (changes requested):
There are a bunch of `map_err` and errors with string as context.

I would suggest turning these Cgroup errors into properly typed errors, and adding a variant for them in the `AgentError`, like `AgentError::CgroupError`, and have:

```rust
#[derive(error)]
enum CgroupError { ... }
```

Then have the error checking in the `From<Error>` implementation, instead of having it in the middle of usage functions.

meowjesty (approved):
:bear: :+1: 

Errors, errors everywhere!

PR #1835, opened 2023-08-22
Changed config verify to return aggregated warnings 

aviramha:
Changed config verify to return aggregated warnings list for user to print instead of warn in current progress - can fix issues with extension where we printed to stderr.

Might fix issues where config is wrong and weird message will be seen in extension (since it sends it to stderr)


meowjesty (approved):
:+1:

PR #1842, opened 2023-08-23
Fix ephemeral agent creation api using agent namespace 

aviramha:
Fix ephemeral agent creation api using agent namespace instead of  target.

### Summary by CodeRabbit

- Bug Fix: Fixed the creation of ephemeral agents by correctly using the agent namespace instead of the target. This ensures that the correct namespace is used when retrieving the Pod resource from the Kubernetes API.
- Documentation: Added a note to the `AgentConfig` struct's `namespace` field documentation, stating that it doesn't work with ephemeral containers. Also, introduced a warning message when using an ephemeral container for the agent and specifying an agent namespace, informing users that the agent namespace is ignored in this case.
- Refactor: Introduced a new parameter `pod_namespace` to the `ContainerApi` trait and its implementations, allowing the namespace to be specified when creating or updating an agent. Also added a new field `pod_namespace` in the `RuntimeData` struct to store the namespace of the associated pod.

### DIFF
--- a/mirrord/kube/src/api/container.rs
+++ b/mirrord/kube/src/api/container.rs
@@ -175,7 +175,7 @@ impl ContainerApi for JobContainer {
         connection_port: u16,
         progress: &P,
         agent_gid: u16,
-        pod_namespace: Option<String>,
+        _pod_namespace: Option<String>,
     ) -> Result<String>
     where
         P: Progress + Send + Sync,
--- a/mirrord/kube/src/api/container.rs
+++ b/mirrord/kube/src/api/container.rs
@@ -52,6 +52,7 @@ pub trait ContainerApi {
         connection_port: u16,
         progress: &P,
         agent_gid: u16,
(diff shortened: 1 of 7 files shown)


mirrord/kube/src/api/container.rs
```diff
@@ -52,6 +52,7 @@ pub trait ContainerApi {
         connection_port: u16,
         progress: &P,
         agent_gid: u16,
+        pod_namespace: Option<String>,
```
DmitryDodzin, line 55: Maybe add to `RuntimeData` instead?

DmitryDodzin: It can't be targetless on Ephemeral

DmitryDodzin (changes requested):
Add the namespace to `RuntimeData` there you always have the namespace and when it's empty it's irrelevant for Ephemeral

DmitryDodzin (approved):

PR #1846, opened 2023-08-24
Fixed issue connecting to ephemeral container when target is in different namespace

aviramha:
Fixed issue connecting to ephemeral container when target is in different namespace

### DIFF
--- a/mirrord/cli/src/connection.rs
+++ b/mirrord/cli/src/connection.rs
@@ -1,7 +1,10 @@
 use std::time::Duration;
 
 use mirrord_config::{feature::network::outgoing::OutgoingFilterConfig, LayerConfig};
-use mirrord_kube::api::{kubernetes::KubernetesAPI, AgentManagment};
+use mirrord_kube::api::{
+    kubernetes::{AgentKubernetesConnectInfo, KubernetesAPI},
+    AgentManagment,
+};
 use mirrord_operator::client::{OperatorApi, OperatorApiError, OperatorSessionInformation};
 use mirrord_progress::Progress;
 use mirrord_protocol::{ClientMessage, DaemonMessage};
@@ -13,7 +16,7 @@ use crate::{CliError, Result};
 pub(crate) enum AgentConnectInfo {
     Operator(OperatorSessionInformation),
     /// Connect directly to an agent by name and port using k8s port forward.
-    DirectKubernetes(String, u16),
+    DirectKubernetes(AgentKubernetesConnectInfo),
 }
 
 pub(crate) struct AgentConnection {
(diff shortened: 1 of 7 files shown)


mirrord/cli/src/connection.rs
```diff
@@ -1,21 +1,44 @@
...
 
+#[derive(Debug, Serialize, Deserialize)]
 pub(crate) enum AgentConnectInfo {
     Operator(OperatorSessionInformation),
     /// Connect directly to an agent by name and port using k8s port forward.
-    DirectKubernetes(String, u16),
+    DirectKubernetes(AgentKubernetesConnectInfo),
 }
 
+const AGENT_CONNECT_INFO_KEY: &str = "MIRRORD_AGENT_CONNECT_INFO";
```
DmitryDodzin, line 24: Maybe should be part of clap args to intproxy so the error will happen as part of clap validation

mirrord/cli/src/connection.rs
```diff
@@ -1,21 +1,44 @@
...
 use tracing::trace;
 
 use crate::{CliError, Result};
 
+#[derive(Debug, Serialize, Deserialize)]
 pub(crate) enum AgentConnectInfo {
     Operator(OperatorSessionInformation),
     /// Connect directly to an agent by name and port using k8s port forward.
-    DirectKubernetes(String, u16),
+    DirectKubernetes(AgentKubernetesConnectInfo),
```
DmitryDodzin, line 21: Maybe add `DirectStream` and not use the value from `config.connect_tcp` inside intproxy but only the `AgentConnectInfo`

mirrord/cli/src/internal_proxy.rs
```diff
@@ -400,25 +403,26 @@ async fn connect(
...
-        Err(InternalProxyError::NoConnectionMethod.into())
+        )),
+        Some(AgentConnectInfo::DirectKubernetes(connect_info)) => {
+            let k8s_api = KubernetesAPI::create(config).await?;
+            let connection = k8s_api.create_connection(connect_info).await?;
+            Ok((connection, None))
+        }
+        None => {
+            if let Some(address) = &config.connect_tcp {
+                let stream = TcpStream::connect(address)
```
DmitryDodzin, line 419: We have a forgotten type inside of `mirrrord-kube` called `Connection` that does this exact code but with `AgentManagement` like `KubernetesAPI`

DmitryDodzin (approved):
Maybe it's better to pass connection info via clap but not critical.

mirrord/cli/src/connection.rs
aviramha: I'd prefer to do it in another PR, as we have tests that rely on CONNECT_TCP env..

mirrord/cli/src/internal_proxy.rs
aviramha: Tried to use it but seems cumbersome and not really adding any value.. (you need to initialize struct with address, then send the address also in the create_connect, all of that instead of just this 2 lines.)

PR #1850, opened 2023-08-25
Faster CI + SIP issue

aviramha:
- use Go cache for integration tests
- Fix macOS SIP potential issues from exec having mirrord loaded into the code sign binary.
- use mac-13 runner for tests

### DIFF
--- a/mirrord/layer/tests/configs/bash_script.json
+++ b/mirrord/layer/tests/configs/bash_script.json
@@ -2,7 +2,7 @@
     "feature": {
         "fs": {
             "mode": "read",
-            "local": ["/var/db/timezone/icutz"]
+            "local": ["/var/db/"]
         }
     }
 }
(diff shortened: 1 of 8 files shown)


mirrord/layer/tests/configs/bash_script.json
```diff
@@ -0,0 +1,8 @@
+{
+    "feature": {
+        "fs": {
+            "mode": "read",
+            "local": ["/var/db/"]
```
eyalb181, line 5: Can you document this?

mirrord/sip/src/codesign.rs
```diff
@@ -3,11 +3,13 @@ use std::{os::unix::process::ExitStatusExt, path::Path, process::Command};
 use crate::error::{Result, SipError};
 
 pub(crate) fn sign<P: AsRef<Path>>(path: P) -> Result<()> {
+    // don't load mirrord into the codesign binary
```
eyalb181, line 6: Doesn't this comment only apply to env_remove?

eyalb181 (approved):

PR #1861, opened 2023-08-28
Change config unstable/deprecations warnings to use the aggregated config warnings

### LINKED ISSUE #1860: mirrord config - change unstable/deprecations to happen once
When we load a `LayerConfig`, if users uses unstable/deprecated feature it alerts each time we load `LayerConfig` which creates a bit of spammy behavior.
There's `verify` method for config, so we should move the logic to happen there.

aviramha:
Change config unstable/deprecations warnings to use the aggregated config warnings
Closes #1860

### DIFF
--- a/mirrord/cli/src/extension.rs
+++ b/mirrord/cli/src/extension.rs
@@ -57,14 +57,12 @@ pub(crate) async fn extension_exec(args: ExtensionExecArgs) -> Result<()> {
         std::env::set_var("MIRRORD_IMPERSONATED_TARGET", target.clone());
         env.insert("MIRRORD_IMPERSONATED_TARGET".into(), target.to_string());
     }
-    let (config, mut warnings) = LayerConfig::from_env_with_warnings()?;
+    let (config, mut context) = LayerConfig::from_env_with_warnings()?;
 
     let mut analytics = AnalyticsReporter::only_error(config.telemetry);
 
-    warnings.extend(config.verify()?);
-    for warning in warnings {
-        progress.warning(&warning);
-    }
+    let mut warnings = config.verify()?;
+    warnings.extend(context.get_warnings().iter().cloned());
 
     #[cfg(target_os = "macos")]
     let execution_result = mirrord_exec(
--- a/mirrord/cli/src/extension.rs
+++ b/mirrord/cli/src/extension.rs
@@ -61,8 +61,10 @@ pub(crate) async fn extension_exec(args: ExtensionExecArgs) -> Result<()> {
 
     let mut analytics = AnalyticsReporter::only_error(config.telemetry);
 
-    let mut warnings = config.verify()?;
-    warnings.extend(context.get_warnings().iter().cloned());
+    config.verify(&mut context)?;
+    for warning in context.get_warnings() {
+        progress.warning(&warning);
+    }
 
     #[cfg(target_os = "macos")]
     let execution_result = mirrord_exec(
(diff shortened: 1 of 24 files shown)


mirrord/cli/src/operator.rs
```diff
@@ -71,7 +71,8 @@ async fn operator_setup(
 
 async fn get_status_api(config: Option<String>) -> Result<Api<MirrordOperatorCrd>> {
     let kube_api = if let Some(config_path) = config {
-        let config = LayerFileConfig::from_path(config_path)?.generate_config()?;
+        let mut warnings = Vec::new();
+        let config = LayerFileConfig::from_path(config_path)?.generate_config(&mut warnings)?;
```
DmitryDodzin, line 75: Maybe add add print for warnings if they happen?

mirrord/config/src/lib.rs
```diff
@@ -278,12 +278,24 @@ pub struct LayerConfig {
 }
 
 impl LayerConfig {
-    pub fn from_env() -> Result<Self, ConfigError> {
+    /// Generate a config from the environment variables and/or a config file.
+    /// On success, returns the config and a vec of warnings.
+    /// To be used from CLI to verify config and print warnings
+    pub fn from_env_with_warnings() -> Result<(Self, Vec<String>), ConfigError> {
+        let mut warnings = Vec::new();
```
DmitryDodzin, line 285: For consistency with other APIs and `from_env` prettier
```suggestion
    pub fn from_env_with_warnings(warnings: &mut Vec<String>) -> Result<(Self, Vec<String>), ConfigError> {
```

mirrord/config/src/util.rs
```diff
@@ -35,11 +35,11 @@ where
 {
     type Generated = T::Generated;
 
-    fn generate_config(self) -> Result<Self::Generated, ConfigError> {
+    fn generate_config(self, warnings: &mut Vec<String>) -> Result<Self::Generated, ConfigError> {
```
DmitryDodzin, line 38: One little thing that will make future work easier on it is maybe wrapping the `Vec<String>` in some `ConfigContext` struct (like if maybe `String` should be `LogMessage` or have some extra data like a pretty error)

mirrord/cli/src/operator.rs
aviramha: don't think we need it in this context

mirrord/config/src/util.rs
aviramha: Nice!

mirrord/config/src/lib.rs
aviramha: I think the internal API shouldn't be consistent with external tbh.. also makes funky code from caller.

mirrord/cli/src/extension.rs
```diff
@@ -57,10 +57,13 @@ pub(crate) async fn extension_exec(args: ExtensionExecArgs) -> Result<()> {
         std::env::set_var("MIRRORD_IMPERSONATED_TARGET", target.clone());
         env.insert("MIRRORD_IMPERSONATED_TARGET".into(), target.to_string());
     }
-    let config = LayerConfig::from_env()?;
+    let (config, mut context) = LayerConfig::from_env_with_warnings()?;
 
     let mut analytics = AnalyticsReporter::only_error(config.telemetry);
 
+    let mut warnings = config.verify()?;
+    warnings.extend(context.get_warnings().iter().cloned());
```
DmitryDodzin, line 65: add the loop with `progress.warning(&warning);` =]

DmitryDodzin (approved):

PR #1867, opened 2023-08-28
Add flag to disable openshift check

aviramha:


aviramha:
Closed for now - in case we find out that it slows down we'll just remove the check.

PR #1874, opened 2023-08-29
Enable ephemeral tests on operator

infiniteregrets:
new cfg attrs to configure tests based on flags

### DIFF
--- a/tests/src/env.rs
+++ b/tests/src/env.rs
@@ -1,3 +1,4 @@
+#![allow(dead_code, unused)]
 #[cfg(test)]
 
 mod env {
--- a/tests/src/env.rs
+++ b/tests/src/env.rs
@@ -7,7 +7,7 @@ mod env {
 
     use crate::utils::{run_exec_with_target, service, EnvApp, KubeService};
 
-    #[cfg(any(feature = "job", not(feature = "ephemeral")))]
+    #[cfg(all(feature = "job", not(feature = "ephemeral")))]
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[timeout(Duration::from_secs(240))]
@@ -18,7 +18,7 @@ mod env {
         remote_env_vars_works(service, application).await;
     }
 
-    #[cfg(any(feature = "job", not(feature = "ephemeral")))]
+    #[cfg(all(feature = "job", not(feature = "ephemeral")))]
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[timeout(Duration::from_secs(120))]
--- a/tests/src/env.rs
+++ b/tests/src/env.rs
@@ -7,6 +7,7 @@ mod env {
 
     use crate::utils::{run_exec_with_target, service, EnvApp, KubeService};
 
+    #[cfg(any(feature = "job", not(feature = "ephemeral")))]
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[timeout(Duration::from_secs(240))]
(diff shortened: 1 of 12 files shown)


tests/src/traffic/steal.rs
```diff
@@ -194,8 +198,10 @@ mod steal {
         let kube_client = kube_client.await;
         let mut flags = vec!["--steal"];
         if let Some(flag) = agent.flag() {
-            flags.extend(flag)
+            flags.extend(flag);
+            Agent::patch_operator(&kube_client).await;
         }
+
```
DmitryDodzin, line 204:
```suggestion
        }
        agent.patch_operator(&kube_client).await;

```
It's a bit ambiguous for the patch to be here I think the better option is to always call the `patch_operator` and let it decide if it needs to be done with `&self` (especially because the patch might flip and we will need to specify when it's not ephemeral)

DmitryDodzin (changes requested):
This can't work because of concurrent tests, the patch needs to happen once

infiniteregrets:
> This can't work because of concurrent tests, the patch needs to happen once

 if let Ok(_) = std::env::var("MIRRORD_OPERATOR_TESTS") && OPERATOR_PATCHED.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed).unwrap() {
 
 I have set a flag already here? are you referring to something else?

tests/src/env.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code, unused)]
```
DmitryDodzin, line 1: This is the reason I was thinking of using ignore instead of `#[cfg]` straight up

tests/src/env.rs
```diff
@@ -7,6 +8,7 @@ mod env {
 
     use crate::utils::{run_exec_with_target, service, EnvApp, KubeService};
 
+    #[cfg(all(feature = "job", not(feature = "ephemeral")))]
```
DmitryDodzin, line 11:
```suggestion
    #[cfg_attr(not(feature = "job"), ignore)]
```

tests/src/file_ops.rs
```diff
@@ -1,12 +1,14 @@
...
 mod file_ops {
 
     use std::time::Duration;
 
     use rstest::*;
 
-    use crate::utils::{run_exec_with_target, service, Agent, FileOps, KubeService};
+    use crate::utils::{run_exec_with_target, service, FileOps, KubeService};
 
+    #[cfg(any(feature = "job", feature = "ephemeral"))]
```
DmitryDodzin, line 11:
```suggestion
    #[cfg_attr(not(any(feature = "ephemeral", feature = "job")), ignore)]
```

tests/src/file_ops.rs
```diff
@@ -16,18 +18,13 @@ mod file_ops {
...
         let service = service.await;
         let _ = std::fs::create_dir(std::path::Path::new("/tmp/fs"));
         let command = ops.command();
 
-        let mut args = vec!["--fs-mode", "write"];
-
-        if let Some(ephemeral_flag) = agent.flag() {
-            args.extend(ephemeral_flag);
-        }
+        let args = vec!["--fs-mode", "write"];
```
DmitryDodzin, line 27:
```suggestion
        let args = vec!["--fs-mode", "write"];
        
        if cfg!(feature = "ephemeral") {
            args.extend("-e");
        }
```

infiniteregrets: why not cfg?
```
#[cfg(feature = "ephemeral")]
mirrord_args.extend(["--ephemeral-container"].into_iter());
```

tests/src/env.rs
infiniteregrets: thank you! i understand your point now (: fixed!

DmitryDodzin (approved):

PR #1885, opened 2023-09-01
style guide: avoid non-optional Options

t4lz:
This suggestion was not discussed anywhere else first, this PR is meant to be the place for objections/discussion/corrections/support.


aviramha (approved):

aviramha:
Great change :)

Issue #1317, opened 2023-04-18
Support mirroring existing sessions


aviramha:
Mirroring works only for new TCP sessions.
We can make it work with ongoing TCP sessions by finding HTTP requests and starting to stream from there.
Should resolve #397

sunng87:
It would be nice to have this supported. This makes mirrord more straight forward with how the doc describes it.

aviramha:
> I just needed this for already established websocket connections

We won't be able to support mirroring established websocket connections because those are stateful connections

aviramha:
> Is it doable for stealing?

It already works this way for stealing.

meowjesty:
With #1887 this should now work if your session is using HTTP!

There is a small caveat in the implementation:

Currently only handle it if the packet data starts with valid HTTP bytes, something along the lines of: `["GET / HTTP"]`. It fails to detect HTTP traffic if the bytes are "messed up", e.g. `["some other data \0 GET / HTTP]`. There is an implicit assumption here that we're going to be hitting the common case of seeing a "well-behaved" packet most of the time.

- A quick fix to this, if it ever poses a problem:

Perform HTTP detection on the packet, and move the bytes' cursor to the start of the HTTP part, example:

```
["some other data \0 GET / HTTP]
 /\                 /\
 |                  |
we're starting here
                    |
                    |
      create a slice that starts from here
```

PR #1887, opened 2023-09-01
Mirror existing connections.

meowjesty:
- Issue: #1317 #397

Adds support for mirroring existing (and long-lived) connections that were set-up before the agent started sniffing.

With this change, mirroring will now look into connections that were already ongoing before it (mirrord) started, trying to see if they contain an HTTP message, and if they do, then we establish a new mirroring session for this connection.

- Related PRs to get this merged:

- [x] [Add test image for issue 1317: mirror existing session](https://github.com/metalbear-co/test-images/pull/19);
- [x] [add missing platform.sh to http-keep-alive](https://github.com/metalbear-co/test-images/pull/20);
- [ ] [Add e2e action for mirror existing connection](https://github.com/metalbear-co/ci/pull/28);


mirrord/agent/src/main.rs
```diff
@@ -712,7 +712,10 @@ async fn main() -> Result<()> {
     tracing_subscriber::registry()
         .with(
             tracing_subscriber::fmt::layer()
-                .with_thread_ids(true)
+                .with_thread_ids(false)
```
aviramha, line 715: different PR please

mirrord/agent/src/sniffer.rs
```diff
@@ -322,6 +324,15 @@ pub struct TcpConnectionSniffer {
 impl TcpConnectionSniffer {
     pub const TASK_NAME: &'static str = "Sniffer";
 
+    /// Used to detect if an existing connection packet contains HTTP traffic that we should be
+    /// mirroring, but we're currently NOT.
+    ///
+    /// **Attention**: Whitespace at the end of this regex is significant!
+    const HTTP_REGEX: LazyLock<Regex> = LazyLock::new(|| {
```
aviramha, line 331: why not re-use the logic we have to http steal?

mirrord/agent/src/sniffer.rs
```diff
@@ -491,7 +508,42 @@ impl TcpConnectionSniffer {
         Ok(())
     }
 
-    #[tracing::instrument(level = "trace", skip(self, eth_packet), fields(bytes = %eth_packet.len()))]
+    /// First it checks the `tcp_flags` with [`is_new_connection`], if that's not the case, meaning
+    /// we have traffic from some existing connection from before mirrord started, then it tries to
+    /// see if `bytes` contains an HTTP request (HTTP/1) of some sort. When an HTTP request is
+    /// detected, then the agent should start mirroring as if it was a new connection.
+    #[tracing::instrument(level = "debug", ret, skip(bytes))]
+    fn treat_as_new_session(tcp_flags: u16, bytes: &[u8]) -> bool {
```
aviramha, line 516: We would want the mirror to start from the valid HTTP stream, and also a potential flow could be
Frame 1: "AAAENDOFPREVIOUSRESPONSE H"
Frame 2: "TTP"
and we wouldn't catch it, so we actually need something like we have for the steal, check current buf is http start marker - no? skip by 1 byte then try again, when bytes are exhausted:

1. if we were in middle of comparison, wait until more data is read.
2. if last byte wasn't part of comparison, you can drop the buffer.

aviramha (commented):
It's very different than what I had in mind but I actually like it - I thought we would find HTTP stream and replay it as http, but you did it so we just "start" the stream from there, which is quite elegant so I'm okay with it. 🐻  in mind that we'd probably later want to sequence/build each HTTP request from the stream, so we can filter like we do for steal.

mirrord/agent/src/sniffer.rs
meowjesty: I was thinking the same, but couldn't make it happen with simple HTTP requests.

Like, they always come in the new packets, maybe if I send some big requests I can trigger this behavior? That's why I'm not even changing the packet bytes to start from where the regex starts matching.

meowjesty: Still works when sending something like `vec![1; 1500]` as the body.

Only way I can see to trigger the invalid frame behavior is if I send this type of frame myself, as a `TcpStream` instead of sending an http request.

aviramha: I'd assume this behavior happens on HTTP proxies or high performant services where they buffer and flush to socket, that can create separate network frames. In the end the sender can't really control how the TCP frames are built. That's part of the OSI spec.

aviramha: You can test it in an integration test also (or just UT tbh)

meowjesty: Yeah, I'm sure this can (and probably will) happen.

Should I:

a. go for a more complete implementation with the whole hyper and stuff right now? or;
b. merge this simple solution, as it kinda works on average cases, then work on (a)?

aviramha: Good q (and idea).. I think yes -  @eyalb181 ?

eyalb181: I'm for option b

aviramha: נ

PR #1896, opened 2023-09-03
Change agent to use current thread runtime

aviramha:
Change agent to use current thread runtime, multi thread is enabled by mistake


mirrord/agent/src/container_handle.rs
```diff
@@ -24,17 +23,15 @@ struct Inner {
...
-                Ok(())
-            }
-        });
-
-        if let Err(err) = result {
-            error!("Could not unpause target container while exiting, got error: {err:?}");
+                if let Err(err) = container.unpause().await {
+                    error!("Could not unpause target container while exiting, got error: {err:?}");
+                }
+            });
```
infiniteregrets, line 34: a changelog entry for this deadlock soln would be great

DmitryDodzin (approved):

Issue #1890, opened 2023-09-03
medschool misses a lot of configuration


eyalb181:
### Bug Description

When I run medschool, a lot of the configuration is missing (biggest example being all of `network.outgoing`, but there are others).

### Steps to Reproduce

N/A

### Your operating system and version

N/A

### Local process

N/A

PR #1900, opened 2023-09-04
Better pattern for if some / else none

meowjesty:
- Issue: #1890 (loosely related, not a fix for the issue)

While investigating #1890 found this `if -> Some / else -> None` pattern that can be simplified with `bool.then()`.

Ok / not ok? Leaving it up for the reviewer if this pattern is ok or not, as we had previously agreed on not using `bool.then` in the middle of functions to have an early return with `?`, but this case is different.


t4lz (approved):
You're going to have to feed the towncrier.

meowjesty:
Hear ye! Hear ye! Alex is refactoring code again.

changelog.d/1890.changed.md
t4lz, line 1: Should be `internal` - users don't care about code refactors.

t4lz (changes requested):

Issue #1899, opened 2023-09-04
remote file isn't accessible dotnet linux (WSL)


aviramha:
[deleteme.zip](https://github.com/metalbear-co/mirrord/files/12514320/deleteme.zip)

In agent logs we saw xstat for the directory that worked, so probably something funny happens there.

aviramha:
seems that we lack a hook for opendir.

aviramha:
@meowjesty will implement the relevant hook and we'll test if that solves the issue.

aviramha:
Fixed in #1901 thanks @meowjesty

PR #1901, opened 2023-09-04
Add hook for opendir, issue #1899

meowjesty:
- Issue: #1899

Add hook for `opendir` by leveraging our `open` and `fdopendir` hooks.

Also adds hooks for `readdir64_r`, `__lxstat`, `__lxstat64`, `__xstat64`.

Refactors `*stat` hooks to use a common function.

Adds an integration test for `opendir`.

### DIFF
--- a/mirrord/layer/src/file/hooks.rs
+++ b/mirrord/layer/src/file/hooks.rs
@@ -11,7 +11,7 @@ use std::{ffi::CString, os::unix::io::RawFd, ptr, slice, time::Duration};
 use errno::{set_errno, Errno};
 use libc::{
     self, c_char, c_int, c_void, dirent, off_t, size_t, ssize_t, stat, statfs, AT_EACCESS,
-    AT_FDCWD, DIR, O_RDONLY,
+    AT_FDCWD, DIR, O_DIRECTORY, O_RDONLY,
 };
 #[cfg(target_os = "linux")]
 use libc::{dirent64, stat64, EBADF, EINVAL, ENOENT, ENOTDIR};
@@ -143,7 +143,7 @@ pub(super) unsafe extern "C" fn open_nocancel_detour(
 /// [`fdopendir`] to convert the [`RawFd`] into a `*DIR` stream (which we treat as `usize`).
 #[hook_guard_fn]
 pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> usize {
-    open_logic(raw_filename, O_RDONLY, 0)
+    open_logic(raw_filename, O_RDONLY, O_DIRECTORY)
         .and_then(fdopendir)
         .unwrap_or_bypass_with(|_| FN_OPENDIR(raw_filename))
 }
--- a/mirrord/layer/src/file/hooks.rs
+++ b/mirrord/layer/src/file/hooks.rs
@@ -302,6 +302,26 @@ pub(crate) unsafe extern "C" fn openat_detour(
     })
 }
 
+/// Equivalent to `open_detour`, **except** when `raw_path` specifies a relative path.
+///
+/// If `fd == AT_FDCWD`, the current working directory is used, and the behavior is the same as
+/// `open_detour`.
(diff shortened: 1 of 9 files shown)


mirrord/layer/src/file/hooks.rs
```diff
@@ -133,6 +133,17 @@ pub(super) unsafe extern "C" fn open_nocancel_detour(
     }
 }
 
+/// Hook for [`libc::opendir`].
+///
+/// Opens the directory with `read` permission using the [`open_logic`] flow, then calls
+/// [`fdopendir`] to convert the [`RawFd`] into a `*DIR` stream (which we treat as `usize`).
+#[hook_guard_fn]
+pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> usize {
+    open_logic(raw_filename, O_RDONLY, 0)
```
aviramha, line 142: need to handle case open succeeds but fdopendir fails (close fd)

aviramha: I'd add O_DIRECTORY just to be on the safe side (though we ignore it, but maybe we wont in the future)

aviramha:
The integration tests pass for me locally when I comment out the "macOS" part - maybe it's x64 thingy though (needing the xstat?)

aviramha:
> The integration tests pass for me locally when I comment out the "macOS" part - maybe it's x64 thingy though (needing the xstat?)

ah I think I know what's the problem - you need to hook `opendir$INODE64` on macOS x64.

aviramha:
After few tests, I found out that for listing files in dir we lack `readdir64_r`
and after that it still won't work - The best way to trigger the same issue is to check if `Directory.Exists` works - and it doesn't :( - I think it might be using ~stat~ func for it.

aviramha:
[patch.patch](https://github.com/metalbear-co/mirrord/files/12523704/patch.patch)
adding patch file for what I did so far

mirrord/layer/src/file/hooks.rs
```diff
@@ -133,6 +137,17 @@ pub(super) unsafe extern "C" fn open_nocancel_detour(
     }
 }
 
+/// Hook for [`libc::opendir`].
+///
+/// Opens the directory with `read` permission using the [`open_logic`] flow, then calls
+/// [`fdopendir`] to convert the [`RawFd`] into a `*DIR` stream (which we treat as `usize`).
+#[hook_guard_fn]
+pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> usize {
+    open_logic(raw_filename, O_RDONLY, O_DIRECTORY)
```
aviramha, line 146: if open succeeds but fdopendir fails, we need to call `close_layer_fd`

meowjesty:
```rust
fn note(&mut self) {
    println!("Remember to update the changelog, as we did more than just add opendir.");
}
```

mirrord/layer/src/file/hooks.rs
```diff
@@ -804,6 +899,7 @@ pub(crate) unsafe fn enable_file_hooks(hook_manager: &mut HookManager) {
     );
 
     replace!(hook_manager, "openat", openat_detour, FnOpenat, FN_OPENAT);
+    replace!(hook_manager, "openat64", openat_detour, FnOpenat, FN_OPENAT);
```
aviramha, line 902: It'd be more preferable to have different detours, for case where `openat` resolves but `openat64` doesn't (as it assigns it to the `FN_OPENAT` (we had cases like this, that's why inode hooks are in cfg, which tbh we should just have similar logic)

mirrord/layer/src/file/hooks.rs
meowjesty: Nice catch.

mirrord/layer/src/file/hooks.rs
meowjesty: Changed.

aviramha (approved):

Issue #1893, opened 2023-09-03
Run IDE e2es on new version branches


eyalb181:
Because the IDE plugins are now decoupled from CLI versions, bugs in the CLI can sneak in without being tested with the e2e tests. However, we don't want to run e2es for both IDE plugins for every PR (would be exhausting). As a middle ground, we want to run the IDE e2es only for release branches (so branches that conform to the X.X.X naming convention).

infiniteregrets:
also refer: https://github.com/metalbear-co/mirrord-vscode/pull/64

PR #1915, opened 2023-09-06
Trigger ide e2e on releases

infiniteregrets:
refer #1893


.github/workflows/ci.yaml
```diff
@@ -131,6 +131,19 @@ jobs:
       # Check that compiles for the supported linux targets (aarch64)
       - run: cargo-zigbuild clippy --lib --bins --all-features --target aarch64-unknown-linux-gnu -- -Wclippy::indexing_slicing -D warnings
 
+  # echo "branch=$([[ ${{ github.head_ref }} =~ "^[0-9]+\.[0-9]+\.[0-9]+$" ]]  && echo "true" || echo "false" )" >> "$GITHUB_OUTPUT"
+  check_if_release_branch:
+    runs-on: ubuntu-latest
+    outputs:
+      release_branch: ${{ steps.release-branch.outputs.branch }}
+    steps:
+      - id: release-branch
```
eyalb181, line 140: Can you explain what's happening in this step and the next?

.github/workflows/ci.yaml
```diff
@@ -425,6 +438,30 @@ jobs:
           config: "markdownlint-config.json"
           args: "README.md"
 
+  build_mirrord_on_release_branch:
```
eyalb181, line 441: Why do we need a specific build step for the release branch? Don't we build mirrord on all branches anyway?

.github/workflows/run-ide-e2e.yaml
```diff
@@ -0,0 +1,136 @@
+name: Run IDE e2e on release
+on:
+  workflow_call:
+
+# this worflow copies the exact IDE e2e and runs it on the release branches
+jobs:
+  intellij-e2e:
```
eyalb181, line 7: Do you need to copy the entire workflow? Can't you trigger the original workflow with something like this? https://github.com/marketplace/actions/trigger-external-workflow

eyalb181 (changes requested):

.github/workflows/run-ide-e2e.yaml
infiniteregrets: - since we want to block in case something fails, i couldnt find a way to do that with this, i also looked at other actions that use repository_dispatch
- we want to test with mirrord binary built on the tag created in the pr, and add it to the path

what we can do is, move the reusable part of the workflow to `mirrord-intellij` and use that here as well

.github/workflows/ci.yaml
infiniteregrets: - we don't explicitly build mirrord? not that i could find it in the steps of integration tests?
- building separately is faster, as we dont have to depend on heavy jobs and we can concurrently proceed with the ide e2e

.github/workflows/ci.yaml
infiniteregrets: in this step, we get the head ref and compare it with a regex to check if we are on a branch of pattern like `*.*.*`

```
[[ ide-workflow-trigger ~= "^[0-9]+\.[0-9]+\.[0-9]+$" ]]  && echo "true" || echo "false"
```
would fail, while something like `3.5.1` would pass

eyalb181 (changes requested):

.github/workflows/ci.yaml
```diff
@@ -131,6 +131,19 @@ jobs:
       # Check that compiles for the supported linux targets (aarch64)
       - run: cargo-zigbuild clippy --lib --bins --all-features --target aarch64-unknown-linux-gnu -- -Wclippy::indexing_slicing -D warnings
 
+  # echo "branch=$([[ ${{ github.head_ref }} =~ "^[0-9]+\.[0-9]+\.[0-9]+$" ]]  && echo "true" || echo "false" )" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
infiniteregrets: old commit btw

.github/workflows/ci.yaml
```diff
@@ -130,6 +130,18 @@ jobs:
...
+  
+  check_if_release_branch:
+    runs-on: ubuntu-latest
+    outputs:
+      release_branch: ${{ steps.release-branch.outputs.branch }}
+    steps:
+      - id: release-branch
+        run: |
+          echo "branch=$([[ ${{ github.head_ref }} =~ "^[0-9]+\.[0-9]+\.[0-9]+$" ]]  && echo "true" || echo "false" )" >> "$GITHUB_OUTPUT"
+      - name: output test
```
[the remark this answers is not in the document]
infiniteregrets: added comments

eyalb181 (approved):

Issue #1840, opened 2023-08-23
Configuration validation from cli


aviramha:
The new command will take a cfg path, and output the debug representation of `LayerConfig` and stderr will be used for outputing list of errors/warnings, so extensions can use it.

meowjesty:
`mirrord verify-config [path]` has been added in #1926

Only the `TargetConfig` part is returned back when validation succeeds.

meowjesty:
Welp, after changing it to a tagged union, the `Failure` portion doesn't serialize to json.

Working on a fix.

PR #1926, opened 2023-09-07
Add verify-config command that checks if a config file is valid.

meowjesty:
- Issue: #1840

Adds the new `mirrord verify-config [path]` command to `mirrord-cli`. When used, this command checks a config file specified by `path` and produces a json output with either success + warnings, or failure.

To be used as a way for the IDE plugins to get easy access to config validation.


mirrord/cli/src/main.rs
```diff
@@ -404,6 +405,43 @@ async fn register_to_waitlist(email: EmailAddress) -> Result<()> {
     Ok(())
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+struct VerifiedConfig {
+    config: Option<String>,
+    errors: Option<String>,
```
aviramha, line 411: make config an object, right now make it return the target
make errors a Vec, and also add warnings.

mirrord/cli/src/main.rs
```diff
@@ -404,6 +405,43 @@ async fn register_to_waitlist(email: EmailAddress) -> Result<()> {
...
+/// ## Usage
+///
+/// ```sh
+/// mirrord verify-config ./config.json
+/// ```
+async fn verify_config(VerifyConfigArgs { path }: VerifyConfigArgs) -> Result<()> {
+    let mut config_context = ConfigContext::default();
+
+    let read_path = path.clone();
+    let verified = match async move { LayerFileConfig::from_path(path) }
```
aviramha, line 425: this seems to be a reverse functional programming which is functional but a mess to read..

mirrord/cli/src/main.rs
```diff
@@ -404,6 +405,43 @@ async fn register_to_waitlist(email: EmailAddress) -> Result<()> {
...
+        .and_then(|config| async move { config.generate_config(&mut config_context) })
+        .and_then(|_| tokio::fs::read_to_string(read_path).map_err(ConfigError::from))
+        .await
+    {
+        Ok(config) => VerifiedConfig {
```
aviramha, line 430: call verify on the generated config.

mirrord/cli/src/main.rs
aviramha: target = pod / deployment / rollout and/or namespace

aviramha (changes requested):

mirrord/cli/src/main.rs
meowjesty: Yup, this was just pure chaos while I was thinking on how to keep the errors and read the file.

Nuked :boom:

mirrord/cli/src/main.rs
meowjesty: Returns the `TargetConfig` right?

aviramha: It doesn't have to be `TargetConfig` exactly - it needs to be what the extension needs - which is namespace if any and pod/deploymento r any (so it can know if it should show a dialog)

mirrord/cli/src/main.rs
```diff
@@ -402,6 +403,62 @@ async fn register_to_waitlist(email: EmailAddress) -> Result<()> {
     Ok(())
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+enum VerifiedConfig {
```
aviramha, line 407: imo it looks better with tagged enum

mirrord/cli/src/main.rs
meowjesty: But isn't this the same thing?

mirrord/cli/src/main.rs
meowjesty:
```json
{
"verified:" {
  "Success": {
    "config": {
      "path": {
        "deployment": "sample-python-deployment",
        "container": null
      },
      "namespace": null
    },
    "warnings": []
  }
}
}
```

You mean this?

aviramha: no - 
```json
{
"type": "success",
"config": {...},
"warnings": []
}
```
or failure:
```json
{
"type": "error",
"errors": ["don't do this"]
}
```

mirrord/cli/src/main.rs
aviramha: I am just saying that it's up to you - we might not want it to be bound by TargetConfig since if we add stuff there it'll be propagated back here, and we might be better with other approach.

meowjesty: Alright, I get it now.

tbh I'm not quite sure what the IDE side would want from this, so maybe leaving it as the `TargetConfig` is best? I don't really want to introduce a struct in the middle here, where we just have to convert for no reason, while we don't add any crazy stuff that would break this.

aviramha (approved):
Looks great. Thank you.

PR #1927, opened 2023-09-08
Refactor code to match newest Rust nightly - 2023-09-07

### LINKED ISSUE #1457: nightly-2023-05-24 doesn't compile

aviramha:
Closes #1457

### DIFF
--- a/medschool/src/main.rs
+++ b/medschool/src/main.rs
@@ -175,13 +175,13 @@ impl Eq for PartialType {}
 impl Eq for PartialField {}
 
 impl Hash for PartialType {
-    fn hash<H: ~const std::hash::Hasher>(&self, state: &mut H) {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.ident.hash(state)
     }
 }
 
 impl Hash for PartialField {
-    fn hash<H: ~const std::hash::Hasher>(&self, state: &mut H) {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.ident.hash(state)
     }
 }
--- a/medschool/src/main.rs
+++ b/medschool/src/main.rs
@@ -192,8 +192,11 @@ impl Display for PartialType {
         let docs = self.docs.last().cloned().unwrap_or_default();
         f.write_str(&docs)?;
 
-        let fields: String = self.fields.iter().map(|field| format!("{field}")).collect();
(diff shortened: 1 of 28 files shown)


medschool/src/main.rs
```diff
@@ -192,8 +192,11 @@ impl Display for PartialType {
         let docs = self.docs.last().cloned().unwrap_or_default();
         f.write_str(&docs)?;
 
-        let fields: String = self.fields.iter().map(|field| format!("{field}")).collect();
-        f.write_str(&fields)
+        self.fields.iter().fold(f, |accum, s| {
+            accum.write_str(&s.to_string()).expect("writing failed");
+            accum
+        });
+        Ok(())
```
t4lz, line 199: Would this work? So that we can avoid that scary `expect`?
```suggestion
        self.fields.iter().try_for_each(|s| {
            f.write_str(&s.to_string())
        })
```

t4lz (approved):
Nice

medschool/src/main.rs
aviramha: It's in medschool and I think it's better to crash if a field can't be translated to str then to quietly fail

PR #1939, opened 2023-09-12
Remove spammy messages from progress

### LINKED ISSUE #1934: Remove spammy progress messages
We're showing some progress messages that aren't really interesting to the user:
- "detecting OpenShift..."
- "checking if node is allocatable...", "node is allocatable", "unable to check if node is allocatable" (I know this is a warning, but it's spammy and doesn't help the user in any way; note that the error "node is not allocatable" should still show)

infiniteregrets:
closes #1934


mirrord/cli/src/connection.rs
```diff
@@ -129,9 +129,8 @@ where
             .await
             .map_err(CliError::KubernetesApiFailed)?;
 
-        let mut detect_openshift_task = progress.subtask("detecting OpenShift...");
-        if (k8s_api.detect_openshift(&mut detect_openshift_task).await).is_err() {
-            detect_openshift_task.warning("couldn't determine OpenShift");
+        if (k8s_api.detect_openshift(progress).await).is_err() {
+            progress.warning("couldn't determine OpenShift");
```
eyalb181, line 133: We don't need this warning either + I think it makes sense to add logs in place of all the messages you remove

eyalb181: Is "OpenShift was not detected." still going to show? If so please remove that as well

eyalb181 (changes requested):

eyalb181 (approved):

PR #1948, opened 2023-09-17
Fix for polling on bad fd (tokio), with uvicorn reload

### LINKED ISSUE #1618: uvicorn with reload crashes
### Bug Description

When running a FastAPI app using Uvicorn and mirrord enabled the process gets tokio errors

### Steps to Reproduce

Create a FastAPI + Uvicorn app from PyCharm template, run it using mirrord.

### Backtrace

```shell
INFO:     Will watch for changes in these directories: ['/Users/aviramhassan/PycharmProjects/fastApiProject1']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [49235] using WatchFiles
thread 'tokio-runtime-worker' panicked at 'unexpected error when polling the I/O driver: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/io/mod.rs:180:23
thread 'tokio-runtime-worker' panicked at 'unexpected error when polling the I/O driver: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/io/mod.rs:180:23
INFO:     Started server process [49238]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
```

### Relevant Logs

_No response_

macOS M1

### Local process

python 3.10

### Local process version

_No response_

### Additional Info

_No response_

infiniteregrets:
wip to fix #1618

### DIFF
--- a/mirrord/layer/src/common.rs
+++ b/mirrord/layer/src/common.rs
@@ -6,7 +6,7 @@ use mirrord_protocol::{file::OpenOptionsInternal, RemoteResult};
 #[cfg(target_os = "macos")]
 use mirrord_sip::{MIRRORD_TEMP_BIN_DIR_CANONIC_STRING, MIRRORD_TEMP_BIN_DIR_STRING};
 use tokio::sync::oneshot;
-use tracing::warn;
+use tracing::{warn, info};
 
 use crate::{
     detour::{Bypass, Detour},
@@ -56,6 +56,7 @@ pub(crate) type ResponseChannel<T> = oneshot::Sender<RemoteResult<T>>;
 ///
 /// [`ClientMessage`]: mirrord_protocol::codec::ClientMessage
 pub(crate) fn blocking_send_hook_message(message: HookMessage) -> HookResult<()> {
+    info!("Sending hook message: {:#?}", message);
     // SAFETY: mutation happens only on initialization.
     unsafe {
         HOOK_SENDER
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -521,10 +521,11 @@ fn layer_start(mut config: LayerConfig) {
             // https://github.com/metalbear-co/mirrord/issues/1792
             // so we leak it.
             std::mem::forget(old_sender);
+        } else {
+            HOOK_SENDER
+                .set(sender)
+                .expect("Setting HOOK_SENDER singleton");
         }
-        HOOK_SENDER
-            .set(sender)
-            .expect("Setting HOOK_SENDER singleton");
     };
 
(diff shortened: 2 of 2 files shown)


infiniteregrets:
its quite weird whats going on here, adding else block fixes the polling issue but now the channel has None values?

```
INFO:     Started reloader process [30798] using WatchFiles
2023-09-17T02:25:16.822011Z  INFO ThreadId(01) mirrord_layer: Initializing mirrord-layer!
2023-09-17T02:25:16.823108Z  INFO ThreadId(01) mirrord_layer: Initializing mirrord-layer!
2023-09-17T02:25:16.832788Z  INFO ThreadId(03) mirrord_layer: Received hook message: None
thread '<unnamed>' panicked at mirrord/layer/src/lib.rs:826:56:
called `Option::unwrap()` on a `None` value
stack backtrace:
2023-09-17T02:25:16.832985Z  INFO ThreadId(01) mirrord_layer::common: Sending hook message: Tcp(
    Close(
        8000,
    ),
)
```

aviramha:
> its quite weird whats going on here, adding else block fixes the polling issue but now the channel has None values? the message to be sent here is the `Close`, but the channel received a None before `Close` as `socket.close` cant get the sender and fails on `Err(CannotGetHookSender)`?

Yes, because you changed it so it doesn't set it when it's not set so it's None

infiniteregrets:
> > its quite weird whats going on here, adding else block fixes the polling issue but now the channel has None values? the message to be sent here is the `Close`, but the channel received a None before `Close` as `socket.close` cant get the sender and fails on `Err(CannotGetHookSender)`?
> 
> Yes, because you changed it so it doesn't set it when it's not set so it's None

yeah I realized that, question: why did we have to leak the sender in the first place on forks?

aviramha:
Because the `Drop` implementation didn't work post-fork.

infiniteregrets:
> Because the `Drop` implementation didn't work post-fork.

from my understanding, i think it is the case that the old sender is being used to send the message after its forgotten, maybe we should keep a hashmap of hook_senders mapping each to parent/child and blocking_sends can lookup the correct sender by thread id?

aviramha:
I have a solution for this - I'll try to implement it soon.
The idea is to re-use the sender/receiver, as it should be okay to re-use between runtimes. On fork, we would call a message to hooksender that will send us back the receiver, then we pass it.

PR #1951, opened 2023-09-18
single-threaded friendly layer

### LINKED ISSUE #1618: uvicorn with reload crashes

aviramha:
Should fix #1618 

In this PR we change the layer to be "nicer" in fork situations (which seems to affect more macOS):

1. Layer doesn't create a thread, instead on hooks it runs the runtime, actually giving it some time to run and do stuff.
2. In fork case, it's UB to call libc stuff if multi-threaded, which is something we caused as a side effect - now since we're single threaded, if the caller process is forking, it should also be single-threaded to respect that rule, and if it doesn't respect it - it's the problematic one!
3. Added a exit hook since I found out thread_local has a DTOR and `atexit` was registered and called us which lead to the following scenario:

- Tokio's dtor thread local gets called, cleans up
- 3rd party app we're loaded into calls our hooks -> tokio runtime crashes because TLS is destroyed.
- :(

Possible issue with the implementation is that if process is on `select`/`kqueue`/`epoll` and there's incoming connection it'd get it only when there's some execution time, which might never happen.

### DIFF
--- a/mirrord/layer/src/file/hooks.rs
+++ b/mirrord/layer/src/file/hooks.rs
@@ -94,11 +94,13 @@ pub(super) unsafe extern "C" fn open_detour(
     if guard.is_none() {
         FN_OPEN(raw_path, open_flags, mode)
     } else {
-        get_runtime().block_on(open_logic(raw_path, open_flags, mode)).unwrap_or_bypass_with(|_bypass| {
-            #[cfg(target_os = "macos")]
-            let raw_path = update_ptr_from_bypass(raw_path, _bypass);
-            FN_OPEN(raw_path, open_flags, mode)
-        })
+        get_runtime()
+            .block_on(open_logic(raw_path, open_flags, mode))
+            .unwrap_or_bypass_with(|_bypass| {
+                #[cfg(target_os = "macos")]
+                let raw_path = update_ptr_from_bypass(raw_path, _bypass);
+                FN_OPEN(raw_path, open_flags, mode)
+            })
     }
 }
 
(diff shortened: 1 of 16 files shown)


mirrord/layer/src/file/hooks.rs
```diff
@@ -148,21 +158,29 @@ pub(super) unsafe extern "C" fn open_nocancel_detour(
...
-    open_logic(raw_filename, O_RDONLY, O_DIRECTORY)
-        .and_then(|fd| match fdopendir(fd) {
-            Detour::Success(success) => Detour::Success(success),
-            Detour::Bypass(bypass) => Detour::Bypass(bypass),
-            Detour::Error(fail) => {
-                close_layer_fd(fd);
-                Detour::Error(fail)
+    get_runtime()
+        .block_on(async {
+            match open_logic(raw_filename, O_RDONLY, O_DIRECTORY).await {
```
meowjesty, line 163: You can keep `and_then` here, drop the `await` and use the `futures::TryStreamExt` (or whatever it's called).

PR #1964, opened 2023-09-24
Added the ability to specify targetless in config file

aviramha:
Added the ability to specify targetless in config file, to allow non-interactive targetless in IDEs

### DIFF
--- a/mirrord/kube/src/api/kubernetes.rs
+++ b/mirrord/kube/src/api/kubernetes.rs
@@ -199,20 +199,13 @@ impl AgentManagment for KubernetesAPI {
     where
         P: Progress + Send + Sync,
     {
-        let runtime_data = if let Some(ref path) = self.target.path {
+        let runtime_data = if let Some(ref path) = self.target.path && !matches!(path, mirrord_config::target::Target::Targetless) {
             let runtime_data = path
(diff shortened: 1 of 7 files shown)


mirrord/kube/src/api/kubernetes.rs
```diff
@@ -199,20 +199,13 @@ impl AgentManagment for KubernetesAPI {
...
     {
-        let runtime_data = if let Some(ref path) = self.target.path {
+        let runtime_data = if let Some(ref path) = self.target.path && !matches!(path, mirrord_config::target::Target::Targetless) {
             let runtime_data = path
                 .runtime_data(&self.client, self.target.namespace.as_deref())
                 .await?;
 
             Some(runtime_data)
         } else {
-            // Most users won't see this, since the default log level is error, and also progress
```
eyalb181, line 209: Why'd you remove this? Also is running the CLI without specifying a target still supported?

aviramha: Because as the comment says, it's never really seen. If we want users to know that, we need to change it to use the progress API.
Yes, it's still supported.

eyalb181 (approved):

Issue #1963, opened 2023-09-24
Test mirroring on mesh


eyalb181:
Although this is now theoretically supported, we've already received two reports from users that couldn't use mirroring on mesh. Need to test mirroring on mesh and make sure it actually works - make sure to test both istio and linkerd.

meowjesty:
Can you give me more details on what's not working? Language/framework, what they're trying to do?

I can replicate the failed behavior by using an unreal tcp communication just to trigger it, but would be helpful to have something closer to reality to test against it as well.

aviramha:
@meowjesty Did you setup a mesh and it worked?

meowjesty:
For people using **istio**, try setting `agent.network_interface` to `"lo"` (or your _loopback_  network interface equivalent). More info on the config [here](https://mirrord.dev/docs/overview/configuration/#agent-network_interface).

Looks like the issue is that we're sniffing on the wrong network interface for istio mesh, and the traffic we get there is encrypted (mTLS).

Working on a fix that doesn't require setting the `agent.network_interface` explicitly.

meowjesty:
Capturing on `lo` should be the default now when we detect `istio`.

PR #1981, opened 2023-09-29
Make mirror (in mesh) capture on the lo interface, instead of eth0 (or equivalent) by default (user setting takes priority).

meowjesty:
- Issue: #1963

Moved `MeshVendor` from the agent crate to the protocol crate, as it's now being used by both kube and agent. The previous implementation stuff we had in `steal/iptables` for it is still there, as an extension trait.

Instead of just detecting that we're in a mesh (kube crate), we now propagate the type of mesh to the agent through the `--mesh` arg, and if it's `istio`, sniffer captures traffic on `lo`, instead of `eth0` (steal should be unaffected by this change).

### DIFF
--- a/mirrord/agent/src/cli.rs
+++ b/mirrord/agent/src/cli.rs
@@ -1,9 +1,15 @@
 #![deny(missing_docs)]
 
+use std::str::FromStr;
+
 use clap::{Parser, Subcommand};
+use mirrord_protocol::MeshVendor;
 
 const DEFAULT_RUNTIME: &str = "containerd";
 
+/// **Heads-up**: Order of arguments passed to this matter, so if you add a new arg after something
+/// like `Targeted`, it won't work, as `Mode` is a `subcomand`. You would need to add the arg in
+/// `Mode::Targeted` for it to work (that's why `--mesh` is there and not here).
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
 pub struct Args {
@@ -49,6 +55,9 @@ pub enum Mode {
         /// Container runtime to use
         #[arg(short = 'r', long, default_value = DEFAULT_RUNTIME)]
         container_runtime: String,
+
+        #[arg(long)]
+        mesh: Option<String>,
     },
     /// Inform the agent to use `proc/1/root` as the root directory.
     Ephemeral,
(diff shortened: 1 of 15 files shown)


mirrord/agent/src/cli.rs
```diff
@@ -62,8 +71,18 @@ impl Mode {
     pub fn is_targetless(&self) -> bool {
         matches!(self, Mode::Targetless)
     }
+
+    pub(super) fn mesh(&self) -> Option<MeshVendor> {
+        match self {
+            Mode::Targeted { mesh, .. } => mesh
+                .as_ref()
+                .and_then(|mesh| MeshVendor::from_str(mesh).ok()),
```
aviramha, line 79: I don't like this - it means if someone enters wrong value it's just ignoring it..

meowjesty: Oh, it panics on the wrong value, see the `from_str` implementation.

mirrord/agent/src/cli.rs
```diff
@@ -49,6 +55,10 @@ pub enum Mode {
         /// Container runtime to use
         #[arg(short = 'r', long, default_value = DEFAULT_RUNTIME)]
         container_runtime: String,
+
+        /// Which kind of mesh the remote pod is in, see [`MeshVendor`].
+        #[arg(long)]
+        mesh: Option<String>,
```
DmitryDodzin, line 61: Can be `MeshVendor` because of `FromStr` no?

mirrord/protocol/src/lib.rs
```diff
@@ -48,3 +49,41 @@ impl Deref for EnvVars {
...
+}
+
+impl FromStr for MeshVendor {
+    type Err = ();
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s {
+            "Linkerd" => Ok(MeshVendor::Linkerd),
+            "Istio" => Ok(MeshVendor::Istio),
+            invalid => panic!(
```
DmitryDodzin, line 84: I fear panic can cause problems with handling error of backwards compatibility if we add new mesh, so something like 
```rust
#[derive(Debug, Error)]
#[error("Invalid {0:?} option for MeshVendor")]
pub struct MeshVendorParseError(String);
```
and let clap print the error

DmitryDodzin (approved):
Im a bit fearful of the panic for backwards compatibility but should be fine but still nice is to let clap do the exit and not panic in the parse function

Issue #1979, opened 2023-09-28
`verify-config` too eager, breaks IDEs when `namespace` is set, but no `target`.


meowjesty:
### Bug Description

If you set a `namespace`, but no `target`, then the IDEs error out thanks to the config check we have.

### Steps to Reproduce

Run mirrord from an IDE plugin with `namespace` set, but no `target`.

### Your operating system and version

not relevant

### Local process

any

meowjesty:
We should probably add a flag, like `mirrord verify-config --ide`, then we can treat these divergences between how strict the `cli` is versus the `ide`.

With this we would have something like (but without these `ifs`):

```rust
if !ide {
    if targetless + mirror {
        error!(":sad:");
    }
    
}
```

PR #1982, opened 2023-10-02
Add --ide flag to verify-config, changing some errors into warnings.

meowjesty:
- Issue: #1979 

Adds the optional `--ide` flag to `mirrord verify-config [--ide] --path {/config/path}`.

Creates a special case for `config::verify` where we may turn some errors (target related) to warnings when we're in an IDE context, as the user may select a target instead of putting on in the config file.

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -292,6 +292,9 @@ fn email_parse(email: &str) -> Result<EmailAddress, String> {
 #[derive(Args, Debug)]
 pub(super) struct VerifyConfigArgs {
     /// Config file path.
+    pub(super) ide: bool,
+
+    /// Config file path.
     pub(super) path: PathBuf,
 }
 
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -62,6 +62,7 @@ pub(super) enum Commands {
     InternalProxy,
 
     /// Verify config file without starting mirrord.
+    #[command(hide = true)]
     VerifyConfig(VerifyConfigArgs),
 }
 
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -62,7 +62,6 @@ pub(super) enum Commands {
     InternalProxy,
 
     /// Verify config file without starting mirrord.
-    #[command(hide = true)]
     VerifyConfig(VerifyConfigArgs),
 }
 
@@ -290,11 +289,14 @@ fn email_parse(email: &str) -> Result<EmailAddress, String> {
 
 /// Args for the [`super::verify_config`] mirrord-cli command.
 #[derive(Args, Debug)]
+#[command(group(ArgGroup::new("verify-config")))]
 pub(super) struct VerifyConfigArgs {
     /// Config file path.
+    #[arg(long)]
     pub(super) ide: bool,
 
     /// Config file path.
+    #[arg(long)]
     pub(super) path: PathBuf,
 }
 
--- a/mirrord/config/src/lib.rs
+++ b/mirrord/config/src/lib.rs
@@ -327,7 +327,7 @@ impl LayerConfig {
     /// - `ide`: Identifies if this is being called from an IDE context, when using
     /// `mirrord verify-config`. Turns some _target missing_ errors into warnings, as the target can
     /// be selected after `verify-config` is run.
-    pub fn verify(&self, context: &mut ConfigContext, ide: bool) -> Result<(), ConfigError> {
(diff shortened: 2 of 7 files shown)


mirrord/config/src/lib.rs
```diff
@@ -381,24 +386,67 @@ impl LayerConfig {
...
+            // In the IDE, a target may be selected after `mirrord verify-config` is run, so we
+            // for this case we treat these as warnings. They'll become errors once mirrord proper
+            // tries to start (if the user somehow managed to not select a target by then).
+            if ide {
+                if self.target.namespace.is_some() {
+                    context.add_warning(
+                        "A target namespace was specified, but no target was specified. \
+                            The IDE will prompt you to select a target."
+                            .into(),
+                    );
```
infiniteregrets, line 402: && !self.target.path...?

mirrord/cli/src/config.rs
```diff
@@ -62,7 +62,6 @@ pub(super) enum Commands {
     InternalProxy,
 
     /// Verify config file without starting mirrord.
-    #[command(hide = true)]
     VerifyConfig(VerifyConfigArgs),
```
infiniteregrets, line 65: i am not sure if i am aware of the context here, do people really want to use it with the cli? why remove the hidden flag?

mirrord/config/src/lib.rs
```diff
@@ -323,7 +323,11 @@ impl LayerConfig {
     /// We don't call it from `from_env` since we want to verify it only once (from cli)
     ///
     /// Fills `context` with the warnings.
-    pub fn verify(&self, context: &mut ConfigContext) -> Result<(), ConfigError> {
+    ///
+    /// - `ide`: Identifies if this is being called from an IDE context, when using
+    /// `mirrord verify-config`. Turns some _target missing_ errors into warnings, as the target can
+    /// be selected after `verify-config` is run.
+    pub fn verify(&self, context: &mut ConfigContext, ide: bool) -> Result<(), ConfigError> {
```
aviramha, line 330: maybe it's better to have it under `ConfigContext`?

mirrord/cli/src/config.rs
meowjesty: Shouldn't be removed, missed this while I was testing some stuff.

mirrord/config/src/lib.rs
meowjesty: You meant just the `ide` arg, right?

aviramha: yes

mirrord/config/src/lib.rs
```diff
@@ -381,24 +386,70 @@ impl LayerConfig {
...
-            if self.pause {
-                Err(ConfigError::Conflict(
-                    "The target pause feature is not compatible with a targetless agent, please either disable this option or specify a target.".into(),
-                ))?;
+            // In the IDE, a target may be selected after `mirrord verify-config` is run, so we
+            // for this case we treat these as warnings. They'll become errors once mirrord proper
+            // tries to start (if the user somehow managed to not select a target by then).
+            if context.ide {
+                if self.target.namespace.is_some() {
+                    context.add_warning(
```
aviramha, line 398: This shouldn't be a warning

infiniteregrets (approved):

mirrord/config/src/lib.rs
```diff
@@ -381,24 +386,70 @@ impl LayerConfig {
...
+                        "A target namespace was specified, but no target was detected in config. \
+                        The IDE will prompt you to select a target."
+                            .into(),
+                    );
+                }
+
+                if self.feature.network.incoming.is_steal() {
```
aviramha, line 405: Same

mirrord/config/src/lib.rs
```diff
@@ -381,24 +386,70 @@ impl LayerConfig {
...
+                    context.add_warning(
+                        "Steal mode is not compatible with a targetless agent.\
+                        The IDE will prompt you to select a target."
+                            .into(),
+                    );
+                }
+
+                if self.agent.ephemeral {
```
aviramha, line 413: Same

mirrord/config/src/lib.rs
```diff
@@ -381,24 +386,70 @@ impl LayerConfig {
...
+                    context.add_warning(
+                        "Using an ephemeral container for the agent is not compatible with a \
+                        targetless agent. The IDE will prompt you to select a target."
+                            .into(),
+                    );
+                }
+
+                if self.pause {
+                    context.add_warning(
```
aviramha, line 422: Same

PR #1990, opened 2023-10-04
Make example conform to its rule in style guide

t4lz:
Alternatively, we can change the rule to say `..._sender` and `..._receiver`. The point of the PR is to have the example and the rule agree with each other.


meowjesty (approved):
`approval_tx.send("👍").await.unwrap()`

PR #2021, opened 2023-10-20
Clarify exceptions in FS config docs

### LINKED ISSUE #2020: Clarify exceptions in FS config docs

t4lz:
closes #2020


mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -19,6 +19,9 @@ use crate::{
 /// 3. `"local"` - Read from the local file system.
 /// 5. `"disable"` - Disable file operations.
```
eyalb181, line 20: I think we're missing number 4 here? Is it localwithoverrides?

mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -19,6 +19,9 @@ use crate::{
...
 ///
+/// > Note: There are pre-defined exceptions to the mode, so some paths will be read locally by
```
eyalb181, line 22:
```suggestion
/// > Note: by default, some specific paths are read locally or remotely, regardless of the selected FS mode. This is described in further detail below.
```

mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -37,8 +40,22 @@ use crate::{
...
+/// 2. There are pre-defined exceptions to the set FS mode.
+///   a. Paths that match
+///      [the patterns defined here](https://github.com/metalbear-co/mirrord/tree/latest/mirrord/layer/src/file/filter/read_local_by_default.rs)
+///      are read locally by default.
+///   b. Paths that match
+///      [the patterns defined here](https://github.com/metalbear-co/mirrord/tree/latest/mirrord/layer/src/file/filter/read_remote_by_default.rs)
+///      are read remotely by default when the mode is `localwithoverrides`.
+///   c. Paths that match
+///      [the patterns defined here](https://github.com/metalbear-co/mirrord/tree/latest/mirrord/layer/src/file/filter/not_found_by_default.rs)
+///      under the running user's home directory will be failed to be found by default when the mode
```
eyalb181, line 52:
```suggestion
///      under the running user's home directory will not be found by the application when the mode
```

mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -37,8 +40,22 @@ use crate::{
...
+///      is not `local`.
+/// In order to override that default setting for a path, or a pattern, include it the appropriate
```
eyalb181, line 54: How do you override the "not found" list? Maybe it's better to add an example of how to override to each list, rather than a general example at the bottom.

eyalb181 (changes requested):

mirrord/config/src/feature/fs/advanced.rs
t4lz: I think you're right. Also, I don't think there is a `disable` mode.
```suggestion
/// 4. `"localwithoverrides"` - perform fs operation locally, unless the path matches a pre-defined or user-specified exception.
```

mirrord/config/src/feature/fs/advanced.rs
t4lz: You have to include a path in another list that specifies what should happen with it - `read_only`/`read_write`/`local`.
I'm afraid if I make this text any longer no one will ever read it.

eyalb181 (approved):

Issue #1946, opened 2023-09-15
error when polling the I/O driver


infiniteregrets:
### Bug Description

 error when polling the I/O driver: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }

### Steps to Reproduce

mix phx.new ./mirrord_demo_app --no-html --no-gettext --no-dashboard --no-assets --no-ecto --no-esbuild --no-mailer --no-tailwind`

run with mirrord

### Backtrace

```shell
via 💧 took 50s ❯ RUST_BACKTRACE=full mirrord exec mix phx.server
⠓ mirrord exec
    ✓ Running on latest (3.67.0)!
    ✓ ready to launch process
      ✓ layer extracted
      ✓ no operator detected
      ✓ OpenShift was not detected.
      ✓ agent pod created
      ✓ pod is ready                                                                                                                                                                                                  thread '<unnamed>' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.32.0/src/runtime/io/driver.rs:164:23:
unexpected error when polling the I/O driver: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }
stack backtrace:
thread '<unnamed>' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.32.0/src/runtime/io/driver.rs:164:23:
unexpected error when polling the I/O driver: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }
stack backtrace:
   0:        0x1037ee234 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h4030c4138bcdf528
   1:        0x1038107d0 - core::fmt::write::h4b06b32fd4efd613
   2:        0x1037eaba8 - std::io::Write::write_fmt::hed3d984009928384
   3:        0x1037ee074 - std::sys_common::backtrace::print::h7e2b50b6cbf90560
   4:        0x1037efbb4 - std::panicking::panic_hook_with_disk_dump::{{closure}}::ha89af3293c7da63a
   5:        0x1037ef8b4 - std::panicking::panic_hook_with_disk_dump::hc4a4f27be6208df3
   6:        0x1037f0170 - std::panicking::rust_panic_with_hook::h037f273058a2da78
   7:        0x1037f0074 - std::panicking::begin_panic_handler::{{closure}}::h9026ba3bb755434a
[...]

PR #2022, opened 2023-10-20
Sync layer refactor

### LINKED ISSUE #1952: Drop tokio runtime from layer
Following #1946 #1618 #1948 #1951 
I think the solution is to get rid of the tokio runtime, and drop async stuff from the layer.
The design I had in mind
1. Create a struct that is responsible for decoding/encoding messages to the internalproxy
2. Use that struct for sending messages, when expecting responses it should have some sort of a "mini runtime" that reads messages until it receives the expected response (and stores unexpected responses, if there are threads waiting for it)
3. Split logic that happens in layer from logic that happens on internal proxy - all the listening/mirror socket stuff should happen in layer, but the actual connection should happen from internal proxy, since we won't have a thread to trigger connections from.
4. When an incoming connection happens, the internal proxy would connect to the layer "mirror" socket, then the layer can send a request to read the connection info from the internal proxy, to do the address manipulation logic.
5. The internal proxy might need to have a "global" state shared among different layer instances, for fork situations #864 
6. When outgoing connection happens, it will actually connect to the internal proxy socket, then it will do the forwarding. (with address manipulation)

Some initial drafts I had:
```rs
/// Message that originates from a layer hook, that needs some logic
/// in the internal proxy.
#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
pub enum HookMessage {}

/// Message that is an answer to a hook message.
#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
pub enum HookResponse {}

/// Layer to internal proxy messages
#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
pub enum LayerToProxyMessage {
    /// Message that is just forwarded to the agent.
    Upstream(ClientMessage),
    /// Hook originated message that we should handle
    Hook(HookMessage),
}

/// Messages from internal proxy to layer
#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
pub enum ProxyToLayerMessage {
    /// Messages that are passed as is from the agent.
    Downstream(DaemonMessage),
    /// Responses to hook messages
    HookResponse(HookResponse),
}
```

Razz4780:
Closes #1952 #1946 #1618 

New PR, because CI is broken in the old one #1960 

1. New crate `mirrord-intproxy`. Contains `IntProxy` - structure encapsulating proxy logic between one agent and multiple layer instances (originating from one ancestor - forks).
2. Changed `mirrord-console`. Now it exposes two kinds of loggers - async (spawns background `tokio::task`, usable in cli) and sync (no extra threads, just a `TcpStream` under a `Mutex`, usable in layer).
3. New layer does not spawn any extra threads at all. Communicates with the `IntProxy` using a global `TcpStream` under a `Mutex`.


t4lz:
> New PR, because CI is broken in the old one

If nothing happens here as well - maybe try resolving conflicts?

t4lz:
Debugging with this branch made me realise: now that a lot of the logic is in the intproxy, it would be very helpful to have a convenient way to get logs from the intproxy - maybe an option to write them to a file?

If it's not a 2 minute, 5 line change - it should be a separate issue though, in order not to delay merging this.

Razz4780:
> Debugging with this branch made me realise: now that a lot of the logic is in the intproxy, it would be very helpful to have a convenient way to get logs from the intproxy - maybe an option to write them to a file?
> 
> If it's not a 2 minute, 5 line change - it should be a separate issue though, in order not to delay merging this.

Hmm, you should be able to see them in `mirrord-console`.
We have some code in the CLI that prevents intproxy from printing logs to stderr, so it doesn't mangle user app's output. Maybe we can remove it? After all, layer also prints logs to stderr

t4lz:
fair enough, mirrord-console is a solution. The intproxy's stderr is redirected to `/dev/null` as part of its daemonization, let's leave it like that.

t4lz (approved):
it's time

PR #2033, opened 2023-10-25
Fix fs config docs markdown

t4lz:
No changes to the text, just indentations and numbering.

Before:

image

After:
image

screenshots are from different viewers, the differences in fonts and colours are irrelevant, just the paragraph structure.


meowjesty (approved):
:

PR #2038, opened 2023-10-27
Fixed python e2e test not running

aviramha:


t4lz (approved):

t4lz:
This was a bait and switch! 😄
Why are we removing this test completely now?

aviramha:
It never worked, and it should be an integration test anyway (that we already have)

t4lz:
`xstatfs` also does stuff in the agent, so I think it makes sense to test that roundtrip.

PR #1989, opened 2023-10-03
Operator Copy Resource

DmitryDodzin:
Initial Implementation

closes #1974

PR #2041, opened 2023-10-30
Copy resource feature in CLI

### LINKED ISSUE #1974: Copy pod/deployment - client side
Add client side which asks operator to copy a target (indicative error when operator doesn't exist) - mark this unstable.

Razz4780:
Closes #1974 

Work based on #1989, but a bit different approach. When `copy_pod` flag is enabled in config, the CLI will make a `create` request to the operator, asking it to create a new custom resource of type `copytarget`. This resource represents a job with a single pod, running a dummy container which never exits. The name of the newly created `copytarget` is the same as the name of the dummy pod, so the CLI gets its name from the `create` request. After that, the CLI can handle it as any other target. This allows for reusing previous `connect` flow and gives us port locks for free.

### DIFF
--- a/mirrord/config/src/feature.rs
+++ b/mirrord/config/src/feature.rs
@@ -83,7 +83,7 @@ pub struct FeatureConfig {
     ///
     /// Creates a new copy of the target. This copy will be used instead of the original target.
     /// This is feature requires mirrord operator.
-    #[config(env = "MIRRORD_COPY_TARGET_ENABLED", default = false)]
+    #[config(env = "MIRRORD_COPY_TARGET_ENABLED", default = false, unstable)]
     pub copy_target: bool,
 }
 
--- a/mirrord/config/src/feature.rs
+++ b/mirrord/config/src/feature.rs
@@ -3,6 +3,7 @@ use mirrord_config_derive::MirrordConfig;
 use schemars::JsonSchema;
 
 use self::{env::EnvConfig, fs::FsConfig, network::NetworkConfig};
+use crate::MirrordConfigSource;
 
 pub mod env;
 pub mod fs;
@@ -56,6 +57,7 @@ pub mod network;
 ///       },
 ///       "dns": false
 ///     },
+///     "copy_target": false
 ///   }
 /// }
 /// ```
@@ -76,6 +78,13 @@ pub struct FeatureConfig {
     /// ## feature.network {#feature-network}
     #[config(nested, toggleable)]
     pub network: NetworkConfig,
+
+    /// ## feature.copy_target {#feature-copy-target}
+    /// 
+    /// Creates a new copy of the target. This copy will be used instead of the original target.
+    /// This is feature requires mirrord operator.
+    #[config(env = "MIRRORD_COPY_TARGET_ENABLED", default = false)]
+    pub copy_target: bool,
 }
 
 impl CollectAnalytics for &FeatureConfig {
@@ -83,5 +92,6 @@ impl CollectAnalytics for &FeatureConfig {
         analytics.add("env", &self.env);
         analytics.add("fs", &self.fs);
         analytics.add("network", &self.network);
+        analytics.add("copy_target", self.copy_target);
     }
 }
--- a/mirrord/config/src/feature.rs
+++ b/mirrord/config/src/feature.rs
@@ -80,7 +80,7 @@ pub struct FeatureConfig {
     pub network: NetworkConfig,
 
     /// ## feature.copy_target {#feature-copy-target}
-    /// 
+    ///
     /// Creates a new copy of the target. This copy will be used instead of the original target.
     /// This is feature requires mirrord operator.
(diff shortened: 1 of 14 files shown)


mirrord/operator/src/client.rs
```diff
@@ -260,87 +303,156 @@ impl OperatorApi {
...
+
+        Ok(OperatorSessionConnection {
+            tx,
+            rx,
+            info: session_info,
+        })
+    }
+
+    #[tracing::instrument(level = "trace", skip(self))]
+    async fn copy_target(
```
t4lz, line 433: please add docs

t4lz (commented):
flushing first comments, will continue later

mirrord/operator/src/client.rs
```diff
@@ -260,87 +303,156 @@ impl OperatorApi {
...
+                )
+            }
+            (true, OperatorSessionTarget::Copied(target)) => {
+                let dt = &();
+                let namespace = self
+                    .target_namespace
+                    .as_deref()
+                    .unwrap_or_else(|| self.client.default_namespace());
+                let api_version = TargetCrd::api_version(dt);
+                let plural = TargetCrd::plural(dt);
```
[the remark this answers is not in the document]
t4lz: Or perhaps we could these methods:
```Rust

fn get_target_part_of_url(&self, target: &OperatorSessionTarget) -> String {
    match target {
        OperatorSessionTarget::Raw(target) => format!(
            "{}?on_concurrent_steal={}&connect=true",
            target.name(),
            self.on_concurrent_steal,
        ),
        OperatorSessionTarget::Copied(target) => format!(
            "{}?connect=true",
            target
                .meta()
                .name
                .as_ref()
                .expect("missing 'copytarget' name"),
        ),
    }
}
```

```Rust

fn get_target_api_resource_url(&self, target: OperatorSessionTarget) -> &str {
    match target {
        OperatorSessionTarget::Raw(_) => self.target_api.resource_url(),
        OperatorSessionTarget::Copied(_) => self.copy_target_api.resource_url(),
    }
}
```


then this match could be replaced with something like

```Rust
let url_target = self.get_target_part_of_url(&target);
if session.metadata.proxy_feature_enabled() {
    let dt = &();
    let namespace = self
        .target_namespace
        .as_deref()
        .unwrap_or_else(|| self.client.default_namespace());
    let api_version = TargetCrd::api_version(dt);
    let plural = TargetCrd::plural(dt);

    format!("/apis/{api_version}/proxy/namespaces/{namespace}/{plural}/{url_target}")
else {
    format!("{}/{url_target}", self.get_target_api_resource_url(target))
}
```

Use it if you like it, you don't have to.

mirrord/config/src/feature.rs
```diff
@@ -76,12 +78,20 @@ pub struct FeatureConfig {
     /// ## feature.network {#feature-network}
     #[config(nested, toggleable)]
     pub network: NetworkConfig,
+
+    /// ## feature.copy_target {#feature-copy-target}
+    ///
+    /// Creates a new copy of the target. This copy will be used instead of the original target.
+    /// This is feature requires mirrord operator.
+    #[config(env = "MIRRORD_COPY_TARGET_ENABLED", default = false, unstable)]
```
t4lz, line 86: I think at some point @aviramha wanted us to stop adding new env vars where not necessary, right?
```suggestion
    #[config(default = false, unstable)]
```

mirrord/config/src/lib.rs
```diff
@@ -418,6 +418,27 @@ impl LayerConfig {
                 ))?
             }
         }
+
+        if self.feature.copy_target {
```
t4lz, line 422: Let's also return an error if `self.operator` is `false`.

t4lz (changes requested):
Missing: indicative error when `copy_target` is set to `true` but there is no operator.

mirrord/config/src/feature.rs
```diff
@@ -76,12 +78,20 @@ pub struct FeatureConfig {
     /// ## feature.network {#feature-network}
     #[config(nested, toggleable)]
     pub network: NetworkConfig,
+
+    /// ## feature.copy_target {#feature-copy-target}
+    ///
+    /// Creates a new copy of the target. This copy will be used instead of the original target.
```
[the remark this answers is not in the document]
Razz4780: Extended the doc a bit, but I don't really know how to elaborate :sweat_smile: 
The client should already know what the "target" is

mirrord/operator/src/client.rs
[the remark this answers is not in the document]
Razz4780: I actually spotted a bug - `api_version` and `plural` should be constructed using `CopyTagetCrd`, so this narrows the repetition. I extracted getting `namespace` into a separate method, it's pretty concise now

mirrord/config/src/feature.rs
Razz4780: That's true. Removed the env var, we can always add it later

t4lz (approved):
Nice, thanks

Issue #1627, opened 2023-07-07
Flakey FastApi integration test on macOS


t4lz:
### Bug Description

https://github.com/metalbear-co/mirrord/actions/runs/5482641038/jobs/9988191840

### Steps to Reproduce

Trigger CI

### Your operating system and version

macOS   12.6.7 (Job runner)

### Local process

uvicorn

Razz4780:
Still happens, last time https://github.com/metalbear-co/mirrord/actions/runs/5937074021

infiniteregrets:
closing in favor of #1883, will reopen if it happens again

infiniteregrets:
happened again: https://github.com/metalbear-co/mirrord/actions/runs/6041689194/job/16395191019#step:37:181

gememma:
it looks like this was fixed by https://github.com/metalbear-co/mirrord/commit/7334b612447da88b3dbe109856627fbc8e57e554#diff-9e46d03ed8a6b605031da36b8a9792030826b634197652aadb5130b78715c561R47 and the issue was just left open by accident

PR #2042, opened 2023-10-30
Remove flakey test label from mirroring_with_http integration test

infiniteregrets:
#1627


eyalb181 (approved):

eyalb181 (changes requested):
Looks like this exact test is flaking out?

infiniteregrets:
> Looks like this exact test is flaking out?

yup, so turns out @meowjesty's pr didn't do it. now I wondering if it has to do with posix_spawn

PR #2054, opened 2023-11-03
Copy target scale down

### LINKED ISSUE #2053: Option to scale down target deployment (`copy target` feature) - client side
We want to add a feature that in case the target is copy + deployment, you can also specify that the deployment should scale down to 0 while the session is active.

Razz4780:
Closes #2053


mirrord/operator/src/client.rs
```diff
@@ -243,9 +245,21 @@ impl OperatorApi {
...
+        let target_to_connect = if config.feature.copy_target.enabled {
             let mut copy_progress = progress.subtask("copying target");
-            let copied = operator_api.copy_target(&metadata, raw_target).await?;
+
+            if config.feature.copy_target.scale_down {
+                let is_deployment = matches!(config.target.path, Some(Target::Deployment(..)));
+                if !is_deployment {
+                    progress.warning(
+                        "cannot scale down while copying target - target is not a deployment",
+                    )
```
t4lz, line 256: I think usually if the user specifies incompatible settings we exit with an error. So we should probably do the same here, unless @eyalb181 says otherwise. So I would also add a check to the config validation. But the check here is still necessary because if I'm not mistaken the target is not guaranteed to be known during config verification.

mirrord/config/src/lib.rs
```diff
@@ -143,7 +143,11 @@ const PAUSE_WITHOUT_STEAL_WARNING: &str =
 ///         "ignore_localhost": false,
 ///         "unix_streams": "bear.+"
 ///       },
-///       "dns": false
+///       "dns": false,
+///       "copy_target": {
+///         "enabled": true,
```
t4lz, line 148: I think there's no `enabled` field in the config schema, right?
```suggestion
```

mirrord/operator/src/crd.rs
```diff
@@ -135,9 +135,12 @@ pub enum OperatorFeatures {
...
 pub struct CopyTargetSpec {
-    /// Original target.
+    /// Original target. Only [`Target::Pod`] and [`Target::Deployment`] are accepted.
     pub target: Target,
     /// How long should the operator keep this pod alive after its creation.
     /// The pod is deleted when this timout has expired and there are no connected clients.
     pub idle_ttl: Option<u32>,
+    /// Should the operator scale down target deployment to 0 while this pod is alive.
+    /// Ignored if [`Target`] is not [`Target::Deployment`].
+    pub scale_down: Option<bool>,
```
t4lz, line 145: Why is this an option? Looks like it's always constructed with a Some, isn't it?

mirrord/operator/src/client.rs
Razz4780: I confirmed with Eyal, we want an error here. Updated

mirrord/config/src/lib.rs
Razz4780: Fixed :+1:

mirrord/config/src/feature/copy_target.rs
```diff
@@ -7,6 +7,17 @@ use crate::{
     util::MirrordToggleableConfig,
 };
 
+/// Default value for [`CopyTargetFileConfig::enabled`] when deserializing.
+/// This way in [`MirrordConfig::generate_config`] we can make distinction between following two
+/// cases:
+/// 1. `{ "features": {} }` - copy target is disabled, we get `enabled: false` from [`Default`]
+///    implemetation.
+/// 2. `{ "features": { "copy_target": {} } }` - copy target is enabled, we get `enabled: true` from
```
t4lz, line 15: This is not the nicest interface to enable a feature. I think it might be a bit weird for users.

I think we could make it take a single value (not an object) and it would be a nicer interface.
Something like:
`true`/`"on"`/`"copy"` -> copy without scaling down.
`"replace"`/`"scale-down"` -> copy with scaling down.
`false`/`"off"`/not specified -> no copy

What do you think?

Either way, doesn't have to be in this PR since the feature is marked as unstable so we can change the interface again later.

t4lz (changes requested):
Let's also add a check to `LayerConfig::verify` to return an error if the target is explicitly set to something other than a deployment and scale down is enabled.

mirrord/config/src/feature/copy_target.rs
Razz4780: I think eventually we'll have more customization here. Right now the user can use all of `true`, `false`, `{}`, `{"scale_down": ...}` or omit this config entirely; so I think it follows the pattern of other configs

t4lz: Oh, if they can just use `true` that's fine. Does `"copy_target": true` mean copy without scaling down?

mirrord/config/src/feature/copy_target.rs
```diff
@@ -0,0 +1,79 @@
...
+use mirrord_analytics::CollectAnalytics;
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+use crate::config::{ConfigContext, FromMirrordConfig, MirrordConfig, Result};
+
+/// Allows the user to target a pod created dynamically from the orignal [`target`](#target).
+/// The new pod inherits most of the original target's specification, e.g. labels.
+///
+/// ```json
```
t4lz, line 16: Let's maybe also add a `"copy_target": true` example somewhere?
It's good to have this example with all the features, but we should also document that `copy_target` can accepts booleans.

t4lz (approved):

PR #2057, opened 2023-11-06
Improve intproxy deps

### LINKED ISSUE #2039: Improve dependencies
After we added `intproxy` we still depend on it in layer/cli which means we bring all of it's dependencies as well just for few structs that can be in a small crate.

Razz4780:
Closes #2039 

1. `layer <-> proxy` protocol and codec extracted to a separate crate. Async stuff feature-gated
2. Async stuff in `mirrord-console` feature-gated


aviramha (approved):
Great stuff!

PR #2060, opened 2023-11-07
Local file filter now applies to directory listing 

aviramha:
Local file filter now applies to directory listing [regex] and not just underlying files


meowjesty (approved):
:+1:

Issue #2055, opened 2023-11-04
DNS Issue on Elixir macOS


aviramha:
Testing the mac build, it seems to work for external to the cluster network calls. The above demo app hits `https://example.com` and returns successfully:

```bash
$ curl localhost:8080/external
Call to https://example.com returned 200
```

Modifying the demo app to hit another pod in the same cluster & namespace seems to fail though:

```bash
## Start mirrord
$ MIX_DEBUG=1 MIRRORD_PROGRESS_MODE=off ./mirrord-debug-2047 exec -- mix run --no-halt
$ curl localhost:8080/internal -ivvv
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /internal HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
HTTP/1.1 500 Internal Server Error
< cache-control: max-age=0, private, must-revalidate
cache-control: max-age=0, private, must-revalidate
< content-length: 49
content-length: 49
< content-type: text/plain; charset=utf-8
content-type: text/plain; charset=utf-8
< date: Fri, 03 Nov 2023 15:36:16 GMT
date: Fri, 03 Nov 2023 15:36:16 GMT
< server: Cowboy
server: Cowboy

< 
* Connection #0 to host localhost left intact
Error: %MatchError{term: {:error, :econnrefused}}

### App logs
08:36:14.517 [info] Making internal (pod) request

08:36:17.575 [error] Error: %MatchError{term: {:error, :econnrefused}}

### Mirrord Agent logs (truncated to the only line that appears when i make the request)
2023-11-03T15:36:53.140421Z  WARN ThreadId(08) mirrord_agent::outgoing: interceptor_task -> Elapsed connect error Elapsed(
    (),
)
```

I execed into the mirrord agent, installed curl, and was able to hit my internal pod directly from there with no issues.

EDIT:

Debugged in Discord, it seems to be DNS resolution that's failing. Directly hitting an internal IP/port gave me a 200 as expected.

_Originally posted by @bradschwartz in https://github.com/metalbear-co/mirrord/issues/2047#issuecomment-1792666033_

aviramha:
Seems that elixir is using https://man7.org/linux/man-pages/man3/gethostbyname.3.html for resolving.. :Q

meowjesty:
Should've been fixed by #2071

PR #2071, opened 2023-11-13
Hook gethosbyname to resolve DNS for erlang/elixir.

meowjesty:
- Issue: #2055 

Adds a hook for [`gethostbyname`](https://www.man7.org/linux/man-pages/man3/gethostbyname.3.html) (yeah it's deprecated).

@aviramha found out that erlang/elixir uses this when resolving DNS.

There is a potential UB in how we cast some pointers that should be `*mut _` from `*const _`, but the manual of the function says that you should do a deep copy of the `*mut hostent` pointer we pass back to the user, so this shouldn't be a problem.

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -947,9 +947,9 @@ pub(super) fn gethostbyname(raw_name: Option<&CStr>) -> Detour<*mut hostent> {
             },
         );
 
-    let mut aliases_ptrs = aliases
+    let mut aliases_ptrs: Vec<*const i8> = aliases
         .iter()
-        .map(|alias| alias.as_ptr() as *const i8)
+        .map(|alias| alias.as_ptr())
         .collect::<Vec<_>>();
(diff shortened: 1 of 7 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -869,6 +895,88 @@ fn remote_hostname_string() -> Detour<CString> {
...
+/// **Safety**:
+/// See the [`GETHOSTBYNAME_ALIASES_PTR`] docs. If you see this function being called and some weird
+/// issue is going on, assume that you might've triggered the UB.
+#[tracing::instrument(level = "trace", ret)]
+pub(super) fn gethostbyname(raw_name: Option<&CStr>) -> Detour<*mut hostent> {
+    let name: String = raw_name
+        .bypass(Bypass::NullNode)?
+        .to_str()
+        .map_err(|fail| {
+            warn!("Failed converting `node` from `CStr` with {:#?}", fail);
```
aviramha, line 911: Node -> name

aviramha (approved):
Nitpick. Great job!

PR #2076, opened 2023-11-21
Added new configuration 'use_proxy'

aviramha:
Added new configuration 'use_proxy' that lets user disable usage of http/s proxy by mirrord even when env is set.

### DIFF
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -125,7 +125,7 @@ impl TcpConnectionStealer {
             command_rx,
             clients: HashMap::with_capacity(8),
             index_allocator: Default::default(),
-            stealer: TcpListener::bind((Ipv4Addr::UNSPECIFIED, 15555)).await?,
+            stealer: TcpListener::bind((Ipv4Addr::UNSPECIFIED, 0)).await?,
             iptables: None, // Initialize on first subscription.
             write_streams: HashMap::with_capacity(8),
             read_streams: StreamMap::with_capacity(8),
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -125,7 +125,7 @@ impl TcpConnectionStealer {
             command_rx,
             clients: HashMap::with_capacity(8),
             index_allocator: Default::default(),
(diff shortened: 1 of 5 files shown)


mirrord/config/src/lib.rs
```diff
@@ -299,6 +299,15 @@ pub struct LayerConfig {
     /// # internal_proxy {#root-internal_proxy}
     #[config(nested)]
     pub internal_proxy: InternalProxyConfig,
+
+    /// ## use_proxy {#root-use_proxy}
+    ///
+    /// When disabled, mirrord will remove `HTTP/S_PROXY` env variables before
+    /// doing any network requests. This is useful when the system sets a proxy
+    /// but you don't want mirrord to use it.
+    /// ```
```
Razz4780, line 308: Leftover ```

mirrord/config/src/lib.rs
```diff
@@ -299,6 +299,15 @@ pub struct LayerConfig {
     /// # internal_proxy {#root-internal_proxy}
     #[config(nested)]
     pub internal_proxy: InternalProxyConfig,
+
+    /// ## use_proxy {#root-use_proxy}
+    ///
+    /// When disabled, mirrord will remove `HTTP/S_PROXY` env variables before
```
Razz4780, line 305:
```suggestion
    /// When disabled, mirrord will remove `HTTP[S]_PROXY` env variables before
```
Just a suggestion

Razz4780 (changes requested):
I think we should clarify semantics of this setting:
1. Are the variables no longer used by the mirrord code?
2. Are the variables no longer accessible in the user code?
3. What if the variables are present in the remote target?

mirrord/agent/src/steal/connection.rs
```diff
@@ -125,7 +125,7 @@ impl TcpConnectionStealer {
             command_rx,
             clients: HashMap::with_capacity(8),
             index_allocator: Default::default(),
-            stealer: TcpListener::bind((Ipv4Addr::UNSPECIFIED, 0)).await?,
+            stealer: TcpListener::bind((Ipv4Addr::UNSPECIFIED, 15555)).await?,
```
Razz4780, line 128: What happened here?

aviramha: crap, good catch

Razz4780 (approved):

PR #2079, opened 2023-11-22
Fix incoming TCP with multiple pods

### LINKED ISSUE #2078: Incoming TCP not working with multiple pod deployments
### Bug Description

When targeting multiple pod deployments, incoming subscription request results in multiple responses being received by the internal proxy. The proxy expects only one, which causes a crash.

### Steps to Reproduce

1. Install operator in the cluster
2. Create a deployment with 2+ pods
3. Run mirrord targeting this deployment

### Backtrace

_No response_

### Relevant Logs

_No response_

---

### Local process

---

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Closes #2078 

## Notable change in logic

Before layer tokio removal PR, TCP subscription failure always terminated user's app. As part of the refactor I changed this behavior to returning regular errors from the `listen` hook. Given the problem described in issue description (intproxy receives multiple responses to a single subscription request) and that most of the time `mirrord_protocol::error::ResponseError` cannot be tracked back to the subscription request, this PR partially restores the previous behavior:

1. Layer makes a subscription request to intproxy.
2. Intproxy makes a subscription request to the agent/s.
3. Intproxy receives the first subscription result. It is either `Ok(port)`, `Err(PortAlreadyStolen(port))` or `Err(some_other_error)`. The first two variants trigger a response to the layer. The last variant terminates intproxy with an error, as we don't know the port. 
4. Intproxy receives subsequent subscription results. First two variants are ignored if they mention port which was already handled. The last variant terminates intproxy.

### DIFF
--- a/changelog.d/2078.fixed.md
+++ b/changelog.d/2078.fixed.md
@@ -1,0 +1,1 @@
+Fixed issues with mirroring incoming TCP connections when targeting multi-pod deployments.
--- a/mirrord/intproxy/protocol/src/lib.rs
+++ b/mirrord/intproxy/protocol/src/lib.rs
@@ -188,6 +188,8 @@ pub enum PortSubscription {
 pub struct PortUnsubscribe {
     /// Port on the remote pod that layer mirrored.
     pub port: Port,
+    /// Local address on which the layer was listening.
+    pub listening_on: SocketAddr,
 }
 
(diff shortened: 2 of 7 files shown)


aviramha (approved):
Great job!

PR #2081, opened 2023-11-23
Fixed `mirrord verify-config` output

Razz4780:
`targetless` target now serialized into the string `"targetless"` instead of `null`. This allows the IDEs to tell difference between a config with unset target and a config with target set to `targetless`


t4lz (approved):
Great, thanks!

Issue #355, opened 2022-09-06
Release CI: provide intellJ/vscode extension zip as a release artifact


infiniteregrets:
It can take a few hours for the released extension to be available on the marketplace, therefore once the release is done we should provide it as an artifact so that we can manually install and check if everything works on a timely basis

PR #2089, opened 2023-11-30
(operator) Check for valid license (when license type is Online).

meowjesty:
- Issue: [346](https://github.com/metalbear-co/operator/issues/346)

Adds a new type `LicenseValidity` to help checking if a license's certificate is expired or not.

Warns the user if their license is close to expiring.

Also adds some docs.

The allow/disallow based on license functionality is implemented in the operator PR [#355](https://github.com/metalbear-co/operator/pull/355).

### DIFF
--- a/changelog.d/2074.fixed.md
+++ b/changelog.d/2074.fixed.md
@@ -1,0 +1,1 @@
+Fix create react apps by adding node related files to default local
--- a/mirrord/layer/src/file/filter/read_local_by_default.rs
+++ b/mirrord/layer/src/file/filter/read_local_by_default.rs
@@ -46,6 +46,10 @@ pub fn regex_set_builder() -> RegexSetBuilder {
         // TODO: `node` searches for this file in multiple directories, bypassing some of our
         // ignore regexes, maybe other "project runners" will do the same.
         r"^.*/package.json$",
+        r"/\.yarnrc$",
+        r"/\.yarnrc.yml$",
+        r"/\.yarnrc.json$",
+        r"/node_modules",
         // asdf
         r".*/\.tool-versions$",
         // macOS


mirrord/auth/src/credentials.rs
```diff
@@ -70,6 +70,119 @@ impl AsRef<Certificate> for Credentials {
...
+            Self::Expired(expiration_date)
+        } else {
+            Self::Good(expiration_date)
+        }
+    }
+
+    /// Checks if this [`LicenseValidity`] represents a date that is close to expiring.
+    ///
+    /// We consider close if the `License` is 2 days away from its `expiration_date`.
+    pub fn close_to_expiring(&self) -> Option<i64> {
```
Razz4780, line 106: Instead of implementing this as a method, I suggest changing the enum to have 3 variants: `Good`, `CloseToExpire` and `Expired`. Rationale:
1. This method checks something, so it should probably return a boolean. 
2. An expired license is technically not close to be expired, so this method should return false.
3. Which would make it not practical :v

mirrord/operator/src/client.rs
```diff
@@ -171,7 +173,36 @@ impl OperatorApi {
...
+    ///
+    /// Afterwards we retrieve this client's [`Certificate`], which is validated in the
+    /// operator.
+    ///
+    /// The last important point is that we search for the target thing (pod or whatever) that
+    /// this client wants to impersonate. This action the first that is blocked by the
+    /// `license_check_middleware` allowing us to continue or not.
+    ///
+    /// Keep in mind that some failures here won't stop mirrord from hooking into the process
+    /// and working, it'll just work without the operator.
```
Razz4780, line 204: I wouldn't leak any info about the private operator repo here

mirrord/operator/src/client.rs
```diff
@@ -331,6 +383,9 @@ impl OperatorApi {
         }
     }
 
+    /// Reaches the `TargetProvider::get_resource` function, which is a route handler in the
```
Razz4780, line 386: Leaks info about private repo

mirrord/cli/src/connection.rs
```diff
@@ -17,6 +17,16 @@ pub(crate) struct AgentConnection {
 }
 
 /// Creates an agent if needed then connects to it.
+///
+/// First it checks if we have an `operator` in the [`config`](LayerConfig), which we do if the
+/// user has installed the mirrord-operator in their cluster, even without a valid license. And
+/// then we create a session with the operator with [`create_operator_session`].
+///
+/// If there is no operator, or the operator session creation failed (e.g. lack of license), then
```
Razz4780, line 25: I don't think this is true, we use `?` here on result from `OperatorApi::create_session`.

mirrord/operator/src/client.rs
```diff
@@ -187,6 +218,26 @@ impl OperatorApi {
...
+            if expiring_soon >= 0 {
+                let expiring_message = format!(
+                    "Operator license will expire soon, in {} days!",
+                    expiring_soon,
+                );
+
+                progress.warning(&expiring_message);
+                warn!(expiring_message);
+            }
+        }
```
Razz4780, line 239: Do you know if exisiting operator installations check expired licenses? If not, maybe we should add a check for this somewhere here and return an error (while making sure that endpoints like `status` are still accessible)? Ofc this check could be bypassed, but we can't do anything about it

Razz4780 (changes requested):

mirrord/auth/src/certificate.rs
```diff
@@ -24,6 +25,9 @@ where
 }
 
 /// Serializable `X509Certificate`
+///
+/// You can access the members of the certificate through it's [`AsRef`] implementation, this
```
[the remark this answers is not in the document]
meowjesty: Changed the docs.

mirrord/auth/src/credentials.rs
```diff
@@ -70,6 +70,119 @@ impl AsRef<Certificate> for Credentials {
...
+            LicenseValidity::Good(expiration_date) => {
+                let until_expiration = (Utc::now() - expiration_date).num_days();
+
+                if until_expiration <= 2 {
+                    Some(until_expiration)
+                } else {
+                    None
+                }
+            }
+            LicenseValidity::Expired(_) => Some(-1),
```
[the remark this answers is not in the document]
meowjesty: Don't like it either. Initially I used the `chrono::Days`, but it was bad.

I could just return `None` for this case, as we don't really care about the expired case right now. And I think I should've done just that (changing it).

mirrord/auth/src/credentials.rs
meowjesty: Guess I should read all the comments before addressing them lol.

Had this a 3-way enum, but on the operator side `CloseToExpiring` is meaningless, so everywhere I was `match`ing on it and doing the same thing as `Good`. So I removed in favor of the current 2-way enum.

I don't want to return a boolean here, so we can return how many days are left before expiration, to give a nice warning to the user (and not have to extract the expiration date and do the math on the warning call-site).

The main idea here is making it easy to get how many days are left, so we can display it.

mirrord/operator/src/client.rs
meowjesty: As far as I remember, there are no checks for license expiration.

We could check here, but wouldn't this break backwards compatibility? Like, if you update mirrord, but do not update the operator, now mirrord will perform this check and say that your license is invalid, even though the operator would allow you to continue.

Also, I think this would be kinda redundant, isn't it better to just have 1 place of authority (the operator itself in the case of these PRs)? We can talk more about how tight up we want this to be on discord.

mirrord/operator/src/client.rs
meowjesty: Moved these into the operator.

mirrord/auth/src/credentials.rs
Razz4780: Maybe something like this would be simplest
```rust
struct LicenseValidity(DateTime<Utc>);

impl LicenseValidity {
    fn is_good(&self) -> bool { ... };
    fn is_expired(&self) -> bool { ... };
    fn days_until_expiration(&self) -> Option<u32> { ... };
}
```
What do you think?

mirrord/cli/src/connection.rs
```diff
@@ -17,6 +17,16 @@ pub(crate) struct AgentConnection {
 }
 
 /// Creates an agent if needed then connects to it.
+///
+/// First it checks if we have an `operator` in the [`config`](LayerConfig), which we do if the
+/// user has installed the mirrord-operator in their cluster, even without a valid license. And
+/// then we create a session with the operator with [`create_operator_session`].
+///
+/// If there is no operator, or we the license is not good enough for starting an operator session,
```
Razz4780, line 25:
```suggestion
/// If there is no operator, or the license is not good enough for starting an operator session,
```

mirrord/cli/src/connection.rs
Razz4780: Good good

mirrord/auth/src/credentials.rs
meowjesty: Hmm if we go this route I think we could drop the type and just have these methods on the license, right?

meowjesty: Nope, not on the `License`, as we also need them for `LicenseInfoOwned`, so I guess they would have to be free functions.

Razz4780: This is also an option:
```rust
trait LicenseValidity {
    fn is_good(&self) -> bool { ... };
    fn days_until_expiration(&self) -> Option<u32> { ... };
}

impl LicenseValidity for DateTime<Utc> { ... }
```

meowjesty: I like this suggestion the most so far, but I'm not a fan of `bools` and prefer types.

Left both implementations for you to compare, on the operator side the change would be instead of `match license.validity` we do `if license.is_good() { ... } else { ... }`.

Leaving the final call on this to you.

Razz4780: I like `LicenseValidityV2` more. Sorry for being a pain in the lower back :sweat_smile:

Razz4780 (approved):
Looks good! Remember to remove one of the `LicenseValidity` variants.
@aviramha said that we want to catch expired licenses here as well, but I think this can be done in a subsequent PR

Razz4780 (approved):

PR #2094, opened 2023-12-04
Reduce some code duplication around agent connection and protocol

t4lz:
This might help doing stuff around the protocol more easily.
Opening this PR now keep PRs small and also to reduce the chance for merge conflicts.

### DIFF
--- a/mirrord/protocol/src/codec.rs
+++ b/mirrord/protocol/src/codec.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     io,
+    marker::PhantomData,
     sync::LazyLock,
 };
 
@@ -142,26 +143,37 @@ pub enum DaemonMessage {
     SwitchProtocolVersionResponse(#[bincode(with_serde)] semver::Version),
 }
 
-pub struct ClientCodec {
+pub struct ProtocolCodec<I, O> {
     config: bincode::config::Configuration,
+    /// Phantom just to associate the message types with the struct.
+    /// Implementing a trait to associate the types is not good enough, because we can't blanket
+    /// implement the foreign `Encoder`/`Decoder` traits, so we have to implement them for a
+    /// struct.
+    _phantom_incoming_message: PhantomData<I>,
+    _phantom_outgoing_message: PhantomData<O>,
 }
 
-impl ClientCodec {
+pub type ClientCodec = ProtocolCodec<DaemonMessage, ClientMessage>;
+pub type DaemonCodec = ProtocolCodec<ClientMessage, DaemonMessage>;
+
+impl<I, O> ProtocolCodec<I, O> {
     pub fn new() -> Self {
-        ClientCodec {
+        Self {
             config: bincode::config::standard(),
+            _phantom_incoming_message: Default::default(),
+            _phantom_outgoing_message: Default::default(),
         }
     }
 }
 
-impl Default for ClientCodec {
+impl<I, O> Default for ProtocolCodec<I, O> {
     fn default() -> Self {
-        ClientCodec::new()
+        Self::new()
     }
 }
 
-impl Decoder for ClientCodec {
-    type Item = DaemonMessage;
+impl<I: bincode::Decode, O> Decoder for ProtocolCodec<I, O> {
+    type Item = I;
     type Error = io::Error;
 
     fn decode(&mut self, src: &mut BytesMut) -> io::Result<Option<Self::Item>> {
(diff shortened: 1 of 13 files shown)


mirrord/protocol/src/codec.rs
```diff
@@ -142,77 +143,37 @@ pub enum DaemonMessage {
     SwitchProtocolVersionResponse(#[bincode(with_serde)] semver::Version),
 }
 
-pub struct ClientCodec {
+pub struct ProtocolCodec<I, O> {
```
meowjesty, line 146: Why not require `I: Decode` and `O: Encode`?

mirrord/protocol/src/codec.rs
```diff
@@ -142,77 +143,37 @@ pub enum DaemonMessage {
...
     config: bincode::config::Configuration,
+    /// Phantom just to associate the message types with the struct.
+    /// Implementing a trait to associate the types is not good enough, because we can't blanket
+    /// implement the foreign `Encoder`/`Decoder` traits, so we have to implement them for a
+    /// struct.
```
meowjesty, line 151: I'm being dumb, what do you mean here?

```rust
trait Codec {
  type In: Decode;
  type Out: Encode;
}
```

This wouldn't work? (not that there is anything wrong with the phantom approach btw, just want to get a better understanding)

mirrord/protocol/src/codec.rs
```diff
@@ -142,77 +143,37 @@ pub enum DaemonMessage {
...
-            }
-            Err(DecodeError::UnexpectedEnd { .. }) => Ok(None),
-            Err(err) => Err(io::Error::new(io::ErrorKind::Other, err.to_string())),
-        }
-    }
-}
-
-impl Encoder<ClientMessage> for ClientCodec {
-    type Error = io::Error;
+pub type ClientCodec = ProtocolCodec<DaemonMessage, ClientMessage>;
```
meowjesty, line 156:
```suggestion
/// Type alias used to deal with the messages we receive from the agent. You'll see this being consumed by our hooks.
pub type ClientCodec = ProtocolCodec<DaemonMessage, ClientMessage>;
```

meowjesty: Or something along those lines.

mirrord/protocol/src/codec.rs
```diff
@@ -142,77 +143,37 @@ pub enum DaemonMessage {
...
+pub type DaemonCodec = ProtocolCodec<ClientMessage, DaemonMessage>;
```
meowjesty, line 157:
```suggestion
/// Type alias that deals with the messages that are sent from mirrord-layer to the agent. These messages are _requests_ from our hooks, and the [`DaemonMessage`]s are the _responses_.
pub type DaemonCodec = ProtocolCodec<ClientMessage, DaemonMessage>;
```

mirrord/protocol/src/codec.rs
```diff
@@ -142,77 +143,37 @@ pub enum DaemonMessage {
...
-    }
-}
-
-pub struct DaemonCodec {
-    config: bincode::config::Configuration,
-}
-
-impl DaemonCodec {
+impl<I, O> ProtocolCodec<I, O> {
     pub fn new() -> Self {
```
meowjesty, line 160:
```suggestion
    fn new() -> Self {
```

Drop the `pub`, we're supposed to build this with `default`, right?

meowjesty (changes requested):
Refactor all the things.

Mostly some questions of things that I'm probably missing.

mirrord/protocol/src/codec.rs
t4lz: What I meant to say was that we can't implement `Codec` for  a generic `ClientCodec` and `DaemonCodec` structs to associate the correct message types with each of them and then just implement `Encoder` and `Decoder` for a generic `C: Codec`, because that `C` could be a foreign type and then we would be implementing a foreign trait (`Encoder`/`Decoder`) for a foreign type (anything that implements `Codec`). I think. But I now see that this comment is really not helpful at all and the code is clearer without it.
```suggestion
    /// Phantom fields to make this struct generic over message types.
```

mirrord/protocol/src/codec.rs
t4lz: There are many existing places where we build Client/Daemon codecs with `new`, but I now changed them all to default and remove this method altogether, see last commit.

mirrord/protocol/src/codec.rs
t4lz: I don't think that comment is correct. How about: 
```suggestion
// Codec to be used by the client side to receive `DaemonMessage`s from the agent and send `ClientMessage`s to the agent.
pub type ClientCodec = ProtocolCodec<DaemonMessage, ClientMessage>;
```

mirrord/protocol/src/codec.rs
t4lz:
```suggestion
// Codec to be used by the agent side to receive `ClientMessage`s from the client and send `DaemonMessage`s to the client.
pub type DaemonCodec = ProtocolCodec<ClientMessage, DaemonMessage>;
```

mirrord/protocol/src/codec.rs
meowjesty: Alright. I'm just a fan of type constraints, and here we always have this constraint.

But it's fine to leave it out, we don't interact with this type much anyway.

mirrord/protocol/src/codec.rs
meowjesty: Hmm I think what confused me most is that, this change is geared towards turning this more generic to help with the protocol versioning thingy, right?

If that's the case, could you leave a comment saying that?

meowjesty (approved):
:+1:

mirrord/protocol/src/codec.rs
t4lz: Should we really comment on a change in code? Future readers will only see the new state, not the old one, so it seems weird to me to discuss that change in the code.
I having a generic `ProtocolVersion` struct is already justified by the fact that we don't implement the `Encoder` and `Decoder` traits twice each with the exact same code. Should I comment that above the struct?

meowjesty: Nah, you're right. I was looking at this as a sort of "TODO: changing this for the protocol", but the change is already being made.

PR #2095, opened 2023-12-05
Fixed config verification in IDEs with 'copy_target' feature

Razz4780:
Very small change - handled the same way as other potential errors related to `target` (see logic above in the same function)

Closes https://github.com/metalbear-co/mirrord-intellij/issues/208
Closes https://github.com/metalbear-co/mirrord-vscode/issues/85


meowjesty (approved):
:+1:

PR #2097, opened 2023-12-05
Extend verify config output with compatible target types

### LINKED ISSUE #2096: Extend `verify-config` with some info useful in IDEs
#2095 fixes IDE errors when using `copy_target` feature without `target` set in the config. However, the user can still pick invalid targets in the popup. `mirrord verify-config` should return a list of available target types, so that the popup is not confusing.

Razz4780:
Closes #2096

### DIFF
--- a/changelog.d/2096.added.md
+++ b/changelog.d/2096.added.md
@@ -1,1 +1,1 @@
-`mirrord verify-config` not outputs a list of available target types.
+`mirrord verify-config` now outputs a list of available target types.
--- a/changelog.d/2096.added.md
+++ b/changelog.d/2096.added.md
@@ -1,0 +1,1 @@
+`mirrord verify-config` not outputs a list of available target types.
--- a/mirrord/cli/src/verify_config.rs
+++ b/mirrord/cli/src/verify_config.rs
@@ -96,6 +96,7 @@ enum VerifiedConfig {
         /// Improper combination of features was requested, but mirrord can still run.
         warnings: Vec<String>,
         /// Target types compatible with the source config.
+        /// Meant to be used by IDE plugins for customizing target selection.
         compatible_target_types: Vec<TargetType>,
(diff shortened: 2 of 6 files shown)


mirrord/cli/src/verify_config.rs
```diff
@@ -56,6 +57,30 @@ impl From<TargetConfig> for VerifiedTargetConfig {
...
+
+impl TargetType {
+    fn all() -> impl Iterator<Item = Self> {
+        [Self::Targetless, Self::Pod, Self::Deployment, Self::Rollout].into_iter()
+    }
+
+    fn compatible_with(&self, config: &FeatureConfig) -> bool {
+        match self {
+            Self::Targetless | Self::Rollout => !config.copy_target.enabled,
+            Self::Pod => !(config.copy_target.enabled && config.copy_target.scale_down),
```
meowjesty, line 78: Do we have these config matches documented anywhere (don't remember for the `copy_target` stuff)?

If we don't could you doc it saying why they're not allowed.

changelog.d/2096.added.md
```diff
@@ -0,0 +1 @@
+`mirrord verify-config` not outputs a list of available target types.
```
meowjesty, line 1:
```suggestion
`mirrord verify-config` now outputs a list of available target types.
```

typo?

meowjesty (approved):
:+1: 

Approved as I only really have doc comments.

changelog.d/2096.added.md
Razz4780: typo typo, thanks

mirrord/cli/src/verify_config.rs
Razz4780: Updated `copy_target` docs, now it mentions compatibility with target types :+1:

PR #2100, opened 2023-12-06
Make operator config optional

Razz4780:
`LayerConfig::operator` is now an `Option` - if the user is not explicit about the operator, try to use it but don't fail on errors that may occur when the operator is not installed


meowjesty (approved):
:+1:

PR #2111, opened 2023-12-12
Fix intproxy timeouts with lengthy build tools

### LINKED ISSUE #2101: Intproxy timeouts on first connection with long builds
### Bug Description

Some `mirrord exec` invocations fail due to a long build step before the actual application run (for example `mirrord exec -- go run ...` on a big project).

# Bug details
By default we skip loading the layer into some known build tools, e.g. `go`, `cc`, `cargo`. However, while these tools are running, the internal proxy is still waiting for the first layer to connect. If the build tool takes to long to finish, the proxy timeouts before the actual application starts and makes a connection. Next, the layer fails to connect to the proxy:
```
thread '<unnamed>' panicked at mirrord/layer/src/lib.rs:318:22:
failed to initialize proxy connection: IoFailed(Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })
```

# Possible solutions
1. In the CLI, open a pipe. Pass write fd to the user process, read fd to the proxy. Instead of a timeout, a separate task in the proxy blocks on reading from a pipe. When it receives EOF, it means that all write fds are down. This solution assumes that the user process passes the write fd across all execs.
2. Change the way we handle skipped processes in layer constructor. Right now we exit right away after collecting process environment (e.g. `std::env::current_exe`, `std::env::args`). Add an extra step, where we make a connection to the proxy.

### Steps to Reproduce

1. Download a big Go project (so that the compilation time is long enough)
2. Set `internal_proxy.start_idle_timeout` to some small value
3. Run the project with `mirrord exec -- go run ...`

### Backtrace

_No response_

### Relevant Logs

_No response_

---

### Local process

---

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Closes #2101 

The layer now initializes a new session with the internal proxy with every load (full, skipped, sip-only).
This could be further optimized by leaking the TCP socket descriptor across execs (no default CLOEXEC) and making the connection only once (storing the flag in env). However, the assumption that processes wouldn't close this descriptor brings us back to the first proposed solution. Not sure about it

### DIFF
--- a/changelog.d/2101.fixed.md
+++ b/changelog.d/2101.fixed.md
@@ -1,0 +1,1 @@
+Fixed an issue with internal proxy timing out when the user application spawns lengthy build processes.
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -65,7 +65,7 @@
 extern crate alloc;
 extern crate core;
 
-use std::{cmp::Ordering, ffi::OsString, panic, sync::OnceLock, time::Duration, net::SocketAddr};
+use std::{cmp::Ordering, ffi::OsString, net::SocketAddr, panic, sync::OnceLock, time::Duration};
 
 use ctor::ctor;
 use error::{LayerError, Result};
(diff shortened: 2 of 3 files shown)


aviramha (approved):

aviramha:
Awesome!

Issue #2099, opened 2023-12-06
golang syscall interception issue the app is using BoltDB


Ngalstyan4:
### Bug Description

Hi,

I have been trying to use `mirrord` with my application and and continually was getting errors when running with mirrord while not getting the same errors without mirrord

I tracked the issue down to etcd, and then to BoltDB inside etcd.

Consider the following golang application

```golang
package main

import (
        "C"
        "log"
        bolt "go.etcd.io/bbolt"
)

func main() {
        // Open the BoltDB file
        db, err := bolt.Open("hello.db", 777, nil)
        if err != nil {
                log.Fatal(err)
        }
        defer db.Close()
}
```
I compile this with `go mod tidy && go build`. 
I have tried `go build -ldflags "-linkmode 'external' -extldflags '-dynamic'"` as well ([details here](https://mt165.co.uk/blog/static-link-go/)). That does not change the outcome.

Compiler info:
```
$ go version
go version go1.20.3 linux/amd64
```

When run with mirrord, the above application produces an error, but runs successfully without mirrord

with mirrord:
```
$ rm -fr hello.db && ./target/debug/mirrord exec ./bolt
2023/12/06 08:30:59 invalid argument
```

w/o mirrord:
```
$ rm -fr hello.db && ./bolt
```
My understanding is the discrepancy originates in the `syscall6` interception in golang code.
Do you know what could cause this? Let me know if any other info is needed to reproduce this. Happy to help debug this, if needed.

### Steps to Reproduce

see above

### Your operating system and version

$ lsb_release -a No LSB modules are available. Distributor ID: Ubuntu Description:    Ubuntu 23.04 Release:        23.04 Codename:       lunar

### Local process

BoltDB

aviramha:
Hey @Ngalstyan4 -  I suspect it wouldn't work because of relative path - mirrord always bypasses relative paths. Could you try this with a full path instead?

aviramha:
Got it! It is most likely we don't handle errors well (i.e given invalid argument instead of other error) - can you run the same with env `RUST_LOG=mirrord=trace` and provide it redacted?

Ngalstyan4:
Here:

[mirrord.txt](https://github.com/metalbear-co/mirrord/files/13596650/mirrord.txt)

Sorry, I tried disabling colors via `TERM=dumb RUST_LOG_STYLE=never RUST_LOG=mirrord=trace`, but it seems that did not succeed so the file has some non-ascii characters

[...]

PR #2114, opened 2023-12-12
Fixes go syscallN being broken for mmap (and adds a few more file system sycalls for go).

meowjesty:
- Issue: #2099

Does not solve the issue itself, instead this allow mirrord to work with a boltdb app when the db file is bypassed.

Uses the [`syscalls`](https://docs.rs/syscalls/latest/syscalls/) crate to handle calling the syscalls for go.

Adds `pwrite64`, `pread64`, `fsync` and `fdatasync` hooks for go; found these out while testing a boltdb app with mirrord.

```go
package main

import (
	"C"
	bolt "go.etcd.io/bbolt"
)

func main() {
	db, err := bolt.Open("/tmp/mirrord.db", 0600, nil)

	if err != nil {
		print("bolt.Open: ")
		println(err.Error())
	} else {
		println(db)
		println("done!")
	}
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3980,6 +3980,7 @@ dependencies = [
  "serde_json",
  "socket2 0.5.5",
  "streammap-ext",
+ "syscalls",
  "test-cdylib",
  "tests",
  "thiserror",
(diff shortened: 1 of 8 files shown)


aviramha (approved):
Great job

PR #2120, opened 2023-12-13
Policies for Operator

t4lz:
for https://github.com/metalbear-co/operator/issues/361.

- Define a `MirrordPolicy` CRD.
- Write that CRD to the setup file.
- Create new ResponseError variant for reporting actions being forbidden by a policy.
  - This means that if a user is running a mirrord version that is older than the installed operator, and the operator blocks an operation because of a policy, mirrord would error because it couldn't parse the message. I think this is better than not sending back the block feedback, since the application would probably error weirdly either way and at least with the extra info we can make newer mirrord versions display helpful information when blocked by a policy. What do you think?

**Edit**: after writing this I realized maybe returning a permission denied IO error + remote log with block info is better and does not break protocol.

**I rebased/squashed and amended commits to group together related changes and make it easier to follow and see the big picture, so I recommend reading 1 commit at a time, and reading the commit messages first.**

Still missing:

- [x] E2E tests for more cases (targeting a deployment, selecting by labels, etc.)

### DIFF
--- a/mirrord/operator/src/crd.rs
+++ b/mirrord/operator/src/crd.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use self::label_selector::LabelSelector;
 
-mod label_selector;
+pub mod label_selector;
 
 pub const TARGETLESS_TARGET_NAME: &str = "targetless";
 
--- a/tests/src/operator/policies.rs
+++ b/tests/src/operator/policies.rs
@@ -2,16 +2,17 @@
 #![cfg(feature = "operator")]
 //! Test that mirrordpolicies work and features are blocked.
 
-use std::{
-    fmt::{Display, Formatter},
-    time::Duration,
-};
+use std::{collections::BTreeMap, time::Duration};
 
 use kube::Api;
-use mirrord_operator::crd::{BlockedFeature, MirrordPolicy, MirrordPolicySpec};
+use mirrord_operator::crd::{
+    label_selector::LabelSelector, BlockedFeature, MirrordPolicy, MirrordPolicySpec,
+};
 use rstest::{fixture, rstest};
 
-use crate::utils::{kube_client, service, Application, KubeService, ResourceGuard, TestProcess};
+use crate::utils::{
+    config_dir, kube_client, service, Application, KubeService, ResourceGuard, TestProcess,
+};
 
 /// Guard that deletes a mirrord policy when dropped.
 struct PolicyGuard(ResourceGuard);
@@ -32,33 +33,33 @@ impl PolicyGuard {
     }
 }
 
-enum Target {
-    Pod(String),
-    Deployment(String),
+enum CanSteal {
+    No,
+    OnlyWithFilter,
+    EvenWithoutFilter,
 }
 
-impl Display for Target {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Target::Pod(name) => write!(f, "{}/{}", "pod", name),
-            Target::Deployment(name) => write!(f, "{}/{}", "deployment", name),
-        }
+impl CanSteal {
+    fn should_work_without_filter(&self) -> bool {
+        matches!(self, EvenWithoutFilter)
+    }
+
+    fn should_work_with_filter(&self) -> bool {
+        matches!(self, EvenWithoutFilter | OnlyWithFilter)
     }
 }
 
+use CanSteal::*;
+
 struct PolicyTestCase {
     policy: MirrordPolicy,
-    service_a_can_steal: bool,
-    service_b_can_steal: bool,
-}
+    service_a_can_steal: CanSteal,
+    service_b_can_steal: CanSteal,
 
-// TODO: test policies:
-//  - with/out path pattern
-//  - with/out label selector
-//  - all steal / unfiltered steal
-//  - deployment target
-//  With every policy - check that stealing is blocked in selected target and not blocked in
-//  other target.
+    /// Use the deployment as a target?
+    target_deploy_a: bool,
+    target_deploy_b: bool,
+}
 
 /// Two services with non-random names, in the same random namespace.
 /// Using random namespace to isolate tests and separate policies (not have policies apply to other
@@ -93,7 +94,7 @@ pub async fn services(
     )
 }
 
-/// Block all stealing for all targets, should not work for either service.
+/// Block all stealing for all targets, stealing should not work for either service.
 fn block_steal_without_qualifiers() -> PolicyTestCase {
     PolicyTestCase {
         policy: MirrordPolicy::new(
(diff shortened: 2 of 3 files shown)


aviramha:
I think a specific error is better.

mirrord/operator/src/crd.rs
```diff
@@ -144,3 +144,23 @@ pub struct CopyTargetSpec {
...
+)]
+pub struct MirrordPolicySpec {
+    /// Specify the targets for which this policy applies, in the pod/my-pod deploy/my-deploy
+    /// notation. Targets can be matched using `*` and `?` where `?` matches exactly one
+    /// occurrence of any character and `*` matches arbitrary many (including zero) occurrences
+    /// of any character. If `None`, the policy will apply for all targets.
+    pub target: Option<String>,
+
+    /// Forbid clients from stealing traffic from the policy target.
+    pub block_steal: Option<bool>,
```
aviramha, line 165: I think this needs to be more future proof i.e

i.e have the field named `disallow`
then it accepts list of stuff it disallows, which the list is an enum, where one of the enum variants is steal, http steal so user can disable one or another or both

mirrord/operator/src/crd.rs
```diff
@@ -144,3 +144,23 @@ pub struct CopyTargetSpec {
...
+    kind = "MirrordPolicy",
+    root = "MirrordPolicyCrd",
+    namespaced
+)]
+pub struct MirrordPolicySpec {
+    /// Specify the targets for which this policy applies, in the pod/my-pod deploy/my-deploy
+    /// notation. Targets can be matched using `*` and `?` where `?` matches exactly one
+    /// occurrence of any character and `*` matches arbitrary many (including zero) occurrences
+    /// of any character. If `None`, the policy will apply for all targets.
+    pub target: Option<String>,
```
aviramha, line 162: it'd be better if we do it based on matching labels, but if it's a lot more work we can have it next version.

mirrord/operator/src/crd.rs
t4lz: Can do, although [this is already future proof](https://docs.rs/kube/latest/kube/derive.CustomResource.html#versioning):
> _if you add new properties behind options, and simply mark old fields as deprecated, then you can safely roll schema out changes without bumping the version_

So I think it's ok to decide by which is the better UI.
Changing to list unless you tell me otherwise.

About http steal - 1. why would an organization want to block filtering? 2. Separate options for header and path filters, or does one string block them both?

mirrord/operator/src/crd.rs
t4lz: Shouldn't be that hard, but I think @eyalb181 didn't want a label selector there.

eyalb181: @aviramha Wouldn't label selector require the customer to add labels to any deployments/pods they want to restrict?

mirrord/operator/src/crd.rs
t4lz: > where one of the enum variants is steal, http steal 

Do we realy want different variants for all possible steal rules (block all stealing, block unfiltered stealing, block filtered stealing, require header filter, require path filter, etc.), or do we want the steal variant to be an object itself so that users can specify like:

for blocking any stealing:
```yaml
disallow:
    - steal: any
```

for blocking unfiltered stealing:
```yaml
disallow:
    - steal: unfiltered
```

for blocking filtered stealing:
```yaml
disallow:
    - steal: filtered
```

If we just have a simple variant (string in the schema) for every mode users can create weird policies:
```yaml
disallow:
    - steal
    # redundant:
    - http steal
```

aviramha: I think the last option will be easiest, but I do understand what you mean about what you suggested.
The advantage of the last option is that you can then provide options to the specific - for example disallow specific ports or filter setc.

mirrord/operator/src/crd.rs
aviramha: Usually applications have an application label, or name so they can just use that, and it'd make more sense to have a label
`mirrord: nostealpolicy` to make pods like that.
I'm up for both options, just need to leave room (for example making not specifying target that it'd apply to all would be a lie if you'd then add labels)

mirrord/layer/src/error.rs
```diff
@@ -242,6 +242,7 @@ impl From<HookError> for i64 {
                 // this could be changed by waiting for the Subscribed response from agent.
                 ResponseError::PortAlreadyStolen(_port) => libc::EINVAL,
                 ResponseError::NotImplemented => libc::EINVAL,
+                ResponseError::Forbidden { .. } => libc::EACCES,
```
aviramha, line 245: imo we should kill the app when this happens, since it's not a "regular flow" - like the app code can try to handle this exception which will just lead to bad UX.

mirrord/operator/src/crd.rs
```diff
@@ -144,3 +145,50 @@ pub struct CopyTargetSpec {
...
+/// Features and operations that can be blocked by a `MirrordPolicy`.
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "kebab-case")] // StealWithoutFilter -> steal-without-filter in yaml.
+pub enum BlockedFeatures {
+    /// Blocks stealing traffic in any way (without or without filter).
+    Steal,
+    /// Blocks stealing traffic without specifying (any) filter. Client can still specify a
+    /// filter that matches anything.
+    StealWithoutFilter,
+    /// Blocks stealing traffic without an HTTP header filter. (Client can still specify a filter
```
aviramha, line 158: imo drop everything besides the first two. it wasn't asked for yet and not sure if that's how we want the options to look like.

mirrord/operator/src/crd.rs
t4lz: Question about selectors:
On deployment runs - should we check the individual pods' labels or just the deployment's?

mirrord/layer/src/error.rs
t4lz: Not sure what that means for this line, but I'll make sure to make the process terminate on blocked operations. Maybe server (operator) side. Maybe both.

aviramha: This line translates the forbidden error into libc errno as part of propagating it to the user app, which should be unreachable if we terminate on that message.

mirrord/operator/src/crd.rs
```diff
@@ -144,3 +145,38 @@ pub struct CopyTargetSpec {
...
+    version = "v1",
+    kind = "MirrordPolicy",
+    root = "MirrordPolicyCrd",
+    namespaced
+)]
+pub struct MirrordPolicySpec {
+    /// Specify the targets for which this policy applies, in the pod/my-pod deploy/my-deploy
+    /// notation. Targets can be matched using `*` and `?` where `?` matches exactly one
+    /// occurrence of any character and `*` matches arbitrary many (including zero) occurrences
+    /// of any character. If `None`, the policy will apply for all targets.
```
aviramha, line 173: If None and no match selection used

mirrord/layer/src/error.rs
t4lz: So what should be here instead? We still need that match arm. Should it be `unreachable!()`? I thought we didn't like panicking (`unwrap` etc.).

aviramha: Either unreachable or call exit with the exit handler (show message etc)

mirrord/operator/src/setup.rs
```diff
@@ -462,6 +462,13 @@ impl OperatorRole {
                     verbs: vec!["impersonate".to_owned()],
                     ..Default::default()
                 },
+                // Allow the operator to list mirrord policies.
+                PolicyRule {
+                    api_groups: Some(vec!["policies.mirrord.metalbear.co".to_owned()]),
+                    resources: Some(vec![MirrordPolicyCrd::plural(&()).to_string()]),
+                    verbs: vec!["list".to_owned()],
```
aviramha, line 469: I'd add get also, we'll end up needing it probably at some point :)

aviramha (approved):
Looks good!

Razz4780 (approved):

PR #2131, opened 2023-12-21
Flush console loggers after each log

t4lz:
While using the console alongside a debugger I noticed that some logs from lines that were executed were not showing up in the console. At some point the intproxy would exit without those logs ever being printed. Flushing after every log solved that.
It makes a big difference because the logs right before the process (either user app with layer or intproxy) exits are usually the most important and interesting.

I think efficiency is not a concern since the console is usually only used by us, not by users.


meowjesty (approved):
:toilet:.flush()

PR #2139, opened 2023-12-26
Remove "unstable" from ignore_localhost

eyalb181:


infiniteregrets (approved):
Missing change log

Are there no rust docs where we mentioned that’s it’s unstable?

eyalb181:
> Missing change log
> 
> Are there no rust docs where we mentioned that’s it’s unstable?

None that i've seen

PR #2144, opened 2023-12-28
Bump dependencies

aviramha:


mirrord/layer/Cargo.toml
```diff
@@ -26,7 +26,6 @@ libc.workspace = true
 nix.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
-# change to upstream once https://github.com/frida/frida-rust/pull/99 is merged
 frida-gum = { git = "https://github.com/metalbear-co/frida-rust", version = "0.13", features = ["auto-download"], branch="capstone_remove" }
```
t4lz, line 29: Can we change to upstream now? If not, should we maybe leave the comment there?

t4lz (approved):

PR #2149, opened 2024-01-01
Fix some test labels

infiniteregrets:
- operator tests were running verify-config tests, which are not needed there
- 1337 tests were not decorated
- ignoring one of the operator tests pending a fix


aviramha (approved):
Great!

tests/src/operator.rs
```diff
@@ -166,6 +166,7 @@ mod operator {
         assert!(!res.success());
     }
 
+    #[ignore]
```
aviramha, line 169: maybe add a comment

aviramha (approved):

aviramha (commented):
Nice, thank you.

aviramha:
Wait, maybe we can split into directory hierarchy to make it easier to follow?
Put all cli tests in cli dir, then `use` the cli dir module with cfg attr, reducing the need to specify for each test?

infiniteregrets:
> Wait, maybe we can split into directory hierarchy to make it easier to follow? Put all cli tests in cli dir, then `use` the cli dir module with cfg attr, reducing the need to specify for each test?

done

PR #2150, opened 2024-01-01
Update to 3.81.0

eyalb181:


infiniteregrets (approved):
we'll have to force merge, the extension e2e wont pass as we haven't had a release on them (we use last released tags, not main)

PR #2151, opened 2024-01-01
Remove http_header_filter

infiniteregrets:


changelog.d/+deprecate-header-filter.internal.md
aviramha, line 1: It should be `removed` instead of `internal` and should mention the alternative use.

changelog.d/+deprecate-header-filter.removed.md
```diff
@@ -0,0 +1 @@
+Remove http_header_filter
```
aviramha, line 1: Add more information on what users should use instead..
"Please use `http_filter` with key `header_filter` instead"

aviramha (approved):
Thank you!

PR #2160, opened 2024-01-09
Fix trace only mode

Razz4780:
1. `fork` and `gethostname` detours still tried to talk to the internal proxy
2. removed unnecessary checks in `layer_start` function, we don't call it from the `fork` detour anymore

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -194,8 +194,8 @@ fn layer_pre_initialization() -> Result<(), LayerError> {
     Ok(())
 }
 
-/// Initialize a new session with the internal proxy.
-/// Sets [`PROXY_CONNECTION`] if not in trace only mode.
+/// Initialize a new session with the internal proxy and set [`PROXY_CONNECTION`]
+/// if not in trace only mode.
 fn load_only_layer_start(config: &LayerConfig) {
     // Check if we're in trace only mode (no agent)
     let trace_only = std::env::var(TRACE_ONLY_ENV)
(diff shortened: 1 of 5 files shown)


mirrord/layer/src/lib.rs
```diff
@@ -495,8 +498,13 @@ pub(crate) unsafe extern "C" fn fork_detour() -> pid_t {
...
             tracing::debug!("Child process initializing layer.");
-            let parent_connection = unsafe { PROXY_CONNECTION.take() }
-                .expect("parent connection doesn't exist in fork");
+            let parent_connection = match unsafe { PROXY_CONNECTION.take() } {
+                Some(conn) => conn,
+                None => {
+                    tracing::debug!("Skipping new inptroxy connection (trace only)");
+                    return res;
+                }
+            };
```
t4lz, line 507: Nitpick: I think this could be a let-else:

```suggestion
            let Some(parent_connection) = unsafe { PROXY_CONNECTION.take() } else {
                tracing::debug!("Skipping new inptroxy connection (trace only)");
                return res;
            };
```

t4lz (approved):
Looks good, comments are optional suggestions, so approved.

mirrord/layer/src/lib.rs
```diff
@@ -195,8 +195,17 @@ fn layer_pre_initialization() -> Result<(), LayerError> {
 }
 
 /// Initialize a new session with the internal proxy.
-/// Sets [`PROXY_CONNECTION`].
+/// Sets [`PROXY_CONNECTION`] if not in trace only mode.
 fn load_only_layer_start(config: &LayerConfig) {
```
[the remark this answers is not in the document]
Razz4780: This is called from the layer constructor when we load to a skipped process, e.g. compiler.
Proxy connection is made here to prevent issues with long builds

mirrord/layer/src/lib.rs
Razz4780: Yeah, this gives me a syntax error because of the `unsafe` block :x

PR #2166, opened 2024-01-10
Remove `unstable` marking from `copy_target`

t4lz:


meowjesty (approved):
Can you add some docs on why it's not unstable anymore? /jk

PR #2170, opened 2024-01-11
Report namespace for operator sessions

infiniteregrets:
- https://github.com/metalbear-co/operator/issues/362
- [X] Test with old operator

### DIFF
--- a/mirrord/cli/src/operator.rs
+++ b/mirrord/cli/src/operator.rs
@@ -235,7 +235,7 @@ Operator License
         sessions.add_row(row![
             session.id.as_deref().unwrap_or(""),
             &session.target,
-            session.namespace.as_deref().unwrap_or(""),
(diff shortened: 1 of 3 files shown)


mirrord/cli/src/operator.rs
```diff
@@ -223,12 +223,19 @@ Operator License
...
+        "Namespace",
+        "User",
+        "Session Duration"
+    ]);
 
     for session in &status.sessions {
         sessions.add_row(row![
             session.id.as_deref().unwrap_or(""),
             &session.target,
+            session.namespace.as_deref().unwrap_or(""),
```
Razz4780, line 238: What do you think about printing `N/A` instead? Empty field might suggest a default namespace, which is not the case

infiniteregrets: great idea, ty

Razz4780 (approved):

Issue #2066, opened 2023-11-09
AWS SQS Queue Filtering


aviramha:
As part of the mirrord for Teams solution, we'd like developers to be able to "fork" a queue based on a filter.
This means the original queue would then split into sub-queues based on a message attribute, so each engineer can consume their relevant messages.
This is currently implemented for SQS, then Rabbit/Kafka (and more later!)

Current plan:

- Implement new CRD - MirrordForkSpec?
- You specify target deployment/rollout
- You can specify list of MirrordForkType - first and only one for now will be SQS
- You specify what env/configmap is used for controlling which SQS queue to use
- You specify how to filter - probably [message metadata is best ](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html)
- User can select fork specs in mirrord config file and specify its filter.
- When user launch session with fork specs, it checks if a fork is already active
- If a fork is active, it registers it's filter to it
- If it is not, it creates a modified configmap / deployment with changed envs to use a new SQS queue
- It spawns a pod (or the operator does it? I think diff pod would be better) that reads from the original queue, then applies the filter then writes to the different queues - "stable" and filtered.
- When filter is removed, remove it's designated queue.
- When no active filters, restore deployment and delete queues.

Example user config map:

## As env from configmap
```
  XX_SQS_FIFO_QUEUE_NAME:
    valueFrom:
      configMapKeyRef:
        key: XX_fifo
        name: sqs-queues
```

## AS mounted configmap

```
apiVersion: v1
data:
  config.yaml: |-
    queue_name: xx-queue-name
```

## Issue state:

- When a user sets a filter to any of the queues defined for their target, we'll set filters for all queues of that target, for the unspecified ones we will use a match-nothing filter.

### Follow-up issues / things we're leaving for future versions:
1. Deployment-wide incoming traffic: on the first version the local process will compete with the existing pods for traffic because we're using an altered copy-target. In the future the local process should receive deployment-wide traffic by default.

### TODOs:

- [x] Plan complete run flow with splitters (~1 day)
- [ ] Remaining Implementation:
  - [x] Session start:
    - [x] Session start when there is already a split:
      - [x] Extracting pod spec from target in SQS controller. (~1 day)
      - [x] Creating new config maps. (~1 day)
      - [x] Creating a copy pod with changed spec. (~1 day)
      - [x] Register filter.
    - [x] Session start for first client: (~2 days)
[...]

PR #2173, opened 2024-01-14
Queue Splitting Feature

t4lz:
#2066


mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,71 @@
...
+struct SplitQueuesConfig(HashMap<String, HashMap<String, QueueMessageFilter>>);
+
+enum QueueMessageFilter {
+    SQS(HashMap<String, SQSAttributeRuleSet>),
+}
+
+/// Set of rules to filter that are required to apply to a single field
+enum SQSAttributeRuleSet {
+    /// A single regex is enough for complex rules for a string.
+    StringAttribute(String),
```
aviramha, line 41: I think we can start with String attribute only to make it a bit easier

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,71 @@
...
+use mirrord_analytics::CollectAnalytics;
+use mirrord_config_derive::MirrordConfig;
+
+/// Maps name of a `MirrordQueueSplitter` resource to a mapping from a fork name to a filter
+/// definition.
+/// ```json
+/// {
+///   "feature": {
+///     "split_queues": {
+///       "whatever-q-splitter": {
```
aviramha, line 12: what's this?

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::HashMap;
+
+use mirrord_analytics::CollectAnalytics;
+use mirrord_config_derive::MirrordConfig;
+
+/// Maps name of a `MirrordQueueSplitter` resource to a mapping from a fork name to a filter
```
aviramha, line 6: fork -> splitter

mirrord/operator/src/crd.rs
```diff
@@ -189,3 +191,71 @@ pub struct MirrordPolicySpec {
...
+}
+
+/// The details of a queue that should be split.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+pub enum SplitQueue {
+    /// Amazon SQS
+    #[serde(rename_all = "camelCase")] // config_map -> configMap in yaml.
+    SQS {
+        /// AWS region to use.
+        region: String,
```
aviramha, line 222: I think we just use the region we're in?

aviramha: like I'm pretty sure it's part of the credentials provided to us

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+///       }
+///     }
+///   }
+/// }
+/// ```
+#[derive(MirrordConfig, Clone, Debug)]
+#[config(derive = "JsonSchema,Eq,PartialEq")]
+pub struct SplitQueuesConfig {
+    /// The name of the `MirrordQueueSplitter` resource to get queue details from.
+    splitter: String,
```
Razz4780, line 39: Not sure yet about the architecture - will the cluster always have one splitter? If not, does it make sense for the user app to use more than one splitter?

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+    queues: HashMap<String, QueueMessageFilter>,
+}
+
+#[derive(Serialize, Deserialize, JsonSchema, Clone, Debug, PartialEq, Eq)]
+#[serde(tag = "queue_type")]
+enum QueueMessageFilter {
+    SQS(HashMap<String, SQSAttributeRuleSet>),
```
Razz4780, line 46: This needs a doc, I think. To let the user know what this mapping actually means (I don't know what SQS attribute is, probably some devs don't either).

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[derive(Serialize, Deserialize, JsonSchema, Clone, Debug, PartialEq, Eq)]
+enum NumberRule {
+    GreaterThanInt(i64),
+    LesserThanInt(i64),
+    // GreaterThanFloat(f64),
+    // LesserThanFloat(f64),
+    EqualsInt(i64),
+    EqualsFloat(i64),
```
Razz4780, line 67:
```suggestion
    EqualsFloat(f64),
```
Right?

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+    DivisibleBy(i64),
+}
+
+#[derive(Serialize, Deserialize, JsonSchema, Clone, Debug, PartialEq, Eq)]
+enum BinaryAttributeRule {
```
Razz4780, line 72: I'd add a doc to explain encoding of these rules, because we're matching against binary data.

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[derive(Serialize, Deserialize, JsonSchema, Clone, Debug, PartialEq, Eq)]
+#[serde(tag = "queue_type")]
+enum QueueMessageFilter {
+    SQS(HashMap<String, SQSAttributeRuleSet>),
+}
+
+#[derive(Serialize, Deserialize, JsonSchema, Clone, Debug, PartialEq, Eq)]
+/// Set of rules to filter that are required to apply to a single field
+enum SQSAttributeRuleSet {
```
Razz4780, line 51:
```suggestion
#[serde(tag = "type", content = "value", rename_all = "snake_case")
enum SQSAttributeRuleSet {
```
Suggestion

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// A single regex is enough for complex rules for a string.
+    StringAttribute(String),
```
Razz4780, line 53:
```suggestion
    MatchesRegex(String),
```

mirrord/config/src/feature/split_queues.rs
t4lz: There can be one splitter per deployment (or Argo rollout). The app won't have to specify the splitter, it will be automatically derived from the target. I still haven't updated the code since the meeting where that was decided.

mirrord/config/src/feature/split_queues.rs
t4lz: Yes, thanks. Though typed rules will probably be dropped from this PR, following https://github.com/metalbear-co/mirrord/pull/2173#discussion_r1451977234.

mirrord/operator/src/client.rs
```diff
@@ -642,16 +646,26 @@ impl OperatorApi<PreparedClientCert> {
     where
         P: Progress,
     {
-        self.check_copy_target_feature_support(config)?;
+        self.check_feature_support(config)?;
 
-        let target = if config.feature.copy_target.enabled {
+        let target = if config.feature.copy_target.enabled
+            // use copy_target for splitting queues
```
Razz4780, line 652: I don't think this should be done transparently. I think if `copy_target` is not enabled it's a config verification error. @aviramha wdyt?

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct SplitQueuesConfig(Option<BTreeMap<QueueId, QueueFilter>>);
+
+impl SplitQueuesConfig {
+    pub fn is_set(&self) -> bool {
+        self.0.is_some()
+    }
+
+    /// Out of the whole queue splitting config, get only the sqs queues.
+    pub fn get_sqs_filter(&self) -> Option<HashMap<String, SqsMessageFilter>> {
+        self.0.as_ref().map(|queue_id2queue_filter| {
```
Razz4780, line 42: I don't know if I like the name of this lambda argument :V

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,96 @@
...
+            queue_id2queue_filter
+                .iter()
+                .filter_map(|(queue_id, queue_filter)| match queue_filter {
+                    QueueFilter::Sqs(filter_mapping) => {
+                        Some((queue_id.clone(), filter_mapping.clone()))
+                    }
+                    #[allow(unreachable_patterns)] // will become reachable when we add variants
+                    _ => None,
```
Razz4780, line 50: Maybe just remove this branch?

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,96 @@
...
+/// {
+///   "feature": {
+///     "split_queues": {
+///       "first-queue": {
+///         "queue_type": "SQS",
+///         "message_filter": {
+///           "wows": "so wows",
+///           "coolz": "^very .*"
+///         }
+///       },
```
Razz4780, line 21:
```suggestion
///       "SQS/first-queue": {
///         "message_filter": {
///           "wows": "so wows",
///           "coolz": "^very .*"
///         }
///       },
```
I think following `target` pattern would make this more concise

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub type SqsMessageFilter = BTreeMap<MessageAttributeName, AttributeValuePattern>;
+
+/// More queue types might be added in the future.
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
+#[serde(tag = "queue_type", content = "message_filter")]
+#[non_exhaustive]
+pub enum QueueFilter {
+    #[serde(rename = "SQS")]
+    Sqs(SqsMessageFilter),
```
Razz4780, line 83: We should document whether this set of filters means a conjunction or an alternative.
Also, it could use even a short description, like "Amazon Simple Queue Service"

mirrord/operator/src/crd.rs
```diff
@@ -137,9 +238,48 @@ pub struct Session {
...
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+pub enum NewOperatorFeature {
+    ProxyApi,
+    CopyTarget,
+    Sqs,
+    SessionManagement,
+    /// This variant is what a client sees when the operator includes a feature the client is not
+    /// yet aware of, because it was introduced in a version newer than the client's.
+    #[serde(other)]
+    FeatureFromTheFuture,
```
Razz4780, line 261:
```suggestion
    Unknown,
```

mirrord/operator/src/crd.rs
```diff
@@ -137,9 +238,48 @@ pub struct Session {
...
     ProxyApi,
+    // DON'T ADD VARIANTS - old clients won't be able to deserialize them.
+    // Add new features in NewOperatorFeature
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+pub enum NewOperatorFeature {
+    ProxyApi,
+    CopyTarget,
+    Sqs,
```
Razz4780, line 256:
```suggestion
    SqsQueueSplitting,
```

mirrord/operator/src/crd.rs
```diff
@@ -161,6 +301,8 @@ pub struct CopyTargetSpec {
     /// Should the operator scale down target deployment to 0 while this pod is alive.
     /// Ignored if [`Target`] is not [`Target::Deployment`].
     pub scale_down: bool,
+    /// queue id -> (attribute name -> regex)
+    pub sqs_filter: Option<HashMap<QueueId, SqsMessageFilter>>,
```
Razz4780, line 305: Maybe we should use a nested struct here? To group all queue filters in a separate type

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,190 @@ pub struct MirrordPolicySpec {
...
+/// Set where the application reads the name of the queue from, so that mirrord can find that queue,
+/// split it, and temporarily change the name there to the name of the branch queue when splitting.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")] // ConfigMap -> configMap in yaml.
+pub enum QueueNameSource {
+    ConfigMap(ConfigMapQueueNameSource),
+    EnvVar(String),
+}
+
+pub type OutputQueueName = String;
```
Razz4780, line 365: Looks like this is not used

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,190 @@ pub struct MirrordPolicySpec {
...
+
+/// The details of a queue that should be split.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(tag = "queueType")]
+// So that controllers that only handle 1 type of queue don't have to adapt when we add more queue
+// types.
+#[non_exhaustive]
```
Razz4780, line 372: I'm not a big fan of this, but I won't complain :v

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,190 @@ pub struct MirrordPolicySpec {
...
+    Sqs(QueueNameSource),
+}
+
+/// A workload that is a consumer of a queue that is being split.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")] // Deployment -> deployment in yaml.
+pub enum QueueConsumer {
+    Deployment(String),
+    Rollout(String),
+}
```
Razz4780, line 389: What about jobs/cronjobs/statefulsets? We're able to patch those as well

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,190 @@ pub struct MirrordPolicySpec {
...
+pub struct SqsSessionStatus {
+    /// Queue ID -> old and new queue names.
+    pub queue_names: BTreeMap<QueueId, QueueNameUpdate>,
+
+    // A bit redundant, because the registry resource status has the mapping from env var name
+    // to queue id, and `queue_names` has the mapping from queue id to name update, but, saving
+    // it here in the form that is useful to reader, for simplicity and readability.
+    /// Env var name -> old and new queue names.
+    pub env_updates: BTreeMap<String, QueueNameUpdate>,
+    // TODO: add updates to config map references.
```
Razz4780, line 494: todo :V

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,190 @@ pub struct MirrordPolicySpec {
...
+}
+
+/// The [`kube::runtime::wait::Condition`] trait is auto-implemented for this function.
+/// To be used in [`kube::runtime::wait::await_condition`].
+pub fn is_session_ready(session: Option<&MirrordSqsSession>) -> bool {
+    session
+        .map(|session| session.status.is_some())
+        .unwrap_or_default()
+}
```
Razz4780, line 503: I don't think this should be here, operator code seems like a better fit

Razz4780 (changes requested):
I think some updates are needed in `LayerConfig::verify_config` command - queue splitting is not compatible with all target types and require operator.

Also, I'd make sure that all crds are forward compatible, as we're gonna add more queue types. Old CLIs/operators must not fail to deserialize crds after update

mirrord/operator/src/crd.rs
t4lz: Changed + added "No time travel" to style guide.

mirrord/operator/src/crd.rs
t4lz: Should I just send over the whole `SplitQueuesConfig`, and do the partitioning by queue types in the operator's side?

mirrord/operator/src/crd.rs
t4lz: What's the problem with that?

mirrord/operator/src/crd.rs
t4lz: Let's add support for those in a separate step.
Is it ok to change the CRD later, or do I have to add them here even if we don't support them yet (and return an error if users try to use them)?

mirrord/operator/src/crd.rs
t4lz: We're going to add support for config map queue name sources in a separate step, this is a deliberate placeholder for when we do that, can I leave it there or should I remove it?

mirrord/operator/src/crd.rs
t4lz: Really? But it depends on the definition of the status struct, and might have to be changed if we change that struct. I think having it here makes more sense and could also help people remember to change it if they change the struct.

mirrord/operator/src/setup.rs
```diff
@@ -432,6 +438,20 @@ impl OperatorRole {
...
+                    verbs: vec!["patch".to_owned()],
+                    ..Default::default()
+                },
+                // For SQS controller to temporarily change Argo Rollouts to use changed queues.
+                PolicyRule {
+                    api_groups: Some(vec!["argoproj.io".to_owned()]),
+                    resources: Some(vec!["rollouts".to_owned()]),
+                    verbs: vec!["patch".to_owned()],
+                    ..Default::default()
+                },
```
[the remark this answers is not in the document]
t4lz: Let's do that in a separate PR.

mirrord/config/src/feature/split_queues.rs
Razz4780: Now that I look at it again, I think it would be too much of a hassle.
`QueueId` would have to store queue type and then we'd have to manually check that the value variant matches. `untagged` sounds a bit scary, because we'd have to ensure that no two variants can be parsed from the same json.
tl;dr it's ok as it is

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl FromMirrordConfig for SplitQueuesConfig {
+    type Generator = Self;
+}
+
+pub type MessageAttributeName = String;
+pub type AttributeValuePattern = String;
+
+/// A filter is a mapping between message attribute names and regexes they should match.
+/// The local application will only receive messages that match ALL the patterns.
```
Razz4780, line 72:
```suggestion
/// The local application will only receive messages that match **all** of the given patterns.
```

mirrord/config/src/feature/split_queues.rs
```diff
@@ -0,0 +1,97 @@
...
+/// This means, only messages that have all the `MessageAttributeName`s in the filter,
```
Razz4780, line 73:
```suggestion
/// This means, only messages that have **all** the `MessageAttributeName`s in the filter,
```

mirrord/operator/src/client.rs
Razz4780: Wdyt about a warning then?

mirrord/operator/src/crd.rs
Razz4780: Yeah, sounds good

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +342,206 @@ pub struct MirrordPolicySpec {
...
+    #[serde(rename = "SQS")]
+    Sqs(QueueNameSource),
+}
+
+/// A workload that is a consumer of a queue that is being split.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")] // Deployment -> deployment in yaml.
+pub enum QueueConsumer {
+    Deployment(String, Option<String>),
+    Rollout(String, Option<String>),
```
Razz4780, line 372: What does it mean when the container is not given? Needs a doc I think

Razz4780 (changes requested):

mirrord/operator/src/crd.rs
Razz4780: I think it's ok to change them later

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,188 @@ pub struct MirrordPolicySpec {
...
+    pub queues: BTreeMap<QueueId, SplitQueue>,
+
+    /// The resource (deployment or Argo rollout) that reads from the queues.
+    pub consumer: QueueConsumer,
+
+    /// These tags will be set for all temporary SQS queues created by mirrord for queues defined
+    /// in this MirrordWorkloadQueueRegistry, alongside with the original tags of the respective
+    /// original queue. In case of a collision, the temporary queue will get the value from the
+    /// tag passed in here.
+    pub tags: Option<HashMap<String, String>>,
```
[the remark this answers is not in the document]
t4lz: Moved, also changed operator side accordingly

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +341,188 @@ pub struct MirrordPolicySpec {
...
+    /// The name of the queue on AWS.
+    pub queue_filters: HashMap<QueueId, SqsMessageFilter>,
+
+    /// The target of this session.
+    pub queue_consumer: QueueConsumer,
+
+    /// The id of the mirrord exec session, from the operator.
+    // The Kubernetes API can't deal with 64 bit numbers (with most significant bit set)
+    // so we save that field as a string even though its source is a u64
+    pub session_id: String,
```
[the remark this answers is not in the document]
t4lz: done (in operator side)

mirrord/operator/src/client.rs
Razz4780: Ok, let it be `tracing::info!`

Razz4780 (approved):
Needs rebasing ^^

mirrord/operator/src/client.rs
t4lz: added.

mirrord/operator/src/crd.rs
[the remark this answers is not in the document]
t4lz: Changed to lower hex now, because I'm using it in resource names (of `MirrordSqsSession`) and resource names have to be lower case.

t4lz:
What changed besides the requested changes:
- Added `#[schemars(skip)]` `#[serde(skip_serializing)]` to unknown (`#[serde(other)]`) CRD enum variants, so that they are not part of the schema, but only deserialized when a peer has a newer version and sends a new, unknown variant.
- Changed `QueueConsumer` to a struct because that's now a better representation after having added a container.

Razz4780 (changes requested):
Only some small ser/de changes

mirrord/operator/src/crd.rs
```diff
@@ -199,3 +342,271 @@ pub struct MirrordPolicySpec {
...
+    pub name: String,
+    /// If a container is not specified, the workload queue registry will apply to every run that
+    /// targets any of the workload's containers.
+    pub container: Option<String>,
+    pub workload_type: QueueConsumerType,
+}
+
+/// A workload that is a consumer of a queue that is being split.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+pub enum QueueConsumerType {
```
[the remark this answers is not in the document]
t4lz: Why snake case? These variants are object types, which I think are UpperCamelCase in k8s.

image
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds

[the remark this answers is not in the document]
Razz4780: Ah, right. Sorry

Razz4780 (approved):

Issue #2077, opened 2023-11-21
Erlang - `Cluster.Strategy.Kubernetes` leads to error


t4lz:
### Bug Description

When using [`Cluster.Strategy.Kubernetes`]( https://hexdocs.pm/libcluster/Cluster.Strategy.Kubernetes.html) as a strategy users get an error:

```
{"time":"2023-11-09T18:17:39.456Z","severity":"error","message":"[libcluster:service_cluster] request to kubernetes failed!: {:failed_connect, [{:to_address, {'kubernetes.default.svc.cluster.local', 443}}, {:inet, [:inet], :nxdomain}]}","metadata":{"domain":["elixir"],"erl_level":"error"}}
```

### Steps to Reproduce

The declaration of the topology is

```Erlang
 topologies = [
            service_cluster: [
              strategy: Cluster.Strategy.Kubernetes,
              config: [
                kubernetes_ip_lookup_mode: :pods,
                kubernetes_node_basename: app_name,
                kubernetes_selector: "app=#{app_name}-web",
                kubernetes_namespace: namespace,
                polling_interval: polling_interval
              ]
            ]
          ]
```

### Your operating system and version

Mac OS 14.1

### Local process

User is developing in Elixir

### Additional Info

https://canary.discord.com/channels/933706914808889356/1169678897420439582/1174664744146309150

User works around the issue by changing their code to use `Elixir.Cluster.Strategy.DNSPoll`.

aviramha:
Should be retested after https://github.com/metalbear-co/mirrord/pull/2179

eyalb181:
@infiniteregrets Can you retest please?

PR #2179, opened 2024-01-17
add preadv and readv to fix erlang file reading

### LINKED ISSUE #2178: elixir file is empty
### Bug Description

when opening a file in Elixir ran by mirrord it returns empty content

### Steps to Reproduce

elixir 1.13.3-otp-24
erlang 24.2

```
mirrord exec -t pod/pod iex
File.read("/etc/hosts")
```

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS m3 14.2.1

### Local process

erlang

### Local process version

_No response_

### Additional Info

_No response_

aviramha:
Could be related to #2077 
Closes #2178

### DIFF
--- a/mirrord/layer/src/file/hooks.rs
+++ b/mirrord/layer/src/file/hooks.rs
@@ -9,7 +9,7 @@ use std::{ffi::CString, os::unix::io::RawFd, ptr, slice, time::Duration};
 
 use errno::{set_errno, Errno};
 use libc::{
-    self, c_char, c_int, c_void, dirent, off_t, size_t, ssize_t, stat, statfs, AT_EACCESS,
+    self, c_char, c_int, c_void, dirent, iovec, off_t, size_t, ssize_t, stat, statfs, AT_EACCESS,
     AT_FDCWD, DIR, EINVAL, O_DIRECTORY, O_RDONLY,
 };
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


mirrord/layer/src/file/hooks.rs
```diff
@@ -985,6 +985,84 @@ unsafe extern "C" fn realpath_darwin_extsn_detour(
...
+            // WARN: Must be careful when it comes to `EOF`, incorrect handling may appear as the
+            // `read` call being repeated.
+            ssize_t::try_from(read_amount).unwrap()
+        })
+        .unwrap_or_bypass_with(|_| FN_READV(fd, iovecs, iovec_count))
+}
+
+/// Hook for `libc::readv`.
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn preadv_detour(
```
meowjesty, line 1028: It's a bit hard to read this with a bunch of github actions warnings in the middle, but, doesn't this duplicate a bunch of stuff from `readv`?

Doesn't this warrant the `hookx_logic` we have for shared implementations?

meowjesty (changes requested):
`writev(["some", "comments"]);`

mirrord/layer/src/file/hooks.rs
```diff
@@ -985,6 +985,84 @@ unsafe extern "C" fn realpath_darwin_extsn_detour(
...
+pub(crate) unsafe extern "C" fn readv_detour(
+    fd: RawFd,
+    iovecs: *const iovec,
+    iovec_count: c_int,
+) -> ssize_t {
+    if iovec_count < 0 {
+        return FN_READV(fd, iovecs, iovec_count);
+    }
+    let iovs = slice::from_raw_parts(iovecs, iovec_count as usize);
+    let read_size: u64 = iovs.iter().map(|iov| iov.iov_len as u64).sum();
```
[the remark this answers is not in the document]
aviramha: I thought so, but seems that reduce expects return value to be same as Iterator::Item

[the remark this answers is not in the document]
aviramha: [Option](https://doc.rust-lang.org/std/option/enum.Option.html)<Self::[Item](https://doc.rust-lang.org/std/iter/trait.Iterator.html#associatedtype.Item)>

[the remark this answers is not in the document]
meowjesty: Ah, `fold` works then, initialize it with `0`.

mirrord/layer/src/file/hooks.rs
aviramha: it's not similar - but the copying to iovec is so I moved it into another function.

mirrord/layer/src/file/hooks.rs
```diff
@@ -985,6 +985,73 @@ unsafe extern "C" fn realpath_darwin_extsn_detour(
 //     result
 // }
 
+unsafe fn vec_to_iovec(bytes: &Vec<u8>, iovecs: &[iovec]) {
```
meowjesty: This is one weird function signature. Can't this be `&[u8]` slice?

meowjesty: Or maybe just pass the `Vec`?

meowjesty: Also, do we need the whole thing to be `unsafe`? Isn't it only the pointer copying bits?

mirrord/layer/src/file/hooks.rs
```diff
@@ -985,6 +985,73 @@ unsafe extern "C" fn realpath_darwin_extsn_detour(
...
+#[hook_guard_fn]
+pub(crate) unsafe extern "C" fn readv_detour(
+    fd: RawFd,
+    iovecs: *const iovec,
+    iovec_count: c_int,
+) -> ssize_t {
+    if iovec_count < 0 {
+        return FN_READV(fd, iovecs, iovec_count);
+    }
+    let iovs = slice::from_raw_parts(iovecs, iovec_count as usize);
```
meowjesty, line 1013: Will probably segfault if `iovecs` is `null` (well, not here, but actually when we try to do stuff with this pointer)?

What is normal `libc` behavior when this happens?

meowjesty (changes requested):
Potential segfault if `iovecs == NULL`?

mirrord/layer/src/file/hooks.rs
aviramha: I thought about passing `Vec` but that'd require keeping len and passing by ref is more gentle

mirrord/layer/src/file/hooks.rs
aviramha: I don't think it does null checks

meowjesty:
```
readv: Bad address
```

Original `readv` returns this when I call `readv(fd, NULL, 3)`.

mirrord/layer/src/file/hooks.rs
meowjesty: Alright!

meowjesty (changes requested):
Also, please add some tests.

A sample program for testing:

```c
#include <stdio.h>
#include <fcntl.h>
#include <sys/uio.h>

int main(int argc, char *argv[]) {
  char first[32], second[32];
  struct iovec iov[2];

  int fd = open("sample.c", O_RDONLY);
  if (fd == -1) {
    perror("open");
    return 1;
  }

  iov[0].iov_base = first;
  iov[0].iov_len = sizeof(first);
  iov[1].iov_base = second;
  iov[1].iov_len = sizeof(second);

  ssize_t result = readv(fd, iov, 2);
  if (result == -1) {
    perror("readv");
    return 1;
  }

  return 0;
}
```

t4lz (approved):

PR #2182, opened 2024-01-18
Add support for shebang containing spaces like asdf's node does

### LINKED ISSUE #2181: mirrord SIP fails on shell script with space
When executing on macOS a shell script containing

```sh
#! /usr/bin/env bash

echo 1
```
mirrord doesn't load into it, probably because of the extra space between the shebang and the command.

aviramha:
Closes #2181


mirrord/sip/src/lib.rs
```diff
@@ -341,13 +341,20 @@ mod main {
...
-                .split_whitespace()
-                .next()
-                .map(ToString::to_string)
+                .split_whitespace();
+
+            let first_token = iterator.next()?;
+            // if first_token is just #!, keep going
+            if first_token == "#!" {
+                let executable = iterator.skip_while(|s| s.is_empty()).next()?;
+                Some(format!("{first_token}{executable}"))
```
t4lz, line 354: Looks like we use the length of the result of this function to find the start fo the rest of the file in `patch_script`, so this could cause bugs.

Example:
first line is `#!<7 spaces>/bin/sh node`
then the new patched script would be
`#!<mirrord-tmp-dir>/bin/sh/bin/sh node`

So if we don't want to make a bigger change to the sip code, this function should return a string that is equal to the shebang as it is in the file, including whitespace.

t4lz (changes requested):

mirrord/sip/src/lib.rs
aviramha: I don't see it this way - reserving the behavior doesn't have any implication other than making our code more confusing..

t4lz (approved):
@aviramha I pushed a fix to this branch. I'm approving so we can merge it, but maybe you could review my commit so that someone that is not me reviews my code.
The commit message contains explanations, please ask me if anything isn't clear.

aviramha:
Great, thank you!

Issue #2186, opened 2024-01-20
Support R on MacOS (ARM64)


danielloader:
I've given Mirrord a go with an R application (just one of the demo applications hosted at https://github.com/rstudio/shiny-examples) - with the intention of working locally and mirroring the traffic from a kubernetes cluster back to my local environment. I'm using vscode but the issue exists with the CLI tool.

I'm using:
- MacOS 14.2
- M1 Pro (ARM64)
- R 4.1.3 
- VS Code 1.85.1

Reproduction steps:

1. Install R - [https://cran.r-project.org/bin/macosx/big-sur-arm64/base/R-4.1.3-arm64.pkg](https://cran.r-project.org/bin/macosx/big-sur-arm64/base/R-4.1.3-arm64.pkg)
2. Install Mirrord via homebrew as directed on quickstart
3. Execute `mirrord exec -- R` to just start the interactive REPL

```shell
$ mirrord exec -- R
⠉ mirrord exec
    ✓ Running on latest (3.84.1)!
    ✓ ready to launch process
      ✓ layer extracted
      ✓ operator not found
      ✓ agent pod created
      ✓ pod is ready                                                                                                                                              
Killed: 9
```

I'm not sure but it sounds like a SIP/signed binary patching issue but thought I'd raise it in case anyone else tries to use mirrord with R.

aviramha:
Hi, thanks for reporting this.
This is indeed a SIP issue. The R binary doesn't have x64 binary, which is our fallback for when SIP patching - in most cases we don't have arm to fallback to (which is the case) since macOS ships with arm64e.
Implementation of fallback to arm64 when available shouldn't be too hard, would you put it as a blocker from your side?

danielloader:
Glad there's an obvious cause. As to it being a blocker, in the strictest sense it's just flat out not working so I guess it counts - Not tried the x64 installation on MacOS with Rosetta2 - though I'm wondering if packaging would become a nightmare as I'm not sure if R looks towards the kernel to detect the uarch for package installation or the R hard coded values based on the installed variant. 

Is it worth leaving this issue open with an amended title to include the ARM64 aspect of the issue? 

As someone not familiar with how this problem is felt on the mirrord side I don't know if it's something even on the roadmap or feasible to fix so I'll defer to you as to what to do with the issue. Thanks for replying.

aviramha:
I did a bit of testing, we can detect cpu at runtime using sysctl crate
then
```rs

[...]

PR #2187, opened 2024-01-22
Allow x86_64 and arm64 when patching a thin binary

t4lz:
In a thin binay there is no arch choice anyway, so patch if we know how to patch that binary type (we know how to patch `x86_64` and `arm_64`).

Up until now we would try to (imperfectly) forbid patching thin `arm64` binaries on `x86_64` systems, but there is no point in that, as that is not a valid situation anyways, so it should not happen, and if it does happen, that's not a mirrord error, so we can let it fail naturally as it would without mirrord.

Addresses the specific case mentioned in #2186, but binary choice in fat binaries is still not correct.

### DIFF
--- a/changelog.d/2186.fixed.md
+++ b/changelog.d/2186.fixed.md
@@ -1,0 +1,1 @@
+Running R on macOS.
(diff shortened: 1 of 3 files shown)


aviramha (approved):
Nice one.

PR #2189, opened 2024-01-23
Change configuration parsing to be strict unallowing unknown fields

aviramha:
We had some users complaining about `http_header_filter` not working, while we removed it - many structs in our configuration already have the `deny_unknown_fields` flag but seems not all - we should keep adding this to any struct that is part of the mirrord configuration if not using `MirrordConfig` derive macro which does it automatically.


infiniteregrets (approved):
great catch!

PR #2192, opened 2024-01-23
Prepare the CLI for license renewals

### LINKED ISSUE #2190: Add some unique user id to certification requests sent to the operator

Razz4780:
Closes #2190 

Turns out we can't inject more data (e.g. unique user id) into the certification request, because it does not accept extensions. However, the key pair used to sign a certification request is generated only when the user has no certificate. If the user already has an expired certificate, the key pair is reused. Therefore, we can use the public key to identify users across renewals.

Changes:
1. Added license subscription id to the operator status CRD.
2. Adjusted `CredentialStore`: currently CLI stores user certificates in an (`operator license fingerprint` -> `certificate`) mapping. The fingerprint will change with every operator license renewal, so I also added a second mapping from operator license subscription id to signing key pair.


aviramha:
Title

mirrord/auth/src/credentials.rs
```diff
@@ -13,60 +13,53 @@ use crate::{
...
+/// Client credentials container for authentication with the operator.
+/// Contains a local [`KeyPair`] and an optional [`Certificate`].
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Credentials {
-    /// Generated Certificate from operator can be None when request isn't signed yet and will
-    /// result in is_ready returning false
-    certificate: Option<Certificate>,
-    /// Local Keypair for creating CertificateRequest for operator
+    /// Certificate generated by the operator based on the sent [`rfc2986::CertificationRequest`].
+    certificate: Certificate,
```
meowjesty, line 21: Optional `Certificate`?

mirrord/auth/src/credentials.rs
```diff
@@ -183,20 +176,58 @@ pub mod client {
     use super::*;
 
     impl Credentials {
-        /// Create `rfc2986::CertificationRequest` and send to operator to replace `Certificate`
-        /// inside of self.certificate making the Credentials ready
-        pub async fn get_client_certificate<R>(
-            &mut self,
+        /// Create a [`rfc2986::CertificationRequest`] and send it to the operator.
+        /// If the `key_pair` is not given, the request is signed with a randomly generated one.
+        pub async fn init<R>(
```
meowjesty, line 181: Any special reason not to call this `new`?

mirrord/auth/src/credential_store.rs
```diff
@@ -33,11 +34,12 @@ static CREDENTIALS_PATH: LazyLock<PathBuf> = LazyLock::new(|| CREDENTIALS_DIR.jo
...
 pub struct CredentialStore {
-    /// User-specified CN to be used in all new certificates in this store. (Defaults to hostname)
-    common_name: Option<String>,
     /// Credentials for operator
-    /// Can be linked to several diffrent operator licenses via diffrent keys
+    /// Can be linked to several different operator licenses via different keys.
+    #[serde(default)]
     credentials: HashMap<String, Credentials>,
+    #[serde(default)]
+    signing_keys: HashMap<String, KeyPair>,
```
meowjesty, line 42: >  Therefore, we can use the public key to identify users across renewals.

This is what's going on here, right?

meowjesty (changes requested):
Mostly me trying to understand it, but I think you missed an `Option`?

mirrord/auth/src/credential_store.rs
Razz4780: Yeah, whenever we create/retrieve user's `Credentials`, we associate the found key pair with operator's subscription id. Then, the operator's license is renewed - its fingerprint changes and we don't have any matching `Credentials`. But the subscription id does not change, so we look up the `signing_keys` mapping to find the key pair we used previously.
Also, subscription id is passed around as an `Option` to make the CLI backwards compatible

mirrord/auth/src/credentials.rs
Razz4780: I guess not, I just left the old name :shrug:

mirrord/auth/src/credentials.rs
Razz4780: It used to be an `Option` due to code logic (the flow included creating "dummy" `Credentials`), but it was never saved to the file before inserting the `Certificate`. I changed the code so that it doesn't have to be an `Option` now

mirrord/auth/src/credential_store.rs
meowjesty: Alright, just add this as docs then.

meowjesty (approved):
renewal prepared

Issue #976, opened 2023-01-23
Simplify SIP handling implementation.


t4lz:
Our SIP handling mechanism is written to deal with multi-step shebangs (when the shebang of a script is also a script with a shebang).

It turns out however that [that's not even allowed on darwin](https://stackoverflow.com/a/9988385/4671300), so we don't have to deal with that case, and we could simplify our implementation a bit.

Arnovsky:
Can I take this?

PR #2195, opened 2024-01-23
Fix spaced shebang bug, remove recursions

### LINKED ISSUE #2193: SIP doesn't work when there are spaces in a script's shebang
### Bug Description

Layer not injected.

### Steps to Reproduce

1. `spaced_shebang.sh`:
```bash
#! /usr/bin/env bash

echo hello
echo "what what" > what
cat what
rm what
```

2. `RUST_LOG=mirrord=info mirrord exec ./spaced_shebang.sh` (no logs about loading into process)

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS 14.2.1

### Local process

env, bash

### Local process version

_No response_

### Additional Info

_No response_

t4lz:
Closes #2193 and https://github.com/metalbear-co/mirrord/issues/976

The bug was that now (since #2182) that we also detect shebangs with spaces, when the sip-detection code was using the result of the shebang extraction as a path to the binary, that path contained whitespace, and so no sip was detected.

I changed the return type of the function that extracts the shebang to a struct to avoid further similar bugs.
We now test those cases with the public `sip_patch` function, (previously we only tested those cases with private functions, testing only part of the whole flow). 
I also eliminated all of the recursive-shebang stuff as we found out that's not allowed on darwin after originally writing that code.

### DIFF
--- a/changelog.d/2193.fixed.md
+++ b/changelog.d/2193.fixed.md
@@ -1,0 +1,1 @@
+Running scripts with whitespaces in the shebang.
--- a/mirrord/sip/src/error.rs
+++ b/mirrord/sip/src/error.rs
@@ -33,9 +33,6 @@ pub enum SipError {
     #[error("Can't perform SIP check - executable file not found at `{0}`")]
     FileNotFound(String),
 
-    #[error("Can't perform SIP check - there is a cycle in the shebang graph, found at `{0}`")]
-    CyclicShebangs(String),
-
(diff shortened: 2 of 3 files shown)


Razz4780 (approved):
Nothing suspicious, LGTM
One question - in multiple places we use `String` instead of `OsString`/`Vec<u8>` (paths, file contents). I wonder if we really need this casting and if it's not something that will bite us

PR #2207, opened 2024-01-28
Make go run work on macOS with go 1.21

### LINKED ISSUE #2202: `go run` does not work with Go 1.21 on macOS
### Bug Description

User gets an error from the go compiler when trying to run with mirrord

```
Build Error: go build -o /Users/<USER>/Workspace/worker/__debug_bin<NUMBERS> -gcflags all=-N -l ./main.go
go: error obtaining buildID for go tool compile: fork/exec /<MIRRORD-tmp-dir>/Users/<USER>/sdk/go1.21.0/pkg/tool/darwin_arm64/compile: no such file or directory (exit status 1)
```

### Steps to Reproduce

1. Install Go with the Go installer.
2. `find /var/folders -delete -name mirrord-bin -type d 2>/dev/null`
3. `MIRRORD_PROGRESS_MODE=off mirrord exec -t deploy/py-serv-deployment go -- run go-e2e`
   ```
   go: error obtaining buildID for go tool compile: fork/exec /var/folders/4l/810mmn597cx7zy5w2bk11clh0000gn/T/mirrord-bin/usr/local/go/pkg/tool/darwin_arm64/compile: no such file or directory
   ```

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS

### Local process

❯ file /Users/<USER>/sdk/go1.21.0/bin/go /Users/<USER>/sdk/go1.21.0/bin/go: Mach-O 64-bit executable arm64

### Local process version

_No response_

### Additional Info

I think maybe what happens is that the go binary does some file operations where we should strip our SIP-patch path, but we currently don't when loaded as SIP-only.

Update:
Seems like go 1.21 does not get the executable's path on macOS via, `_NSGetExecutablePath`, that we hook, but [by reading the memory after argv and envv](https://github.com/golang/go/blob/d278d5bbdddd0e976c272d1dd3ecc41eeb37daf9/src/runtime/os_darwin.go#L457-L472), so the executable knows its real location in our SIP directory, and then tries to read some files relative to that location, but they don't exist there. The nicest solution would be to find a way to make go get the original path as its executable path. Otherwise we would have to add path-strippnig hooks for all file hooks on SIP-only loading mode, which would suck.

t4lz:
Fixes #2202 

1. Strip our tmp dir from the path in exec.
2. Do code signing on a temp file, and move the ready and signed file to the final location, because two processes might try to execute a binary at the same time and both of them trying to sign the same binary simultaneously leads to an error.

Also manually tested that this fixes the issue the way the user discovered it: by running a go application in vscode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "ahash",
- "base64 0.21.5",
+ "base64 0.21.7",
  "bitflags 2.4.1",
  "brotli",
  "bytes",
@@ -40,7 +40,7 @@ dependencies = [
  "flate2",
  "futures-core",
  "h2",
- "http",
+ "http 0.2.11",
  "httparse",
  "httpdate",
  "itoa",
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.43",
+ "syn 2.0.48",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


aviramha (approved):

aviramha:
Great work!

PR #2217, opened 2024-02-05
Disabled unix sockets being wrongfully sent to the agent when socket isn't connected

aviramha:
Disabled unix sockets being wrongfully sent to the agent when socket isn't connected

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -1137,9 +1137,9 @@ pub(super) fn send_to(
 
     // we don't support unix sockets which don't use `connect`
     if (destination.is_unix() || user_socket_info.domain == AF_UNIX)
-        && user_socket_info.user_socket_info.state != SocketState::Connected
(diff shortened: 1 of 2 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -1135,6 +1135,13 @@ pub(super) fn send_to(
         .remove(&sockfd)
         .ok_or(Bypass::LocalFdNotFound(sockfd))?;
 
+    // we don't support unix sockets which don't use `connect`
+    if (destination.is_unix() || user_socket_info.domain == AF_UNIX)
+        && user_socket_info.user_socket_info.state != SocketState::Connected
+    {
+        return Bypass::UnixSocket(destination);
```
t4lz, line 1142: Doesn't make a difference right now, but I think the intended bypass variant for this case is `Domain`, not `UnixSocket`, according to their respective documentations.

t4lz (approved):

PR #2226, opened 2024-02-09
Allow for HTTP upgrades in intproxy (HTTP filter)

### LINKED ISSUE #2224: Adjust intproxy to allow stealing upgraded connections
To handle stolen incoming connections, internal proxy spawns two types of tasks - `RawInterceptor`s for raw tcp connections and `HttpInterceptor`s for http connections. Since we want to be able to switch from http to raw tcp (#2126), some adjustments are needed

Razz4780:
Closes #2224 

* `RawInterceptor` and `HttpInterceptor` merged into one type, that freely switches between being a TCP proxy and an HTTP gateway
* Added version requirement to `mirrord-protocol`
* Added a unit test for HTTP upgrade


mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -368,13 +323,9 @@ impl IncomingProxy {
...
-        if let Some(handle) = self.interceptors_raw.get(&interceptor_id) {
-            Some(&handle.subscription)
-        } else if let Some(handle) = self.interceptors_http.get(&interceptor_id) {
-            Some(&handle.subscription)
-        } else {
-            None
-        }
+        self.interceptors
+            .get(&interceptor_id)
+            .map(|h| &h.subscription)
```
meowjesty, line 328:
```suggestion
            .map(|handle| &handle.subscription)
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,53 +1,89 @@
...
+    match version {
+        Version::HTTP_2 => {
+            let (sender, connection) =
+                http2::handshake(TokioExecutor::default(), TokioIo::new(target_stream)).await?;
+            tokio::spawn(connection);
 
-                Ok(Self::V2(sender))
-            }
+            let (upgrade_tx, upgrade_rx) = oneshot::channel();
+            let _ = upgrade_tx.send(Err(Error::UpgradeNotSupported(version)));
```
meowjesty, line 55: I know that this might be a sad request, but feels a bit weird to preemptively send a "can't upgrade" here like this, and I'm assuming whoever takes the `upgrade_rx` deals with it when calling `recv`?

Why not express this as an `Option<Receiver>` in the return of `handshake`?

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -0,0 +1,557 @@
...
+    Raw(Vec<u8>),
+}
+
+/// Messages produced by the [`Interceptor`] when it runs as a [`BackgroundTask`].
+#[derive(Debug)]
+pub enum MessageOut {
+    /// Response received from the user application.
+    Http(HttpResponseFallback),
+    /// Data received from the user application.
+    Bytes(Vec<u8>),
```
meowjesty, line 39: Let's either call both `Raw` or both `Bytes`.

I prefer `Raw` if it's `Raw(Vec<u8>)`, as `Bytes` might get confused with the `bytes` crate stuff.

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -0,0 +1,557 @@
...
+
+impl From<Vec<u8>> for MessageIn {
+    fn from(value: Vec<u8>) -> Self {
+        Self::Raw(value)
+    }
+}
+
+/// Errors that can occur when [`Interceptor`] runs as a [`BackgroundTask`].
+#[derive(Error, Debug)]
+pub enum Error {
```
meowjesty, line 56: I'm not a big fan of using rust names like this (same for how we do `type Result = ...` in some places).

I would suggest changing it to `InterceptorError`, but I don't think we have a style rule around this yet, so it's up to you.

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -0,0 +1,557 @@
...
+            let new_conn = match conn {
+                GenericConnection::Raw(raw_conn) => raw_conn
+                    .run(message_bus)
+                    .await?
+                    .map(GenericConnection::Http),
+                GenericConnection::Http(http_conn) => http_conn
+                    .run(message_bus)
+                    .await?
+                    .map(GenericConnection::Raw),
+            };
```
meowjesty, line 130: Wait, I don't get this.

You create a binding `conn = GenericConnection::Raw(RawConnection { stream })`, then `match conn`? Why are you `match`ing if it's guaranteed to be `RawConnection` (also, why even have the `GenericConnection` enum at all?

meowjesty (changes requested):
Still looking into `incoming/interceptor.rs`, but I found some minor stuff to comment on.

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -0,0 +1,557 @@
...
+                                request,
+                                StatusCode::BAD_GATEWAY,
+                                "mirrord",
+                            )
+                        }),
+                ),
+            }
+    }
+
+    async fn send(&mut self, req: HttpRequestFallback) -> Result<HttpResponseFallback> {
```
meowjesty, line 234: docs please

mirrord/intproxy/src/proxies/incoming/interceptor.rs
Razz4780: It's in a loop :eyes:

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -0,0 +1,557 @@
...
+                    message_bus.send(MessageOut::Http(res)).await;
+                }
+            }
+        }
+
+        Ok(None)
+    }
+}
+
+struct RawConnection {
```
meowjesty, line 296: docs

mirrord/intproxy/src/proxies/incoming/interceptor.rs
Razz4780: Sometimes this `conn` variable is `RawConnection`, sometimes it's `HttpConnection`.

meowjesty (changes requested):
Just docs missing requests I guess.

mirrord/intproxy/src/proxies/incoming/interceptor.rs
meowjesty: Oh, that's why it's `mut`.

I always struggle with patterns like this :face_with_spiral_eyes:

Razz4780: Yeah, I tried to implement `async fn run(&mut self, ...)` on `GenericConnection`, but I need to move field values when converting. Couldn't work it out :<

mirrord/intproxy/src/proxies/incoming/interceptor.rs
Razz4780: Changes both to `InterceptorError`/`InterceptorResult` :+1:

mirrord/intproxy/src/proxies/incoming/interceptor.rs
Razz4780: Added

mirrord/intproxy/src/proxies/incoming/interceptor.rs
Razz4780: Added

meowjesty (approved):
:rocket:

PR #2231, opened 2024-02-13
Change tests to not require multi threaded runtime

aviramha:


t4lz (approved):

tests/src/traffic/steal.rs
```diff
@@ -2,15 +2,15 @@
...
+        io::{BufRead, Read, Write},
+        net::SocketAddr,
         time::Duration,
     };
 
     use kube::Client;
     use reqwest::{header::HeaderMap, Url};
     use rstest::*;
-    use tokio::time::sleep;
+    use tokio::{io::BufReader, net::TcpStream, time::sleep};
```
t4lz, line 13: Maybe this alone is the fix? we were using blocking inside async code.

tests/src/traffic/steal.rs
```diff
@@ -567,7 +567,7 @@ mod steal {
 
         let addr = SocketAddr::new(host.trim().parse().unwrap(), port as u16);
         let mut stream = TcpStream::connect(addr).await.unwrap();
-        stream.write(tcp_data.as_bytes()).unwrap();
+        stream.write(tcp_data.as_bytes()).await.unwrap();
```
t4lz:
```suggestion
        stream.write_all(tcp_data.as_bytes()).await.unwrap();
```

Like @Razz4780 and Clippy suggested.

PR #427, opened 2022-09-17
Intellij Ext: Fix "access is allowed from event dispatch thread"

infiniteregrets:
refer #317 
refer: https://stackoverflow.com/questions/54826606/intellij-plugin-access-is-allowed-from-event-dispatch-thread-only
https://intellij-support.jetbrains.com/hc/en-us/community/posts/7602780460306-Access-is-allowed-from-event-dispatch-thread-only

PR #2235, opened 2024-02-13
Added session management routes + crd + cli interaction.

meowjesty:
- Issue: [#217](https://github.com/metalbear-co/operator/issues/217)
- operator PR: [#427](https://github.com/metalbear-co/operator/pull/427)

Creates a new cli command to forcefully kill one or more `OperatorSession`s.

Adds a new `CustomResource`, as this new command is a a separate route in the operator from the ones that we had.


mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// `mirrord operator session kill --id {id}`: kills the operator session specified by `id`.
+#[tracing::instrument(level = "trace", ret)]
+pub(super) async fn operator_session_kill_one(id: u64) -> Result<()> {
+    let (progress, api, sub_progress) = operator_session_prepare().await?;
+
+    sub_progress.print("killing session with id {session_id}");
+
+    let result = api
+        .delete(&format!("kill_one/{id}"), &DeleteParams::default())
```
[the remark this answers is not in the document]
meowjesty: I did this on purpose, to make it easier when searching.

If you see a function `kill_one`, search for it and you should find all `kill_one` related code. Same for the others. At this point only the `kill` command is less searchable due to dropping the `_one` part.

As this is internal stuff, I would rather keep it easier for us to find/search for this stuff, than to look more RESTish (as you kinda know where I stand on REST...).

wdyt?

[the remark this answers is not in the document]
aviramha: It's not internal, it's an API we expose in the operator.. K8s adheres to REST and we should try to follow the guidelines.

mirrord/cli/src/config.rs
```diff
@@ -223,6 +223,39 @@ pub(super) enum OperatorCommand {
...
+
+    /// Kills _inactive_ sessions, might be useful if an undead session is still being stored in
+    /// the session storage.
+    #[clap(hide(true))]
+    RetainActive,
+}
+
+/// Parses the operator session id from hex (without `0x` prefix) into `u64`. It also accepts the
+/// non-hex version of the id.
+fn hex_id(raw: &str) -> Result<u64, String> {
```
Razz4780, line 255: Session id is generated randomly in `mirrord_operator` crate. What if its hex representation is also a valid decimal representation? :X
E.g. the user could pass `0x1111` id as `1111` (because we omit the `0x` prefix) and it would parse to a different value.
`mirrord operator status` always returns session ids in hex, so I think it would be better to parse only hex values

mirrord/cli/src/config.rs
```diff
@@ -223,6 +223,39 @@ pub(super) enum OperatorCommand {
...
+}
+
+/// `mirrord operator session` family of commands.
+///
+/// Allows the user to forcefully kill operator sessions, use with care!
+#[derive(Debug, Subcommand, Clone, Copy)]
+pub(crate) enum SessionCommand {
+    /// Kills the session specified by `id`.
+    Kill {
+        /// Id of the session.
```
Razz4780, line 240: I'd mention format, like `given as hexadecimal string`

mirrord/config/src/target.rs
```diff
@@ -290,6 +293,19 @@ pub struct PodTarget {
     pub container: Option<String>,
 }
 
+impl Display for PodTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!(
+            "{}{}",
+            self.container
+                .clone()
```
Razz4780, line 301:
```suggestion
                .as_ref()
```
No need to clone the string, I think

mirrord/config/src/target.rs
```diff
@@ -290,6 +293,19 @@ pub struct PodTarget {
     pub container: Option<String>,
 }
 
+impl Display for PodTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!(
```
Razz4780, line 298:
```suggestion
        write!(f, "{}{}", ...)
```

mirrord/operator/src/client.rs
```diff
@@ -175,6 +178,13 @@ pub struct OperatorSessionConnection {
     pub info: OperatorSessionInformation,
 }
 
+/// Allows us to access the operator's [`SessionCrd`] [`Api`].
+pub async fn session_api(config: Option<String>) -> Result<Api<SessionCrd>> {
+    let kube_api: Client = create_kube_api(false, config, None).await.unwrap();
```
Razz4780, line 183: No `unwrap` please :X

mirrord/operator/src/crd.rs
```diff
@@ -121,6 +121,25 @@ pub struct Session {
     pub locked_ports: Option<Vec<(u16, String, Option<String>)>>,
 }
 
+/// Resource used to access the operator's session management routes.
+///
+/// - `kind = Session` controls the how kube generates the route, in this case it becomes
```
Razz4780, line 126:
```suggestion
/// - `kind = Session` controls how [`kube`] generates the route, in this case it becomes
```

mirrord/cli/src/operator/session.rs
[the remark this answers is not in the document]
Razz4780: I'd suggest 
`DELETE /sessions/id`
`DELETE /sessions`
`DELETE /sessions?active=false`

mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,102 @@
...
+    Api<mirrord_operator::crd::SessionCrd>,
+    ProgressTracker,
+)> {
+    let progress = ProgressTracker::from_env("Operator session action");
+
+    let session_api = session_api(None)
+        .await
+        .inspect_err(|fail| error!("Failed to even get sessio_api {fail:?}!"))?;
+
+    let sub_progress = progress.subtask("preparing to execute session operation...");
```
Razz4780, line 21:
```suggestion
    let sub_progress = progress.subtask("preparing...");
```

mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,102 @@
...
+async fn operator_session_prepare() -> Result<(
+    ProgressTracker,
+    Api<mirrord_operator::crd::SessionCrd>,
+    ProgressTracker,
+)> {
+    let progress = ProgressTracker::from_env("Operator session action");
+
+    let session_api = session_api(None)
+        .await
+        .inspect_err(|fail| error!("Failed to even get sessio_api {fail:?}!"))?;
```
Razz4780, line 19:
```suggestion
        .inspect_err(|fail| error!(error = ?fail, "Failed to create session API"))?;
```

mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,102 @@
...
+    mut progress: ProgressTracker,
+) -> Result<()> {
+    if let Some(status) = result {
+        sub_progress.success(Some(&format!(
+            "session operation completed with {status:?}"
+        )));
+    }
+
+    sub_progress.success(Some("session operation finished"));
+    progress.success(Some("Done with session stuff!"));
```
Razz4780, line 40: Something else maybe? :sweat_smile:

mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Handles the cleanup part of progress after an operator session command.
+#[tracing::instrument(level = "trace", ret)]
+fn operator_session_finished(
+    result: Option<Status>,
+    mut sub_progress: ProgressTracker,
+    mut progress: ProgressTracker,
+) -> Result<()> {
+    if let Some(status) = result {
+        sub_progress.success(Some(&format!(
+            "session operation completed with {status:?}"
```
Razz4780, line 35: Don't show debug impl, maybe just `... completed successfuly (status code ...)` or `... failed (status code ...)`

Razz4780 (changes requested):
I feel like this also needs another operator feature, like `session_management: bool`. Otherwise the users will see real ugliness :skull:

mirrord/cli/src/config.rs
meowjesty: I'm afraid some users would put `0x` thinking it's required, as we just show them `Session ID`.

mirrord/cli/src/config.rs
meowjesty: Only hex now, good catch.

mirrord/cli/src/operator/session.rs
meowjesty: Changed to give out more messages based on status.

meowjesty:
> I feel like this also needs another operator feature, like `session_management: bool`. Otherwise the users will see real ugliness 💀

Ideally we would have some credentials/permissions system in place, so a random user would **not** be able to just `kill-all` sessions, instead they could only `kill` their own. While some admin level user could do whatever.

We could add a route that enables/disables session management, but this seems like extra annoyance for users, due to us not having a system in place to handle this stuff?

aviramha:
> > I feel like this also needs another operator feature, like `session_management: bool`. Otherwise the users will see real ugliness 💀
> 
> 
> 
> Ideally we would have some credentials/permissions system in place, so a random user would **not** be able to just `kill-all` sessions, instead they could only `kill` their own. While some admin level user could do whatever.
> 
> 
> 
> We could add a route that enables/disables session management, but this seems like extra annoyance for users, due to us not having a system in place to handle this stuff?

That's controlled via k8s RBAC. To use the route users need to have the delete verb on the sessions resource.

Razz4780:
> > > I feel like this also needs another operator feature, like `session_management: bool`. Otherwise the users will see real ugliness 💀
> > 
> > 
> > Ideally we would have some credentials/permissions system in place, so a random user would **not** be able to just `kill-all` sessions, instead they could only `kill` their own. While some admin level user could do whatever.
> > We could add a route that enables/disables session management, but this seems like extra annoyance for users, due to us not having a system in place to handle this stuff?
> 
> That's controlled via k8s RBAC. To use the route users need to have the delete verb on the sessions resource.

To be honest, I was not thinking about access control at all. Rather compatibility with older operators, which don't have these routes at all. Like with `copy_target`, the operator should announce that it supports killing sessions, so that the users don't see generic `404` or `503` (?) when it doesn't

mirrord/cli/src/operator/session.rs
Razz4780: Splendid :laughing:

mirrord/cli/src/config.rs
Razz4780: Okok. Just leaving this without specifying format will work as well, given that operator status prints ids in hex :+1:

aviramha:
Ah it could be nice the cli to check for feature existence, yes.

mirrord/operator/src/client.rs
```diff
@@ -160,6 +160,9 @@ pub struct OperatorApi {
     client: Client,
     target_api: Api<TargetCrd>,
     copy_target_api: Api<CopyTargetCrd>,
+
+    /// [`Api`] to the operator's [`SessionCrd`] resources (routes).
+    _sessions_api: Api<SessionCrd>,
```
Razz4780, line 165: I think we can remove this, right?

meowjesty:
> Ah it could be nice the cli to check for feature existence, yes.

Hmm we could just convert an `404` for hitting a missing route in an error saying "The operator doesn't seem to support this operation.". Maybe the message could also contain the operator version? I think the whole handling for this can be done on the cli-side?

meowjesty:
```sh
killing all sessions
  x Session operation `kill-all` failed!
    x `operator session kill-all` is not supported by the mirrord-operator found in your cluster, consider updating it!                                                                                             2024-02-26T18:02:45.664986Z ERROR mirrord::operator::session: not_supported="`operator session kill-all` is not supported by the mirrord-operator found in your cluster, consider updating it!"
Error:   × Failed to connect to the operator, probably due to RBAC: getting status failed: ApiError: "": Failed to parse error data (ErrorResponse { status: "404 Not Found", message: "\"\"", reason: "Failed to parse
  │ error data", code: 404 })

(...)
```

Progress now shows a nicer error if operator version doesn't support the session ops.

mirrord/cli/src/operator/session.rs
```diff
@@ -0,0 +1,150 @@
...
+    let result = api
+        .delete("inactive", &DeleteParams::default())
+        .await
+        .inspect_err(|kube_fail| match kube_fail {
+            kube::Error::Api(response) if response.code == 404 && response.reason.contains("parse") => {
+                let not_supported = "`operator session retain-active` is not supported by the mirrord-operator found in your cluster, consider updating it!";
+
+                sub_progress.failure(Some(not_supported));
+                progress.failure(Some("Session operation `retain-active` failed!"));
+                error!(not_supported);
```
Razz4780, line 137: This is always printed and looks kinda ugly, I'd remove it

Razz4780 (changes requested):
Looks mostly good, but there are 3 things:
1. Do we really need the `response.reason.contains("parse")` part when checking operator response? I think checking status code should be enough.
2. Same error handling logic is repeated in all 3 command handlers
3. Using `OperatorApiError::KubeError` with `OperatorOperation::GettingStatus` is not a good idea. User sees `Failed to connect to the operator, probably due to RBAC: getting status failed` in the CLI output :x You can use `OperatorApiError::UnsupportedFeature` for this. It would require fetching the operator status first (you'd need the operator version), but I think it's even better - we can check if the operator is installed before attempting to kill sessions

mirrord/cli/src/operator/session.rs
meowjesty: I'll use this box to address your review comment.

```sh
executing `mirrord operator kill-all`
  x Session management operation failed!
    x mirrord operator foobar does not support feature session management                                                                                                                                           Error:   × Feature `session management` is not supported in mirrord operator foobar.
  help: 
        
        - If you're still stuck:
        
        >> Please open a new bug report at https://github.com/metalbear-co/mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help in #mirrord-help
        
        >> Or email us at hi@metalbear.co
```

Doesn't this error look beautiful now or what? It's just missing the proper operator version (wip).

meowjesty: > 1. Do we really need the response.reason.contains("parse") part when checking operator response? I think checking status code should be enough

I think we do need it. I'm filling the error in the operator side, so now we're going to have nicer error messages coming from there (at least for this feature), and as such, we need to distinguish from the "random generic kube api error due to version mismatch", and the proper `reason` errors being returned.

meowjesty:
```sh
executing `mirrord operator kill-all`
  x Session management operation failed!
    x mirrord operator 3.76.0 does not support feature session management                                                                                                                                           Error:   × Feature `session management` is not supported in mirrord operator 3.76.0.
  help: 
        
        - If you're still stuck:
        
        >> Please open a new bug report at https://github.com/metalbear-co/mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help in #mirrord-help
        
        >> Or email us at hi@metalbear.co

```

We have version!

meowjesty: And the error in case we don't find an id to remove:

```
executing `mirrord operator kill --id 257`
  x Session operation failed!
    x `mirrord operator kill --id 257` failed due to `Session 257 not found!` with code `404`!                                                                                                                      Error:   × Operator returned a failure status for `mirrord operator kill --id 257`!
  help: Session 257 not found!
```

Razz4780 (approved):
Wonderful :cat2:

PR #2243, opened 2024-02-16
Support copying rollouts

Razz4780:
1. Removed target type checks from the `copy_target` flow, because these are done in the operator anyway. Thanks to removing the checks from the CLI, we don't have to add another `feature` to operator status (whether copying rollouts is enabled)
2. Added permissions for `rollouts/scale` to operator setup

### DIFF
--- a/changelog.d/+removed-copy-target-checks.changed.md
+++ b/changelog.d/+removed-copy-target-checks.changed.md
@@ -1,0 +1,2 @@
+1. mirrord CLI now does not check target type when the `copy_target` feature is enabled. The check is now done only in the operator.
+2. `mirrord operator setup` not includes permissions to read and change rollouts scale.
--- a/mirrord/config/src/lib.rs
+++ b/mirrord/config/src/lib.rs
@@ -20,7 +20,6 @@ use config::{ConfigContext, ConfigError, MirrordConfig};
 use mirrord_analytics::CollectAnalytics;
 use mirrord_config_derive::MirrordConfig;
 use schemars::JsonSchema;
-use target::Target;
 use tera::Tera;
 use tracing::warn;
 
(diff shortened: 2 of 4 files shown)


changelog.d/+removed-copy-target-checks.changed.md
```diff
@@ -0,0 +1,2 @@
+1. mirrord CLI now does not check target type when the `copy_target` feature is enabled. The check is now done only in the operator.
```
aviramha, line 1: why skip the client side change?

changelog.d/+removed-copy-target-checks.changed.md
```diff
@@ -0,0 +1,2 @@
...
+2. `mirrord operator setup` not includes permissions to read and change rollouts scale.
```
aviramha, line 2: remember to update it in the helm chart repo please.

changelog.d/+removed-copy-target-checks.changed.md
Razz4780: New operators will support rollouts, older operators will not. We can do the check if we extend operator status again (new field for feature flag), but I'm not sure if there's any gain here, since the operator already does this check and returns a nice error

changelog.d/+removed-copy-target-checks.changed.md
Razz4780: https://github.com/metalbear-co/charts/pull/39

changelog.d/+removed-copy-target-checks.changed.md
aviramha: Got it, sounds good.

aviramha (approved):

PR #2263, opened 2024-02-22
change kubeconfig path expansion to use env as well

### LINKED ISSUE #2262: use relative path to kubeconfig files in mirrord.json
we have lots of different git repositories that work with different kubernetes clusters; so we tend to check in local kube config files to each git repository. e.g. relative to the git directory they may be in: `./kubeconfigs/dev/.kube/config` for the `Dev` environment

however I’ve only been able to refer to these project local kube config files using an absolute path in the `.mirrord/mirrord.json `file; all attempts at relative paths don't seem to work. I guess this is due to the mirrord plugin/agent running in a different temp directory maybe?

I wonder is it possible to use a relative path or expression/environment variable for the current IntelliJ project directory inside the `kubeconfig` value in the `.mirrord/mirrord.json` file? 

As I’d love to be able to check in the `.mirrord/mirrord.json` file to git but right now it only works on 1 persons laptop due to the absolute path.

If this isn’t supported yet I’ll happy try figure out a patch; any hints/tips greatly appreciated - thanks!

aviramha:
Closes #2262

### DIFF
--- a/changelog.d/2262.changed.md
+++ b/changelog.d/2262.changed.md
@@ -1,0 +1,1 @@
+change kubeconfig path expansion to use env as well
--- a/mirrord/kube/src/api/kubernetes.rs
+++ b/mirrord/kube/src/api/kubernetes.rs
@@ -280,7 +280,8 @@ where
     };
 
     let mut config = if let Some(kubeconfig) = kubeconfig {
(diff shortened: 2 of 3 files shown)


infiniteregrets (approved):
🅰️

PR #2264, opened 2024-02-22
Use the same output rules for both meshed and not meshed networks

### LINKED ISSUE #2255: Steal + istio + port forward doesn't work
User reported in FastAPI app running on macOS that when sending requests via port forward, the requests reach the remote service instead of the local.

Probably relevant is istio in the setup.

DmitryDodzin:
closes #2255

According to `-m owner --gid-owner {gid} -p tcp -j RETURN` rule the agent traffic should be unaffected and `-o lo` should limit to localhost requests only.

Looks working =]


DmitryDodzin:
Checked with sidecars and everything worked as expected

mirrord/agent/src/steal/ip_tables/mesh.rs
```diff
@@ -114,53 +79,38 @@ where
...
-            .await?;
-
-        let redirect_rule = format!(
-            "{} -m tcp -p tcp --dport {redirected_port} -j REDIRECT --to-ports {target_port}",
-            self.own_packet_filter
-        );
-
-        self.managed.remove_rule(&redirect_rule)?;
+        tokio::try_join!(
+            self.prerouteing
```
aviramha, line 110: do we want this to run concurrently? why?
should be quite quick and just makes it more complex per my understanding

aviramha (changes requested):
Nitpick

mirrord/agent/src/steal/ip_tables/mesh.rs
DmitryDodzin: There is no real reason, It can be just sequential awaits

aviramha: I think it should be then, it'd be clearer

changelog.d/2255.fixed.md
```diff
@@ -0,0 +1 @@
+Use the same `OUTPUT` rules for both meshed and not meshed networks
```
aviramha, line 1: Mention what it fixes

aviramha (approved):

PR #2265, opened 2024-02-23
Increase internal proxy timeout

aviramha:
Increase internal proxy timeout from 5 seconds to 10 seconds to fix long agent ops


Razz4780 (approved):
:+1:

PR #2266, opened 2024-02-23
Concurrent DNS queries (agent)

Razz4780:
1. All DNS queries are processes concurrently in the agent
2. DNS queries do not block the client session


meowjesty (approved):
it's always dns

PR #2273, opened 2024-02-26
add support for mesh using nftables by passing flag to agent config

### LINKED ISSUE #2272: Support for Istio Mesh + nft
When users use Istio Mesh + nftables (not iptables legacy) our rules run before mesh, leading to undefined behavior.
We probably should have a flag - iptables_mode which can be either legacy or nft then we use the appropriate and all should work the same (in theory).
Need to manually test it to know better though.
Linkerd has similar modes. `iptablesMode`

aviramha:
Closes #2272

### DIFF
--- a/mirrord/agent/src/steal/ip_tables.rs
+++ b/mirrord/agent/src/steal/ip_tables.rs
@@ -118,7 +118,7 @@ pub struct IPTablesWrapper {
 }
 
 /// wrapper around iptables::new that uses nft or legacy based on env
-pub fn new_iptables() -> Result<iptables::IPTables, Box<dyn std::error::Error>> {
+pub fn new_iptables() -> iptables::IPTables {
     if let Some(val) = std::env::var("MIRRORD_AGENT_NFTABLES")
         && val.lowercase() == "true"
     {
@@ -126,6 +126,7 @@ pub fn new_iptables() -> Result<iptables::IPTables, Box<dyn std::error::Error>> 
     } else {
         iptables::new_with_cmd("/usr/sbin/iptables-legacy")
     }
+    .expect("IPTables initialization may not fail!")
 }
 
 impl Debug for IPTablesWrapper {
--- a/mirrord/agent/src/steal/ip_tables.rs
+++ b/mirrord/agent/src/steal/ip_tables.rs
@@ -120,7 +120,7 @@ pub struct IPTablesWrapper {
 /// wrapper around iptables::new that uses nft or legacy based on env
 pub fn new_iptables() -> iptables::IPTables {
     if let Ok(val) = std::env::var("MIRRORD_AGENT_NFTABLES")
-        && val.lowercase() == "true"
+        && val.to_lowercase() == "true"
     {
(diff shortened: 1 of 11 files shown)


mirrord/config/src/agent.rs
```diff
@@ -253,6 +253,15 @@ pub struct AgentConfig {
     #[config(default = false)]
     pub privileged: bool,
 
+    /// ### agent.nftables {#agent-nftables}
+    ///
+    /// Use iptables-nft instead of iptables-legacy.
+    /// Defaults to `false`.
+    ///
+    /// Needed if your mesh uses nftables instead of iptables-legacy,
```
meowjesty, line 261: Only for mesh?

mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -117,6 +117,17 @@ pub struct IPTablesWrapper {
     tables: Arc<iptables::IPTables>,
 }
 
+/// wrapper around iptables::new that uses nft or legacy based on env
+pub fn new_iptables() -> Result<IPTables, Box<dyn Error>> {
```
meowjesty: Do we have `Box<dyn Error>` anywhere else in the agent?

- If we do: then this gets a pass.
- If we don't: please wrap in an error.

meowjesty (changes requested):

mirrord/config/src/agent.rs
aviramha: I don't know of other situations where you need (k8s doesn't use nftables for now)

mirrord/agent/src/steal/ip_tables.rs
aviramha: It's just a propagation of the underlying error, didn't want to change the whole code for that.

mirrord/agent/src/steal/ip_tables.rs
```diff
@@ -117,6 +117,17 @@ pub struct IPTablesWrapper {
...
+/// wrapper around iptables::new that uses nft or legacy based on env
+pub fn new_iptables() -> Result<iptables::IPTables, Box<dyn std::error::Error>> {
+    if let Some(val) = std::env::var("MIRRORD_AGENT_NFTABLES")
+        && val.lowercase() == "true"
+    {
+        iptables::new_with_cmd("/usr/sbin/iptables-nft")
+    } else {
+        iptables::new_with_cmd("/usr/sbin/iptables-legacy")
+    }
+}
```
meowjesty, line 129:
```suggestion
pub fn new_iptables() -> iptables::IPTables {
    if let Some(val) = std::env::var("MIRRORD_AGENT_NFTABLES")
        && val.lowercase() == "true"
    {
        iptables::new_with_cmd("/usr/sbin/iptables-nft")
    } else {
        iptables::new_with_cmd("/usr/sbin/iptables-legacy")
    }
    .expect("IPTables initialization may not fail!")
}
```

And remove the `unwrap`s from the other pieces of code. We're just `unwrap`ping this anyway right?

meowjesty (approved):
:+1:

PR #2280, opened 2024-02-28
Allowed for non camel case types in hook macros

Razz4780:


meowjesty (approved):
Next on the "things that we should've fixed years ago" list: fix the `check-rust-docs` we get in github actions.

PR #2285, opened 2024-03-01
Fixed DNS resolution in `OutgoingSelector`

### LINKED ISSUE #2283: Outgoing local filter breaks all outgoing connections
### Bug Description

When using an outgoing local filter with a hostname that does not exist locally, all outgoing connections seem to be broken.

Example config:
```json
{
    "feature": {
        "network": {
            "outgoing": {
                "tcp": true,
                "filter": {
                    "local": ["i-do-not-exist-at-all"]
                }
            },
            "dns": true
        }
    }
}

```
Example app:
```rust
use std::net::TcpStream;

fn main() {
    let _stream = TcpStream::connect("py-serv:80").expect("py-serv should exist in the cluster");
}
```

### Steps to Reproduce

1. Deploy the `py-serv` service in the cluster
2. Run the example app and config from bug description
3. `TcpStream::connect` fails with `Os { code: 5, kind: Uncategorized, message: "Input/output error" }`. Layer prints `2024-03-01T10:28:50.776415Z ERROR ThreadId(01) mirrord_layer::error: Error occured in Layer >> IO(Custom { kind: Uncategorized, error: "failed to lookup address information: Temporary failure in name resolution" })`

### Backtrace

_No response_

### Relevant Logs

_No response_

Linux 6.5.0-21-generic #21~22.04.1-Ubuntu

### Local process

---

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Closes #2283 

When the user app wants to initiate an outgoing connection, all outgoing filters are resolved in order to determine whether the connection should be done remotely or locally. The issue was that all resolution failures were treated as fatal errors (and were aborting the `connect` call). This PR filters out resolution errors which mean only that the name does not map to any known IP address (which is the case when the user puts a non-existent host in the filter).

Also, this PR adds a `GetAddrInfoResponse` cache on the internal proxy level. No need to do multiple dns queries each time an outgoing connection is initiated

### DIFF
--- a/changelog.d/2283.fixed.md
+++ b/changelog.d/2283.fixed.md
@@ -1,0 +1,1 @@
+Fixed a bug where non-existent hosts in outgoing filter would prevent the application from making initiating outgoing connections.
--- a/changelog.d/2283.fixed.md
+++ b/changelog.d/2283.fixed.md
@@ -1,1 +1,1 @@
-Fixed a bug where non-existent hosts in outgoing filter would prevent the application from making initiating outgoing connections.
+Fixed a bug where non-existent hosts in outgoing filter would prevent the application from initiating outgoing connections.
--- a/mirrord/intproxy/src/proxies/simple.rs
+++ b/mirrord/intproxy/src/proxies/simple.rs
@@ -1,11 +1,11 @@
 //! The most basic proxying logic. Handles cases when the only job to do in the internal proxy is to
 //! pass requests and responses between the layer and the agent.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 
 use mirrord_intproxy_protocol::{LayerId, MessageId, ProxyToLayerMessage};
 use mirrord_protocol::{
-    dns::{GetAddrInfoRequest, GetAddrInfoResponse},
+    dns::{DnsLookup, GetAddrInfoRequest, GetAddrInfoResponse},
     file::{CloseDirRequest, CloseFileRequest, OpenDirResponse, OpenFileResponse},
     ClientMessage, FileRequest, FileResponse, GetEnvVarsRequest, RemoteResult,
 };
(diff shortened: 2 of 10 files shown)


mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -187,3 +206,43 @@ impl BackgroundTask for SimpleProxy {
...
+    /// [`GetAddrInfoRequest`] this struct knows of.
+    ///
+    /// # Panic
+    ///
+    /// This method panics if this struct does not now of any incomplete [`GetAddrInfoRequest`].
+    fn response_came(&mut self, lookup: DnsLookup) {
+        let host = self
+            .outstanding_requests
+            .pop_front()
+            .expect("received too many GetAddrInfoResponses");
```
meowjesty, line 239: Does this ever happen?

Wouldn't it be better if this was just a noop in this case? (maybe also adding a log so we could debug it)

Something like:

```rust
fn foo() -> Option<()> {
   let host = foo.pop_front()?;
    ...

    Some(()) // and we never check this anywhere
}
```

changelog.d/2283.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed a bug where non-existent hosts in outgoing filter would prevent the application from making initiating outgoing connections.
```
meowjesty, line 1: > from making initiating outgoing connections

"making initiating" sounds a bit weird? Maybe it's just me though.

meowjesty (changes requested):
YADI®

changelog.d/2283.fixed.md
Razz4780: Ah sorry, should be just "initiating"

mirrord/intproxy/src/proxies/simple.rs
Razz4780: This can literally never happen :x

meowjesty: Not with that attitude!

PR #2294, opened 2024-03-11
Hook `statx`

### LINKED ISSUE #2204: support statx on new libc
Currently we don't support statx - we have commented out code, not sure why but the comment says we can use it when libc is updated.
I think we can also backport the relevant structs if needed.
This impacted user having `fs.exists` failing on Rust Linux

Razz4780:
Closes #2204 

`statx` is not supported in Linux kernel prior to 4.11 or glibc prior to glibc 2.28. However, I don't see any problem with having this hooked, regardless of the user app's libc version. Either `statx` symbol is found and we should hook it or it's not found and nothing happens :eyes:

### DIFF
--- a/mirrord/layer/src/file/ops.rs
+++ b/mirrord/layer/src/file/ops.rs
@@ -1,9 +1,9 @@
 use core::ffi::CStr;
 use std::{env, ffi::CString, io::SeekFrom, os::unix::io::RawFd, path::PathBuf, time::Duration};
 
-use libc::{c_char, c_int, iovec, unlink, AT_FDCWD, FILE};
 #[cfg(target_os = "linux")]
-use libc::{statx, statx_timestamp};
+use libc::{c_char, statx, statx_timestamp};
+use libc::{c_int, iovec, unlink, AT_FDCWD, FILE};
 use mirrord_protocol::file::{
     OpenFileRequest, OpenFileResponse, OpenOptionsInternal, ReadFileResponse, SeekFileResponse,
     WriteFileResponse, XstatFsResponse, XstatResponse,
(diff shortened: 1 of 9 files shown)


mirrord/layer/src/file/ops.rs
```diff
@@ -444,6 +450,124 @@ pub(crate) fn xstat(
...
 }
 
+/// Logic for the `libc::statx` function.
+/// See [manual](https://man7.org/linux/man-pages/man2/statx.2.html) for reference.
+///
+/// # Warning
+///
+/// Due to backwards compatibility on the [`mirrord_protocol`] level, we use [`XstatRequest`] to get
+/// the remote file metadata.
+/// Because of this, we're not able to fill all field of the [`statx`] structure. Missing fields
```
aviramha, line 460: maybe we should zero the buffer before starting to write to it?

aviramha: Inb4 apps that don't check for that flag and rely on zeroing...

aviramha: Example
> If a filesystem does not support a field or if it has an
       unrepresentable value (for instance, a file with an exotic type),
       then the mask bit corresponding to that field will be cleared in
       stx_mask even if the user asked for it and a dummy value will be
       filled in for compatibility purposes if one is available (e.g., a
       dummy UID and GID may be specified to mount under some
       circumstances).

Razz4780: Ok, I'm convinced :+1:

mirrord/layer/src/file/ops.rs
```diff
@@ -444,6 +450,128 @@ pub(crate) fn xstat(
...
+/// caller about respective fields being skipped.
+#[cfg(target_os = "linux")]
+pub(crate) fn statx_logic(
+    dir_fd: RawFd,
+    path_name: *const c_char,
+    flags: c_int,
+    mask: c_int,
+    statx_buf: *mut statx,
+) -> Detour<c_int> {
+    let statx_buf = unsafe {
```
aviramha, line 477: Move the zero to line 541 where we are writing to the buffer for sure.

aviramha (approved):

PR #2308, opened 2024-03-15
Remove intproxy dns cache

Razz4780:
That was not the best idea


meowjesty (approved):
YADI won this one, but don't lose hope, you'll have another chance.

PR #2310, opened 2024-03-15
feature(jsonschema): `target` Simple Case is 'string'

### LINKED ISSUE #2188: Short form string Target missing from schema
### Bug Description

In the docs, a [series of options are given](https://mirrord.dev/docs/overview/configuration/#target-path) for passing the `target` as a string. However, these are not included in the JSON schema, so the schema is treated as invalid.

### Steps to Reproduce

1. Install VSCode Mirrord extension (which uses the schema at the route of this repo)
2. Create mirrord.json file with `{ "target": "deployment/foo/container/bar" }`
3. View "Problems" tab and observe "Incorrect type" warning.

### Backtrace

_No response_

### Relevant Logs

_No response_

Fedora Linux 38

### Local process

N/A

### Local process version

_No response_

### Additional Info

These options are supported by test cases in the code here, so it seems like it makes sense to add them into the schema.

One fix would be to add a `{ "type": "string", "pattern": "regex here" }` to the [existing `anyOf` in the schema](https://github.com/metalbear-co/mirrord/blob/main/mirrord-schema.json#L990), with a regex that supported the various options shown in the docs. I would be happy to open a PR with that change!

bradschwartz:
fixes #2188

NOTE: Currently I think this does actually break having `target` be just the Pod/Rollout/Deploy objects directly, so maybe this shouldn't be merged as-is, still working on a solution. I don't think schemars lets me _add_ a type to allow, so would the expected solution be to add a new `TargetFileConfig::Basic` variant that is just a string as allowed?

For JSONSchema used by VSCode, configure schemars for the simple `target` case to allow just a string. The complex/Advanced case is still going to be the object, although that is still broken for `target.path` being a string.

Currently there is no validation of what the string is. I could probably hardcode a regex for container/pod/deployment/rollout, but that seems like it could get out of sync in the future. Unsure if there's a way to just coerce Target variants to strings and regex on that, but that seems like it could be complicated.

## Examples

This is now valid and doesn't throw warnings in vscode:

```json
{
	"target": "deploy/my-deploy"
}
```

This is still invalid / throws warnings per the schema:

```json
{
	"target": {
		  "path": "deploy/my-deploy"
	}
}
```


mirrord-schema.json
```diff
@@ -1058,14 +1058,7 @@
     "TargetFileConfig": {
       "anyOf": [
         {
-          "anyOf": [
-            {
-              "$ref": "#/definitions/Target"
```
bradschwartz, line 1063: Didn't meant to _remove_ this as a usable schema value, so need to adjust

DmitryDodzin: Ah, The problem as I see it is our use of `#[serde(deserialize_with = "string_or_struct_option")]` so schemars doesn't quite have the information that it can be a string and when providing the `#[schemars(with = "String")]` it overrides the generation process completely. 

Luckily it seems that schemars [supports passing a function](https://github.com/GREsau/schemars/blob/master/schemars_derive/src/schema_exprs.rs#L96) to create the schema and it seems simple as `Fn(&mut SchemaGenerator) -> Schema` with [`SubschemaValidation `](https://docs.rs/schemars/latest/schemars/schema/struct.SubschemaValidation.html) as a nice helper to create the correct any-of (also relevant for line [31](https://github.com/metalbear-co/mirrord/pull/2310/files#diff-c4893d2d826b5d74ff657b6e3c3f26710530785e179b4ac46e2fbb915d0cbe3aL31))

bradschwartz: Thanks for pointing that out, I had tried a few things with changing the Enum but didn't love how it turned out. Let me give that a shot instead!

bradschwartz: @DmitryDodzin I'm definitely happier with this new solution, thanks for pointing it out! The string itself is still unvalidated but it feels like a good jumping off point for the future.

aviramha (approved):
Great solution, thank you!

PR #2322, opened 2024-03-26
Handle connection shutdowns

### LINKED ISSUE #2309: Hung connections
### Bug Description

Hung connections were observed when using `curl` to fetch static files from a `webpack` dev server (incoming steal) - `curl` stops progressing.

[agent.log](https://github.com/metalbear-co/mirrord/files/14617694/agent.log)

[intproxy.txt](https://github.com/metalbear-co/mirrord/files/14617682/intproxy.txt)

### Steps to Reproduce

---

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS 14.2.1

### Local process

node running a webpack dev server

### Local process version

18.19.1

### Additional Info

Kubernetes: standard AWS EKS with managed node groups
k8s 1.26
runner version 1.26.12-20240213

Razz4780:
Closes #2309

### DIFF
--- a/mirrord/agent/src/outgoing.rs
+++ b/mirrord/agent/src/outgoing.rs
@@ -100,8 +100,11 @@ impl TcpOutgoingApi {
 /// Handles outgoing connections for one client (layer).
 struct TcpOutgoingTask {
     next_connection_id: ConnectionId,
+    /// Writing halves of peer connections made on layer's requests.
     writers: HashMap<ConnectionId, WriteHalf<SocketStream>>,
+    /// Reading halves of peer connections made on layer's requests.
     readers: StreamMap<ConnectionId, ReaderStream<ReadHalf<SocketStream>>>,
+    /// Optional pid of agent's target. Used in [`SocketStream::connect`].
     pid: Option<u64>,
     layer_rx: Receiver<LayerTcpOutgoing>,
     daemon_tx: Sender<DaemonTcpOutgoing>,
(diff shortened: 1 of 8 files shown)


mirrord/agent/src/outgoing.rs
```diff
@@ -160,89 +78,266 @@ impl TcpOutgoingApi {
...
+                );
+
+                let daemon_message = DaemonTcpOutgoing::Read(Ok(DaemonRead {
+                    connection_id,
+                    bytes: vec![],
+                }));
+
+                self.daemon_tx.send(daemon_message).await?;
+
+                if !self.writers.contains_key(&connection_id) {
```
t4lz, line 222: sorry if you told me that already, but why can't we remove the reader of this connection in this case?
If we're here it means the user's application shut down from its side, and the connection peer has shut down from its side, right?

t4lz (commented):
flushing question 😬

mirrord/agent/src/outgoing.rs
```diff
@@ -160,89 +78,267 @@ impl TcpOutgoingApi {
...
+                    Some(writer) => writer.write_all(&bytes).await.map_err(ResponseError::from),
+
+                    None => Err(ResponseError::NotFound(connection_id)),
+                };
+
+                match write_result {
+                    Ok(()) if bytes.is_empty() => {
+                        self.writers.remove(&connection_id);
+
+                        if !self.readers.contains_key(&connection_id) {
```
t4lz, line 309: I couldn't find when we ever remove a connection's reader before we remove the writer, so maybe document what this case is and when does it happen.

PR #2325, opened 2024-03-27
Simplify KubernetesAPI

DmitryDodzin:
Changes:

* Remove `AgentManagment` because only `KubernetesAPI` implements it now and there is no need for this abstraction and moved the used functions straight onto `KubernetesAPI`.
* Split off pod template to a separate `ContainerVariant` implementation used inside of `JobVariant` and `JobTargetedVariant`.


aviramha:
is the pod variant actually used or to be used?

DmitryDodzin:
> is the pod variant actually used or to be used? also I think it might belong in the operator repo as it's only used by it?

`JobVariant<PodVariant<'c>>` It's currently used inside of `JobVariant` as well as why I needed to make it generic now

Razz4780 (approved):
Looks good to me :+1:

PR #2329, opened 2024-03-28
graceful_exit on proxyerror

aviramha:


mirrord/layer/src/error.rs
```diff
@@ -212,6 +212,11 @@ impl From<HookError> for i64 {
             HookError::SocketUnsuportedIpv6 => {
                 info!("{fail}")
             }
+            HookError::ProxyError(err) => {
+                graceful_exit!(
+                    "Proxy error, connectivity issue or a bug. Please report if not connectivity.\n{err}"
```
meowjesty, line 217:
```suggestion
                    "Proxy error, connectivity issue or a bug. \n\
                    You may report it to us on https://github.com/metalbear-co/mirrord/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml \n{err}"
```

meowjesty: We may want people to report this stuff to us? Or have you seen cases where it's just a small issue on the user side?

meowjesty (approved):
Should we always ask for a report?

aviramha:
> Should we always ask for a report?

I think connection might be dropping which is okay and not a bug.

PR #2333, opened 2024-03-29
Fix some open/fd potential issues

aviramha:


meowjesty (approved):
:potable_water: :wrench:

PR #2350, opened 2024-04-04
mirrord now listens on 0.0.0.0 when requested

### LINKED ISSUE #2319: Capability to modify the local listen address for mirrord
_Feature request born out of the issue comment found here: https://github.com/metalbear-co/mirrord/issues/1658#issuecomment-2016755958_

I'd like to be able to bind mirrord to `0.0.0.0` so I can expose a mirrord tunnelled application from inside a local container.

This would permit containers in the same network namespace or the host OS applications if the container is port mapped, access to the kubernetes internal network via proxy services, e.g. tinyproxy, microsocks, squid etc.

aviramha:
mirrord now listens on 0.0.0.0 when requested and changes address to localhost only when needed

Closes #2319


mirrord/layer/src/socket/ops.rs
```diff
@@ -229,12 +236,13 @@ pub(super) fn bind(
 
     // Listen port was specified
     if let Some(port) = listen_port {
-        bind_port(sockfd, socket.domain, port)
+        let address = SocketAddr::new(requested_address.ip(), port);
+        bind_address(sockfd, socket.domain, &ddress)
```
DmitryDodzin, line 240:
```suggestion
        bind_address(sockfd, socket.domain, &address)
```

DmitryDodzin (approved):

aviramha:
Closed in favor of #2364

PR #2363, opened 2024-04-08
Pass operator's TLS cert to agent container

### LINKED ISSUE #2362: Authenticate agent connection
When TLS is in use (#2299), provide the agent with an X509 certificate. The agent should make connections only with TLS servers using this certificate.

**Note**: this is an operator-only feature

Razz4780:
Closes #2362 

The original idea was to provide the agent only with a public key. However, I didn't want to go to deep into `rustls` and write my own verifier, because I'm not sure how to get this right. So I made the agent accept the whole certificate and use it as root cert.


mirrord/protocol/src/lib.rs
```diff
@@ -89,11 +89,13 @@ impl FromStr for MeshVendor {
     }
 }
 
-/// Name of environment variable that can be used to instruct the agent to use TLS when accepting
-/// client connections. Should store only values that parse into [`bool`].
+/// Name of environment variable that can be used to provide the agent with a PEM-encoded X509
+/// certificate. Given the certificate, the agent will secure the incoming connections with TLS.
+/// The agent will act as TLS client and will make successful connections only with TLS servers
+/// using the given certificate. The agent will still be the side accepting TCP connections.
```
aviramha, line 95: didn't understand last statement

Razz4780: Rephrased a bit

mirrord/kube/src/api.rs
```diff
@@ -95,14 +95,15 @@ pub trait AgentManagment {
     /// # Params
     ///
     /// * `config` - if passed, will be checked against cluster setup
-    /// * `extra_env` - appended to agent container definition
+    /// * `tls_cert` - value for
+    ///   [`AGENT_OPERATOR_CERT_ENV`](mirrord_protocol::AGENT_OPERATOR_CERT_ENV)
```
t4lz, line 99:
```suggestion
    /// * `tls_cert` - value for
    ///   [`AGENT_OPERATOR_CERT_ENV`](mirrord_protocol::AGENT_OPERATOR_CERT_ENV),
    ///   for creating an agent from the operator. In usage from this repo this is always `None`.
```
What do you think?

t4lz (approved):

PR #2371, opened 2024-04-12
Extend port stealing config with `ports` field, change `http_filter.ports` behavior

### LINKED ISSUE #2366: Change port stealing configuration behavior
This is meant to address two user issues:

- When specifying steal + HTTP filtering, all ports that aren't configured under HTTP filtering are stolen in their entirety. This is even weirder when the user accidentally configures the wrong ports for HTTP filtering, and just end up stealing all traffic coming to the service
- When users only want to steal a specific port, and ignore all other ports, they have to actually list all other ports (which can be a lot) under `ignore_ports`

The new behavior should be as follows:
```
{
  "feature": {
    "network": {
      "incoming": {
        "mode": "steal"
      }
    }
    }
}
```
This will steal on all ports.

```
{
  "feature": {
    "network": {
      "incoming": {
        "mode": "steal"
        "ports": [1,2,3]
      }
    }
    }
}
```
`ports` is a new configuration value. This will only steal on ports 1,2,3.

```
  "feature": {
    "network": {
      "incoming": {
        "http_filter": {
          "ports": [80, 100]
        },
      }
    }
    }
}
```
This will only steal on ports 80, 100 with an HTTP filter. This is different from today's behavior, where these ports would be stolen with a filter, and all other ports would be stolen completely.

```
{
  "feature": {
    "network": {
      "incoming": {
        "http_filter": {
          "ports": [80, 100]
        },
        "ports": [1,2,3]
      }
    }
    }
}
```
This will steal with a filter on ports 80, 100, and steal completely on ports 1,2,3.

Razz4780:
Closes #2366


changelog.d/2366.changed.md
```diff
@@ -0,0 +1,3 @@
+Changed port stealing configuration:
+1. Added new `port` field to the `incoming` configuration. The field lists ports that should be stolen/mirrored. Other ports remain local.
```
aviramha, line 2: typo in `port`

aviramha (approved):

changelog.d/2366.changed.md
```diff
@@ -0,0 +1,3 @@
+Changed port stealing configuration:
+1. Added new `ports` field to the `incoming` configuration. The field lists ports that should be stolen/mirrored. Other ports remain local.
```
aviramha, line 2: Logic should be that if `ignore_ports` exists and `ports` is empty, then all ports will be stolen except for those in `ignore_ports`

Razz4780: The new field itself is optional, so I think the logic checks out. Should I extend the changelog here?

aviramha: nvm I think I was confused

PR #2373, opened 2024-04-16
Change reqwest to use rustls with native certificates to work in more cases

aviramha:


Razz4780 (changes requested):
I think it needs an update in `mirrord-analytics`

aviramha:
> I think it needs an update in `mirrord-analytics`

It's using workspace?..

Razz4780 (approved):

Razz4780:
> > I think it needs an update in `mirrord-analytics`
> 
> It's using workspace?..

Sorry, got confused

PR #2376, opened 2024-04-16
Connection task tracing + proper TLS shutdown

Razz4780:
Minor changes to tokio task spawned in `wrap_raw_connection`:

1. Added `codec.close()` at the end of the task. It doesn't change much when we're connecting to the agent from the CLI or from the operator via plain TCP, but it removes some ugly IO errors produced by `tokio-rustls` when operator talks to the agent with TLS.
2. Removed logging `DaemonMessage::LogMessage`s. Given that this code is used in the operator, we I believe we don't want to log agent->client messages globally. We should leave them to be handled by internal proxy.
3. Improved tracing a bit


aviramha (approved):
!

PR #2377, opened 2024-04-16
Update nightly to 2024-04-15.

meowjesty:
Removing deprecated function call from `whoami`, and some unused stuff.

Bunch of clippy fixes.

With this, we should be able to run `cargo update` again.


aviramha:
Need to update CI base images first, no?

aviramha (approved):

PR #2388, opened 2024-04-22
Added JetBrains debugger to build tools

Razz4780:
Closes https://github.com/metalbear-co/mirrord-intellij/issues/258

Fixes mirrord compatibility with Rider 2024.1


t4lz (approved):
Do we need to test it doesn't break other JetBrains IDE plugins?

Razz4780:
> Do we need to test it doesn't break other JetBrains IDE plugins?

Hmm, nice catch. I'd be surprised, but I'll check

PR #2393, opened 2024-04-23
Fix medschool dropping fields sometimes

### LINKED ISSUE #1580: `medschool` drops fields (sometimes).
### Bug Description

When running `medschool` in our `mirrord-config` project, sometimes a few fields are dropped (so far I've only seen the `operator` config field being lost).

I'm not quite sure why this happens, probably due to how we search for inner types in the fields of a struct, and when we have reached an edge we go back, and repeat the process until all edges have been found. So the bug probably lies in the "going back" part.

A somewhat decent way of checking if a field has been dropped is to run `medschool` a few times, and check the amount of lines produce in `configuration.md`. The correct one will have more lines (keep in mind that the tool wasn't tested for overwriting an already existing `configuration.md`, so to guarantee you're looking at a correct file, before you re-run it, change the existing file name / remove it / copy it) .

### Steps to Reproduce

1. `cd mirrord/config`;
2. run `../../target/medschool`;
3. `cat -n configuration.md`;
4. check line numbers (currently should be `>= 790`
5. `mv configuration.md.old`;
6. run `../../target/medschool`;
7.  `cat -n configuration.md`;
8. compare line count with previous run;

### Backtrace

_No response_

### Relevant Logs

_No response_

N.A.

### Local process

medschool

### Local process version

_No response_

### Additional Info

_No response_

infiniteregrets:
closes #1580 

- went through the logic where we resolve references by building a tree, there is nothing really going wrong there however the logic can be simplified..
- the only fishy place was when we parse the rust files and somehow end up assuming there can be duplicate fields in a struct and convert to a unique set using a HashSet which somehow ends dropping a field?


meowjesty (approved):
:+1: 

But please leave some docs on the possible duplication that might happen due to us not checking for unique types any more.

PR #2404, opened 2024-04-28
Same Session ID for copy-target

DmitryDodzin:
Make sure copy-target request has the same `session_id` as the subsequent connection.


aviramha:
The session ID should be set from the operator side, so I think any change/addition of logic in this side should be avoided..

PR #2406, opened 2024-04-29
fallback to OSS on license expiry

aviramha:


mirrord/operator/src/client.rs
```diff
@@ -287,6 +285,13 @@ impl OperatorApi {
                 progress.info(&good_validity_message);
                 info!(good_validity_message);
             }
+        } else {
+            let no_license_message = "Operator has no valid license, falling back to OSS usage.";
```
eyalb181, line 289:
```suggestion
            let no_license_message = "No valid license found for mirrord for Teams, falling back to OSS usage. Visit https://app.metalbear.co to purchase or renew your license.";
```

mirrord/operator/src/client.rs
```diff
@@ -102,6 +102,9 @@ pub enum OperatorApiError {
         operation: String,
         status: Box<kube::core::Status>,
     },
+
+    #[error("Operator has expired license, falling back to OSS usage.")]
+    NoLicense,
```
meowjesty, line 107:
```suggestion
    ExpiredLicense,
```

Nit.

meowjesty: And if you go with this, change the `no_license` vars to `expired_license`.

meowjesty (approved):
Just a nit on the error name.

mirrord/operator/src/client.rs
meowjesty: Sure, it was only a nit anyway.

PR #2407, opened 2024-04-30
Fix case were agent log message causes startup failure

aviramha:


t4lz (commented):
The error handling is just an optional suggestion. Let me know if you disagree about the timeout.

aviramha:
> The error handling is just an optional suggestion. Let me know if you disagree about the timeout.

 The log would emit a clear message anyway (debug representation so we know exactly what we had received)..
The timeout is a good catch.

t4lz (approved):
Looks fine. If we're using the same error variant for all scenarios, I suggest changing its name from `AgentClosedConnection` (as that might be a bit confusing when that is not what happen) to `NoPong`, and also changing that error's text (in `#[error("...")]`) to `Did not get a 'pong' from the agent: could be a timeout, an invalid message, or the agent closed the connection`.

t4lz (approved):
thanks

PR #2418, opened 2024-05-02
feat: `config_file` Shell Completions Hint for Files

bradschwartz:
Currently, shell completions for `mirrord exec --config-file <TAB><TAB>` only will complete to the next match of the path provided, and then assume it is done and append a space. We can provide a value_hint to the macro so that the generated completions only finish for a real file. Test like:

```bash
source <(cargo run completions bash)
mirrord exec --config-file ~/partial/path-<TAB><TAB>
```

*Open Question for Maintainers*: I only did this for the `exec` subcommand - is it the same config file used in all the other subcommands as well? If so, I can also add this value hint for them, just wasn't sure.


aviramha (approved):
Great, thank you!

PR #2423, opened 2024-05-03
Improve tera templating error, by digging into the source.

meowjesty:
- Issue: [#261](https://github.com/metalbear-co/mirrord-intellij/issues/261)

Improves the error message we get when a tera template config error happens.

### DIFF
--- a/mirrord/config/src/config.rs
+++ b/mirrord/config/src/config.rs
@@ -3,6 +3,8 @@ pub mod from_env;
 pub mod source;
 pub mod unstable;
 
+use std::error::Error;
+
 use thiserror::Error;
 
 /// <!--${internal}-->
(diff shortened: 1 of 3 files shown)


mirrord/config/src/config.rs
```diff
@@ -49,8 +51,23 @@ pub enum ConfigError {
...
+}
+
+impl From<tera::Error> for ConfigError {
+    fn from(fail: tera::Error) -> Self {
+        let mut fail_message = fail.to_string();
+        let mut source = fail.source();
+
+        while let Some(fail_source) = source {
+            fail_message.push_str(" -> ");
+            fail_message.push_str(&fail_source.to_string());
```
infiniteregrets, line 65: You can use format! here no?

infiniteregrets (approved):

PR #2424, opened 2024-05-03
Refactor `medschool` to use a simple DFS

infiniteregrets:
A few changes to medschool - refactored the code, changed the algorithm taking into consideration we don't ever drop fields.


meowjesty (changes requested):
Still going through it, so here are just a few things.

Could you add a better sanity test than the basic tests we currently have? I was thinking it would be swell if we had a test that tried to parse 2 files, with a couple of `struct`s, `enum`s, and out of order fields, so we can have better future-proofing.

medschool/Cargo.toml
```diff
@@ -29,3 +29,8 @@ regex.workspace = true
 
 [dev-dependencies]
 rand = "0.8"
+
+
+
+
+
```
meowjesty, line 36: random empty lines

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+            }
+            _ => break,
+        }
+    }
+
+    inner_type
+}
+
+/// Converts a [`syn::ItemMod`] into a [`PartialType`].
+fn parse_item_mod(item_mod: syn::ItemMod) -> Option<PartialType> {
```
meowjesty, line 78: Why is this an `Option` if the return is always `Some`?

meowjesty, line 78: It would be neat to implement this as `From<syn::ItemMod> for PartialType`.

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    // We only care about types that have docs.
+    (!thing_docs_untreated.is_empty() && !is_internal).then(|| PartialType {
+        ident: item.ident.to_string(),
+        docs: thing_docs_untreated,
+        fields: Default::default(),
+    })
+}
+
+/// Converts a [`syn::ItemStruct`] into a [`PartialType`].
```
meowjesty, line 102:
```suggestion
/// Converts a [`syn::ItemStruct`] into a [`PartialType`].
///
/// We remove docs that have `<!--${internal}-->` (they're not converted to [`PartialType`]s).
```

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+            return None;
+        }
+
+        // `trim` is too aggressive, we just want to remove 1 whitespace
+        if doc.starts_with(' ') {
+            doc.remove(0);
+        }
+    }
+
+    docs.push("\n".to_string());
```
meowjesty, line 122:
```suggestion
    // New line to separate this types' docs from the next types' docs.
    docs.push("\n".to_string());
```

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+
+use syn::{Attribute, Expr, Ident, Meta, Type, TypePath};
+
+use crate::{
+    types::{PartialField, PartialType},
+    DocsError,
+};
+
+#[tracing::instrument(level = "trace", ret)]
+pub fn docs_from_attributes(attributes: Vec<Attribute>) -> Vec<String> {
```
meowjesty, line 13:
```suggestion
pub fn docs_from_attributes(attributes: Vec<Attribute>) -> Option<Vec<String>> {
```

wdyt? Then in every function that uses this, you can just do `let docs = docs_from_attributes(attrs)?;`

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+    (!fields.is_empty()).then(|| PartialType {
+        ident: item.ident.to_string(),
+        docs,
+        fields,
+    })
+}
+
+/// Converts a [`syn::Item`] into a [`PartialType`], if possible.
+#[tracing::instrument(level = "trace", ret)]
+fn parse_syn_item_into_partial_type(item: syn::Item) -> Option<PartialType> {
```
meowjesty, line 146: Another good candidate for becoming a `From<syn::Item> for PartialType`, we're dealing with `Option`s in some of these cases, so this would have to be `TryFrom`, and we `.ok()` or ignore the errors.

```rust
impl TryFrom<syn::Item> for PartialType {
    type Error = ();
}
```

meowjesty (changes requested):
moar suggestions!

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+fn parse_item_struct(item: syn::ItemStruct) -> Option<PartialType> {
+    let mut docs = docs_from_attributes(item.attrs);
+
+    if docs.is_empty() {
+        return None;
+    }
+
+    for doc in docs.iter_mut() {
+        // removes docs that we don't want in `configuration.md`
+        if doc.contains(r"<!--${internal}-->") {
```
meowjesty, line 112: Instead of checking for this on every call site, if you turn `docs_from_attributes` to return an `Option<Vec>`, you could check for this there, and return `None`, everywhere else would become just `let mut docs = docs_from_attributes(whatever)?;`.

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+        ident: item_mod.ident.to_string(),
+        docs: thing_docs_untreated,
+        fields: Default::default(),
+    })
+}
+
+/// Converts a [`syn::ItemEnum`] into a [`PartialType`].
+fn parse_item_enum(item: syn::ItemEnum) -> Option<PartialType> {
+    let thing_docs_untreated = docs_from_attributes(item.attrs);
+    let is_internal = thing_docs_untreated
```
meowjesty, line 90: See other `internal` suggestion related to `Option<Vec>`.

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    for doc in docs.iter_mut() {
+        // removes docs that we don't want in `configuration.md`
+        if doc.contains(r"<!--${internal}-->") {
+            return None;
+        }
+
+        // `trim` is too aggressive, we just want to remove 1 whitespace
+        if doc.starts_with(' ') {
```
meowjesty, line 117: Can't we also move this to the `docs_from_attributes`? Or would it mess up non-`ItemStruct`s?

medschool/src/types.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// These docs will be merged with the type docs of this field's type, if it is a sub-type of
+    /// any type that we declared.
+    pub docs: Vec<String>,
+}
+
+impl PartialField {
+    /// Converts a [`syn::Field`] into [`PartialField`], using
+    /// [`get_ident_from_field_skipping_generics`] to get the field type.
+    #[tracing::instrument(level = "trace", ret)]
+    pub fn new(field: syn::Field) -> Option<Self> {
```
meowjesty, line 56: Convert this to `TryFrom`?

medschool/src/types.rs
```diff
@@ -0,0 +1,180 @@
...
+                    .get_ident()
+                    .cloned()
+                    .or_else(|| get_ident_from_field_skipping_generics(type_path))
+            }
+            _ => None,
+        }?;
+
+        let mut docs = docs_from_attributes(field.attrs);
+
+        // TODO: this part can be encapsulated into a function
```
meowjesty, line 71: Encapsulate it :pill:

medschool/src/types.rs
```diff
@@ -0,0 +1,180 @@
...
+impl Display for PartialType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Prevents crashing the tool if the type is not properly documented.
+        let docs = self.docs.last().cloned().unwrap_or_default();
+        f.write_str(&docs)?;
+
+        self.fields.iter().fold(f, |accum, s| {
+            accum
+                .write_str(&s.ident.to_string())
+                .expect("writing failed");
```
meowjesty, line 167: Think you can use `try_fold` to avoid `expect`.

medschool/src/types.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+impl Display for PartialType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Prevents crashing the tool if the type is not properly documented.
+        let docs = self.docs.last().cloned().unwrap_or_default();
+        f.write_str(&docs)?;
+
+        self.fields.iter().fold(f, |accum, s| {
```
meowjesty, line 164:
```suggestion
        self.fields.iter().fold(f, |final_docs, type_| {
```

meowjesty: Or something along those lines, let's avoid single chars and names that are too generic.

meowjesty (changes requested):
Some improvements (also to existing code).

medschool/src/main.rs
```diff
@@ -11,547 +11,43 @@
...
-            }
+pub(crate) fn parse_files(path: PathBuf) -> Result<Vec<syn::File>, DocsError> {
+    let mut paths = glob::glob(&format!("{}/**/*.rs", path.to_string_lossy()))?.peekable();
+    paths.peek().ok_or_else(|| DocsError::NoFiles)?;
+
+    paths
+        .map(|path| {
+            let mut file = File::open(path?)?;
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)?;
```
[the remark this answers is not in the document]
infiniteregrets: great idea, as I see now, calling `clear` is much cheaper than repeated allocations. thanks!

medschool/src/parse.rs
infiniteregrets: because other related functions do, like `parse_item_struct` and can return None. all these functions are used in `parse_syn_item_into_partial_type`

infiniteregrets: it would lead to some inconsistent code, we can for sure have this, but not for other functions..

medschool/src/parse.rs
```diff
@@ -0,0 +1,286 @@
...
+use crate::{
+    types::{PartialField, PartialType},
+    DocsError,
+};
+
+#[tracing::instrument(level = "trace", ret)]
+pub fn docs_from_attributes(attributes: Vec<Attribute>) -> Vec<String> {
+    attributes
+        .into_iter()
+        .filter_map(|attribute| {
```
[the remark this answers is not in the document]
infiniteregrets: I had difficulty reading it that way so I changed it lol, now coming back to it I agree with you

medschool/src/parse.rs
infiniteregrets: what is the use of it tho?

medschool/src/parse.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/// DFS helper function to resolve the references of the types. Returns the docs of the fields of
+/// the field we're currently looking at search till its leaf nodes. The leaf here means a primitive
+/// type for which we don't have a [`PartialType`].
+fn dfs_fields<'a>(
+    field: &PartialField<'a>,
+    types: &'a HashSet<PartialType>,
+    cache: &mut HashMap<&'a str, Vec<String>>,
+    recursion_level: &mut usize,
```
meowjesty, line 206: I think we need a cap for this, to avoid crashing if the recursion goes too deep.

Probably a `const` on the function that we check to abort.

meowjesty (changes requested):
Just need to set a max recursion to avoid crashing/hanging before we can merge this!

meowjesty (approved):
:medical_symbol:

PR #2433, opened 2024-05-08
Add edits to CONTRIBUTING.md

gememma:
- Add GCC to dependencies
- Add job limit fix for RAM exhaustion
- Add missing argument to mirrord-console command

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@ The following guide details the steps to setup a local development environment f
 
 ### Prerequisites
 
+- [GCC](https://gcc.gnu.org/) - only on Linux, GCC is needed for Go dynamic linking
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -21,6 +21,7 @@ The following guide details the steps to setup a local development environment f
 
 ### Prerequisites
 
+- [GCC](https://gcc.gnu.org/)
```
aviramha, line 24: only on Linux, no?. It is needed for golang to do dynamic linking (maybe note that as well?)

gememma: good point, thanks

aviramha (approved):

aviramha:
Congrats on the first PR 🥳

PR #2435, opened 2024-05-09
don't error on permission denied hookerror

aviramha:


mirrord/layer/src/error.rs
```diff
@@ -215,6 +215,9 @@ impl From<HookError> for i64 {
             HookError::SocketUnsuportedIpv6 => {
                 info!("{fail}")
             }
+            HookError::IO(e) if (e.kind() == std::io::ErrorKind::PermissionDenied) => {
```
Razz4780, line 218: Suggestion - you can add this case to `is_ignored_code` function (used above)

Razz4780 (approved):

PR #2436, opened 2024-05-10
Fix check-rust-docs warnings

### LINKED ISSUE #1399: Fix all check-rust-docs warnings
Existing warnings are shown in the "Unchanged files with check annotations" parts of PRs' code, e.g. https://github.com/metalbear-co/mirrord/pull/1343/files#file-mirrord-protocol-src-outgoing-rs-L102.

We should fix all of those issues so that there are no preexisting warnings, then not merge new warning-inducing code into main.

## Where to begin?
Please refer to the [CONTRIBUTING.md file](https://github.com/metalbear-co/mirrord/blob/main/CONTRIBUTING.md)

gememma:
Closes #1399 

- Removed some broken links to items that exist but are not in scope - if they are to be linked, the path needs to be specified manually
- Added TODOs where items appear to not exist in the codebase - possibly they refer to something that has since been renamed
- Added angle brackets around bare links in doc comments
- Fixed spellings on some links
- Added backticks to e.g. `<DATA>` that rust-doc interprets as an HTML tag
- Allowed public → private links

### DIFF
--- a/mirrord/layer/src/socket/hooks.rs
+++ b/mirrord/layer/src/socket/hooks.rs
@@ -109,9 +109,9 @@ pub(crate) unsafe extern "C" fn gethostname_detour(
 /// Hook for `libc::gethostbyname` (you won't find this in rust's `libc` as it's been deprecated and
 /// removed).
 ///
-/// Resolves DNS `raw_name` and allocates a `static` [`libc::hostent`] that we change the inner
-/// values whenever this function is called. The address itself of `*mut hostent` has to remain the
-/// same (thus why it's a `static`).
+/// Resolves DNS `raw_name` and allocates a `static` [`libc::hostent`] that we change the
+/// inner values whenever this function is called. The address itself of `*mut hostent` has to
+/// remain the same (thus why it's a `static`).
 #[hook_guard_fn]
 unsafe extern "C" fn gethostbyname_detour(raw_name: *const c_char) -> *mut hostent {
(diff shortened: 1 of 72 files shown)


gememma:
CI fails if the check-rust-docs job fails, which occurs when cargo doc emits any warnings (warnings are turned into errors with `-Dwarnings` in `RUSTDOCFLAGS`)

mirrord/layer/src/socket/hooks.rs
```diff
@@ -415,7 +415,7 @@ pub(super) unsafe extern "C" fn recvmsg_detour(
 
 /// Not a faithful reproduction of what [`libc::sendmsg`] is supposed to do, see [`sendmsg`].
 ///
-/// TODO(alex): We are ignoring the control message header [`libc::cmgshdr`].
+/// TODO(alex): We are ignoring the control message header [`libc::cmsghdr`].
```
meowjesty, line 418:
```suggestion
// TODO(alex): We are ignoring the control message header `libc::cmsghdr`.
```

Don't think this was supposed to be a doc.

meowjesty (approved):
Thanks for your service in the mines :pick: !

meowjesty:
Changes to docs in `mirrord/config` require running:

```sh
cargo test -p mirrord-config check_schema_file_exists_and_is_valid_or_create_it -- --ignored --nocapture
```

to generate the config schema file.

PR #2440, opened 2024-05-14
Fix outgoing connect on bound sockets

### LINKED ISSUE #2438: mirrord fails to handle remote connection for bound sockets
### Bug Description

When the user application binds the socket before connecting, `mirrord-layer` bypasses the `connect` call. It rejects the socket as being in invalid state (layer expects that `bind` is always followed by `listen`).

### Steps to Reproduce

-

### Backtrace

_No response_

### Relevant Logs

_No response_

-

### Local process

-

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Closes #2438


aviramha (approved):
Nice!

PR #2447, opened 2024-05-15
Give config info at session start

### LINKED ISSUE #2367: Provide explanation before session start
When user starts a session, provide information to it about their config, and what it would do.

For example:
```
Running binary X with args Y. mirrord Configuration file is C.
mirrord will target PODNAME. It will load all of its environment variables, provide access to its file for read only and will mirror all ports you're listening on locally
```

gememma:
Closes #2367

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -104,32 +104,67 @@ where
     binary_args.insert(0, args.binary.clone());
 
     // Print config details for the user
-    sub_progress.info(
-        &format!(
-            "Running binary \"{}\" with arguments: {:?}.",
-            binary, args.binary_args
-        )[..],
+    let mut sub_progress_config = progress.subtask("config summary");
+    print_config(
+        &sub_progress_config,
+        Some(&binary),
+        Some(&args.binary_args),
+        &config,
+        false,
     );
+    sub_progress_config.success(None);
+
+    sub_progress.success(Some("ready to launch process"));
+    // The execve hook is not yet active and does not hijack this call.
+    let err = execvp(binary.clone(), binary_args.clone());
+    error!("Couldn't execute {:?}", err);
+    analytics.set_error(AnalyticsError::BinaryExecuteFailed);
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    if let exec::Error::Errno(errno) = err {
+        if Into::<i32>::into(errno) == 86 {
+            // "Bad CPU type in executable"
+            if _did_sip_patch {
+                return Err(CliError::RosettaMissing(binary));
+            }
+        }
+    }
+    Err(CliError::BinaryExecuteFailed(binary, binary_args))
+}
+
+fn print_config<P>(
+    progress_subtask: &P,
+    binary: Option<&String>,
+    binary_args: Option<&Vec<String>>,
+    config: &LayerConfig,
+    single_msg: bool,
+) where
+    P: Progress + Send + Sync,
+{
+    let mut messages = vec![];
+    if let Some(b) = binary
+        && let Some(a) = binary_args
+    {
+        messages.push(format!("Running binary \"{}\" with arguments: {:?}.", b, a));
+    }
 
     let target_info = if let Some(target) = &config.target.path {
         &format!("mirrord will target: {}", target)[..]
     } else {
-        // TODO: check accuracy of statement
         "mirrord will run without a target"
     };
     let config_info = if let Ok(path) = std::env::var("MIRRORD_CONFIG_FILE") {
-        &format!("a configuration file was loaded from: {}. ", path)[..]
+        &format!("a configuration file was loaded from: {} ", path)[..]
     } else {
         "no configuration file was loaded"
     };
-    sub_progress.info(&format!("{}, {}", target_info, config_info)[..]);
+    messages.push(format!("{}, {}", target_info, config_info));
 
     let operator_info = match config.operator {
         Some(true) => "be used",
         Some(false) => "not be used",
         None => "be used if possible",
     };
-    sub_progress.info(&format!("operator: the operator will {}", operator_info)[..]);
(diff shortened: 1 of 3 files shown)


aviramha:
Can you attach a GIF to show how it looks in CLI/IDE please?

gememma:
Ah. I completely forgot about the plugins - I'll mark this as a draft and make sure they look good too

gememma:
For the record, this is how it looks in the CLI:
print_config

mirrord/cli/src/main.rs
```diff
@@ -102,6 +103,85 @@ where
...
+        "all"
+    };
+    sub_progress.info(&format!("env: {} environment variables will be fetched", env_info)[..]);
+
+    let fs_info = match config.feature.fs.mode {
+        FsModeConfig::Read => "read only from the remote",
+        FsModeConfig::Write => "read and write from the remote",
+        _ => "read and write locally",
+    };
+    sub_progress.info(&format!("fs: file operations will default to {}", fs_info)[..]);
```
aviramha, line 159: maybe mention the defaults local/remote that are overriding (unless explicitly specified otherwise)

mirrord/cli/src/main.rs
```diff
@@ -102,6 +103,85 @@ where
     // Put original executable in argv[0] even if actually running patched version.
     binary_args.insert(0, args.binary.clone());
 
+    // Print config details for the user
+    sub_progress.info(
```
Razz4780, line 107: If you use a separate `Progress` subtask for printing this out (`Progress::subtask`), it will be easier to find this info in the plugins (you'll get a JSON message like `{"type": "new_task", "name": "you_choose"}`)

Razz4780: Also, you could put this in a separate function ^^

mirrord/cli/src/main.rs
gememma: Can you give an example? I don't think I can visualise it

Razz4780: [here](https://github.com/metalbear-co/mirrord/tree/main/mirrord/layer/src/file/filter) we have default regex sets for file ops. There's quite a lot of them tbh, I don't know if we should list them all :eyes: 

I think the ideal solution would be to have them at `mirrord.dev` docs and only provide a link in this message (probably not in scope of this PR though). @aviramha wdyt?

gememma:
## VSCode plugin
### collapsed (default)
popup_custom_config_collapsed
### expanded
popup_custom_config
## IntelliJ plugin
### collapsed (default)
popup_custom_intellij_collapsed
### expanded
popup_custom_intellij

aviramha:
> IMO doesn't look great, but perhaps there's a way for us to show this in a persistent UI component within the IDE? Even just mirrord dropdown menu -> Session Status which would make a popup show up. This has the advantage of letting the user check the status throughout the entire session, not just at the start.

Having a persistent UI component has been itching me for a while but I think it should be out of scope for this specific issue. It'd require more design and architecture decisions.
We can consider not having this feature in IDEs for now.

eyalb181:
> > IMO doesn't look great, but perhaps there's a way for us to show this in a persistent UI component within the IDE? Even just mirrord dropdown menu -> Session Status which would make a popup show up. This has the advantage of letting the user check the status throughout the entire session, not just at the start.
> 
> Having a persistent UI component has been itching me for a while but I think it should be out of scope for this specific issue. It'd require more design and architecture decisions. We can consider not having this feature in IDEs for now.

@gememma In that case, lets just not have this in the IDEs for the moment. Instead, please open follow-up issues in IntelliJ and VS Code to display this in a persistent component.

Razz4780:
> @aviramha @Razz4780 @gememma Thoughts about this?
> 
> IMO doesn't look great, but perhaps there's a way for us to show this in a persistent UI component within the IDE? Even just mirrord dropdown menu -> Session Status which would make a popup show up. This has the advantage of letting the user check the status throughout the entire session, not just at the start.

1. VSCode - I'm not sure if we can do better. I don't think notifications respect newlines + there's no "popup" concept in VSCode. We can only use notifications, quickpick, open a new editor tab prefilled with text :shrug: 
2. IntelliJ - we can do a popup, yeah

Regardless, this is what it's going to look like in old plugin/extension versions. Unless we come up with something very cunning

eyalb181:
> > @aviramha @Razz4780 @gememma Thoughts about this?
> > IMO doesn't look great, but perhaps there's a way for us to show this in a persistent UI component within the IDE? Even just mirrord dropdown menu -> Session Status which would make a popup show up. This has the advantage of letting the user check the status throughout the entire session, not just at the start.
> 
> 1. VSCode - I'm not sure if we can do better. I don't think notifications respect newlines + there's no "popup" concept in VSCode. We can only use notifications, quickpick, open a new editor tab prefilled with text 🤷
> 2. IntelliJ - we can do a popup, yeah
> 
> Regardless, this is what it's going to look like in old plugin/extension versions. Unless we come up with something very cunning

The left sidebar is probably a good candidate for this on vs code

gememma:
> > Having a persistent UI component has been itching me for a while but I think it should be out of scope for this specific issue. It'd require more design and architecture decisions. We can consider not having this feature in IDEs for now.
> 
> @gememma In that case, lets just not have this in the IDEs for the moment. Instead, please open follow-up issues in IntelliJ and VS Code to display this in a persistent component.

I agree that this wall-of-text-style notification isn't very nice to look at or parse for details. I'll remove the call in the extension code and at least we know that the information itself is easily sent to the IDE when we figure out the best way to present it. I can also open some issues to track that 👍 thanks all @aviramha @eyalb181 @Razz4780

mirrord/cli/src/main.rs
aviramha: Yeah that's what I had in mind as well. Thanks @Razz4780

Razz4780 (approved):

Issue #2155, opened 2024-01-05
Add JSON log output format


aviramha:
Add to `AgentConfig` a field that controls the log format, default being what we have now - Textual, and new would be JSON.
Change Agent to read that env and apply that.

PR #2449, opened 2024-05-15
Adds `agent.json_log` config for priting agent logs as json.

meowjesty:
- Issue: #2155 

- New config under `agent`:
  - `json_log` allows the user to control whether to use the normal tracing format, or json tracing;


mirrord/cli/src/config.rs
```diff
@@ -103,6 +103,10 @@ pub(super) struct ExecArgs {
     #[arg(short = 'l', long)]
     pub agent_log_level: Option<String>,
 
+    /// Controls whether the agent produce logs in readable tracing mode or json.
+    #[arg(long)]
```
aviramha, line 107: We don't add new flags to the cli unless those are crucial.

aviramha (changes requested):

mirrord/cli/src/config.rs
meowjesty: Removed!

aviramha (approved):

PR #2450, opened 2024-05-16
Improve `mirrord-kube` errors

Razz4780:
1. Made `RuntimeData::from_pod` public to avoid refetching multiple objects in the operator (no need to do n queries for deployment/rollout targets, consistent view of target state)
2. Improved `KubeApiError`. Error messages were starting with `mirrord-layer: `, which was confusing. Some errors produced really weird messages, like ```"mirrord-layer: Deployment: `Label for rollout: my-rollout, not found! not found!`"```
3. Payloads to kube POST requests are no longer constructed with `json!` macro and then parsed :skull: 
4. Changed `AgentConfig::image_pull_secrets` a bit to catch malformed secrets before attempting to create the agent and be more user friendly

### DIFF
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, path::Path};
+use std::{fmt, path::Path};
 
 use k8s_openapi::api::core::v1::{ResourceRequirements, Toleration};
 use mirrord_analytics::CollectAnalytics;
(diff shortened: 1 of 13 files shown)


mirrord/config/src/agent.rs
```diff
@@ -337,6 +350,14 @@ impl AgentImageFileConfig {
     }
 }
 
+/// <!--${internal}-->
+/// Specifies a secret reference for the agent pod.
+#[derive(Clone, Debug, PartialEq, Eq, JsonSchema, Deserialize, Serialize)]
+pub struct AgentPullSecret {
+    /// Name of the secret.
+    pub name: String,
+}
```
meowjesty, line 359:
```suggestion
///
/// Name of the secret.
pub struct AgentPullSecret(String);
```

The most complex suggestion to date.

mirrord/kube/src/api/container/job.rs
```diff
@@ -129,29 +132,44 @@ where
         self.inner.params()
     }
 
-    fn as_update(&self) -> Result<Self::Update> {
+    fn as_update(&self) -> Self::Update {
```
meowjesty, line 135: This was already the case, but, why is this function called `as_update`?

meowjesty (approved):
Have only 1 very important thing that might be worth changing.

But I'm going to risk it and approve anyway.

mirrord/config/src/agent.rs
Razz4780: Yeah, it would be simpler. But would break compatibility :sweat_smile:

mirrord/kube/src/api/container/job.rs
Razz4780: No idea :eyes: 
Possible interpretation - an update to cluster state?

PR #2461, opened 2024-05-23
Give config printout subtask a success message

gememma:
Fixes the last line of the config printout being repeated due to `None` success message.


mirrord/cli/src/main.rs
```diff
@@ -112,9 +114,8 @@ where
         &config,
         false,
     );
-    sub_progress_config.success(None);
```
aviramha, line 115: I think just "config summary" is okay

aviramha: maybe rename `sub_progress_config` to `cfg_summary_progress`

aviramha (approved):
Nice!

PR #2463, opened 2024-05-24
Verify target state before creating the agent

### LINKED ISSUE #462: Non-indicative error when running on a Completed pod
### Bug Description

When running on a Pod with state=Completed, we get a non-informative error as can be seen in the logs. We need to detect that the pod is in a running state, and if it isn't, display a more helpful error.

### Steps to Reproduce

Run mirrord with IMPERSONATED_POD_NAME=<name of a pod with state=Completed)

### Backtrace

_No response_

### Relevant Logs

```shell
❯ mirrord exec -c --steal --pod-namespace some-namespace --pod-name some-pod ./bin/main
thread '<unnamed>' panicked at 'unexpected response - expected env vars response None', mirrord-layer/src/lib.rs:407:13
fatal runtime error: failed to initiate panic, error 5
[1]    60108 abort      mirrord exec -c --steal --pod-namespace some-namespace --pod-name  ./bin/main
```

MacOS

### Local process

Go

### Local process version

_No response_

### Additional Info

_No response_

Razz4780:
Closes #462 

Added some extra checks on the targeted pod (described in `RuntimeData::from_pod` doc). Example output when targeting a pod in deletion:
Screenshot from 2024-05-24 13-47-21

Additionaly, when targeting multi-pod deployments/rollouts, mirrord scans the list of matching pod and picks the first one that passes status check (instead of always using the first one on the list)

### DIFF
--- a/mirrord/kube/src/api/runtime.rs
+++ b/mirrord/kube/src/api/runtime.rs
@@ -75,6 +75,12 @@ pub struct RuntimeData {
 }
 
 impl RuntimeData {
+    /// Extracts data needed to create the mirrord-agent targeting the given [`Pod`].
+    /// Verifies that the [`Pod`] is ready to be a target:
+    /// 1. pod is in "Running" phase,
+    /// 2. pod is not in deletion,
+    /// 3. all regular containers are ready (ephemeral containers are not inspected),
+    /// 4. viable target container is present.
     pub fn from_pod(pod: &Pod, container_name: Option<&str>) -> Result<Self> {
         let pod_name = pod
             .metadata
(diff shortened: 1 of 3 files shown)


mirrord/kube/src/api/runtime.rs
```diff
@@ -75,24 +75,53 @@ pub struct RuntimeData {
 }
 
 impl RuntimeData {
+    /// Extracts data needed to create the mirrord-agent targeting the given [`Pod`].
+    /// Verifies that the [`Pod`] is ready to be a target:
+    /// 1. pod is in "Running" phase,
+    /// 2. pod is not in deletion,
+    /// 3. all regular containers are ready (ephemeral containers are not inspected),
```
t4lz, line 82: Why do we need all regular containers to be ready and not just a viable target one?
And why don't we care about `initContainerStatuses`?

t4lz (commented):
Very nice, just 2 questions:

mirrord/kube/src/api/runtime.rs
Razz4780: Regarding second question - I think all `ready` containers imply that init containers have finished successfully

t4lz: > init containers run to completion before any app containers start

But not [sidecar containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/):
> If an init container is created with its restartPolicy set to Always, it will start and remain running during the entire life of the Pod.

t4lz (approved):
Approving because it's an improvement as is either way.

Maybe we should ask @aviramha about:
1. Using a pod when some containers are not ready, but the target (or **a** posiible target) is ready.
2. Requiring all init containers (for sidecar containers) to have a valid status.

aviramha:
> Approving because it's an improvement as is either way.
> 
> 
> 
> Maybe we should ask @aviramha about:
> 
> 1. Using a pod when some containers are not ready, but the target (or **a** posiible target) is ready.
> 
> 2. Requiring all init containers (for sidecar containers) to have a valid status.

The only precondition should be the target container imo

t4lz (approved):

PR #2477, opened 2024-05-29
Concurrent steal tests fixed

Razz4780:
Following https://github.com/metalbear-co/mirrord/pull/2476

### DIFF
--- a/mirrord/operator/src/crd.rs
+++ b/mirrord/operator/src/crd.rs
@@ -21,9 +21,6 @@ pub const TARGETLESS_TARGET_NAME: &str = "targetless";
 pub struct TargetSpec {
     /// None when targetless.
     pub target: Option<Target>,
-    /// This field is deprecated.
-    /// New operator versions always return [`None`].
-    pub port_locks: Option<Vec<TargetPortLock>>,
 }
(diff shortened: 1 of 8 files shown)


mirrord/operator/src/crd.rs
```diff
@@ -21,7 +21,8 @@ pub const TARGETLESS_TARGET_NAME: &str = "targetless";
 pub struct TargetSpec {
     /// None when targetless.
     pub target: Option<Target>,
-    #[deprecated = "port locks are enforced only in the operator, new operator versions always return an empty list"]
```
meowjesty, line 24: Why remove this in favor of docs?

meowjesty (approved):
:+1: 

Have 1 question though.

mirrord/operator/src/crd.rs
Razz4780: Too many warnings in the operator code, I got tired of copy-pasting `#[allow(deprecated)]`. But I figured I can just remove the field, it's backwards-compatible ^^

meowjesty: Nice!

PR #2484, opened 2024-05-31
Add support for chunked HTTP data requests

### LINKED ISSUE #2478: Support Chunked Request Messages in HTTP Header Filter
See issue #929

gememma:
Closes #2478

### DIFF
--- a/changelog.d/2478.added.md
+++ b/changelog.d/2478.added.md
@@ -1,0 +1,1 @@
+Add support for Chunked Request Messages in HTTP Header Filter.
--- a/changelog.d/2478.added.md
+++ b/changelog.d/2478.added.md
@@ -1,1 +1,1 @@
-Add support for Chunked Request Messages in HTTP Header Filter.
+Added support for intercepting streaming HTTP requests with an HTTP filter.
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -1,11 +1,9 @@
 use std::{
     collections::{HashMap, HashSet},
     net::{IpAddr, Ipv4Addr, SocketAddr},
-    task::Poll,
 };
 
 use fancy_regex::Regex;
-use futures::poll;
 use http_body_util::BodyExt;
 use hyper::{
     body::Incoming,
@@ -14,9 +12,9 @@ use hyper::{
 };
 use mirrord_protocol::{
     tcp::{
-        ChunkedFrame, ChunkedRequest, ChunkedRequestBody, ChunkedRequestError, DaemonTcp,
-        HttpRequest, HttpResponseFallback, InternalHttpBody, InternalHttpRequest, StealType,
-        TcpClose, TcpData, HTTP_CHUNKED_VERSION, HTTP_FILTERED_UPGRADE_VERSION,
+        ChunkedRequest, ChunkedRequestBody, ChunkedRequestError, DaemonTcp, HttpRequest,
+        HttpResponseFallback, InternalHttpBody, InternalHttpBodyFrame, InternalHttpRequest,
+        StealType, TcpClose, TcpData, HTTP_CHUNKED_VERSION, HTTP_FILTERED_UPGRADE_VERSION,
         HTTP_FRAMED_VERSION,
     },
     ConnectionId, Port,
@@ -35,7 +33,7 @@ use crate::{
         connections::{
             ConnectionMessageIn, ConnectionMessageOut, StolenConnection, StolenConnections,
         },
-        http::HttpFilter,
+        http::{Frames, HttpFilter, IncomingExt},
         orig_dst,
         subscriptions::{IpTablesRedirector, PortSubscriptions},
         Command, StealerCommand,
@@ -171,136 +169,67 @@ impl Client {
                     },
                     mut body,
                 ) = request.request.into_parts();
-                let _ = tx
-                    .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Start(
-                        HttpRequest {
-                            internal_request: InternalHttpRequest {
-                                method,
-                                uri,
-                                headers,
-                                version,
-                                body: (),
-                            },
-                            connection_id,
-                            request_id,
-                            port: request.port,
-                        },
-                    )))
-                    .await;
-                tracing::error!("sent header");
-
-                // If frames are immediately available, send them
-                let mut ready_frames: Vec<ChunkedFrame> = vec![];
-                let mut frame_fut: http_body_util::combinators::Frame<Incoming> = body.frame();
-                // Poll the future without moving the frame
-                let mut is_last = false;
-                while let Poll::Ready(ready_frame) = poll!(&mut frame_fut) {
-                    tracing::error!("{:?}", ready_frame);
-                    match ready_frame {
-                        Some(Ok(ok_frame)) => {
-                            let frame = if ok_frame.is_data() {
-                                ChunkedFrame::Data(ok_frame.into_data().unwrap().into())
-                            } else if ok_frame.is_trailers() {
-                                ChunkedFrame::Trailer(ok_frame.into_trailers().unwrap())
-                            } else {
-                                // ignore
-                                frame_fut = body.frame();
-                                continue;
-                            };
-                            ready_frames.push(frame);
-                            frame_fut = body.frame();
-                        }
-                        Some(Err(_)) => {
-                            // Error occured while sending request frame
-                            let frame_err = ChunkedRequest::Error(ChunkedRequestError {
+                match body.next_frames(true).await {
+                    Err(..) => return,
+                    Ok(Frames { frames, is_last }) => {
+                        let frames = frames
+                            .into_iter()
+                            .map(InternalHttpBodyFrame::try_from)
+                            .filter_map(Result::ok)
+                            .collect();
+                        let message =
+                            DaemonTcp::HttpRequestChunked(ChunkedRequest::Start(HttpRequest {
+                                internal_request: InternalHttpRequest {
+                                    method,
+                                    uri,
+                                    headers,
+                                    version,
+                                    body: frames,
+                                },
                                 connection_id,
                                 request_id,
-                            });
-                            let _ = tx.send(DaemonTcp::HttpRequestChunked(frame_err)).await;
+                                port: request.port,
+                            }));
+                        if tx.send(message).await.is_err() || is_last {
                             return;
                         }
-                        None => {
-                            is_last = true;
-                            break;
-                        }
                     }
                 }
+                tracing::error!("sent header");
 
-                if !(ready_frames.is_empty()) || is_last {
-                    let chunked = ChunkedRequestBody {
-                        frames: ready_frames,
-                        is_last,
-                        connection_id,
-                        request_id,
-                    };
-                    if tx
-                        .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(chunked)))
-                        .await
-                        .is_err()
-                    {
-                        return;
-                    }
-                }
-
-                if is_last {
-                    return;
-                }
-
-                // Start processing incoming frames
-                while let Some(incoming_frame) = frame_fut.await {
-                    // if !self.subscribed_connections.contains(&request.connection_id) {
-                    //     // if client has unsubscibed, stop sending data
-                    //     break;
-                    // }
-
-                    tracing::error!("{:?}", incoming_frame);
-                    match incoming_frame {
-                        Ok(ok_frame) => {
-                            let frame = if ok_frame.is_data() {
-                                ChunkedFrame::Data(ok_frame.into_data().unwrap().into())
-                            } else if ok_frame.is_trailers() {
-                                ChunkedFrame::Trailer(ok_frame.into_trailers().unwrap())
-                            } else {
-                                // ignore
-                                frame_fut = body.frame();
-                                continue;
-                            };
-                            let chunked = ChunkedRequestBody {
-                                frames: vec![frame],
-                                is_last: false,
-                                connection_id,
-                                request_id,
-                            };
-                            if tx
-                                .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(chunked)))
-                                .await
-                                .is_err()
-                            {
+                loop {
+                    match body.next_frames(false).await {
+                        Ok(Frames { frames, is_last }) => {
+                            let frames = frames
+                                .into_iter()
+                                .map(InternalHttpBodyFrame::try_from)
+                                .filter_map(Result::ok)
+                                .collect();
+                            let message = DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(
+                                ChunkedRequestBody {
+                                    frames,
+                                    is_last,
+                                    connection_id,
+                                    request_id,
+                                },
+                            ));
+                            if tx.send(message).await.is_err() || is_last {
                                 return;
                             }
                         }
                         Err(_) => {
-                            // Error occured while sending request frame
-                            let frame_err = ChunkedRequest::Error(ChunkedRequestError {
-                                connection_id,
-                                request_id,
-                            });
-                            let _ = tx.send(DaemonTcp::HttpRequestChunked(frame_err)).await;
+                            let _ = tx
+                                .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Error(
+                                    ChunkedRequestError {
+                                        connection_id,
+                                        request_id,
+                                    },
+                                )))
+                                .await;
                             return;
                         }
                     }
-
-                    frame_fut = body.frame();
                 }
-                let chunked = ChunkedRequestBody {
-                    frames: vec![],
-                    is_last: true,
-                    connection_id,
-                    request_id,
-                };
-                let _ = tx
-                    .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(chunked)))
-                    .await;
             } else if framed {
                 let Ok(request) = request.into_serializable().await else {
                     return;
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -1,9 +1,11 @@
 use std::{
     collections::{HashMap, HashSet},
     net::{IpAddr, Ipv4Addr, SocketAddr},
+    task::Poll,
 };
 
 use fancy_regex::Regex;
+use futures::poll;
 use http_body_util::BodyExt;
 use hyper::{
     body::Incoming,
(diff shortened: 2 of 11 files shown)


mirrord/protocol/src/tcp.rs
```diff
@@ -73,6 +73,14 @@ pub enum DaemonTcp {
     SubscribeResult(RemoteResult<Port>),
     HttpRequest(HttpRequest<Vec<u8>>),
     HttpRequestFramed(HttpRequest<InternalHttpBody>),
+    HttpRequestChunked(Result(ChunkedRequest, hyper::Error)),
+}
+
+/// Contents of a chunked message from server
+pub enum ChunkedRequest {
+    Start(HeaderMap),
```
Razz4780, line 81: This should probably be `HttpRequest<()>` - the same as in `DaemonTcp::HttpRequest` and `DaemonTcp::HttpRequestFramed`, just without the body. We need more info than just the headers, e.g. path, original destination port, request id, probably more

mirrord/protocol/src/tcp.rs
```diff
@@ -73,6 +73,14 @@ pub enum DaemonTcp {
...
+    Body(TcpData),
```
Razz4780, line 82: I'd name this one `Frame` and make it contain whatever the original `hyper` frame contains (bytes or `HeaderMap`). It also needs the connection id and the request id (so that the internal proxy knows which request this belongs to)

mirrord/protocol/src/tcp.rs
```diff
@@ -73,6 +73,14 @@ pub enum DaemonTcp {
...
+    End,
```
Razz4780, line 83: Also needs connection id and request id

Razz4780 (changes requested):
Further optimizations o consider:
1. Stop sending frames once the client disconnects/unsubscribes
2. Remove the `ChunkedRequest::End` variant and inline this information. Requires changing the `ChunkedRequest::Body` variant into a struct like:
```rust
struct ChunkedRequestBody {
    frame: ChunkedRequestFrame,
    is_last_frame: bool,
}
```
3. If frames are already available from hyper's `Incoming` stream (we can get them without `.await` or we know that `.await` will not block, send them together to minimize amount of agent-client messages we send.

mirrord/agent/src/steal/connection.rs
```diff
@@ -147,10 +157,66 @@ impl Client {
...
+                        Poll::Ready(Some(ready_frame)) => {
+                            ready_frames.push(ready_frame.unwrap());
+                            frame = body.frame();
+                            continue;
+                        }
+                        _ => break,
+                    }
+                }
+                for frame in ready_frames {
+                    let Ok(request) = request.into_serializable_chunked(frame.into_data()).await
```
Razz4780, line 182: The other variant of `Frame` is also interesting (`frame.into_trailers()`), we want to send it as well

mirrord/agent/src/steal/connection.rs
```diff
@@ -147,10 +157,66 @@ impl Client {
...
+                        // if client has unsubscibed, stop sending data
+                        break;
+                    }
+
+                    let request = match incoming_frame {
+                        Err(error) => Err(error),
+                        Ok(ok_frame) => {
+                            return if ok_frame.is_data() {
+                                ok_frame.into_data()
+                            };
```
Razz4780, line 201: This does not compile, does it? :sweat_smile:

Razz4780 (changes requested):
In general, we never want to exit early from this background task. If anything goes wrong, we want to send the error to the client

mirrord/agent/src/steal/connection.rs
```diff
@@ -147,10 +150,86 @@ impl Client {
...
+                }
+                let chunked = ChunkedRequestBody {
+                    frames: vec![],
+                    is_last: true,
+                    connection_id,
+                    request_id,
+                };
+                let _ = tx
+                    .send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(chunked)))
+                    .await;
```
Razz4780, line 231:
```suggestion
                loop {
                    match body.next_frames(false).await {
                        Ok(Frames { frames, is_last }) => {
                            let frames = frames.into_iter().filter_map(|frame| if frame.is_data() {
                                frame.into_data().ok().map(ChunkedFrame::Data)
                            } else if frame.is_trailers() {
                                frame.into_trailers().ok().map(ChunkedFrame::Trailers)
                            } else {
                                None
                            }).collect();
                            let message = DaemonTcp::HttpRequestChunked(ChunkedRequest::Body(ChunkedRequestBody {
                                frames,
                                is_last,
                                connection_id,
                                request_id,
                            }));

                            if tx.send(message).await.is_err() || is_last {
                                return;
                            }
                        }
                        Err(_) => {
                            let _ = tx.send(DaemonTcp::HttpRequestChunked(ChunkedRequest::Error { connection_id, request_id })).await;
                            return;
                        }
                    }
                }
```

mirrord/agent/src/steal/connection.rs
```diff
@@ -147,10 +150,86 @@ impl Client {
...
+                                headers,
+                                version,
+                                body: (),
+                            },
+                            connection_id,
+                            request_id,
+                            port: request.port,
+                        },
+                    )))
+                    .await;
```
Razz4780, line 188:
```suggestion
                match body.next_frames(true).await {
                    Err(..) => return,
                    Ok(Frames { frames, is_last }) => {
                        let message = todo!(); // prepare first message
                        if tx.send(message).await.is_err() || is_last {
                            return;
                        }
                    }
                }
```
We'd like to send the ready frames withing the first mirrord-protocol message

mirrord/agent/src/steal/connection.rs
```diff
@@ -147,10 +149,108 @@ impl Client {
...
+                                if frame.is_data() {
+                                    frame
+                                        .into_data()
+                                        .ok()
+                                        .map(|x| ChunkedFrame::Data(x.to_vec()))
+                                } else if frame.is_trailers() {
+                                    frame.into_trailers().ok().map(ChunkedFrame::Trailer)
+                                } else {
+                                    None
+                                }
```
Razz4780, line 187: Since we're going to do conversion between `ChunkedFrame` and `Frame<Bytes>` in multiple places, you can implement some traits ^^
```rust
impl From<ChunkedFrame> for Frame<Bytes> {
    ...
}

impl TryFrom<Frame<Bytes>> for ChunkedFrame {
    type Error = Frame<Bytes>;

    ...
}
```
Then it's just `let frames = frames.into_iter().map(ChunkedFrame::try_from).filter_map(Result::ok).collect()`

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +274,40 @@ impl IncomingProxy {
...
+                                body: (),
+                            },
+                            connection_id: req.connection_id,
+                            request_id: req.request_id,
+                            port: req.port,
+                        };
+                        let http_stream = ReceiverStreamReplay::new(rx);
+
+                        // TODO: pass frames to tx
+                        for frame in req.internal_request.body.iter().cloned() {
```
Razz4780, line 297:
```suggestion
                        for frame in req.internal_request.body {
```
No need to clone, right?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -53,26 +53,29 @@ pub async fn handshake(
...
     pub async fn send(
         &mut self,
-        req: HttpRequestFallback,
+        request: HttpRequestFallback,
     ) -> InterceptorResult<Response<Incoming>, InterceptorError> {
+        let request = if let HttpRequestFallback::Chunked(_req, replay) = request {
+            Request::builder()
+                .version(Version::HTTP_11)
+                .body(StreamBody::new(replay.stream()).boxed())
+                .unwrap()
```
Razz4780, line 62: Gotta add more stuff like uri and headers
I think now that `HttpRequestFallback::Chunked` variant holds the replay stream, we can move this logic into `HttpRequestFallback::into_hyper`? It's ok to consume the value (given that replay stream implements `Clone` in correct way)

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -258,6 +257,16 @@ impl HttpConnection {
                         .await
                         .map(HttpResponseFallback::Fallback)
                     }
+                    HttpRequestFallback::Chunked(_, _) => {
+                        HttpResponse::<Vec<u8>>::from_hyper_response(
```
Razz4780, line 261: I think we want to use the `HttpResponse<InternalHttpBody>` type (it's preferred over `HttpResponse<Vec<u8>>` and we know it's supported by the agent - we got a chunked request)

mirrord/protocol/src/streamed_request.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    // send `_request_1`, inspect response
+
+    // if we need to try again:
+    let _request_2 = Request::builder()
+        .version(Version::HTTP_11)
+        .body(StreamBody::new(replay.stream()).boxed())
+        .unwrap();
+    // send `_request_2`
+}
```
Razz4780, line 95: Delet :laughing:

mirrord/protocol/src/streamed_request.rs
```diff
@@ -0,0 +1,95 @@
...
+            idx: 0,
+        }
+    }
+
+    pub fn stream(&self) -> Self {
+        Self {
+            origin: self.origin.clone(),
+            idx: 0,
+        }
+    }
```
Razz4780, line 35: Now that it has correct `Clone`, we can remove this method

Razz4780 (changes requested):
I think we can move the `HttpRequestFallback` type to `mirrord-intproxy` crate, there's really no need for it in `mirrord-protocol`

changelog.d/2478.added.md
```diff
@@ -0,0 +1 @@
+Add support for Chunked Request Messages in HTTP Header Filter.
```
Razz4780, line 1: This should be understandable to the user, so probably something like `Added support for intercepting streaming HTTP requests with an HTTP filter`

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -151,6 +157,8 @@ pub struct IncomingProxy {
     background_tasks: BackgroundTasks<InterceptorId, MessageOut, InterceptorError>,
     /// For managing intercepted connections metadata.
     metadata_store: MetadataStore,
+    /// For managing streamed [`DaemonTcp::HttpRequestChunked`] request channels.
+    senders: HashMap<(ConnectionId, RequestId), Sender<InternalHttpBodyFrame>>,
```
Razz4780, line 161:
```suggestion
    request_body_txs: HashMap<(ConnectionId, RequestId), Sender<InternalHttpBodyFrame>>,
```
Just a suggestion

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +278,53 @@ impl IncomingProxy {
                     interceptor.send(req).await;
                 }
             }
+            DaemonTcp::HttpRequestChunked(req) => {
+                tracing::error!("{:?}", req);
```
Razz4780, line 282: Delet :wastebasket:

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +278,53 @@ impl IncomingProxy {
...
+                                body: http_stream,
+                            },
+                            connection_id: req.connection_id,
+                            request_id: req.request_id,
+                            port: req.port,
+                        };
+
+                        for frame in req.internal_request.body {
+                            let _ = tx.send(frame).await;
+                        }
```
Razz4780, line 302: Sending these should happen after passing the request to the interceptor. No one is receiving from the other end of the channel yet, so if you we have more than 12 frames ready, it's a deadlock.

Also, we should not ignore the send errors here. If sending fails, we should drop the sender and return early

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +278,53 @@ impl IncomingProxy {
...
+                        let interceptor = self.get_interceptor_for_http_request(&http_req)?;
+                        if let Some(interceptor) = interceptor {
+                            interceptor.send(http_req).await;
+                        }
+                    }
+                    ChunkedRequest::Body(body) => {
+                        let key = &(body.connection_id, body.request_id);
+                        if let Some(tx) = self.senders.get(key) {
+                            for frame in body.frames {
+                                let _ = tx.send(frame).await;
```
Razz4780, line 316: If send fails, we should remove the sender from `self.senders`

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +278,53 @@ impl IncomingProxy {
...
+                            }
+                        }
+                        if body.is_last {
+                            let _ = self.senders.remove(key);
```
Razz4780, line 320:
```suggestion
                            self.senders.remove(key);
```

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -270,6 +278,53 @@ impl IncomingProxy {
...
+                        }
+                    }
+                    ChunkedRequest::Error(err) => {
+                        tracing::trace!("ChunkedRequest error received for connection id: {:?}, request id: {:?}", err.connection_id, err.request_id);
```
Razz4780, line 324: We gotta remove the sender from `self.senders`

mirrord/intproxy/src/proxies/incoming.rs
Razz4780, line 161: We need to remove some senders once the interceptor task finishes

mirrord/intproxy/src/proxies/incoming.rs
Razz4780, line 324:
```suggestion
                        tracing::trace!(?err, "ChunkedRequest error received");
```

mirrord/intproxy/src/proxies/incoming/interceptor.rs
```diff
@@ -258,6 +257,16 @@ impl HttpConnection {
                         .await
                         .map(HttpResponseFallback::Fallback)
                     }
+                    HttpRequestFallback::Streamed(..) => {
+                        HttpResponse::<InternalHttpBody>::from_hyper_response(
```
Razz4780, line 261:
```suggestion
                        // Returning `HttpResponseFallback::Framed` variant is safe - streaming requests require an even higher mirrord-protocol version
                        HttpResponse::<InternalHttpBody>::from_hyper_response(
```

mirrord/protocol/src/tcp.rs
```diff
@@ -221,38 +259,117 @@ where
...
+#[derive(Clone, Debug)]
 pub enum HttpRequestFallback {
     Framed(HttpRequest<InternalHttpBody>),
     Fallback(HttpRequest<Vec<u8>>),
+    Streamed(HttpRequest<StreamingBody>),
+}
+
+#[derive(Debug)]
+pub struct StreamingBody {
+    origin: Arc<Mutex<(Receiver<InternalHttpBodyFrame>, Vec<InternalHttpBodyFrame>)>>,
```
Razz4780, line 293:
```suggestion
    /// Shared with instances acquired via [`Clone`].
    /// Allows the clones to receive a copy of the data.
    origin: Arc<Mutex<(Receiver<InternalHttpBodyFrame>, Vec<InternalHttpBodyFrame>)>>,
```

mirrord/protocol/src/tcp.rs
```diff
@@ -221,38 +259,117 @@ where
...
+    idx: usize,
```
Razz4780, line 294:
```suggestion
    /// Index of the next frame to return from the buffer.
    /// If outside of the buffer, we need to poll the stream to get the next frame.
    /// Local state of this instance, zeroed when cloning.
    idx: usize,
```

Razz4780 (changes requested):
Cool, it works! ^^
Some changes need to be done in intproxy's `IncomingProxy`.

It could also use unit tests on both sides (intproxy and agent).

Unit test on the intproxy side could be done as follows:
1. Create `BackgroundTasks` instance and register `IncomingProxy` task there.
2. Simulate user's HTTP server: open `TcpListener`, spawn a background task that accepts a connection and uses `hyper` to receive a request from it (use `Incoming` body type)
3. Use the registered `IncomingProxy` to send some `ChunkedRequest`
4. Verify that `Incoming` body from our dummy HTTP server receives frames as they come

Unit test on the agent side can be modeled on other tests in `agent/steal/connections/filtered.rs`. Just gotta make sure that the `FilteredStealTask` produces frames as they come

mirrord/protocol/src/tcp.rs
```diff
@@ -73,6 +74,43 @@ pub enum DaemonTcp {
...
+
+/// Contents of a chunked message from server.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum ChunkedRequest {
+    Start(HttpRequest<Vec<InternalHttpBodyFrame>>),
+    Body(ChunkedRequestBody),
+    Error(ChunkedRequestError),
+}
+
+impl ChunkedRequest {}
```
Razz4780, line 88: Delett' :wastebasket:

mirrord/protocol/src/tcp.rs
```diff
@@ -272,6 +395,11 @@ impl HttpRequestFallback {
 pub static HTTP_FRAMED_VERSION: LazyLock<VersionReq> =
     LazyLock::new(|| ">=1.3.0".parse().expect("Bad Identifier"));
 
+/// Minimal mirrord-protocol version that allows [`DaemonTcp::HttpRequestChunked`] instead of
+/// [`DaemonTcp::HttpRequest`].
+pub static HTTP_CHUNKED_VERSION: LazyLock<VersionReq> =
+    LazyLock::new(|| ">=1.6.0".parse().expect("Bad Identifier")); // TODO: check version number
```
Razz4780, line 401: Delett' the TODO
Also, I think Alex was quicker and already bumped the crate to `1.6.0`. Gotta bump again

mirrord/agent/src/steal/connection.rs
```diff
@@ -581,3 +659,152 @@ impl TcpConnectionStealer {
...
+    use mirrord_protocol::tcp::{ChunkedRequest, DaemonTcp, InternalHttpBodyFrame};
+    use tokio::{
+        net::{TcpListener, TcpStream},
+        sync::{
+            mpsc::{self, Receiver, Sender},
+            oneshot,
+        },
+    };
+    use tokio_stream::wrappers::ReceiverStream;
+    use x509_parser::nom::AsBytes;
```
Razz4780, line 685: Whaat happened here :laughing: 
Maybe let's not use this `nom` trait, this is confusing. Instead of `b"string".as_bytes().into()` you can do `b"string".to_vec().into()`

Razz4780 (changes requested):
Last nitpicks and we're mergin'

mirrord/protocol/src/tcp.rs
```diff
@@ -272,6 +393,11 @@ impl HttpRequestFallback {
 pub static HTTP_FRAMED_VERSION: LazyLock<VersionReq> =
     LazyLock::new(|| ">=1.3.0".parse().expect("Bad Identifier"));
 
+/// Minimal mirrord-protocol version that allows [`DaemonTcp::HttpRequestChunked`] instead of
+/// [`DaemonTcp::HttpRequest`].
+pub static HTTP_CHUNKED_VERSION: LazyLock<VersionReq> =
+    LazyLock::new(|| ">=1.6.1".parse().expect("Bad Identifier"));
```
Razz4780, line 399:
```suggestion
    LazyLock::new(|| ">=1.7.0".parse().expect("Bad Identifier"));
```

Razz4780 (approved):

PR #2485, opened 2024-06-01
Add warning for unnecessary port mappings

t4lz:
When you search for "port_mapping" on our discord server, most of the results are configuration files that contain a port mapping that maps each port their application uses to that same identical port.

Let's help users declutter their config files and be less confused, by explaining that field if they don't understand it and nudge them to remove unnecessary mappings.


aviramha (approved):
Great!

PR #2486, opened 2024-06-01
Improve http filter port UX

t4lz:
The semantics of port configurations change when filter is set and that might be a bit confusing.

This PR includes some changes that try to help users that might already be confused:
- When a filter is set - display all the filtered and unfiltered ports in the progress messages on start.
- When a filter is set and a port is bound, that is not included in the filter ports, and no unfiltered ports were specified, emit a warning. Most users probably won't see that as the default log level seems to be `error`, so we should not consider this UX hazard resolved, but I think it's still better to add this warning than not to.
- Consider the configuration invalid if the same port is included in both `feature.network.incoming.ports` and `feature.network.incoming.http_filter.ports`, as it's a bit suspicious and could mean the user does not understand those fields.
  - This one I'm not sure about, might be too strict. So let me know if I should remove it from the PR.
- Document default http filter ports.


mirrord/cli/src/main.rs
```diff
@@ -202,6 +202,56 @@ fn print_config<P>(
...
+        let filtered_ports_str = config
+            .feature
+            .network
+            .incoming
+            .http_filter
+            .get_filtered_ports()
+            .and_then(|filtered_ports| match filtered_ports.len() {
+                0 => None,
+                #[allow(clippy::indexing_slicing)]
+                1 => Some(format!("port {} (filtered)", filtered_ports[0])),
```
aviramha, line 219: use get instead of overriding clippy

aviramha (approved):
Great! Small nitpick.

PR #2490, opened 2024-06-03
Change Strings in SIP crate to PathBufs/OsStrings

### LINKED ISSUE #2198: Improve String handling in SIP crate
As pointed out in https://github.com/metalbear-co/mirrord/pull/2195#pullrequestreview-1841106655, we incorrectly use Strings in the SIP crate in places where an `OsString` or a `Vec<u8>` would be appropriate.
I think if we do everything correctly, no error handling should be needed for string conversions, and also nothing like `to_string_lossy`.

There isn't any known issue caused by this yet, but it might be possible to construct an example where this incorrect handling would lead to SIP-sidestepping not working, maybe with some special characters or encoding in file paths or script shebangs.

gememma:
Closes #2198 

This results in some lossy conversions where the SIP code is called from other crates which could be eliminated if all the Strings in the whole codebase were evaluated - there are plenty that would make more sense as OsStrings or Paths - but that would be a much larger issue. Eventually I think we could probably eliminate all the to_string_lossy() calls and avoid conversion to String altogether.


mirrord/layer/src/file/hooks.rs
```diff
@@ -57,7 +57,9 @@ fn update_ptr_from_bypass(ptr: *const c_char, bypass: Bypass) -> *const c_char {
         // inside mirrord's temp bin dir. The detour has returned us the original path of the file
         // (stripped mirrord's dir path), so now we carry out the operation locally, on the stripped
         // path.
-        Bypass::FileOperationInMirrordBinTempDir(stripped_ptr) => stripped_ptr,
+        Bypass::FileOperationInMirrordBinTempDir(path_buf) => {
+            path_buf.as_os_str().as_encoded_bytes().as_ptr() as _
```
t4lz, line 61: Returning ptr into a `PathBuf` that goes out of scope here, not good.

mirrord/layer/src/detour.rs
```diff
@@ -147,7 +144,7 @@ pub(crate) enum Bypass {
     /// We hooked a file operation on a path in mirrord's bin directory. So do the operation
     /// locally, but on the original path, not the one in mirrord's dir.
     #[cfg(target_os = "macos")]
-    FileOperationInMirrordBinTempDir(*const c_char),
+    FileOperationInMirrordBinTempDir(PathBuf),
```
t4lz, line 147: Why do you want to return a new `PathBuf` here?
Why not the new pointer like before, or just the length of the removed prefix?

mirrord/layer/src/exec_utils.rs
```diff
@@ -269,7 +282,8 @@ pub(crate) unsafe extern "C" fn _nsget_executable_path_detour(
     let res = FN__NSGET_EXECUTABLE_PATH(path, buflen);
     if res == 0 {
         let path_buf_detour = CheckedInto::<PathBuf>::checked_into(path as *const c_char);
-        if let Bypass(FileOperationInMirrordBinTempDir(later_ptr)) = path_buf_detour {
+        if let Bypass(FileOperationInMirrordBinTempDir(path_buf)) = path_buf_detour {
+            let later_ptr: *const i8 = path_buf.as_os_str().as_encoded_bytes().as_ptr() as _;
```
t4lz, line 286: If you change what `later_ptr` is, you also have to change the code that  uses that variable in line 292 and after that, because it does not hold the information it used to hold anymore.

The calculation of `prefix_len` is now wrong. Making the `prefix_len` correct would be enough to make the rest of the function correct.

`prefix_len` is supposed to hold the length of the prefix that should be removed from the path. 

It is currently calculated as
```Rust
            let prefix_len = later_ptr.offset_from(path);
```

But after the change `later_ptr` and `path` are pointers into two different buffers, so taking their offset makes no sense.

If you really want `FileOperationInMirrordBinTempDir` to hold a `PathBuf` (which doesn't look convenient to me), then you could calculate `prefix_len` by reducing the length of `path_buf` from `*buflen`.
However I think it would be nicer to either leave the value of that variant a pointer as it was, or hold the prefix length directly (but then you would have to calculate the new pointer in the places in the code where the returned pointer was used as-is).

mirrord/layer/src/detour.rs
gememma: We can't return a pointer into the original `String` because the switch over to using `PathBuf` means that operations like stripping a prefix result in a new `PathBuf`, where before they kept the same `String` buffer - so the `FileOperationInMirrordBinTempDir` needs to own the value, otherwise the pointer will be dangling

mirrord/layer/src/exec_utils.rs
gememma: I think calculating the length in a new way should be sufficient - I don't believe you can have `FileOperationInMirrordBinTempDir` contain a meaningful pointer with these changes

mirrord/layer/src/detour.rs
t4lz: I disagree. You don't need to return a `PathBuf`, as all the information the callers (that handle that variant) ever need, is how many bytes of the path to skip (this is equivalent to a new pointer into the original path buffer, either one works). If we still want to create new `PathBuf`s (which I don't think is necessary) while finding out that number, we can still at the end just check the length difference of the old and new `PathBuf`s, and only return that number.

PR #2493, opened 2024-06-04
Add CronJob and StatefulSet targetting when copy_target is enabled.

meowjesty:
- Issue: [#404](https://github.com/metalbear-co/operator/issues/404) and [#543](https://github.com/metalbear-co/operator/issues/543)

1. Adds `CronJob` as a possible `target` (requires `copy_target` feature to be enabled);
2. Adds `StatefulSet` as a possible `target` (requires `copy_target` feature to be enabled);


mirrord/config/src/target/cron_job.rs
```diff
@@ -0,0 +1,51 @@
...
+            f,
+            "{}{}",
+            self.container
+                .as_ref()
+                .map(|c| format!("{c}/"))
+                .unwrap_or_default(),
+            self.cron_job.clone()
+        )
+    }
+}
```
Razz4780, line 29: I don't think this impl is needed, most other target types (excluding `JobTarget`) don't have it :eyes: 
Could you either remove it or change the order of elements (right now the container is displayed before the cron job name). Applies to both `CronJobTarget` and `JobTarget`

Razz4780 (changes requested):
There's also one thing I missed when reviewing previous `JobTarget` PR - we need a feature flag in the operator status to check whether this target type is supported

Razz4780:
> * Issue: [#404](https://github.com/metalbear-co/operator/issues/404)

Issue not found ;/

aviramha:
> There's also one thing I missed when reviewing previous `JobTarget` PR - we need a feature flag in the operator status to check whether this target type is supported

We could trust the `mirrord ls` in terms of UX, no?

Razz4780:
> > There's also one thing I missed when reviewing previous `JobTarget` PR - we need a feature flag in the operator status to check whether this target type is supported
> 
> We could trust the `mirrord ls` in terms of UX, no?

Yes, job targets should not be listed unless operator supports them

aviramha:
> > > There's also one thing I missed when reviewing previous `JobTarget` PR - we need a feature flag in the operator status to check whether this target type is supported
> > 
> > 
> > We could trust the `mirrord ls` in terms of UX, no?
> 
> Yes, job targets should not be listed unless operator supports them

I'm saying we don't need a feature flag because it either returns it as an option or it doesn't.

aviramha:
If we want a feature flag for nicer error, I'd make a "target not supported" error with appropriate message and re-use it, as we're adding more and more targets, and it's a bit weird to have a feature flag for each

Razz4780:
> If we want a feature flag for nicer error, I'd make a "target not supported" error with appropriate message and re-use it, as we're adding more and more targets, and it's a bit weird to have a feature flag for each

Okok, makes sense

mirrord/config/src/target.rs
```diff
@@ -402,6 +419,12 @@ impl CollectAnalytics for &TargetConfig {
                         flags |= TargetAnalyticFlags::CONTAINER;
                     }
                 }
+                Target::CronJob(cron_job) => {
+                    flags |= TargetAnalyticFlags::JOB;
```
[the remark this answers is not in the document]
meowjesty: Changed.

mirrord/config/src/target/cron_job.rs
meowjesty: Removed from both places.

mirrord/config/src/lib.rs
```diff
@@ -389,7 +389,7 @@ impl LayerConfig {
                 .target
                 .path
                 .as_ref()
-                .map(|target| matches!(target, target::Target::Job(_)))
+                .map(|target| matches!(target, target::Target::Job(_) | target::Target::CronJob(_)))
```
Razz4780, line 392: `StatefulSet` as well

mirrord/config/src/target.rs
```diff
@@ -184,6 +188,7 @@ mirrord-layer failed to parse the provided target!
     >> deploy/<deployment-name>[/container/container-name]
     >> pod/<pod-name>[/container/container-name]
     >> job/<job-name>[/container/container-name]
+    >> cronjob/<cronjob-name>[/container/container-name]
```
Razz4780, line 191: `StatefulSet` as well

mirrord/config/src/target.rs
```diff
@@ -207,6 +212,7 @@ mirrord-layer failed to parse the provided target!
 /// - `container/{sample-container}`;
 /// - `containername/{sample-container}`.
 /// - `job/{sample-job}`;
+/// - `cronjob/{sample-job}`;
```
Razz4780, line 215: `StatefulSet` as well

mirrord/kube/src/api/runtime/stateful_set.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn container(&self) -> Option<&str> {
+        self.container.as_deref()
+    }
+
+    async fn get_labels(resource: &Self::Resource) -> Result<BTreeMap<String, String>> {
+        resource
+            .spec
+            .as_ref()
+            .and_then(|spec| spec.selector.match_labels.clone())
+            .ok_or_else(|| KubeApiError::missing_field(resource, ".spec.selector.match_labels"))
```
Razz4780, line 25:
```suggestion
            .ok_or_else(|| KubeApiError::missing_field(resource, ".spec.selector.matchLabels"))
```
It's the original name used in yaml

Razz4780 (changes requested):

Razz4780 (approved):

Issue #2488, opened 2024-06-03
readlink implementation


aviramha:
When node uses `fs.realpathSync` it sometimes calls `readlink` which isn't hooked by mirrord
we probably need to add a new message to the protocol, so we can support this hook if agent > specific version I guess.

PR #2495, opened 2024-06-05
Add file hook for readlink.

meowjesty:
- Issue: #2488

1. Adds the `readlink` hook;
    - gated behind `experimental.readlink` config to avoid protocol breakage;
3. Introduces new message in the protocol (new `FileRequest`);


mirrord/protocol/src/codec.rs
```diff
@@ -113,6 +114,7 @@ pub enum FileResponse {
     Open(RemoteResult<OpenFileResponse>),
     Read(RemoteResult<ReadFileResponse>),
     ReadLimited(RemoteResult<ReadFileResponse>),
+    ReadLink(RemoteResult<ReadLinkFileResponse>),
```
aviramha, line 117: I think this needs to be at end of the list to not break.

mirrord/layer/src/file/hooks.rs
```diff
@@ -1097,6 +1122,14 @@ pub(crate) unsafe fn enable_file_hooks(hook_manager: &mut HookManager) {
         FN__PREAD_NOCANCEL
     );
 
+    replace!(
```
aviramha, line 1125: Don't forget readlinkat

mirrord/agent/src/file.rs
```diff
@@ -401,6 +408,18 @@ impl FileManager {
...
     }
 
+    #[tracing::instrument(level = "trace", skip_all)]
+    pub(crate) fn read_link(
+        &mut self,
+        stripped_path: impl AsRef<Path>,
+    ) -> RemoteResult<ReadLinkFileResponse> {
+        let full_path = self.root_path.join(stripped_path);
+
+        read_link(full_path)
```
aviramha, line 418: Need to make sure the path we returned is normalized to be "from container" standpoint
Since we're basically calling readlink("/proc/PID/root/path..") I think it might return similar thing, which needs to be normalized..

aviramha (changes requested):

aviramha:
I just realized we have a hook for realpath that doesn't add new message type and I think should work pretty much the same.

aviramha:
Also this needs to go thru our file filter.

aviramha:
tbh we could open an issue for `readlinkat` and `freadlink` as not sure those would be required.

aviramha:
Sending user an alternative implementation https://github.com/metalbear-co/mirrord/tree/2488

aviramha:
My branch seems to be working - let's wrap it for now with an integration test and merge.

aviramha:
Okay actually Node keeps calling readlink + xstat until it doesn't see it's a symlink T_T

mirrord/agent/src/file.rs
meowjesty: When I try this out:

```
ln -s "/gatos/rajado.txt" "/gatos/tigrado.txt"
readlink("/gatos/tigrado.txt"); // in C program
> "rajado.txt"
```

So we're not messing up the path, right?

mirrord/layer/src/file/hooks.rs
meowjesty: Moved to issue #2503

mirrord/agent/src/file.rs
aviramha: shouldn't it be `/gatos/rajado.txt`?

aviramha:
Make the sample read path from argv so we can run more than one case easily.

aviramha:
User confirmed it fixes their issue.

mirrord/agent/src/file.rs
meowjesty: It is, my example was off.

```
read_link:make_request_with_response: mirrord_layer::proxy_connection: return=Ok(Ok(ReadLinkFileResponse { path: "/gatos/rajado.txt" })) path=Success("/gatos/tigrado.txt") request=ReadLinkFileRequest { path: "/gatos/tigrado.txt" }
```

aviramha (approved):
Great!

PR #2504, opened 2024-06-10
Fixed mirrord-agent/cli protocol negotation

aviramha:


mirrord/agent/src/entrypoint.rs
```diff
@@ -454,15 +453,19 @@ impl ClientConnectionHandler {
                 ))
                 .await?;
             }
-            ClientMessage::SwitchProtocolVersion(version) => {
+            ClientMessage::SwitchProtocolVersion(client_version) => {
+                let our_version = mirrord_protocol::VERSION.get().clone();
```
DmitryDodzin, line 457: Need also to have some initial value of `1.2.1` as `protocol_version` variable above

DmitryDodzin: Also don't need get it's a LazyLock

DmitryDodzin (changes requested):
Still need `1.2.1` version to start with & fix compilation in agent =]

mirrord/agent/src/entrypoint.rs
aviramha: got it - I think we can remove it since the switch protocol exists for 10 months already

DmitryDodzin: 😅

aviramha: I'm just afraid that having a base which is a very "old" version while none actually uses it (according to metrics) will create hard bugs (why is it this version??)

DmitryDodzin (approved):
Alright, Would be strange when return value of negotiation downgrades version rather than upgrading it from older version, though I can't think of any current problems with it just throwing it out there

aviramha:
@DmitryDodzin I thought about it too, but essentially that's how protocol neg should work 🤣 
also, this is what we do in the operator handling side (where we even emulate the message!)

PR #2505, opened 2024-06-10
cross rust analyzer

aviramha:
cc @gememma who started to look into it


gememma (approved):
Works on my machine too, LGTM :)

PR #2506, opened 2024-06-10
Remove from tests host name that often can't be found

t4lz:
See example flake here: https://github.com/metalbear-co/mirrord/actions/runs/9450763078/job/26030239466?pr=2486#step:10:252

I don't know that this is not actually an issue with mirrord, but since [we removed the last website with which the tests often flaked](https://github.com/metalbear-co/mirrord/pull/2471) I'm assuming we can just remove this one as well.


aviramha:
The flake is internet connectivity issue of GitHub.. :#

Issue #2386, opened 2024-04-21
agent stays when passing mirrord a non-existing executable


t4lz:
### Bug Description

When you call `mirrord exec` with an executable that doesn't exist, an agent pod keeps running until it is manually deleted.

Impact on users: low - not reported by any users so far AFAIK. Won't prevent them from doing anything, will just leave lingering unwanted agent pods running, that they have to delete themselves, if they ever have a typo when typing their executable's name.

### Steps to Reproduce

```
mirrord exec no-such-binary
```

### Your operating system and version

macOS 14.4.1

### Local process

does-not-exist

aviramha:
Are you sure that it's lingering for more than 60s? (Startup timeout)

t4lz:
> Are you sure that it's lingering for more than 60s? (Startup timeout)

yes:

![image](https://github.com/metalbear-co/mirrord/assets/12370607/67015121-44eb-4bce-bc10-c76628be90fd)

meowjesty:
Looking into this, found out it terminates after around 11 minutes lol.

![image](https://github.com/metalbear-co/mirrord/assets/43983236/218f870b-f856-4e37-8ead-f9eddf824e4c)

Working on decreasing this time though.

PR #2510, opened 2024-06-11
Kill intproxy on a bogus binary.

meowjesty:
- Issue: #2386


aviramha (approved):

mirrord/cli/src/main.rs
```diff
@@ -129,6 +129,10 @@ where
             }
         }
     }
+
+    // Kills the intproxy, freeing the agent.
+    execution_info.stop().await;
```
t4lz, line 134: 1. clippy.
2. please move above the `if`, so that we don't `return` without killing the intproxy.

mirrord/cli/src/execution.rs
```diff
@@ -362,4 +364,16 @@ impl MirrordExecution {
...
+    ///
+    /// Used when mirrord execution fails inside `execvp`.
+    pub async fn stop(self) {
+        let Self { mut child, .. } = self;
+
+        child
+            .kill()
+            .await
+            .expect("Could not kill intproxy child process!");
+    }
```
Razz4780, line 378:
```suggestion
    pub fn stop(self) {
        let Self { mut child, .. } = self;

        let _ = child.start_kill();
    }
```
We don't really need to wait for process to die and we can't do much in case of failure (e.g. intproxy has already died on its own for some reason)

mirrord/cli/src/main.rs
meowjesty: Changed Mr. Wackz.

PR #2520, opened 2024-06-12
Improve operator detection + errors in CLI

### LINKED ISSUE #2487: Improve operator discovery UX (CLI)
Currently decision whether to use the operator is made based on `.operator` field in the config:
1. Set to `true` - we try to use the operator and abort in case of failure
2. Set to `false` - we do not try to use the operator
3. Unset - we try to use the operator but fall back to running without it if we're not sure whether it's installed or not (we failed to fetch operator resource)

The problem is that failure described in 3. can occur even if the operator is installed (due to RBAC). To improve to flow we can use `kube::discovery::pinned_kind` to confirm that the operator is not installed (before we fall back to running raw). If `kube::discovery::pinned_kind` says that the operator is installed yet we failed to fetch it, we should abort and prompt the user with the usual RBAC warning

### LINKED ISSUE #2522: Improve seat count exceeded error on CLI
Currently errors with 403 and not very clear message.

Razz4780:
Closes #2487
Closes #2522

When user is not explicit about whether we should use the operator or not and simple GET on the operator resource fails, we use kube discovery API to confirm that the operator is not installed.

Also improved `CliError` type - more consistent diagnostical messages etc.


aviramha:
Looks like great improvement - I'd like to address maximum seat count error at the same time please.

Razz4780:
> Looks like great improvement - I'd like to address maximum seat count error at the same time please.

This is what it looks like now:

image

aviramha:
CI fails and it seems for valid reason?

Razz4780:
> CI fails and it seems for valid reason?

Yeah, I have to investigate. Works with containerd though, which is weird

Razz4780:
> CI fails and it seems for valid reason?

Found out it was docker being lazy with port forwarding, all works now ^^

DmitryDodzin (approved):
LGTM =] A couple of nitpicks

DmitryDodzin (approved):

PR #2532, opened 2024-06-17
Removed misleading expiration warning from the CLI

### LINKED ISSUE #2526: remove expiry message when not on trial
on self serve license expires every last few days of the billing month, creating a suboptimal experience.
We should consider adding this back when/if we know that subscription is cancelled somehow.

Razz4780:
Closes #2526


meowjesty (approved):
:+1:

PR #2549, opened 2024-06-25
When using linkerd mesh use `lo` interface for mirroring traffic

### LINKED ISSUE #2452: gRPC traffic is not mirrored when using the Linkerd service mesh
### Bug Description

When debugging a mirrored C# (.Net 8) micro service in Rider, breakpoints for incoming gRPC requests are not hit.

My service listens for HTTP requests on port 8080 and for gRPC requests on port 5000. I see that HTTP handlers are mirrored properly (well, at least the `/healthz` endpoint which is called periodically by the k8s probe). But the gRPC requests are not. Those gRPC requests actually come into the ingress (Contour/Envoy) as grpc-web requests, which are translated to gRPC by Envoy. Then, Envoy sends a meshed request to my service on port 5000. 

Removing the Linkerd proxy from my mirrored pod fixes the issue (my breakpoints are hit)

### Steps to Reproduce

1. Have a Contour ingress, injected with the Linkerd proxy
2. Have a gRPC service in a pod that is also injected by the Linkerd proxy
3. Use mirrord to mirror the service's traffic
4. Make a grpc-web call to the Contour ingress
5. Observe that the mirrored process doesn't receive the request
6. Remove the Linkerd proxy from the gRPC service pod
7. Make a grpc-web call to the Contour ingress
8. Observe that the mirrored process _does_ receive the request

### Backtrace

_No response_

### Relevant Logs

_No response_

macOS 14.4.1 (23E224)

### Local process

A C# (.Net 8) process

### Local process version

_No response_

### Additional Info

_No response_

DmitryDodzin:
closes #2452

### DIFF
--- a/mirrord/agent/src/sniffer.rs
+++ b/mirrord/agent/src/sniffer.rs
@@ -141,8 +141,8 @@ async fn prepare_sniffer(
     network_interface: Option<String>,
     mesh: Option<MeshVendor>,
 ) -> Result<RawCapture, AgentError> {
-    // Priority is whatever the user set as an option to mirrord, then we check if we're in an istio
-    // mesh, otherwise we try to get the appropriate interface.
+    // Priority is whatever the user set as an option to mirrord, then we check if we're in a mesh
+    // to use `lo` interface, otherwise we try to get the appropriate interface.
     let interface = match network_interface.or_else(|| mesh.map(|_| "lo".to_string())) {
         Some(interface) => interface,
         None => resolve_interface()
--- a/mirrord/agent/src/sniffer.rs
+++ b/mirrord/agent/src/sniffer.rs
@@ -143,12 +143,7 @@ async fn prepare_sniffer(
 ) -> Result<RawCapture, AgentError> {
     // Priority is whatever the user set as an option to mirrord, then we check if we're in an istio
     // mesh, otherwise we try to get the appropriate interface.
-    let interface = match network_interface.or_else(|| {
-        mesh.and_then(|mesh_vendor| match mesh_vendor {
-            MeshVendor::Linkerd | MeshVendor::Istio => Some("lo".to_string()),
-            MeshVendor::Kuma => None,
-        })
-    }) {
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/sniffer.rs
```diff
@@ -145,8 +145,8 @@ async fn prepare_sniffer(
     // mesh, otherwise we try to get the appropriate interface.
     let interface = match network_interface.or_else(|| {
         mesh.and_then(|mesh_vendor| match mesh_vendor {
-            MeshVendor::Linkerd | MeshVendor::Kuma => None,
-            MeshVendor::Istio => Some("lo".to_string()),
+            MeshVendor::Linkerd | MeshVendor::Istio => Some("lo".to_string()),
+            MeshVendor::Kuma => None,
```
aviramha, line 149: I think kuma is same, no?

DmitryDodzin: Yep, I looked at their iptables should also be lo

DmitryDodzin: Did update

mirrord/agent/src/sniffer.rs
```diff
@@ -143,12 +143,7 @@ async fn prepare_sniffer(
 ) -> Result<RawCapture, AgentError> {
     // Priority is whatever the user set as an option to mirrord, then we check if we're in an istio
     // mesh, otherwise we try to get the appropriate interface.
-    let interface = match network_interface.or_else(|| {
-        mesh.and_then(|mesh_vendor| match mesh_vendor {
-            MeshVendor::Linkerd | MeshVendor::Kuma => None,
-            MeshVendor::Istio => Some("lo".to_string()),
-        })
-    }) {
+    let interface = match network_interface.or_else(|| mesh.map(|_| "lo".to_string())) {
```
aviramha, line 146: maybe add a comment explanining this

aviramha (approved):
LGTM, small nitpick.

PR #2550, opened 2024-06-26
Extend operator headers

### LINKED ISSUE #2466: Add client certificate header to all operator APIs
We send the client certificate header only on connect, but we should send it in regards to any communication with the operator to provide smoother UX (failing earlier)

Razz4780:
Closes #2466 

CLI version and user name/hostname headers are sent with all operator requests.
Client certificate header:
1. `mirrord operator status`: we don't know operator license fingerprint before getting the resource, so header is not sent.
2. `mirrord exec`: as soon as we get the operator resource, we prepare the client cert header. If we fail, command aborts.
3. `mirrord ls` / `mirrord operator session`: as soon as we get the operator resource, we try to prepare the client cert header. If we fail, we continue without it

I kind of went full aggro on the operator client code and made more changes than necessary :V
1. Copied to `mirrord ls` the extra discovery step present in `mirrord exec` (https://github.com/metalbear-co/mirrord/issues/2487)
2. Moved some parts of operator client code to new separate submodules (errors, http upgrade request logic, connection wrapper, operator discovery)
3. Unified format of progress messages

### DIFF
--- a/mirrord/cli/src/connection.rs
+++ b/mirrord/cli/src/connection.rs
@@ -37,7 +37,7 @@ where
 {
     let mut operator_subtask = progress.subtask("checking operator");
 
-    let Some(mut api) = OperatorApi::try_new(config, analytics, &operator_subtask).await? else {
+    let Some(api) = OperatorApi::try_new(config, analytics, &operator_subtask).await? else {
         operator_subtask.success(Some("proceeding without operator"));
         return Ok(None);
     };
@@ -66,7 +66,7 @@ where
     }
 
     let mut user_cert_subtask = operator_subtask.subtask("preparing user credentials");
-    api.prepare_client_cert(analytics).await?;
+    let api = api.prepare_client_cert(analytics).await.into_certified()?;
     user_cert_subtask.success(Some("user credentials prepared"));
 
     let mut session_subtask = operator_subtask.subtask("starting session");
--- a/mirrord/cli/src/connection.rs
+++ b/mirrord/cli/src/connection.rs
@@ -61,14 +61,14 @@ where
     let mut version_cmp_subtask = operator_subtask.subtask("checking version compatibility");
     let compatible = api.check_operator_version(&version_cmp_subtask);
     if compatible {
-        version_cmp_subtask.success(None);
+        version_cmp_subtask.success(Some("operator version compatible"));
     } else {
-        version_cmp_subtask.failure(None);
+        version_cmp_subtask.failure(Some("operator version may not be compatible"));
     }
 
     let mut license_subtask = operator_subtask.subtask("checking license");
     match api.check_license_validity(&license_subtask) {
-        Ok(()) => license_subtask.success(None),
+        Ok(()) => license_subtask.success(Some("operator license valid")),
         Err(error) => {
             license_subtask.failure(Some("operator license expired"));
 
@@ -83,11 +83,13 @@ where
 
     let mut user_cert_subtask = operator_subtask.subtask("preparing user credentials");
     api.prepare_client_cert(analytics).await?;
-    user_cert_subtask.success(None);
+    user_cert_subtask.success(Some("user credentials prepared"));
 
     let mut session_subtask = operator_subtask.subtask("starting session");
     let connection = api.connect_in_new_session(config, &session_subtask).await?;
-    session_subtask.success(None);
+    session_subtask.success(Some("session started"));
+
+    operator_subtask.success(Some("using operator"));
 
     Ok(Some(connection))
 }
--- a/mirrord/cli/src/connection.rs
+++ b/mirrord/cli/src/connection.rs
@@ -36,9 +36,13 @@ where
     R: Reporter,
 {
     let mut operator_subtask = progress.subtask("checking operator");
+    if config.operator == Some(false) {
+        operator_subtask.success(Some("operator disabled"));
+        return Ok(None);
+    }
 
-    let Some(api) = OperatorApi::try_new(config, analytics, &operator_subtask).await? else {
-        operator_subtask.success(Some("proceeding without operator"));
+    let Some(api) = OperatorApi::try_new(config, analytics).await? else {
+        operator_subtask.success(Some("operator not found"));
         return Ok(None);
     };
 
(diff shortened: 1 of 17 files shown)


mirrord/operator/src/types.rs
```diff
@@ -12,3 +13,26 @@ pub struct LicenseInfoOwned {
     /// Subscription id encoded in the operator license extension.
     pub subscription_id: Option<String>,
 }
+
+/// HTTP header containing CLI version.
+/// Sent with each request to the mirrord operator.
+pub static MIRRORD_CLI_VERSION_HEADER: HeaderName =
```
aviramha, line 19: const?

Razz4780: Clippy complains :<

aviramha:
We don't want a license to be needed for the non actual usage APIs. It should be fallible and optional.

Razz4780:
> We don't want a license to be needed for the non actual usage APIs. It should be fallible and optional.

Do you mean operator license or user license?

aviramha:
> > We don't want a license to be needed for the non actual usage APIs. It should be fallible and optional.
> 
> Do you mean operator license or user license?

I meant user license but we shouldn't check the operator license for getting it's status anyway

Razz4780:
> > > We don't want a license to be needed for the non actual usage APIs. It should be fallible and optional.
> > 
> > 
> > Do you mean operator license or user license?
> 
> I meant user license but we shouldn't check the operator license for getting it's status anyway

So, summarizing:
1. operator status -> no check at all, plain GET with version/name/hostname headers
2. target connection -> operator license validity check, user license required
3. session management or listing targets -> no operator license validity check, fallible and optional user license
Correct?

aviramha:
I'd add the user license to the status too, just make it optional.

mirrord/operator/src/client.rs
```diff
@@ -1,557 +1,709 @@
...
-        let client_hash = self
-            .client_certificate
-            .as_ref()
-            .map(|cert| cert.public_key_data())
-            .as_deref()
-            .map(AnalyticsHash::from_bytes);
+    /// Prepares client [`Certificate`] to be sent in all subsequent requests to the operator.
+    /// In case of failure, state of this API instance does not change.
+    #[tracing::instrument(level = Level::TRACE, skip(reporter), err)]
+    pub async fn prepare_client_cert<R>(&mut self, reporter: &mut R) -> OperatorApiResult<()>
```
DmitryDodzin, line 316: Having some sort of middleware layer for `kube::Client` using their builder can be an elegant solution for keeping the certificate up to date for each request without having to call it after creating the operator client though there might be some complications and double clienting, alternatively there is maybe some merit in making it a type state machine where you have `OperatorApi<T = Unauthorised>` and only post upgrade of `prepare_client_cert` you have `OperatorApi<WithCertificate>` that has all the methods that require credentials

mirrord/operator/src/client.rs
```diff
@@ -1,557 +1,709 @@
...
-    /// Create websocket connection to operator.
-    #[tracing::instrument(level = "trace", skip(self))]
-    async fn connect_target(
-        &self,
-        session_info: OperatorSessionInformation,
-    ) -> Result<OperatorSessionConnection> {
-        let UserIdentity { name, hostname } = UserIdentity::load();
+    /// Checks features specified in the given [`LayerConfig`] against what is supported by the
+    /// detected operator installation.
+    fn check_feature_support(&self, config: &LayerConfig) -> OperatorApiResult<()> {
```
DmitryDodzin, line 638: rename this to `check_copy_target_feature_support`

DmitryDodzin (commented):
I think operator api can be a changed bit to prevent any issues that may happen post future changes

mirrord/cli/src/main.rs
```diff
@@ -565,11 +568,25 @@ async fn print_targets(args: &ListTargetArgs) -> Result<()> {
     // Try operator first if relevant
     let mut targets = match &layer_config.operator {
         Some(true) | None => {
-            let operator_targets = OperatorApi::list_targets(&layer_config).await;
-            match operator_targets {
-                Ok(targets) => {
-                    // adjust format to match non-operator output
-                    targets
+            let api = if layer_config.operator.is_some() {
```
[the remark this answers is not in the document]
Razz4780: Looks nice, thanks!

mirrord/operator/src/client.rs
Razz4780: Type state machine sound cool. I don't want to play with adding middleware though, Aviram added extra headers field to `kube::Config` and imho it's simpler

mirrord/cli/src/main.rs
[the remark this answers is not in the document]
Razz4780: I moved reading `config.operator` setting to `OperatorApi::try_new`, but I left the return type as it was (`Result<Option<_>, _>`). Imho there's no need for an extra type. Wdyt?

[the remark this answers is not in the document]
DmitryDodzin: Should be fine, I was just thinking of it as a helper so you won't have to do `.ok_or_else(|| /* Error that indicates that the operation requires a installed and valid operator */)`

mirrord/operator/src/client.rs
Razz4780: Updated ^^

mirrord/cli/src/main.rs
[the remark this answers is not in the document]
Razz4780: When the operation requires operator I just call `OperatorApi::new`

mirrord/cli/src/connection.rs
```diff
@@ -35,29 +35,13 @@ where
...
-
-    let api = if operator_required {
-        OperatorApi::new(config, analytics).await?.into()
-    } else {
-        OperatorApi::try_new(config, analytics).await?
-    };
-    let Some(mut api) = api else {
-        find_operator_subtask.success(Some("operator not found"));
+    let Some(api) = OperatorApi::try_new(config, analytics, &operator_subtask).await? else {
         operator_subtask.success(Some("proceeding without operator"));
```
DmitryDodzin, line 41: Feels like this may be a bit much first it's "checking operator" and inside of it it's "detecting operator" I would keep only one of them

DmitryDodzin (approved):
I would double check that the progress bar isn't a bit much but other than that looks good =]

PR #2552, opened 2024-06-26
Changed http path filter to include query params in match

### LINKED ISSUE #2551: http path filter runs on path only
http path filter runs on the path only:
image

```rs
            Self::Path(filter) => {
                let path = request.uri().path();
                filter
                    .is_match(path)
                    .inspect_err(|error| {
                        tracing::error!(path, ?error, "Error while matching path");
                    })
                    .unwrap_or(false)
            }
```
but actually it'd be more useful to run on the entire uri - I guess naming might confuse but I also guess that most users would assume path includes the query params.

aviramha:
Closes #2551

### DIFF
--- a/mirrord/agent/src/steal/http/filter.rs
+++ b/mirrord/agent/src/steal/http/filter.rs
@@ -58,15 +58,20 @@ impl HttpFilter {
                 headers.has_match(filter)
             }
 
-            Self::Path(filter) => {
-                let path = request.uri().path();
-                filter
-                    .is_match(path)
-                    .inspect_err(|error| {
-                        tracing::error!(path, ?error, "Error while matching path");
-                    })
-                    .unwrap_or(false)
-            }
+            Self::Path(filter) => request
+                .uri()
+                .into_parts()
+                .path_and_query
+                .map(|path_and_query| {
+                    let path = path_and_query.as_str();
+                    filter
+                        .is_match(path)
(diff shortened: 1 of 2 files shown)


Razz4780:
One remark - this is not backwards compatible. Filters like `^/my/exact/path$` may stop working

aviramha:
> One remark - this is not backwards compatible. Filters like `^/my/exact/path$` may stop working

Right, I'll make it an or then (try on path and then on path+query)

mirrord/agent/src/steal/http/filter.rs
```diff
@@ -58,15 +58,33 @@ impl HttpFilter {
...
+                .into_parts()
+                .path_and_query
+                .as_ref()
+                .map(|path_and_query| {
+                    // For backward compatability, we first match path then we match path and query
+                    // together and return true if any of them matches
+                    if filter
+                        .is_match(path_and_query.path())
+                        .inspect_err(|error| {
+                            tracing::error!(path, ?error, "Error while matching path");
```
Razz4780, line 72:
```suggestion
                    let path = path_and_query.path();
                    if filter
                        .is_match(path)
                        .inspect_err(|error| {
                            tracing::error!(path, ?error, "Error while matching path");
```
Compilation issue

Razz4780 (approved):

Issue #2515, opened 2024-06-12
Add `Unknown` variant for `Target`


Razz4780:
We're adding more target types to the operator and we want older CLIs to still be able to list targets. We basically need to deserialize `TargetCrd`s with forward compatibility in mind

PR #2553, opened 2024-06-26
Adds an `Unknown` variant to the list of possible mirrord targets for future compat.

meowjesty:
- Issue: #2515 

1. Changes the `TargetSpec` from having an `Option<Target>` field to an `KubeTarget` enum, which can take either a `Known(Target)` or an `Unknown(String)`:
    - custom deserialization, allowing us to convert a valid kubernetes target, which we don't know about in mirrord, into an `KubeTarget::Unknown`;


Razz4780:
Have you tried changing `TargetCrd` instead? Maybe we can wrap `Target` in there into some enum :V
Sth like:

```rust
#[derive(Deserialize, ...)
#[serde(untagged)]
pub enum SafeTarget {
  KnownTarget(Target),
  UnknownTarget {
    #[serde(rename = "type")]
    type_: String,
  },
}
```

The necessary use of `unreachable`s here is uncanny and I'm sad :<
Also `Unknown` being displayed as an empty string and other code depending on it :<

meowjesty:
Changed it to an enum, please look into the [operator PR](https://github.com/metalbear-co/operator/pull/569) to see if it's better now.

Name suggestions are welcome, I was thinking `KubeTarget`, as this type encapsulates any kube target, known and unknown to us.

Razz4780:
> Changed it to an enum, please look into the [operator PR](https://github.com/metalbear-co/operator/pull/569) to see if it's better now.
> 
> Name suggestions are welcome, I was thinking `KubeTarget`, as this type encapsulates any kube target, known and unknown to us.

Looks better imho ^^

mirrord/config/src/target.rs
```diff
@@ -252,6 +252,7 @@ pub enum Target {
 
     /// <!--${internal}-->
     /// Spawn a new pod.
+    #[default]
```
Razz4780, line 255: This looks kind of error prone tbh, I wouldn't add `Default` here

Razz4780 (changes requested):
I think this change could touch only `mirrord_operator`. Would be nice, because unknown targets are only relevant for maintaining forward compatibility with the operator. New wrapper type could be moved to `mirrord_operator::crd` and the rest of OSS code could stay unaware of its existence. You wouldn't need to change `mirrord_config` or `mirrord_kube`

mirrord/config/src/target.rs
meowjesty: Removed

mirrord/operator/src/client.rs
```diff
@@ -165,9 +165,14 @@ impl OperatorSessionTarget {
...
         match (use_proxy, self) {
-            (true, OperatorSessionTarget::Raw(target)) => {
-                let name = target.name();
-                let namespace = target
+            (true, OperatorSessionTarget::Raw(crd)) => {
+                let name = TargetCrd::urlfied_name(
+                    crd.spec
+                        .target
+                        .known()
+                        .expect("[BUG] Unknown target should never be seen here!"),
```
Razz4780, line 173: This could be a new variant of `OperatorApiError`. I see `expect` - I panic

Razz4780, line 173: This should be a new variant of `OperatorApiError`. I see `expect` - I panic

mirrord/operator/src/client.rs
```diff
@@ -178,9 +183,14 @@ impl OperatorSessionTarget {
...
 
-            (false, OperatorSessionTarget::Raw(target)) => {
-                let name = target.name();
-                let namespace = target
+            (false, OperatorSessionTarget::Raw(crd)) => {
+                let name = TargetCrd::urlfied_name(
+                    crd.spec
+                        .target
+                        .known()
+                        .expect("[BUG] Unknown target should never be seen here!"),
```
Razz4780, line 191: Same as above, should be an error

mirrord/kube/src/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum KubeApiError {
     #[error("None runtime data for non-targetless agent. This is a bug.")]
     MissingRuntimeData,
 
+    #[error("Failed trying to target unknown `{0}` resource!")]
+    UnknownTarget(String),
+
```
Razz4780, line 34: Do we need this here?

mirrord/operator/src/crd.rs
```diff
@@ -51,22 +58,20 @@ impl TargetCrd {
...
 
 impl From<TargetCrd> for TargetConfig {
     fn from(crd: TargetCrd) -> Self {
         TargetConfig {
-            path: crd.spec.target,
+            path: Some(
+                crd.spec
+                    .target
+                    .known()
+                    .expect("[BUG] Unknown target should never be seen here!")
```
Razz4780, line 72: Should be an error, so `TryFrom` instead of `From`

Razz4780 (changes requested):
Some nits + could you add a unit tests asserting that this change is backwards compatible? In test submodule, you could declare a struct like `TargetSpecLegacy` (with all the derives) that matches previous format and check how:
1. Serialized legacy type with target `None` deserializes into new type
2. Serialized legacy type with some target deserializes into new type
3. New type deserializes from some unknown target type
4. Legacy type deserializes from the serialized new type

mirrord/kube/src/error.rs
meowjesty: Removed.

mirrord/operator/src/crd.rs
meowjesty: Changed. Implemented a couple of `TryFrom`s.

Razz4780 (approved):
Neat!

PR #2556, opened 2024-06-27
Fs path re-mapping

### LINKED ISSUE #2068: Path mapping
Users would like to be able to specify a mapping `m` so that when an application tries to read file `A` mirrord switches the path so that the app then reads the path it is mapped to, `m(A)`, locally.

This is for cases when a user wants to provide a local file for the app to read but the path can't be written to locally (e.g. `/run/secrets` on macOS).
Currently the user has to change the path in their code in order to read the local file instead of the remote one.

In the specific case motivating this issue reading the file locally is a workaround for a [file read being too slow](https://github.com/metalbear-co/mirrord/issues/2067).

DmitryDodzin:
closes #2068


changelog.d/2068.added.md
```diff
@@ -0,0 +1,9 @@
+Add fs mapping, under `feature.fs.mapping` now it's possible to specify regex match and replace for paths while running mirrord exec.
+
+Example:
+
+```toml
+[feature.fs.mapping]
+"/var/app/temp" = "/tmp" # Will replace all calls to read/write/scan for "/var/app/temp/sample.txt" to "/tmp/sample.txt"
+"/var/app/.cache" = "/workspace/mirrord$0" # Will replace "/var/app/.cache/sample.txt" to "/workspace/mirrord/var/app/.cache/sample.txt" see [Regex::replace](https://docs.rs/regex/latest/regex/struct.Regex.html#method.replace)
```
aviramha, line 8: I'd like users to have the ability to group different parts of the path
for example
`"/(var)/(app)/.cache" = "/$0/bbb/$1"`
is it possible?

mirrord/layer/src/file/mapper.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use regex::{Regex, RegexSet, RegexSetBuilder};
+
+#[derive(Debug)]
+pub struct FileRemapper {
+    filter: RegexSet,
+    mapping: Vec<(Regex, String)>,
+}
+
+impl FileRemapper {
```
aviramha, line 11: Add UT

mirrord/layer/src/file/ops.rs
```diff
@@ -159,13 +181,9 @@ fn close_remote_file_on_failure(fd: u64) -> Result<()> {
 /// [`OPEN_FILES`].
 #[mirrord_layer_macro::instrument(level = "trace", ret)]
 pub(crate) fn open(path: Detour<PathBuf>, open_options: OpenOptionsInternal) -> Detour<RawFd> {
-    let path = path?;
-
-    if path.is_relative() {
-        // Calls with non absolute paths are sent to libc::open.
-        Detour::Bypass(Bypass::RelativePath(path.clone()))?
-    };
+    let path = remap_path!(path?);
```
aviramha, line 184: remap should happen after relative

aviramha (changes requested):
Great progress. See inline comments.

changelog.d/2068.added.md
DmitryDodzin: yep, capture groups work even named ones like "/(?<base>var)/(?<sub>app)/.cache"  = "/.cache/$base-$sub"

mirrord/layer/src/file/ops.rs
DmitryDodzin: But won't we want the ability to remap relative paths to absolute ones?

aviramha: Not at first step at least (haven't seen use case)

aviramha (approved):
Great work :)

aviramha (changes requested):
Somehow breaks SIP - need to investgiate
dylib fails loading on Tomcat IntelliJ integration

aviramha:
False flag on SIP - can merge post another review by team @Razz4780 @DmitryDodzin

aviramha (approved):

mirrord/config/src/feature/fs/advanced.rs
```diff
@@ -106,6 +108,11 @@ pub struct FsConfig {
     ///
     /// Specify file path patterns that if matched will be treated as non-existent.
     pub not_found: Option<VecOrSingle<String>>,
+
+    /// ### feature.fs.mapping {#feature-fs-mapping}
+    ///
+    /// Specify map of patterns that if matched will replace the path according to specification
```
Razz4780, line 114: This doc could use some examples

mirrord/layer/src/file/mapper.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    #[tracing::instrument(level = "trace", skip(self), ret)]
+    pub fn change_path_str<'p>(&self, path_str: &'p str) -> Cow<'p, str> {
+        let matches = self.filter.matches(path_str);
+
+        if let Some(index) = matches.iter().next() {
+            let (pattern, value) = self
+                .mapping
+                .get(index)
+                .expect("RegexSet matches returned an imposible index");
```
Razz4780, line 38:
```suggestion
                .expect("RegexSet matches returned an impossible index");
```

Razz4780 (approved):
Overall looks good, one doc suggestion ^^

PR #2561, opened 2024-07-01
Improve mirroring for high stress

### LINKED ISSUE #2529: Research sessions failing under high stress mirroring
There is sometimes a problem in running mirrord on high traffic/high throughput services where the required throughput of `mirrord-agent` -> `mirrord-int-proxy` is higher than available (via port-forward or operator connection) and the agent buffers start overflowing.

Possible solution is to allow http_filters (or some tcp filters like source_addr or something like that) to work somehow on mirroring or to have some sort of sampling of connections.

Razz4780:
Closes #2529 

1. `TcpConnectionSniffer` uses `broadcast` channels (capacity - 512 TCP packets) to send incoming data in a non-blocking way.
2. `TcpSnifferApi` notifies the client about closed connection (and sends an error log) when it's `broadcast::Receiver` is lagging.
3. `TcpConnectionSniffer` notifies `TcpSnifferApi`s about new connections via `mpsc` channels (capacity - 128 new sniffed connections). When a new connection is sniffed, the sniffer attempts to notify each client without blocking. If the client already has 128 queued new connections, client is not notified. There's some space for improvement here - instead of dropping new connection, we could be dropping the oldest one in the queue. However, `mpsc::Sender`'s api does not allow for it and we would need to implement some async circular FIFO
4. BPF filter used by the raw socket is now updated only when needed - `PortSubscribe`/`PortUnsubscribe` from a user actually changed the set of subscribed ports
5. Also some minor performance optimizations like using `hash_map::Entry` instead of `get() -> remove()`

### DIFF
--- a/mirrord/agent/src/sniffer.rs
+++ b/mirrord/agent/src/sniffer.rs
@@ -3,25 +3,16 @@ use std::{
     fmt,
     future::Future,
     hash::{Hash, Hasher},
-    io,
-    net::{Ipv4Addr, SocketAddr},
+    net::Ipv4Addr,
     pin::Pin,
     task::{Context, Poll},
 };
 
 use futures::{stream::FuturesUnordered, StreamExt};
 use mirrord_protocol::{MeshVendor, Port};
-use nix::sys::socket::SockaddrStorage;
-use pnet::packet::{
-    ethernet::{EtherTypes, EthernetPacket},
-    ip::IpNextHeaderProtocols,
-    ipv4::Ipv4Packet,
-    tcp::{TcpFlags, TcpPacket},
-    Packet,
-};
-use rawsocket::RawCapture;
+use pnet::packet::tcp::TcpFlags;
+use tcp_capture::TcpCapture;
 use tokio::{
-    net::UdpSocket,
     select,
     sync::{
         broadcast,
@@ -31,7 +22,10 @@ use tokio::{
 use tokio_util::sync::CancellationToken;
 use tracing::Level;
 
-use self::messages::{SniffedConnection, SnifferCommand, SnifferCommandInner};
+use self::{
+    messages::{SniffedConnection, SnifferCommand, SnifferCommandInner},
+    tcp_capture::RawSocketTcpCapture,
+};
 use crate::{
     error::AgentError,
     http::HttpVersion,
@@ -40,6 +34,7 @@ use crate::{
 
 pub(crate) mod api;
 pub(crate) mod messages;
+pub(crate) mod tcp_capture;
 
 /// [`Future`] that resolves to [`ClientId`] when the [`TcpConnectionSniffer`] client drops their
 /// [`TcpSnifferApi`](api::TcpSnifferApi).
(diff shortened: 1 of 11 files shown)


mirrord/agent/src/sniffer.rs
```diff
@@ -20,27 +23,56 @@ use rawsocket::RawCapture;
...
 use crate::{
     error::AgentError,
     http::HttpVersion,
-    util::{ClientId, IndexAllocator, Subscriptions},
-    watched_task::TaskStatus,
+    util::{ClientId, Subscriptions},
 };
 
+pub mod api;
+pub mod messages;
```
meowjesty, line 42: nit: with just `pub` we're not warned on unused, so if you need this everywhere in the agent change to `pub(crate)` please.

mirrord/agent/src/sniffer.rs
```diff
@@ -20,27 +23,56 @@ use rawsocket::RawCapture;
...
+        let future = std::pin::pin!(self.get_mut().client_tx.closed());
+        std::task::ready!(future.poll(cx));
+
+        Poll::Ready(client_id)
+    }
+}
+
 #[derive(Debug, Eq, Copy, Clone)]
-pub(crate) struct TcpSessionIdentifier {
+pub struct TcpSessionIdentifier {
```
meowjesty, line 68: more `pub` nits in this struct.

mirrord/agent/src/sniffer.rs
```diff
@@ -136,7 +162,7 @@ async fn resolve_interface() -> Result<Option<String>, AgentError> {
 //
 // And to make matters worse, the error reported back to the user is the very generic:
 // "mirrord-layer received an unexpected response from the agent pod!"
-#[tracing::instrument(level = "trace")]
+#[tracing::instrument(level = Level::DEBUG, err)]
```
meowjesty, line 165: Keeping it as `debug` intentionally?

mirrord/agent/src/sniffer.rs
```diff
@@ -150,7 +176,11 @@ async fn prepare_sniffer(
             .unwrap_or_else(|| "eth0".to_string()),
     };
 
-    trace!("Using {interface:#?} interface.");
+    tracing::debug!(
```
meowjesty, line 179: Same here, intentional `debug`?

mirrord/agent/src/sniffer.rs
```diff
@@ -203,147 +234,59 @@ fn get_tcp_packet(eth_packet: Vec<u8>) -> Option<(TcpSessionIdentifier, TcpPacke
...
-            command,
-        };
-
-        if self.sender.send(command).await.is_ok() {
-            Ok(())
-        } else {
-            Err(self.task_status.unwrap_err().await)
-        }
-    }
+pub(crate) struct TcpConnectionSniffer {
```
meowjesty, line 237: You touch it you add docs. That's the rules?

mirrord/agent/src/sniffer.rs
```diff
@@ -203,147 +234,59 @@ fn get_tcp_packet(eth_packet: Vec<u8>) -> Option<(TcpSessionIdentifier, TcpPacke
...
-    connection_id_to_tcp_identifier: HashMap<ConnectionId, TcpSessionIdentifier>,
-    index_allocator: IndexAllocator<ConnectionId, 100>,
-}
-
 impl TcpConnectionSniffer {
     pub const TASK_NAME: &'static str = "Sniffer";
 
     /// Runs the sniffer loop, capturing packets.
-    #[tracing::instrument(level = "trace", skip(self))]
+    #[tracing::instrument(level = Level::DEBUG, skip(cancel_token), err)]
```
meowjesty, line 262: `debug` intentional?

mirrord/agent/src/sniffer/messages.rs
```diff
@@ -0,0 +1,28 @@
+use mirrord_protocol::Port;
+use tokio::sync::{broadcast, mpsc::Sender, oneshot};
+
+use super::TcpSessionIdentifier;
+use crate::util::ClientId;
+
+#[derive(Debug)]
+pub enum SnifferCommandInner {
```
meowjesty, line 8: Less `pub`s?

mirrord/agent/src/sniffer/messages.rs
```diff
@@ -0,0 +1,28 @@
+use mirrord_protocol::Port;
```
meowjesty, line 1: You touch you docs (part 2)?

meowjesty (changes requested):
Getting rid of the nits and requests for docs first.

mirrord/agent/src/sniffer.rs
Razz4780: Yup, this way this trace's data will be included in some debug-level logs. I added some debug-level logs for stuff that should not generate lots of spam and kept spam on trace level

mirrord/agent/src/sniffer.rs
Razz4780: yup

mirrord/agent/src/sniffer.rs
Razz4780: yup

mirrord/agent/src/sniffer.rs
```diff
@@ -203,147 +234,81 @@ fn get_tcp_packet(eth_packet: Vec<u8>) -> Option<(TcpSessionIdentifier, TcpPacke
...
+/// incoming TCP data streams and sends copy of the traffic to all subscribed clients.
+///
+/// Can be easily used via [`api::TcpSnifferApi`].
+///
+/// # Notes on behavior under high load
+///
+/// Because this struct does not talk directly with the remote peers, we can't apply any back
+/// pressure on the incoming connections. There is no reliable mechanism to ensure that all
+/// subscribed clients receive all of the traffic. If we wait too long when distributing data
+/// between the clients, raw socket's recv buffer will overflow and we'll loose packets.
```
meowjesty, line 248:
```suggestion
/// between the clients, raw socket's recv buffer will overflow and we'll lose packets.
```

mirrord/agent/src/sniffer.rs
```diff
@@ -203,147 +234,81 @@ fn get_tcp_packet(eth_packet: Vec<u8>) -> Option<(TcpSessionIdentifier, TcpPacke
...
+///
+/// Having this in mind, this struct distributes incoming data using [`broadcast`] channels. If the
+/// clients are not fast enough to pick up TCP packets, they will loose them
```
meowjesty, line 251:
```suggestion
/// clients are not fast enough to pick up TCP packets, they will lose them
```

mirrord/agent/src/sniffer/messages.rs
```diff
@@ -0,0 +1,42 @@
...
+        Sender<SniffedConnection>,
+    ),
+    /// Client wants to start receiving connections incoming to a specific port.
+    Subscribe(
+        // Number of port to subscribe.
+        Port,
+        // Channel to notify with `()` when the subscription is done.
+        oneshot::Sender<()>,
+    ),
+    /// Client no longer wats to receive connections incoming to a specific port.
```
meowjesty, line 22:
```suggestion
    /// Client no longer wants to receive connections incoming to a specific port.
```

mirrord/agent/src/util.rs
```diff
@@ -31,51 +31,60 @@ where
     C: Eq + Hash + Clone + Copy,
 {
     /// Add a new subscription to a topic for a given client.
-    pub fn subscribe(&mut self, client: C, topic: T) {
-        self._inner.entry(topic).or_default().insert(client);
+    /// Returns whether this resulted in adding a new topic to this mapping.
+    pub fn subscribe(&mut self, client: C, topic: T) -> bool {
+        match self._inner.entry(topic) {
```
meowjesty, line 36: nit: I'm not sure why this is called `_inner`, can we rename it in the struct?

mirrord/agent/src/util.rs
```diff
@@ -31,51 +31,60 @@ where
...
+            Entry::Occupied(mut e) => {
+                e.get_mut().insert(client);
+                false
+            }
+            Entry::Vacant(e) => {
+                e.insert([client].into_iter().collect());
```
meowjesty, line 42:
```suggestion
                e.insert(From::from([client]));
```

Doesn't this work?

meowjesty (approved):
Just some more minor stuff, but it's worthy of approval.

Thanks for the docs, I'm happy now.

mirrord/agent/src/util.rs
Razz4780: Oh, it does. Cool!

Razz4780:
Something is not working and I don't yet know what ;/
Added some unit tests to `TcpConnectionSniffer`, but they all pass

meowjesty (approved):
:+1:

PR #2566, opened 2024-07-02
Send a copy of the SOCKETS to intproxy, and have it initialize a new layer with those SOCKETS.

meowjesty:
(wip)

### DIFF
--- a/mirrord/intproxy/protocol/src/lib.rs
+++ b/mirrord/intproxy/protocol/src/lib.rs
@@ -6,7 +6,9 @@ use std::{
     collections::HashMap,
     fmt,
     net::{IpAddr, SocketAddr},
+    num::ParseIntError,
     os::fd::RawFd,
+    str::FromStr,
 };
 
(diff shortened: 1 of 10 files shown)


Razz4780 (changes requested):
I think you got it a bit wrong :V

Forking is not a problem, because the child process inherits both descriptor and the memory (`SOCKETS` map). So layer actually works fine with forks. What you need to hook is the `exec` functions family. When the user app does exec, a new process is spawned with fresh memory and the layer initializes from scratch (from lib constructor). The problem is that this new process does inherit some of the descriptors (those without the `CLOEXEC` flag).

I imagine a flow like this:
1. In `exec*` detours, set parent layer id in env.
2. In layer's constructor, read parent layer id from env and clear it. If it was set, it means that the current process came from `exec`. At this point you should fetch sockets info from the intproxy, maybe with a new variant of `NewSessionRequest`

Razz4780:
Superseded by #2610

PR #2570, opened 2024-07-03
Always run discovery when fetching operator fails

### LINKED ISSUE #1730: mirrord operator status bad error message
When operator isn't installed, this is the output we get from `mirrord operator status`

```
x Operator Status
  x unable to get status
Error:   × Failed to connect to the operator. We have found the operator and unable to connect to it. KubeApiError(
  │     KubeError(
  │         Api(
  │             ErrorResponse {
  │                 status: "404 Not Found",
  │                 message: "\"404 page not found\\n\"",
  │                 reason: "Failed to parse error data",
  │                 code: 404,
  │             },
  │         ),
  │     ),
  │ )
```
This is probably because the code is shared with part that asserts operator existence.

Razz4780:
Closes #1730 

Running API discovery always when we fail to fetch operator resource (also if user explicitly enabled operator or the command requires it). Provides a better error message to the user.
Now it looks like this:
image

### DIFF
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -240,7 +240,7 @@ pub(crate) enum CliError {
     #[error("mirrord operator was not found in the cluster.")]
     #[diagnostic(help(
         "Command requires the mirrord operator or operator usage was explicitly enabled in the configuration file.
-        Install the operator using the `mirrord operator setup` command or the Helm chart at https://github.com/metalbear-co/charts/tree/main/mirrord-operator.{GENERAL_HELP}"
+        Read more here: https://mirrord.dev/docs/overview/quick-start/#operator.{GENERAL_HELP}"
     ))]
     OperatorNotInstalled,
 }
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -239,7 +239,7 @@ pub(crate) enum CliError {
 
     #[error("mirrord operator was not found in the cluster.")]
     #[diagnostic(help(
-        "Used command requires a mirrord operator or operator usage was explicitly enabled in the configuration file.
+        "Command requires the mirrord operator or operator usage was explicitly enabled in the configuration file.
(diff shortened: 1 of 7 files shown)


mirrord/cli/src/error.rs
```diff
@@ -236,6 +236,13 @@ pub(crate) enum CliError {
     #[error("Failed to prepare mirrord operator client certificate: {0}")]
     #[diagnostic(help("{GENERAL_BUG}"))]
     OperatorClientCertError(String),
+
+    #[error("mirrord operator was not found in the cluster.")]
+    #[diagnostic(help(
+        "Used command requires a mirrord operator or operator usage was explicitly enabled in the configuration file.
```
meowjesty, line 242:
```suggestion
        "Command requires the mirrord operator or operator usage was explicitly enabled in the configuration file.
```

Something like this?

mirrord/cli/src/error.rs
```diff
@@ -236,6 +236,13 @@ pub(crate) enum CliError {
...
+        Install the operator using the `mirrord operator setup` command or the Helm chart at https://github.com/metalbear-co/charts/tree/main/mirrord-operator.{GENERAL_HELP}"
```
meowjesty, line 243:
```suggestion
        You can install the operator with `mirrord operator setup` or by using the Helm chart at https://github.com/metalbear-co/charts/tree/main/mirrord-operator.{GENERAL_HELP}"
```

Maybe? Should we also mention you can get an operator trial free + no credit card?

meowjesty (approved):
Only 2 small message suggestions, nothing worth blocking the PR, so approved.

mirrord/cli/src/error.rs
Razz4780: Hmm, not sure. @aviramha ?
Context - this error is shown when we need the operator and we not for sure that it is not installed

Razz4780: @eyalb181 maybe?  : D

eyalb181: This needs to address two cases: 1. user registered for mfT but hasn't installed the Operator, 2. user hasn't even registered for mfT.
I think it should just be "Read more here" with this link: https://mirrord.dev/docs/overview/quick-start/#operator

PR #2580, opened 2024-07-09
Fix mirrord-operator compilation with only `crd` feature enabled

Razz4780:
I found myself fixing the same thing again, so I added an extra CI check for this


gememma (approved):
Nice!

PR #2583, opened 2024-07-10
Don't include non-running pods in node capacity check

### LINKED ISSUE #2582: Node capacity check counts non-running pods
The node capacity check we have also counts non-running pods (completed jobs for example).
```rs
        let mut pod_count = 0;
        let mut list_params = ListParams {
            field_selector: Some(format!("spec.nodeName={}", self.node_name)),
            ..Default::default()
        };
```
probably need this `status.phase=Running`

aviramha:
Fixes #2582

### DIFF
--- a/changelog.d/2582.fixed.md
+++ b/changelog.d/2582.fixed.md
@@ -1,0 +1,1 @@
+Don't include non-running pods in node capacity check
--- a/mirrord/kube/src/api/runtime.rs
+++ b/mirrord/kube/src/api/runtime.rs
@@ -184,7 +184,10 @@ impl RuntimeData {
 
         let mut pod_count = 0;
         let mut list_params = ListParams {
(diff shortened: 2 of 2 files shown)


Razz4780:
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
Maybe we should rather filter out succeeded and failed pods, wdyt?

aviramha:
> [kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) Maybe we should rather filter out succeeded and failed pods, wdyt?

It's a good question - I think this would only help with Pending (since not sure what our behavior should be with Unknown) and pending might be a pod not starting and we might be still be able to run

Razz4780 (approved):

PR #2588, opened 2024-07-11
Don't filter outgoing traffic to target

### LINKED ISSUE #2572: Outgoing connections to same service are not stolen
### Bug Description

1. Application steals from port 8081
2. Makes a request to its own k8s service
3. Agent makes outgoing connection
4. The connection should be stolen by the same agent, but it's not 

### Steps to Reproduce

---

### Backtrace

_No response_

### Relevant Logs

_No response_

---

### Local process

---

### Local process version

_No response_

### Additional Info

_No response_

DmitryDodzin:
Updated so if requests comes from target address don't fliter them out in our OUTPUT chain

closes #2572

### DIFF
--- a/mirrord/agent/src/cli.rs
+++ b/mirrord/agent/src/cli.rs
@@ -41,9 +41,6 @@ pub struct Args {
     /// If not given, the agent will not use TLS.
     #[arg(long, env = AGENT_OPERATOR_CERT_ENV)]
     pub operator_tls_cert_pem: Option<String>,
-
-    #[arg(long)]
-    pub pod_ips: Option<String>,
 }
 
 #[derive(Clone, Debug, Default, Subcommand)]
--- a/mirrord/agent/src/cli.rs
+++ b/mirrord/agent/src/cli.rs
@@ -41,6 +41,9 @@ pub struct Args {
     /// If not given, the agent will not use TLS.
     #[arg(long, env = AGENT_OPERATOR_CERT_ENV)]
     pub operator_tls_cert_pem: Option<String>,
+
+    #[arg(long)]
+    pub pod_ips: Option<String>,
 }
 
 #[derive(Clone, Debug, Default, Subcommand)]
--- a/mirrord/agent/src/steal/ip_tables/output.rs
+++ b/mirrord/agent/src/steal/ip_tables/output.rs
@@ -20,12 +20,18 @@ where
 {
     const ENTRYPOINT: &'static str = "OUTPUT";
 
-    pub fn create(ipt: Arc<IPT>, chain_name: String) -> Result<Self> {
+    pub fn create(ipt: Arc<IPT>, chain_name: String, pod_ips: Option<&str>) -> Result<Self> {
         let managed = IPTableChain::create(ipt, chain_name)?;
+
+        let exclude_source_ips = pod_ips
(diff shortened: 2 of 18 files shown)


mirrord/agent/src/steal/ip_tables/output.rs
```diff
@@ -23,18 +21,13 @@ where
     pub fn create(ipt: Arc<IPT>, chain_name: String) -> Result<Self> {
         let managed = IPTableChain::create(ipt, chain_name)?;
 
-        let gid = getgid();
-        managed
-            .add_rule(&format!("-m owner --gid-owner {gid} -p tcp -j RETURN"))
-            .inspect_err(|_| {
-                warn!("Unable to create iptable rule with \"--gid-owner {gid}\" filter")
-            })?;
+        managed.add_rule("! -o lo -j RETURN")?;
```
Razz4780, line 24: This change combined with changes to `add_redirect`/`remove_redirect` are a noop, right?

Razz4780 (commented):
How can it possibly work without filtering by gid? :eyes:

mirrord/agent/src/steal/ip_tables/output.rs
DmitryDodzin: Yes this should result in same result

mirrord/agent/src/steal/ip_tables/output.rs
```diff
@@ -20,12 +20,18 @@ where
 {
     const ENTRYPOINT: &'static str = "OUTPUT";
 
-    pub fn create(ipt: Arc<IPT>, chain_name: String) -> Result<Self> {
+    pub fn create(ipt: Arc<IPT>, chain_name: String, pod_ips: Option<&str>) -> Result<Self> {
         let managed = IPTableChain::create(ipt, chain_name)?;
 
+        let exclude_source_ips = pod_ips
+            .map(|pod_ips| format!("! -s {pod_ips}"))
+            .unwrap_or_default();
```
Razz4780, line 28: Wait, so the complete rule is now like "ignore TCP packets sent from agent's GID, unless they're sent from pod's IPs"? How does it help? Isn't this part a noop? :eyes:

DmitryDodzin: Not quite because the outgoing handler runs in the target process namespace so the request do originate from the agent (checking what happens on ephemerals)

aviramha (changes requested):

mirrord/agent/src/cli.rs
```diff
@@ -41,6 +41,9 @@ pub struct Args {
     /// If not given, the agent will not use TLS.
     #[arg(long, env = AGENT_OPERATOR_CERT_ENV)]
     pub operator_tls_cert_pem: Option<String>,
+
+    #[arg(long)]
+    pub pod_ips: Option<String>,
```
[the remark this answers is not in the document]
DmitryDodzin: updated, should be fine now

mirrord/agent/src/steal/ip_tables/output.rs
DmitryDodzin: ephemeral agent works just fine

aviramha (approved):
Approving changes requested by me - still need additional approval.

Razz4780 (approved):
LGTM ^^

PR #2592, opened 2024-07-12
Fix empty request streaming hanging forever

### LINKED ISSUE #2590: HTTP Request filtering hangs until timeout since 3.107.0
macOS
HTTP2
Node
Started in https://github.com/metalbear-co/mirrord/tree/3.107.0
Probably related to the streaming feature

aviramha:
Closes #2590

### DIFF
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -27,6 +27,7 @@ use tokio::{
     sync::mpsc::{Receiver, Sender},
 };
 use tokio_util::sync::CancellationToken;
+use tracing::warn;
 
 use crate::{
     error::{AgentError, Result},
@@ -169,6 +170,7 @@ impl Client {
                     },
                     mut body,
                 ) = request.request.into_parts();
+                tracing::trace!(?connection_id, "starting request");
                 match body.next_frames(true).await {
                     Err(..) => return,
                     Ok(Frames { frames, is_last }) => {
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/steal/connection.rs
```diff
@@ -190,9 +192,25 @@ impl Client {
...
+                                    request_id,
+                                },
+                            ));
+                            if let Err(e) = tx.send(message).await {
+                                warn!(?e, ?connection_id, ?request_id, ?request.port, "failed to send chunked request empty body");
+                                return;
+                            }
+                        }
                     }
                 }
```
Razz4780, line 215: I think you can just remove this entirely and don't use `is_last` from first `next_frames()` call at all. Same thing will happen in the first iteration of the loop below

mirrord/agent/src/steal/connection.rs
```diff
@@ -169,6 +170,7 @@ impl Client {
                     },
                     mut body,
                 ) = request.request.into_parts();
+                tracing::trace!(?connection_id, "starting request");
```
Razz4780, line 173: Maybe move this to the top of this `tokio::task`? And also log `framed` and `chunked` variables

Razz4780 (changes requested):

mirrord/agent/src/steal/connection.rs
aviramha: Smart

Razz4780 (approved):

PR #2603, opened 2024-07-15
Add test for empty body streamed request in agent

### LINKED ISSUE #2593: Write a test that covers empty request with http filtering
To avoid a regression like https://github.com/metalbear-co/mirrord/issues/2590
A UT should be enough I assume

gememma:
Closes #2593


Razz4780 (approved):
Noice

Issue #864, opened 2022-12-28
TCP Steal/mirror doesn't work with shared sockets


aviramha:
### Bug Description

When a parent process binds a socket, then children sockets listen/accept on that socket mirrord doesn't intercept those calls because the fd isn't managed (doesn't exist in our fork).
Using this issue to document use cases where this happens for now:

- uvicorn when using `--reload` flag.
- hypercorn
		

### Steps to Reproduce

```python
from fastapi import FastAPI
import sys
app = FastAPI()


@app.get("/")
async def root():
    print("cake")
    return {"message": "Hello World"}


@app.get("/hello/{name}")
async def say_hello(name: str):
    print("pake")
    return {"message": f"Hello {name}"}

```

run with `mirrord exec --target pod/pod uvicorn --  --reload --port 80 main:app`

### Your operating system and version

macOS

### Local process

python

aviramha:
Possible solutions:

1. hooking fork/exec, then passing down the `SOCKET/FILES` fd struct as environment variable for fds that aren't CLOEXEC
2. Use shared memory (shmem) for `SOCKET/FILES` (dunno about this..)
3. on `listen` with missing fd (this happens on uvicorn + reloader case) use getsockname to resolve addr and then decide if to add the socket and backfill the information. This is pretty good approach but has some caveats:
  - It would work only on the listen case, not if listen is called in parent and accept in child (not sure if this happens and what happens in this case)
  - the socket information we would get from getsockname would be the "fake port" and not the one the remote would need to listen for, so maybe this is impossible ;(

aviramha:
Please let us know if this is impacting anyone so we can know if we need to prioritize it.

danielloader:
This would probably affect me if not already blocked by this [issue](https://github.com/metalbear-co/mirrord/issues/1618)

yoav-orca:
I'm not blocked by this, but it does affect my productivity.
MacOS M1 Pro

aviramha:
> I'm not blocked by this, but it does affect my productivity.
> 
> MacOS M1 Pro

Hey, does it still happen to you on latest version?

yoav-orca:
I'm running `3.75.0` and I'm not getting traffic with `--reload` and I do get traffic without it

aviramha:
Also doesn't work when using Uvicorn with more than one worker process.

surbas:
Happening to me on 3.84.1. I am not using the operator. Not sure if that matters.

[...]

PR #2610, opened 2024-07-22
Hook `execve` and `execv` to allow shared sockets.

meowjesty:
- Issue: #864 

We now share the `SOCKETS` created by the parent process with its children (yay happy family) through an env var `MIRRORD_SHARED_SOCKETS` (the creativity is astonishing) that is set up in `exec*` calls (and also `posix_spawn` for macos). The `SOCKETS` are rebuilt from this env var (if it exists) in the child process.

- Adds a new hook `execv` for linux;
  - Sets an env var with Rust's `set_var`; 
- Makes the previously macos only `execve` hook also available on linux;
  - Changes `execve` behaviour to insert a new env var before calling `FN_EXECVE`;
- Adds a new integration test for an `uvicorn` app with the `--reload` flag, which triggers the issue;


mirrord/layer/src/setup.rs
```diff
@@ -126,15 +126,6 @@ impl LayerSetup {
             .unwrap_or(true)
     }
 
-    #[cfg(target_os = "macos")]
-    pub fn sip_binaries(&self) -> Vec<String> {
-        self.config
-            .sip_binaries
```
aviramha, line 132: Why?

mirrord/layer/src/exec_hooks/hooks.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    if let Detour::Success(modified_envp) = execve(checked_envp) {
+        #[cfg(target_os = "macos")]
+        match patch_sip_for_new_process(path, argv, modified_envp) {
+            Detour::Success((new_path, new_argv, new_envp)) => FN_EXECVE(
+                new_path.into_raw().cast_const(),
+                new_argv.leak(),
+                new_envp.leak(),
+            ),
+            _ => FN_EXECVE(path, argv, envp),
```
DmitryDodzin, line 116:
```suggestion
            _ => FN_EXECVE(path, argv, modified_envp),
```

Should It be the modified one in this branch?

mirrord/layer/src/exec_hooks/hooks.rs
```diff
@@ -0,0 +1,132 @@
...
+/// by the child process.
+#[mirrord_layer_macro::instrument(
+    level = Level::DEBUG,
+    ret,
+    fields(
+        pid = std::process::id(),
+        parent_pid = parent_id(),
+    )
+)]
+pub(crate) fn execve(env_vars: Detour<Argv>) -> Detour<*const *const c_char> {
```
DmitryDodzin, line 43: I'd rename this to something like `prepare_execve_envp`

mirrord/layer/src/exec_utils.rs
```diff
@@ -279,21 +205,22 @@ pub(crate) unsafe extern "C" fn posix_spawn_detour(
...
+                new_path.into_raw().cast_const(),
                 file_actions,
                 attrp,
-                new_argv.as_ptr() as *const *const c_char,
-                new_envp.as_ptr() as *const *const c_char,
-            )
+                new_argv.leak(),
+                new_envp.leak(),
+            ),
+            _ => FN_POSIX_SPAWN(pid, path, file_actions, attrp, argv, envp),
```
DmitryDodzin, line 220:
```suggestion
            _ => FN_POSIX_SPAWN(pid, path, file_actions, attrp, argv, modified_envp),
```
Same here no?

mirrord/layer/src/exec_hooks/hooks.rs
```diff
@@ -0,0 +1,132 @@
...
+    path: *const c_char,
+    argv: *const *const c_char,
+    envp: *const *const c_char,
+) -> c_int {
+    // Hopefully `envp` is a properly null-terminated list.
+    let checked_envp = envp.checked_into();
+
+    if let Detour::Success(modified_envp) = execve(checked_envp) {
+        #[cfg(target_os = "macos")]
+        match patch_sip_for_new_process(path, argv, modified_envp) {
```
DmitryDodzin, line 110: Shouldn't we patch sip regardless of we managed to change the envp? (like I see this shouldn't happen mostly but it may still happen and then we won't patch the sip)

mirrord/layer/src/exec_utils.rs
```diff
@@ -279,21 +205,22 @@ pub(crate) unsafe extern "C" fn posix_spawn_detour(
...
 ) -> c_int {
-    match patch_sip_for_new_process(path, argv, envp) {
-        Success((new_path, new_argv, new_envp)) => {
-            let new_argv = new_argv.null_vec();
-            let new_envp = new_envp.null_vec();
+    let checked_envp = envp.checked_into();
 
-            FN_POSIX_SPAWN(
+    if let Detour::Success(modified_envp) = hooks::execve(checked_envp) {
+        match patch_sip_for_new_process(path, argv, modified_envp) {
```
DmitryDodzin, line 211: Same here for sip on if the `Detour` isn't `Success`, I think we should patch sip regardless of if we managed to change the envp

DmitryDodzin (changes requested):
Minor changes regarding sip but looks good otherwise =]

mirrord/layer/src/exec_hooks/hooks.rs
meowjesty: Changed it to use the same name everywhere, maybe shadowing is better here so we can't have this mismatched issues.

mirrord/layer/src/exec_utils.rs
meowjesty: :ninja:

mirrord/layer/src/exec_hooks/hooks.rs
meowjesty: Changed to `patch_sip` first, and separated the detours for macos and non-macos. wdyt?

mirrord/layer/src/exec_utils.rs
meowjesty: Calling `patch_sip` first now, and if the `execve` preparation fails we still use the patched stuff.

DmitryDodzin (approved):
🥷

PR #2613, opened 2024-07-25
Warn when automatically picking a container out of many

t4lz:
* When there are multiple containers on the target, and the user did not specify a target container, show a warning with 
 the chosen container name.
 Target has multiple containers, mirrord picked
 Currently shows only when not connected to an operator. Adding that warning when running with the operator is a bit more complicated because we don't have the target resource (pod/deployment/etc.) at the client side. So we would need to send back a signal that a warning should be displayed. I think we should leave that to a separate PR as this PR is just a quick detour to pick low hanging fruit.

* Also, added Knative's "queue-proxy" container to the skipped list.

### DIFF
--- a/mirrord/kube/src/api/container.rs
+++ b/mirrord/kube/src/api/container.rs
@@ -132,38 +132,28 @@ pub fn choose_container<'a>(
         }
     });
 
-    // How many containers does mirrord have to choose from:
-    // 0: no containers, or container was specified.
-    // 1: only one, no choice.
-    // 2: 2 or more choices (non-mesh, or only non-mesh, but multiple of them).
-    let mut possible_containers = 0u8;
+    let mut picked_from_many = false;
 
     let container = if let Some(name) = container_name {
         container_statuses
             .iter()
             .find(|&status| status.name == name)
     } else {
-        // Choose any container that isn't part of the skip list
-        let container_refs = container_statuses
+        let mut container_refs = container_statuses
             .iter()
-            .take_while(|&status| !SKIP_NAMES.contains(status.name.as_str()))
-            .inspect(|_| possible_containers += 1)
-            // Iterate up to 2 even though we're taking the first container, so that we know if
-            // there are other possible choices for target container (for the purposes of showing a
-            // warning).
-            .take(2)
-            // Collect so that counting happens before we only take the first.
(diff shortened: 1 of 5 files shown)


mirrord/kube/src/api/runtime.rs
```diff
@@ -126,6 +129,8 @@ impl RuntimeData {
             .and_then(|status| status.container_statuses.as_ref())
             .ok_or_else(|| KubeApiError::missing_field(pod, ".status.containerStatuses"))?;
 
+        let guessed_container = container_name.is_none() && container_statuses.len() > 1;
```
Razz4780, line 132: I'd consider issuing a warning only if we have multiple containers that are not on the default skip list. Otherwise we'd be warning pretty much always, if the cluster uses proxies :thinking:

t4lz: Fixed. Let me know if you think the iterator code is now too ugly and loops would be better.

mirrord/kube/src/api/container.rs
```diff
@@ -131,17 +132,38 @@ pub fn choose_container<'a>(
...
             .iter()
             .find(|&status| status.name == name)
     } else {
         // Choose any container that isn't part of the skip list
-        container_statuses
+        let container_refs = container_statuses
             .iter()
-            .find(|&status| !SKIP_NAMES.contains(status.name.as_str()))
-            .or_else(|| container_statuses.first())
+            .take_while(|&status| !SKIP_NAMES.contains(status.name.as_str()))
```
Razz4780, line 149:
```suggestion
            .filter(|&status| !SKIP_NAMES.contains(status.name.as_str()))
```

Razz4780 (changes requested):
The loop is indeed quite weird :sweat_smile: 
I'd do sth like:
```rust
let mut picked_from_many = false;
// ...
let mut container_refs = container_statuses
    .iter()
    .filter(...);
let picked = container_refs.next().or_else(...);
picked_from_many = container_refs.next().is_some();

```

mirrord/kube/src/api/container.rs
t4lz: Thanks, changed to that.

Razz4780 (approved):

PR #2619, opened 2024-07-30
Revert "Hook `execve` and `execv` to allow shared sockets. (#2610)"

Razz4780:
Reverts broken #2610


Razz4780:
The bug was fixed with #2620

PR #2620, opened 2024-07-30
Fix exec hooks

Razz4780:


aviramha (approved):

meowjesty (approved):
It looks like it should work!

PR #2627, opened 2024-08-01
Istio Ambient Patch

### LINKED ISSUE #2456: Istio Ambient Mesh Compatibility
Mirrord works great with Istio's traditional sidecar deployment mode, and even gives a helpful message about `--steal` when it detects service meshes.

However, I'm not sure if it's a bug or not implemented that mirrord doesn't work with Istio's ambient mode. Is this on the roadmap? The agent pod runs fine but ATM I need to remove a `istio.io/dataplane-mode: ambient` label from the namespace for mirrord to successfully steal traffic.

DmitryDodzin:
closes #2456

Add support to istio with `istio.io/dataplane-mode: ambient`

*Note update does require privileged agent to work.*
```json
{
  "agnet": {
    "privileged": true
  }
}
```

### DIFF
--- a/changelog.d/2456.fixed.md
+++ b/changelog.d/2456.fixed.md
@@ -1,0 +1,1 @@
+Add custom handling for istio ambient mode where we set `/proc/sys/net/ipv4/conf/all/route_localnet` to `1` so it does require `agent.privilated = true` to work. (See [#2456](https://github.com/metalbear-co/mirrord/issues/2456))
--- a/changelog.d/2456.fixed.md
+++ b/changelog.d/2456.fixed.md
@@ -1,1 +1,1 @@
-Add custom handling for istio ambient mode where we set `/proc/sys/net/ipv4/conf/all/route_localnet` to `1` so it does require `agent.privilated = true` to work. (See [#2456](https://github.com/metalbear-co/mirrord/issues/2456))
+Add custom handling for istio ambient mode where we set `/proc/sys/net/ipv4/conf/all/route_localnet` to `1` so it does require `agent.privileged = true` to work. (See [#2456](https://github.com/metalbear-co/mirrord/issues/2456))
--- a/mirrord/agent/src/steal/ip_tables/mesh/istio.rs
+++ b/mirrord/agent/src/steal/ip_tables/mesh/istio.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 
 pub(crate) struct AmbientRedirect<IPT: IPTables> {
-    prerouteing: PreroutingRedirect<IPT>,
+    prerouting: PreroutingRedirect<IPT>,
     output: OutputRedirect<true, IPT>,
 }
 
@@ -21,23 +21,17 @@ where
     IPT: IPTables,
 {
     pub fn create(ipt: Arc<IPT>, pod_ips: Option<&str>) -> Result<Self> {
-        let prerouteing = PreroutingRedirect::create(ipt.clone())?;
+        let prerouting = PreroutingRedirect::create(ipt.clone())?;
         let output = OutputRedirect::create(ipt, IPTABLE_MESH.to_string(), pod_ips)?;
 
-        Ok(AmbientRedirect {
-            prerouteing,
-            output,
-        })
+        Ok(AmbientRedirect { prerouting, output })
     }
 
     pub fn load(ipt: Arc<IPT>) -> Result<Self> {
-        let prerouteing = PreroutingRedirect::load(ipt.clone())?;
+        let prerouting = PreroutingRedirect::load(ipt.clone())?;
         let output = OutputRedirect::load(ipt, IPTABLE_MESH.to_string())?;
 
-        Ok(AmbientRedirect {
-            prerouteing,
-            output,
-        })
+        Ok(AmbientRedirect { prerouting, output })
     }
 }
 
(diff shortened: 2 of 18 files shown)


aviramha:
Can we warn user if we detect ambient mesh and privileged is false?

changelog.d/2456.fixed.md
```diff
@@ -0,0 +1 @@
+Add custom handling for istio ambient mode where we set `/proc/sys/net/ipv4/conf/all/route_localnet` to `1` so it does require `agent.privilated = true` to work. (See [#2456](https://github.com/metalbear-co/mirrord/issues/2456))
```
aviramha, line 1: Typo in privileged

DmitryDodzin:
> Can we warn user if we detect ambient mesh and privileged is false?

It's a bit convoluted currently because we do the final detection in the iptables-guard agent process before any connection is even made but a "hacky" way with env variables is possible

aviramha:
> > Can we warn user if we detect ambient mesh and privileged is false?
> 
> It's a bit convoluted currently because we do the final detection in the iptables-guard agent process before any connection is even made but a "hacky" way with env variables is possible

isn't `MeshVendor` detected as part of runtime, which is created in the cli?

mirrord/agent/src/steal/ip_tables/mesh/istio.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::{
+    error::Result,
+    steal::ip_tables::{
+        output::OutputRedirect, prerouting::PreroutingRedirect, redirect::Redirect, IPTables,
+        IPTABLE_IPV4_ROUTE_LOCALNET_ORIGINAL, IPTABLE_MESH,
+    },
+};
+
+pub(crate) struct AmbientRedirect<IPT: IPTables> {
+    prerouteing: PreroutingRedirect<IPT>,
```
Razz4780, line 15:
```suggestion
    prerouting: PreroutingRedirect<IPT>,
```

Razz4780 (approved):
Looks good, one typo is field name

DmitryDodzin:
> isn't `MeshVendor` detected as part of runtime, which is created in the cli?

Well it's partially true, we do pass the runtime via cli args but also perform detection inside safe iptables

aviramha:
> > isn't `MeshVendor` detected as part of runtime, which is created in the cli?
> 
> Well it's partially true, we do pass the runtime via cli args but also perform detection inside safe iptables

So let's warn in that case.

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -227,18 +228,31 @@ impl KubernetesAPI {
...
+            progress.info(&format!("service mesh detected: {mesh}"));
+
+            let incoming_mode = config.map(|config| config.feature.network.incoming.mode);
+
+            if matches!(incoming_mode, Some(IncomingMode::Mirror)) {
+                progress.warning(
+                    "mirrord has detected that you might be running on a cluster with a \
+                     service mesh and `network.incoming.mode = \"mirror\"`, which is currently \
+                     unsupported. You can set `network.incoming.mode` to \"steal\" (check out the\
+                     `http_filter` configuration value if you only want to steal some of the traffic).",
```
Razz4780, line 241: Huh, wait. Is this still true? :eyes: 
We even have custom logic in the sniffer when mesh is detected

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -227,18 +228,31 @@ impl KubernetesAPI {
...
+
+                let privileged = config
+                    .map(|config| config.agent.privileged)
+                    .unwrap_or_default();
+
+                if matches!(mesh, MeshVendor::IstioAmbient) && !privileged {
+                    progress.warning(
+                        "mirrord has detected that there is an ambient istio service mesh but\
+                         agent will not be created with a privileged SecurtyContext,\
+                         please set `agent.privileged = true` otherwise agent will not be able to start.",
```
Razz4780, line 252:
```suggestion
                        "mirrord detected an ambient Istio service mesh but\
                         the agent is not configured to run in a privileged SecurityContext.\
                         Please set `agent.privileged = true`, otherwise the agent will not be able to start.",
```

mirrord/kube/src/api/container.rs
```diff
@@ -104,23 +104,26 @@ where
...
+    if pod
+        .metadata
+        .annotations
+        .as_ref()
+        .and_then(|annotations| annotations.get("ambient.istio.io/redirection"))
+        .map(|annotation| annotation == "enabled")
+        .unwrap_or_default()
+    {
+        return Some(MeshVendor::IstioAmbient);
+    }
```
Razz4780, line 122: Now that we're actually creating `IstioAmbient` variant and passing it to agent CLI, there's a problem - what if the agent is older and doesn't recognize this variant?

Razz4780 (changes requested):

mirrord/kube/src/api/container.rs
DmitryDodzin: I though about it though I'm not sure how to solve it elegantly (also older version of agent won't work properly on mesh proxy anyway no?)

mirrord/kube/src/api/kubernetes.rs
DmitryDodzin: Yep it works now, I removed it

mirrord/kube/src/api/container.rs
DmitryDodzin: Ok I did a compatibility change

PR #2630, opened 2024-08-05
medschool: validate json snippets in config files

### LINKED ISSUE #2525: Validate JSON snippets in `configuration.md` (or directly in the rust docs)
We generate the documentation for mirrord's configuration from the doc comments in the rust code of the configuration, using `medschool`.

In this documentation, we usually include examples in the form of JSON.

Currently there is nothing preventing us from including invalid JSON file snippets in this documentation. Those invalid snippets then make it into `configuration.md` via `medschool`, and at some point into https://github.com/metalbear-co/mirrord.dev.

It would be nice to have a validation mechanism that would prevent us from merging invalid JSON snippets into `main`.

We could also consider validating the one full mirrord configuration at the top of the documentation against our JSON schema, to make sure it is always valid and up to date, even when we make changes to our configuration schema. But this should probably be a separate PR at a later time.

infiniteregrets:
closes #2525

### DIFF
--- a/medschool/Cargo.toml
+++ b/medschool/Cargo.toml
@@ -26,6 +26,8 @@ clap = { version = "4", features = ["derive"] }
 tracing.workspace = true
 tracing-subscriber.workspace = true
 regex.workspace = true
+jsonschema = "0.18.0"
+serde_json = "1.0.122"
 
 [dev-dependencies]
 rand = "0.8"
--- a/medschool/src/main.rs
+++ b/medschool/src/main.rs
@@ -17,7 +17,9 @@
 
 use std::{fs, fs::File, io::Read, path::PathBuf};
 
+use jsonschema::JSONSchema;
 use parse::parse_docs_into_set;
+use serde_json::Value;
 use tracing_subscriber::{fmt::format::FmtSpan, prelude::*};
 
 use crate::{error::DocsError, parse::resolve_references};
@@ -25,6 +27,7 @@ use crate::{error::DocsError, parse::resolve_references};
 mod error;
 mod parse;
 mod types;
+mod validate;
 
 // TODO(alex): Support specifying a path.
 /// Converts all files in the [`glob::glob`] pattern defined within, in the current directory,
@@ -64,6 +67,9 @@ struct MedschoolArgs {
     /// Defaults to `./configuration.md`.
     #[arg(short, long)]
     output: Option<PathBuf>,
+
+    #[arg(short, long)]
+    schema: Option<PathBuf>,
 }
 
 /// # Attention when using `RUST_LOG`
@@ -89,11 +95,24 @@ fn main() -> Result<(), DocsError> {
         prepend,
         output,
         input,
+        schema,
     } = <MedschoolArgs as clap::Parser>::parse();
 
     let files = parse_files(input.unwrap_or_else(|| PathBuf::from("./src")))?;
     let type_docs = parse_docs_into_set(files)?;
-    let resolved = resolve_references(type_docs.clone());
+
+    let schema = if let Some(schema) = schema {
+        let schema = fs::read_to_string(schema)?;
+        let schema: Value =
+            serde_json::from_str(&schema).map_err(|e| DocsError::Json(e, schema))?;
+        let compiled_schema =
+            JSONSchema::compile(&schema).map_err(|e| DocsError::JsonSchema(e.to_string()))?;
+        Some(compiled_schema)
+    } else {
+        None
+    };
+
+    let resolved = resolve_references(type_docs.clone(), schema)?;
 
     if let Some(produced) = resolved {
         let mut final_docs = produced.produce_docs();
@@ -258,7 +277,9 @@ mod test {
         let type_docs = super::parse_docs_into_set(files).unwrap();
         println!("parsed {type_docs:#?}");
 
-        let root_type = resolve_references(type_docs.clone()).unwrap();
+        let root_type = resolve_references(type_docs.clone(), None)
+            .unwrap()
+            .unwrap();
 
         let final_docs = root_type.produce_docs();
         println!("final_docs {final_docs:#?}");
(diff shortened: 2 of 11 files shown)


medschool/Cargo.toml
```diff
@@ -26,6 +26,8 @@ clap = { version = "4", features = ["derive"] }
 tracing.workspace = true
 tracing-subscriber.workspace = true
 regex.workspace = true
+jsonschema = "0.18.0"
+serde_json = "1.0.122"
```
meowjesty, line 30:
```suggestion
jsonschema = "0.18"
serde_json = "1.0"
```

We usually just use major.min everywhere else right?

medschool/src/main.rs
```diff
@@ -89,11 +95,24 @@ fn main() -> Result<(), DocsError> {
...
+    let schema = if let Some(schema) = schema {
+        let schema = fs::read_to_string(schema)?;
+        let schema: Value =
+            serde_json::from_str(&schema).map_err(|e| DocsError::Json(e, schema))?;
+        let compiled_schema =
+            JSONSchema::compile(&schema).map_err(|e| DocsError::JsonSchema(e.to_string()))?;
+        Some(compiled_schema)
+    } else {
+        None
+    };
```
meowjesty, line 113: `if let Some {} else None` is just `.map`.

medschool/src/parse.rs
```diff
@@ -200,43 +203,53 @@ pub fn parse_docs_into_set<'a>(
 /// DFS helper function to resolve the references of the types. Returns the docs of the fields of
 /// the field we're currently looking at search till its leaf nodes. The leaf here means a primitive
 /// type for which we don't have a [`PartialType`].
+#[tracing::instrument(level = "trace", ret)]
```
meowjesty, line 206:
```suggestion
#[tracing::instrument(level = Level::TRACE, ret)]
```

For some reason we started doing it like this now.

medschool/src/parse.rs
```diff
@@ -279,48 +292,59 @@ fn dfs_fields<'a, const MAX_RECURSION_LEVEL: usize>(
 /// be where we resolve all references and return the same HashSet and let the caller
 /// decide what the root should be.
 #[tracing::instrument(level = "trace", ret)]
```
meowjesty, line 294:
```suggestion
#[tracing::instrument(level = Level::TRACE, ret)]
```

Also change it here, since you're touching this function.

medschool/src/parse.rs
```diff
@@ -279,48 +292,59 @@ fn dfs_fields<'a, const MAX_RECURSION_LEVEL: usize>(
...
-pub fn resolve_references(types: HashSet<PartialType>) -> Option<PartialType> {
-    /// Maximum recursion level for safety.
+pub fn resolve_references(
+    types: HashSet<PartialType>,
+    schema: Option<JSONSchema>,
+) -> Result<Option<PartialType>, DocsError> {
+    // Maximum recursion level for safety.
```
meowjesty, line 299:
```suggestion
    /// Maximum recursion level for safety.
```

medschool/src/parse.rs
```diff
@@ -200,43 +203,53 @@ pub fn parse_docs_into_set<'a>(
...
     if *recursion_level >= MAX_RECURSION_LEVEL {
-        return vec!["Recursion limit reached".to_string()];
+        return Ok(vec!["Recursion limit reached".to_string()]);
+    }
+    if let Some(schema) = schema {
+        validate_json_blocks(&field.docs, schema)
+            .inspect_err(|e| error!(error = ?e, context=?field, "Error validating JSON blocks"))?;
+    }
+    // get the type of the field from the types set to recurse into it's fields
+    match types.get(&field.ty) {
```
meowjesty, line 222: I don't think you need to change this so much just due to the function returning a `Result` now?

The logic inside this `match` looks very similar to the iterator (at a glance at least).

medschool/src/parse.rs
```diff
@@ -279,48 +292,59 @@ fn dfs_fields<'a, const MAX_RECURSION_LEVEL: usize>(
...
+) -> Result<Option<PartialType>, DocsError> {
+    // Maximum recursion level for safety.
     const MAX_RECURSION_LEVEL: usize = 10;
     // Cache to perform memoization between recursive calls so we don't have to resolve the same
     // type multiple times. Mapping between `ident` -> `resolved_docs`.
     // For example, if we have a types [`A`, `B`, `C`] and A has a field of type `B` and `B` has a
     // field of type `C`, and `C` has already been resolved, we don't want to resolve `C` again
     // as we iterate over the types. A -> (B -> C), (B -> C), (C)
     let mut cache = HashMap::with_capacity(types.len());
+    let mut max_area = 0;
```
meowjesty, line 307: Same thing here, the logic looks pretty much the same, except it's more `Result`-y now? You could use `try_collect` for example.

meowjesty (changes requested):
Thanks for the PR my bhai.

It's missing docs, overall looks fine, but I wonder why the changes to remove iterator stuff?

infiniteregrets:
> Thanks for the PR my bhai.
> 
> It's missing docs, overall looks fine, but I wonder why the changes to remove iterator stuff?

of course bhai, the whole closure/functional thing made the code a mess to be honest, its much easier and clear to read it with a for loop

infiniteregrets:
> Aw, I thought it would be just a matter of changing some `collect` to `try_collect`, and maybe some options to results before going into `and_then`. I asked for this because the diff felt like a bit too much, when the change is really just `validate_json_blocks(&type_.docs, schema)`.

hi, coming back to this, id like to keep the loop there just to make it more readable! will address the comments! sorry for the delay

meowjesty:
> id like to keep the loop

Alright! I'm not a `loop` person, but it was more of a suggestion due to how the bigger the diff got than what I expected.

> sorry for the delay

No worries bhai! After your previous fix the tool is already working way better.

Razz4780:
Hey @infiniteregrets, is there any work in progress on this one?

Razz4780:
Seems like the PR is dead :skull_and_crossbones:

Issue #2611, opened 2024-07-23
Buffer directory iteration


aviramha:
When user iterates a directory, we fetch each entry on its own, causing it to be very slow when reading big directory.
Similar in nature to https://github.com/metalbear-co/mirrord/issues/2069

Razz4780:
Note: this will probably require extending `mirrord-protocol`, so operator will have to updated as well

Razz4780:
Closed with #2637
Operator counterpart [here](https://github.com/metalbear-co/operator/issues/618)

PR #2637, opened 2024-08-07
Send readdir as a batched request, where the response comes with a list of dirs, instead of just one.

meowjesty:
- Issue: #2611 

- Adds a new pair of messages to the protocol: `ReadDirBatchRequest` and `ReadDirBatchResponse`;
  - `ReadDirBatchRequest` is sent from the intproxy to the agent when the intproxy receives a `ReadDirRequest` (the layer doesn't deal with the _batch_ stuff). It gets an iterator of `InternalDirEntry` from the agent (where previously this iterator was kept in the agent only);
  - `ReadDirBatchResponse` sends the iterator to the intproxy, and the length is set in the intproxy;


mirrord/agent/src/file.rs
```diff
@@ -727,6 +721,17 @@ impl FileManager {
         Ok(result)
     }
 
+    #[tracing::instrument(level = Level::DEBUG, skip(self), err)]
+    pub(crate) fn read_dir_batch(&mut self, fd: u64) -> RemoteResult<ReadDirBatchResponse> {
+        let result = self
+            .get_dir_stream(fd)?
+            .map(DirEntryInternal::try_from)
+            .try_collect::<Vec<_>>()
+            .map(|dir_entries| ReadDirBatchResponse { fd, dir_entries })?;
```
Razz4780, line 730: Do we want to read the *whole* directory? A directory on ext4 can have like 64k entries

aviramha: +1, I'd expect some buffering

meowjesty: Changed it! Now it's being done in steps. Should I add the `amount` we read as config somewhere?

Razz4780: Nah, I think we can enable configuration later if at all

mirrord/protocol/src/file.rs
```diff
@@ -385,11 +385,23 @@ pub struct ReadDirRequest {
...
+
 #[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
 pub struct ReadDirResponse {
     pub direntry: Option<DirEntryInternal>,
 }
 
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct ReadDirBatchResponse {
+    pub fd: u64,
+    pub dir_entries: Option<Vec<DirEntryInternal>>,
```
Razz4780, line 402: We could probably skip the `Option`? Empty `Vec` means no more entries anyway

tests/go-e2e-dir/main.go
```diff
@@ -1,3 +1,4 @@
+// TODO(alex) [high]: Change the integration to readdirbatchrequest
```
Razz4780, line 1: todoo

mirrord/layer/tests/apps/issue1123/issue1123.rs
```diff
@@ -1,3 +1,4 @@
+// TODO(alex) [high]: Remote resource out of sync, is this the `None`?
```
Razz4780, line 1: todoo

mirrord/layer/tests/apps/env_bash_cat.sh
```diff
@@ -14,4 +14,4 @@
 # cat is a SIPed binary on mac.
 cat /very_interesting_file
 # sleep so we get close request (else bash might exit before we have time to close it)
-sleep 1
\ No newline at end of file
+sleep 10
```
Razz4780, line 17: Why? :eyes:

mirrord/layer/src/lib.rs
```diff
@@ -483,7 +483,7 @@ fn sip_only_layer_start(mut config: LayerConfig, patch_binaries: Vec<String>) {
 /// - `enabled_remote_dns`: replaces [`libc::getaddrinfo`] and [`libc::freeaddrinfo`] when this is
 ///   `true`, see [`NetworkConfig`](mirrord_config::feature::network::NetworkConfig), and
 ///   [`hooks::enable_socket_hooks`](socket::hooks::enable_socket_hooks).
-#[mirrord_layer_macro::instrument(level = "trace")]
+// #[mirrord_layer_macro::instrument(level = Level::TRACE)]
```
Razz4780, line 486: todoo

mirrord/intproxy/protocol/src/lib.rs
```diff
@@ -372,6 +365,13 @@ impl_request!(
     res_path = ProxyToLayerMessage::File => FileResponse::ReadDir,
 );
 
+impl_request!(
+    req = ReadDirBatchRequest,
+    res = RemoteResult<ReadDirBatchResponse>,
+    req_path = LayerToProxyMessage::File => FileRequest::ReadDirBatch,
+    res_path = ProxyToLayerMessage::File => FileResponse::ReadDirBatch,
+);
```
Razz4780, line 373: This is required only for the code to compile, right? Layer should not send batch requests

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -55,8 +80,15 @@ impl BackgroundTask for SimpleProxy {
     type MessageOut = ProxyMessage;
 
     async fn run(mut self, message_bus: &mut MessageBus<Self>) -> Result<(), RequestQueueEmpty> {
+        let mut protocol_version = VERSION.clone();
```
Razz4780, line 83: Should be initialized to `None`. Agent version might be lower and might not allow for batched readdir

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -27,20 +32,40 @@ pub enum SimpleProxyMessage {
...
+#[derive(Clone)]
+pub(crate) enum FileResource {
+    File,
+    Dir {
+        dirs_iter: Option<IntoIter<DirEntryInternal>>,
+    },
+}
+
+impl FileResource {
+    fn next_dir(&mut self) -> Option<DirEntryInternal> {
```
Razz4780, line 53: This should return an error when called on a regular file

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -85,6 +117,45 @@ impl BackgroundTask for SimpleProxy {
...
+                                    ReadDirResponse {
+                                        direntry: Some(dir),
+                                    },
+                                ))),
+                                layer_id,
+                            })
+                            .await;
+                    } else {
+                        self.file_reqs.insert(message_id, layer_id);
+                        // let protocol_version = Version::new(0, 0, 0);
```
Razz4780, line 143: Delet :wastebasket:

Razz4780 (changes requested):

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -85,6 +117,45 @@ impl BackgroundTask for SimpleProxy {
...
+                        let request = if protocol_version >= Version::new(1, 9, 0) {
```
Razz4780, line 144: This minimal version should be a static defined in `mirrord_protocol` crate

Razz4780 (changes requested):
One more thing - could you add a UT to `SimpleProxy` to check the new messages? Should check readdir behavior when agent protocol version (does not) match version requirement.

mirrord/protocol/src/codec.rs
```diff
@@ -121,6 +115,7 @@ pub enum FileResponse {
     Xstat(RemoteResult<XstatResponse>),
     XstatFs(RemoteResult<XstatFsResponse>),
     ReadDir(RemoteResult<ReadDirResponse>),
+    ReadDirBatch(RemoteResult<ReadDirBatchResponse>),
```
Razz4780, line 118: This **must** be the last variant in the enum

mirrord/protocol/src/codec.rs
```diff
@@ -79,6 +72,7 @@ pub enum FileRequest {
     XstatFs(XstatFsRequest),
     FdOpenDir(FdOpenDirRequest),
     ReadDir(ReadDirRequest),
+    ReadDirBatch(ReadDirBatchRequest),
```
Razz4780, line 75: This **must** be the last variant of the enum

Razz4780 (changes requested):

mirrord/layer/tests/apps/env_bash_cat.sh
meowjesty: Was testing stuff.

mirrord/intproxy/src/proxies/simple.rs
meowjesty: Changed it to return proper errors when we hit an error in the cached readdir.

mirrord/intproxy/protocol/src/lib.rs
meowjesty: Should I break the pattern? Kept it like this to have it looking like all the other requests.

mirrord/protocol/src/file.rs
```diff
@@ -5,11 +5,18 @@ use std::fs::DirEntry;
...
+};
 
 use bincode::{Decode, Encode};
 #[cfg(target_os = "linux")]
 use nix::sys::statfs::Statfs;
+use semver::VersionReq;
+
+/// Minimal mirrord-protocol version that allows [`ReadDirBatchRequest`].
+pub static READDIR_BATCH_VERSION: LazyLock<VersionReq> =
+    LazyLock::new(|| ">=1.8.2".parse().expect("Bad Identifier"));
```
Razz4780, line 19: Gotta bump the crate version as well

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -27,20 +34,103 @@ pub enum SimpleProxyMessage {
...
+
+impl FileResource {
+    fn next_dir(&mut self, remote_fd: u64) -> Result<Option<DirEntryInternal>, FileError> {
+        match self {
+            FileResource::Dir { dirs_iter } => dirs_iter.next().map(Ok).transpose(),
+            FileResource::File => Err(FileError::DirOnFile(remote_fd)),
+        }
+    }
+
+    async fn handle_readdir(
```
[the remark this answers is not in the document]
meowjesty: Moved it there.

mirrord/protocol/src/file.rs
meowjesty: Should this be an `x.x.+` bump, or a `x.+.0`?

Razz4780: I think `x.+.0`, it's a new feature in the protocol

mirrord/protocol/src/file.rs
```diff
@@ -385,11 +392,31 @@ pub struct ReadDirRequest {
...
 pub struct ReadDirResponse {
     pub direntry: Option<DirEntryInternal>,
 }
 
+/// `readdir` response with the list of items (length depends on the [`ReadDirBatchRequest`]'s
+/// `amount`), and the `remote_fd` of the dir (for convenience).
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct ReadDirBatchResponse {
+    /// Remote fd of the dir.
+    pub fd: u64,
```
Razz4780, line 414: What do we need this for? Agent handles file requests sequentially, so file responses don't usually contain the descriptor. Just a question though, I don't mind if it stays here

Razz4780 (approved):
:cat2:

gememma (approved):
noice 👍

mirrord/protocol/src/file.rs
meowjesty: Was the easiest way I found to updating the iterator:

```rust
                    if let Some(FileResource::Dir { dirs_iter }) =
                        self.remote_fds.get_mut(&layer_id, &RemoteFd::Dir(fd))
                    {
                        *dirs_iter = entries_iter;
                    }

```

Razz4780: Ah, okok

PR #2643, opened 2024-08-08
Add Execution Kind

DmitryDodzin:


mirrord/analytics/src/lib.rs
```diff
@@ -29,6 +29,45 @@ pub enum AnalyticsError {
     Unknown,
 }
 
+#[derive(Default, Debug, Serialize, Deserialize, Clone, Copy)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecutionKind {
+    Container,
+    #[default]
```
aviramha, line 36: we don't send strings in the analytics on purpose (see target type) - set it to an int enum (0 for exec, 1 for container and so on)

DmitryDodzin: I can make it bitflags so we can do extension + exec or (extension + contianer in the future)

mirrord/analytics/src/lib.rs
```diff
@@ -155,30 +194,33 @@ pub struct AnalyticsReporter {
...
-    pub fn new(enabled: bool, watch: drain::Watch) -> Self {
+    pub fn new(enabled: bool, execution_kind: ExecutionKind, watch: drain::Watch) -> Self {
         AnalyticsReporter {
             analytics: Analytics::default(),
             error_only_send: false,
             enabled,
             error: None,
             operator_properties: None,
             start_instant: Instant::now(),
+            execution_kind,
```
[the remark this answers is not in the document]
DmitryDodzin: We can push it event properties in analytcis-server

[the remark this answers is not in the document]
aviramha: Why would we want to introduce another change instead of just pushing it here into event_properties?

mirrord/analytics/src/lib.rs
aviramha: I don't think we need, we can make "extension" a flag on it's own then specify which kind of extension.

gememma (approved):
real nice 👍

mirrord/analytics/src/lib.rs
DmitryDodzin: Ok did it simple

mirrord/analytics/src/lib.rs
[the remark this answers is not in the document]
DmitryDodzin: Ok updated so it will just create Analytics with it included

PR #2646, opened 2024-08-12
3.113.0

Razz4780:


CHANGELOG.md
```diff
@@ -8,6 +8,76 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
...
+
+  ```
+  mirrord container [options] -- <docker/podman> run ...
+  ```
+
+  Because we need to run internal proxy process on the same network as the
+  process loaded with `mirrord-layer`, to keep config and kubernetes
+  comparability the communication to mirrord agent is made via external proxy
+  that will run on the host machine.
+  ```
```
DmitryDodzin, line 26: I just have a dilemma, I like the graphic but. not sure should be in changelog

Razz4780: Nah, I think the graphic is cool. Why not?

DmitryDodzin: It's just Iv'e never seen one in any changelog

Razz4780: @eyalb181 wdyt?

eyalb181: Let's keep it in, it's attention grabbing

PR #2649, opened 2024-08-12
Add support for port forwarding

### LINKED ISSUE #567: Support pod port-forwarding
I would like to use mirrord to connect to managed RDS service.

I can use something like:
`mirrord exec --no-fs -s "" --target <deployment>   -- /opt/homebrew/opt/libpq/bin/psql -h rds-addr -U admin`
What if I want to use GUI applications such as pgAdmin?

I think something like
`mirrord proxy --target <depoyment> <remote_addr:remote_port> <local_addr:local_port>` could be useful for debugging

Related issues and projects:
https://github.com/kubernetes/kubernetes/issues/72597
https://github.com/knight42/krelay

gememma:
Closes #567

### DIFF
--- a/changelog.d/567.added.md
+++ b/changelog.d/567.added.md
@@ -1,0 +1,5 @@
+Add port forwarding feature which can be used to proxy data from a local port to a remote one -
+if the local port is not specified, it will default to the same as the remote
+```
+mirrord portforward [options] -L <local_port:destination_ip:remote_port/destination_ip:remote_port>
+```
--- a/changelog.d/567.added.md
+++ b/changelog.d/567.added.md
@@ -1,5 +1,5 @@
 Add port forwarding feature which can be used to proxy data from a local port to a remote one -
 if the local port is not specified, it will default to the same as the remote
 ```
-mirrord portforward [options] -L <local_port:destination_ip:remote_port/destination_ip:remote_port>
+mirrord portforward [options] -L [local_port:]remote_ip:remote_port
 ```
--- a/changelog.d/567.added.md
+++ b/changelog.d/567.added.md
@@ -1,5 +1,5 @@
 Add port forwarding feature which can be used to proxy data from a local port to a remote one -
 if the local port is not specified, it will default to the same as the remote
 ```
-mirrord portforward [options] -L [local_port:]remote_ip:remote_port
+mirrord port-forward [options] -L [local_port:]remote_ip:remote_port
 ```
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -67,7 +67,7 @@ pub(super) enum Commands {
     #[command(hide = true, name = "intproxy")]
     InternalProxy,
 
-    /// Port forwarding
+    /// Port forwarding - UNSTABLE FEATURE
     #[command(hide = true, name = "portforward")]
     PortForward(Box<PortForwardArgs>),
 
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -1,11 +1,19 @@
 #![deny(missing_docs)]
 
-use std::{collections::HashMap, ffi::OsString, fmt::Display, path::PathBuf};
+use std::{
+    collections::HashMap,
+    ffi::OsString,
+    fmt::Display,
+    net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
+    path::PathBuf,
+    str::FromStr,
+};
 
 use clap::{ArgGroup, Args, Parser, Subcommand, ValueEnum, ValueHint};
 use clap_complete::Shell;
 use mirrord_config::MIRRORD_CONFIG_FILE_ENV;
 use mirrord_operator::setup::OperatorNamespace;
+use thiserror::Error;
 
 use crate::error::CliError;
 
@@ -57,6 +65,10 @@ pub(super) enum Commands {
     /// Internal proxy - used to aggregate connections from multiple layers
     #[command(hide = true, name = "intproxy")]
     InternalProxy,
+
+    /// Port forwarding
+    #[command(hide = true, name = "portforward")]
+    PortForward(Box<ExecArgs>),
 
     /// Verify config file without starting mirrord.
     #[command(hide = true)]
@@ -301,6 +313,143 @@ pub(super) struct ExecArgs {
 
     /// Arguments to pass to the binary.
     pub(super) binary_args: Vec<String>,
+}
+
+#[derive(Args, Debug)]
+#[command(group(ArgGroup::new("port-forward")))]
+pub(super) struct PortForwardArgs {
+    /// Target name to mirror.    
+    /// Target can either be a deployment or a pod.
+    /// Valid formats: deployment/name, pod/name, pod/name/container/name
+    #[arg(short = 't', long)]
+    pub target: Option<String>,
+
+    /// Namespace of the pod to mirror. Defaults to "default".
+    #[arg(short = 'n', long)]
+    pub target_namespace: Option<String>,
+
+    /// Namespace to place agent in.
+    #[arg(short = 'a', long)]
+    pub agent_namespace: Option<String>,
+
+    /// Agent log level
+    #[arg(short = 'l', long)]
+    pub agent_log_level: Option<String>,
+
+    /// Agent image
+    #[arg(short = 'i', long)]
+    pub agent_image: Option<String>,
+
+    /// Agent TTL
+    #[arg(long)]
+    pub agent_ttl: Option<u16>,
+
+    /// Agent Startup Timeout seconds
+    #[arg(long)]
+    pub agent_startup_timeout: Option<u16>,
+
+    /// Accept/reject invalid certificates.
+    #[arg(short = 'c', long)]
+    pub accept_invalid_certificates: bool,
+
+    /// Use an Ephemeral Container to mirror traffic.
+    #[arg(short, long)]
+    pub ephemeral_container: bool,
+
+    /// Disable telemetry. See <https://github.com/metalbear-co/mirrord/blob/main/TELEMETRY.md>
+    #[arg(long)]
+    pub no_telemetry: bool,
+
+    #[arg(long)]
+    /// Disable version check on startup.
+    pub disable_version_check: bool,
+
+    /// Load config from config file
+    #[arg(short = 'f', long, value_hint = ValueHint::FilePath)]
+    pub config_file: Option<PathBuf>,
+
+    /// Kube context to use from Kubeconfig
+    #[arg(long)]
+    pub context: Option<String>,
+
+    /// Mappings
+    pub mappings: Vec<PortMapping>,
+}
+
+#[derive(Clone, Debug)]
+pub struct PortMapping {
+    local: SocketAddr,
+    remote: SocketAddr,
+}
+
+impl FromStr for PortMapping {
+    type Err = PortMappingParseErr;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // expected format = local_port:dest_server:remote_port
+        let vec: Vec<&str> = s.split(":").collect();
+        let (local_port, remote_ip, remote_port) = match vec.len() {
+            3 => {
+                // local port included
+                let local_port = match vec[0].parse::<u16>() {
+                    Ok(port) => port,
+                    Err(_error) => return Err(PortMappingParseErr::PortParseErr(vec[0].into())),
+                };
+                let remote_port = match vec[2].parse::<u16>() {
+                    Ok(0) => return Err(PortMappingParseErr::PortZeroInvalid(s.into())),
+                    Ok(port) => port,
+                    Err(_error) => return Err(PortMappingParseErr::PortParseErr(vec[2].into())),
+                };
+                let ip_parts: Vec<u8> = vec[1]
+                    .split(".")
+                    .filter_map(|s| s.parse::<u8>().ok())
+                    .collect();
+                let remote_ip = match ip_parts.len() {
+                    4 => Ipv4Addr::new(ip_parts[0], ip_parts[1], ip_parts[2], ip_parts[3]),
+                    _ => return Err(PortMappingParseErr::IpParseErr(vec[1].into())),
+                };
+                (local_port, remote_ip, remote_port)
+            }
+            2 => {
+                // local port excluded, default to same as remote port
+                let remote_port = match vec[1].parse::<u16>() {
+                    Ok(port) => port,
+                    Err(_error) => return Err(PortMappingParseErr::PortParseErr(vec[0].into())),
+                };
+                let ip_parts: Vec<u8> = vec[0]
+                    .split(".")
+                    .filter_map(|s| s.parse::<u8>().ok())
+                    .collect();
+                let remote_ip = match ip_parts.len() {
+                    4 => Ipv4Addr::new(ip_parts[0], ip_parts[1], ip_parts[2], ip_parts[3]),
+                    _ => return Err(PortMappingParseErr::IpParseErr(vec[1].into())),
+                };
+                (remote_port.clone(), remote_ip, remote_port)
+            }
+            num @ _ => {
+                return Err(PortMappingParseErr::NumSubArgs(num, s.into()));
+            }
+        };
+        Ok(Self {
+            local: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), local_port),
+            remote: SocketAddr::new(IpAddr::V4(remote_ip), remote_port),
+        })
+    }
+}
+
+#[derive(Error, Debug)]
+enum PortMappingParseErr {
+    #[error("Incorrect number of sub-arguments found for port forwarding: expected 2 or 3, found {0} in argument `{1}`")]
+    NumSubArgs(usize, String),
+
+    #[error("Failed to parse `{0}` into port (u16)")]
+    PortParseErr(String),
+
+    #[error("Failed to parse `{0}` into IP address (u8.u8.u8.u8)")]
+    IpParseErr(String),
+
+    #[error("Failed to set remote port for port forwarding while parsing `{0}`: `0` is never a valid remote port")]
+    PortZeroInvalid(String),
 }
 
 #[derive(Args, Debug)]
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -69,7 +69,7 @@ pub(super) enum Commands {
 
     /// Port forwarding
     #[command(hide = true, name = "portforward")]
-    PortForward(Box<ExecArgs>),
+    PortForward(Box<PortForwardArgs>),
 
     /// Verify config file without starting mirrord.
     #[command(hide = true)]
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -13,7 +13,6 @@ use clap::{ArgGroup, Args, Parser, Subcommand, ValueEnum, ValueHint};
 use clap_complete::Shell;
 use mirrord_config::MIRRORD_CONFIG_FILE_ENV;
 use mirrord_operator::setup::OperatorNamespace;
-use rstest::rstest;
 use thiserror::Error;
 
 use crate::error::CliError;
@@ -454,10 +453,10 @@ impl FromStr for PortMapping {
 
 #[derive(Error, Debug, PartialEq)]
 pub enum PortMappingParseErr {
-    #[error("Invalid format of argument `{0}`, expected `[local-port]:remote-ipv4:remote-port`")]
+    #[error("Invalid format of argument `{0}`, expected `[local-port:]remote-ipv4:remote-port`")]
     InvalidFormat(String),
 
-    #[error("Failed to parse port `{0}` in argument: `{1}`")]
+    #[error("Failed to parse port `{0}` in argument `{1}`")]
     PortParseErr(String, String),
 
     #[error("Failed to parse IPv4 address `{0}` in argument `{1}`")]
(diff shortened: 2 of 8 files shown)


changelog.d/567.added.md
```diff
@@ -0,0 +1,5 @@
+Add port forwarding feature which can be used to proxy data from a local port to a remote one -
+if the local port is not specified, it will default to the same as the remote
+```
+mirrord portforward [options] -L <local_port:destination_ip:remote_port/destination_ip:remote_port>
```
Razz4780, line 4:
```suggestion
mirrord portforward [options] -L [local_port:]remote_ip:remote_port
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+        tracing::trace!("Port forwarding setup complete");
+
+        loop {
+            select! {
+                _ = tokio::time::sleep_until(self.ping_pong_timeout.into()) => {
+                    if self.waiting_for_pong {
+                        // no pong received before timeout
+                        break Err(PortForwardError::AgentError("Agent failed to respond to Ping".into()));
+                    }
+                    let _ = self.agent_connection.sender.send(ClientMessage::Ping).await;
```
Razz4780, line 145: Error should not be ignored :V

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    None => {
+                        tracing::trace!("None message received, connection ended with agent");
+                        break Ok(())
+                    },
+                },
+
+                // stream coming from the user app
+                message = self.listeners.next() => match message {
+                    Some(message) => self.handle_listener_stream(message)?,
+                    None => break Ok(()),
```
Razz4780, line 161:
```suggestion
                    None => unreachable!("created listener sockets are never closed"),
```
Given that we check `parsed_mappings` non-emptiness in `PortForwarder::new`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                },
+
+                message = self.internal_msg_rx.recv() => match message {
+                    Some(message) => self.handle_msg_from_task(message).await?,
+                    None => {
+                        tracing::trace!("PortForwardMessage Sender disconnected while sending");
+                    },
+                },
```
Razz4780, line 169:
```suggestion
                message = self.internal_msg_rx.recv() => {
                    self.handle_msg_from_task(message.expect("this channel is never closed")).await?;
                },
```
I think we can assume that this channel is never closed, because the `Receiver` never "escapes" encapsulation in this struct

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+            }
+        }
+    }
+
+    async fn handle_msg_from_agent(
```
Razz4780, line 174:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip(self), err)]
    async fn handle_msg_from_agent(
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                return Err(PortForwardError::AgentError(format!(
+                    "Unexpected message from Agent: {other:?}"
+                )));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn handle_listener_stream(
```
Razz4780, line 268:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip(self), err)]
    fn handle_listener_stream(
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+            (socket, Err(error)) => {
+                // error from TcpStream
+                tracing::error!("Error occured while listening to local socket {socket}: {error}");
+                self.listeners.remove(&socket);
+            }
+        }
+        Ok(())
+    }
+
+    async fn handle_msg_from_task(
```
Razz4780, line 328:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip(self), err)]
    async fn handle_msg_from_task(
```

mirrord/cli/src/error.rs
```diff
@@ -327,6 +329,10 @@ pub(crate) enum CliError {
     #[error("mirrord returned a target resource of unknown type: {0}")]
     #[diagnostic(help("{GENERAL_BUG}"))]
     OperatorReturnedUnknownTargetType(String),
+
+    #[error("An error occurred in the port-forwarding process: {0}")]
+    #[diagnostic(help("{GENERAL_BUG}"))]
+    PortForwardingError(PortForwardError),
```
Razz4780, line 335:
```suggestion
    PortForwardingError(#[from] PortForwardError),
```
Instead of manually implementing `From<PortForwardError> for CliError` ^^

mirrord/cli/src/config.rs
```diff
@@ -303,6 +316,175 @@ pub(super) struct ExecArgs {
...
 }
 
+#[derive(Args, Debug)]
+#[command(group(ArgGroup::new("port-forward")))]
+pub(super) struct PortForwardArgs {
+    /// Target name to mirror.    
+    /// Target can either be a deployment or a pod.
+    /// Valid formats: deployment/name, pod/name, pod/name/container/name
+    #[arg(short = 't', long)]
+    pub target: Option<String>,
```
DmitryDodzin, line 326: I see that there a quite a few duplicate keys with `ExecConfig` maybe there is a case for doing 
```rust
#[derive(Args, Debug)]
pub(super) struct TargetParams {
    /// Target name to mirror.    
    /// Target can either be a deployment or a pod.
    /// Valid formats: deployment/name, pod/name, pod/name/container/name
    #[arg(short = 't', long)]
    pub target: Option<String>,

    /// Namespace of the pod to mirror. Defaults to "default".
    #[arg(short = 'n', long)]
    pub target_namespace: Option<String>,
}

impl TargetParams {
    pub fn as_env_vars(&self) -> Result<HashMap<String, OsString>, CliError> {
        let mut envs: HashMap<String, OsString> = HashMap::new();

        if let Some(target) = &self.target {
            envs.insert("MIRRORD_IMPERSONATED_TARGET".into(), target.into());
        }
        if let Some(namespace) = &self.target_namespace {
            envs.insert("MIRRORD_TARGET_NAMESPACE".into(), namespace.into());
        }
        
        envs
    }
}
```
And use it as 
```rust
#[clap(flatten)]
pub target: TargetParams,
```
Similar to how `ExecParams` is reused across `ExecArgs` and `ContainerArgs`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                };
+                self.tasks.remove(&socket);
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum PortForwardError {
```
Razz4780, line 388: Error messages here should probably start with a lowercase letter (this probably is not always the case across the code base, but nvm) to make the final user-facing message look better. Same for nested messages

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+    #[error("Failed to setup connection with Agent with error: `{0}`")]
+    AgentSetupError(String),
+
+    #[error("Error occurred in the Agent: `{0}`")]
+    AgentError(String),
```
Razz4780, line 393:
```suggestion
    #[error("agent closed connection with error: `{0}`")]
    AgentError(String),
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum PortForwardError {
+    #[error("Failed to setup connection with Agent with error: `{0}`")]
+    AgentSetupError(String),
```
Razz4780, line 390: I think you can add this instead:
```rust
pub enum PortForwardError {
    ...
    #[error("connection with the agent failed")]
    AgentConnectionFailed,
    ...
}

impl From<mpsc::error::SendError<ClientMessage>> for PortForwardError {
    fn from(_: mpsc::error::SendError<ClientMessage>) -> Self {
        Self::AgentConnectionFailed
    }
}
```
Use it whenever you fail to send/recv using `agent_connection`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                            ));
+                        };
+                        let Some(channel) = self.oneshots.pop_front() else {
+                            return Err(PortForwardError::ReadyTaskNotFound(
+                                remote_socket,
+                                res.connection_id,
+                            ));
+                        };
+                        self.sockets.insert(res.connection_id, remote_socket);
+                        let _ = channel.send(res.connection_id);
```
Razz4780, line 194: If this fails, close the remote connection with `LayerTcpOutgoing::Close`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                        tracing::trace!("Successful connection to remote address {remote_socket}, connection ID is {}", res.connection_id);
+                    }
+                    Err(error) => {
+                        return Err(PortForwardError::AgentError(format!(
+                            "Problem receiving DaemonTcpOutgoing::Connect {error}"
+                        )))
+                    }
```
Razz4780, line 201: This should not kill the whole port forwarder. This error can happen when the user specifies unreachable remote address. You can just `self.oneshots.pop_front()`, kill the local connection task and log error

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                },
+                DaemonTcpOutgoing::Read(res) => match res {
+                    Ok(res) => {
+                        let Some(&remote_socket) = self.sockets.get(&res.connection_id) else {
+                            return Err(PortForwardError::SocketFromIdNotFound(res.connection_id));
+                        };
```
Razz4780, line 207: This may happen due to inherent local<->agent lag. Just ignore agent messages for unknown connection ids

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                        let Some(sender) = self.tasks.get(&remote_socket) else {
+                            return Err(PortForwardError::SenderNotFound(
+                                remote_socket,
+                                res.connection_id,
+                            ));
+                        };
```
Razz4780, line 213: This can be an `unreachable!()`, the struct is encapsulated and you can ensure it never happens

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                        let Ok(_) = sender.send(res.bytes).await else {
+                            return Err(PortForwardError::InternalMessageError(
+                                "Failed to send data from remote to task".into(),
+                            ));
+                        };
```
Razz4780, line 218: This can also happen if the task fails, just close the connection in the agent and do necessary cleanup

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    Err(error) => {
+                        return Err(PortForwardError::AgentError(format!(
+                            "Problem receiving DaemonTcpOutgoing::Read {error}"
+                        )))
+                    }
+                },
+                DaemonTcpOutgoing::Close(connection_id) => {
+                    let Some(remote_socket) = self.sockets.remove(&connection_id) else {
+                        return Err(PortForwardError::SocketFromIdNotFound(connection_id));
+                    };
```
Razz4780, line 229: Again, it's ok to ignore

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    let Some(sender) = self.tasks.remove(&remote_socket) else {
+                        return Err(PortForwardError::SenderNotFound(
+                            remote_socket,
+                            connection_id,
+                        ));
+                    };
```
Razz4780, line 235: Again, can be `unreachable!()`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    drop(sender);
```
Razz4780, line 236: Seems unnecessary :eyes:

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    tracing::trace!(
+                        "Connection closed for remote socket {remote_socket}, connection {connection_id}"
+                    );
+                }
+            },
+            DaemonMessage::LogMessage(log_message) => match log_message.level {
+                LogLevel::Warn => tracing::warn!(log_message.message),
+                LogLevel::Error => tracing::error!(log_message.message),
```
Razz4780, line 244:
```suggestion
                LogLevel::Warn => tracing::warn!("Agent log: {}", log_message.message),
                LogLevel::Error => tracing::error!("Agent log: {}", log_message.message),
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+            },
+            DaemonMessage::Pong => {
+                if !self.waiting_for_pong {
+                    // message not expected
+                    return Err(PortForwardError::AgentError(
+                        "Unexpected message from Agent: DaemonMessage::Pong".into(),
+                    ));
```
Razz4780, line 251: I feel like I see a lot of `PortForwardError::AgentError`s with similar message. Maybe you can add `UnexpectedMessage(DaemonMessage)` variant to the error enum?

mirrord/cli/src/config.rs
gememma: ooh looks nice, will do 🫡

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+    fn handle_listener_stream(
+        &mut self,
+        message: (SocketAddr, Result<TcpStream, std::io::Error>),
+    ) -> Result<(), PortForwardError> {
+        match message {
+            (socket, Ok(stream)) => {
+                let task_internal_tx = self.internal_msg_tx.clone();
+                let Some(remote_socket) = self.mappings.get(&socket).cloned() else {
+                    return Err(PortForwardError::SocketMappingNotFound(socket));
+                };
```
Razz4780, line 277: Could be `unreachable!()`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                let (response_tx, mut response_rx) = mpsc::channel(16);
```
Razz4780, line 278:
```suggestion
                let (response_tx, mut response_rx) = mpsc::channel(256);
```
Ah, why not

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                self.tasks.insert(remote_socket, response_tx);
+                tokio::spawn(async move {
+                    let (read, mut write) = stream.into_split();
+                    let mut read_stream = ReaderStream::new(read);
+                    let (oneshot_tx, oneshot_rx) = oneshot::channel::<ConnectionId>();
+                    let _ = task_internal_tx
+                        .send(PortForwardMessage::Connect(remote_socket, oneshot_tx))
+                        .await;
```
Razz4780, line 286: Wait until you get first data from `read_stream`

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                    let Ok(connection_id) = oneshot_rx.await else {
+                        return Err(PortForwardError::InternalMessageError(format!("Error while trying to receive connection ID for socket address {socket}")));
+                    };
+
+                    loop {
+                        select! {
+                            message = read_stream.next() => match message {
+                                Some(Ok(message)) => {
+                                    let _ = task_internal_tx.send(PortForwardMessage::Send(connection_id, message.into())).await;
```
Razz4780, line 295: Should exit the task in case of failure

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                                },
+                                Some(Err(error)) => {
+                                    break Err(PortForwardError::TcpListenerError(error));
```
Razz4780, line 298: I think no one inspects this error. You can just log and exit task

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                                },
+                                None => {
+                                    let _ = task_internal_tx.send(PortForwardMessage::Close(remote_socket, connection_id)).await;
+                                    break Ok(());
+                                },
+                            },
+
+                            message = response_rx.recv() => match message {
+                                Some(message) => {
+                                    match write.write(message.as_ref()).await {
```
Razz4780, line 308:
```suggestion
                                    match write.write_all(message.as_ref()).await {
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+            (socket, Ok(stream)) => {
+                let task_internal_tx = self.internal_msg_tx.clone();
+                let Some(remote_socket) = self.mappings.get(&socket).cloned() else {
+                    return Err(PortForwardError::SocketMappingNotFound(socket));
+                };
+                let (response_tx, mut response_rx) = mpsc::channel(16);
+                self.tasks.insert(remote_socket, response_tx);
+                tokio::spawn(async move {
+                    let (read, mut write) = stream.into_split();
+                    let mut read_stream = ReaderStream::new(read);
```
Razz4780, line 282: With 64kb capacity pls

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                                    let _ = task_internal_tx.send(PortForwardMessage::Close(remote_socket, connection_id)).await;
+                                    break Ok(());
+                                },
+                            },
+
+                            message = response_rx.recv() => match message {
+                                Some(message) => {
+                                    match write.write(message.as_ref()).await {
+                                        Ok(_) => continue,
+                                        Err(error) => break Err(PortForwardError::InternalMessageError(format!("Failed to write data to local port in task: {error}"))),
```
Razz4780, line 310: I think no one inspects this error, you can just log and exit the task

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+                            }
+                        }
+                    }
+                });
+            }
+            (socket, Err(error)) => {
+                // error from TcpStream
+                tracing::error!("Error occured while listening to local socket {socket}: {error}");
+                self.listeners.remove(&socket);
+            }
```
Razz4780, line 323: If you started this whole function with
```rust
let stream = match message.1 {
    Ok(stream) => stream,
    Err(error) => {
        // log and remove listener
        return Ok(());
    }
};
```
it would flatten the code a bit :V

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,689 @@
...
+    ) -> Result<(), PortForwardError> {
+        match message {
+            (socket, Ok(stream)) => {
+                let task_internal_tx = self.internal_msg_tx.clone();
+                let Some(remote_socket) = self.mappings.get(&socket).cloned() else {
+                    return Err(PortForwardError::SocketMappingNotFound(socket));
+                };
+                let (response_tx, mut response_rx) = mpsc::channel(16);
+                self.tasks.insert(remote_socket, response_tx);
+                tokio::spawn(async move {
```
Razz4780, line 280: I feel like this task should be extracted to its own struct, like
```rust
struct LocalConnectionTask { ... }
impl LocalConnectionTask {
    fn new(...) -> Self { ... }
    fn run(mut self) { ... }
}
```

Razz4780 (changes requested):

mirrord/cli/src/main.rs
```diff
@@ -454,6 +457,87 @@ async fn print_targets(args: &ListTargetArgs) -> Result<()> {
...
+    (&config).collect_analytics(analytics.get_mut());
+
+    config.verify(&mut context)?;
+    for warning in context.get_warnings() {
+        progress.warning(warning);
+    }
+
+    let (_connection_info, connection) =
+        create_and_connect(&config, &mut progress, &mut analytics).await?;
+    let mut port_forward = PortForwarder::new(connection, args.port_mappings.clone()).await?;
```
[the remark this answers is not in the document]
gememma: args.port_mappings is behind a shared ref so it cant be moved :/

mirrord/cli/src/port_forward.rs
gememma: I think the restructuring so far has already eliminated this 👍

mirrord/cli/src/main.rs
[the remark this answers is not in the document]
Razz4780: Doesn't have to be tho, no? :eyes:

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+            remote_socket,
+            task_internal_tx,
+            response_rx,
+            read_stream,
+            write,
+            local_socket,
+        }
+    }
+
+    pub async fn run(&mut self) -> Result<(), PortForwardError> {
```
Razz4780, line 393: This function exits early in multiple places. Make sure that main port-forwarder loop always knows about finished tasks (to do necessary clean-up).

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+                            Ok(_) => (),
+                            Err(error) => {
+                                break Err(PortForwardError::InternalMessageError(format!(
+                                    "failed to send data to main loop: {error}"
+                                )))
+                            }
+                        };
+                    },
+                    Some(Err(error)) => {
+                        tracing::error!("tcp listener error: {error}");
```
Razz4780, line 454:
```suggestion
                        tracing::warn!(
                            %error,
                            local_socket = %self.local_socket,
                            remote_socket = %self.remote_socket,
                            "local connection failed",
                        );
```

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+                        break Ok(());
+                    },
+                },
+
+                message = self.response_rx.recv() => match message {
+                    Some(message) => {
+                        match self.write.write_all(message.as_ref()).await {
+                            Ok(_) => continue,
+                            Err(error) => {
+                                tracing::error!("failed to write remote response to local destination: {error}");
```
Razz4780, line 468:
```suggestion
                                tracing::error!(
                                    %error,
                                    local_socket = %local_socket,
                                    remote_socket = %remote_socket,
                                    "local connection failed",
                                );
```

Razz4780 (changes requested):
Mostly suggestions you can just apply ^^

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+    mappings: HashMap<SocketAddr, SocketAddr>,
+    /// accepts connections from the user app in the form of a stream
+    listeners: StreamMap<SocketAddr, TcpListenerStream>,
+    /// oneshot channels for sending connection IDs to tasks
+    oneshots: VecDeque<oneshot::Sender<ConnectionId>>,
+    /// identifies remote socket addresses by their corresponding connection ID
+    sockets: HashMap<ConnectionId, SocketAddr>,
+    /// identifies tasks by their corresponding remote socket address
+    /// for sending data from the remote socket to the local address
+    tasks: HashMap<SocketAddr, Sender<Vec<u8>>>,
```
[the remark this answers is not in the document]
gememma: The docs aren't wrong, this is janky and I forgot that the mappings arent 1:1. Honestly I think it's going to be better to change `tasks` to map an ID to the sender (as you originally suggested) and get around the fact that the tasks get spawned before a connection ID is established with the agent some other way, like having IDs internal to portforwarder and mapping between them and agen IDs in the main loop

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+    ping_pong_timeout: Instant,
+}
+
+/// Used by tasks for individual forwarding connections to send instructions to [`PortForwarder`]'s
+/// main loop.
+#[derive(Debug)]
+enum PortForwardMessage {
+    Connect(SocketAddr, oneshot::Sender<ConnectionId>),
+    Send(ConnectionId, Vec<u8>),
+    Close(SocketAddr, ConnectionId),
```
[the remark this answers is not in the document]
gememma: its needed for `self.tasks.remove` - the individual tasks cant remove items from fields in `PortForwarder` - but I want to change this behaviour anyway (see other comment on tasks logic)

mirrord/cli/src/port_forward.rs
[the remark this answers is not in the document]
Razz4780: Why not use `tcp_stream.local_addr()` as identifier? (`tcp_stream` that comes from accepting a connection on listener)

mirrord/cli/src/port_forward.rs
```diff
@@ -0,0 +1,779 @@
...
+                    }
+                },
+                DaemonTcpOutgoing::Close(connection_id) => {
+                    let Some(remote_socket) = self.sockets.remove(&connection_id) else {
+                        // ignore unknown connection IDs
+                        return Ok(());
+                    };
+                    let Some(_sender) = self.tasks.remove(&remote_socket) else {
+                        unreachable!("sender is always created before this point")
+                    };
```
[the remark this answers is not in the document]
gememma: tasks and sockets get cleaned, what else needs removing?

mirrord/cli/src/port_forward.rs
gememma: good point, i'm going to redesign run() slightly so that the loop returns a value which gets handled at the end of the function

mirrord/cli/src/port_forward.rs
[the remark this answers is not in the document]
Razz4780: Ah sorry, missed that

Razz4780 (approved):
image

PR #2657, opened 2024-08-14
Agent housekeeping

Razz4780:
Some minor improvements to agent code:
1. `TcpStealerApi` no longer need to hold a shared channel permit to close session with the background task
2. Removed last uses `IndexAllocator`
3. Added some more logs to outgoing task
4. Improved agent errors

### DIFF
--- a/mirrord/agent/src/outgoing.rs
+++ b/mirrord/agent/src/outgoing.rs
@@ -15,6 +15,7 @@ use tokio::{
 };
 use tokio_stream::StreamExt;
 use tokio_util::io::ReaderStream;
+use tracing::Level;
 
 use crate::{
     error::Result,
@@ -53,7 +54,7 @@ impl TcpOutgoingApi {
     /// # Params
     ///
     /// * `pid` - process id of the agent's target container
-    #[tracing::instrument(level = "trace")]
+    #[tracing::instrument(level = Level::TRACE)]
     pub(crate) fn new(pid: Option<u64>) -> Self {
         let (layer_tx, layer_rx) = mpsc::channel(1000);
         let (daemon_tx, daemon_rx) = mpsc::channel(1000);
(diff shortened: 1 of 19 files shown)


mirrord/agent/src/outgoing.rs
```diff
@@ -89,7 +90,7 @@ impl TcpOutgoingApi {
     }
 
     /// Receives a [`DaemonTcpOutgoing`] message from the background task.
-    pub(crate) async fn daemon_message(&mut self) -> Result<DaemonTcpOutgoing> {
+    pub(crate) async fn recv_from_task(&mut self) -> Result<DaemonTcpOutgoing> {
```
meowjesty, line 93:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip(self), err)]
    pub(crate) async fn recv_from_task(&mut self) -> Result<DaemonTcpOutgoing> {
```

Add tracing here?

mirrord/agent/src/outgoing.rs
```diff
@@ -148,45 +149,51 @@ impl TcpOutgoingTask {
     }
 
     /// Runs this task as long as the channels connecting it with [`TcpOutgoingApi`] are open.
+    /// This routine never fails and returns [`Result`] only due to [`WatchedTask`] constraints.
+    #[tracing::instrument(level = Level::TRACE, skip(self))]
```
meowjesty, line 153:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip(self), err)]
```

We're going to see the error elsewhere, but maybe add it in the tracing here just for consistency?

mirrord/agent/src/runtime/error.rs
```diff
@@ -0,0 +1,39 @@
...
+            error: error.to_string(),
+        }
+    }
+
+    pub(crate) fn unknown_runtime<N: ToString>(name: N) -> Self {
+        Self::UnknownRuntimeName(name.to_string())
+    }
+}
+
+pub(crate) type Result<T, E = ContainerRuntimeError> = core::result::Result<T, E>;
```
meowjesty, line 39: This pattern kinda always annoys me, when the alias is named just `Result`, and I want to check something out by writing a quick function that returns a `Result<(), SomeOtherError>`.

Could you rename it to `ContainerResult` (or something else you find appropriate)?

meowjesty (approved):
Approving, as I only have minor stuff to comment on.

mirrord/agent/src/outgoing.rs
Razz4780: There is no error, this task never fails :sweat_smile:

meowjesty: Yup, saw it after the comment ...

Issue #2599, opened 2024-07-15
Detect using "docker" as binary and alert user it won't do what they think it does


aviramha:

eyalb181:
Wait until #1658 is merged, because it affects what the message should be

PR #2663, opened 2024-08-14
Add warning when mirrord exec docker is detected, hinting the user towards mirrord container.

meowjesty:
- Issue: #2599 

- Adds a new warning when the user does `mirrord exec docker`;
  - Thought about making the message a bit longer, explaining what's going on (similar to the "multi-pod" warning), but it feels like too much text;


mirrord/progress/src/messages.rs
```diff
@@ -11,3 +11,9 @@ pub const MULTIPOD_WARNING: (&str, &str) = (
         You can get started with mirrord for Teams at this link: \
         https://mirrord.dev/docs/overview/teams/",
 );
+
+pub const EXEC_DOCKER_BINARY: (&str, &str) = (
```
aviramha, line 15: why the `(&str, &str)` format? I don't see `.0` being used anywhere

Razz4780:
Something weird happened in the `test_agent` job :eyes:

meowjesty:
> Something weird happened in the test_agent job 👀

What? How?

Razz4780:
> > Something weird happened in the test_agent job 👀
> 
> What? How?

Aliens

mirrord/progress/src/messages.rs
```diff
@@ -11,3 +11,6 @@ pub const MULTIPOD_WARNING: (&str, &str) = (
         You can get started with mirrord for Teams at this link: \
         https://mirrord.dev/docs/overview/teams/",
 );
+
+pub const EXEC_DOCKER_BINARY: &str = "`mirrord exec docker` detected! \
+    Please use `mirrord container [options] exec` instead.";
```
Razz4780, line 16:
```suggestion
pub const EXEC_DOCKER_BINARY: &str = "`mirrord exec <docker|podman|nerdctl> ...` detected! \
    If you try to run a container with mirrord, please use `mirrord container [options] -- <docker|podman|nerdctl> ...` instead.";
```

mirrord/cli/src/main.rs
```diff
@@ -321,6 +321,10 @@ async fn exec(args: &ExecArgs, watch: drain::Watch) -> Result<()> {
         args.binary, args.binary_args
     );
 
+    if args.binary.contains("docker") {
```
Razz4780, line 324:
```suggestion
    if args.binary.contains("docker") || args.binary.contains("podman") || args.binary.contains("nerdctl") {
```

Razz4780 (changes requested):
1. I'd handle all container CLIs we support with `mirrord container ...`
2. Small fix in suggested command line

mirrord/cli/src/main.rs
meowjesty: Added it as a regex, so we can just add `|whatever`, wdyt? (also is the regex ok? it'll match on things like `foo-docker`)

mirrord/cli/src/main.rs
```diff
@@ -321,6 +322,12 @@ async fn exec(args: &ExecArgs, watch: drain::Watch) -> Result<()> {
         args.binary, args.binary_args
     );
 
+    let container_detection = Regex::new(".*(docker|podman|nerdctl).*")
```
Razz4780, line 325:
```suggestion
    let container_detection = Regex::new("docker|podman|nerdctl")
```
Should be the same (doesn't have to match full string)

Razz4780 (approved):

PR #2667, opened 2024-08-15
Small fix on shared sockets

### LINKED ISSUE #864: TCP Steal/mirror doesn't work with shared sockets

DmitryDodzin:
closes #864

### DIFF
--- a/changelog.d/864.fixed.md
+++ b/changelog.d/864.fixed.md
@@ -1,0 +1,1 @@
+Fix small error in shared sockets that resulted in it adding the shared socket env several times.
--- a/mirrord/layer/src/exec_hooks.rs
+++ b/mirrord/layer/src/exec_hooks.rs
@@ -33,18 +33,19 @@ impl Argv {
         self.0.push(item);
     }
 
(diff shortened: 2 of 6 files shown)


Razz4780 (approved):
Nice! I'd also enable the shared sockets test on macOS as part of this PR

PR #2686, opened 2024-08-19
Intproxy logs in tests.

meowjesty:
- Issue: N/A

1. Adds intproxy logs to the tests;
    - produces a file with the test name (based on rust's thread test name) with the logs;


Razz4780 (commented):
+1 for the thread name trick ^^

I'd made two changes:
1. Read thread name in `init_tracing`
2. If thread name does not exist or file creation fails, redirect logs to stderr instead and log an error

You could also add a doc to `init_tracing` and mention it in `CONTRIBUTING.md` (there's a section there on integration tests)

Razz4780 (approved):

PR #2698, opened 2024-08-22
(Operator) Adds `statefulsets`, `cronjobs`, and `jobs` to the `mirrord ls` test.

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/602


Razz4780 (commented):
I don't see the any check that `mirrord ls` actually *does* print the new types :eyes:

Razz4780 (changes requested):
Now we have 6 different `mirrord ls` tests in the CI :eyes: 
Can you just add 4 more `assert!`s to the existing `mirrord_ls` test?
```rust
assert!(targets
        .iter()
        .any(|output| output.starts_with(&format!("pod/{}", service.name))));
assert!(targets
        .iter()
        .any(|output| output.starts_with(&format!("deployment/{}", service.name))));
assert!(targets
        .iter()
        .any(|output| output.starts_with(&format!("statefulset/{}", service.name))));
...
```

meowjesty:
> Now we have 6 different mirrord ls tests in the CI 

Changed! That was helping me figure out what was going wrong.

Razz4780 (approved):
:+1:

Issue #2500, opened 2024-06-06
npm module not found when running turbo app


t4lz:
### Bug Description

Getting module not found when running some npm applications.

Seems like it could be related to #1956 and #1957, as we're getting the same error.

What is happening is that node is looking for module in a location that is relative to a path in our temp dir, but we want it to look for modules relative to the original path.
Not sure why our protections for that (hooking `_NSGetExecutablePath`, passing non-patched paths in exec args) did not prevent this. Maybe now the exec path is retrieved in a different way.

Volta-installed node is SIP-protected, unlike most node installations we've seen so far. I think this might be related to this issue.

### Steps to Reproduce

```
git clone git@github.com:t4lz/mirrord-sip-npm-issue.git
```
```
cd mirrord-sip-npm-issue
```
you might have to install vollta and then install npm and npx via vollta at this point.
```
npm run repro
```

### Backtrace

```shell
│ node:internal/modules/cjs/loader:1148
│   throw err;
│   ^
│ 
│ Error: Cannot find module '../lib/cli.js'
│ Require stack:
│ - /private/var/folders/4l/810mmn597cx7zy5w2bk11clh0000gn/T/mirrord-bin/Users/tal/.volta/tools/image/node/20.14.0/bin/npm
│     at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
│     at Module._load (node:internal/modules/cjs/loader:986:27)
│     at Module.require (node:internal/modules/cjs/loader:1233:19)
│     at require (node:internal/modules/helpers:179:18)
│     at Object.<anonymous> (/private/var/folders/4l/810mmn597cx7zy5w2bk11clh0000gn/T/mirrord-bin/Users/tal/.volta/tools/image/node/20.14.0/bin/npm:2:1)
│     at Module._compile (node:internal/modules/cjs/loader:1358:14)
│     at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
│     at Module.load (node:internal/modules/cjs/loader:1208:32)
│     at Module._load (node:internal/modules/cjs/loader:1024:12)
│     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12) {
│   code: 'MODULE_NOT_FOUND',
│   requireStack: [
│     '/private/var/folders/4l/810mmn597cx7zy5w2bk11clh0000gn/T/mirrord-bin/Users/tal/.volta/tools/image/node/20.14.0/bin/npm'
│   ]
│ }
[...]

PR #2702, opened 2024-08-25
wip patching env for turbo

aviramha:
working on patching Turbo to add missing env for fixing #2500 on Linux


aviramha:
still doesn't add env, I suspect we need to add another hook :\

PR #2704, opened 2024-08-27
Removed all operator-managed resources from mirrord ls output

Razz4780:
Relevant for https://github.com/metalbear-co/operator/issues/635

Regardless of the linked operator issue, nice to have due to following scenario:
1. Operator is installed in the cluster
2. User runs `mirrord ls` without explicitly enabling the operator
3. For some reason we fail to list targets using the operator
4. We fall back to manual listing
5. We end up showing copied pods


Razz4780:
Superseded by #2707

PR #2708, opened 2024-08-28
Patch operator setup to not use impersonate api

DmitryDodzin:
Small update to operator rbac setup to use `SubjectAccessReview` api

Added:
```yaml
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
```

Removed:
```yaml
- apiGroups:
  - ""
  - authentication.k8s.io
  resources:
  - groups
  - users
  - userextras/accesskeyid
  - userextras/arn
  - userextras/canonicalarn
  - userextras/sessionname
  - userextras/iam.gke.io/user-assertion
  - userextras/user-assertion.cloud.google.com
  - userextras/principalid
  - userextras/oid
  - userextras/username
  - userextras/licensekey
  verbs:
  - impersonate
```


aviramha:
It doesn't need `get` on the resource?

DmitryDodzin:
> It doesn't need `get` on the resource?

It's a create only api (the metadata is required to be empty) and you get the response in `status` on the create request

PR #2710, opened 2024-08-29
Rejecting empty composite HTTP filters

Razz4780:
If we get an empty list of filters, it's probably by user's mistake and the filter doesn't do anything useful (accepts/rejects all messages)


gememma (approved):
🐙

PR #2711, opened 2024-08-29
Allow hostnames in port-forward args

### LINKED ISSUE #2696: Support DNS resolution in port-forward command
Port forwarding only currently supports IP addresses, not DNS

gememma:
Closes #2696

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -399,7 +399,16 @@ pub(super) struct PortForwardArgs {
 #[derive(Clone, Debug, PartialEq)]
 pub struct PortMapping {
     pub local: SocketAddr,
-    pub remote: SocketAddr,
+    pub remote: (RemoteAddr, u16),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum RemoteAddr {
+    // if the remote is given as an IPv4
+    Ip(Ipv4Addr),
+    // if the remote needs DNS resolution, we'll delay until PortForwarder can use the
+    // AgentConnection
+    Hostname(String),
 }
 
 impl FromStr for PortMapping {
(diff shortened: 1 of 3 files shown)


mirrord/cli/src/config.rs
```diff
@@ -399,7 +399,16 @@ pub(super) struct PortForwardArgs {
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum RemoteAddr {
+    // if the remote is given as an IPv4
+    Ip(Ipv4Addr),
+    // if the remote needs DNS resolution, we'll delay until PortForwarder can use the
+    // AgentConnection
+    Hostname(String),
 }
```
Razz4780, line 412: At this point I'd extend the doc for `PortForwardArgs::port_mappings` and describe what an argument should look like and what it does and that hostnames are resolved remotely in the cluster (the docs are visible to the user when they do `mirrord port-forward --help`)

mirrord/cli/src/config.rs
```diff
@@ -417,39 +426,35 @@ impl FromStr for PortMapping {
...
             }
-            [remote_ip, remote_port] => {
+            [remote_ip_str, remote_port] => {
                 let remote_port = parse_port(remote_port, string)?;
-                let remote_ip = parse_ip(remote_ip, string)?;
-                (remote_port, remote_ip, remote_port)
+                (remote_port, remote_ip_str, remote_port)
             }
             _ => {
                 return Err(PortMappingParseErr::InvalidFormat(string.to_string()));
```
Razz4780, line 450: Error message for `PortMappingParseErr::InvalidFormat` needs an update

mirrord/cli/src/port_forward.rs
```diff
@@ -55,44 +56,82 @@ pub struct PortForwarder {
...
+
         for mapping in parsed_mappings {
             if listeners.contains_key(&mapping.local) {
                 // two mappings shared a key thus keys were not unique
                 return Err(PortForwardError::PortMapSetupError(mapping.local));
             }
+            let resolved_remote = match mapping.remote.0 {
+                RemoteAddr::Ip(ip) => IpAddr::V4(ip),
+                RemoteAddr::Hostname(hostname) => {
+                    PortForwarder::resolve_hostname(&mut agent_connection, hostname)
```
Razz4780, line 122: I think we want to resolve the hostname lazily (only right before making remote connection via agent). Probably requires another variant of `PortForwardMessage`

Razz4780 (changes requested):

mirrord/cli/src/port_forward.rs
```diff
@@ -268,6 +288,45 @@ impl PortForwarder {
                     );
                 }
             },
+            DaemonMessage::GetAddrInfoResponse(GetAddrInfoResponse(message)) => match message {
```
Razz4780, line 291: Not quite sure what's going on in this branch.
1. The record can contain more than 1 entry
2. `else` block inside `Ok(DnsLookup(record))` branch looks different than `Err(error)` branch, but it's practically the same case, I think

I'd do sth like
```rust
DaemonMessage::GetAddrInfoResponse(GetAddrInfoResponse(message)) => match message {
    Ok(DnsLookup(record)) if !record.is_empty() => {
        // yay
    }
    error_or_empty => {
        // nay
    }
}
```

mirrord/cli/src/config.rs
```diff
@@ -392,14 +392,27 @@ pub(super) struct PortForwardArgs {
     pub context: Option<String>,
 
     /// Mappings for port forwarding
+    /// Expected format is: '-L \[local_port:\]remote_ip_or_hostname:remote_port'
+    /// If the remote is given as an ip, this is parsed as soon as mirrord starts
+    /// Otherwise, the remote is assumed to be a hostname and lookup is performed in the cluster
+    /// after a connection is made to the target
```
Razz4780, line 398:
```suggestion
    /// after a local connection is accepted.
```
Also, can you separate sentences with periods? I don't think clap respects newlines when producing `--help` message from docs

Razz4780 (changes requested):
Noice, two small things

Razz4780 (approved):
:+1:

PR #2721, opened 2024-09-03
Add attribution to mfT docs links

### LINKED ISSUE #2703: Add attribution to redirections to mfT
We redirect users to mfT from within the product (e.g. "You're trying to target a deployment, this is supported in mfT"). We need to be able to attribute website visits to these links in order to analyze them, do A/B tests, etc.

gememma:
Part of #2703

### DIFF
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -289,7 +289,7 @@ pub(crate) enum CliError {
     #[error("Feature `{0}` requires using mirrord operator")]
     #[diagnostic(help(
         "The mirrord operator is part of mirrord for Teams. \
-        You can get started with mirrord for Teams at this link: https://mirrord.dev/docs/overview/teams/?utm_source=errreqop"
+        You can get started with mirrord for Teams at this link: https://mirrord.dev/docs/overview/teams/?utm_source=errreqop&utm_medium=cli"
     ))]
     FeatureRequiresOperatorError(String),
 
@@ -308,7 +308,7 @@ pub(crate) enum CliError {
 
     If you want to run without the operator, please set `\"operator\": false` in the mirrord configuration file.
 
-    Please remember that some features are supported only when using mirrord operator (https://mirrord.dev/docs/overview/teams/#supported-features?utm_source=erropfailed).{GENERAL_HELP}"))]
+    Please remember that some features are supported only when using mirrord operator (https://mirrord.dev/docs/overview/teams/#supported-features?utm_source=erropfailed&utm_medium=cli).{GENERAL_HELP}"))]
     OperatorApiFailed(OperatorOperation, kube::Error),
 
     #[error("mirrord operator rejected {0}: {1}")]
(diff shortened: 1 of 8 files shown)


gememma:
- utm_source = {one per msg}
    - multipodwarn - warning that multipod mode is on but no operator
    - copytarget - generated config from doc comments, no platform
    - teamscmd - cli only, mirrord Teams command
    - errreqop - cli only, error message for if config has copy target enabled but no operator
    - erropfailed - cli only, error message for when something failed while using the operator (general OperatorApiError::KubeError or OperatorApiError::StatusFailure)
- utm_medium = {cli, (plugin), vscode, intellij}
    - parsed from plugin → specific in the plugin code
    - vscode: can use replace() on link string

mirrord/cli/src/error.rs
```diff
@@ -289,7 +289,7 @@ pub(crate) enum CliError {
     #[error("Feature `{0}` requires using mirrord operator")]
     #[diagnostic(help(
         "The mirrord operator is part of mirrord for Teams. \
-        You can get started with mirrord for Teams at this link: https://mirrord.dev/docs/overview/teams/"
+        You can get started with mirrord for Teams at this link: https://mirrord.dev/docs/overview/teams/?utm_source=errreqop&utm_medium=cli"
```
t4lz, line 292: Does this error not reach the IDE user? Either way, the plugins should probably also replace `utm_medium=cli`, right?

mirrord/cli/src/teams.rs
```diff
@@ -1,6 +1,7 @@
 use tokio::process::Command;
 
-const MIRRORD_FOR_TEAMS_URL: &str = "https://mirrord.dev/docs/overview/teams/";
+const MIRRORD_FOR_TEAMS_URL: &str =
```
t4lz, line 3: We could make this const local to the function that uses it so that we don't use the same utm_source in other links by mistake.

mirrord/config/src/feature.rs
```diff
@@ -85,7 +85,7 @@ pub struct FeatureConfig {
     /// ## feature.copy_target {#feature-copy_target}
     ///
     /// Creates a new copy of the target. mirrord will use this copy instead of the original target
-    /// (e.g. intercept network traffic). This feature requires a [mirrord operator](https://mirrord.dev/docs/overview/teams/).
+    /// (e.g. intercept network traffic). This feature requires a [mirrord operator](https://mirrord.dev/docs/overview/teams/?utm_source=copytarget).
```
t4lz, line 88: do we actually want those query parameters in the website? not only in links that show in the CLI/IDE?

t4lz (approved):
Looks good, had a couple of questions, but nothing critical.

mirrord/cli/src/error.rs
gememma: pretty sure this specific one doesn't (at least in vscode) but i agree they should check for both

mirrord/config/src/feature.rs
gememma: maybe @eyalb181 has an opinion?

eyalb181: I didn't think about it, but sure, why not. Would be interesting to learn if the docs are what leads people to mfT.

PR #2723, opened 2024-09-04
Change `dns_configuration_copy` to return remote value

### LINKED ISSUE #2713: Can't connect to Mongo on Node 20 + macOS
Node 20.17.0 + 20.16.0 doesn't work
Node 18 does work
Doesn't work both via CLI or IDE (WebStorm)
macOS
dotenv ^8.2.0
mongodb ^4.17.0
typescript 4.7.4
using `new MongoClient`

DmitryDodzin:
closes #2713

### DIFF
--- a/mirrord/layer/Cargo.toml
+++ b/mirrord/layer/Cargo.toml
@@ -49,10 +49,10 @@ exec.workspace = true
 syscalls = { version = "0.6", features = ["full"] }
 null-terminated = "0.3"
 base64.workspace = true
-resolv-conf = "0.7"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 mirrord-sip = { path = "../sip" }
+resolv-conf = "0.7"
 
 [dev-dependencies]
 mirrord-intproxy = { path = "../intproxy" }
--- a/mirrord/layer/Cargo.toml
+++ b/mirrord/layer/Cargo.toml
@@ -49,6 +49,7 @@ exec.workspace = true
 syscalls = { version = "0.6", features = ["full"] }
 null-terminated = "0.3"
 base64.workspace = true
+resolv-conf = "0.7"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 mirrord-sip = { path = "../sip" }
--- a/mirrord/layer/src/socket/hooks.rs
+++ b/mirrord/layer/src/socket/hooks.rs
@@ -445,13 +445,15 @@ pub(super) unsafe extern "C" fn sendmsg_detour(
 #[cfg(target_os = "macos")]
 #[allow(non_camel_case_types)]
 mod macos {
-    #[repr(C, align(4))]
+    use alloc::ffi::CString;
+
+    #[repr(C, packed(4))]
     pub struct dns_sortaddr_t {
         pub address: libc::in_addr,
         pub mask: libc::in_addr,
     }
 
-    #[repr(C, align(4))]
+    #[repr(C, packed(4))]
     pub struct dns_resolver_t {
         pub domain: *mut libc::c_char,
         pub n_nameserver: i32,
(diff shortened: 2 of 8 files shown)


mirrord/layer/Cargo.toml
```diff
@@ -49,6 +49,7 @@ exec.workspace = true
 syscalls = { version = "0.6", features = ["full"] }
 null-terminated = "0.3"
 base64.workspace = true
+resolv-conf = "0.7"
```
Razz4780, line 52: Should be moved to `[target.'cfg(target_os = "macos")'.dependencies]` section

DmitryDodzin: Correct

mirrord/layer/src/socket/hooks.rs
```diff
@@ -442,68 +444,25 @@ pub(super) unsafe extern "C" fn sendmsg_detour(
...
-        pub reserved: [u32; 5],
-    }
-}
-
-#[cfg(target_os = "macos")]
-use macos::*;
-
 /// This implementation is actually enough for Netty case, since it seems to use the "standard"
 /// approach if resolver returned here is null TODO: return a real resolver based on remote
 /// resolv.conf
```
t4lz, line 449: That TODO is DONE, right?
Could you please also explain this detour in the function documentation?
What is this function supposed to do normally, what we do instead and why.

mirrord/layer/src/socket/hooks.rs
```diff
@@ -442,68 +444,25 @@ pub(super) unsafe extern "C" fn sendmsg_detour(
...
-        n_scoped_resolver: 0,
-        scoped_resolver: std::ptr::null_mut(),
-        reserved: [0; 5],
-    }))
+    remote_dns_configuration_copy().unwrap_or_bypass_with(|_| FN_DNS_CONFIGURATION_COPY())
 }
 
 #[cfg(target_os = "macos")]
 #[hook_guard_fn]
 unsafe extern "C" fn dns_configuration_free_detour(config: *mut dns_config_t) {
```
t4lz, line 458: please document what this detour is for - what do we do different from the original and why.

mirrord/layer/src/socket/ops.rs
```diff
@@ -1392,3 +1421,111 @@ pub(super) fn sendmsg(
...
+        let _ = Box::from_raw(nameserver);
+    }
+
+    let searchs = Vec::from_raw_parts(resolver.search, resolver.n_search as usize, 0);
+
+    for search in searchs {
+        let _ = CString::from_raw(search);
+    }
+
+    let _ = CString::from_raw(resolver.options);
```
t4lz, line 1481: Why do these pointers need to be freed but not `domain` and `sortaddr`?

mirrord/layer/src/socket/ops.rs
```diff
@@ -1392,3 +1421,111 @@ pub(super) fn sendmsg(
...
+}
+
+#[cfg(target_os = "macos")]
+#[mirrord_layer_macro::instrument(level = "trace", ret)]
+pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
+    let remote = read_remote_resolv_conf()?;
+
+    // TODO: replace error
```
t4lz, line 1489: If this TODO is meant to stay there, it needs more details.

mirrord/layer/src/socket/ops.rs
DmitryDodzin: Well I only free any memory we only allocated, I changed so the or_bypass_with will return an empty one so the drop won't accidentally break or leak any memory

mirrord/layer/src/socket/ops.rs
DmitryDodzin: I changed it to be more detailed

mirrord/layer/src/socket/hooks.rs
DmitryDodzin: Updated

mirrord/layer/src/socket/ops.rs
```diff
@@ -1392,3 +1421,123 @@ pub(super) fn sendmsg(
...
+
+    let _ = CString::from_raw(resolver.options);
+}
+
+/// reconstruct a macos specific [`dns_config_t`] api from parsing the `/etc/resolv.conf` file from
+/// remote container and fill only the first resolver in the response
+#[cfg(target_os = "macos")]
+#[mirrord_layer_macro::instrument(level = "trace", ret)]
+pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
+    let remote = read_remote_resolv_conf()?;
```
t4lz, line 1498: shouldn't we at some point check the user's mirrord config? What if they disabled DNS?

DmitryDodzin: Ah

DmitryDodzin: Nvm the detours are not hooked when no dns

t4lz (approved):

PR #2729, opened 2024-09-04
Update sidecar error printing

### LINKED ISSUE #2726: Mirrord containers show unintuitive error
### Bug Description

Unintuitive error when using containers, if the sidecar fails it is deleted
```
Error:   × Comand failed to execute command [docker logs be6fe8a8bcc8bb8f721851fb5fd2c560666c0b03261f97c0db0dd94f74bf0e1f]: Error response from daemon: No such
  │ container: be6fe8a8bcc8bb8f721851fb5fd2c560666c0b03261f97c0db0dd94f74bf0e1f
```

### Steps to Reproduce

run where result of `local_ip_address::local_ip` is not accessible from container: 
```
mirrord container -- docker run -it debian
```

### Backtrace

_No response_

### Relevant Logs

_No response_

linux

### Local process

docker

### Local process version

_No response_

### Additional Info

_No response_

DmitryDodzin:
Closes #2726

### DIFF
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -133,8 +133,14 @@ async fn create_sidecar_intproxy(
     sidecar_command.add_env(MIRRORD_INTPROXY_CONTAINER_MODE_ENV, "true");
     sidecar_command.add_envs(connection_info);
 
-    let sidecar_container_command =
-        ContainerCommand::run(["-d", &config.container.cli_image, "mirrord", "intproxy"]);
+    let sidecar_container_command = ContainerCommand::run(
+        config
+            .container
+            .cli_extra_args
+            .iter()
+            .map(String::as_str)
+            .chain(["-d", &config.container.cli_image, "mirrord", "intproxy"]),
+    );
 
     let (runtime_binary, sidecar_args) = sidecar_command
         .with_command(sidecar_container_command)
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -6,7 +6,6 @@ use std::{
 };
 
 use exec::execvp;
-use local_ip_address::local_ip;
 use mirrord_analytics::{
     AnalyticsError, AnalyticsReporter, CollectAnalytics, ExecutionKind, Reporter,
 };
@@ -251,13 +250,7 @@ pub(crate) async fn container_command(
     let _external_proxy_tls_guards = if config.external_proxy.tls_certificate.is_none()
         || config.external_proxy.tls_key.is_none()
     {
-        let external_proxy_subject_alt_names = config
-            .external_proxy
-            .address
-            .or_else(|| local_ip().ok())
-            .map(|item| item.to_string())
-            .into_iter()
-            .collect();
+        let external_proxy_subject_alt_names = vec![config.external_proxy.address.clone()];
 
         let (external_proxy_cert, external_proxy_key) =
             create_self_signed_certificate(external_proxy_subject_alt_names)?;
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -28,7 +28,7 @@ use crate::{
     config::{ContainerArgs, ContainerCommand},
     connection::AGENT_CONNECT_INFO_ENV_KEY,
     container::command_builder::RuntimeCommandBuilder,
-    error::{ContainerError, Result},
+    error::{CliError, ContainerError, Result},
     execution::{
         MirrordExecution, LINUX_INJECTION_ENV_VAR, MIRRORD_CONNECT_TCP_ENV,
         MIRRORD_EXECUTION_KIND_ENV,
@@ -128,7 +128,7 @@ async fn create_sidecar_intproxy(
     config: &LayerConfig,
     base_command: &RuntimeCommandBuilder,
     connection_info: Vec<(&str, &str)>,
-) -> Result<(String, SocketAddr), ContainerError> {
+) -> Result<(String, SocketAddr)> {
     let mut sidecar_command = base_command.clone();
 
     sidecar_command.add_env(MIRRORD_INTPROXY_CONTAINER_MODE_ENV, "true");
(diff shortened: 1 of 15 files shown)


mirrord/cli/src/container.rs
```diff
@@ -158,6 +153,16 @@ async fn create_sidecar_intproxy(
             )
         })?;
 
+    let mut sidecar_watcher_command = Command::new(std::env::current_exe().unwrap());
```
meowjesty, line 156:
```suggestion
    let mut sidecar_watcher_command = Command::new(std::env::current_exe().expect("Could not get the full path of the binary (??)!"));
```

Maybe dress the error up a bit if it happens for a user (even though it's extremely unlikely)?

mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +27,18 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
 #[config(map_to = "ExternalProxyFileConfig", derive = "JsonSchema")]
 #[cfg_attr(test, config(derive = "PartialEq"))]
 pub struct ExternalProxyConfig {
+    /// ### external_proxy.listen {#external_proxy-listen}
```
meowjesty, line 30: Think it would be helpful if you added a bit of reasoning on why you might need to set this up?

mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +27,18 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
...
+    ///
+    /// Provide a specific address to listen to for external proxy
+    /// (will try and bind local machine ip if not specified)
+    pub listen: Option<IpAddr>,
+
+    /// ### external_proxy.address {#external_proxy-address}
```
meowjesty, line 36: Same for this config. The user might have to set this up why?

mirrord/cli/src/sidecar_watcher.rs
```diff
@@ -0,0 +1,78 @@
+use std::time::Duration;
+
+use tokio::process::Command;
+use tokio_retry::strategy::{jitter, ExponentialBackoff};
+
+use crate::{config::ContainerRuntime, error::Result};
+
+pub async fn watcher(runtime: ContainerRuntime, container_id: &str) -> Result<()> {
```
meowjesty, line 8: Docs. Should be pretty similar to what you put in the changelog.

mirrord/cli/src/sidecar_watcher.rs
```diff
@@ -0,0 +1,78 @@
...
+                    tokio::time::sleep(backoff_timeout).await
+                } else {
+                    tracing::error!(
+                        ?container_id,
+                        ?error,
+                        ?runtime,
+                        "timout reached for waiting for <runtime> attach to <container_id>"
+                    );
+
+                    // TODO: maybe throw an error though the stdout is detached, the tracing is
```
meowjesty, line 50: TODO?

mirrord/cli/src/sidecar_watcher.rs
```diff
@@ -0,0 +1,78 @@
...
+                    // mainly for mirrord console
+                    return Ok(());
+                }
+            }
+        }
+    }
+
+    tracing::info!("exit detected waiting 5s before delete");
```
meowjesty, line 58: Is this message supposed to be seem by the user? If so, then maybe we should go for something more descriptive? "(whatever operation) exiting. Container will be deleted after 5 seconds.". Something along those lines?

meowjesty (changes requested):
:whale:

mirrord/cli/src/sidecar_watcher.rs
DmitryDodzin: I don't intend for `sidecar-watcher` to have stdin/out but all these traces can be used with `MIRRORD_CONSOLE_ADDR`

mirrord/cli/src/sidecar_watcher.rs
DmitryDodzin: There is no stdin/out/err so adding an error would only result in a different exit code but the child_handle is ignored in `mirrord container` command right now

mirrord/cli/src/container.rs
DmitryDodzin: I did it like the rest

mirrord/config/src/external_proxy.rs
DmitryDodzin: I updated it a bit but can change it to hostname sadly

mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +27,35 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
...
 #[cfg_attr(test, config(derive = "PartialEq"))]
 pub struct ExternalProxyConfig {
+    /// ### external_proxy.listen {#external_proxy-listen}
+    ///
+    /// Provide a specific address to listen to for external proxy
+    /// (will try and bind local machine ip if not specified)
+    ///
+    /// This is a workaround for when the local machine ip that is default is bound is not
+    /// accecible from the container (local address resoved to `192.168.0.2` but said ip is
+    /// inacceible from within a container)
```
meowjesty, line 37:
```suggestion
    /// This is a workaround for when the local machine ip that is bound by default, is not
    /// accessible from the container (local address resolved to `192.168.0.2` but this ip is
    /// inaccessible from within a container)
```

mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +27,35 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
...
+
+    /// ### external_proxy.address {#external_proxy-address}
+    ///
+    /// Specify an address that is accessible from within the container runtime to the host machine
+    ///
+    /// This is a workaround where the listen address should be different from the one the
+    /// container is connecting to, example can be where `host.docker.internal -> 192.168.127.254`
+    /// and we intend to utilize the network briging
+    ///
+    /// Note: this needs to be an ip because of TLS nagotiation.
```
meowjesty, line 48:
```suggestion
    /// and we intend to utilize the network bridging
    ///
    /// Note: this needs to be an ip because of TLS negotiation.
```

meowjesty (approved):
Just some typos left to squash.

PR #2732, opened 2024-09-04
Do not sort targets from mirrord ls based on string values, instead let the iterator chain handle it.

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord-vscode/issues/147

Remove a string sort that we had in the targets, when running `mirrord ls`. We have a neat chain of iterators that sorts the targets by type, starting with `pods`, which is nicer than alphabetical ordering.

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -444,10 +444,9 @@ async fn print_targets(args: &ListTargetArgs) -> Result<()> {
         remove_proxy_env();
     }
 
-    let mut targets = list_targets(&layer_config, args).await?;
-
-    targets.sort();
-
+    // The targets come sorted in the following order:
+    // `pods - deployments - rollouts - jobs - cronjobs - statefulsets`
+    let targets = list_targets(&layer_config, args).await?;
     let json_obj = json!(targets);
     println!("{json_obj}");
     Ok(())


Razz4780 (approved):

Razz4780:
Note - targets are still sorted in the plugin [here](https://github.com/metalbear-co/mirrord-vscode/blob/cc8e8f10f28cf30abe4fb6ad668dd669aed8d896/src/api.ts#L357)

aviramha:
Can we switch the order so deployments will always be first?

Razz4780:
> Can we switch the order so deployments will always be first?

Maybe deployments -> rollouts -> whatever?

aviramha:
> > Can we switch the order so deployments will always be first?
> 
> Maybe deployments -> rollouts -> whatever?

sure

meowjesty:
I don't have any rollout, but cli now presents it like this:

```
$ mirrord ls
["deployment/sample-python-deployment","statefulset/web","cronjob/autoquit-cronjob","pod/sample-python-deployment-796d8bcd4f-zv62c","pod/web-0","pod/web-1"]
```

And the extension:

image

mirrord/cli/src/main.rs
```diff
@@ -444,10 +444,9 @@ async fn print_targets(args: &ListTargetArgs) -> Result<()> {
         remove_proxy_env();
     }
 
-    let mut targets = list_targets(&layer_config, args).await?;
-
-    targets.sort();
-
+    // The targets come sorted in the following order:
+    // `pods - deployments - rollouts - jobs - cronjobs - statefulsets`
```
Razz4780, line 448:
```suggestion
    // `deployments - rollouts - statefulsets - cronjobs - jobs - pods`
```

Razz4780 (approved):
Only comment needs an update :+1:

PR #2745, opened 2024-09-09
Enable lint for unused crates, cleanup unused crates

aviramha:


aviramha:
ok this lint doesn't work well enough ;(

PR #2747, opened 2024-09-09
Suggest mfT when user uses HTTP filter

### LINKED ISSUE #2701: Write a message about mfT when user uses HTTP filter
Usage of HTTP filter is a strong indicator that the user is running on a shared environment. We should use this opportunity to inform them of mfT, which is better suited for shared environments.

gememma:
Closes #2701

### DIFF
--- a/mirrord/cli/src/connection.rs
+++ b/mirrord/cli/src/connection.rs
@@ -9,7 +9,8 @@ use mirrord_kube::{
 };
 use mirrord_operator::client::{OperatorApi, OperatorSessionConnection};
 use mirrord_progress::{
-    messages::MULTIPOD_WARNING, IdeAction, IdeMessage, NotificationLevel, Progress,
+    messages::{HTTP_FILTER_WARNING, MULTIPOD_WARNING},
+    IdeAction, IdeMessage, NotificationLevel, Progress,
 };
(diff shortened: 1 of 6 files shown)


mirrord/cli/src/connection.rs
```diff
@@ -120,40 +121,44 @@ where
...
-            },
-        })?);
-        // This is CLI Only because the extensions also implement this check with better messaging.
-        progress.print("When targeting multi-pod deployments, mirrord impersonates the first pod in the deployment.");
-        progress.print("Support for multi-pod impersonation requires the mirrord operator, which is part of mirrord for Teams.");
-        progress.print("You can get started with mirrord for Teams at this link: https://mirrord.dev/docs/overview/teams/?utm_source=multipodwarn&utm_medium=cli");
-    }
+        (true, true) => {
+            // only show user one of the two msgs - each user should always be shown same msg
+            if mid::get("mirrord")
```
DmitryDodzin, line 141: Maybe wrap this logic to a function just so it will understandable that this is for not spamming user with random messages (`user_persistent_random_message_select() -> bool`?)

DmitryDodzin (approved):
Just a nitpick other than that LGTM

PR #2752, opened 2024-09-12
Refactor ResolvedTarget from operator to mirrord. Warn when targeting a multi-container pod (and no container was specified).

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/605

1. Moves `ResolvedTarget` from the operator to mirrord;
    - It's in the `mirrord-operator` crate, but it could be moved to some other shared crate if/when we want to unify target resolving;
    - Brings the `GET target` checks from the operator to mirrord;
2. Adds a new warning when a random container was selected by the operator (multi-container pod, and user did not specify the container).


mirrord/operator/src/client.rs
```diff
@@ -628,39 +548,39 @@ impl OperatorApi<PreparedClientCert> {
...
+            && config
+                .target
+                .path
+                .as_ref()
+                .is_some_and(|target| target.container().is_none())
+        {
+            progress.warning(&format!(
+                "Target has multiple containers, but no container was specified by user!\
+                    mirrord will pick a random container `{}`.",
+                target.container().unwrap_or_default()
```
t4lz, line 561: No point in including this, because inside this `if`, `target.container()` always returns `None`, doesn't it? and then we'd always get the default `""` here.

mirrord/operator/src/client.rs
```diff
@@ -628,39 +548,39 @@ impl OperatorApi<PreparedClientCert> {
...
-            }
-            _ => false,
-        };
+        if target.containers_count() > 1
+            && config
+                .target
+                .path
+                .as_ref()
+                .is_some_and(|target| target.container().is_none())
+        {
```
t4lz, line 557: Warning logic should match the logic when running without an operator - don't warn if there is only 1 non-skipped container (skipped containers are e.g. containers inserted by service meshes etc.).
If those two cases could share code for the common logic that could be nice.

t4lz (commented):
Not done reviewing.

mirrord/kube/src/resolved.rs
```diff
@@ -0,0 +1,399 @@
...
+    Job(Job, Option<String>),
+    CronJob(CronJob, Option<String>),
+    StatefulSet(StatefulSet, Option<String>),
+    Pod(Pod, Option<String>),
+    Targetless(String),
+}
+
+impl ResolvedTarget {
+    /// Gets a target from k8s with the [`Client`] that is passed here, and checks if it's
+    /// a valid mirrord target with [`ResolvedTarget::assert_valid_mirrord_target`].
```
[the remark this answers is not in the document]
meowjesty: The lie has been removed.

mirrord/kube/src/resolved.rs
```diff
@@ -0,0 +1,397 @@
...
+        }?;
+
+        Ok(target)
+    }
+
+    /// Check if the target can be used as a mirrord target.
+    ///
+    /// 1. [`ResolvedTarget::Deployment`] or [`ResolvedTarget::Rollout`] - has available replicas
+    ///    and the target container, if specified, is found in the spec
+    /// 2. [`ResolvedTarget::Pod`] - passes target-readiness check of `StatusObserver`
```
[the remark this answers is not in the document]
meowjesty: Changed to point to the `get_full_runtime_data` function instead of `StatusObserver`.

mirrord/kube/src/resolved.rs
```diff
@@ -0,0 +1,397 @@
...
+        .and_then(|status| status.container_statuses.as_ref())
+        .into_iter()
+        .flatten()
+        .find(|status| status.name == runtime_data.container_name)
+        .expect(
+            "container should exist in containerStatuses, we just found it with RuntimeData::from_pod",
+        )
+        .restart_count;
+
+    Ok((runtime_data, restart_count))
```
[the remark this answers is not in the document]
meowjesty: The result is only used in the operator to check active status. In the `ResolvedTarget` check this makes no sense, good catch.

mirrord/cli/src/connection.rs
```diff
@@ -81,8 +82,21 @@ where
     let api = api.prepare_client_cert(analytics).await.into_certified()?;
     user_cert_subtask.success(Some("user credentials prepared"));
 
+    // TODO(alex) [high] 3: I don't think we can `ResolveTarget` here, it must be done only
+    // after we decide if it should be copy or not? Or at least, if we can't find a valid
+    // target, then it should become a copy?
```
t4lz, line 87: Then why is it here, I don't get it 😬

mirrord/cli/src/error.rs
```diff
@@ -359,6 +360,17 @@ pub(crate) enum CliError {
...
     "))]
     KubeAuthExecFailed(String),
+
+    #[error("Failed while resolving target while using the mirrord-operator: {0}")]
+    #[diagnostic(help(
+        "
+        mirrord failed to resolve or validate a target.
+        Target resolution failure happens when the target cannot be found, or doesn't exist.
+        Validation may fail for a variety of reasons, such as: target is in an invalid state, or missing required fields.
+        Please check that your user (in mirrord-operator context) has access to the target, and that the target actually exists in the cluster.
```
t4lz, line 370: What's your user in the mirrord-operator context?

Would this be correct:

```suggestion
        Please check that your Kubernetes user has access to the target, and that the target actually exists in the cluster.
```

mirrord/operator/src/client.rs
```diff
@@ -747,6 +654,12 @@ impl OperatorApi<PreparedClientCert> {
     ///
     /// `copy_target` feature is not available for all target types.
     /// Target type compatibility is checked by the operator.
+    // TODO(alex) [high] 2: Now I need to fix this somehow. Ideally, it would make a request
+    // to the operator to create the copy, then we would list the targets and turn it into
+    // a `ResolvedTarget`, or maybe `impl From<CopyTargetCrd> for ResolvedTarget`, instead
+    // of listing.
+    // I think we could come here after `ResolvedTarget` has been used, which is not ideal,
+    // as there will be no target to resolve?
```
t4lz, line 662: Not sure I understand you, but this sounds like a merge blocker, no?

Can't you just not resolve the target before calling this, and leave the whole flow as it was before this change?

mirrord/cli/src/connection.rs
meowjesty: To avoid this kind of confusion, I was thinking of _statefying_ the `ResolvedTarget`. Something along the lines of `ResolvedTarget<Unchecked>` and `ResolvedTarget<Checked>`, wdyt?

mirrord/cli/src/error.rs
meowjesty: I guess it is. I was probably being too explicit with the notion that we're talking about the kubernetes' user config in regards to the operator setup. The idea is how we can build a kube-client for some general context, and a kube-client that has operator stuff in it.

mirrord/operator/src/client.rs
meowjesty: > Can't you just not resolve the target before calling this, and leave the whole flow as it was before this change?

Yup, that's kinda what's in the code now. It was an outdated comment from the `ResolvedTarget::new` again.

Resolving a target should be harmless to the copy target stuff (as far as I'm aware), if we don't check anything in it.

mirrord/cli/src/connection.rs
meowjesty: Maybe something like this [moving states into a type parameter](https://willcrichton.net/rust-api-type-patterns/typestate.html#moving-states-into-a-type-parameter)

mirrord/operator/src/client.rs
meowjesty: And to clarify, we have to resolve a target before anything `copy_target` related, to check if it's an empty deployment.

mirrord/operator/src/client.rs
meowjesty: Got it.

How far should I go to merge these apis? I spent some time playing around with it, and I think it would be nice if we leaned towards using `ResolvedTarget` more. What I mean is, create a type for each variant, then implemented the `RuntimeData*` stuff that we have for `config::Target` for the variants.

The quickest approach would be to just pass the `config::Target` to the `connect_in_new_session`, and call `target.runtime_data(operator_client, ...)`.

The end result will be pretty much the same, but if we implement this in `ResolvedTarget`, we kinda decouple our code more from config stuff, and I think it would look a bit less confusing.

:thinking: :thought_balloon:  Thoughts?

mirrord/kube/src/resolved.rs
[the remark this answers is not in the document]
t4lz: There is no `get_full_runtime_data` anymore in here, I think.

mirrord/operator/src/client.rs
t4lz: Didn't completely understand, but IMO the best thing for this PR would be to stay with the current behaviour and logic, especially if it does not mean writing a lot of new code. I think it's best to separate refactors from actual changes when possible, and only introduce new stuff together with refactors if keeping the old logic would mean writing new code that would immediately be replaced on the change to the new logic.
If this is the case here (keeping old logic would require a significant amount of new code), let's do call and discuss those options while looking at the code together.

mirrord/cli/src/connection.rs
t4lz: Not sure if it's necessary, but I'm fine with that, I think.
And BTW, is state as generic type parameter better than state as a const generics bool?

meowjesty: Considering that it's just 2 states, const generics bool is way easier to implement for an enum!

mirrord/kube/src/resolved/cron_job.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct ResolvedCronJob {
+    pub resource: CronJob,
+    pub container: Option<String>,
+}
+
+impl RuntimeDataFromLabels for ResolvedResource<CronJob> {
+    type Resource = CronJob;
+
+    fn name(&self) -> &str {
+        self.resource.metadata.name.as_ref().unwrap().as_str()
```
t4lz, line 21: Is this code used or could be used by the operator? If so, maybe let's not `unwrap`, even though k8s is promises to populate that field? Maybe just change to `unwrap_or_default`? That way in the highly unlikely case that we get invalid information from k8s, the operator does not panic, but just has 1 session with this wrong name. Of course we could make this method return a `Result` or an `Option`, but IMO the complication of the interface is not worth if for the tiny chance this would happen. Up to you.

t4lz, line 21: This method can be implemented in the trait definition instead of in all implementations, as the associated type has a bound to `Resource`.

mirrord/kube/src/resolved.rs
```diff
@@ -10,25 +10,40 @@ use mirrord_config::{feature::network::incoming::ConcurrentSteal, target::Target
...
+    Rollout(ResolvedResource<Rollout>),
+    Job(ResolvedResource<Job>),
+    CronJob(ResolvedResource<CronJob>),
+    StatefulSet(ResolvedResource<StatefulSet>),
+    Pod(ResolvedResource<Pod>),
     Targetless(String),
 }
 
+#[derive(Debug, Clone)]
+pub struct ResolvedResource<R>
```
t4lz, line 40: Wondering if we even need `ResolvedTarget` if we have this struct.

mirrord/kube/src/resolved/cron_job.rs
meowjesty: Only if we have a method `resource` in the trait, so we can do `self.resource().meta().*`, but the `config::Target` stuff doesn't hold the `Resource`, so it can't really be done.

e.g.

```rust
impl RuntimeDataFromLabels for RolloutTarget {
    type Resource = Rollout;

    fn resource(&self) -> &Self::Resource { 
        // what could go here?
        // self.resource <- only exists for the `Resource<Rollout>` type
    }
```

meowjesty: `Cow` to the rescue, changed to `unwrap_or_default`. Had to change the return though, as we can't return a temporary (the _default).

mirrord/kube/src/resolved/deployment.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use super::ResolvedResource;
+use crate::{api::runtime::RuntimeDataFromLabels, error::KubeApiError};
+
+#[derive(Debug, Clone)]
+pub struct ResolvedDeployment {
+    pub resource: Deployment,
+    pub container: Option<String>,
+}
+
```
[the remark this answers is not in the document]
meowjesty: Removed!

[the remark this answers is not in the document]
meowjesty: Think there was another stray struct, that was also removed.

mirrord/kube/src/resolved.rs
meowjesty: Maybe not, if we convert the `config::Target` to a typed `ResolvedResource<Deployment>`, for example, but this could get a bit messy to initialize.

- A simplified example of this type creation (converting directly from `Target`):

```rust
impl<R> ResolvedResource<R> {
  // Warning: returning different types!
  fn new(target: config::Target) -> Self {
    match target { 
      Target::Deployment(dep) => ResolvedResource { resource: dep },
      Target::Job(job) => ResolvedResource { resource: job },
    }
  }
}
```

mirrord/operator/src/client.rs
meowjesty: Alright, now I think we're using the right stuff to check if we're in multi-container.

mirrord/kube/src/resolved.rs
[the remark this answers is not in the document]
meowjesty: Yup! Changed the misleading docs again.

image

mirrord/kube/src/resolved/cron_job.rs
```diff
@@ -0,0 +1,47 @@
...
+            .as_ref()
+            .map(Cow::from)
+            .unwrap_or_default()
+    }
+
+    fn container(&self) -> Option<&str> {
+        self.container.as_deref()
+    }
+
+    async fn get_labels(resource: &Self::Resource) -> Result<BTreeMap<String, String>> {
```
t4lz, line 27: I suggest renaming this method to `get_selector_match_labels`, because `get_labels` sounds a lot like it returns the resource's labels (which are under metadata.labels), like [`ResourceExt::labels()`](https://docs.rs/kube/latest/kube/trait.ResourceExt.html#tymethod.labels) does.

t4lz (approved):
Nice. One more optional suggestion.
And don't make the towncrier cry, give him a change log.

mirrord/operator/src/crd.rs
```diff
@@ -3,7 +3,7 @@ use std::{
     fmt::{Display, Formatter},
 };
 
-use kube::CustomResource;
+use kube::{CustomResource, Resource};
```
t4lz, line 6: You need to undo this for the docs check not to return an error.

tests/src/operator/queue_splitting.rs
t4lz, line 1: you can drop all changes done to this file, unless you think there is a good reason to have longer timeouts.

meowjesty: Left the change of `from queue1` to `from queue 1`, the most important change of all time.

mirrord/operator/src/crd.rs
meowjesty: Need this for the `meta()` calls. Removed the `kube::Resource` part from the docs instead.

t4lz (approved):

Issue #2750, opened 2024-09-12
Enable internal proxy log by default to enable easier troubleshooting


aviramha:
We have quite few issues that raise from layer saying internal proxy died, but users can't know why it happened.
I suggest the following:

1. Make internal proxy log by default to tempdir with `mirrord_datetime_rand4chars.log`
2. Prompt user to look into it and upload it to the issue they open when they get an internal proxy error.

https://github.com/metalbear-co/mirrord/issues/2742

meowjesty:
Now that 1 has been done, I think 2 should be part of the error report, but do we also to add it in the "new issue" template thingy?

Issue #2652, opened 2024-08-14
Mirrord often exits during debugging


izturn:
### Bug Description

when debug with vsc, the mirrord often exits during debugging with output likes this:
`
Proxy error, connectivity issue or a bug. 
You may report it to us on https://github.com/metalbear-co/mirrord/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml 
io failed: Resource temporarily unavailable (os error 11)
`

the config likes:
```
{
    "feature": {
        "network": {
            "incoming": {
                "mode": "steal"
              }
        },
        "fs": "read",
        "env": true
    },

    "use_proxy": false,
    "kubeconfig": "~/.kube/config",
    "target": {
        "path": {
            "pod": "xx-59b69f96b5-7bp9v",
            "container": "xx-container"
        },
       "namespace": "xx-system"
      },
    "internal_proxy": {
        "idle_timeout": 180,
        "log_destination": "/tmp/internal_proxy.log",
        "log_level": "mirrord=trace"
    },
    "agent": {
        "log_level": "mirrord=trace,debug,warn",
        "namespace": "default",
        "image": "ghcr.io/metalbear-co/mirrord:latest"
    }
}


```

### Steps to Reproduce

1.  just run the app in mode: `debug` with `mirrord` attached.
2. sometime past, the debugged process is exited as above.

### Relevant Logs

```shell
2024-08-14T06:56:11.212235Z TRACE handle_agent_read{read=Ok(DaemonRead { connection_id: 51, bytes (length): 20635 }) protocol=Stream}: mirrord_intproxy::proxies::outgoing: outgoing interceptor 51-STREAM does not exist, received data for connection that is already closed
2024-08-14T06:56:11.212249Z TRACE handle_agent_read{read=Ok(DaemonRead { connection_id: 51, bytes (length): 12381 }) protocol=Stream}: mirrord_intproxy::proxies::outgoing: outgoing interceptor 51-STREAM does not exist, received data for connection that is already closed
2024-08-14T06:56:11.212260Z TRACE handle_agent_read{read=Ok(DaemonRead { connection_id: 51, bytes (length): 3956 }) protocol=Stream}: mirrord_intproxy::proxies::outgoing: outgoing interceptor 51-STREAM does not exist, received data for connection that is already closed
[...]

PR #2754, opened 2024-09-13
Increase timeout of intproxy connection to a ludicrous amount (avoids intproxy issues when debugging with breakpoints).

meowjesty:
- Issue: #2652 
#2750 (default intproxy logging catching a ride)

I couldn't really reproduce the issue, but this should be enough to avoid it.


mirrord/cli/src/internal_proxy.rs
```diff
@@ -53,25 +56,42 @@ pub(crate) async fn proxy(watch: drain::Watch) -> Result<(), InternalProxyError>
...
+        .log_destination
+        .as_ref()
+        .map(PathBuf::from)
+        .unwrap_or_else(|| {
+            let random_name: String = rand::thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(7)
+                .map(char::from)
+                .collect();
+            let timestamp = Utc::now().timestamp();
```
Razz4780, line 71:
```suggestion
            let timestamp = Utc::now().timestamp();
```
```suggestion
            let timestamp = SystemTime::UNIX_EPOCH.elapsed().unwrap();
```
Can be done with Rust std, no need to add `chrono` :p

mirrord/cli/src/internal_proxy.rs
```diff
@@ -53,25 +56,42 @@ pub(crate) async fn proxy(watch: drain::Watch) -> Result<(), InternalProxyError>
...
+        .open(&log_destination)
+        .map_err(|fail| {
+            InternalProxyError::OpenLogFile(log_destination.to_string_lossy().to_string(), fail)
+        })?;
+
+    let log_level = config
+        .internal_proxy
+        .log_level
+        .clone()
+        .unwrap_or_else(|| "mirrord=trace".to_string());
```
Razz4780, line 88:
```suggestion
        .clone()
        .unwrap_or_else(|| "mirrord=trace".to_string());
```
```suggestion
        .as_deref()
        .unwrap_or("mirrord=trace");
```
We don't need owned, I think

mirrord/config/src/internal_proxy.rs
```diff
@@ -84,6 +84,22 @@ pub struct InternalProxyConfig {
...
+    /// Sometimes the cpu is too busy with other tasks and the internal proxy is blocked
+    /// from performing its keep-alive check. Increase this value if mirrord exits while
+    /// you're debugging with breakpoints.
+    ///
+    /// ```json
+    /// {
+    ///   "internal_proxy": {
+    ///     "idle_timeout": 31536000
+    ///   }
+    /// }
```
Razz4780, line 98: This is kind of misleading :v
1. The default value is already at max, so "Increase this value if mirrord exits while you're debugging with breakpoints" does not sound right
2. Snippet contains `idle_timeout` field
3. The explanation itself is kinda off, the problem is not about keep-alive checks

This setting is not really for internal proxy but for the layer, since we set the timeout there. I propose we move this into `experimental` section under `.experimental.intproxy_socket_timeout` without any instructions (just a doc saying that it's a read/write timeout for layer<->intproxy sockets, the users should not normally touch it without our directions)

mirrord/cli/src/internal_proxy.rs
Razz4780, line 88: Not sure if we should default to `trace`. That's gonna produce large files for long mirrord sessions.
Maybe we should:
1. Move default logging change to separate PR
2. Use `mirrord=info` as default level
3. Make sure that intproxy logs on `info` are not spammy and still informative (only layer connects/disconnects and files/connections/subscriptions state changes)
 @aviramha wdyt?

Razz4780 (changes requested):

mirrord/config/src/internal_proxy.rs
meowjesty: I don't think the user should be seeing this either, but it feels like it belong under the `internal_proxy` config. I mean, this is a setting for an internal proxy startup thingy. Aren't the `experimental` configs for things we're not quite sure we want? This one feels like it doesn't hurt to have, so I see no reason to remove it in the future.

Razz4780: Tbh I think `experimental` is mainly for hacks :V

> this is a setting for an internal proxy startup thingy

Wait, no. You're thinking about `.internal_proxy.idle_timeout`. My comment is for your new setting. The timeout specified in it is set on layer's TCP socket for communicating with intproxy (so applies to any request<->response exchange)

meowjesty: > layer's TCP socket for communicating with intproxy

Yeah I got that, but isn't that part of the proxy startup? When we have a new layer connection? That's what I mean.

meowjesty: Anyway, I changed the docs.

mirrord/cli/src/internal_proxy.rs
meowjesty: New PR just for this at #2764

aviramha: sounds like a plan.

mirrord/layer/src/lib.rs
```diff
@@ -164,7 +164,7 @@ static EXECUTABLE_PATH: OnceLock<String> = OnceLock::new();
 
 /// Proxy Connection timeout
 /// Set to 10 seconds as most agent operations timeout after 5 seconds
```
Razz4780, line 166:
```suggestion
/// Read/write timeout for layer<->intproxy TCP sockets.
/// Can be configured in the [`LayerConfig`].
```

mirrord/cli/Cargo.toml
```diff
@@ -68,6 +68,8 @@ tokio-stream = { workspace = true, features = ["net"] }
 tokio-retry = "0.3"
 regex.workspace = true
 mid = "2.1.0"
+rand.workspace = true
+chrono = { version = "0.4" }
```
Razz4780, line 72: Delet :wastebasket:

mirrord/cli/src/internal_proxy.rs
```diff
@@ -53,25 +55,42 @@ pub(crate) async fn proxy(watch: drain::Watch) -> Result<(), InternalProxyError>
...
+        .internal_proxy
+        .log_level
+        .as_deref()
+        .unwrap_or("mirrord=trace");
+
+    tracing_subscriber::fmt()
+        .with_writer(output_file)
+        .with_ansi(false)
+        .with_env_filter(EnvFilter::builder().parse_lossy(log_level))
+        .init();
```
Razz4780, line 93: Delet :wastebasket:

mirrord/layer/src/lib.rs
```diff
@@ -626,7 +630,10 @@ pub(crate) unsafe extern "C" fn fork_detour() -> pid_t {
             let new_connection = ProxyConnection::new(
                 parent_connection.proxy_addr(),
                 NewSessionRequest::Forked(parent_connection.layer_id()),
-                PROXY_CONNECTION_TIMEOUT,
+                PROXY_CONNECTION_TIMEOUT
+                    .get()
+                    .copied()
+                    .unwrap_or_else(|| Duration::from_secs(31536000)),
```
Razz4780, line 636: Should be initialized, I'd just do `.get().expect(...)`, so we don't need this value hardcoded here

mirrord-schema.json
```diff
@@ -1247,6 +1247,16 @@
       "description": "Configuration for the internal proxy mirrord spawns for each local mirrord session that local layers use to connect to the remote agent\n\nThis is seldom used, but if you get `ConnectionRefused` errors, you might want to increase the timeouts a bit.\n\n```json { \"internal_proxy\": { \"start_idle_timeout\": 30, \"idle_timeout\": 5 } } ```",
       "type": "object",
       "properties": {
+        "breakpoint_timeout": {
```
Razz4780, line 1250: Needs regeneration

Razz4780 (changes requested):

mirrord/config/src/internal_proxy.rs
Razz4780: Oke, let's leave it in `.internal_proxy` section

Razz4780 (approved):

PR #2764, opened 2024-09-17
Enable intproxy logs by default.

meowjesty:
- Issue: #2750


Razz4780:
Gotta make sure that `info` level produces informative logs on layer<->intproxy<->agent requests and responses, except messages that only transfer data (like in file or remote connections reads/writes)

aviramha:
@Razz4780 to be on the same page - you want to check there's no spam, not to add more messages, right?

Razz4780:
> @Razz4780 to be on the same page - you want to check there's no spam, not to add more messages, right?

I think this will only require moving some message from `trace` to `info`

aviramha:
> > @Razz4780 to be on the same page - you want to check there's no spam, not to add more messages, right?
> 
> I think this will only require moving some message from `trace` to `info`

I'm quite against this - I think it'd be premature optimization - I'd guess that most important stuff are getting errors/warnings and if we need to add more things let's do it when we realize that.

Razz4780 (approved):

Issue #2518, opened 2024-06-12
readlink stabilization


aviramha:
In few months time, when we stabilize readlink, make sure to warn at first at unused config `readlink` under `experimental` and don't fail, for another period of time.

eyalb181:
On September 20th, check if we can stabilize this according to the versions currently in use.

aviramha:
tbh we can use the protocol detection to use it, and we already have `READDIR_BATCH_VERSION` which includes it for sure

aviramha:
c

aviramha:
gggg

aviramha:
b

PR #2766, opened 2024-09-18
Make readlink stable (enabled by default) now, with the protocol part gated on version support.

meowjesty:
- Issue: #2518

### DIFF
--- a/mirrord/agent/src/file.rs
+++ b/mirrord/agent/src/file.rs
@@ -412,7 +412,7 @@ impl FileManager {
     }
 
     /// Handles our `readlink_detour` with [`std::fs::read_link`].
-    #[tracing::instrument(level = "trace", skip_all)]
+    #[tracing::instrument(level = Level::TRACE, skip_all, err(level = Level::ERROR))]
     pub(crate) fn read_link(&mut self, path: PathBuf) -> RemoteResult<ReadLinkFileResponse> {
         let path = path
             .strip_prefix("/")
--- a/mirrord/agent/src/file.rs
+++ b/mirrord/agent/src/file.rs
@@ -412,7 +412,7 @@ impl FileManager {
     }
 
     /// Handles our `readlink_detour` with [`std::fs::read_link`].
-    #[tracing::instrument(level = Level::TRACE, skip_all, err(level = Level::ERROR))]
+    #[tracing::instrument(level = Level::TRACE, skip_all)]
     pub(crate) fn read_link(&mut self, path: PathBuf) -> RemoteResult<ReadLinkFileResponse> {
         let path = path
             .strip_prefix("/")
--- a/mirrord/config/src/lib.rs
+++ b/mirrord/config/src/lib.rs
@@ -526,7 +526,7 @@ impl LayerConfig {
 
         if self.experimental.readlink {
             context.add_warning(
-                "experimenta.readlink config has been deprecated, and `readlink` is now\
+                "experimental.readlink config has been deprecated, and `readlink` is now\
                     enabled by default! You may remove it from your config."
                     .into(),
             );
--- a/mirrord/config/src/lib.rs
+++ b/mirrord/config/src/lib.rs
@@ -524,6 +524,14 @@ impl LayerConfig {
         self.feature.network.dns.verify(context)?;
         self.feature.network.outgoing.verify(context)?;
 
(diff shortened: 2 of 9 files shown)


mirrord/agent/src/file.rs
```diff
@@ -412,7 +412,7 @@ impl FileManager {
     }
 
     /// Handles our `readlink_detour` with [`std::fs::read_link`].
-    #[tracing::instrument(level = "trace", skip_all)]
+    #[tracing::instrument(level = Level::TRACE, skip_all, err(level = Level::ERROR))]
```
Razz4780, line 415:
```suggestion
    #[tracing::instrument(level = Level::TRACE, skip_all)]
```
This should not be logged globally as an agent error, user might simply use a non-existent path

mirrord/config/src/lib.rs
```diff
@@ -524,6 +524,14 @@ impl LayerConfig {
         self.feature.network.dns.verify(context)?;
         self.feature.network.outgoing.verify(context)?;
 
+        if self.experimental.readlink {
+            context.add_warning(
+                "experimenta.readlink config has been deprecated, and `readlink` is now\
```
Razz4780, line 529:
```suggestion
                "experimental.readlink config has been deprecated, and `readlink` is now\
```

mirrord/layer/tests/common/mod.rs
```diff
@@ -741,6 +741,7 @@ pub enum Application {
     RustRecvFrom,
     RustListenPorts,
     Fork,
+    /// `mirrord/layer/tests/apps/readlink/readlink.c`
```
Razz4780, line 744: :eyes:

Razz4780 (changes requested):
Small stuff only

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -217,8 +217,34 @@ impl BackgroundTask for SimpleProxy {
...
+                SimpleProxyMessage::FileReq(
+                    message_id,
+                    layer_id,
+                    req @ FileRequest::ReadLink(_),
+                ) => {
+                    if protocol_version
+                        .as_ref()
+                        .is_some_and(|version| READDIR_BATCH_VERSION.matches(version))
+                    {
+                        self.file_reqs.insert(message_id, layer_id);
```
aviramha, line 230: This means that if user calls readlink, they'll just crash - that's not what we want

aviramha (changes requested):

mirrord/layer/src/file/ops.rs
```diff
@@ -295,24 +295,18 @@ pub(crate) fn pread(local_fd: RawFd, buffer_size: u64, offset: u64) -> Detour<Re
...
 
 /// Resolves the symbolic link `path`.
-///
-/// Bypassed if the `experimental.readlink` config is not set to `true`.
-#[mirrord_layer_macro::instrument(level = "trace", ret)]
+#[mirrord_layer_macro::instrument(level = Level::TRACE, ret)]
 pub(crate) fn read_link(path: Detour<PathBuf>) -> Detour<ReadLinkFileResponse> {
-    if crate::setup().experimental().readlink {
-        let path = remap_path!(path?);
+    let path = remap_path!(path?);
```
aviramha, line 300: We need here to replace the check of the flag with `if version > BATCHED..` then do logic if not None? (like existing) so it'll call bpass

mirrord/intproxy/src/proxies/simple.rs
meowjesty: My bad. I changed it to `Bypass` on `NotImplemented` error. This way we avoid having to check the version in the layer.

mirrord/layer/tests/common/mod.rs
meowjesty: Just thought it would be nice to have the test files there. Should I remove or may I keep it?

aviramha (approved):
Approving to remove blocker - fixes looks ok but haven't looked at the whole thing.

mirrord/layer/tests/common/mod.rs
meowjesty: Nah, at that point I've already done my `gr` command to see which file it is.

Razz4780 (approved):

Issue #1898, opened 2023-09-04
Respect socket's non blocking connect


aviramha:
Current implementation of the connect detour always waits for the other side to accept before it returns, while most applications nowadays set the socket to non blocking, then connect should immediately return with EINPROGRESS error code, indicating that the connect is in progress.
We can improve our implementation to check if socket is non blocking, and if it is, we setup the mirror listen socket, but don't accept until agent returns the result, to reproduce similar behavior (as much as possible).

Razz4780:
It would be a bit tricky to implement it this way. The usual flow is to call `connect` and expect `EINPROGRESS`, then use `poll` to wait for result. `getsockopt` can also be used to get connected/error state of the socket. We'd have to asynchronously let the layer know that the listener is ready (which doesn't fit well into our request-response model and lack of background threads :X)

A simpler solution would be to follow the usual TCP proxy pattern - eagerly accept connection, kill it if we fail to connect to the real peer

aviramha:
Why is it tricky though?
Layer sends ConnectRequest
intProxy responds ConnectInProgress
layer continues
Intproxy accepts connection only after connection is really made on the remote side, if not returns error

PR #2776, opened 2024-09-23
Respect non-blocking connect

DmitryDodzin:
relevant for #1898


mirrord/layer/src/socket/ops.rs
```diff
@@ -569,7 +639,17 @@ pub(super) fn connect(
 
     let unix_streams = crate::setup().remote_unix_streams();
 
-    trace!("in connect {:#?}", SOCKETS);
+    let socket_fd_flags = nix::fcntl::fcntl(sockfd, nix::fcntl::FcntlArg::F_GETFL)
```
aviramha, line 642: Why do we need to check it?
If socket is blocking, it will block the app, if not it won't (and we'll get einprogress) - why does intproxy care?

DmitryDodzin: Because we block to get response from intproxy before we call `connect` to intproxy interceptor addr so my solution was to send einprogress as soon as intproxy handles the connect request and not when the reply from the agent is returned

aviramha: but you can just call the original `CONNECT` and if you get einprogress then you know it's not blocking and you pass it to caller and if it doesn't return it then you need to wait for a reply from upstream

DmitryDodzin: But connect to what address? intproxy doesn't yet start the interceptor until it has a response from the agent

aviramha: change order - start interceptor before getting answer from agent.

aviramha: (but don't accept until you get response, to simulate real stuff)

DmitryDodzin: What to do with errors? like when error is an actual error?

DmitryDodzin: Like right now I also don't have a good solution

aviramha: what happens when einprogress errors? I think it doesn't?
the internal proxy should just timeout it's socket/accept

mirrord/layer/src/socket/ops.rs
```diff
@@ -413,6 +417,53 @@ pub(super) fn listen(sockfd: RawFd, backlog: c_int) -> Detour<i32> {
     }
 }
 
+fn non_blocking_connect_watcher(
```
meowjesty, line 420: I'm trying to understand what's going on with this, and wondering if there isn't a more straightforward solution to the issue.

I wonder: is there a reason why we can't have a secondary thread in the layer, that loops checking for some sort of `BackgroundMessage`, and drop this necessity of having all of our messages be request/responses? Wouldn't something like this simplify all of this?

Elaborating on what I'm thinking is something like: layer sends the usual connect request, intproxy responds to it immediately (either with a success or `EINPROGRESS`), user gets the result of this operation, and we put the socket in some `InProgress(remote_addr/whatever)` state. On this secondary thread, at some point we receive `Connected(id, stuff)`, and change the socket to be in a connected state.

I think something like this would be sort of a `poll` implementation?

mirrord/layer/src/socket/ops.rs
meowjesty: I also don't think we need this?

mirrord/layer/src/socket/ops.rs
meowjesty: If my idea is dumb, then please add docs for this function.

meowjesty (changes requested):
Not done reviewing.

Overall this feels a bit confusing to me, on an already confusing part of the code. I suggested something that, hopefully, would be simpler (big question mark)

mirrord/cli/src/internal_proxy.rs
aviramha, line 1: Please either create a separate PR for this or roll it back

mirrord/intproxy/protocol/src/lib.rs
```diff
@@ -117,13 +118,49 @@ impl fmt::Display for NetProtocol {
     }
 }
 
+bitflags! {
```
aviramha, line 121: I don't think int proxy should know whether it's blocking or nonblocking.

aviramha (changes requested):

aviramha:
I don't understand current implementation - why does int proxy need to know this is non blocking? this should be purely from layer perspective, and even that shouldn't care.
Can you draw a diagram of how it's supposed to work, and why can't we do it without intproxy knowing?

mirrord/intproxy/protocol/src/lib.rs
aviramha: but.. getpeeraddr is hooked by us anyway, why not modify the struct/hook to support it?

DmitryDodzin:
```
                                                                                                        
                                                                                                        
  NonBlockTask      Layer                   Intproxy             Agent  socket       getpeeraddr        
                                                                        state                           
        │             │      Connect(r)         │                  │  ◄───────┐           │             
        │             ├────────────────────────►│     Connect(r)   │          │        Not Ready        
        │             │                         ├─────────────────►│   Init   │           │             
        │             │     InProgress(l)       │                  │  ◄───────┤           │             
        │             │◄────────────────────────┤                  │          │           │             
        │             │                         │                  │          │           │             
        │  wait_reply │     libc_connect(l)     │                  │          │           │             
        │◄────────────┼────────────────────────►│                  │Connecting│     127.0.0.1:23451     
        │             │◄────────────────────────┤                  │          │           │             
        │             │  libc_connected(l)      │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │  accept_libc_connect(l) │   Connected(r)   │          │           │             
        │◄────────────┼─────────────────────────┤◄─────────────────┤  ◄───────┤           │             
        │             │                         │                  │          │       8.8.8.8:53        
        │             │      connected(l)       │                  │          │           │             
        ├─────────────┼────────────────────────►│                  │          │           │             
        │             │                         │                  │ Connected│           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │          │           │             
        │             │                         │                  │  ◄───────┘           │             
                                                                                                        
                                                                                                        
```

aviramha:
> ```
>                                                                                                         
>                                                                                                         
>   NonBlockTask      Layer                   Intproxy             Agent  socket       getpeeraddr        
>                                                                         state                           
>         │             │      Connect(r)         │                  │  ◄───────┐           │             
>         │             ├────────────────────────►│     Connect(r)   │          │        Not Ready        
>         │             │                         ├─────────────────►│   Init   │           │             
>         │             │     InProgress(l)       │                  │  ◄───────┤           │             
>         │             │◄────────────────────────┤                  │          │           │             
>         │             │                         │                  │          │           │             
>         │  wait_reply │     libc_connect(l)     │                  │          │           │             
>         │◄────────────┼────────────────────────►│                  │Connecting│     127.0.0.1:23451     
>         │             │◄────────────────────────┤                  │          │           │             
>         │             │  libc_connected(l)      │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │  accept_libc_connect(l) │   Connected(r)   │          │           │             
>         │◄────────────┼─────────────────────────┤◄─────────────────┤  ◄───────┤           │             
>         │             │                         │                  │          │       8.8.8.8:53        
>         │             │      connected(l)       │                  │          │           │             
>         ├─────────────┼────────────────────────►│                  │          │           │             
>         │             │                         │                  │ Connected│           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │          │           │             
>         │             │                         │                  │  ◄───────┘           │             
>                                                                                                         
>                                                                                                         
> ```

beautiful - why can't we change socket state to connecting then return info?

DmitryDodzin:
> 
> beautiful - why can't we change socket state to connecting then return info?

We don't yet have the address from the agent, and If I just return `EINPROGRESS` then it breaks blocking sockets because they do block but only for the connection to internal proxy and not until the internal proxy gets the response from the agent

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -199,11 +208,40 @@ impl OutgoingProxy {
...
+
+            message_bus
+                .send(ToLayer {
+                    message: ProxyToLayerMessage::OutgoingConnect(response),
+                    message_id,
+                    layer_id,
+                })
+                .await;
+
+            if did_error {
```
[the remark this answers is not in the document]
DmitryDodzin: Yep, I moved the adding to queue lower

Razz4780:
We found the problem quite complex. Solving it brings more risk than value and we don't know of anyone affected. Closing for now.

PR #2781, opened 2024-09-26
Reverse port-forwarding

### LINKED ISSUE #2609: Support reverse port-forwarding
#567 like that, but stealing/mirroring traffic to a local port

gememma:
Closes #2609

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -488,7 +488,7 @@ pub enum PortMappingParseErr {
     PortZeroInvalid(String),
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Copy)]
 pub struct PortOnlyMapping {
     pub local: LocalPort,
     pub remote: RemotePort,
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -398,6 +398,10 @@ pub(super) struct PortForwardArgs {
     /// after a connection is made to the target.
     #[arg(short = 'L', long)]
     pub port_mappings: Vec<PortMapping>,
+
+    /// Enable reverse port forwarding
+    #[arg(short = 'r', long)]
+    pub reverse_port_forward: bool,
 }
 
 #[derive(Clone, Debug, PartialEq)]
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -398,7 +398,7 @@ pub(super) struct PortForwardArgs {
     /// after a connection is made to the target.
     /// Multiple forwarding mappings are each passed with -L.
     #[arg(short = 'L', long)]
-    pub port_mappings: Option<Vec<AddrPortMapping>>, // TODO: remove options
+    pub port_mappings: Vec<AddrPortMapping>,
 
     /// Mappings for reverse port forwarding.
     /// Expected format is: '-R \[remote_port:\]local_port'.
@@ -407,7 +407,7 @@ pub(super) struct PortForwardArgs {
     /// target:remote_port.
     /// Multiple reverse mappings are each passed with -R.
     #[arg(short = 'R', long)]
-    pub reverse_port_mappings: Option<Vec<PortOnlyMapping>>,
+    pub reverse_port_mappings: Vec<PortOnlyMapping>,
 }
 
 #[derive(Clone, Debug, PartialEq)]
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -396,15 +396,16 @@ pub(super) struct PortForwardArgs {
     /// If the remote is given as an ip, this is parsed as soon as mirrord starts.
     /// Otherwise, the remote is assumed to be a hostname and lookup is performed in the cluster
     /// after a connection is made to the target.
+    /// Multiple forwarding mappings are each passed with -L.
     #[arg(short = 'L', long)]
     pub port_mappings: Option<Vec<AddrPortMapping>>,
 
-    /// Enable reverse port forwarding
-    #[arg(short = 'r', long)]
-    pub reverse_port_forward: bool,
-
     /// Mappings for reverse port forwarding.
     /// Expected format is: '-R \[remote_port:\]local_port'.
+    /// In reverse port forwarding, traffic to the remote_port on the target pod is stolen or
+    /// mirrored to localhost:local_host. If stealing, the response is sent back to the
+    /// target:remote_port.
+    /// Multiple reverse mappings are each passed with -R.
     #[arg(short = 'R', long)]
     pub reverse_port_mappings: Option<Vec<PortOnlyMapping>>,
 }
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -486,9 +486,12 @@ pub enum PortMappingParseErr {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct PortOnlyMapping {
-    pub local: u16,
-    pub remote: u16,
+    pub local: LocalPort,
+    pub remote: RemotePort,
 }
+
+pub type LocalPort = u16;
+pub type RemotePort = u16;
 
 impl FromStr for PortOnlyMapping {
     type Err = PortMappingParseErr;
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -408,9 +408,9 @@ pub(super) struct PortForwardArgs {
     #[arg(short = 'R', long)]
     pub reverse_port_mappings: Option<Vec<PortOnlyMapping>>,
 
-    /// Steal traffic from the remote. If false, the traffic is mirrored
-    #[arg(long)]
-    pub steal_traffic: bool,
+    /// Steal TCP instead of mirroring
+    #[arg(long = "steal")]
+    pub tcp_steal: bool,
 }
 
 #[derive(Clone, Debug, PartialEq)]
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -341,7 +341,7 @@ impl TargetParams {
 }
 
 #[derive(Args, Debug)]
-#[command(group(ArgGroup::new("port-forward").args(["port_mappings", "reverse_port_mappings"]).required(true)))]
+#[command(group(ArgGroup::new("port-forward").args(["port_mapping", "reverse_port_mapping"]).required(true)))]
 pub(super) struct PortForwardArgs {
     /// Parameters for the target
     #[clap(flatten)]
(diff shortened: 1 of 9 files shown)


gememma:
> Before I even start to review it: reverse port-forwarding should obviously be called port-backwarding.

I only narrowly decided to call the struct `ReversePortForwarder` instead of `PortBackwarder` and all my notes say port-backwarding 😔 regrets

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +435,271 @@ impl PortForwarder {
     }
 }
 
+pub struct ReversePortForwarder {
+    /// details for traffic mirroring or stealing
+    incoming_mode: IncomingMode,
+    /// communicates with the agent (only TCP supported).
+    agent_connection: AgentConnection,
+    /// associates resolved destination ports with local ports.
+    mappings: HashMap<RemotePort, LocalPort>,
```
t4lz, line 444: What does "resolved" mean here?

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +435,271 @@ impl PortForwarder {
...
+        }
+
+        for mapping in &parsed_mappings {
+            // check destinations are unique
+            if mappings.contains_key(&mapping.remote) {
+                // two mappings shared a key thus keys were not unique
+                return Err(PortForwardError::ReversePortMapSetupError(mapping.remote));
+            }
+            mappings.insert(mapping.remote, mapping.local);
+        }
```
t4lz, line 475: How about making the type of the port mapping command argument a tuple type that holds a `HashMap`, and do this parsing and verification in the `FromStr` implementation for that type?

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +435,271 @@ impl PortForwarder {
...
+            return Err(PortForwardError::NoMappingsError());
+        }
+
+        for mapping in &parsed_mappings {
+            // check destinations are unique
+            if mappings.contains_key(&mapping.remote) {
+                // two mappings shared a key thus keys were not unique
+                return Err(PortForwardError::ReversePortMapSetupError(mapping.remote));
+            }
+            mappings.insert(mapping.remote, mapping.local);
```
t4lz, line 474: nitpick
```suggestion
            if mappings.insert(mapping.remote, mapping.local).is_some() {
                // two mappings shared a key thus keys were not unique
                return Err(PortForwardError::ReversePortMapSetupError(mapping.remote));
            }
```

t4lz (commented):
Not done reviewing.

mirrord/cli/src/config.rs
```diff
@@ -397,11 +397,20 @@ pub(super) struct PortForwardArgs {
     /// Otherwise, the remote is assumed to be a hostname and lookup is performed in the cluster
     /// after a connection is made to the target.
     #[arg(short = 'L', long)]
-    pub port_mappings: Vec<PortMapping>,
+    pub port_mappings: Option<Vec<AddrPortMapping>>,
```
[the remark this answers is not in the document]
gememma: The original reason that mapping args are not positional and use `-L` is to imitate the ssh port forward command as closely as possible, with `-R` then being chosen for the reverse. After some internal discussion, it has been decided that the user should be able to use `-L` and `-R` arguments in the same command. In this light, I can remove the -r flag (you're right, it isn't needed) and go from there - I'll play around with the clap `required`-adjacent stuff to see if it can enforce the presence of at least one of `-L, `-R`

mirrord/cli/src/port_forward.rs
gememma: I am not entirely sure this is possible within clap (see [this discussion](https://github.com/clap-rs/clap/discussions/4291)) - I think this check has to be done in ReversePortForwarder

t4lz: Ok so we can use a `Vec`, but the verification that there are no duplicates should happen early (before creating an agent etc.), and the error message should not say "This is a bug". We could accept a single delimited list (`80:80;8080:8080`) and parse it into a hashmap with `FromStr`, but apparently we want to copy the ssh interface, so we will have to live with verification after parsing.

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+        for remote_port in self.mappings.keys() {
+            let subscription = self.incoming_mode.subscription(*remote_port);
+            let msg = match subscription {
+                PortSubscription::Steal(steal_type) => {
+                    ClientMessage::TcpSteal(LayerTcpSteal::PortSubscribe(steal_type))
+                }
+                PortSubscription::Mirror(_) => {
+                    ClientMessage::Tcp(LayerTcp::PortSubscribe(*remote_port))
+                }
+            };
```
t4lz, line 511: Can this existing method be used here?

```suggestion
            let msg = subscription.agent_subscribe();
```

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+            self.agent_connection.sender.send(msg).await?
+        }
+
+        loop {
+            select! {
+                _ = tokio::time::sleep_until(self.ping_pong_timeout.into()) => {
```
t4lz, line 517: I think you can just use [`sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) instead of `sleep_until` and remove the struct member and the calculation, no?

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+                message = self.agent_connection.receiver.recv() => match message {
+                    Some(message) => {
+                        self.handle_msg_from_agent(message).await?},
+                    None => {
+                        break Err(PortForwardError::AgentError("unexpected end of connection with agent".into()));
+                    },
+                },
+
+                Some((task_id, update)) = self.background_tasks.next() => {
+                    // dbg!((&task_id, &update));
```
t4lz, line 536:
```suggestion
```

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+        task_id: MainTaskId,
+        update: TaskUpdate<ProxyMessage, IntProxyError>,
+    ) -> Result<(), PortForwardError> {
+        match (task_id, update) {
+            (MainTaskId::IncomingProxy, TaskUpdate::Message(message)) => match message {
+                ProxyMessage::ToAgent(message) => {
+                    if matches!(
+                        message,
+                        ClientMessage::TcpSteal(LayerTcpSteal::PortSubscribe(_))
+                    ) || matches!(message, ClientMessage::Tcp(LayerTcp::PortSubscribe(_)))
```
t4lz, line 592: I think this could be shortened a bit with something like this, but up to you.

```suggestion
                    if matches!(
                        message,
                        ClientMessage::TcpSteal(LayerTcpSteal::PortSubscribe(_))
                        | ClientMessage::Tcp(LayerTcp::PortSubscribe(_)))
```

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+                    {
+                        // suppress additional subscription requests
+                        return Ok(());
```
t4lz, line 595: Can this ever happen? If not, maybe document that.

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+                ProxyMessage::ToLayer(ToLayer { message, .. }) => match message {
+                    ProxyToLayerMessage::Incoming(IncomingResponse::PortSubscribe(res)) => {
+                        if let Err(error) = res {
+                            return Err(PortForwardError::from(IntProxyError::from(
+                                IncomingProxyError::SubscriptionFailed(error),
+                            )));
+                        }
+                    }
+                    _ => {
+                        unreachable!("message type are never used in port forwarding: {message:?}")
```
t4lz, line 608:
```suggestion
                        unreachable!("message type is never used in port forwarding: {message:?}")
```

right?

t4lz: Thinking about it, maybe we should not panic if this happen, and instead just log and then ignore the message.
I'm thinking about what if we add new message types in the future, that can originate in the agent without the client sending a request first. WDYT?

mirrord/cli/src/port_forward.rs
```diff
@@ -442,12 +420,255 @@ impl PortForwarder {
...
+                    }
+                },
+                ProxyMessage::FromLayer(_)
+                | ProxyMessage::FromAgent(_)
+                | ProxyMessage::NewLayer(_) => {
+                    unreachable!("message type are never used in port forwarding: {message:?}")
```
t4lz, line 614:
```suggestion
                    unreachable!("message type is never used in port forwarding: {message:?}")
```

mirrord/cli/src/port_forward.rs
gememma: The first port subscription is sent explicitly [here during setup](https://github.com/metalbear-co/mirrord/pull/2781/files#diff-73a2e5bd8b740e52a5f910d8ecf50cd4c7249be55d6d58f987ddefa5b2c1c747R458) (port_forward.rs, Ln 458). Then, when the agent sends a new connection, `IncomingProxy` attempts to send another subscription request (I assume because it handles connections lazily in regular `mirrord exec`) - if this second sub is sent to the agent, it generates a "port already subscribed to" error.

The reason for this eager vs lazy eval is because the portforwarder drives port management in `port-forward`, and `IncomingProxy` is being used outside the intended context.

mirrord/cli/src/port_forward.rs
gememma: yep, sounds sensible 👍

t4lz (approved):
Very nice.

mirrord/cli/src/config.rs
```diff
@@ -396,12 +396,22 @@ pub(super) struct PortForwardArgs {
     /// If the remote is given as an ip, this is parsed as soon as mirrord starts.
     /// Otherwise, the remote is assumed to be a hostname and lookup is performed in the cluster
     /// after a connection is made to the target.
+    /// Multiple forwarding mappings are each passed with -L.
     #[arg(short = 'L', long)]
-    pub port_mappings: Vec<PortMapping>,
+    pub port_mappings: Vec<AddrPortMapping>,
```
t4lz, line 401: I still think this should be called `port_mapping`, singular, BTW.
A mapping maps multiple keys two multiple values.
And even if a mapping would only mean one key->value item, I would still find it nicer to write `... --port-mapping 80:3000 --port-mapping 8080:3030` than `... --port-mappings 80:3000 --port-mappings 8080:3030`.
We could ask for more feedback from the team if you're not convinced :)

t4lz: We can leave `--port-mappings` as an [`alias`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.alias) for backwards compatibility.

gememma: i agree, the plural is misleading

Issue #2788, opened 2024-09-30
Cluster with IPv6 leads to agent crash when stealing


aviramha:
Originally from here
https://github.com/metalbear-co/mirrord/issues/2728#issuecomment-2382721602

```
2024-09-30T10:02:38.653317Z TRACE handle_agent_message{Close("Background task `Stealer` failed with `IPTables failed with `code: 2, msg: iptables v1.8.9 (legacy): host/network `fc00:1000::fa24' not found\nTry `iptables -h' or 'iptables --help' for more information.\n``")}: mirrord_intproxy: return=Err(AgentFailed("Background task `Stealer` failed with `IPTables failed with `code: 2, msg: iptables v1.8.9 (legacy): host/network `fc00:1000::fa24' not found\nTry `iptables -h' or 'iptables --help' for more information.\n``"))
```

aviramha:
@sahityacodes does your pod have both ipv6/ipv4 or only ipv6? because if only ipv6 this will be harder to fix.

sahityacodes:
Our pod is deployed with istio proxy container. The main application itself contains IPV4 only, however when deployed in a cluster that supports dual stack configurations, istio proxy can work with both IPv4 and IPv6 I suppose. So, I guess it has both IPv4 and IPv6

aviramha:
Ok so I believe my fix would work for your case :)

PR #2790, opened 2024-09-30
internal proxy now explicitly logs exit errors

aviramha:
This error https://github.com/metalbear-co/mirrord/issues/2788
could've been diagnosed easily given it was marked as an error, and logged by default by the new settings.

### DIFF
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -145,6 +145,9 @@ pub(crate) async fn proxy(watch: drain::Watch) -> Result<(), InternalProxyError>
         .run(first_connection_timeout, consecutive_connection_timeout)
         .await
         .map_err(InternalProxyError::from)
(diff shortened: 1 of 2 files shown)


mirrord/cli/src/internal_proxy.rs
```diff
@@ -145,6 +145,9 @@ pub(crate) async fn proxy(watch: drain::Watch) -> Result<(), InternalProxyError>
         .run(first_connection_timeout, consecutive_connection_timeout)
         .await
         .map_err(InternalProxyError::from)
+        .inspect_err(|e| {
+            tracing::error!("internal proxy exiting: {e:?}");
+        })
```
Razz4780, line 150:
```suggestion
        .inspect_err(|error| {
            tracing::error!(%error, "Internal proxy encountered an error, exiting");
        })
```

Razz4780 (approved):

PR #2793, opened 2024-10-01
Add support for manually setting listen ports for int/extproxy

DmitryDodzin:


mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +28,12 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
 #[config(map_to = "ExternalProxyFileConfig", derive = "JsonSchema")]
 #[cfg_attr(test, config(derive = "PartialEq"))]
 pub struct ExternalProxyConfig {
+    /// <!--${internal}-->
+    ///
+    /// Whether to use a tls or plain connection
+    #[config(env = MIRRORD_EXTERNAL_TLS_ENABLE_ENV, default = true)]
```
aviramha, line 34: Can we drop the env?

DmitryDodzin: I want to use the env in the "automatic" script for the minikube execution instead of requesting the user to change his config

aviramha: Why do we need to disable TLS?

mirrord/config/src/external_proxy.rs
```diff
@@ -27,6 +28,12 @@ pub static MIRRORD_EXTERNAL_TLS_KEY_ENV: &str = "MIRRORD_EXTERNAL_TLS_KEY";
 #[config(map_to = "ExternalProxyFileConfig", derive = "JsonSchema")]
 #[cfg_attr(test, config(derive = "PartialEq"))]
 pub struct ExternalProxyConfig {
+    /// <!--${internal}-->
+    ///
+    /// Whether to use a tls or plain connection
```
Razz4780, line 33:
```suggestion
    /// Whether to use TLS or a plain TCP connection.
```

Razz4780 (approved):
Looks fine to me, but I second Aviram's question

PR #2795, opened 2024-10-02
Only set env vars in process to be executed locally

### LINKED ISSUE #2783: analytics fails on certificate access?
To reproduce:

```
RUST_BACKTRACE=full MIRRORD_PROGRESS_MODE=off RUST_LOG=mirrord=trace mirrord exec -t deployment/ip-visit-counter ./sha2.sh
```

Logs:

```
2024-09-27T15:20:07.198345Z TRACE start: mirrord_sip::main: Checking the SIP status of ./sha2.sh (or of the binary in its shebang, if applicable) failed with FileNotFound("./sha2.sh"). Continuing without SIP-sidestepping.This is not necessarily an error.
2024-09-27T15:20:07.199366Z ERROR mirrord: Couldn't execute Errno(Errno { code: 2, description: Some("No such file or directory") })
2024-09-27T15:20:07.199501Z TRACE mirrord::execution: watch_stderr finished
thread 'main' panicked at /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/reqwest-0.12.5/src/async_impl/client.rs:1901:38:
Client::new(): reqwest::Error { kind: Builder, source: Custom { kind: NotFound, error: "could not load certs from dir /etc/ssl/certs/ca-certificates.crt: No such file or directory (os error 2)" } }
stack backtrace:
   0:        0x105a7f294 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hd504cdf75dd790a5
   1:        0x105aa2eac - core::fmt::write::h5707604b140ad434
   2:        0x105a7b32c - std::io::Write::write_fmt::h4786975223352ad5
   3:        0x105a7f0ec - std::sys_common::backtrace::print::hbbeb9ed95a7b80ec
   4:        0x105a8082c - std::panicking::default_hook::{{closure}}::h9824320c6941fbee
   5:        0x105a8051c - std::panicking::default_hook::h9842bf3cae562040
   6:        0x105a810fc - std::panicking::rust_panic_with_hook::h83d90448f6646b96
   7:        0x105a80b18 - std::panicking::begin_panic_handler::{{closure}}::hf0c46de292f71ea2
   8:        0x105a7f724 - std::sys_common::backtrace::__rust_end_short_backtrace::h19b73a882a1ec51f
   9:        0x105a80884 - _rust_begin_unwind
  10:        0x105af4384 - core::panicking::panic_fmt::h5317ec8667b1cbec
  11:        0x105af4708 - core::result::unwrap_failed::h217ba6ef626fc10a
  12:        0x105475dfc - <reqwest::async_impl::client::Client as core::default::Default>::default::hf57e83cc782d88cb
  13:        0x1053f1928 - <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll::h814e37c1f5112c2e
  14:        0x1053ee23c - <mirrord_analytics::AnalyticsReporter as core::ops::drop::Drop>::drop::{{closure}}::h4de94e8426f1a031
  15:        0x1053eddc4 - tokio::runtime::task::core::Core<T,S>::poll::h03f9b3b40bf7b246
  16:        0x1053fbb50 - tokio::runtime::task::harness::Harness<T,S>::poll::hecb794b564fc5099
  17:        0x104d5a7cc - tokio::runtime::scheduler::current_thread::Context::enter::hdece761f632c71c0
  18:        0x104f0eff8 - tokio::runtime::context::scoped::Scoped<T>::set::hae9abe92c267b491
  19:        0x104e3dc5c - tokio::runtime::context::set_scheduler::h123108e7126cf435
  20:        0x104d6df38 - mirrord::main::heef16227dda6075d
  21:        0x104f4b0ec - std::sys_common::backtrace::__rust_begin_short_backtrace::h77d3657619d5e245
  22:        0x104ee9d20 - std::rt::lang_start::{{closure}}::h48d0474fc67780db
  23:        0x105a73a5c - std::rt::lang_start_internal::h9fea95fc6a17b76c
  24:        0x104d86468 - _main
```

([sha2.sh](http://sha2.sh) need not exist)

gememma:
Closes #2783

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -129,7 +129,7 @@ where
     let args = binary_args
         .clone()
         .into_iter()
-        .map(|s| CString::new(s))
+        .map(CString::new)
         .collect::<Result<Vec<_>, _>>()?;
     let env = env_vars
         .into_iter()
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -104,10 +104,7 @@ where
 
     // Collect environment variables curretn local vars and add those from agent + layer settings.
     let mut env_vars: HashMap<String, String> = vars().collect();
-    for (key, val) in execution_info.environment.clone() {
-        // configured vars should overwrite existing vars
-        let _ = env_vars.insert(key, val);
-    }
+    env_vars.extend(execution_info.environment.clone());
 
     for key in &execution_info.env_to_unset {
         env_vars.remove(key);
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -141,7 +141,7 @@ where
             // check if binary is present in each path
             path.push(binary.clone());
             if path.is_executable() {
-                resolved = path.to_string_lossy().into();
+                resolved = CString::new(path.as_ref().as_os_str().as_bytes())?;
                 break;
             }
         }
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -3,7 +3,7 @@
 #![feature(try_blocks)]
 #![warn(clippy::indexing_slicing)]
 
-use std::{collections::HashMap, net::SocketAddr, sync::LazyLock, time::Duration};
+use std::{collections::HashMap, ffi::CString, net::SocketAddr, sync::LazyLock, time::Duration};
 
 use clap::{CommandFactory, Parser};
 use clap_complete::generate;
@@ -11,7 +11,6 @@ use config::*;
 use connection::create_and_connect;
 use container::container_command;
 use diagnose::diagnose_command;
-use exec::execvp;
 use execution::MirrordExecution;
 use extension::extension_exec;
 use extract::extract_library;
@@ -35,6 +34,8 @@ use mirrord_intproxy::agent_conn::{AgentConnection, AgentConnectionError};
 use mirrord_kube::api::kubernetes::{create_kube_config, seeker::KubeResourceSeeker};
 use mirrord_operator::client::OperatorApi;
 use mirrord_progress::{messages::EXEC_CONTAINER_BINARY, Progress, ProgressTracker};
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+use nix::errno::Errno;
 use operator::operator_command;
 use port_forward::{PortForwardError, PortForwarder, ReversePortForwarder};
 use regex::Regex;
@@ -99,13 +100,11 @@ where
     // Stop confusion with layer
     std::env::set_var(mirrord_progress::MIRRORD_PROGRESS_ENV, "off");
 
-    // Set environment variables from agent + layer settings.
-    for (key, value) in &execution_info.environment {
-        std::env::set_var(key, value);
-    }
+    // Collect environment variables from agent + layer settings.
+    let mut env_vars = execution_info.environment.clone();
 
     for key in &execution_info.env_to_unset {
-        std::env::remove_var(key);
+        env_vars.remove(key);
     }
 
     let mut binary_args = args.binary_args.clone();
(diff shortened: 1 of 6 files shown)


aviramha:
> Doesn't this change mean that local environment variables will no longer be set for the local process? That doesn't seem desirable for me, I think it would break a lot of user applications.

How come?

t4lz:
> How come?

because with `execvp` the new process would inherit the environment variables from the mirrord binary's process, but after this change we call `execve` and pass an environment that looks to me like it only includes the remote environment, minus env_to_unset. Or am I missing where the local environment is added?

aviramha:
> > How come?
> 
> 
> 
> because with `execvp` the new process would inherit the environment variables from the mirrord binary's process, but after this change we call `execve` and pass an environment that looks to me like it only includes the remote environment, minus env_to_unset. Or am I missing where the local environment is added?

Excellent catch

gememma:
> Doesn't this change mean that local environment variables will no longer be set for the local process? That doesn't seem desirable for me, I think it would break a lot of user applications.

Yeah good point, the env vars should merge the vars from config into the local extant vars while overwriting conflicts - I expect this is why CI e2e is failing 🙃

mirrord/cli/src/main.rs
```diff
@@ -98,13 +99,15 @@ where
...
 
-    // Set environment variables from agent + layer settings.
-    for (key, value) in &execution_info.environment {
-        std::env::set_var(key, value);
+    // Collect environment variables curretn local vars and add those from agent + layer settings.
+    let mut env_vars: HashMap<String, String> = vars().collect();
+    for (key, val) in execution_info.environment.clone() {
+        // configured vars should overwrite existing vars
+        let _ = env_vars.insert(key, val);
     }
```
t4lz, line 107:
```suggestion
    env_vars.extend(execution_info.environment.clone());
```

mirrord/cli/src/main.rs
```diff
@@ -124,21 +131,49 @@ where
...
 
+    // since execvpe doesn't exist on macOS, resolve path manually and use execve
+    let binary_path: String = if let Some(path) = var_os("PATH") {
+        let mut resolved = binary.clone();
+        let paths = split_paths(&path).collect::<Vec<_>>();
+        for mut path in paths {
+            // check if binary is present in each path
+            path.push(binary.clone());
+            if path.is_executable() {
+                resolved = path.to_string_lossy().into();
```
t4lz, line 142: I think we could avoid the lossy conversion, if you want

```suggestion
                resolved = CString::new(path.as_ref().as_os_str().as_bytes())?;
```

mirrord/cli/src/main.rs
```diff
@@ -124,21 +131,49 @@ where
     );
     sub_progress_config.success(Some("config summary"));
 
+    // since execvpe doesn't exist on macOS, resolve path manually and use execve
+    let binary_path: String = if let Some(path) = var_os("PATH") {
```
t4lz, line 135: `PATH` based resolving should only happen if the given executable does not contain a `/` character.
From the execvp man page:
> The functions execlp(), execvp(), and execvP() will duplicate the actions of the shell in searching for an executable file if the specified file name does not contain a slash “/” character.  For execlp() and execvp(), search path is the path specified in the environment by “PATH” variable.  If this variable
     is not specified, the default path is set according to the _PATH_DEFPATH definition in <paths.h>, which is set to “/usr/bin:/bin”.

Also [`relibc`](https://github.com/redox-os/relibc/blob/a91cb17e4eaa82ebc807155285b3b661d8ebe0e5/src/header/unistd/mod.rs#L300) does that check.

Without that check, when executing a relative path, if that path also exists relatively to one of the paths in `PATH`, then that path would be executed, instead of the one relative to the current directory.

Regarding the last part from the docs quote - if `PATH` is not set they check relatively to /usr/bin and to /bin. Maybe you could find out what happens on shells (bash/zsh) and try to copy their behaviour. If running a certain path without mirrord works, we want it to also work with mirrord.

gememma: sounds good to me 🫡

PR #2797, opened 2024-10-04
Add default bash interpreter shebang to scripts without one

### LINKED ISSUE #2784: running shell scripts without shebang causes SIP detection to fail
Latest macOS update, Sequoia (15.0) broke running shell scripts using mirrord. Specifically, it looks like it broke the SIP detection somehow.

To reproduce create a script called fun.sh:
```sh
#/bin/sh
curl http://ip-info
```

then
 on 
`MIRRORD_PROGRESS_MODE=off RUST_LOG=mirrord=trace mirrord exec -t deployment/ip-visit-counter -- ./fun.sh`

you get 
```
2024-09-27T15:23:39.388168Z TRACE start: mirrord_sip::main: No SIP detected on "./sha.sh"
* Could not resolve host: ip-info
* Closing connection
curl: (6) Could not resolve host: ip-info
```

gememma:
Closes #2784

### DIFF
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -444,6 +444,7 @@ mod main {
 
     /// SIP check for binaries.
     fn is_binary_sip(path: &Path, patch_binaries: &[String]) -> Result<bool> {
+        dbg!(path, patch_binaries);
         // Patch binary if it is in the list of binaries to patch.
         // See `ends_with` docs for understanding better when it returns true.
         Ok(patch_binaries.iter().any(|x| path.ends_with(x))
@@ -480,7 +481,15 @@ mod main {
             return Ok(NoSip);
         }
 
-        if let Some(shebang) = read_shebang_from_file(&complete_path)? {
+        if let Ok(option) = read_shebang_from_file(&complete_path) {
+            let shebang = match option {
+                Some(shebang) => shebang,
+                None => ScriptShebang {
+                    interpreter_path: which("bash")?,
+                    start_of_rest_of_file: 0, /* FIX: this causes the first line of the script to
+                                               * be ignored */
+                },
+            };
             let interpreter_complete_path = get_complete_path(&shebang.interpreter_path)?;
             if is_in_mirrord_tmp_dir(&interpreter_complete_path)? {
                 return Ok(NoSip);
@@ -563,7 +572,7 @@ mod main {
     /// If it is not, `Ok(None)`.
     /// Propagate errors.
     pub fn sip_patch(binary_path: &str, patch_binaries: &[String]) -> Result<Option<String>> {
-        match get_sip_status(binary_path, patch_binaries) {
+        match dbg!(get_sip_status(binary_path, patch_binaries)) {
             Ok(SipScript { path, shebang }) => {
                 let patched_interpreter = patch_binary(&shebang.interpreter_path)?;
                 let patched_script = patch_script(
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -444,7 +444,6 @@ mod main {
 
     /// SIP check for binaries.
     fn is_binary_sip(path: &Path, patch_binaries: &[String]) -> Result<bool> {
-        dbg!(path, patch_binaries);
         // Patch binary if it is in the list of binaries to patch.
         // See `ends_with` docs for understanding better when it returns true.
         Ok(patch_binaries.iter().any(|x| path.ends_with(x))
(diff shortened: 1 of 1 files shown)


mirrord/sip/src/lib.rs
```diff
@@ -480,7 +480,15 @@ mod main {
             return Ok(NoSip);
         }
 
-        if let Some(shebang) = read_shebang_from_file(&complete_path)? {
+        if let Ok(option) = read_shebang_from_file(&complete_path) {
+            let shebang = match option {
+                Some(shebang) => shebang,
+                None => ScriptShebang {
+                    interpreter_path: which("bash")?,
```
aviramha, line 487: use `$SHELL`

mirrord/sip/src/lib.rs
```diff
@@ -480,7 +480,15 @@ mod main {
             return Ok(NoSip);
         }
 
-        if let Some(shebang) = read_shebang_from_file(&complete_path)? {
+        if let Ok(option) = read_shebang_from_file(&complete_path) {
+            let shebang = match option {
+                Some(shebang) => shebang,
+                None => ScriptShebang {
```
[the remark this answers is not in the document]
gememma: i think that non-text files will return an `Err` instead of an `Ok(Option)` here from `read_shebang_from_file()`, but maybe i'm wrong 👀 I'll test with binaries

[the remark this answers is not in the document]
gememma: you were correct, this breaks the expected distrinction between SipScript and SipBinary 🙃 needs an additional check

mirrord/sip/src/lib.rs
```diff
@@ -480,7 +486,28 @@ mod main {
...
+                if is_sip {
+                    SipBinary(complete_path)
+                } else {
+                    NoSip
+                }
+            })
+        } else {
+            // find shebang or use the default from $SHELL
+            let shebang = match read_shebang_from_file(&complete_path)? {
+                Some(shebang) => shebang,
```
DmitryDodzin, line 503: I'd do it with `.unwrap_or_else` instead of match statement here

mirrord/sip/src/lib.rs
```diff
@@ -480,7 +486,28 @@ mod main {
             return Ok(NoSip);
         }
 
-        if let Some(shebang) = read_shebang_from_file(&complete_path)? {
+        // determine if file is object file
+        let data = std::fs::read(&complete_path)?;
```
DmitryDodzin, line 490: We load the target binary in it's entirety twice into memory here because the data is only passed as ref and not consumed by it, I don't think this is an issue in most cases unless the target binary is huge for some reason (multiple hundreds of MB or even several GB). I would take a look at maybe changing `is_binary_sip ` & `is_code_signed` to use the `MachFile` directly instead of paths (or with both if necessary) instead of creating them inside of them

DmitryDodzin (changes requested):
I would update some apis to prevent double loading of target binary into memory to prevent any possible issues with large binaries right now and so it'll be more straightforward to change it if large binaries do turn out of being an issue + couple of nitpicks =]

mirrord/sip/src/lib.rs
```diff
@@ -8,7 +8,7 @@ mod rpath;
 
 mod main {
     use std::{
-        env,
+        env::{self},
```
DmitryDodzin, line 11:
```suggestion
        env,
```

mirrord/sip/src/lib.rs
```diff
@@ -443,11 +444,11 @@ mod main {
     }
 
     /// SIP check for binaries.
-    fn is_binary_sip(path: &Path, patch_binaries: &[String]) -> Result<bool> {
+    fn is_binary_sip(path: &Path, data: &Vec<u8>, patch_binaries: &[String]) -> Result<bool> {
```
DmitryDodzin, line 447:
```suggestion
    fn is_binary_sip(path: &Path, data: &[u8], patch_binaries: &[String]) -> Result<bool> {
```

mirrord/sip/src/lib.rs
```diff
@@ -419,12 +425,7 @@ mod main {
...
     /// Checks if binary is signed with either `RUNTIME` or `RESTRICTED` flags.
     /// The code ignores error to allow smoother fallbacks.
-    fn is_code_signed(path: &Path) -> bool {
-        let data = match std::fs::read(path) {
-            Ok(data) => data,
-            Err(_) => return false,
-        };
-
+    fn is_code_signed(data: &Vec<u8>) -> bool {
         if let Ok(mach) = MachFile::parse(data.as_ref()) {
```
DmitryDodzin, line 429:
```suggestion
    fn is_code_signed(data: &[u8]) -> bool {
        if let Ok(mach) = MachFile::parse(data) {
```

DmitryDodzin (approved):
LGTM =] Some tiny nitpicks

PR #2813, opened 2024-10-09
Add integration test, fix wrong warning

### LINKED ISSUE #2812: Add test for double bind on port 0
follow up on [https://github.com/metalbear-co/mirrord/pull/2811](https://github.com/metalbear-co/mirrord/pull/2811)

also found this false warning:

```
2024-10-09T13:52:17.702198Z  WARN ThreadId(03) bind: mirrord_layer::socket::ops: Port 0 was not included in the filtered ports, and also not in the non-filtered ports, and will therefore be bound locally. If this is intentional, ignore this warning. If you want the http filter to apply to this port, add it to `feature.network.incoming.http_filter.ports`. If you want to steal all the traffic of this port, add it to `feature.network.incoming.ports`. sockfd=59 address_length=28 pid=4038
```

t4lz:
Closes #2812

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -262,8 +262,8 @@ pub(super) fn bind(
     }
 
     // To handle #1458, we don't ignore port `0` for UDP.
-    if (is_ignored_tcp_port(&requested_address, incoming_config)
-        && matches!(socket.kind, SocketKind::Tcp(_)))
+    if (matches!(socket.kind, SocketKind::Tcp(_)))
+        && is_ignored_tcp_port(&requested_address, incoming_config)
         || crate::setup().is_debugger_port(&requested_address)
         || incoming_config.ignore_ports.contains(&requested_port)
     {
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -24,7 +24,7 @@ use mirrord_protocol::{
     dns::{GetAddrInfoRequest, LookupRecord},
     file::{OpenFileResponse, OpenOptionsInternal, ReadFileResponse},
 };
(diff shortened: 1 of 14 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -258,8 +258,8 @@ pub(super) fn bind(
     }
 
     // To handle #1458, we don't ignore port `0` for UDP.
-    if (is_ignored_tcp_port(&requested_address, incoming_config)
```
aviramha, line 261: why?

aviramha: oh I realize now, but that's not a solution -this warning can also happen with tcp.. (just more rare)

aviramha: I think it should still fail if it was ignored, since we try to match local to remote ports when it comes to binding..

t4lz: > I think it should still fail if it was ignored, since we try to match local to remote ports when it comes to binding..

I don't understand @aviramha

aviramha: Perhaps clarify your original intent since I think we might not be talking about the same thing

t4lz: Nevermind. Is [this](https://github.com/metalbear-co/mirrord/pull/2813/commits/a8e3dee2d12d3b7d2af466f821e54190d0e8a2fa) a solution? - moved the warning to `listen` so that we don't show it if the user app binds a port that is ignored for incoming but then uses it for outgoing.

aviramha: I'm okay with this, would get another review tho.

Razz4780 (approved):
:+1:

PR #2808, opened 2024-10-09
Allow using IPv6 sockets with mirrord

aviramha:
Closes #2807

PR #2829, opened 2024-10-11
Handle IPv4 in IPv6, fix regressions

aviramha:
Should fix any regression caused by https://github.com/metalbear-co/mirrord/pull/2808


meowjesty (approved):
It's canon now.

ps. Could you add a module docs in `socket.rs` talking about the need to use `to_canonical`, mentioning the ipv4 in ipv6 and how we need this to have ipv6 not break our ipv4 `loopback`.

Issue #2287, opened 2024-03-04
DNS failure because lack of privileged container


aviramha:
A user encountered a case where DNS resolution failed because the agent lacked the `privileged` flag to access the target container's `/etc/resolv.conf`. 
In this line we just propagate the error, converting it to unknown dns failure error. I'm not sure how we can improve it but it'd be nice if we could check for permission denied then let user know somehow.
https://github.com/metalbear-co/mirrord/blob/879fd097bd5949bf694f8f6610900139323e1f33/mirrord/agent/src/dns.rs#L71

meowjesty:
Bit of a better error was added, hopefully it's friendlier to users now.

PR #2830, opened 2024-10-11
Add error log when `resolv.conf` fails to `open` due to `PermissionDenied`.

meowjesty:
- Issue: #2287 

Is this what we want to warn the user?


aviramha:
I can understand how this solution was thought of but the function that actually fails is the getaddrinfo in the agent side - afaik when this happens it's not even trying to open the file.
Ideally we would like to have the agent notify when such error happens as part of resolution

meowjesty:
Now we get a nice error on the layer side:

```
ERROR ThreadId(02) remote_getaddrinfo: mirrord_layer::socket::ops: error=Failed performing `getaddrinfo` with: the agent could not find a DNS related file, such as `/etc/resolv.conf` or `/etc/hosts`! node="www.mirrord.dev"
```

p.s.: This error is for `NotFound`, as I was using it to trigger an issue in the `fs::read` call (easier than changing permissions).

mirrord/agent/src/dns.rs
```diff
@@ -75,7 +76,7 @@ impl DnsWorker {
     ///
     /// We could probably cache results here.
     /// We cannot cache the [`AsyncResolver`] itself, becaues the configuration in `etc` may change.
-    #[tracing::instrument(level = "trace")]
+    #[tracing::instrument(level = Level::TRACE, ret, err)]
```
Razz4780, line 79: This should not be a global agent error imho. User puts bad url in their app -> error spam in the agent. I'd log error when we fail to build the resolver though. You can put building the resolver in a `try` block (excluding only the lookup itself) and log the error returned from that block

mirrord/agent/src/dns.rs
```diff
@@ -110,12 +111,16 @@ impl DnsWorker {
...
+    #[tracing::instrument(level = Level::TRACE, skip(self))]
     fn handle_message(&self, message: DnsCommand) {
         let etc_path = self.etc_path.clone();
         let timeout = self.timeout;
         let attempts = self.attempts;
         let lookup_future = async move {
-            let result = Self::do_lookup(etc_path, message.request.node, attempts, timeout).await;
+            let result = Self::do_lookup(etc_path, message.request.node, attempts, timeout)
+                .await
+                .inspect_err(|fail| tracing::error!(?fail, "DNS lookup failed!"));
```
Razz4780, line 122:
```suggestion
                .inspect_err(|fail| tracing::trace!(?fail, "DNS lookup failed!"));
```
I wouldn't log on error level here, resolver build failure should already be logged internally in `do_lookup`. Lookup errors themselves could be spammy

mirrord/layer/src/error.rs
```diff
@@ -41,7 +41,7 @@ mod ignore_codes {
 /// These errors are converted to [`libc`] error codes, and are also used to [`set_errno`].
 #[derive(Error, Debug)]
 pub(crate) enum HookError {
-    #[error("mirrord-layer: Failed while getting a response!")]
+    #[error(transparent)]
```
Razz4780, line 44: All other variants have `mirrord-layer` prefix :X

mirrord/layer/src/error.rs
```diff
@@ -126,7 +126,7 @@ pub(crate) enum HookError {
 /// handling [`ProxyToLayerMessage`](mirrord_intproxy_protocol::ProxyToLayerMessage).
 #[derive(Error, Debug)]
 pub(crate) enum LayerError {
-    #[error("mirrord-layer: Failed while getting a response!")]
+    #[error(transparent)]
```
Razz4780, line 129: All other variants have `mirrord-layer` prefix :X

mirrord/protocol/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mirrord-protocol"
-version = "1.11.4"
+version = "1.11.5"
```
Razz4780, line 3: Needs an update I think, version is now different on main

mirrord/protocol/src/error.rs
```diff
@@ -151,13 +151,17 @@ pub struct RemoteIOError {
 
 /// Our internal version of Rust's `std::io::Error` that can be passed between mirrord-layer and
 /// mirrord-agent.
+///
+/// [`ResolveErrorKindInternal`] has a nice [`core::fmt::Display`] implementation that
+/// should be user friendly, and can be appended to the generic error message here.
 #[derive(Encode, Decode, Debug, PartialEq, Clone, Eq, Error)]
-#[error("Failed performing `getaddrinfo` with {kind:?}!")]
+#[error("Failed performing `getaddrinfo` with: {kind}!")]
```
Razz4780, line 158:
```suggestion
#[error("Failed performing `getaddrinfo`: {kind}")]
```

mirrord/protocol/src/error.rs
```diff
@@ -151,13 +151,17 @@ pub struct RemoteIOError {
...
 pub struct DnsLookupError {
     pub kind: ResolveErrorKindInternal,
 }
 
 impl From<io::Error> for ResponseError {
+    #[tracing::instrument(level = Level::TRACE, ret)]
```
Razz4780, line 164: Looks unnecessary :v

mirrord/protocol/src/error.rs
```diff
@@ -167,6 +171,7 @@ impl From<io::Error> for ResponseError {
 }
 
 impl From<ResolveError> for ResponseError {
+    #[tracing::instrument(level = Level::TRACE, ret)]
```
Razz4780, line 174: Looks unnecessary :v

mirrord/protocol/src/error.rs
```diff
@@ -233,6 +238,33 @@ pub enum ResolveErrorKindInternal {
...
+
+impl core::fmt::Display for ResolveErrorKindInternal {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            ResolveErrorKindInternal::Message(message) => write!(f, "{message}"),
+            ResolveErrorKindInternal::NoConnections => write!(f, "no connections"),
+            ResolveErrorKindInternal::NoRecordsFound(records) => {
+                write!(f, "no records found {records}")
+            }
+            ResolveErrorKindInternal::Proto => write!(f, "protocol"),
```
Razz4780, line 253:
```suggestion
            ResolveErrorKindInternal::Proto => write!(f, "protocol error"),
```

mirrord/protocol/src/error.rs
```diff
@@ -233,6 +238,33 @@ pub enum ResolveErrorKindInternal {
...
+            ResolveErrorKindInternal::Timeout => write!(f, "timeout"),
+            ResolveErrorKindInternal::Unknown => write!(f, "unknown"),
```
Razz4780, line 255:
```suggestion
            ResolveErrorKindInternal::Unknown => write!(f, "unknown error"),
```

Razz4780 (changes requested):
Just nits around tracing/messages

mirrord/layer/src/error.rs
meowjesty: Re-added it, but I think we should look into our error messages, while trying this out I was getting some terrible ones, along the lines of:

`mirrord-layer: Failed to get Response due to: Failed to DNS something with: KubeError failed { kind: Thing, error: UnknownSomething }`.

We should probably prepare a good initial message for the inner error, and just keep everything else `transparent` (for our internal errors), example:

```rust
#[error("We have an io error: {0}!")]
IO(#[from] std::IO),

#[error(transparent)]
Response(#[from] protocol::ResponseError),
```

mirrord/agent/src/dns.rs
meowjesty: Alright, for some reason the `try` block approach did not work. Rust could not figure out the type, and I tried putting the type in, both on the `let` side and on the `Ok` side, still kept complaining.

Using an `async` closure to the same effect.

mirrord/layer/src/error.rs
Razz4780: Yeah these error messages require some extra care

mirrord/agent/src/dns.rs
Razz4780: Tadaa. This compiles

```rust
    #[tracing::instrument(level = Level::TRACE, ret, err(level = Level::TRACE))]
    async fn do_lookup(
        etc_path: PathBuf,
        host: String,
        attempts: usize,
        timeout: Duration,
    ) -> RemoteResult<DnsLookup> {
        let resolver: Result<_, ResponseError> = try {
            let resolv_conf_path = etc_path.join("resolv.conf");
            let hosts_path = etc_path.join("hosts");

            let resolv_conf = fs::read(resolv_conf_path).await?;
            let hosts_conf = fs::read(hosts_path).await?;

            let (config, mut options) = parse_resolv_conf(resolv_conf)?;
            options.server_ordering_strategy =
                hickory_resolver::config::ServerOrderingStrategy::UserProvidedOrder;
            options.timeout = timeout;
            options.attempts = attempts;
            options.ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4Only;

            let mut resolver = AsyncResolver::tokio(config, options);

            let hosts = Hosts::default().read_hosts_conf(hosts_conf.as_slice())?;
            resolver.set_hosts(Some(hosts));

            resolver
        };

        let lookup = resolver
            .inspect_err(|error| tracing::error!(%error, "failed to build DNS resolver"))?
            .lookup_ip(host)
            .await
            .inspect(|lookup| tracing::trace!(?lookup, "Lookup finished"))?
            .into();

        Ok(lookup)
    }
```

Razz4780 (approved):
Noice. I recommend using the try block tho

mirrord/agent/src/dns.rs
meowjesty: Thank you!

Changed.

PR #2832, opened 2024-10-13
Update github actions dependencies

aviramha:


.github/workflows/ci.yaml
```diff
@@ -154,8 +154,8 @@ jobs:
           toolchain: nightly-2024-04-15
           components: rustfmt, clippy
           target: aarch64-unknown-linux-gnu,x86_64-unknown-linux-gnu
-      - uses: actions/setup-python@v3 # For http mirroring tests with Flask and FastAPI.
-      - run: pip3 install cargo-zigbuild # For http mirroring test with Flask.
+      - uses: actions/setup-python@v5 # For http mirroring tests with Flask and FastAPI.
+      - run: pip3 install --break-system-packages cargo-zigbuild # For http mirroring test with Flask.
```
DmitryDodzin, line 158: Is maybe switching to uv better in this case like for towncrier instead of setting `--break-system-packages`?

.github/workflows/ci.yaml
```diff
@@ -471,10 +471,8 @@ jobs:
           candidate: java
           version: 17.0.6-tem
       - run: java -version
-      - uses: actions/setup-python@v3 # For http mirroring tests with Flask and FastAPI.
-      - run: pip3 install flask # For http mirroring test with Flask.
-      - run: pip3 install fastapi # For http mirroring test with FastAPI.
-      - run: pip3 install uvicorn[standard] # For http mirroring test with FastAPI.
+      - uses: actions/setup-python@v5 # For http mirroring tests with Flask and FastAPI.
+      - run: pip3 install --break-system-packages flask fastapi uvicorn[standard] # For http mirroring test with Flask.
```
DmitryDodzin, line 475: Same here uv?

.github/workflows/ci.yaml
```diff
@@ -288,26 +288,26 @@ jobs:
         with:
           node-version: 14
       - run: npm install express # For http mirroring test with node.
-      - uses: actions/setup-python@v3 # For http mirroring tests with Flask and FastAPI.
-      - run: pip3 install flask fastapi uvicorn[standard] # For http mirroring test with Flask.
+      - uses: actions/setup-python@v5 # For http mirroring tests with Flask and FastAPI.
+      - run: pip3 install --break-system-packages flask fastapi uvicorn[standard] # For http mirroring test with Flask.
```
DmitryDodzin, line 292: Same here uv?

.github/workflows/ci.yaml
aviramha: problem is it's quite hard to make everything use uv (requires a venv and activating it..)

DmitryDodzin (approved):

PR #2834, opened 2024-10-13
Add nodeSelector option to agent config

dbeilin:
Hello, this is my first PR so please go easy on me.
When I run `mirrord exec` I missed the option to tell the agent on which node to spawn, so I added the functionality after seeing that `labels` and `annotations` were already supported (since they're more or less the same, except node selector sits under the `spec` key).

I tested it on my own cluster using this config JSON:

```json
{
  "agent": {
    "node_selector":
      {
        "node1": "label1"
      }
  }
}
```

I was interested in adding this option in the CLI too, but this seemed like a quicker proof-of-concept. If this is something you're willing to merge, I can work on the CLI version as well.

Let me know what you think 😄

### DIFF
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -329,7 +329,8 @@ pub struct AgentConfig {
 
     /// ### agent.node_selector {#agent-node_selector}
     ///
-    /// Allows setting up custom node selector for the agent Job and Pod.
+    /// Allows setting up custom node selector for the agent Pod. Applies only to targetless runs,
+    /// as targeted agent always runs on the same node as its target container.
     ///
     /// ```json
     /// {
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -328,10 +328,12 @@ pub struct AgentConfig {
     pub annotations: Option<HashMap<String, String>>,
 
     /// ### agent.node_selector {#agent-node_selector}
-    /// Allows setting up node selector for the agent Job and Pod.
+    ///
+    /// Allows setting up custom node selector for the agent Job and Pod.
+    ///
     /// ```json
     /// {
-    ///  "node_selector": { "kubernetes.io/hostname": "node1" }
+    ///   "node_selector": { "kubernetes.io/hostname": "node1" }
     /// }
     /// ```
     pub node_selector: Option<HashMap<String, String>>,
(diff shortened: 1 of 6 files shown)


aviramha:
Thanks for sending this!
We'd prefer not to expose it from cli, as we want to keep the cli api clean and concise.

mirrord/kube/src/api/container/job.rs
```diff
@@ -165,9 +165,19 @@ where
...
         ]));
 
+        let node_selector = config
+            .node_selector
+            .clone()
+            .map(BTreeMap::from_iter)
+            .unwrap_or_default();
+
         pod.labels_mut().extend(labels.clone());
         pod.annotations_mut().extend(annotations.clone());
```
aviramha, line 175: run `cargo fmt --all`

dbeilin:
@aviramha Took a while to figure out the `schema_file_is_up_to_date` test. I've been fighting with with for a while until I realized my VS Code auto-saves and also auto-adds white-spaces to all files, which caused the `mirrord-schema.json` to always have a diff 😩 

Let me know if anything else is missing.

Thanks

mirrord/config/src/agent.rs
```diff
@@ -327,6 +327,17 @@ pub struct AgentConfig {
     /// ```
     pub annotations: Option<HashMap<String, String>>,
 
+    /// ### agent.node_selector {#agent-node_selector}
+    ///
+    /// Allows setting up custom node selector for the agent Job and Pod.
```
Razz4780, line 332:
```suggestion
    /// Allows setting up custom node selector for the agent Pod. Applies only to targetless runs, as targeted agent always runs on the same node as its target container.
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -165,9 +165,19 @@ where
...
 
+        if let Some(spec) = pod.spec.as_mut() {
+            spec.node_selector = Some(node_selector);
+        }
+
```
Razz4780, line 180: I'd move this injection to `mirrord-kube::api::container::pod::PodVariant`, so that it also works in mirrord for Teams (we don't use jobs there)

Razz4780 (changes requested):

dbeilin:
@aviramha In addition to this PR, I believe adding the option of choosing a service account would be beneficial as well. I'm willing to give it another go and open a PR, if that's something you'd approve of.

Issue #2836, opened 2024-10-14
Rollback allowing AF_INET6


aviramha:
Closes #2828 
Closes #2835 
Closes #2820

PR #2837, opened 2024-10-14
Disallow using IPv6 sockets with mirrord

DmitryDodzin:
relevant for #2836


aviramha:
just to make sure, we want to rollback all attempts to fix regression as well (as I fear those might have introduced even more issues ;( )

aviramha (approved):

PR #2841, opened 2024-10-15
Better error display for mirrord-container sidecar

DmitryDodzin:
relevant for #2726


gememma:
@DmitryDodzin do you have any before/ after examples?

DmitryDodzin:
> @DmitryDodzin do you have any before/ after examples?

Well right now any error there is a high likelihood that the container logs will be lost and only this error will be shown
```
Error:   × Comand failed to execute command [docker logs be6fe8a8bcc8bb8f721851fb5fd2c560666c0b03261f97c0db0dd94f74bf0e1f]: Error response from daemon: No such
  │ container: be6fe8a8bcc8bb8f721851fb5fd2c560666c0b03261f97c0db0dd94f74bf0e1f
```
this happens because the sidecar is deleted (due to `--rm` flag) before the `<runtime> logs <container_id>` is executed

This changes the command to use `<runtime> logs -f <container_id>` so the output can be something like
```
Error:   × Command output indicates an error [podman logs -f 7a3aaf9257b89b785074195b9a2bcdf85cb0f2e1895cbfc31e0815c40431f988]: thread 'main' panicked at mirrord/cli/src/internal_proxy.rs:130:5:
  │ test panic
  │ note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  │
  help: This is a bug. Please report it in our Discord or GitHub repository.
```

gememma (approved):
Looks really nice 👍

Issue #2807, opened 2024-10-09
listening on ::1:8080 fails - node macOS


aviramha:
```
Webpack: Starting ...
 INFO  Starting development server...
 ERROR  Error: listen EAFNOSUPPORT: address family not supported ::1:8080
Error: listen EAFNOSUPPORT: address family not supported ::1:8080
    at Server.setupListenHandle [as _listen2] (node:net:1886:21)
    at listenInCluster (node:net:1965:12)
    at node:net:2171:7
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

aviramha:
How to reproduce this issue:
```
HOST=::1 /Users/aviramhassan/Code/mirrord/target/universal-apple-darwin/debug/mirrord exec -t deployment/ip-visit-counter -- npm run serve
New mirrord version available: 3.120.0. To update, run: `"curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash"`.
To disable version checks, set env variable MIRRORD_CHECK_VERSION to 'false'.
* Running binary "npm" with arguments: ["run", "serve"].
* mirrord will target: deployment/ip-visit-counter, no configuration file was loaded
* operator: the operator will be used if possible
* env: all environment variables will be fetched
* fs: file operations will default to read only from the remote
* incoming: incoming traffic will be mirrored
* outgoing: forwarding is enabled on TCP and UDP
* dns: DNS will be resolved remotely
⠲ mirrord exec
    ✓ update to 3.120.0 available
    ✓ ready to launch process
      ✓ layer extracted
      ✓ using operator
        ✓ operator version compatible
        ✓ operator license valid
        ✓ user credentials prepared
        ✓ session started
          ✓ connected to the target
    ✓ config summary                                                            
> my-project@0.1.0 serve
> vue-cli-service serve

 INFO  Starting development server...
 ERROR  Error: listen EAFNOSUPPORT: address family not supported ::1:80
Error: listen EAFNOSUPPORT: address family not supported ::1:80
    at Server.setupListenHandle [as _listen2] (node:net:1734:21)
    at listenInCluster (node:net:1799:12)
    at doListen (node:net:1948:7)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21)

```
this issue is probably because of remote environment having `HOST`. need to exclude it

aviramha:
[...]

PR #2842, opened 2024-10-15
Add `getifaddrs` hook that removes ipv6 addresses

Razz4780:
Issue #2807

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -1615,6 +1615,7 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
     // I don't see any reason why we could not do this,
     // I also did not find anything suspicious in glibc [implementation](https://github.com/lattera/glibc/blob/master/sysdeps/gnu/ifaddrs.c).
     if !ipv6_head.is_null() {
+        // Safety: we constructed a valid `libc::ifaddrs` list using pointers received from libc.
         unsafe { libc::freeifaddrs(ipv6_head) };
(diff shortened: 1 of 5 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -1568,3 +1568,80 @@ pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
 
     Detour::Success(config)
 }
+
+/// Calls [`libc::getifaddrs`] and removes IPv6 addresses from the list.
+#[mirrord_layer_macro::instrument(level = "trace", ret, err)]
```
meowjesty, line 1573:
```suggestion
#[mirrord_layer_macro::instrument(level = Level::TRACE, ret, err)]
```

mirrord/layer/tests/apps/issue2807.js
```diff
@@ -0,0 +1,5 @@
+const portfinder = require('portfinder')
```
meowjesty, line 1: Nit: but can we make this an `.mjs`? Most of our js tests are `mjs`.

meowjesty (approved):
Only have a couple of nits.

We could use `NonNull`, but I don't think it's worth the hassle, the `.as_mut` already checks `null` for us so ... Approving.

Razz4780:
> Only have a couple of nits.
> 
> We could use `NonNull`, but I don't think it's worth the hassle, the `.as_mut` already checks `null` for us so ... Approving.

`NonNull` works on arrays of pointers, no? This is a linked list

Issue #2825, opened 2024-10-11
mirrord overrides kube config's 


aviramha:
Follow up on https://github.com/metalbear-co/mirrord/issues/2821
User had no `accept_invalid_certificates`, kubectl works, mirrord doesn't
I assume is because of 
```rs
    config.accept_invalid_certs = accept_invalid_certificates;
```
we probably need to allow null, remove it from the auto generated config (in intellij, vscode, and mirrord) and set it when it's not null only.

EDIT: this setting is not present in auto generated configs, only parsing needs an update

PR #2844, opened 2024-10-16
Don't override kubeconfig `insecure-skip-tls-verify` option when `accept_invalid_certificates` is not set

Razz4780:
Issue #2825


meowjesty (approved):
:+1:

Issue #2824, opened 2024-10-11
Detect invalid certificate error and print better message


aviramha:
Follow up on https://github.com/metalbear-co/mirrord/issues/2821
failure was:
```
⠤ mirrord exec
    ✓ running on latest (3.119.1)!
⠖ mirrord exec
    ✓ running on latest (3.119.1)!
  ⠚ preparing to launch process
  x mirrord exec
    ✓ running on latest (3.119.1)!
    x preparing to launch process
      ✓ layer extracted
      ✓ operator disabled                                                                                                                                                                     Error:   × Failed to create mirrord-agent: Kube failed: ServiceError: client error (Connect)
  help: 1. Please check the status of the agent pod, using `kubectl get pods` in the relevant namespace.
                2. If you don't see any `mirrord-agent-[...]` pods, then try running `kubectl get jobs`, and `kubectl describe mirrord-agent-[...]`.
                Make sure it is able to fetch the agent image, it didn't fail due to lack of resources, etc.
        
        - If you're still stuck:
        
        >> Please open a new bug report at https://github.com/metalbear-co/mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help in #mirrord-help
        
        >> Or email us at hi@metalbear.co
```

but in logs we could see that it was invalid certificate
```
 DEBUG start:create_and_connect: mirrord::connection: Failed to detect OpenShift error=KubeError(Service(hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) } })))

```
need to improve error experience here.

PR #2852, opened 2024-10-16
Better error for invalid certificate.

meowjesty:
- Issue: #2824 

Improves the error logs, should be more specific now in case of an `InvalidCertificate`.


Razz4780 (changes requested):
I'd move detecting invalid cert error to `CliError`. It offers better support for help messages/fix instructions. See `CliError::KubeAuthExecFailed` and `CliError::auth_exec_error_or`. I'd inject new logic into this function

meowjesty:
Moved the error.

This error feels like it's nested deep into some `box`es, and I'm not sure how to reproduce the issue to get the error, that's why I'm a bit unsure on how to go about `downcast`. Do you know how to get an invalid cert error, so I can change it to `downcast` instead of checking if error `contains` a string?

mirrord/kube/src/error.rs
```diff
@@ -7,7 +7,9 @@ pub type Result<T, E = KubeApiError> = std::result::Result<T, E>;
 
 #[derive(Debug, Error)]
 pub enum KubeApiError {
-    #[error("Kube failed: {0}")]
+    /// We manually implement `From<kube::Error>` to give a better error in case of
+    /// kube failing due to an invalid certificate.
+    #[error(transparent)]
```
Razz4780, line 12: No longer relevant

mirrord/cli/src/error.rs
```diff
@@ -400,7 +409,12 @@ impl CliError {
 
         match error {
             KubeApiError::KubeError(Error::Auth(AuthError::AuthExec(error))) => {
-                Self::KubeAuthExecFailed(error)
+                Self::KubeAuthExecFailed(error.to_owned())
+            }
+            KubeApiError::KubeError(Error::Service(ref fail))
+                if fail.to_string().contains("InvalidCertificate") =>
+            {
+                Self::InvalidCertificate(error)
```
Razz4780, line 417: I'd change function name now, it's no longer just `auth_exec_error_or`. Maybe sth like `detect_known_error_or_else`? Maybe sth shorter :v

mirrord/cli/src/error.rs
```diff
@@ -187,6 +187,15 @@ pub(crate) enum CliError {
     ))]
     AgentConnectionFailed(KubeApiError),
 
+    /// Friendlier version of the invalid certificate error that comes from a
+    /// [`kube::Error::Service`].
+    #[error("Kube API operation failed due to missing or invalid certificate: {0}")]
+    #[diagnostic(help(
+        r"1. Consider enabling `accept_invalid_certificates` in your `mirrord.json`, or;
+          2. Running `mirrord exec` with the `-c` flag."
+    ))]
```
Razz4780, line 196:
```suggestion
    #[diagnostic("Consider enabling `accept_invalid_certificates` in your `mirrord.json` or running `mirrord exec` with the `-c` flag."))]
```

Razz4780 (changes requested):

Razz4780:
> Moved the error.
> 
> This error feels like it's nested deep into some `box`es, and I'm not sure how to reproduce the issue to get the error, that's why I'm a bit unsure on how to go about `downcast`. Do you know how to get an invalid cert error, so I can change it to `downcast` instead of checking if error `contains` a string?

I'd try something like this in the test:
1. Setup TLS server with a randomly generated self-signed certificate
2. Pass its address as `cluster_url` in [kube::Config](https://docs.rs/kube/latest/kube/struct.Config.html)
3. Create a kube client and try listing pods

mirrord/cli/src/error.rs
```diff
@@ -462,3 +480,144 @@ impl From<OperatorApiError> for CliError {
...
+            assert!(format!("{fail:?}").contains("InvalidCertificate"),
+                "We were expecting this error {fail:?} to have `InvalidCertificate` but it's {fail}!"
+            );
+        }
+    }
+
+    /// Creates an [`hyper`] server with a broken certificate and a _catch-all_ route on
+    /// `localhost:9669`.
+    ///
+    /// The `.pem` files live in `/mirrord/tests/issue-2824-certs`.
```
Razz4780, line 564: Can you maybe generate this cert dynamically? :eyes:

Razz4780 (approved):
I'd generate the test cert at runtime, but file is also fine by me

mirrord/cli/src/error.rs
meowjesty: Sure, but why? I grabbed this cert from the hyper examples repo.

Razz4780: Dunno, just a suggestion. I don't like random files in the repo. But I don't insist ^^

meowjesty: Changed!

PR #2853, opened 2024-10-16
Added support for Istio CNI

### LINKED ISSUE #2851: istio mTLS causes issue
Mirantis Kubernetes Engine k8s 1.24
istio mesh with istio-cni

> Health checks are received well and plain text by mirrord, but it also receives some TLS traffic that looks destined to the mTLS/istio (i.e before TLS termination)

aviramha:
Closes #2851
Scenarios tested:

1. exec into pod and curl localhost -> stolen
2. via istiogateway -> stolen
3. port forward to pod -> stolen

### DIFF
--- a/mirrord/kube/src/api/container.rs
+++ b/mirrord/kube/src/api/container.rs
@@ -112,6 +112,7 @@ pub fn check_mesh_vendor(pod: &Pod) -> Option<MeshVendor> {
     const ISTIO: [&str; 2] = ["istio-proxy", "istio-init"];
     const LINKERD: [&str; 2] = ["linkerd-proxy", "linkerd-init"];
     const KUMA: [&str; 2] = ["kuma-sidecar", "kuma-init"];
+    const ISTIO_CNI: [&str; 2] = ["istio-proxy", "istio-validation"];
 
     if pod
         .metadata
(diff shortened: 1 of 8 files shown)


mirrord/kube/src/api/container.rs
```diff
@@ -125,18 +126,33 @@ pub fn check_mesh_vendor(pod: &Pod) -> Option<MeshVendor> {
...
+                .iter(),
+        )
+        .map(|status| status.name.as_str())
+        .collect::<Vec<&str>>();
+
+    // check that all the containers are present
+    // we had a case where istio cni was detected as istio while
+    // the init was only present.
+    // leave old logic for detection to not break existing setups
+    if ISTIO_CNI.iter().all(|name| container_names.contains(name)) {
```
DmitryDodzin, line 145: Is there a reason for contains? like are the container names different from the lookup? (container named `my-custom-istio-validation` will pass this filter) I'd maybe do a straight up compare

DmitryDodzin (approved):
LGTM, have a nitpick on container name specificity

mirrord/kube/src/api/container.rs
aviramha: Nice catch!

aviramha: Ah sorry, `container_names` is a Vec of names

aviramha: meaning, it does exact match

PR #2858, opened 2024-10-17
No arm64e in fat layer binary

### LINKED ISSUE #2846: use proper arm64e
until now, we had a patch that generates a "dummy" arm64e by just patching the architecture of our arm64 binary to be recognized as arm64e.
It worked on most occasions, because arm64 binaries can load arm64e libraries, and arm64e binaries are usually protected so they won't load us anyway.
When we upgraded to newer rust #2826 it caused a regression that crashed binaries such as dlv
so we downgraded to https://github.com/metalbear-co/mirrord/pull/2845

that lead me to investigate further, finding that the issue is with our fake arm64e:
```
➜  report DYLD_INSERT_LIBRARIES=/Users/aviramhassan/Code/mirrord/target/aarch64-apple-darwin/debug/libmirrord_layer.dylib /Users/aviramhassan/go/bin/dlv 
thread '<unnamed>' panicked at mirrord/layer/src/lib.rs:229:10:
missing internal proxy address
mirrord layer setup panicked
➜  report DYLD_INSERT_LIBRARIES=/Users/aviramhassan/Code/mirrord/target/aarch64-apple-darwin/debug/libmirrord_layer_arm64e.dylib /Users/aviramhassan/go/bin/dlv 
dyld[3826]: bad bind opcode 0x00
[1]    3826 abort      DYLD_INSERT_LIBRARIES= /Users/aviramhassan/go/bin/dlv
```

in the above example, if we use plain arm64, we manage to load. if we use arm64e, we crash.
macOS prefers arm64e libraries, so our fat binary makes it prefer always arm64e. The reason we need arm64e is debugserver which is the exception - it's an arm64e binary that we can load into, and if the fat binary lacks arm64e it crashes.

My proposed solution is to create a C shim (Since Rust can't yet compile arm64e properly, also lack of CI, etc) that loads our arm64 always.
Perhaps have a check if it's in arm64e mode really, then if not only load.

The cli will actually include 2 dylibs:
1. x64 (real layer), arm64e - fake layer
2. arm64 layer

When executing, it'll extract arm64 on arm builds then set environment variable to it, arm64e fake layer will load, find that env and load that if it matches it's conditions.

t4lz:
Closes #2846 (but instead of using a proper arm64e, we're not using any arm64e binary at all).

### DIFF
--- a/.github/workflows/arch_dylib_sign.json
+++ b/.github/workflows/arch_dylib_sign.json
@@ -1,15 +1,14 @@
 {
-    "source": [
-        "target/aarch64-apple-darwin/release/libmirrord_layer.dylib",
-        "target/aarch64-apple-darwin/release/libmirrord_layer_arm64e.dylib",
-        "target/x86_64-apple-darwin/release/libmirrord_layer.dylib"
-    ],
-    "bundle_id": "com.metalbear.mirrord",
-    "apple_id": {
-        "username": "@env:AC_USERNAME",
-        "password": "@env:AC_PASSWORD"
-    },
-    "sign": {
-        "application_identity": "Developer ID Application: METALBEAR TECH LTD (8W42TQ6PFA)"
-    }
+  "source": [
+    "target/aarch64-apple-darwin/release/libmirrord_layer.dylib",
+    "target/x86_64-apple-darwin/release/libmirrord_layer.dylib"
+  ],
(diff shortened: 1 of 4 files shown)


t4lz:
See https://github.com/metalbear-co/mirrord/issues/2846#issuecomment-2419632439

PR #2872, opened 2024-10-25
Change build_fat_mac.sh script to use arm64e shim

gememma:


mirrord/cli/build.rs
```diff
@@ -5,4 +5,13 @@ fn main() {
             std::env::var("CARGO_CDYLIB_FILE_MIRRORD_LAYER").unwrap()
         );
     };
+
+    // TODO: is it better to fail here or pass in cargo layer file?
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    if std::env::var("MIRRORD_MACOS_ARM64_LIBRARY").is_err() {
```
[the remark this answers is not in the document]
gememma: yeah it triggers if you try to build for aarch64-apple-darwin target without the var - it works fine without this whole if statement as long as you dont try to build the layer for aarch without the variable but i would prefer a nicer error message than the one rust spits out:

```
   Compiling mirrord v3.121.1 (/Users/gem/source/mirrord/mirrord/cli)
error: environment variable `MIRRORD_MACOS_ARM64_LIBRARY` not defined at compile time
  --> mirrord/cli/src/extract.rs:84:31
   |
84 |     let extension = Path::new(env!("MIRRORD_MACOS_ARM64_LIBRARY"))
   |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: use `std::env::var("MIRRORD_MACOS_ARM64_LIBRARY")` to read the variable at run time
   = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: environment variable `MIRRORD_MACOS_ARM64_LIBRARY` not defined at compile time
   --> mirrord/cli/src/extract.rs:100:36
    |
100 |         let bytes = include_bytes!(env!("MIRRORD_MACOS_ARM64_LIBRARY"));
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: use `std::env::var("MIRRORD_MACOS_ARM64_LIBRARY")` to read the variable at run time
    = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `mirrord` (bin "mirrord") due to 2 previous errors
```

i'll remove this for now and figure out the error improvement later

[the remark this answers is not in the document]
gememma: ah, the if statement in build.rs also silences a compiler diagnostic complaining that the var isnt defined at compile time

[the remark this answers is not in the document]
gememma: oh i missed this

> also the file should be loaded another layer (the arm64 only one)

what do you mean?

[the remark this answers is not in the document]
aviramha: I thought library attachment happens here but now I understand you're just creating a way to do an error - which I don't understand why you println the rust stuff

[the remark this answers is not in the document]
gememma: i was using it as a fallback for if the var isnt defined (if value isnt present use the one for layer instead) but i dont think that makes sense honestly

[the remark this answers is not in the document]
aviramha: yeah this is error prone, I'd remove it

aviramha:
Gave a test build to user (compiled using the script locally) - fixes their issue! 😄

.github/workflows/ci.yaml
```diff
@@ -397,6 +398,9 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: clippy x64
         run: cargo clippy -p mirrord -p mirrord-layer -p mirrord-sip --target=x86_64-apple-darwin --tests -- -Wclippy::indexing_slicing -D warnings
+      # this is required for clippy lint because the layer path variable is checked at compile time
+      - name: create fake layer file for arm 64
```
aviramha, line 402: I think it'd be better to just add clippy ignore to that line with the warning..

gememma: Oh actually i dont think this will work since its a compilation error rather than a clippy failure - if the var isnt given, that varibale wont have a value

scripts/shim.c
t4lz, line 1: imo scripts is not the right location for this. The scripts here should be files that are used by humans, I think.
I would put it in the layer's directory.

.github/workflows/release.yaml
```diff
@@ -72,28 +72,21 @@ jobs:
...
+      - name: Sign layer binaries and code shim
         env:
           AC_USERNAME: ${{ secrets.APPLE_DEVELOPER }}
           AC_PASSWORD: ${{ secrets.APPLE_DEVELOPER_PASSWORD }}
         run: gon -log-level=debug -log-json .github/workflows/arch_dylib_sign.json
       - name: create universal target dir
         run: mkdir -p /tmp/target/universal-apple-darwin/release
       - name: create universal dylib
-        run: lipo -create -output /tmp/target/universal-apple-darwin/release/libmirrord_layer.dylib target/aarch64-apple-darwin/release/libmirrord_layer.dylib target/x86_64-apple-darwin/release/libmirrord_layer.dylib target/aarch64-apple-darwin/release/libmirrord_layer_arm64e.dylib
+        run: lipo -create -output /tmp/target/universal-apple-darwin/release/libmirrord_layer.dylib target/aarch64-apple-darwin/release/shim.dylib target/x86_64-apple-darwin/release/libmirrord_layer.dylib
```
t4lz, line 89: Are you sure macOS will always just load the arm64e library if there is no arm64 library in the fat binary? Otherwise we could include the arm64 binary in the fat binary.

.github/workflows/ci.yaml
```diff
@@ -382,6 +382,7 @@ jobs:
     if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
     env:
       MIRRORD_TEST_USE_EXISTING_LIB: ../../target/x86_64-apple-darwin/debug/libmirrord_layer.dylib
+      MIRRORD_LAYER_FILE_MACOS_ARM64: /tmp/file.dylib
```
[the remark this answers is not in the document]
gememma: ah, i should have removed this in the commit that switched to using clippy ignore 👀 however, i dont think that approach works so I'll revert to before (which includes a comment about why this is needed)

.github/workflows/release.yaml
gememma: @aviramha suggested only including the x86 and the shim, i dont know if there are any situations where macos would take arm64 and not arm64e?

aviramha: I've tested few cases and it always preferred arm64e, but I can't tell for sure ofc..

gememma: it would be very easy to add it in if we ever needed to

gememma: I guess my only concern is that it introduces a decision that macos can make without our input, which could lead to subtle bugs if we cant tell/ are incorrect about which binary is loaded from the fat binary

t4lz (approved):
Please verify `build_fat_mac.sh` and the release action both work, before merging.

.github/workflows/release.yaml
t4lz: I think there is no risk in including the arm64 library in the layer's fat binary, but there is cost, in binary size and compilation time, for including it twice in the mirrord binary. There is a certain risk in not including the arm64 library directly in the layer's fat binary, as there might be some scenarios we're unaware of, where an arm64e library cannot be loaded but an arm64 library can be. And if there aren't any such scenarios, they could be introduced by apple in the future.

aviramha: I took a bullet and dove into the best codebase:
https://github.com/apple-oss-distributions/dyld/blob/65bbeed63cec73f313b1d636e63f243964725a9d/common/MachOFile.cpp#L267
+
https://github.com/apple-oss-distributions/dyld/blob/65bbeed63cec73f313b1d636e63f243964725a9d/common/MachOFile.cpp#L315
Basically each "runtime" arch has an array of prioritized architectures it can load - you can see that arm64e always prefers arm64e - there's no "arm" runtime on macOS, only arm64e so we should be good!

aviramha:
> Please verify `build_fat_mac.sh` and the release action both work, before merging.

To clarify further, make sure you download the release build from the action and see that it works

PR #2874, opened 2024-10-25
Added experimental disable_reuseaddr to bypass the issue

### LINKED ISSUE #2819: mirrord stealing to different app
Some of our users want to use mirrord over `kubectl proxy` which runs on a bastion host they SSH and port-forward to.
They're running 
`ssh -L 8080:localhost:8080`
then inside the ssh
`kubectl proxy`

then on their local machine
```
kubectl config set-cluster testt --server=http://localhost:8080 --insecure-skip-tls-verify=true
kubectl config set-context testt --cluster=testt
kubectl config use-context testt
```

then they `mirrord exec` a Java Bootspring app that listens on 8080, requests gets stolen but when it's stolen it is sent to... the kubectl proxy (ssh forward) instead of the app :|
happens on macOS

ssh version:
OpenSSH_9.7p1, LibreSSL 3.3.6

lsof output (`lsof -nP -iTCP:8080 -sTCP:LISTEN`):
```
ssh     36023 user    7u  IPv6 0x.ccc      0t0  TCP [::1]:8080 (LISTEN)

ssh     36023 user    8u  IPv4  0xbb...      0t0  TCP [127.0.0.1:8080](http://127.0.0.1:8080/) (LISTEN)

nc      36503 user    5u  IPv4 0xaaa..      0t0  TCP *:8080 (LISTEN)
```

aviramha:
Closes #2819

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -1441,7 +1441,7 @@
       "properties": {
         "filter": {
           "title": "feature.network.outgoing.filter {#feature.network.outgoing.filter}",
-          "description": "Unstable: the precise syntax of this config is subject to change.",
+          "description": "Filters that are used to send specific traffic from either the remote pod or the local app",
           "anyOf": [
             {
               "$ref": "#/definitions/OutgoingFilterConfig"
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -758,7 +758,7 @@
       "properties": {
         "disable_reuseaddr": {
           "title": "_experimental_ disable_reuseaddr {#experimental-disable_reuseaddr}",
-          "description": "Disables the `SO_REUSEADDR` socket option on sockets that mirrord steals/mirrors. On macOS the application can use the same address many times but then we don't steal it correctly. This probably should be on by default but we want to gradually roll it out. https://github.com/metalbear-co/mirrord/issues/2819 This option applies only on macOS.",
+          "description": "Disables the `SO_REUSEADDR` socket option on sockets that mirrord steals/mirrors. On macOS the application can use the same address many times but then we don't steal it correctly. This probably should be on by default but we want to gradually roll it out. <https://github.com/metalbear-co/mirrord/issues/2819> This option applies only on macOS.",
           "type": [
             "boolean",
             "null"
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -756,6 +756,14 @@
       "description": "mirrord Experimental features. This shouldn't be used unless someone from MetalBear/mirrord tells you to.",
       "type": "object",
       "properties": {
+        "disable_reuseaddr": {
+          "title": "_experimental_ disable_reuseaddr {#experimental-disable_reuseaddr}",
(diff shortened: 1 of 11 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -275,6 +274,22 @@ pub(super) fn bind(
         Err(HookError::AddressAlreadyBound(requested_address))?;
     }
 
+    #[cfg(target_os = "macos")]
+    unsafe {
```
Razz4780, line 278: Can we restrict the scope of this `unsafe` to bare minimum?

mirrord/layer/src/socket/ops.rs
```diff
@@ -275,6 +274,22 @@ pub(super) fn bind(
...
+        let experimental = crate::setup().experimental();
+        if experimental.disable_reuseaddr {
+            tracing::debug!("here!!");
```
Razz4780, line 281: Delet :wastebasket:

mirrord/layer/src/socket/ops.rs
```diff
@@ -275,6 +274,22 @@ pub(super) fn bind(
...
+            let socket = socket2::Socket::from_raw_fd(sockfd);
+            if let Err(e) = socket.set_reuse_address(false) {
+                tracing::debug!(?e, "Failed to set SO_REUSEADDR to false");
+            }
+            if let Err(e) = socket.set_reuse_port(false) {
+                tracing::debug!(?e, "Failed to set SE_REUSEPORT to false");
+            }
+            // get ownership of fd back
+            let _ = socket.into_raw_fd();
```
Razz4780, line 290: I'd use [nix function](https://docs.rs/nix/latest/nix/sys/socket/fn.setsockopt.html) and `BorrowedFd::borrow_raw` for this. Just a suggestion tho

Razz4780 (changes requested):

mirrord/layer/src/socket/ops.rs
gememma, line 281: 🚨 👀 debug statement

mirrord-schema.json
```diff
@@ -756,6 +756,14 @@
       "description": "mirrord Experimental features. This shouldn't be used unless someone from MetalBear/mirrord tells you to.",
       "type": "object",
       "properties": {
+        "disable_reuseaddr": {
+          "title": "_experimental_ disable_reuseaddr {#experimental-disable_reuseaddr}",
+          "description": "Disables the `SO_REUSEADDR` socket option on sockets that mirrord steals/mirrors. on macOS the application can use the same address many times but then we don't steal it correctly. This probably should be on by default but we want to gradual roll it out. https://github.com/metalbear-co/mirrord/issues/2819 This option applies only on macOS.",
```
gememma, line 761:
```suggestion
          "description": "Disables the `SO_REUSEADDR` socket option on sockets that mirrord steals/mirrors. On macOS the application can use the same address many times but then we don't steal it correctly. This probably should be on by default but we want to gradually roll it out. https://github.com/metalbear-co/mirrord/issues/2819 This option applies only on macOS.",
```

aviramha: Thanks for the suggestion! P.S, next time send the review on the comments where it's being generated from :D

mirrord/layer/src/socket/ops.rs
aviramha: nice!

Razz4780 (approved):

PR #2879, opened 2024-10-29
Upper hex debug display for operator session id

Razz4780:


mirrord/operator/src/client.rs
```diff
@@ -118,7 +118,7 @@ pub struct OperatorSession {
 impl fmt::Debug for OperatorSession {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("OperatorSession")
-            .field("id", &self.id)
+            .field("id", &format!("{:X}", self.id))
```
gememma, line 121: is it worth doing `"0x{:X}"` in case all the digits are 0-9 to make it clear the value is hex?

gememma (approved):
nit 🔬

mirrord/operator/src/client.rs
Razz4780: Hmm, maybe. But we don't do it in the operator and I wanted to have the exact same display

Issue #2868, opened 2024-10-22
Intproxy tracing-subscriber print to console in container mode


DmitryDodzin:
When sidecar is in container mode it should print error to console if encountered one at startup so cli can print to the user an indicative error.

PR #2881, opened 2024-10-29
Update intproxy tracing subscriber print to console in container

DmitryDodzin:
relevant for #2868


mirrord/cli/src/container/sidecar.rs
```diff
@@ -0,0 +1,158 @@
...
+            .stdout
+            .poll_next_line(cx)
+            .map(|result| result.transpose())
+        {
+            return result;
+        }
+
+        this.stderr
+            .poll_next_line(cx)
+            .map(|result| result.transpose())
```
t4lz, line 156: Maybe this is a nitpick - which logs does the sidecar send to stdout and which ones to stderr? Is it ok to be biased towards stdout?
If you think it's ok as-is that's fine with me. If you think we should change it, maybe [tokio's `merge`](https://docs.rs/tokio-stream/0.1.17/tokio_stream/trait.StreamExt.html#method.merge) is helpful.

mirrord/cli/src/container.rs
```diff
@@ -426,11 +303,57 @@ pub(crate) async fn container_command(
...
+/// This spawns: "agent" - "external proxy" - "intproxy sidecar" - "execution container"
+pub(crate) async fn container_command(
+    runtime_args: RuntimeArgs,
+    exec_params: ExecParams,
+    watch: drain::Watch,
+) -> CliResult<i32> {
+    let mut progress = ProgressTracker::from_env("mirrord container");
+
+    if runtime_args.command.has_publish() {
+        progress.warning("mirrord container may have problems with \"-p\" directly container in command, please add to \"contanier.cli_extra_args\" in config if you are planning to publish ports");
```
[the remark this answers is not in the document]
DmitryDodzin: I updated the text a bit wdyt?

mirrord/cli/src/container/sidecar.rs
DmitryDodzin: Ok this entire thing is not needed because of `LinesStream`

PR #2888, opened 2024-10-30
add .class to be always local

aviramha:


mirrord/layer/src/file/filter/read_local_by_default.rs
```diff
@@ -13,6 +13,7 @@ pub fn regex_set_builder() -> RegexSetBuilder {
         r"^.+\.pyc$",
         r"^.+\.py$",
         r"^.+\.jar$",
+        r"^.+\.class$",
```
t4lz, line 16: This and all similar regexes could be simplified to `r".\.class$"`, but doesn't have to be in this PR.

t4lz (approved):

PR #2895, opened 2024-11-06
hide ipv6 interfaces by default

### LINKED ISSUE #2849: Make `hide_ipv6_interfaces` on by default

aviramha:
Closes #2849

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -790,7 +790,7 @@
         },
         "readlink": {
           "title": "_experimental_ readlink {#experimental-readlink}",
-          "description": "Enables the `readlink` hook.",
+          "description": "DEPRECATED, WILL BE REMOVED",
           "type": [
             "boolean",
             "null"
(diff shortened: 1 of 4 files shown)


mirrord-schema.json
```diff
@@ -790,7 +790,7 @@
         },
         "readlink": {
           "title": "_experimental_ readlink {#experimental-readlink}",
-          "description": "Enables the `readlink` hook.",
+          "description": "DEPRECATED, WILL BE REMOVED",
```
DmitryDodzin, line 793: This is for ipv6 only right?

DmitryDodzin (approved):
LGTM, just maybe mention the deprecation in the changelog

mirrord/config/src/experimental.rs
```diff
@@ -40,7 +40,7 @@ pub struct ExperimentalConfig {
     /// ### _experimental_ hide_ipv6_interfaces {#experimental-hide_ipv6_interfaces}
```
gememma, line 40: should it still be experimental if it's enabled by default?

mirrord-schema.json
aviramha: no, I just nitpicked a fix (readlink was deprecated but docs stayed same which was weird)

mirrord/config/src/experimental.rs
aviramha: yes, so users can turn it off if it affects them badly until we roll a fix.

PR #2905, opened 2024-11-07
Change getifaddrs to allocate new list for addresses

gememma:


mirrord/layer/src/socket/ops.rs
```diff
@@ -1594,26 +1594,56 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
...
+    unsafe {
+        while let Some(ifaddr) = count_head.as_mut() {
+            entry_count += 1;
+            count_head = ifaddr.ifa_next;
+        }
+    }
+
+    // Allocate new list so we can safely free the original list later
+    let new_list_start: *mut libc::ifaddrs = unsafe {
+        libc::malloc((mem::size_of::<libc::ifaddrs>() as libc::size_t) * entry_count)
```
meowjesty, line 1609: In theory, `malloc` should be harmless, but we've been avoiding using it.

@aviramha came up with a clever solution around this "problem" in a similar PR [#938](https://github.com/metalbear-co/mirrord/pull/938). Would require hooking the `freeifaddrs` to check if the pointers are ours, or if they are users'.

mirrord/layer/src/socket/ops.rs
```diff
@@ -1594,26 +1594,56 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
...
                 // If not ipv6, advance to the next interface address in the original list.
                 // Move both `previous` and `inspected`.
                 None => {
-                    previous = inspected;
-                    inspected = ifaddr.ifa_next;
-                    continue;
+                    // Append the address to the new list by copying ifaddr to the list head, and
+                    // setting next of previous then moving head
+                    copy_nonoverlapping::<libc::ifaddrs>(inspected, next_new, 1);
+                    match previous_new_entry.as_mut() {
```
meowjesty, line 1632: Could be just `if let Some`.

meowjesty (changes requested):
Some superficial comments.

changelog.d/2903.fixed.md
```diff
@@ -0,0 +1 @@
+Change getifaddrs hook to allocate memory for a new list instead of modifyinging list returned form libc call.
```
meowjesty, line 1:
```suggestion
Change getifaddrs hook to allocate memory for a new list instead of modifying list returned from libc call.
```

mirrord/layer/src/socket/ops.rs
```diff
@@ -1594,26 +1594,56 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
...
+                        None => {
+                            // inspected is head of list
+                        }
+                    }
+
+                    previous_new_entry = next_new;
+                    // clippy lint fails on this line but the pointer arithmetic is fine
+                    #[allow(clippy::size_of_in_element_count)]
+                    {
+                        next_new = next_new.wrapping_add(mem::size_of::<libc::ifaddrs>());
```
meowjesty, line 1645: Is this really fine? Shouldn't `next` point to `1 * size_of(ifaddrs)`?

meowjesty, line 1645:
```suggestion
                    {
                        next_new = next_new.add(mem::size_of::<libc::ifaddrs>());
```

Can't we just use `ptr::add` here instead? We control the allocations, offsets, we should be able to satisfy `add`.

meowjesty (changes requested):
pointery comments

mirrord/layer/src/socket/ops.rs
meowjesty: Which in `ptr::add` terms, means `add(1)`?

mirrord/layer/src/socket/ops.rs
```diff
@@ -1594,26 +1594,56 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
...
-    let mut ipv6_head: *mut libc::ifaddrs = std::ptr::null_mut();
+    // Count entries for new_list_start malloc
+    let mut entry_count = 0;
+    let mut count_head: *mut libc::ifaddrs = original_head;
+    unsafe {
+        while let Some(ifaddr) = count_head.as_mut() {
+            entry_count += 1;
+            count_head = ifaddr.ifa_next;
+        }
+    }
```
meowjesty, line 1605: If we take control of the allocations, by using `HashSet` or whatever is appropriate, we can get rid of this and build the C linked list at the end, when we have all the pointers ready.

meowjesty (commented):
C linkedin lists comment.

mirrord/layer/src/socket/ops.rs
gememma: yeah you're right, i misinterpreted the docs for `wrapping_add()` 🥲

mirrord/layer/src/socket/ops.rs
gememma: i did suggest that approach, @aviramha didnt think it was necessary

meowjesty: Hmm I like it better due to how it would help us avoid pointer math (right?). And we already have this pattern elsewhere, so would be more familiar.

But! Don't let this be a blocker, I'll leave the decision to you.

mirrord/layer/src/socket/ops.rs
aviramha: nope, since `freeifaddrs` calls `free` underneath we should be good (as long as we use malloc)

meowjesty: And as long as the user is also using `malloc`.

mirrord/layer/src/socket/ops.rs
```diff
@@ -1594,26 +1594,47 @@ pub(super) fn getifaddrs() -> HookResult<*mut libc::ifaddrs> {
...
+
+    // Allocate new list so we can safely free the original list later
+    let new_list_start: *mut libc::ifaddrs = unsafe {
```
meowjesty, line 1608:
```suggestion
    // Safety: We assume `libc::malloc` is the same allocator as the user's system.
    let new_list_start: *mut libc::ifaddrs = unsafe {
```

meowjesty (changes requested):
Just add a safety notice so we know where a potential future seg fault is coming from.

Also, tests!

mirrord/layer/tests/hide_ipv6_interfaces.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use rand::Rng;
+use rstest::rstest;
+
+mod common;
+
+pub use common::*;
+
+/// Verifies the IPv6 interfaces are hidden when feature `hide_ipv6_interfaces` is used
+/// via a call to os.networkInterfaces(), which calls C function getifaddrs()
```
meowjesty, line 17: Add a link to the gh issue here.

mirrord/layer/tests/common/mod.rs
```diff
@@ -702,6 +702,7 @@ pub enum Application {
     EnvBashCat,
     NodeFileOps,
     NodeSpawn,
+    NodeInterfaceAddrs,
```
meowjesty, line 705: Maybe change the name of the app to the issue (we have quite a few like that).

`NodeIssue2903`.

mirrord/layer/tests/apps/interface_addrs.mjs
```diff
@@ -0,0 +1,3 @@
+import os from "os";
```
meowjesty, line 1: Also change the name of the file here, to `issue2903`.

tests/src/utils.rs
```diff
@@ -172,10 +172,18 @@ impl TestProcess {
...
 
+    pub async fn assert_stdout_doesnt_contain(&self, string: &str) {
+        assert!(!self.get_stdout().await.contains(string));
+    }
+
     pub async fn assert_stderr_contains(&self, string: &str) {
         assert!(self.get_stderr().await.contains(string));
     }
 
+    pub async fn assert_stderr_doesnt_contain(&self, string: &str) {
```
meowjesty, line 183: This is not being used?

meowjesty (approved):
✨

tests/src/utils.rs
gememma: no, but i figured if i was adding the other one it might be useful at some point

PR #2906, opened 2024-11-07
Add operator permissions for argocd appications

DmitryDodzin:
add a clusterrole rule to allow applications integration
```yaml
- apiGroups:
  - argoproj.io
  resources:
  - applications
  verbs:
  - list
  - get
  - patch
  ```


aviramha:
Not approving because I'm waiting for env that enables application patch (need to add in the operator first)

Razz4780 (approved):

PR #2908, opened 2024-11-07
Fix file mapping doesn't affect xstat

aviramha:


meowjesty (approved):
Can't we have `check_relative_paths!` also remap the path? Not for this PR btw, just wondering.

PR #2910, opened 2024-11-08
Fix agent crash on sniffer failure

### LINKED ISSUE #2909: mirrord fails when sniffer fails on old kernel
was supposed to be fixed in 3.124.1 thanks to https://github.com/metalbear-co/mirrord/pull/2897 but apparently not

aviramha:
Closes #2909

### DIFF
--- a/mirrord/agent/src/entrypoint.rs
+++ b/mirrord/agent/src/entrypoint.rs
@@ -535,10 +535,8 @@ async fn start_agent(args: Args) -> Result<()> {
                     if let Err(err) = res {
                         error!(%err, "Sniffer failed");
                     }
-                } else {
-                    if let Err(error) = sniffer_init_tx.send(false) {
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/entrypoint.rs
```diff
@@ -514,7 +514,7 @@ async fn start_agent(args: Args) -> Result<()> {
     } else {
         let cancellation_token = cancellation_token.clone();
         let is_mesh = args.is_mesh();
-
+        let (sniffer_init_tx, sniffer_init_rx) = tokio::sync::oneshot::channel::<bool>();
```
meowjesty, line 517:
```suggestion
        // We're using this to avoid crashing on old kernels when initializing the `RawSocketTcpCapture`.
        let (sniffer_init_tx, sniffer_init_rx) = tokio::sync::oneshot::channel::<bool>();
```

PR #2913, opened 2024-11-15
Add sqs to `mirrord operator status` report.

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/629

Adds SQS information to `mirrord operator status`, by sending a header containing the SQS config when we're connecting to the operator (`connect_resource`).

### DIFF
--- a/mirrord/cli/src/operator.rs
+++ b/mirrord/cli/src/operator.rs
@@ -21,7 +21,7 @@ use crate::{
     config::{OperatorArgs, OperatorCommand},
     error::{CliError, OperatorSetupError},
     util::remove_proxy_env,
-    CliResult, OperatorSetupParams,
+    OperatorSetupParams, Result,
 };
 
 mod session;
@@ -33,7 +33,7 @@ struct OperatorVersionResponse {
 }
 
 /// Fetches latest version of mirrord operator from our API
-async fn get_last_version() -> CliResult<String, reqwest::Error> {
+async fn get_last_version() -> Result<String, reqwest::Error> {
     let client = reqwest::Client::builder().build()?;
     let response: OperatorVersionResponse = client
         .get("https://version.mirrord.dev/v1/operator/version")
@@ -57,7 +57,7 @@ async fn operator_setup(
         sqs_splitting,
         kafka_splitting,
     }: OperatorSetupParams,
-) -> CliResult<(), OperatorSetupError> {
+) -> Result<(), OperatorSetupError> {
     if !accept_tos {
         eprintln!("Please note that mirrord operator installation requires an active subscription for the mirrord Operator provided by MetalBear Tech LTD.\nThe service ToS can be read here - https://metalbear.co/legal/terms\nPass --accept-tos to accept the TOS");
 
@@ -120,7 +120,7 @@ async fn operator_setup(
 }
 
 #[tracing::instrument(level = Level::TRACE, ret)]
-async fn get_status_api(config: Option<&Path>) -> CliResult<Api<MirrordOperatorCrd>> {
+async fn get_status_api(config: Option<&Path>) -> Result<Api<MirrordOperatorCrd>> {
     let layer_config = if let Some(config) = config {
         let mut cfg_context = ConfigContext::default();
         LayerFileConfig::from_path(config)?.generate_config(&mut cfg_context)?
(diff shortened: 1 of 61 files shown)


mirrord/cli/src/operator.rs
```diff
@@ -147,158 +144,15 @@ async fn get_status_api(config: Option<&Path>) -> Result<Api<MirrordOperatorCrd>
...
-Operator License
-    name: {name}
-    organization: {organization}
-    expire at: {expire_at}
-"#
-    );
-
-    let Some(status) = &api.operator().status else {
-        return Ok(());
-    };
```
meowjesty, line 207: I'm not sure why it was like this, so I changed it to a normal error.

mirrord/cli/src/operator/status.rs
```diff
@@ -0,0 +1,190 @@
...
+    organization: {organization}
+    expire at: {expire_at}
+"#
+        );
+
+        let status = api
+            .operator()
+            .status
+            .as_ref()
+            .ok_or(CliError::OperatorStatusNotFound)?;
```
meowjesty, line 89: 1 of the 2 actual changes in this file.

mirrord/cli/src/operator/status.rs
```diff
@@ -0,0 +1,190 @@
...
+                sqs.add_row(row![session.id.as_deref().unwrap_or_default(), session_sqs]);
+            }
+        }
+
+        sessions.printstd();
+
+        if sqs.get_row(1).is_some() {
+            println!("\nActive SQS:");
+            sqs.printstd();
+        }
```
meowjesty, line 186: Change no. 2 of 2

mirrord/operator/src/client.rs
```diff
@@ -585,14 +592,13 @@ impl OperatorApi<PreparedClientCert> {
...
-                        .split_queues
-                        .is_set()
-                        .then(|| layer_config.feature.split_queues.clone()),
+                    target, scale_down, namespace,
+                    // layer_config
+                    //     .feature
+                    //     .split_queues
+                    //     .is_set()
+                    //     .then(|| layer_config.feature.split_queues.clone()),
+                    None,
```
t4lz, line 601: ?

mirrord/operator/src/client.rs
```diff
@@ -748,23 +754,37 @@ impl OperatorApi<PreparedClientCert> {
...
+        let request = if layer_config.feature.split_queues.is_set() {
+            request.header(
+                HeaderName::from_static(CLIENT_SQS_HEADER),
+                HeaderValue::from_bytes(&serde_json::to_vec(&layer_config.feature.split_queues)?)?,
+            )
+        } else {
+            request
+        };
+
+        let request = request
```
t4lz, line 785: Why change the way we send the SQS config?

mirrord/operator/src/client.rs
meowjesty: This is just what I was using to test the feature, since I don't have a way to use SQS (nor any idea how to mock it).

mirrord/operator/src/client.rs
meowjesty: It's not changing it. This is being consumed on the operator side `connect_resource` because there we don't get any SQS info, only that we're connecting to a copy target. I've tried changing it so SQS would bleed over to `connect_resource`, but it seemed like I would need to introduce a bunch of changes (some potentially breaking due to CRDs?).

Just throwing the SQS stuff into a header seemed way simpler.

aviramha: @t4lz I think the review should start from the operator side.
Trusting user input to determine state is wrong - operator should have information about what current splitting is going on and shouldn't need to rely on user passing the information for it.
In general I think the SQS (and other splitting) status should be in another resource, that can be fetched as part of mirrord operator status (so the status will be built from more than one api call) but that's not a must.

t4lz: I suggest the following flow for SQS status:
1. Client queries mirrord operator's status.
2. mirrord operator determines SQS operator status via Kube API, either:
  - mirrord operator gets the status of all `MirrordWorkloadQueueRegistry` objects (nice if we want to group by workload), or
  - mirrord operator gets the status of all `MirrordSQSSession` objects, or
  - the SQS operator keeps its status in a CR, and the mirrord operator reads that status from there.
3. mirrord operator returns complete status (including SQS) to client.

What do you guys think?

meowjesty: As I don't have a good understanding of what SQS is:

- Is the SQS something that lives longer than the user session that started it, like a copy target pod?
- To implement any of these suggestions, I'll need help getting a running SQS to test stuff.

meowjesty:
This one was too dumb, closing in favor of https://github.com/metalbear-co/mirrord/pull/2928 after implementing the smart suggestions.

PR #2915, opened 2024-11-18
mirrord container port-publish warning

DmitryDodzin:
Add a warning to `mirrord container` command if `-p` is provided as part of run command.


aviramha (approved):

aviramha (approved):
LGTM. remember to change the comment + PR title so commit will be correct.

PR #2921, opened 2024-11-19
Simplify local-read regexes and anchor current dir and exe

t4lz:
1. Remove unneeded anchoring - since we are searching for a match anywhere in the text, we can remove things like `^.*` and `.*$`. This is not meant to cause any change in behaviour (strictly speaking there is the small change that the changed regexes will now correctly match multiline paths).
2. Anchor the current dir path regex to the start of the path - so instead of matching any path that contains the current dir's path anywhere we only match paths that start with it.
3. Anchor the current exe's path to the end of the path - so instead of matching any path that contains the exe's path anywhere, we only match paths that end with it.


mirrord/layer/src/file/filter/read_local_by_default.rs
```diff
@@ -37,16 +37,16 @@ pub fn regex_set_builder() -> RegexSetBuilder {
...
+        r"\.pdb$",
         // dotnet: `/home/{username}/{project}.dll`
-        r"^.*\.dll$",
+        r"\.dll$",
         // jvm.cfg or ANYTHING/jvm.cfg
         r".*(^|/)jvm\.cfg$",
         // TODO: `node` searches for this file in multiple directories, bypassing some of our
         // ignore regexes, maybe other "project runners" will do the same.
-        r"^.*/package.json$",
+        r"/package.json$",
```
Razz4780, line 49:
```suggestion
        r"/package\.json$",
```

mirrord/layer/src/file/filter/read_local_by_default.rs
```diff
@@ -81,10 +81,10 @@ pub fn regex_set_builder() -> RegexSetBuilder {
     .collect();
 
     if let Ok(cwd) = env::current_dir() {
-        patterns.push(format!("^.*{}.*$", cwd.to_string_lossy()));
+        patterns.push(format!("^{}", cwd.to_string_lossy()));
```
Razz4780, line 84: Does `env::current_dir` return the path with slash at the end? Maybe there's a bug here we could fix as well. If the path does not contain the slash, we'd match `/some/dir2` with `/some/dir`

t4lz: In my tests it did not include the slash, but I guess if current dir is root, it will.
Also I'm not really sure it will never include a slash in other cases (different OSs, different FSs). How about something like:
```suggestion
        patterns.push(format!("^{}(/|$)", cwd.to_string_lossy().trim_end_matches('/')));
```

Razz4780: Yeah, I think this works :+1:

Razz4780 (approved):

PR #2923, opened 2024-11-20
Escape paths used in regexes

t4lz:
Otherwise, a path like "/my/cool/executable (1)" would be used in the regex, and that regex would not match that path (it would match "/my/cool/executable 1").


Razz4780 (approved):
:+1:

PR #2924, opened 2024-11-20
Support passing env vars from a file

Razz4780:


mirrord/cli/src/config.rs
```diff
@@ -200,6 +206,10 @@ pub(super) struct ExecParams {
     /// Kube context to use from Kubeconfig
     #[arg(long)]
     pub context: Option<String>,
+
+    /// Path to env file that should be used for the execution.
```
meowjesty, line 210:
```suggestion
    /// Path to env file that should be used for the execution.
    ///
    /// Allows for passing environment variables from an env file.
    ///
    /// These variables will override environment fetched from the remote target.
```

meowjesty (approved):
:+1: 

Just a docs thing.

PR #2931, opened 2024-11-25
Scaledown prep

Razz4780:
New scaledown implementation in the operator patches replicas directly in the workload spec (to keep things nice and simple)


t4lz (approved):
There should be a matching PR for the helm chart, yeah?

Razz4780:
> There should be a matching PR for the helm chart, yeah?

Yup

Issue #2927, opened 2024-11-21
mirrord container -- docker run nginx fails


aviramha:
```
mirrord container -t deployment/ip-visit-counter -- docker run gninx
```
leads to
```
! mirrord container is currently an unstable feature                                                                                                                                                                                                   x ! mirrord container is currently an unstable feature
    ✓ preparing to launch process
      ✓ using operator
        ✓ operator version compatible
        ✓ operator license valid
        ✓ user credentials prepared
        ✓ session started
          ✓ connected to the target                                                                                                                                                                                                                  Error:   × Command output indicates an error [docker logs -f 0c51e080abe659a39ee4ebc4356a9ececc253961eb239f285c6acade6fef1a64]: stdout and stderr were empty
  help: This is a bug. Please report it in our Discord or GitHub repository.
        
        >> Please open a new bug report at https://github.com/metalbear-co/mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help in #mirrord-help
        
        >> Or email us at hi@metalbear.co
        
```
macOS, Sequoia 15
Docker 4.34.3  Docker Engine 27.2.0

PR #2933, opened 2024-11-26
mirrord container docker check run

DmitryDodzin:
For mirrord container command check that the sidecar container is started after it is created when running with docker runtime

relevant for #2927

### DIFF
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -192,35 +192,40 @@ async fn create_sidecar_intproxy(
             )
         })?;
 
+    // For Docker runtime sometimes the sidecar doesn't start so we double check.
+    // See [#2927](https://github.com/metalbear-co/mirrord/issues/2927)
     if matches!(base_command.runtime(), ContainerRuntime::Docker) {
-        let mut inspect_command = Command::new(&runtime_binary);
-        inspect_command
+        let mut container_inspect_command = Command::new(&runtime_binary);
+        container_inspect_command
             .args(["inspect", &sidecar_container_id])
             .stdout(Stdio::piped());
 
-        let inspect_output = inspect_command.output().await.map_err(|err| {
+        let container_inspect_output = container_inspect_command.output().await.map_err(|err| {
             ContainerError::UnsuccesfulCommandOutput(
-                format_command(&inspect_command),
+                format_command(&container_inspect_command),
                 err.to_string(),
             )
         })?;
 
-        let inspect =
-            serde_json::from_slice::<serde_json::Value>(&inspect_output.stdout).unwrap_or_default();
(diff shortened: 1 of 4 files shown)


mirrord/cli/src/container/command_builder.rs
```diff
@@ -57,19 +61,28 @@ impl RuntimeCommandBuilder {
         }
     }
 
-    pub fn add_volume<H, C>(&mut self, host_path: H, container_path: C)
+    pub fn add_volume<H, C>(&mut self, host_path: H, container_path: C, readonly: bool)
```
meowjesty, line 64: Do we need the functions in this file to be `pub`? Reason I'm not a fan of `pub` all the things is that it doesn't warn you if they're not being used.

meowjesty, line 64: Is `readonly` something that can be controlled by the user?

If it's not, then I would suggest changing it from a param to a const generic `<const READONLY: bool = true>` on this function.

mirrord/cli/src/container.rs
```diff
@@ -193,6 +192,41 @@ async fn create_sidecar_intproxy(
             )
         })?;
 
+    if matches!(base_command.runtime(), ContainerRuntime::Docker) {
```
meowjesty, line 195: Could you add some docs here, saying that we're checking the sidecar because sometimes it doesn't start properly (was this the issue?)

mirrord/cli/src/container.rs
```diff
@@ -193,6 +192,41 @@ async fn create_sidecar_intproxy(
...
+            .stdout(Stdio::piped());
+
+        let inspect_output = inspect_command.output().await.map_err(|err| {
+            ContainerError::UnsuccesfulCommandOutput(
+                format_command(&inspect_command),
+                err.to_string(),
+            )
+        })?;
+
+        let inspect =
```
meowjesty, line 208: I would suggest changing the name from `inspect` to something that says what we're inspecting + `_stdout`.

mirrord/cli/src/container.rs
```diff
@@ -193,6 +192,41 @@ async fn create_sidecar_intproxy(
...
+            serde_json::from_slice::<serde_json::Value>(&inspect_output.stdout).unwrap_or_default();
+
+        let status = inspect
```
meowjesty, line 211: `status` is the status of the container?

meowjesty (changes requested):
Nits and docs.

mirrord/cli/src/container/command_builder.rs
DmitryDodzin: is `pub(super)` fine?

meowjesty (approved):
Omnissiah favours you :computer:

Issue #1706, opened 2023-07-23
feature: support for default config name/path for the CLI


Kerwood:
When using the `mirrord` CLI it would be nice if it would check for an existing config file to use instead of having to specify it everytime you run it.
On execution is could look for `mirrord.json`, `mirrord.yaml`, and `mirrord.toml` in the current directory and use the first file available.

Having a mirrord config file in a project repository is convenient when launching a development environment. Having to not specify the config file would be even more convenient.

This feature will not revolutionize mirrord, but it would make it a little bit nicer :)

aviramha:
Thanks for the suggestion! I believe to avoid breakage it'd be best to have it as a flag, wdyt?

Kerwood:
My personal opinion is that it should pick up the config file without any extra arguments. This it what the majority of cli tools do.
I am not sure I would categorize this feature as breaking.

waveywaves:
I would like to contribute to this issue. This is something I thought that the CLI should have instead of having to mention the mirrord.json path every time. Can it be assigned to me ?

aviramha:
Thanks @waveywaves  - happy to accept contribution. Please note that it should be a new flag and not a default behavior.

aviramha:
Sounds okay to me

PR #2939, opened 2024-11-27
feat: use default config path when no arg for -f is provided

waveywaves:
part of https://github.com/metalbear-co/mirrord/issues/1706

### DIFF
--- a/changelog.d/1706.fixed.md
+++ b/changelog.d/1706.fixed.md
@@ -1,0 +1,1 @@
+make `-f` flag accept optional argument, defaulting to "./mirrord.json" when no argument is provided
--- a/changelog.d/1706.fixed.md
+++ b/changelog.d/1706.fixed.md
@@ -1,1 +1,1 @@
-make `-f` flag accept optional argument, defaulting to "./mirrord.json" when no argument is provided
+make `-f` flag accept optional argument, defaulting to "./.mirrord/mirrord.json" when no argument is provided
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -204,7 +204,8 @@ pub(super) struct ExecParams {
     pub disable_version_check: bool,
 
     /// Load config from config file
-    #[arg(short = 'f', long, value_hint = ValueHint::FilePath)]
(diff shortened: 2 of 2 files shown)


Razz4780 (changes requested):
Looks like these no longer have to be `Option`s, no? :eyes:

waveywaves:
> Looks like these no longer have to be `Option`s, no? 👀

@Razz4780  think it does because PathBuf can either be `Some` (if `-f` is provided with or without a value) and `None` (-f is not provided at all). When we use `default_value` instead of `default_missing_value` then we wouldn't need `Option` because there would always be `Some` value which is the default value we pass to `default_value`. `default_missing_value` checks if `-f` has a value. If it does not then it uses `./mirrord.json` as the argument to the `-f` flag.

Let me know if I am making sense.

Razz4780:
> > Looks like these no longer have to be `Option`s, no? 👀
> 
> @Razz4780 think it does because PathBuf can either be `Some` (if `-f` is provided with or without a value) and `None` (-f is not provided at all). When we use `default_value` instead of `default_missing_value` then we wouldn't need `Option` because there would always be `Some` value which is the default value we pass to `default_value`. `default_missing_value` checks if `-f` has a value. If it does not then it uses `./mirrord.json` as the argument to the `-f` flag.
> 
> Let me know if I am making sense.

Ah yes, this is correct. Sorry for confusion

Razz4780 (approved):

t4lz (commented):
Two small comments on the PR:
1. The IDE plugins use a default path of `.mirrord/mirrord.json`, I think the default path in the CLI should be the same.
2. Currently users are able to specify a config file without a `=` sign, like this: `-f my-mirrord-config.json`. After this change, will that still work? If not, this is a breaking change.

waveywaves:
@t4lz 
1. I will update the default path to be the same then.

2. I tested this PR with `-f my-mirrord-config.json` like config with the following command 
```
./target/universal-apple-darwin/debug/mirrord exec -f ./my-mirrord-mirrord.json -t deployment/kotlin-guestbook bash
```
and it works, don't believe this will break the existing user experience.

waveywaves:
@t4lz @Razz4780 PR has been updated 🫡

PR #2943, opened 2024-11-28
Prevent unintended stdout print

DmitryDodzin:
#2933 added a start patch but unintentionally added a small print that is piped from the underlying command, this removes the print.

### DIFF
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -224,8 +224,8 @@ async fn create_sidecar_intproxy(
             container_start_command
                 .args(["start", &sidecar_container_id])
                 .stdin(Stdio::null())
-                .stdout(Stdio::piped())
-                .stderr(Stdio::piped());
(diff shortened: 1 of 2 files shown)


mirrord/cli/src/container.rs
```diff
@@ -221,7 +221,11 @@ async fn create_sidecar_intproxy(
         {
             let mut container_start_command = Command::new(&runtime_binary);
 
-            container_start_command.args(["start", &sidecar_container_id]);
+            container_start_command
+                .args(["start", &sidecar_container_id])
+                .stdin(Stdio::null())
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped());
```
Razz4780, line 228: Looks like we don't read these streams anywhere, so maybe you could use `Stdio::null()` instead?

Razz4780 (changes requested):

Razz4780 (approved):

PR #2945, opened 2024-11-28
Fix race condition with container run caused by regression

### LINKED ISSUE #2944: mirrord container regression
in https://github.com/metalbear-co/mirrord/pull/2933
There was an improvement of not using execvp, but that actually exposed a bug - when we used execvp we didn't drop the `NamedTempFile`s so they actually persisted, then docker was able to access it.
After the change, mirrord process ends because container starts successfully (supposedly) then layer fails on loading with:
```
2024-11-28 14:29:18 mirrord layer setup failed with Config(TemplateRenderingFailed("Couldn't open template '\"/tmp/mirrord-config.json\"' -> No such file or directory (os error 2)"))
```
since the host file was deleted.

aviramha:
Fixes #2944

### DIFF
--- a/changelog.d/2944.fixed.md
+++ b/changelog.d/2944.fixed.md
@@ -1,0 +1,1 @@
+Fix race condition with container run caused by regression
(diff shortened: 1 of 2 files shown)


meowjesty (approved):
:+1:

PR #2946, opened 2024-11-28
3.125.2

aviramha:
continue of #2945 + 3.125.2

### DIFF
--- a/mirrord/cli/src/container.rs
+++ b/mirrord/cli/src/container.rs
@@ -435,7 +435,9 @@ pub(crate) async fn container_command(
         .status()
         .await;
 
-    let _ = composed_config_file.close();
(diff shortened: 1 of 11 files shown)


mirrord/cli/src/container.rs
```diff
@@ -435,6 +435,19 @@ pub(crate) async fn container_command(
         .status()
         .await;
 
+    // Keep the files, since process is going to exit before process is actually ran and uses it
```
t4lz, line 438:
```suggestion
    // Keep the files, since the process is going to exit before the new process is actually run and uses them
```
Maybe also be more specific and write which process we're talking about instead of just "the process", like "the internal proxy's process" if that's the process we're talking about.

t4lz (approved):

Issue #2221, opened 2024-02-06
feature fs - agent mkdir


klingenm:
Currently, configuring 

```
{
  "feature": {
    "fs": {
      "read_write": "^/mnt/data/.*"
    }
  }
}
```

will only successfully create "ordinary" file entries in the remote filesystem. If the executable running in `mirrord` tries to create a directory the remote filesystem the mkdir call will actually be called locally.

The current behaviour is quite confusing, as xstat calls are forwarded to the agent (checking if directory exists) but actual mkdir call is done locally.

danielloader:
Yeah this has hit me before too, good to see an issue raised.

AlbertZheng:
Yeah this has hit me too, good to see an issue raised. BTW, `rmdir` also has the same confusing behaviour.
please consider to support too: feature fs - agent rmdir

aviramha:
Done in https://github.com/metalbear-co/mirrord/pull/2949
Thank you @facundopoblete

PR #2949, opened 2024-11-29
Add mkdir support

facundopoblete:
Add support to forward the `mkdir` command to the agent

- Issue: https://github.com/metalbear-co/mirrord/issues/2221


aviramha:
This looks great, thank you for taking a stab at it - I know it's still in draft, but need to remember that we try to maintain agent<>cli compatability, and the way we add new messages is by checking protocol version before using new message types (see `READDIR_BATCH_VERSION`)

mirrord/agent/src/file.rs
```diff
@@ -483,35 +487,64 @@ impl FileManager {
 
         let path = resolve_path(path, &self.root_path)?;
 
-        match std::fs::create_dir(Path::new(&path)) {
-            Ok(_) => Ok(MakeDirResponse {
+        let c_path = match std::ffi::CString::new(path.to_string_lossy().as_bytes()) {
```
aviramha, line 490: We'd prefer to avoid libc - is it to obtain underlying error? We'd prefer to convert the errors to our "standard" (which would be Rust's probably). It's less consistent but easier to maintain in case of introducing new platforms from each side.

mirrord/protocol/src/codec.rs
```diff
@@ -126,6 +128,8 @@ pub enum FileResponse {
     OpenDir(RemoteResult<OpenDirResponse>),
     GetDEnts64(RemoteResult<GetDEnts64Response>),
     ReadLink(RemoteResult<ReadLinkFileResponse>),
+    MakeDir(RemoteResult<MakeDirResponse>),
+    MakeDirAt(RemoteResult<MakeDirAtResponse>),
```
Razz4780, line 132: New variants must be added at the end of enum definition here (due to how our serialization works) ;)

mirrord/agent/src/file.rs
facundopoblete: This is one of the questions that I was about to ask later. I did it when I saw that the mkdirat had a custom logic for relative paths, so I just created a 1-1 match for the libc functions. I see that the same happens with `openat`  (which is already implemented), so I will use that as an example to make this change.

Regarding file/directory permissions, is there something that we should take into account?

aviramha: Nothing that comes to mind.

aviramha:
Is this still WIP? (I see it's not marked as draft)

facundopoblete:
> Is this still WIP? (I see it's not marked as draft)

Done. Sorry for the delay, I had to make some changes and I needed the CI to validate them.

mirrord/layer/src/file/ops.rs
```diff
@@ -337,6 +337,60 @@ pub(crate) fn read_link(path: Detour<PathBuf>) -> Detour<ReadLinkFileResponse> {
     }
 }
 
+#[mirrord_layer_macro::instrument(level = Level::TRACE, ret)]
+pub(crate) fn mkdir(pathname: Detour<PathBuf>, mode: u32) -> Detour<MakeDirResponse> {
+    let pathname = pathname?;
+
+    check_relative_paths!(pathname);
```
facundopoblete, line 344: I copied this from `open` function, and I just wanted to validate if that validation is there because mirrord is not able to get the working directory of the original process (the one that runs on the target pod)

Razz4780: Yeah, getting the working directory would be problematic (what if there were multiple processes in the target container, what if workdir changes, stuff like this). Good call to include this check ^^

mirrord/protocol/src/codec.rs
```diff
@@ -127,6 +129,8 @@ pub enum FileResponse {
     GetDEnts64(RemoteResult<GetDEnts64Response>),
     ReadLink(RemoteResult<ReadLinkFileResponse>),
     ReadDirBatch(RemoteResult<ReadDirBatchResponse>),
+    MakeDir(RemoteResult<MakeDirResponse>),
+    MakeDirAt(RemoteResult<MakeDirResponse>),
```
Razz4780, line 133: These could be merged into one variant (like `FileRequest::Open` and `FileRequest::OpenRelative` both result in `FileResponse::Open`).
Additionaly, `MakeDirResponse` contains only error code, which should already be handled by `RemoteResult` wrapper. So the variant can look like:
```rust
pub enum FileResponse {
  ...
  
  MakeDir(RemoteResult<()>),
}
```

mirrord/agent/src/file.rs
```diff
@@ -472,6 +485,67 @@ impl FileManager {
             })
     }
 
+    pub(crate) fn mkdir(&mut self, path: &Path, mode: u32) -> RemoteResult<MakeDirResponse> {
```
Razz4780, line 488: Instead of using separate calls to `std::fs::create_dir` and `fs::set_permissions` in this function, you can use one call to [`nix::unistd::mkdir`](https://docs.rs/nix/latest/nix/unistd/fn.mkdir.html) (requires enabling `fs` feature on `nix` dependency)

mirrord/agent/src/file.rs
```diff
@@ -472,6 +485,67 @@ impl FileManager {
...
+        let permissions = fs::Permissions::from_mode(mode);
+        fs::set_permissions(path, permissions)?;
+
+        Ok(MakeDirResponse {
+            result: 0,
+            errno: 0,
+        })
+    }
+
+    pub(crate) fn mkdirat(
```
Razz4780, line 509: Like above, you can use just one call to [`nix::sys::stat::mkdirat`](https://docs.rs/nix/latest/nix/sys/stat/fn.mkdirat.html)

Razz4780 (changes requested):
Looks neat! Only some small improvement suggestions

mirrord/protocol/src/codec.rs
facundopoblete: By using the MakeDirResponse, all error codes were returned to the mirrord client. I think that not all [mkdir](https://man7.org/linux/man-pages/man2/mkdir.2.html)/libc error codes are mapped on error.rs if we go for the ResponseError option. I think this also applies to other libc functions, Are we ok with that?

mirrord/agent/src/file.rs
facundopoblete: 👍 Done, thanks

mirrord/agent/src/file.rs
facundopoblete: 👍 Done, thanks

mirrord/protocol/src/codec.rs
Razz4780: Yeah, that's ok ^^

Razz4780 (approved):
Sweet! Thanks for the contribution!

PR #2950, opened 2024-11-29
Added local filter for hidden paths inside $HOME directory

Razz4780:


DmitryDodzin (approved):
LGTM

mirrord/layer/src/file/filter/read_local_by_default.rs
```diff
@@ -92,5 +92,13 @@ pub fn regex_set_builder() -> RegexSetBuilder {
         patterns.push(format!("{}$", regex::escape(&executable.to_string_lossy())));
     }
 
+    // Hidden files and directories in $HOME.
+    if let Ok(home) = env::var("HOME") {
+        patterns.push(format!(
+            "^{}/\\.",
```
t4lz, line 98:
```suggestion
            "^{}/\.",
```

Otherwise this requires a `\` in the path, doesn't it?

PR #2953, opened 2024-12-03
Support kafka settings from secrets with load_from_secret field

gememma:


mirrord/operator/src/crd/kafka.rs
```diff
@@ -23,6 +24,11 @@ pub struct MirrordKafkaClientConfigSpec {
     ///
     /// The list of all available properties can be found [here](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).
     pub properties: Vec<MirrordKafkaClientProperty>,
+
+    /// Properties to set from a secret.
+    ///
+    /// The secret is resolved when resolving the rest of the final configuration.
```
Razz4780, line 30:
```suggestion
    /// Namespace and name of a `Secret` to use as another source of properties.
    ///
    /// The secret is fetched and entries in its `.data` field are resolved to properties.
    /// Properties resolved this way override properties from the parent configuration.
    /// Properties defined inline in this object override properties resolved from the secret.
    ///
    /// Example value: `default/my-secret`
```

changelog.d/+kafka-settings-from-secret.internal.md
```diff
@@ -0,0 +1 @@
+Add optional load_from_secret field to Kafka client config spec to allow setting properties from a Secret.
```
Razz4780, line 1:
```suggestion
Add optional `loadFromSecret` field to Kafka client config spec to allow setting properties from a Secret.
```
We rename fields to camelCase, I think it's better to use the name from resource definition (changelog is for the users)

mirrord/operator/src/crd/kafka.rs
```diff
@@ -23,6 +24,11 @@ pub struct MirrordKafkaClientConfigSpec {
...
+    pub load_from_secret: Option<Secret>,
```
Razz4780, line 31:
```suggestion
    pub load_from_secret: Option<String>,
```
We don't want to include the whole `Secret` object, we want to reference an exisiting `Secret`

Razz4780, line 31: You can also add a `printcolumn` entry for this field (makes `kubectl get ...` output nicer)

Razz4780 (changes requested):
You can a

mirrord/operator/src/crd/kafka.rs
gememma: ohh i see 👀 ty

Razz4780 (approved):

PR #2954, opened 2024-12-03
Add a tokio sleep after MirrordExecution::start so we can await on it and drive the runtime, preventing the websocket reset.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/79


aviramha (approved):
Fix check-rust-docs - lgtm otherwise :)

PR #2960, opened 2024-12-04
Add 'mirrord-operator-secret-access' role to operator

gememma:
Related to #2953 - the operator needs explicit access to read Secrets from the cluster

### DIFF
--- a/mirrord/operator/src/setup.rs
+++ b/mirrord/operator/src/setup.rs
@@ -147,7 +147,7 @@ impl Operator {
         let (role, role_binding) = kafka_splitting
             .then(|| {
                 let role = OperatorRole::new(&namespace);
-                let role_binding = OperatorRoleBinding::new(&role, &service_account);
+                let role_binding = OperatorRoleBinding::new(&role, &service_account, &namespace);
                 (role, role_binding)
             })
             .unzip();
@@ -759,10 +759,15 @@ impl OperatorRole {
 pub struct OperatorRoleBinding(RoleBinding);
 
 impl OperatorRoleBinding {
-    pub fn new(role: &OperatorRole, sa: &OperatorServiceAccount) -> Self {
+    pub fn new(
+        role: &OperatorRole,
+        sa: &OperatorServiceAccount,
+        namespace: &OperatorNamespace,
+    ) -> Self {
         let role_binding = RoleBinding {
             metadata: ObjectMeta {
                 name: Some(OPERATOR_ROLE_BINDING_NAME.to_owned()),
+                namespace: Some(namespace.name().to_owned()),
                 ..Default::default()
             },
             role_ref: role.as_role_ref(),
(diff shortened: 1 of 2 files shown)


mirrord/operator/src/setup.rs
```diff
@@ -44,6 +44,7 @@ static OPERATOR_LICENSE_SECRET_FILE_NAME: &str = "license.pem";
 static OPERATOR_LICENSE_SECRET_VOLUME_NAME: &str = "license-volume";
 static OPERATOR_SERVICE_ACCOUNT_NAME: &str = "mirrord-operator";
 static OPERATOR_SERVICE_NAME: &str = "mirrord-operator";
+static OPERATOR_SECRET_ROLE_NAME: &str = "mirrord-operator-secret-access";
```
Razz4780, line 47:
```suggestion
static OPERATOR_ROLE_NAME: &str = "mirrord-operator";
```
plus rename exisiting `OPERATOR_ROLE_NAME` to `OPERATOR_CLUSTER_ROLE_NAME`

mirrord/operator/src/setup.rs
```diff
@@ -105,6 +106,7 @@ pub struct Operator {
     namespace: OperatorNamespace,
     role: OperatorRole,
     role_binding: OperatorRoleBinding,
+    kafka_secret_role: Option<OperatorSecretRole>,
```
Razz4780, line 109:
```suggestion
    role: Option<OperatorSecretRole>,
```
plus rename exisiting `role` to `cluster_role`

mirrord/operator/src/setup.rs
```diff
@@ -691,6 +701,41 @@ impl OperatorRoleBinding {
     }
 }
 
+#[derive(Debug)]
+pub struct OperatorSecretRole(Role);
+
+impl OperatorSecretRole {
+    pub fn new(namespace: &OperatorNamespace, kafka_splitting: bool) -> Option<Self> {
```
Razz4780, line 708:
```suggestion
    pub fn new(namespace: &OperatorNamespace) -> Self {
```
And call it like `kafka_splitting.then(|| todo!("constructor here"))`. Wdyt?

Razz4780 (changes requested):
I think we need `RoleBinding` as well, but I'm not sure :eyes: 
We have analogous `ClusterRoleBinding` for the `ClusterRole`

mirrord/operator/src/setup.rs
gememma: oh nice, yeah I wasn't a fan of a `new()` function returning `Option` - this is way nicer

mirrord/operator/src/setup.rs
```diff
@@ -667,18 +692,75 @@ impl OperatorRole {
...
+    }
+}
+
+#[derive(Debug)]
+pub struct OperatorRoleBinding(RoleBinding);
 
 impl OperatorRoleBinding {
     pub fn new(role: &OperatorRole, sa: &OperatorServiceAccount) -> Self {
-        let role_binding = ClusterRoleBinding {
+        let role_binding = RoleBinding {
```
Razz4780, line 763: This `RoleBinding` needs a namespace, I think

Razz4780 (changes requested):

mirrord/operator/src/setup.rs
Razz4780: Nvm, my bad

Razz4780 (approved):

Issue #2920, opened 2024-11-18
override env with regex


aviramha:
Add a new configuration under `env` called `override_pattern`
It should work like `fs.mapping` where any env key that matches the pattern, will be set to the value and can use groups from the env key that was captured.
for example:
```json
{"env": {"override_pattern": {".+_TIMEOUT": "10000"}}
```

will set all keys that end with `TIMEOUT` to value `10000`

PR #2961, opened 2024-12-04
Add env vars remapper, allowing the user to change multiple env vars values based on a regex matched against env var keys.

meowjesty:
- Issue: #2920

### DIFF
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -6,7 +6,8 @@ use std::{
 
 use mirrord_analytics::{AnalyticsError, AnalyticsReporter, Reporter};
 use mirrord_config::{
-    config::ConfigError, internal_proxy::MIRRORD_INTPROXY_CONNECT_TCP_ENV, LayerConfig,
+    config::ConfigError, feature::env::mapper::EnvVarsRemapper,
+    internal_proxy::MIRRORD_INTPROXY_CONNECT_TCP_ENV, LayerConfig,
 };
 use mirrord_intproxy::agent_conn::AgentConnectInfo;
 use mirrord_operator::client::OperatorSession;
@@ -504,8 +505,8 @@ impl MirrordExecution {
             env_vars.extend(envs_from_file);
         }
 
-        if let Some(overrides) = &config.feature.env.r#override {
-            env_vars.extend(overrides.iter().map(|(k, v)| (k.clone(), v.clone())));
+        if let Some(mapping) = config.feature.env.mapping.clone() {
+            env_vars = EnvVarsRemapper::new(mapping, env_vars).remapped();
         }
 
         Ok(env_vars)
--- a/mirrord/config/src/feature/env.rs
+++ b/mirrord/config/src/feature/env.rs
@@ -10,6 +10,8 @@ use crate::{
     util::{MirrordToggleableConfig, VecOrSingle},
 };
 
+pub mod mapper;
+
 pub const MIRRORD_OVERRIDE_ENV_VARS_INCLUDE_ENV: &str = "MIRRORD_OVERRIDE_ENV_VARS_INCLUDE";
 pub const MIRRORD_OVERRIDE_ENV_VARS_EXCLUDE_ENV: &str = "MIRRORD_OVERRIDE_ENV_VARS_EXCLUDE";
 pub const MIRRORD_OVERRIDE_ENV_FILE_ENV: &str = "MIRRORD_OVERRIDE_ENV_VARS_FILE";
@@ -111,6 +113,26 @@ pub struct EnvConfig {
     /// These variables will override environment fetched from the remote target.
     #[config(env = MIRRORD_OVERRIDE_ENV_FILE_ENV)]
     pub env_file: Option<PathBuf>,
+
+    /// ### feature.env.mapping {#feature-env-mapping}
+    ///
+    /// Specify map of patterns that if matched will replace the value according to specification.
+    ///
+    /// *Capture groups are allowed.*
+    ///
+    /// Example:
+    /// ```json
+    /// {
+    ///   ".+_TIMEOUT": "10000"
+    ///   "LOG_.+_VERBOSITY": "debug"
+    /// }
+    /// ```
+    ///
+    /// Will do the next replacements for environment variables that match:
+    ///
+    /// `CONNECTION_TIMEOUT: 500` => `CONNECTION_TIMEOUT: 10000`
+    /// `LOG_FILE_VERBOSITY: info` => `LOG_FILE_VERBOSITY: debug`
+    pub mapping: Option<HashMap<String, String>>,
 }
 
 impl MirrordToggleableConfig for EnvFileConfig {
@@ -129,6 +151,7 @@ impl MirrordToggleableConfig for EnvFileConfig {
             env_file: FromEnv::new(MIRRORD_OVERRIDE_ENV_FILE_ENV)
                 .source_value(context)
                 .transpose()?,
+            mapping: None,
         })
     }
 }
@@ -164,6 +187,13 @@ impl CollectAnalytics for &EnvConfig {
                 .unwrap_or_default(),
         );
         analytics.add("env_file_used", self.env_file.is_some());
+        analytics.add(
+            "env_mapping_count",
+            self.mapping
+                .as_ref()
+                .map(|v| v.len() as u32)
+                .unwrap_or_default(),
+        );
     }
 }
 
--- a/mirrord/config/src/feature/env/mapper.rs
+++ b/mirrord/config/src/feature/env/mapper.rs
@@ -1,0 +1,101 @@
+use std::collections::HashMap;
+
+use fancy_regex::Regex;
+use tracing::Level;
+
+#[derive(Debug)]
+pub struct EnvVarsRemapper {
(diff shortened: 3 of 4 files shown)


mirrord/config/src/feature/env.rs
```diff
@@ -111,6 +113,26 @@ pub struct EnvConfig {
...
+    /// {
+    ///   ".+_TIMEOUT": "10000"
+    ///   "LOG_.+_VERBOSITY": "debug"
+    /// }
+    /// ```
+    ///
+    /// Will do the next replacements for environment variables that match:
+    ///
+    /// `CONNECTION_TIMEOUT: 500` => `CONNECTION_TIMEOUT: 10000`
+    /// `LOG_FILE_VERBOSITY: info` => `LOG_FILE_VERBOSITY: debug`
```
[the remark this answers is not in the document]
meowjesty: Reuse capture groups? Can you elaborate, I'm not following?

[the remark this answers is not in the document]
aviramha: like fs.mapping
```
    /// ### feature.fs.mapping {#feature-fs-mapping}
    ///
    /// Specify map of patterns that if matched will replace the path according to specification.
    ///
    /// *Capture groups are allowed.*
    ///
    /// Example:
    /// ```json
    /// {
    ///   "^/home/(?<user>\\S+)/dev/tomcat": "/etc/tomcat"
    ///   "^/home/(?<user>\\S+)/dev/config/(?<app>\\S+)": "/mnt/configs/${user}-$app"
    /// }
    /// ```
    /// Will do the next replacements for any io operaton
    ///
    /// `/home/johndoe/dev/tomcat/context.xml` => `/etc/tomcat/context.xml`
    /// `/home/johndoe/dev/config/api/app.conf` => `/mnt/configs/johndoe-api/app.conf`
    ///
    /// - Relative paths: this feature (currently) does not apply mappings to relative paths, e.g.
    ///   `../dev`.
 ```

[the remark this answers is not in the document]
Comment (meowjesty) [reply mirrord/config/src/feature/env.rs:134]: Hmm but doesn't `fancy_regex` also support capture groups? Like if the user has a config of (imagining these are env vars instead of paths):

```json
{
  "^/home/(?<user>\\S+)/dev/tomcat": "/etc/tomcat"
  "^/home/(?<user>\\S+)/dev/config/(?<app>\\S+)": "/mnt/configs/${user}-$app"
}
```

it should work the same?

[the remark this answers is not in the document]
aviramha: you use match, not replace - but in any case now that I think about it it's less relevant in this use case anyway.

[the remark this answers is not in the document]
meowjesty: Ah I see.

mirrord/cli/src/execution.rs
```diff
@@ -504,8 +505,8 @@ impl MirrordExecution {
             env_vars.extend(envs_from_file);
         }
 
-        if let Some(overrides) = &config.feature.env.r#override {
-            env_vars.extend(overrides.iter().map(|(k, v)| (k.clone(), v.clone())));
+        if let Some(mapping) = config.feature.env.mapping.clone() {
+            env_vars = EnvVarsRemapper::new(mapping, env_vars).remapped();
```
Razz4780, line 509: Why remove overrides? :eyes:

mirrord/config/src/feature/env.rs
```diff
@@ -111,6 +113,26 @@ pub struct EnvConfig {
     /// These variables will override environment fetched from the remote target.
     #[config(env = MIRRORD_OVERRIDE_ENV_FILE_ENV)]
     pub env_file: Option<PathBuf>,
+
+    /// ### feature.env.mapping {#feature-env-mapping}
+    ///
+    /// Specify map of patterns that if matched will replace the value according to specification.
+    ///
+    /// *Capture groups are allowed.*
```
Razz4780, line 121: Can you add an example with capture groups?

mirrord/config/src/feature/env.rs
```diff
@@ -111,6 +113,26 @@ pub struct EnvConfig {
...
+    ///   ".+_TIMEOUT": "10000"
+    ///   "LOG_.+_VERBOSITY": "debug"
+    /// }
+    /// ```
+    ///
+    /// Will do the next replacements for environment variables that match:
+    ///
+    /// `CONNECTION_TIMEOUT: 500` => `CONNECTION_TIMEOUT: 10000`
+    /// `LOG_FILE_VERBOSITY: info` => `LOG_FILE_VERBOSITY: debug`
+    pub mapping: Option<HashMap<String, String>>,
```
Razz4780, line 135: We need to verify regexes during config validation

mirrord/config/src/feature/env/mapper.rs
```diff
@@ -0,0 +1,114 @@
...
+                    .iter()
+                    .find(|(k, _)| k.is_match(&key).ok().is_some_and(|matched| matched))
+                    .map(|(_, to_value)| to_value.clone())
+                {
+                    Some(to_value) => acc.insert(key, to_value),
+                    None => acc.insert(key, value),
+                };
+
+                acc
+            })
```
Razz4780, line 45: This is a bit weird :eyes: 
Can't we do
```rust
env_vars
  .iter_mut()
  .for_each(|(name, value)| {
    ....
  });

env_vars
```

Also, I don't see capture groups support :eyes:

mirrord/layer/src/lib.rs
```diff
@@ -455,6 +455,10 @@ fn fetch_env_vars() -> HashMap<String, String> {
         env_vars.extend(overrides.iter().map(|(k, v)| (k.clone(), v.clone())));
     }
 
+    if let Some(mapping) = setup().env_config().mapping.clone() {
+        env_vars = EnvVarsRemapper::new(mapping, env_vars).remapped();
+    }
```
Razz4780, line 460: I think override should be applied after this step. Same goes for fetching env vars in the CLI

mirrord/config/src/feature/env/mapper.rs
meowjesty: Oops.

Changed it to `captures`.

mirrord/cli/src/execution.rs
meowjesty: Argh, think I wrote over it when copy pasting the `if let`. Nice catch.

mirrord/layer/src/lib.rs
meowjesty: Changed the order.

mirrord/config/src/feature/env/mapper.rs
meowjesty: Hmm I think we run into some annoying borrowing issues.

mirrord/config/src/feature/env.rs
meowjesty: Added to the `verify` fn.

mirrord/config/src/config.rs
```diff
@@ -76,6 +76,9 @@ pub enum ConfigError {
 
     #[error("Queue splitting config is invalid: {0}")]
     QueueSplittingVerificationError(#[from] QueueSplittingVerificationError),
+
+    #[error("Regex creation failed with: {0}")]
+    Regex(#[from] Box<fancy_regex::Error>),
```
Razz4780, line 81: This error could be more specific, mentioning env mapping directly (and containing the bad regex)

mirrord/config/src/feature/env/mapper.rs
```diff
@@ -0,0 +1,134 @@
...
+                match mapping.iter().find_map(|(k, new_value)| {
+                    k.captures(&key).ok()??.get(0).map(|_| new_value.clone())
+                }) {
+                    Some(new_value) => modified_env_vars.insert(key, new_value),
+                    None => modified_env_vars.insert(key, value),
+                };
+
+                modified_env_vars
+            },
+        )
```
Razz4780, line 46:
```suggestion
    env_vars.iter_mut().for_each(|(name, value)| {
        let regex_match = mapping
            .iter()
            .find(|(regex, _)| regex.is_match(name).unwrap_or(false));
        if let Some((_, replace_with)) = regex_match {
            *value = replace_with.clone();
        }
    });

    env_vars
```
This works :eyes:

Razz4780 (changes requested):

mirrord/config/src/feature/env/mapper.rs
Razz4780: > Changed it to `captures`.

Sorry, I was wrong. I thought capture groups support meant that the user can use capture groups in value (in other words, value is a template that can refer to the captured portions of the key). `is_match` is fine here

mirrord/config/src/feature/env/mapper.rs
meowjesty: Oh I see. I was re-inserting it.

mirrord/config/src/config.rs
meowjesty: Improved a bit.

mirrord/config/src/feature/env/mapper.rs
meowjesty: Change to `is_match` thanks to your other suggestion.

Razz4780 (approved):
:rocket: :kiribati: :cocktail:

PR #2964, opened 2024-12-06
Stop mirrord ending unexpectedly when port forwarding in reverse

gememma:
Changed the ping/pong routine of `ReversePortForwarder` to match that of `PortForwarder` - this prevents the timeout being repeatedly reset when recieving messages more frequently than every 30 seconds (causing the ping to never be sent)


Razz4780 (approved):
:+1:

PR #2967, opened 2024-12-06
Add `MirrordPolicy` that blocks traffic mirroring

Razz4780:
I tried versioning `MirrordPolicy`, but to no success. In a scenario when the cluster has the new policy and the operator does not recognize it, there are only two options:
1. Operator will fail to deserialize this policy
2. Operator will silently ignore the new blocked feature

This is because `kube::Api` fetches and attempts to deserialize all resources of the given kind, regardless of their version

Since option 1 is less scary, we can as well just add another enum variant, without introducing the versioning boilerplate.
I also added a catch-all variant to prevent this problem in the future.

### DIFF
--- a/changelog.d/+block-mirroring-policy.added.md
+++ b/changelog.d/+block-mirroring-policy.added.md
@@ -1,0 +1,1 @@
+`MirrordPolicy` can now block traffic mirroring (requires operator support).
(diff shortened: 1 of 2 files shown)


aviramha (approved):

aviramha:
Update docs :)

PR #2976, opened 2024-12-10
IPv6 support for traffic stealing

### LINKED ISSUE #2956: Support and test traffic stealing with IPv6

t4lz:
Closes #2956

Add config flag to enable IPv6 support.
When enabled, mirrord will try to also steal IPv6 connections.
The plan is to enable it by default some day, so I made the behaviour of the agent when enabled what I think the future default behaviour should be: try to setup steeling for both IPv4 and IPv6. If at least one of them can be set up, that's fine, also if it's IPv4. Only fail if none of them could be set up.

Did not add any handling for a mesh + IPv6 case.

Added an E2E test.


tests/src/utils.rs
```diff
@@ -575,6 +586,7 @@ pub async fn run_exec(
     // minikube load image test:latest
     let mut base_env = HashMap::new();
     base_env.insert("MIRRORD_AGENT_IMAGE", "test");
+    base_env.insert("MIRRORD_AGENT_TTL", "30");
```
DmitryDodzin, line 589: Do we need the extra ttl? for logs or is there some other reason?

tests/src/traffic/steal.rs
```diff
@@ -63,6 +71,47 @@ mod steal_tests {
...
+        #[future] ipv6_service: KubeService,
+        #[future] kube_client: Client,
+    ) {
+        let application = Application::PythonFastApiHTTPIPv6;
+        let service = ipv6_service.await;
+        let kube_client = kube_client.await;
+
+        let mut flags = vec!["--steal"];
+
+        // if cfg!(feature = "ephemeral") {
```
DmitryDodzin, line 88: We can keep it uncommented if we are planing to test the "ephemeral" variant at some point

DmitryDodzin (approved):
LGTM, just a couple nitpicks

tests/src/utils.rs
t4lz: Yeah was supposed to be for logs, but we removed the new test from CI anyways so I'll just remove that too.

Issue #2069, opened 2023-11-10
Connecting to a PostgreSQL DB with JDBC takes forever 


t4lz:
### Bug Description

User connects to the DB with a URL like `jdbc:postgresql://<DB-LOCATION>?sslmode=verify-ca&sslrootcert=/run/secrets/rds_public_key/some_cert.pem`, the PostreSQL jdbc driver reads the provided cert file byte-by-byte, so we send a request to the agent for each byte in the file and that is way too slow.

### Steps to Reproduce

Write a Java app that connects to a URL like `jdbc:postgresql://<DB-LOCATION>?sslmode=verify-ca&sslrootcert=/run/secrets/rds_public_key/some_cert.pem` and run with mirrord

### Your operating system and version

macOS

### Additional Info

We could do #2067 in order to optimise reading the file.

As a temporary workaround the user downloads the cert file and tells mirrord to read it locally, but that path can't be written to on macOS so currently the user also has to change the cert file's path in the URL argument in the app's code, which is suboptimal. #2068 could help with that as at least the user would not have to change the app's code for that workaround.

aviramha:
Fixed in #2979 - requires config for now:
```json
{"experimental": {"readonly_file_buffer": 4096}}
```

PR #2979, opened 2024-12-13
Buffering readonly files

Razz4780:
Issue #2069 

Moved file ops to its own proxy inside intproxy. The logic around files is getting to complex to be kept in `SimpleProxy` (which was supposed to just pass messages back and forth).

Haven't tested the solution with jdbc yet

### DIFF
--- a/mirrord/layer/tests/common/mod.rs
+++ b/mirrord/layer/tests/common/mod.rs
@@ -109,7 +109,7 @@ impl TestIntProxy {
             let agent_conn = AgentConnection::new_for_raw_address(fake_agent_address)
                 .await
                 .unwrap();
-            let intproxy = IntProxy::new_with_connection(agent_conn, listener, false);
+            let intproxy = IntProxy::new_with_connection(agent_conn, listener, 0);
             intproxy
                 .run(Duration::from_secs(5), Duration::from_secs(5))
                 .await


mirrord/intproxy/src/proxies/files.rs
```diff
@@ -0,0 +1,1345 @@
...
+                    .send(ClientMessage::FileRequest(FileRequest::OpenRelative(open)))
+                    .await;
+            }
+
+            // Try to use local buffer if possible.
+            FileRequest::Read(read) => match self.files_data.get_mut(&read.remote_fd) {
+                // File is buffered.
+                Some(data) => {
+                    let from_buffer = data.read_from_buffer(read.buffer_size, data.fd_position);
+                    if let Some(from_buffer) = from_buffer {
```
meowjesty, line 307: We read from the buffer, if the user asked to read more than what the buffer has, then we go into the `None` case which will re-fill the buffer with new data?

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -0,0 +1,1345 @@
...
+                        let bytes = from_buffer.to_vec();
+                        data.fd_position += read.buffer_size;
```
meowjesty, line 309: `fd_position`? Do you mean that this is the cursor position in this buffer for the `fd` of the file we are reading?

mirrord/intproxy/src/request_queue.rs
```diff
@@ -14,44 +14,59 @@
...
-        self.inner.push_back((message_id, layer_id));
+        self.inner
+            .push_back((message_id, layer_id, Default::default()));
+    }
+}
+
+impl<T: fmt::Debug> RequestQueue<T> {
+    /// Retrieve and remove a request from the front of this queue.
+    #[tracing::instrument(level = Level::TRACE)]
+    pub fn get(&mut self) -> Option<(MessageId, LayerId)> {
```
meowjesty, line 56: I'll forever resent that this is called `get` when we have `&mut self` going on. Shouldn't we change this name to `pop_front`, `pop`, `remove`?

mirrord/intproxy/src/request_queue.rs
```diff
@@ -14,44 +14,59 @@
...
+    #[tracing::instrument(level = Level::TRACE)]
+    pub fn insert_with_data(&mut self, message_id: MessageId, layer_id: LayerId, data: T) {
+        self.inner.push_back((message_id, layer_id, data));
     }
 
     /// Retrieve and remove a request from the front of this queue.
     #[tracing::instrument(level = Level::TRACE)]
-    pub fn get(&mut self) -> Result<(MessageId, LayerId), RequestQueueEmpty> {
-        self.inner.pop_front().ok_or(RequestQueueEmpty)
+    pub fn get_with_data(&mut self) -> Option<(MessageId, LayerId, T)> {
```
meowjesty, line 69: Same here, feels weird to call this `get_`.

meowjesty (changes requested):
I mostly have some questions to make sure I'm understanding some refactored stuff correctly.

mirrord/intproxy/src/proxies/files.rs
Razz4780: Yeah, `None` case will send a `ReadLimited` request to the agent for the next chunk

mirrord/intproxy/src/request_queue.rs
Razz4780: Changed to `pop_front[_with_data]`, `push_back[_with_data]`

mirrord/intproxy/src/proxies/files.rs
Razz4780: I mean position of the file descriptor in the file, simulated in here. Normally fd position of remote files is handled in the agent - the agent opens a raw OS file descriptor, position of which is managed by the kernel. Following this approach for buffered files was a bit problematic. So I moved managing fd position to the intproxy (see `File buffering` section of `FilesProxy` doc)

mirrord/intproxy/src/request_queue.rs
meowjesty: Are we objective-c?

meowjesty (approved):
:+1:

PR #2981, opened 2024-12-16
Add node's --inspect debugger to port detection

gememma:


mirrord/layer/src/debugger_ports.rs
```diff
@@ -91,81 +99,142 @@ impl FromStr for DebuggerType {
...
+                            let vec = flag.split('=').collect::<Vec<_>>();
+                            match vec.as_slice() {
+                                // you can use --inspect, -wait and -brk all together at once -
+                                // we need to ignore them all
+                                &[_flag, port] => port.parse::<u16>().ok(),
+                                _ => None,
+                            }
+                        })
+                        .map(|port| SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), port))
+                        .collect::<Vec<SocketAddr>>()
```
Razz4780, line 232: wdyt about this:
```rust
    var.split_ascii_whitespace()
        .filter_map(|flag| match flag.split_once('=')? {
            ("--inspect" | "--inspect-brk" | "--inspect-wait", port) => port.parse::<u16>().ok(),
            _ => None,
        })
        .map(|port| SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port))
        .collect::<Vec<_>>()
```

mirrord/layer/src/debugger_ports.rs
```diff
@@ -194,23 +264,28 @@ impl DebuggerPorts {
...
+                        })
+                        .ok()
+                }) {
+                Some(debugger) => debugger.get_port(&std::env::args().collect::<Vec<_>>()),
+                None => vec![],
+            };
+        if !detected.is_empty() {
+            detected
+                .iter()
+                .for_each(|port| env::set_var(MIRRORD_IGNORE_DEBUGGER_PORTS_ENV, port.to_string()));
```
Razz4780, line 286: Doing this in a loop will just overwrite the env. You'll end up with `MIRRORD_IGNORE_DEBUGGER_PORTS=<last_port>` in your env

Razz4780 (changes requested):

mirrord/layer/src/debugger_ports.rs
Razz4780: Also, can these flags use default values? Like if you encounter just `--inspect`, does it work and use a default port?

mirrord/config/src/util.rs
```diff
@@ -250,7 +250,6 @@ where
     deserializer.deserialize_any(StringOrStruct(PhantomData))
 }
 
-#[cfg(test)]
```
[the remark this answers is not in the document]
gememma: so i can import `with_env_vars()` to use in the test

mirrord/layer/src/debugger_ports.rs
gememma: i think the reason i didnt split on whitespace is because path names can contain whitespace but this is better. As far as defaults go, if you invoke with `--inspect`, the flag gets transformed by node to `--inspect=xyz` with the port number :)

mirrord/config/src/util.rs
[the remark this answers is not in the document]
Razz4780: Hmm, why can't we just use clean functions? Like in other tests (`detect_debugpy_port`, `detect_pydevd_port`, ...)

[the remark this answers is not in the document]
Razz4780: Ah ok, I see now ;_;
I'd rather have it clean tho, so maybe we can change
```rust
impl DebuggerType {
    fn get_port(self, args: &[String] -> Vec<u16> { ... }
}
```
into
```rust
impl DebuggerType {
    fn get_port_2<F: FnMut(&str) -> Option<String>>(self, args: &[String], get_env: F) -> Vec<u16> { ... }
}
```

In prod code you'd call it like
```rust
debugger_type.get_port(&args, |name| std::env::var(name).ok())
```

mirrord/layer/src/debugger_ports.rs
gememma: why do we set this var anyway if we return `Self::Detected(detected)` straight after? when would that value get read?

Razz4780: In children processes. They have different command line, so we wouldn't be able to detect the port

mirrord/layer/src/debugger_ports.rs
```diff
@@ -175,14 +241,16 @@ impl DebuggerType {
...
     }
 }
 
 /// Local ports used by the debugger running the process.
 /// These should be ignored by the layer.
 #[derive(Debug)]
 pub enum DebuggerPorts {
-    Detected(u16),
+    Detected(Vec<u16>),
+    Multiple(Vec<u16>),
```
Razz4780, line 253: Why do we need two variants now? :eyes: 
Shouldn't it be just `Detected(Vec<u16>)`?

mirrord/layer/src/debugger_ports.rs
```diff
@@ -194,23 +262,52 @@ impl DebuggerPorts {
...
+                    }),
+                None => vec![],
+            };
+        if !detected.is_empty() {
+            let mut value = String::new();
+            detected
+                .iter()
+                .for_each(|port| value.push_str(format!("{port},").as_str()));
+            // remove trailing comma
+            value.pop();
```
Razz4780, line 290:
```suggestion
            let value = detected.iter().map(ToString::to_string).collect::<Vec<_>>().join(",");
```
Only a suggestion tho

Razz4780 (approved):
:+1:

mirrord/layer/src/debugger_ports.rs
```diff
@@ -175,15 +276,24 @@ impl DebuggerType {
...
+            _ => None,
+        };
+        (ports, next_type)
     }
 }
 
 /// Local ports used by the debugger running the process.
 /// These should be ignored by the layer.
 #[derive(Debug)]
 pub enum DebuggerPorts {
```
Razz4780, line 293: wdyt about
```rust
pub enum DebuggerPorts {
  Single(u16),
  Range(RangeInclusive<u16>),
  Combination(Vec<DebuggerPorts>),
  None,
}
```

mirrord/layer/src/debugger_ports.rs
```diff
@@ -205,41 +315,68 @@ impl DebuggerPorts {
...
+                                    MIRRORD_IGNORE_DEBUGGER_PORTS_ENV,
+                                );
+                                None
+                            },
+                        };
+                        if let Some(ports) = range {
+                            v.push(Self::FixedRange(ports));
+                        }
+                    });
+                None
```
Razz4780, line 376: I think that at this point we should implement `FromStr` and `fmt::Display` on `DebuggerPorts` :eyes:

mirrord/layer/src/debugger_ports.rs
```diff
@@ -91,81 +111,162 @@ impl FromStr for DebuggerType {
...
+                    })
+                    .map(|port| SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port))
+                    .collect::<Vec<_>>()
+                } else if let Some(value) = get_env("NODE_INSPECTOR_INFO") {
+                    value.split(',').filter_map(|var| match var.split_once(':')? {
+                        ("inspectorURL", url) => url.parse::<Uri>().ok()?.port_u16(),
+                        _ => None,
+                    })
+                    .map(|port| SocketAddr::new(Ipv4Addr::LOCALHOST.into(), port))
+                    .collect::<Vec<_>>()
```
Razz4780, line 264: This env value is a json, right? We could use here sth like
```rust
#[derive(Deserialize)
struct NodeInspectorInfo {
    inspector_url: Uri,
}

serde_json::from_str::<NodeInspectorInfo>(&value)
    .inspect_err(|error| tracing::warn!(%error, value, "Failed to deserialize NODE_INSPECTOR_INFO"))
    .ok()
    .map(|uri| {
        vec![SocketAddr::new(Ipv4Addr::LOCALHOST.into(), uri.port_u16())]
    })
    .unwrap_or_default()
```

Manual parsing is prone to changes in the representation. Like, at some point they might add extra whitespace. It's highly unlikely, so leaving it up to you

Razz4780 (approved):
LGTM
Left some optional suggestions

mirrord/layer/src/debugger_ports.rs
gememma: agreed, this code is gross

mirrord/layer/src/debugger_ports.rs
gememma: i like this but i think it might be overkill here? ill leave it out for now

mirrord/layer/src/debugger_ports.rs
```diff
@@ -175,26 +276,93 @@ impl DebuggerType {
...
+                            "Failed to decode debugger ports from {} env variable: expected a port or range of ports",
+                            MIRRORD_IGNORE_DEBUGGER_PORTS_ENV,
+                        );
+                    },
+                };
+            });
+        if !vec.is_empty() {
+            Ok(Self::Combination(vec))
+        } else {
+            Err(format!("couldn't parse ports from string: {s}"))
```
Razz4780, line 332:
```suggestion
            Ok(Self::None)
```
and `<DebuggerPorts as FromStr>::Err = std::convert::Infallible`

PR #2982, opened 2024-12-16
E2e test for mirrord cluster policy

Razz4780:


t4lz (approved):
Looks good. If you think there is value in verifying stealing works when mirroring is blocked and vice versa (I'm not sure it's necessary), we could add `run_mirrord_and_verify_steal_result` calls to the end of the new test, and/or change `PolicyTestCaes` to also verify expected mirroring capabilities.

PR #2993, opened 2024-12-25
Add policy to exclude env vars.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/103
- Related PRs: https://github.com/metalbear-co/operator/pull/727


mirrord/operator/src/crd/policy.rs
```diff
@@ -78,6 +86,12 @@ pub struct MirrordClusterPolicySpec {
     // TODO: make the k8s list type be set/map to prevent duplicates.
     /// List of features and operations blocked by this policy.
     pub block: Vec<BlockedFeature>,
+
+    /// List of environment variables that should be excluded when using mirrord.
+    ///
+    /// These environment variables won't be retieved from the target even if the user
```
aviramha, line 92: typo - retrieved

mirrord/operator/src/crd/policy.rs
```diff
@@ -78,6 +86,12 @@ pub struct MirrordClusterPolicySpec {
...
+    /// specifies them in their `feature.env.include` mirrord config.
+    pub env_vars_exclude: HashSet<String>,
```
aviramha, line 94: I think `excluded_env_vars` sounds better

aviramha, line 94: is it backward compatible? would policies without env_vars_exclude work?

meowjesty: I agree, but we kinda use "env vars exclude" everywhere else :sweat: 

Like in the env config, and in all internal stuff we have (this one doesn't really matter though). I think this name ends up being more consistent, since `mirrord.json` navigation would also be more similar as it's `feature.env.exclude`.

wdyt? Am I overthinking this name?

meowjesty: Meh, thought it would just have some default value, but it appears in the "required" section.

Changed to `Option`.

Razz4780: I think it works if you mark it with `#[serde(default)]`, no?

mirrord/operator/src/crd/policy.rs
```diff
@@ -51,6 +53,12 @@ pub struct MirrordPolicySpec {
     // TODO: make the k8s list type be set/map to prevent duplicates.
     /// List of features and operations blocked by this policy.
     pub block: Vec<BlockedFeature>,
+
+    /// List of environment variables that should be excluded when using mirrord.
+    ///
+    /// These environment variables won't be retrieved from the target even if the user
+    /// specifies them in their `feature.env.include` mirrord config.
+    pub env_vars_exclude: Option<HashSet<String>>,
```
Razz4780, line 61: Should we go for literal keys or key regexes? :thinking:

meowjesty: I thought we wanted regexes just like the env vars exclude filter thingy? I don't see why we would not want that?

meowjesty: Changed the docs.

mirrord/operator/src/crd/policy.rs
meowjesty: Yup! Removed the `Option`.

mirrord/operator/src/crd/policy.rs
Razz4780, line 61: Requires same update as `MirrordClusterPolicySpec::env_vars_exclude`:
1. No `Option`
2. `#[serde(default)]`
3. Wildcard doc

Razz4780 (changes requested):

mirrord/operator/src/crd/policy.rs
meowjesty: Ooops. Changed.

Razz4780 (approved):

mirrord/operator/src/crd/policy.rs
aviramha: Perhaps to have a better UX it'd be better to make it closer to mirrord.json
so then we'd have `env` and under it `exclude` and perhaps later add include.

meowjesty: That's a good UX idea, I'm also working on the file policies, think I should do the same thing there?

aviramha: Yes

meowjesty: Changed.

mirrord/operator/src/crd/policy.rs
```diff
@@ -51,6 +53,10 @@ pub struct MirrordPolicySpec {
     // TODO: make the k8s list type be set/map to prevent duplicates.
     /// List of features and operations blocked by this policy.
     pub block: Vec<BlockedFeature>,
+
+    /// Allows the mirrord-operator to control which environment variables can affected by mirrord.
```
Razz4780, line 57:
```suggestion
    /// Controls how mirrord-operator handles user requests to fetch environment variables from the target.
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -78,6 +84,25 @@ pub struct MirrordClusterPolicySpec {
...
```
Razz4780, line 88:
```suggestion
    /// Controls how mirrord-operator handles user requests to fetch environment variables from the target.
```

Razz4780 (commented):
suggestions

PR #2997, opened 2025-01-01
Add policy for file ops.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/104
- Related PRs:
  - https://github.com/metalbear-co/operator/pull/729


mirrord/operator/src/crd/policy.rs
```diff
@@ -104,9 +110,26 @@ pub struct EnvPolicy {
...
+#[serde(rename_all = "kebab-case")]
+pub struct FsPolicy {
+    #[serde(default)]
+    pub read_only: HashSet<String>,
+
+    #[serde(default)]
+    pub read_write: HashSet<String>,
+
+    #[serde(default)]
+    pub local: HashSet<String>,
```
aviramha, line 127: What would local do? client will see that and fallback to local?

meowjesty: That's a good idea, but not what I had in mind.

I thought that `local` here would mean that this file has to be read locally, so the operator returns `NotFound` if the user does not put it under `config.fs.local`.

Returning a `OpenLocal` (or something like that) and having mirrord turn this `open` call into a local one would be neat. Should I do this?

aviramha: I'd assume the notfound to be returned on the next proeprty (not_found) - but yeah I like that idea:)

tests/src/operator/policies.rs
t4lz, line 1: How about an E2E test for FS policies?

meowjesty: Since you're the one reviewing this tal, do you see any value in the `read_write` policy? I was thinking that we could be blocking everything that doesn't match `read_only` and `read_write`, but it seems that's not what's supposed to happen, so `read_write` looks pretty useless.

tests/src/operator/policies.rs
```diff
@@ -1,5 +1,5 @@
 #![cfg(test)]
-#![cfg(feature = "operator")]
+// #![cfg(feature = "operator")]
```
t4lz, line 2: why?

tests/src/operator/policies/fs.rs
```diff
@@ -0,0 +1,95 @@
...
+) {
+    let kube_client = kube_client.await;
+    let service = service.await;
+
+    // Create policy, delete it when test exits.
+    // TODO(alex) [high]: Should probably be namespaced, to avoid it infecting other tests?
+    let _policy_guard = PolicyGuard::namespaced(
+        kube_client,
+        &MirrordPolicy::new(
+            "e2e-test-block-mirror-with-path-pattern",
```
t4lz, line 41:
```suggestion
            "e2e-test-fs-policy-with-path-pattern",
```

tests/src/operator/policies/fs.rs
```diff
@@ -0,0 +1,95 @@
...
+#[timeout(Duration::from_secs(60))]
+pub async fn create_cluster_fs_policy_and_try_file_operations(
+    #[future] service: KubeService,
+    #[future] kube_client: kube::Client,
+) {
+    let kube_client = kube_client.await;
+    let service = service.await;
+
+    // Create policy, delete it when test exits.
+    // TODO(alex) [high]: Should probably be namespaced, to avoid it infecting other tests?
```
t4lz, line 37:
```suggestion
```
Isn't it already namespace?

tests/src/operator/policies/fs.rs
```diff
@@ -0,0 +1,95 @@
...
+            Some(&service.namespace),
+            Some(vec!["--fs-mode=write"]),
+            None,
+        )
+        .await;
+
+    test_process
+        .wait_for_line(Duration::from_secs(40), "daemon subscribed")
+        .await;
+
```
t4lz, line 74: But the test app does not listen on any port, does it? 🤔 

```suggestion
```

tests/src/operator/policies/fs.rs
```diff
@@ -0,0 +1,95 @@
...
+    test_process
+        .assert_stdout_contains("SUCCESS /app/file.read-only")
+        .await;
+    test_process
+        .assert_stdout_contains("SUCCESS /app/file.read-write")
+        .await;
+
+    test_process.child.kill().await.unwrap()
```
t4lz, line 94: This is redundant after calling `wait_assert_success` earlier.
```suggestion
    test_process.child.kill().await.unwrap()
```

tests/src/operator/policies.rs
meowjesty: Oops, changed it so I can get rust-analyzer to not ignore these files.

Removed my change.

tests/src/operator/policies/fs.rs
meowjesty: Yeah, removed.

tests/src/operator/policies/fs.rs
meowjesty: Removed.

tests/src/operator/policies.rs
t4lz: I agree, we can remove `read_write`.

t4lz (approved):
👍

tests/src/operator/policies.rs
meowjesty: Removed `read_write`.

Issue #2992, opened 2024-12-24
agent `open` fails with `getaddrinfo` error, instead of proper file error.


meowjesty:
### Bug Description

If you try to `open` a file that doesn't exist in the target pod, the agent fn `FileManager::open` logs the error as `IO failed for remote operation with `Failed performing `getaddrinfo` with Some(2) and kind NotFound!!`, even though this has nothing to do with `getaddrinfo`.

### Steps to Reproduce

1. Local process tries to `open` a file remotely;
2. File does not exist in the target pod;
3. Wrong error is reported in the agent;

### Your operating system and version

Linux

### Local process

node sample that opens a file

PR #3001, opened 2025-01-02
Change misleading error when agent fails IO operation.

meowjesty:
- Issue: #2992

### DIFF
--- a/mirrord/protocol/src/error.rs
+++ b/mirrord/protocol/src/error.rs
@@ -154,7 +154,7 @@ impl From<AddrParseError> for RemoteError {
 /// Our internal version of Rust's `std::io::Error` that can be passed between mirrord-layer and
 /// mirrord-agent.
 #[derive(Encode, Decode, Debug, PartialEq, Clone, Eq, Error)]
-#[error("Failed performing `getaddrinfo` with {raw_os_error:?} and kind {kind:?}!")]
+#[error("Failed IO operation with {raw_os_error:?} and kind {kind:?}!")]
 pub struct RemoteIOError {
     pub raw_os_error: Option<i32>,
     pub kind: ErrorKindInternal,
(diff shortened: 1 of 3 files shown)


mirrord/protocol/src/error.rs
```diff
@@ -154,7 +154,7 @@ impl From<AddrParseError> for RemoteError {
 /// Our internal version of Rust's `std::io::Error` that can be passed between mirrord-layer and
 /// mirrord-agent.
 #[derive(Encode, Decode, Debug, PartialEq, Clone, Eq, Error)]
-#[error("Failed performing `getaddrinfo` with {raw_os_error:?} and kind {kind:?}!")]
+#[error("Failed IO operation with {raw_os_error:?} and kind {kind:?}!")]
```
Razz4780, line 157: The complete message for `ResponseError::RemoteIO` still looks a bit ugly: 
```
IO failed for remote operation with `Failed IO operation with None and kind AddrInUse!!
```
Maybe sth like:

```rust
#[derive(Encode, Decode, Debug, PartialEq, Clone, Eq, Error)]
pub struct RemoteIOError {
    pub raw_os_error: Option<i32>,
    pub kind: ErrorKindInternal,
}

impl fmt::Display for RemoteIOError {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        write!(f, "{:?}", self.kind)?;
        if let Some(code) = self.raw_os_error {
            write!(f, " (error code {code})")?;
        }
        Ok(())
    }
}

...

enum ResponseError {
    ...
    #[error("IO failed for remote operation: {0}!")]
    RemoteIO(RemoteIOError),
}
```
This would produce:
```
IO failed for remote operation: AddrInUse!
```
or
```
IO failed for remote operation: AddrInUse (error code 98)!
```

meowjesty: Nice! Changed it.

Razz4780: Did you push? 👀

meowjesty: [Should I seek help?](https://www.mayoclinic.org/diseases-conditions/alzheimers-disease/in-depth/memory-loss/art-20046326)

Razz4780 (approved):

PR #3003, opened 2025-01-03
Add namespaces to `mirrord ls` output

Razz4780:
Issue #2999

Since `mirrord ls` currently returns a list of target paths, it was not possible to implement this without adding an extra env. IDEs will have to inspect mirrord binary version or try parsing the new format with fallback to the old one.

Additionally, fixed `mirrord::port_forward` module visibility and removed some dead code.


mirrord/cli/src/list.rs
```diff
@@ -0,0 +1,208 @@
...
+#[derive(Serialize)]
+struct FoundTargets {
+    /// In order:
+    /// 1. deployments
+    /// 2. rollouts
+    /// 3. statefulsets
+    /// 4. cronjobs
+    /// 5. jobs
+    /// 6. pods
+    targets: Vec<FoundTarget>,
```
meowjesty, line 49: Wouldn't it be better if this was a `HashSet` so we get no duplicates by default?

mirrord/cli/src/list.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    /// Current lookup namespace.
+    ///
+    /// Taken from [`LayerConfig::target`], defaults to [`Client`]'s default namespace.
+    current_namespace: String,
+
+    /// Available lookup namespaces.
+    namespaces: Vec<String>,
```
meowjesty, line 57: Same `HashSet` suggestion here.

meowjesty (changes requested):
Some nits here and there.

The main thing is, I would prefer to have kube interactions under the kube resource struct, as I find it's easier to think about (my thought process is usually: I want kube stuff, create this struct, now it offers me an api with exactly what I want ... Oh it doesn't have what I want? I add it to the struct). But we never talked about api design (not that I remember), maybe we should bring this up to the team?

mirrord/cli/src/list.rs
```diff
@@ -0,0 +1,208 @@
...
+        list.end()
+    }
+}
+
+/// Controls whether we support listing all targets or just the open source ones.
+static ALL_TARGETS_SUPPORTED_OPERATOR_VERSION: LazyLock<VersionReq> =
+    LazyLock::new(|| ">=3.84.0".parse().expect("verion should be valid"));
+
+/// Fetches mirrord targets from the cluster and prints output to stdout.
+pub async fn print_targets(args: ListTargetArgs, rich_output: bool) -> CliResult<()> {
```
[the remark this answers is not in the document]
Razz4780: Can be `pub(super)`, yeah

mirrord/cli/src/list.rs
meowjesty: Yeah it would have to be `iter().sort()`, which seems annoying, considering it's just serializing this.

Nvm then.

meowjesty: Oh! https://doc.rust-lang.org/std/collections/struct.BTreeSet.html

Razz4780: This would require custom `PartialEq` and `PartialOrd` implementations, can we keep this simple? :V

mirrord/cli/src/list.rs
Razz4780: I'd keep it simpleeee

mirrord/cli/src/list.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// Right now this is always true. Some preliminary checks are done in the
+    /// [`KubeResourceSeeker`] and results come filtered.
+    ///
+    /// This field is here for forward compatibility, because in the future we might want to return
+    /// unavailable targets as well (along with some validation error message) to improve UX.
+    available: bool,
+}
+
+/// Result of mirrord targets lookup in the cluster.
```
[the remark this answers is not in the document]
Razz4780: Added docs

mirrord/cli/src/list.rs
```diff
@@ -0,0 +1,208 @@
...
+            targets,
+            current_namespace,
+            namespaces,
+        })
+    }
+}
+
+/// Thin wrapper over [`FoundTargets`] that implements [`Serialize`].
+/// Its serialized format is a sequence of available target paths.
+struct SimpleDisplay<'a>(&'a FoundTargets);
```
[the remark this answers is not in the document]
Razz4780: Changed name to `FoundTargetsList`, since it's not really about `fmt::Display` :V

mirrord/cli/src/list.rs
meowjesty: Sure.

I thought it would be nice to guarantee uniqueness, and ordering with the btreeset, but this is not a thing that we use anywhere else right? So whatever you want I'm fine with.

mirrord/kube/src/api/kubernetes/list.rs
```diff
@@ -0,0 +1,93 @@
...
+            params.continue_token.replace(continue_token);
+        }
+    }
+}
+
+/// Returns a [`Stream`] of all objects in the cluster.
+///
+/// 1. `field_selector` can be used for filtering.
+/// 2. Our own resources are excluded.
+pub fn list_all_clusterwide<R>(
```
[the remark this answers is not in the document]
Razz4780: Oke, moved both listing functions to `KubeResourceSeeker`

mirrord/cli/src/list.rs
Razz4780: Yeah, this is private and only for this printout. The struct definition is basically just for `Serialize` derive.

mirrord/cli/src/list.rs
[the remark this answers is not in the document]
meowjesty: But isn't it about "a simpler way to display"?

meowjesty (approved):
And here you prefer present for all the docs. I feel trolled.

:troll:

mirrord/cli/src/list.rs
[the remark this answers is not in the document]
Razz4780: Kind of, but more like "a simpler way to serialize". I feared confusion with `fmt::Display`

PR #3007, opened 2025-01-07
fix: add loopback exclusion rule to prevent local port loops

jboykin-bread:


mirrord/agent/src/steal/ip_tables/mesh.rs
```diff
@@ -101,11 +92,9 @@ where
     }
 
     async fn add_redirect(&self, redirected_port: Port, target_port: Port) -> Result<()> {
-        if self.vendor != MeshVendor::IstioCni {
-            self.prerouting
-                .add_redirect(redirected_port, target_port)
-                .await?;
-        }
+        self.prerouting
```
aviramha, line 95: I can't see how that would work, or specifically how it'd fix the issue - you're actually redirecting another port by having this change, which wouldn't affect outbound connections, and if it would it'll send those to the agent which sounds like an issue

PR #3010, opened 2025-01-09
Fix fs policy test.

meowjesty:


t4lz (approved):

changelog.d/+104-policy-fs-test.fixed.md
t4lz, line 1: Should be internal if at all. Actually there shouldn't even be a changelog entry, since there was not a release since the test was added, so fixing the test is not a change that will be introduced in the next release. So either delete this file and just make some small change to another towncrier entry file just to pass the check (imo the way to go), or if you really want to have this entry, change it to `internal`, as users don't care we fixed a test.

meowjesty: Changed the previous changelog.

PR #3011, opened 2025-01-10
Prevent config rendering on every process

### LINKED ISSUE #2936: mirrord + vscode + templating seems to get ide to hang
when running a launch configuration as follows (next dev project, node 18, macOS)
```json
 {  "version": "0.2.0",
  "configurations": [   {
      "name": "NEW",
      "request": "launch",
      "runtimeArgs": [
        "next",
        "dev"
      ],
      "runtimeExecutable": "npx",
      "skipFiles": [
        "<node_internals>/**"
      ],
      "env": {
        "NODE_OPTIONS": "--inspect"
      },
      "type": "node",
      "outFiles": [
        "${workspaceFolder}/x/**/*.(m|c|)js",
        "!**/node_modules/**"
      ],
      "cwd": "${workspaceFolder}/apps/x",
    }
  ]
}
```

mirrord.json:
```
{
  "target": {
    "path": "x",
    "namespace": "y"
  },

  "feature": {
    "env": {
      "override": {
        "NODE_EXTRA_CA_CERTS": "{{ get_env(name="HOMEBREW_PREFIX", default="/opt/homebrew") }}/etc/ca-certificates/cert.pem",
        "NODE_ENV": "development"
      }
    },
    "fs": false,
    "network": {
      "incoming": {
        "mode": "steal",
        "http_filter": {
          "header_filter": "x-intercept: {{ get_env(name="USER") }}"
        },
        "port_mapping": [[4003, 8080]]
      },
      "outgoing": false
    }
  },
  "operator": true
}
```

gememma:
Closes #2936

### DIFF
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -289,10 +289,9 @@ impl MirrordExecution {
             })?;
 
         // Provide details for layer to connect to agent via internal proxy
-        env_vars.insert(
-            MIRRORD_CONNECT_TCP_ENV.to_string(),
-            format!("127.0.0.1:{}", address.port()),
-        );
+        let mut config = config.clone();
+        config.connect_tcp = Some(format!("127.0.0.1:{}", address.port()));
+        config.update_env()?;
 
         // Fixes <https://github.com/metalbear-co/mirrord/issues/1745>
         // by disabling the fork safety check in the Objective-C runtime.
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -177,7 +177,7 @@ impl MirrordExecution {
     /// [`tokio::time::sleep`] or [`tokio::task::yield_now`] after calling this function.
     #[tracing::instrument(level = Level::TRACE, skip_all)]
     pub(crate) async fn start<P>(
-        config: &LayerConfig,
+        config: &mut LayerConfig,
         // We only need the executable on macos, for SIP handling.
         #[cfg(target_os = "macos")] executable: Option<&str>,
         progress: &mut P,
@@ -289,9 +289,8 @@ impl MirrordExecution {
             })?;
 
         // Provide details for layer to connect to agent via internal proxy
-        let mut config = config.clone();
         config.connect_tcp = Some(format!("127.0.0.1:{}", address.port()));
-        config.update_env()?;
+        config.update_env_var()?;
 
         // Fixes <https://github.com/metalbear-co/mirrord/issues/1745>
         // by disabling the fork safety check in the Objective-C runtime.
--- a/mirrord/config/Cargo.toml
+++ b/mirrord/config/Cargo.toml
@@ -17,8 +17,8 @@ edition.workspace = true
 workspace = true
 
 [dependencies]
-mirrord-config-derive = { path = "./derive"}
-mirrord-analytics = { path = "../analytics"}
+mirrord-config-derive = { path = "./derive" }
+mirrord-analytics = { path = "../analytics" }
 
 serde.workspace = true
 serde_json.workspace = true
@@ -27,13 +27,14 @@ tracing.workspace = true
 serde_yaml.workspace = true
 toml = "0.8"
 schemars.workspace = true
-bimap = "0.6"
+bimap = { version = "0.6", features = ["serde"] }
 nom = "7.1"
 ipnet.workspace = true
 bitflags = "2"
 k8s-openapi = { workspace = true, features = ["schemars", "earliest"] }
 tera = "1"
 fancy-regex.workspace = true
+base64.workspace = true
 
 [dev-dependencies]
 rstest.workspace = true
--- a/mirrord/config/Cargo.toml
+++ b/mirrord/config/Cargo.toml
@@ -27,7 +27,7 @@ tracing.workspace = true
 serde_yaml.workspace = true
 toml = "0.8"
 schemars.workspace = true
-bimap = { version = "0.6", features = ["serde"] }
+bimap = { version = "0.6" }
 nom = "7.1"
 ipnet.workspace = true
 bitflags = "2"
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -67,7 +67,7 @@ impl fmt::Display for LinuxCapability {
 ///   }
 /// }
 /// ```
-#[derive(MirrordConfig, Clone, Debug, Serialize)]
+#[derive(MirrordConfig, Clone, Debug, Serialize, Deserialize)]
 #[config(map_to = "AgentFileConfig", derive = "JsonSchema")]
 #[cfg_attr(test, config(derive = "PartialEq"))]
 pub struct AgentConfig {
@@ -477,7 +477,7 @@ impl AgentFileConfig {
     }
 }
 
-#[derive(MirrordConfig, Default, PartialEq, Eq, Clone, Debug, Serialize)]
+#[derive(MirrordConfig, Default, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
 #[config(derive = "JsonSchema")]
 #[cfg_attr(test, config(derive = "PartialEq, Eq"))]
 pub struct AgentDnsConfig {
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -354,7 +354,7 @@ pub struct AgentConfig {
     /// Create an agent that returns an error after accepting the first client. For testing
     /// purposes. Only supported with job agents (not with ephemeral agents).
     #[cfg(all(debug_assertions, not(test)))] // not(test) so that it's not included in the schema json.
-    #[serde(skip_serializing)]
+    #[serde(skip_serializing, skip_deserializing)]
     #[config(env = "MIRRORD_AGENT_TEST_ERROR", default = false, unstable)]
     pub test_error: bool,
 }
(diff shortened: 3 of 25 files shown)


mirrord/config/Cargo.toml
```diff
@@ -17,8 +17,8 @@ edition.workspace = true
 workspace = true
 
 [dependencies]
-mirrord-config-derive = { path = "./derive"}
-mirrord-analytics = { path = "../analytics"}
+mirrord-config-derive = { path = "./derive" }
+mirrord-analytics = { path = "../analytics" }
```
meowjesty, line 21: For a change this big I require an explanation.

mirrord/config/src/feature/network/incoming.rs
```diff
@@ -306,19 +306,6 @@ pub struct IncomingAdvancedFileConfig {
     pub ports: Option<Vec<u16>>,
 }
 
-fn serialize_bi_map<S>(map: &BiMap<u16, u16>, serializer: S) -> Result<S::Ok, S::Error>
```
meowjesty, line 309: Results in the same thing as whatever `bimap` serde does right? Do we have config tests to check that this wouldn't be breaking config?

mirrord/config/src/lib.rs
```diff
@@ -327,15 +331,50 @@ pub struct LayerConfig {
 }
 
 impl LayerConfig {
+    /// Given an encoded complete config from the [`MIRRORD_RESOLVED_CONFIG_ENV`]
+    /// env var, attempt to decode it into [`LayerConfig`].
+    /// Intended to avoid re-resolving the config in every process mirrord in loaded into
```
meowjesty, line 336:
```suggestion
    /// Intended to avoid re-resolving the config in every process mirrord is loaded into.
```

mirrord/config/src/lib.rs
```diff
@@ -327,15 +331,50 @@ pub struct LayerConfig {
...
+        Ok(serde_json::from_str::<Self>(serialized)?)
+    }
+
+    pub fn to_env_var(&self) -> Result<String, ConfigError> {
+        let serialized = serde_json::to_string(self)
+            .map_err(|error| ConfigError::EnvVarEncodeError(error.to_string()))?;
+        Ok(BASE64_STANDARD.encode(serialized))
+    }
+
+    pub fn update_env(&self) -> Result<(), ConfigError> {
```
meowjesty, line 352:
```suggestion
    pub fn update_env_var(&self) -> Result<(), ConfigError> {
```

Nit, since the others are `*_env_var`.

mirrord/config/src/lib.rs
```diff
@@ -986,4 +1025,20 @@ mod tests {
 
         assert_eq!(existing_content.replace("\r\n", "\n"), compare_content);
     }
+
+    /// related to issue #2936: https://github.com/metalbear-co/mirrord/issues/2936
+    /// checks that resolved config written to [`MIRRORD_RESOLVED_CONFIG_ENV`] can be
+    /// transformed back into a [`LayerConfig`]
+    #[test]
+    fn encode_and_decode_config() {
```
meowjesty, line 1033: I wonder if we should have a test value with a filled config too, instead of just default values.

Wdyt?

mirrord/config/src/lib.rs
```diff
@@ -327,15 +331,50 @@ pub struct LayerConfig {
...
+    pub fn from_env_var(encoded_value: String) -> Result<Self, ConfigError> {
+        let decoded = BASE64_STANDARD
+            .decode(encoded_value)
+            .map_err(|error| ConfigError::EnvVarDecodeError(error.to_string()))?;
+        let serialized = std::str::from_utf8(&decoded)
+            .map_err(|error| ConfigError::EnvVarDecodeError(error.to_string()))?;
+        Ok(serde_json::from_str::<Self>(serialized)?)
+    }
+
+    pub fn to_env_var(&self) -> Result<String, ConfigError> {
```
meowjesty, line 346: Docs for this new friend.

mirrord/config/src/lib.rs
meowjesty, line 352: Docs for this new friend.

mirrord/config/src/lib.rs
```diff
@@ -327,15 +331,50 @@ pub struct LayerConfig {
 }
 
 impl LayerConfig {
+    /// Given an encoded complete config from the [`MIRRORD_RESOLVED_CONFIG_ENV`]
+    /// env var, attempt to decode it into [`LayerConfig`].
+    /// Intended to avoid re-resolving the config in every process mirrord in loaded into
+    pub fn from_env_var(encoded_value: String) -> Result<Self, ConfigError> {
```
meowjesty, line 337: Do we need this to be `pub`?

mirrord/config/src/lib.rs
meowjesty, line 346: Do we need this to be `pub`?

mirrord/cli/src/execution.rs
```diff
@@ -289,10 +289,9 @@ impl MirrordExecution {
             })?;
 
         // Provide details for layer to connect to agent via internal proxy
-        env_vars.insert(
-            MIRRORD_CONNECT_TCP_ENV.to_string(),
-            format!("127.0.0.1:{}", address.port()),
-        );
+        let mut config = config.clone();
```
meowjesty, line 292: Instead of `clone` to `mut` here, why not take this as `&mut` in the function args?

meowjesty (changes requested):
Just some small stuff.

So if we save the config to an env var, we don't get the templating breaking on `get_env` secondary calls? Since we only do it once, is that what this PR is solving?

gememma:
> So if we save the config to an env var, we don't get the templating breaking on get_env secondary calls? Since we only do it once, is that what this PR is solving?

@meowjesty  That's correct - the config template variables get resolved when the config file is rendered by Tera, which previously happened on every process. This causes an issue when a process tries to, for example, call `get_env()` to render the config but the process doesnt have access to the same env as previous ones - doing the `get_env()` call only once will solve this :) plus I think this way is more sensible in terms of reducing pointless recalculation and having fewer sources of truth

mirrord/config/src/feature/network/incoming.rs
gememma: we have the cli sanity tests in `tests/src/cli/sanity.rs` to check for breaking changes to config, but I dont really see how this _could_ break anything unless a user was interacting with an already serialised config instead of a json/ whatever format config file and env vars?

mirrord/config/src/agent.rs
```diff
@@ -354,7 +354,7 @@ pub struct AgentConfig {
     /// Create an agent that returns an error after accepting the first client. For testing
     /// purposes. Only supported with job agents (not with ephemeral agents).
     #[cfg(all(debug_assertions, not(test)))] // not(test) so that it's not included in the schema json.
-    #[serde(skip_serializing)]
+    #[serde(skip_serializing, skip_deserializing)]
```
DmitryDodzin, line 357:
```suggestion
    #[serde(skip)]
```

mirrord/config/src/feature/network/incoming.rs
meowjesty: Alright, I was just afraid of this changing something weird that we might've been doing somewhere.

mirrord/cli/src/execution.rs
DmitryDodzin, line 292: Im a bit worried for container feature behaviour where we split up the env variables for sidecar and main container relying on `MIRRORD_CONNECT_TCP_ENV` as env value

mirrord/config/src/lib.rs
gememma: good idea, i can hit some more unusual cases 🫡

mirrord/config/src/lib.rs
```diff
@@ -327,15 +331,50 @@ pub struct LayerConfig {
...
+    }
+
+    pub fn to_env_var(&self) -> Result<String, ConfigError> {
+        let serialized = serde_json::to_string(self)
+            .map_err(|error| ConfigError::EnvVarEncodeError(error.to_string()))?;
+        Ok(BASE64_STANDARD.encode(serialized))
+    }
+
+    pub fn update_env(&self) -> Result<(), ConfigError> {
+        std::env::set_var(MIRRORD_RESOLVED_CONFIG_ENV, self.to_env_var()?);
```
DmitryDodzin, line 353: There might be a problem with using env variables due to size (especially if tls certs are included),my local test macos can handle `~512mb` but can possibly cause a problem on some linux distribution

DmitryDodzin (commented):
I would check `mirrord container` due to changes to `MIRRORD_CONNECT_TCP_ENV` in exec, it should work but I would double check

mirrord/config/src/lib.rs
meowjesty: Good point. Think we could save it to `/tmp` and refresh it from there?

gememma: this honestly didnt even occur to me 🙃 @aviramha @Razz4780 any opinions on this? i feel like there was maybe a reason to avoid using `/tmp` ?

gememma: I think this can be left as-is for the moment, but if it ever becomes a problem we can compress the config or something else 👀

mirrord/cli/src/execution.rs
gememma: shouldnt it get passed through by the config anyway? the only time the env var is accessed is when a new config is created so the value should already be in there, right?

meowjesty (approved):
:earth_americas:  :t-rex:

PR #3012, opened 2025-01-10
Fixed fs policy E2E test

Razz4780:
1. The test was using a bad target path
2. Added some messages to assertions and extra debug prints in test utils
3. Changed case of `FsPolicy` fields to camelCase, because everything in our crds is camelCase
4. Changed `FsPolicy` docs a bit


meowjesty (approved):
:+1:

PR #3016, opened 2025-01-13
Fix hanging requests with filtered steal

Razz4780:
So...

This started as a small refactor in order with some hope to fix the hanging requests issue. I could not find any bug that could cause the problem and only later I found out that there was no problem. The repro application I was using was handling only one connection at a time and the first HTTP connection was not closed by the k8s proxy (most probably to be reused later). And so the second request would hang on intproxy's HTTP handshake attempt. Since we want to be user friendly, this PR introduces reusing local HTTP connections, which solves the problem. However, since it started as a refactor, it's big. Sorry.

Changes summarized:
1. `StreamingBody` was moved from `mirrord-protocol` to `mirrord-intproxy` without any notable changes. There was no need for it to be in the protocol crate.
2. `BodyExt` trait in `mirrord-protocol` was renamed to `BatchedBody`. The only notable change is moving from custom `Future` implementation (`FramesFut`) to using `now_or_never`. I was afraid of it in the past, now I'm not. I tested this with heavy load and did not detect any difference. Using `now_or_never` simplifies things, because some code no longer needs to be async
3. All requests in the intproxy are now of `HttpRequest<StreamingBody>` type, to remove ugly generics and match expressions. `HttpRequestFallback` enum, along with lots of conversion code, was removed from `mirrord-protocol`.
4. `HttpResponseFallback` type was moved to the agent without any notable changes. There was no need for it to be in the protocol crate.
5. `ReversePortForwarder` and its tests were fixed. It was never streaming responses' bodies, because `IncomingProxy` was not notified about agent protocol version. This change is not related to the issue, but the problem came up in the CI.
6. Removed `h2::Error::is_reset` check and the dependency on `h2` completely. Instead of checking if the HTTP error is transient, we check if it's **not** transient (using `hyper::Error` methods, e.g `hyper::Error::is_user`). I think it's simpler and safer, since retrying a request is not harmful.
7. Added a simple `BoundTcpSocket` struct to the incoming proxy, which wraps logic for binding the same interface as user socket. Now we can actually see the bound socket address in tracing.
8. Added a `ClientStore` struct that caches unused local HTTP connection and cleans them up after some timeout.
9. HTTP requests stolen with a filter are now handled completely independently on the local side. Since HTTP is stateless, this is fine. Each HTTP request has its own dedicated `HttpGatewayTask` inside `IncomingProxy`. To reuse connections, they share a `ClientStore` instance.
10. Improved how connections stolen/mirrored in whole are now handled in the `IncomingProxy`. Each connection is handled by its own `TcpProxyTask`. The task knows whether the connection is stolen or mirrored. If it's mirrored, the data is no longer being sent to the main `IncomingProxy` task, it is immediately discarded. If it's stolen, the connection is no longer artificially kept alive until silent for a second (this mechanism makes sense only with the mirror mode, can introduce weird behavior in steal mode).
11. `Interceptor` task removed completely, now we have two separate tasks: `HttpGatewayTask` and `TcpProxyTask`
12. `MetadataStore` was moved to its own module without any notable changes
13. Added a unit test that verifies connection reuse (original issue)
14. `IncomingProxy` now optimizes HTTP response variant. If the whole response body is available when the response head is received, we no longer send the chunked response variant. Instead we respond with the framed variant. This allows us to use only one `mirrord_protocol` message.
15. `IncomingProxy` now does subscription checks when receiving a new connection/request. If we receive a connection/request on a remote port that we no longer subscribe, we unsubscribe immediately, without attempting to connect to the user application.
16. Improved tracing around `IncomingProxy`, e.g added time spent on polling response frames


mirrord/agent/src/steal/http.rs
```diff
@@ -3,11 +3,12 @@
...
 
-pub use filter::HttpFilter;
-
-pub(crate) use self::reversible_stream::ReversibleStream;
+pub(crate) use filter::HttpFilter;
+pub(crate) use response_fallback::{HttpResponseFallback, ReceiverStreamBody};
+pub(crate) use reversible_stream::ReversibleStream;
 
 /// Handy alias due to [`ReversibleStream`] being generic, avoiding value mismatches.
 pub(crate) type DefaultReversibleStream = ReversibleStream<{ HttpVersion::MINIMAL_HEADER_SIZE }>;
```
meowjesty, line 14: Considering that this refactor is already changing a bunch of stuff around these areas, what do you think of we dropping most of this in favor of [`TcpStream::peek`](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.peek)?

mirrord/protocol/src/batched_body.rs
```diff
@@ -0,0 +1,86 @@
...
+                    frames.frames.push(result?);
+                }
+            }
+        }
+
+        Ok(frames)
+    }
+}
+
+pub struct Frames<D> {
```
meowjesty, line 83: Please add some docs, and while at it, either put in the docs what `D` stands for, or change the parameter to `Data` or something byte related.

mirrord/protocol/src/batched_body.rs
```diff
@@ -0,0 +1,86 @@
...
+                    break;
+                }
+                Some(Some(result)) => {
+                    frames.frames.push(result?);
+                }
+            }
+        }
+
+        Ok(frames)
+    }
```
meowjesty, line 80: Considering this is pretty much the same as the other method, maybe create a helper function that takes the `Frames` we're pushing?

meowjesty (changes requested):
Mostly some doc requests while I'm still going through stuff.

mirrord/intproxy/src/background_tasks.rs
```diff
@@ -35,6 +36,21 @@ impl<T: BackgroundTask> MessageBus<T> {
             msg = self.rx.recv() => msg,
         }
     }
+
+    pub fn closed(&self) -> Closed<T> {
+        Closed(self.tx.clone())
+    }
+}
+
+pub struct Closed<T: BackgroundTask>(Sender<T::MessageOut>);
```
meowjesty, line 45: Docs.

Should probably add why we're using this, and how it helps.

mirrord/intproxy/src/background_tasks.rs
```diff
@@ -35,6 +36,21 @@ impl<T: BackgroundTask> MessageBus<T> {
...
+
+impl<T: BackgroundTask> Closed<T> {
+    pub async fn cancel_on_close<F: Future>(&self, future: F) -> Option<F::Output> {
```
meowjesty, line 48: Docs

mirrord/intproxy/src/background_tasks.rs
meowjesty: Do we need this stuff to be `pub`?

mirrord/intproxy/src/background_tasks.rs
meowjesty: Do we need this to be `pub`?

meowjesty (changes requested):
Moar stuff

mirrord/agent/src/steal/http/response_fallback.rs
```diff
@@ -0,0 +1,58 @@
...
+            HttpResponseFallback::Fallback(req) => req.request_id,
+            HttpResponseFallback::Streamed(req) => req.request_id,
+        }
+    }
+
+    pub fn into_hyper<E>(self) -> Response<BoxBody<Bytes, E>> {
+        match self {
+            HttpResponseFallback::Framed(req) => req
+                .internal_response
+                .map_body(|body| body.map_err(|_| unreachable!()).boxed())
```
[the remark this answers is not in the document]
Razz4780: Because we need a specific body type. Some requests are passed to the clients, and the response body implements `Body<Data = Bytes, Error = Infallible>`. Other requests are passed to the original destination, and the response body implements `Body<Data = Bytes, Error = hyper::Error>`. With this trick we unify (the function passed to `Body::map_err` is of type `Fn(Infallible) -> hyper::Error`

mirrord/agent/src/steal/http.rs
Razz4780: Unfortunately, we can't use peek. It may return too few bytes for us to determine connection type, and `Successive calls return the same data` :<

mirrord/cli/src/port_forward.rs
```diff
@@ -609,21 +580,18 @@ impl ReversePortForwarder {
                     )
                 }
             },
-            (MainTaskId::IncomingProxy, TaskUpdate::Finished(result)) => match result {
+
+            TaskUpdate::Finished(result) => match result {
                 Ok(()) => {
```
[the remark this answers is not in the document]
Razz4780: We should never, unless there's a bug in the `IncomingProxy`. I'm up for doing `unreachable!` here, if you think it's better

mirrord/intproxy/src/proxies/incoming/http/response_mode.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// and the previous variants.
+    Chunked,
+    /// Agent supports
+    /// [`LayerTcpSteal::HttpResponseFramed`](mirrord_protocol::tcp::LayerTcpSteal::HttpResponseFramed)
+    /// and the previous variant.
+    Framed,
+    /// Agent supports only
+    /// [`LayerTcpSteal::HttpResponse`](mirrord_protocol::tcp::LayerTcpSteal::HttpResponse)
+    #[default]
+    Basic,
```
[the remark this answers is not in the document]
Razz4780: I don't see a reason to drop this now. Like, this introduces ~100 lines of code in the whole project maybe :man_shrugging:

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Reuses or creates a new [`LocalHttpClient`].
+    #[tracing::instrument(level = Level::TRACE, skip(self), ret, err(level = Level::WARN))]
+    pub async fn get(
+        &self,
+        server_addr: SocketAddr,
+        version: Version,
+    ) -> Result<LocalHttpClient, LocalHttpError> {
+        let ready = {
+            let mut guard = self.clients.lock().unwrap();
```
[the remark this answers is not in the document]
Razz4780: This will pretty much never happen, unless there's a bug in std/tokio. And if it happens, it will make the whole local HTTP unusable (because data under the mutex will be in undefined state and you'd need `unsafe` to handle this case). This panic will make the whole `HttpGatewayTask` panic, which will result in an error log in the `IncomingProxy`. Like, I don't think we need to add extra code to cover this case :sweat_smile:

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    loop {
+        let Some(clients) = clients.upgrade() else {
+            break;
+        };
+
+        let now = Instant::now();
+        let mut min_last_used = None;
+        let notified = {
+            let mut guard = clients.lock().unwrap();
```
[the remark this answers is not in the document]
Razz4780: Done

mirrord/cli/src/port_forward.rs
```diff
@@ -1044,90 +1011,148 @@ mod test {
...
+        }
+
+        async fn auto_responder(
+            mut rx: mpsc::Receiver<ClientMessage>,
+            tx_to_test_code: mpsc::Sender<ClientMessage>,
+            tx_to_port_forwarder: mpsc::Sender<DaemonMessage>,
+        ) {
+            loop {
+                let Some(message) = rx.recv().await else {
+                    break;
```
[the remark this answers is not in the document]
Razz4780: Done, `unwrap`ing on send errors

mirrord/protocol/src/batched_body.rs
Razz4780: Done

mirrord/protocol/src/batched_body.rs
Razz4780: Done

changelog.d/3013.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed an issue where HTTP requests stolen with a filter would hang with a single-threaded local HTTP server.
```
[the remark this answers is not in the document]
Razz4780: Done

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/// Cleans up stale [`LocalHttpClient`]s from the [`ClientStore`].
+async fn cleanup_task(store: ClientStore, idle_client_timeout: Duration) {
+    let clients = Arc::downgrade(&store.clients);
+    let notify = store.notify.clone();
+    std::mem::drop(store);
+
+    loop {
+        let Some(clients) = clients.upgrade() else {
```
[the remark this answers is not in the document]
Razz4780: Added comments

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
```diff
@@ -0,0 +1,213 @@
...
+            break;
+        };
+
+        let now = Instant::now();
+        let mut min_last_used = None;
+        let notified = {
+            let mut guard = clients.lock().unwrap();
+            let notified = notify.notified();
+            guard.retain(|client| {
+                if client.last_used + idle_client_timeout > now {
```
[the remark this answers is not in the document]
Razz4780: Added comments

mirrord/agent/src/steal/api.rs
```diff
@@ -40,7 +39,7 @@ pub(crate) struct TcpStealerApi {
     /// View on the stealer task's status.
     task_status: TaskStatus,
 
-    response_body_txs: HashMap<(ConnectionId, RequestId), Sender<hyper::Result<Frame<Bytes>>>>,
+    response_body_txs: HashMap<(ConnectionId, RequestId), ResponseBodyTx>,
```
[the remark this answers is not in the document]
Razz4780: Added a doc, wdyt?

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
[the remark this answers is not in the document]
Razz4780: Removed the `unwrap` from the cleanup background task tho, it wasn't needed there.

mirrord/intproxy/src/background_tasks.rs
Razz4780: Added docs.

>Do we need this stuff to be `pub`?

Yeah, why not? We use it in other modules

mirrord/intproxy/src/background_tasks.rs
Razz4780: Yeah why, not? We use it in other modules

mirrord/agent/src/steal/http.rs
meowjesty: Dang. I swear that a couple of months after we did our own `ReversibleStream`, tokio added or made `pub` their own thing, but I can't remember the name.

mirrord/cli/src/port_forward.rs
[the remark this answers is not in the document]
meowjesty: Yeah, `unreachable` with a nice message seems better, makes it more obvious to us that something went terribly wrong if we ever need to debug it.

mirrord/intproxy/src/proxies/incoming/http/client_store.rs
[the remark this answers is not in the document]
meowjesty: Alright so it's kinda pointless then.

mirrord/intproxy/src/background_tasks.rs
meowjesty: It's just that `pub` ignores "unused", so when we change stuff we end up keeping the unused pub stuff ...

mirrord/intproxy/src/proxies/incoming/http/response_mode.rs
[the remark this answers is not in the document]
meowjesty: It was worth a try asking to remove this legacy unused stuff.

meowjesty (changes requested):
Even moar stuff

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
-    V1(http1::SendRequest<BoxBody<Bytes, Infallible>>),
-    V2(http2::SendRequest<BoxBody<Bytes, Infallible>>),
+/// An HTTP client used to pass requests to the user application.
+pub struct LocalHttpClient {
+    /// Established HTTP connection with the user application.
+    sender: HttpSender,
+    /// Address of the user application's HTTP server.
+    local_server_address: SocketAddr,
+    /// Address of this client's TCP socket.
+    self_address: SocketAddr,
```
meowjesty, line 32:
```suggestion
    address: SocketAddr,
```

This one time I think reducing the name here might be better, since it's kinda implied that this is kinda `self.address`, and `local_server_address` exists. Plus the name of the struct makes it hard to think of "what other address could this be".

Nit though.

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
-    match version {
-        Version::HTTP_2 => {
-            let (sender, connection) =
-                http2::handshake(TokioExecutor::default(), TokioIo::new(target_stream)).await?;
-            tokio::spawn(connection);
-
-            Ok(HttpSender::V2(sender))
+impl LocalHttpClient {
+    /// Makes an HTTP connection with the given server and creates a new client.
+    pub async fn new(
```
meowjesty, line 37: No `tracing` love here?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+
+        Ok(Self {
+            sender,
+            local_server_address,
+            self_address,
+        })
+    }
+
+    /// Send the given `request` to the user application's HTTP server.
+    #[tracing::instrument(level = Level::DEBUG, err(level = Level::WARN), ret)]
```
meowjesty, line 60: `DEBUG` wanted or forgot to change it?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    pub async fn send_request(
+        &mut self,
+        request: HttpRequest<StreamingBody>,
+    ) -> Result<Response<Incoming>, LocalHttpError> {
+        self.sender.send_request(request).await
+    }
+
+    /// Returns the address of the local server to which this client is connected.
+    pub fn local_server_address(&self) -> SocketAddr {
```
meowjesty, line 69: No `tracing` love (2)?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+            .field("self_address", &self.self_address)
+            .field("is_http_1", &matches!(self.sender, HttpSender::V1(..)))
+            .finish()
+    }
+}
+
+/// Errors that can occur when sending an HTTP request to the user application.
+#[derive(Error, Debug)]
+pub enum LocalHttpError {
+    #[error("handshake failed: {0}")]
```
meowjesty, line 96:
```suggestion
    #[error("Handshake with local proxy HTTP server failed: `{0}`!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    HandshakeFailed(#[source] hyper::Error),
+
+    #[error("{0:?} is not supported")]
```
meowjesty, line 99:
```suggestion
    #[error("Local proxy HTTP server does not support HTTP version {0:?}!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    UnsupportedHttpVersion(Version),
+
+    #[error("sending the request failed: {0}")]
```
meowjesty, line 102:
```suggestion
    #[error("Local proxy HTTP server failed sending the request: `{0}`!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    SendFailed(#[source] hyper::Error),
+
+    #[error("setting up TCP socket failed: {0}")]
```
meowjesty, line 105:
```suggestion
    #[error("Failed setting up TCP socket for the local proxy HTTP server: `{0}`!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    SocketSetupFailed(#[source] io::Error),
 
-        _http_v1 => {
-            let (sender, connection) = http1::handshake(TokioIo::new(target_stream)).await?;
+    #[error("making a TPC connection failed: {0}")]
```
meowjesty, line 108:
```suggestion
    #[error("Failed making a TCP connection with to the user socket with the local proxy HTTP server: `{0}`!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+    ConnectTcpFailed(#[source] io::Error),
 
-            tokio::spawn(connection.with_upgrades());
+    #[error("reading the response body failed: {0}")]
```
meowjesty, line 111:
```suggestion
    #[error("Local HTTP proxy server failed reading the response body: `{0}`!")]
```

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
+            .field("local_server_address", &self.local_server_address)
+            .field("self_address", &self.self_address)
+            .field("is_http_1", &matches!(self.sender, HttpSender::V1(..)))
+            .finish()
+    }
+}
+
+/// Errors that can occur when sending an HTTP request to the user application.
+#[derive(Error, Debug)]
+pub enum LocalHttpError {
```
meowjesty, line 95: I guess my suggestions are not quite right since we call this `LocalHttpClient`, but hopefully you get the gist of it?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -1,86 +1,256 @@
...
-            Ok(HttpSender::V1(sender))
+impl LocalHttpError {
+    /// Checks if we can retry sending the request, given that the previous attempt resulted in this
+    /// error.
+    pub fn can_retry(&self) -> bool {
+        match self {
+            Self::SocketSetupFailed(..) | Self::UnsupportedHttpVersion(..) => false,
+            Self::ConnectTcpFailed(..) => true,
+            Self::HandshakeFailed(err) | Self::SendFailed(err) | Self::ReadBodyFailed(err) => {
+                !(err.is_parse()
```
meowjesty, line 123: You have chosen the wrong side in the `!` vs `.not()` war.

mirrord/intproxy/src/proxies/incoming/http.rs
meowjesty, line 1: Stuff in this file needs to be `pub`?

mirrord/intproxy/src/proxies/incoming/tcp_proxy.rs
```diff
@@ -0,0 +1,185 @@
...
+    AfterUpgrade(OnUpgrade),
+}
+
+/// [`BackgroundTask`] of [`IncomingProxy`](super::IncomingProxy) that handles a remote
+/// stolen/mirrored TCP connection.
+#[derive(Debug)]
+pub struct TcpProxyTask {
+    /// The local connection between this task and the user application.
+    connection: LocalTcpConnection,
+    /// Whether this task should silently discard data coming from the user application.
```
meowjesty, line 38: Could you also detail "why", or "when" should this data be discarded please.

mirrord/intproxy/src/proxies/incoming/metadata_store.rs
meowjesty, line 1: Same `pub` question as the others.

mirrord/intproxy/src/background_tasks.rs
Razz4780: Oke, now it's `pub(crate)`

mirrord/intproxy/src/background_tasks.rs
Razz4780: Oke, now it's `pub(crate)`

mirrord/intproxy/src/proxies/incoming/metadata_store.rs
Razz4780: This whole module is private in `incoming`, so I think this is fine. If it was unused, clippy would complain

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: It's simpler this way :V
The whole module is private to `incoming`, so clippy will complain about unused stuff

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: Oke, it's `.not()` now

mirrord/intproxy/src/proxies/incoming/tcp_proxy.rs
Razz4780: Done

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: Changed

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: Rephrased messages, wdyt now?

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: tracing here would be just spam, this address is contained in `impl Debug for LocalHttpClient`

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: wdym? :eyes:

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: oke, added

mirrord/intproxy/src/proxies/incoming/http/streaming_body.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+impl Clone for StreamingBody {
+    fn clone(&self) -> Self {
+        Self {
+            shared_state: self.shared_state.clone(),
+            // Setting idx to 0 in order to replay the previous frames.
+            idx: 0,
```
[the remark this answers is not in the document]
Razz4780: Having `Clone` here is very convenient :sweat_smile: 
I extended the doc

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
+                if let Some(gateways) = self.http_gateways.get_mut(&connection_id) {
+                    gateways.remove(&request_id);
+                };
+            }
+        }
     }
 
     /// Handles all agent messages.
-    #[tracing::instrument(level = Level::TRACE, skip(self, message_bus))]
     async fn handle_agent_message(
```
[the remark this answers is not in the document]
Razz4780: All info should already be traced in `IncomingProxy::handle_message`.
If you look into `IncomingProxy::run`, each function called there is instrumented. So we should not be losing any info. Since these functions are basically just `match` expressions, I think more instrumentation would be just spam :eyes:

mirrord/intproxy/src/proxies/incoming/tasks.rs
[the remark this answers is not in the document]
Razz4780: Added

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
-    agent_protocol_version: Option<semver::Version>,
+    /// What HTTP response flavor we produce.
+    response_mode: ResponseMode,
+    /// Cache for [`LocalHttpClient`](http::LocalHttpClient)s.
+    client_store: ClientStore,
+    /// Each mirrored remote connection is mapped to a [TcpProxyTask] in mirror mode.
+    ///
+    /// Each entry here maps to a connection that is in progress both locally and remotely.
+    mirror_tcp_proxies: HashMap<ConnectionId, TaskSender<TcpProxyTask>>,
+    /// Each remote connection stolen in whole is mapped to a [TcpProxyTask] in steal mode.
```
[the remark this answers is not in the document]
Razz4780: Done

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
-    request_body_txs: HashMap<(ConnectionId, RequestId), Sender<InternalHttpBodyFrame>>,
-    /// For managing streamed [`LayerTcpSteal::HttpResponseChunked`] response streams.
-    response_body_rxs: StreamMap<(ConnectionId, RequestId), StreamNotifyClose<ReceiverStreamBody>>,
-    /// Version of [`mirrord_protocol`] negotiated with the agent.
-    agent_protocol_version: Option<semver::Version>,
+    /// What HTTP response flavor we produce.
+    response_mode: ResponseMode,
+    /// Cache for [`LocalHttpClient`](http::LocalHttpClient)s.
+    client_store: ClientStore,
+    /// Each mirrored remote connection is mapped to a [TcpProxyTask] in mirror mode.
```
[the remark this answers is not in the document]
Razz4780: Done

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
+    ///
+    /// Each entry here maps to a connection that is in progress both locally and remotely.
+    mirror_tcp_proxies: HashMap<ConnectionId, TaskSender<TcpProxyTask>>,
+    /// Each remote connection stolen in whole is mapped to a [TcpProxyTask] in steal mode.
+    ///
+    /// Each entry here maps to a connection that is in progress both locally and remotely.
+    steal_tcp_proxies: HashMap<ConnectionId, TaskSender<TcpProxyTask>>,
+    /// Each remote HTTP request stolen with a filter is mapped to a [HttpGatewayTask].
```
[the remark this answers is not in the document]
Razz4780: Done

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
+///    independently). If a request fails locally, we send a
+///    [`StatusCode::BAD_GATEWAY`](hyper::http::StatusCode::BAD_GATEWAY) response.
+///
+/// We are notified about stolen requests with the [`HttpRequest`] messages.
+///
+/// # HTTP upgrades
+///
+/// An HTTP request stolen with a filter can result in an HTTP upgrade.
+/// When this happens, the TPC connection is recovered and passed to a new [`TcpProxyTask`].
+/// The TCP connection is then treated as stolen in whole.
```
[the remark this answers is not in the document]
Razz4780: Change to "without a filter"

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -112,544 +68,647 @@ pub enum IncomingProxyMessage {
...
+                request_id,
+            }) => {
+                tracing::debug!(
+                    connection_id,
+                    request_id,
+                    "Received an error in an HTTP request body",
+                );
+
+                if let Some(gateways) = self.http_gateways.get_mut(&connection_id) {
+                    gateways.remove(&request_id);
```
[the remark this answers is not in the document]
Razz4780: We remove it in more than one place, I extended `IncomingProxy` doc

mirrord/intproxy/src/background_tasks.rs
Razz4780: Breaks me doctest tho https://github.com/Razz4780/mirrord/actions/runs/12882853677/job/35915934768
:<

mirrord/intproxy/src/proxies/incoming/http.rs
meowjesty: I meant if you had forgotten this as `DEBUG` level trace, or if you want this to be `DEBUG`.

meowjesty (approved):
Think there's a place with `TPC` instead of `TCP`, other than that

:+1:

PR #3018, opened 2025-01-14
Add statfs support

facundopoblete:
Add support to forward the `statfs` command to the agent


meowjesty (approved):
lgtm

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -274,6 +274,12 @@ impl FilesProxy {
             {
                 Err(FileResponse::RemoveDir(Err(ResponseError::NotImplemented)))
             }
+            FileRequest::StatFs(..)
+                if protocol_version
+                    .is_some_and(|version: &Version| !STATFS_VERSION.matches(version)) =>
```
Razz4780, line 279: Uhm, I think all `if` conditions in this function are wrong :eyes: 
Should be
```rust
protocol_version
    .map(|v| !SOMETHING_VERSION.matches(v))
    .unwrap_or(true)
```
We want `true` the `protocol_version` is missing

mirrord/protocol/src/codec.rs
```diff
@@ -77,6 +77,7 @@ pub enum FileRequest {
     Access(AccessFileRequest),
     Xstat(XstatRequest),
     XstatFs(XstatFsRequest),
+    StatFs(StatFsRequest),
```
Razz4780, line 80: New enum variants should always be added at the end in `mirrord_protocol` (if the enum is sent on the wire)

Razz4780, line 80: Can you add a dedicated `FileResponse` variant as well? Every file request has one and I'm not sure what would be the implications of not following this pattern

Razz4780 (changes requested):

mirrord/intproxy/src/proxies/files.rs
Razz4780: Or better `protocol_version.is_none_or(|v| !SOMETHING_VERSION.matches(v))`

mirrord/protocol/src/codec.rs
facundopoblete: I just wanted to reuse the `XstatFsResponse` since they are the same. If I am not wrong, it is the same that we talked about for the mkdir command (https://github.com/metalbear-co/mirrord/pull/2949#discussion_r1889832568)

mirrord/intproxy/src/proxies/files.rs
facundopoblete: Thanks, I thought about this case, but I wasn't sure what would be the case where there is no protocol and what should happen. I like the `protocol_version.is_none_or` 👍

mirrord/protocol/src/codec.rs
facundopoblete: I completely forgot about it

Razz4780 (approved):

PR #3020, opened 2025-01-14
Update configuration.md and improve config env.mapping list.

meowjesty:
- Issue: n/a

The list of env var config mapping was being displayed in a single line, so ugly. Changed it to a proper list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,49 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
 
 <!-- towncrier release notes start -->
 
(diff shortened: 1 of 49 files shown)


Razz4780:
It's still a single line in the schema :eyes: 
Maybe you need `*` instead of `-`?

meowjesty:
> It's still a single line in the schema 👀 Maybe you need `*` instead of `-`?

The schema doesn't render as proper markdown? I'll update it with new lines then, but I think it'll still be a single line.

Razz4780 (approved):
Yeah, the schema still shows it in a single line (checked in vscode). Though it's an improvement either way, so approving.

If you want, you can try with `*`s or maybe newlines between the entries

PR #3025, opened 2025-01-21
use older builder base image for aarch64 to support centos-7 libc

aviramha:


Cross.toml
```diff
@@ -5,8 +5,7 @@ passthrough = [
 # Dockerfile used for building mirrord-layer for x64 with very old libc
 # this to support centos7 or Amazon Linux 2.
 [target.x86_64-unknown-linux-gnu]
-image = "ghcr.io/metalbear-co/ci-layer-build:latest"
+image = "ghcr.io/metalbear-co/ci-layer-build:8ca4a4e9757a5749c384c57c3435e56c2b442458e14e4cb7ecbaec4d557f5d69"
```
DmitryDodzin, line 8: maybe we need to change `mirrord/cli/Dockerfile`? I think this is relevant for the cli image (We build cli there with zigbuild but maybe better to do something that downloads github artifacts for the image)

aviramha: I think this is out of scope of this PR anyway?

DmitryDodzin (approved):

PR #3031, opened 2025-01-23
Fixed HTTP filter stuck forever

Razz4780:


mirrord/agent/src/steal/connection.rs
```diff
@@ -719,18 +735,75 @@ mod test {
...
+            TcpConnectionStealer, TcpStealerApi,
+        },
+        watched_task::TaskStatus,
+    };
+
+    /// Notification about a requested redirection operation.
+    ///
+    /// Produced by [`NotifyingRedirector`].
+    #[derive(Debug, PartialEq, Eq)]
+    enum RedirectNotification {
```
DmitryDodzin, line 767: Nice, it's a bit better neater solution for testing rather than setting up the mock with `mockall`

DmitryDodzin (approved):
LGTM, there is an option for using pin_project and just `Box<dyn Future>` to prevent some unpins but it's rather nit-picky and shouldn't really matter.

PR #3033, opened 2025-01-23
Add tolerations for targetless agent.

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/737

### DIFF
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -240,8 +240,10 @@ pub struct AgentConfig {
 
     /// ### agent.tolerations {#agent-tolerations}
     ///
-    /// Set pod tolerations. (not with ephemeral agents)
-    /// Default is
+    /// Set pod tolerations. (not with ephemeral agents).
+    ///
+    /// Defaults to `operator: Exists`.
+    ///
     /// ```json
     /// [
     ///   {
--- a/mirrord/config/src/agent.rs
+++ b/mirrord/config/src/agent.rs
@@ -245,7 +245,7 @@ pub struct AgentConfig {
     /// ```json
     /// [
     ///   {
-    ///     "operator": "Exists"
+    ///     "key": "meow", "operator": "Exists", "effect": "NoSchedule"
     ///   }
     /// ]
     /// ```
--- a/mirrord/kube/src/api/container/pod.rs
+++ b/mirrord/kube/src/api/container/pod.rs
@@ -195,7 +195,7 @@ impl ContainerVariant for PodTargetedVariant<'_> {
         let agent = self.agent_config();
         let params = self.params();
 
-        let tolerations = agent.tolerations.as_ref().unwrap_or(&DEFAULT_TOLERATIONS);
+        let tolerations = agent.tolerations.clone();
 
         let env = self.runtime_data.mesh.map(|mesh_vendor| {
             let mut env = vec![EnvVar {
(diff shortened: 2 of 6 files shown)


aviramha:
Wait - the sentence in the issue might be tricky but we don't want to use default tolerations on the agent when running targetless.

meowjesty:
What should happen then? I thought it was about: if user sets tolerations in operator.agent-config, use those, otherwise ... do?

aviramha:
> What should happen then? I thought it was about: if user sets tolerations in operator.agent-config, use those, otherwise ... do?

Otherwise don't set tolerations.

meowjesty:
Alright!

meowjesty:
Removed the default, now it'll just pass the `Option`.

mirrord/kube/src/api/container/pod.rs
```diff
@@ -195,6 +195,8 @@ impl ContainerVariant for PodTargetedVariant<'_> {
         let agent = self.agent_config();
         let params = self.params();
 
+        let tolerations = agent.tolerations.clone();
```
aviramha, line 198: but that's the targeted variant

aviramha (dismissed):

mirrord/kube/src/api/container/pod.rs
meowjesty: That's the one that changes when the target is set to `targetless`. I have no idea why.

aviramha: ok need @Razz4780 or @DmitryDodzin

mirrord/config/src/agent.rs
```diff
@@ -245,7 +245,7 @@ pub struct AgentConfig {
     /// ```json
     /// [
     ///   {
-    ///     "operator": "Exists"
+    ///     "key": "meow", "operator": "Exists", "effect": "NoSchedule"
```
Razz4780, line 248: Pretty sure that's not the default :laughing:

Razz4780 (changes requested):
Not sure why we even have tolerations :eyes: :

`The default Kubernetes scheduler takes taints and tolerations into account when selecting a node to run a particular Pod. However, if you manually specify the .spec.nodeName for a Pod, that action bypasses the scheduler`

We either:
1. (targetless) Don't care about the node at all
2. (targeted) Require a specific node and we use `.spec.nodeName`

mirrord/kube/src/api/container/pod.rs
Razz4780: > That's the one that changes 

Not sure what this means, but `PodTargetedVariant` (targeted) builds upon `PodVariant` (targetless). See what happens at the end of `PodTargetedVariant::as_update`

aviramha:
I believe we saw agent pods failing to start because they lacked the tolerations

mirrord/config/src/agent.rs
Razz4780: Yeah but this is not an example, this describes the default we'll use. Right? :eyes:

mirrord/kube/src/api/container/pod.rs
meowjesty: Ok, but wait, we don't want tolerations at all in targetless?

mirrord/config/src/agent.rs
meowjesty: Changed the wording around. wdyt?

Razz4780 (approved):
Looks good ^^

mirrord/kube/src/api/container/pod.rs
Razz4780: Ah yes, nice catch

PR #3039, opened 2025-01-27
Add ping pong subtask to extproxy

DmitryDodzin:
Add ping timer to extproxy so at least one connection is kept up while it is running

### DIFF
--- a/mirrord/cli/src/external_proxy.rs
+++ b/mirrord/cli/src/external_proxy.rs
@@ -36,7 +36,7 @@ use local_ip_address::local_ip;
 use mirrord_analytics::{AnalyticsReporter, CollectAnalytics, Reporter};
 use mirrord_config::LayerConfig;
 use mirrord_intproxy::agent_conn::{AgentConnection, ConnectionTlsError};
-use mirrord_protocol::DaemonCodec;
+use mirrord_protocol::{ClientMessage, DaemonCodec, DaemonMessage, LogLevel, LogMessage};
 use tokio::net::{TcpListener, TcpStream};
 use tokio_rustls::server::TlsStream;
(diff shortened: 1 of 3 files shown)


mirrord/cli/src/external_proxy.rs
```diff
@@ -101,6 +101,12 @@ pub async fn proxy(listen_port: u16, watch: drain::Watch) -> CliResult<()> {
         config.external_proxy.start_idle_timeout,
     )));
 
+    let mut own_agent_conn = connect_and_ping(&config, agent_connect_info.clone(), &mut analytics)
+        .await
+        .inspect_err(|_| cancellation_token.cancel())?;
```
t4lz, line 106: At this point the cancellation token still isn't used anywhere, so I don't think this line is needed, is it?

mirrord/cli/src/external_proxy.rs
```diff
@@ -101,6 +101,12 @@ pub async fn proxy(listen_port: u16, watch: drain::Watch) -> CliResult<()> {
         config.external_proxy.start_idle_timeout,
     )));
 
+    let mut own_agent_conn = connect_and_ping(&config, agent_connect_info.clone(), &mut analytics)
```
t4lz, line 104:
```suggestion
    // This connection is just to keep the agent alive as long as the client side is running.
    let mut own_agent_conn = connect_and_ping(&config, agent_connect_info.clone(), &mut analytics)
```
Or explain in your own words :)

t4lz (approved):

mirrord/cli/src/external_proxy.rs
DmitryDodzin: Yep, not needed here

PR #3048, opened 2025-01-29
mirrord policy for enforcing header pattern

DmitryDodzin:
Update to `MirrordPolicy` and `MirrordClusterPolicy` to allow enforcing header pattern for steal-with-filter


mirrord/operator/src/crd/policy.rs
```diff
@@ -150,6 +157,33 @@ pub struct FsPolicy {
...
 
+/// Network operations policy that partialy mimics the mirrord network config.
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct NetworkPolicy {
+    #[serde(default)]
+    pub incoming: IncomingNetworkPolicy,
+}
+
+/// Incoming network operations policy that partialy mimics the mirrord network.incoming config.
```
Razz4780, line 168:
```suggestion
/// Incoming network operations policy that partialy mimics the mirrord `network.incoming` config.
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -150,6 +157,33 @@ pub struct FsPolicy {
...
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct IncomingNetworkPolicy {
+    #[serde(default)]
+    pub http_filter: HttpFilterPolicy,
+}
+
+/// Http filter policy that allows to specify any filter requirments that users must specify in
+/// their config for a successful network steal
```
Razz4780, line 177:
```suggestion
/// Http filter policy that allows to specify requirements for the HTTP filter used in a session.
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -150,6 +157,33 @@ pub struct FsPolicy {
...
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct HttpFilterPolicy {
+    /// Require user's filter to match this regex if filter is provided. (this works in tandom with
+    /// `steal-without-filter` block to require the user to specify a header filter for network
+    /// steal feature)
```
Razz4780, line 183:
```suggestion
    /// Require the user's header filter to match this regex if such filter is provided.
    /// 
    /// This works in tandem with the `steal-without-filter` block
    /// to require that the user to specifies a header filter for the network steal feature.
```

mirrord/protocol/src/error.rs
```diff
@@ -59,10 +59,11 @@ pub enum ResponseError {
     #[error("Operation is not yet supported by mirrord.")]
     NotImplemented,
 
-    #[error("{blocked_action} is forbidden by {} for this target (your organization does not allow you to use this mirrord feature with the chosen target).", policy_name_string(.policy_name.clone()))]
+    #[error("{blocked_action} is forbidden by {} for this target ({}).", policy_name_string(.policy_name.as_deref()), policy_reason(.reason.as_deref()))]
     Forbidden {
         blocked_action: BlockedAction,
         policy_name: Option<String>,
+        reason: Option<String>,
```
Razz4780, line 66: This breaks compatibility, we can't add new fields.

mirrord/operator/src/crd/policy.rs
```diff
@@ -150,6 +157,33 @@ pub struct FsPolicy {
...
+    pub header_filter: Option<String>,
```
Razz4780, line 184: What if multiple filters are used? Right now the user can specify any_of/all_of filters.

I think the semantics should be that an HTTP filter is accepted only when it has a header filter matching this pattern, and this header filter is **required** to match for the request to be stolen (so `any_of(some-path-filter, valid-header-filter)` would not pass the policy check). If so, this should probably be named like `requires_header_filter` or `required_header_filter`, and the doc should mention the case of composed filters.

Razz4780 (changes requested):

mirrord/protocol/src/error.rs
DmitryDodzin: It should be fine because it's at the end of all messages that contain this result and it looks working for old layer > new operator and new layer > old operator

DmitryDodzin: Like we don't have any message where `RemoteResult` isn't the last field (well all the places it's the only field)

mirrord/operator/src/crd/policy.rs
DmitryDodzin: There is also the possibility of multiple rules being applied on single target so making a correct regex can be tricky

mirrord/protocol/src/error.rs
Razz4780: This test mocks new intproxy and old operator, decoding the first message fails :<

```rust
#[test]
fn compat_test() {
    #[derive(Encode, Decode, PartialEq, Eq, Debug)]
    enum ResponseErrorOld {
        IdsExhausted(String),
        NotFound(u64),
        NotDirectory(u64),
        NotFile(u64),
        RemoteIO(RemoteIOError),
        DnsLookup(DnsLookupError),
        Remote(RemoteError),
        PortAlreadyStolen(Port),
        NotImplemented,
        Forbidden {
            blocked_action: BlockedAction,
            policy_name: Option<String>,
        },
        StripPrefix(String),
        OpenLocal,
    }

    #[derive(Encode, Decode, PartialEq, Eq, Debug)]
    enum ResponseErrorNew {
        IdsExhausted(String),
        NotFound(u64),
        NotDirectory(u64),
        NotFile(u64),
        RemoteIO(RemoteIOError),
        DnsLookup(DnsLookupError),
        Remote(RemoteError),
        PortAlreadyStolen(Port),
        NotImplemented,
        Forbidden {
            blocked_action: BlockedAction,
            policy_name: Option<String>,
            reason: Option<String>,
        },
        StripPrefix(String),
        OpenLocal,
    }

    let mut codec_old_sender = ProtocolCodec::<(), ResponseErrorOld>::default();
    let mut codec_new_receiver = ProtocolCodec::<ResponseErrorNew, ()>::default();
    let mut buffer = BytesMut::default();

    codec_old_sender
        .encode(
            ResponseErrorOld::Forbidden {
                blocked_action: BlockedAction::Mirror(80),
                policy_name: Some("name_1".into()),
            },
            &mut buffer,
        )
        .unwrap();
    codec_old_sender
        .encode(
            ResponseErrorOld::Forbidden {
                blocked_action: BlockedAction::Steal(StealType::All(99)),
                policy_name: Some("name_2".into()),
            },
            &mut buffer,
        )
        .unwrap();

    let error_1 = codec_new_receiver.decode(&mut buffer).unwrap();
    assert_eq!(
        error_1,
        Some(ResponseErrorNew::Forbidden {
            blocked_action: BlockedAction::Mirror(80),
            policy_name: Some("name_1".into()),
            reason: None
        })
    );

    let error_2 = codec_new_receiver.decode(&mut buffer).unwrap();
    assert_eq!(
        error_2,
        Some(ResponseErrorNew::Forbidden {
            blocked_action: BlockedAction::Steal(StealType::All(99)),
            policy_name: Some("name_2".into()),
            reason: None
        })
    );
}
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -63,6 +63,10 @@ pub struct MirrordPolicySpec {
     /// the user config.
     #[serde(default)]
     pub fs: FsPolicy,
+
+    /// Fine grained control over network features like specifiying required http filters.
+    #[serde(default)]
+    pub network: NetworkPolicy,
```
t4lz, line 69: The network policy is now kind of distributed between `BlockedFeature`s and this.
I guess `env` and `fs` should have also been `BlockedFeature`s, but it's too late for that. But with network I find it a bit more confusing because there is already a `StealWithoutFilter` blocked feature.
I would expect the header requirement to be in a new, similar variant, that also includes a header filter specification.

DmitryDodzin: I feel like it's going to be more convoluted once a path filter is introduced and this is is already quite convoluted, here I like that it's simmilar to config

mirrord/operator/src/crd/policy.rs
```diff
@@ -63,6 +63,10 @@ pub struct MirrordPolicySpec {
     /// the user config.
     #[serde(default)]
     pub fs: FsPolicy,
+
+    /// Fine grained control over network features like specifiying required http filters.
```
Razz4780, line 67:
```suggestion
    /// Fine grained control over network features like specifying required HTTP filters.
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -150,6 +157,33 @@ pub struct FsPolicy {
...
+}
+
+/// Http filter policy that allows to specify requirements for the HTTP filter used in a session.
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct HttpFilterPolicy {
+    /// Require the user's header filter to match this regex if such filter is provided.
+    ///
+    /// This works in tandem with the `steal-without-filter` block
+    /// to require that the user to specifies a header filter for the network steal feature.
```
Razz4780, line 183:
```suggestion
    /// Require the user's header filter to match this regex, if such filter is provided.
    ///
    /// This works in tandem with the `steal-without-filter` block
    /// to require that the user specifies a header filter for the network steal feature.
    ///
    /// # Composed filters
    ///
    /// When the user requests an `all_of` HTTP filter, at least one of the nested filters
    /// must be a header filter that matches this regex. At least one nested filter is required.
    /// 
    /// When the user requests an `any_of` HTTP filter, all nested filters
    /// must be header filters and match this regex. At least one nested filter is required.
```

Razz4780 (changes requested):

Razz4780 (approved):

PR #3050, opened 2025-01-31
Fix mirrord ls test

Razz4780:


tests/src/operator/sanity.rs
```diff
@@ -16,15 +16,33 @@ pub async fn mirrord_ls(#[future] service_for_mirrord_ls: KubeService) {
...
+        "cronjob",
+        "job",
+        "replicaset",
+        "service",
+    ];
+
+    let re = Regex::new(&format!(
+        r"^({})/.+(/container/.+)?$",
+        expected_target_types
+            .into_iter()
```
gememma, line 33: why turn it into an iterator? can you not do `let expected_target_types = vec![blah, blah, blah]` and then simplify to `expected_target_types.join("|")` in regex?

Razz4780: ah yes, right

Razz4780: fixed

gememma (approved):
🐿️

Issue #3044, opened 2025-01-28
error in 1.131.x crashes go process


aviramha:
User sees

```

2025-01-28T19:12:11.654285Z ERROR ThreadId(02) mirrord_layer::error: Error occured in Layer >> ResponseError(NotImplemented)
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd30786]
```

go 1.23.2 - running on arm64 macOS inside Docker

aviramha:
I'm not sure it solves users' issue yet

aviramha:
My fix didn't fix 😒

aviramha:
I believe the issue is that go calls the syscall `statfs` which uses the `statfs64` struct in libc, and we use the `statfs` struct which is backward compatible, hence when coming from go we should actually call `statfs64_detour` instead.
this is probably relevant to other syscalls that were added as part of the PR (I think there are 2 structs/functions?)

t4lz:
Is there a reproduction you guys worked with?

aviramha:
I couldn't reproduce, but I am pretty sure that if you write a go app that calls statfs when using mirrord it'd get wrong data in the last fields, and that's how we can add some verification to it.

aviramha:
(On Linux)

t4lz:
We (implicitly) use the statfs64 struct in the agent when we do the call (we call `nix::sys::statfs::statfs` and [on linux that just calls statfs64](https://docs.rs/nix/latest/aarch64-linux-android/src/nix/sys/statfs.rs.html#33)), and when we fill the struct in the layer the memory is the same as in the 64 bit struct. However we just don't send over some fields, and instead set them to 0 in the user application's struct.
I'm adding a new protocol message so that we can send over all of the fields.
I don't know if that will prevent the crash, as we don't know for sure the missing fields were the cause for the crash (note the memory of the missing fields was not uninitialized, it was zeroed).

aviramha:
Still happens after rollback of Go

aviramha:
It also happens on macOS natively :|

aviramha:
could be related
https://github.com/metalbear-co/mirrord/issues/3082

PR #3051, opened 2025-01-31
statfs in Go on Linux

t4lz:
Might prevent #3044 

We were already calling statfs64 in the agent.
Now sending back all of the statfs64 data back to the client with a new message type.
Added new statfs64 detours, and reactivated the Go detour, using the statfs64 detours.

Added an E2E test that runs Go's statfs in a deployed test image and with mirrord, and compares the result (we do not require the exact same data, because some fields change from call to call).

### DIFF
--- a/mirrord/intproxy/src/proxies/files.rs
+++ b/mirrord/intproxy/src/proxies/files.rs
@@ -6,7 +6,7 @@ use mirrord_protocol::{
     file::{
         CloseDirRequest, CloseFileRequest, DirEntryInternal, ReadDirBatchRequest, ReadDirResponse,
         ReadFileResponse, ReadLimitedFileRequest, SeekFromInternal, MKDIR_VERSION,
-        READDIR_BATCH_VERSION, READLINK_VERSION, RMDIR_VERSION, STATFS_VERSION,
+        READDIR_BATCH_VERSION, READLINK_VERSION, RMDIR_VERSION, STATFS_V2_VERSION, STATFS_VERSION,
     },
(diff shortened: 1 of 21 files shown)


mirrord/intproxy/src/proxies/files.rs
```diff
@@ -539,6 +539,32 @@ impl FilesProxy {
...
+                    .as_ref()
+                    .is_none_or(|version| !STATFS_V2_VERSION.matches(version)) =>
+            {
+                self.request_queue.push_back(message_id, layer_id);
+                message_bus
+                    .send(ClientMessage::FileRequest(FileRequest::XstatFs(
+                        xstatfs_v2.into(),
+                    )))
+                    .await;
+            }
```
Razz4780, line 567: Shouldn't we rather convert to the old format? This will break some user apps until they update agent/operator

mirrord/protocol/src/codec.rs
```diff
@@ -94,6 +94,12 @@ pub enum FileRequest {
     Unlink(UnlinkRequest),
     UnlinkAt(UnlinkAtRequest),
     StatFs(StatFsRequest),
+
+    /// Same as XstatFs, but results in the V2 response.
+    XstatFsV2(XstatFsRequestV2),
+
+    /// Same as StatFs, but results in the V2 response.
+    StatFsV2(StatFsRequestV2),
```
Razz4780, line 102: Do we need extra request variants? The agent could simply check client protocol version and decide which response variant to send

PR #3054, opened 2025-02-03
Remove faulty statfs hooks in Go apps

Razz4780:


aviramha:
We can just remove the go hooks part since that's the broken part.

Razz4780:
> We can just remove the go hooks part since that's the broken part.

Done :+1:

meowjesty (approved):
:bug:

Issue #3072, opened 2025-02-06
High latency in local HTTP


Razz4780:
### Bug Description

Requests stolen with filter experience high latency due to some delay in the local HTTP (intproxy<->user app)

### Steps to Reproduce

---

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### mirrord CLI version

>=3.131.0

### mirrord-operator version (if relevant)

operator present

### Your operating system and version

---

### Local process

---

PR #3084, opened 2025-02-12
Made idle timeout configurable for local HTTP clients

Razz4780:
Issue #3072

The situation improved when we started reusing local connections, which eliminated long-standing idle connections. My suspicion is that the user application does not handle multiple open HTTP connections very well. With this config, they'll be able to have the connections killed right after the request

### DIFF
--- a/mirrord/config/src/experimental.rs
+++ b/mirrord/config/src/experimental.rs
@@ -70,6 +70,22 @@ pub struct ExperimentalConfig {
     /// <https://github.com/metalbear-co/mirrord/issues/2069>
     #[config(default = 128000)]
     pub readonly_file_buffer: u64,
+
+    /// ### _experimental_ idle_local_http_connection_timeout {#experimental-idle_local_http_connection_timeout}
+    ///
+    /// Sets a timeout for idle local HTTP connections (in milliseconds).
+    ///
+    /// HTTP requests stolen with a filter are delivered to the local application
+    /// from a HTTP connection made from the local machine. Once a request is delivered,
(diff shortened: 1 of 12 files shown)


mirrord/config/src/experimental.rs
```diff
@@ -70,6 +70,22 @@ pub struct ExperimentalConfig {
...
+    /// HTTP requests stolen with a filter are delivered to the local application
+    /// from a HTTP connection made from the local machine. Once a request is delivered,
+    /// the connection is cached for some time, so that it can be reused to deliver
+    /// the next request.
+    ///
+    /// This timeout determines for how long such connections are cached.
+    ///
+    /// Set to 0 to disable caching local HTTP connections (connections will be dropped as soon as
+    /// the request is delivered).
+    #[config(default = 3000)]
```
DmitryDodzin, line 87: I would add the default value to doc

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -126,7 +126,6 @@ struct HttpGatewayHandle {
 /// An HTTP request stolen with a filter can result in an HTTP upgrade.
 /// When this happens, the TCP connection is recovered and passed to a new [`TcpProxyTask`].
 /// The TCP connection is then treated as stolen without a filter.
-#[derive(Default)]
```
DmitryDodzin, line 129: We can also remove `Default` from `ClientStore` to just to prevent any confusion with `IDLE_CLIENT_DEFAULT_TIMEOUT`

DmitryDodzin (approved):
LGTM a couple of nitpicks

PR #3088, opened 2025-02-13
Fix wrong bypassing checks in FS detours

### LINKED ISSUE #3083: Wrong `ensure_not_ignored!` calls
Some of the calls to `ensure_not_ignored!` pass `false` for write accesses, which I think is wrong.
Let's go over all the calls to `ensure_not_ignored!` and make sure they're all correct.

Some examples:

https://github.com/metalbear-co/mirrord/blob/971c6caa0d2e6b28433b08ddddaf2aaadec365b7/mirrord/layer/src/file/ops.rs#L350-L358

https://github.com/metalbear-co/mirrord/blob/971c6caa0d2e6b28433b08ddddaf2aaadec365b7/mirrord/layer/src/file/ops.rs#L400-L408

t4lz:
Fix #3083


DmitryDodzin (approved):
LGTM

Issue #3076, opened 2025-02-10
ephemeral targets wrong file system / pid


aviramha:
### Bug Description

while debugging an issue with a customer, we noticed that `/proc/1/cmdline` was `/pause` binary, and the actual PID of the desired container was quite high (43~).
We probably should find a way to get the pid of the desired container and use that instead of using the static `1` pid

### Steps to Reproduce

I don't have clear instructions, but they had crowdstrike init container + sidecar, perhaps the culprit is init containers + sidecars?

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### Your operating system and version

n/A

### Local process

N/A

meowjesty:
- `reproduce.yaml` so you can just `kubectl apply -f reproduce.yaml`;

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: py-serv-deployment
  labels:
    app: py-serv
spec:
  replicas: 1
  selector:
    matchLabels:
      app: py-serv
  template:
    metadata:
      labels:
        app: py-serv
    spec:
      # Without this, the issue doesn't reproduce.
      shareProcessNamespace: true
      initContainers:
      - name: debian
        image: debian
        command: ['sh', '-c', 'echo "Hello, World!"']
      containers:
        - name: nginx1
          image: nginx
        - name: py-serv
          image: ghcr.io/metalbear-co/mirrord-pytest:latest
          ports:
            - containerPort: 9999
          env:
            - name: MIRRORD_FAKE_VAR_FIRST
              value: mirrord.is.running
            - name: MIRRORD_FAKE_VAR_SECOND
              value: "7777"
            - name: PORT
              value: "9999"
        - name: nginx2
          image: debian
          command: ['sleep', '300000']

---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: py-serv
  name: py-serv
spec:
  ports:
    - port: 80
      protocol: TCP
      targetPort: 80
      nodePort: 30000
  selector:
    app: py-serv
  sessionAffinity: None
  type: NodePort
```

PR #3091, opened 2025-02-13
Look for the correct pid that matches the targets container_id  (by searching /proc/pid/cgroup).

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/3076


mirrord/agent/src/cli.rs
```diff
@@ -95,6 +95,10 @@ pub enum Mode {
         // This argument is being kept here only for compatibility with very old CLIs.
         #[arg(long)]
         mesh: Option<String>,
+
+        /// Container id to get traffic from
+        #[arg(short, long)]
+        container_id: Option<String>,
```
Razz4780, line 101: Don't use command line arguments, you'll break old agents :V

Razz4780: Use the fancy `mirrord-agent-env` crate

meowjesty: How does this break old agents though? If this is not present, they would keep working, no? And if this is present, then they have this subtle bug where (mainly fs) operations don't work as expected.

I'll change it to env, just curious.

Razz4780: Yeah, but at least they start :shrug: 
We're moving from cli arguments in the agent, that's for sure

mirrord/agent/src/runtime.rs
```diff
@@ -269,13 +270,90 @@ impl ContainerRuntime for ContainerdContainer {
...
+                }
+                Ok(None) => return Ok(1),
+                Err(fail) => {
+                    tracing::warn!(?fail, "Searching for container pid!");
+                    continue;
+                }
+            }
+        }
+    } else {
+        Ok(1)
```
aviramha, line 342: We should warn on fallback. (I guess this applies above too)

mirrord/agent/env/src/envs.rs
```diff
@@ -51,3 +51,10 @@ pub const POD_IPS: CheckedEnv<Vec<IpAddr>> = CheckedEnv::new("MIRRORD_AGENT_POD_
 ///
 /// Should follow `tracing` format, e.g `mirrord=trace`.
 pub const LOG_LEVEL: CheckedEnv<String> = CheckedEnv::new("RUST_LOG");
+
+/// Container id to get traffic from.
```
aviramha, line 55: Container id to "attach"/"bind"/"work with"
It's not only for traffic.

mirrord/agent/env/src/envs.rs
```diff
@@ -51,3 +51,10 @@ pub const POD_IPS: CheckedEnv<Vec<IpAddr>> = CheckedEnv::new("MIRRORD_AGENT_POD_
...
+///
+/// For `Mode::Ephemeral`, this is also used to get the target's `pid`, when this is not
```
aviramha, line 57: This makes it sound like it has any other purpose other than ephemeral, which could be true but the env is also called ephemeral so need to have clearer docs.

aviramha:
Good job on the docs - need some improvement though :)

mirrord/agent/env/src/envs.rs
meowjesty: Good one, the naming was very confusing. Changed it and made the docs more explicit.

mirrord/agent/env/src/envs.rs
meowjesty: Changed.

mirrord/agent/src/runtime.rs
meowjesty: Warnings everywhere now!

mirrord/agent/env/src/envs.rs
```diff
@@ -51,3 +51,15 @@ pub const POD_IPS: CheckedEnv<Vec<IpAddr>> = CheckedEnv::new("MIRRORD_AGENT_POD_
 ///
 /// Should follow `tracing` format, e.g `mirrord=trace`.
 pub const LOG_LEVEL: CheckedEnv<String> = CheckedEnv::new("RUST_LOG");
+
+/// Container id of the target we're attaching to, e.g. `mirrord exec -t
+/// pod/glorious-cat/container/[cat-container]`, this is the id of `cat-container` that you
+/// can retrieve with `kubectl describe glorious-cat`.
+///
+/// For `Mode::Ephemeral`, this is also used to get the target's `pid`. When this is not
```
aviramha, line 59: is it used for other thing other than `Mode::Ephemeral` ? if not, "this is used"

meowjesty: Changed.

meowjesty: The "For Mode::Ephemeral" still might make it seem like it's used in other places, but I'm failing to come up with a phrase that says it's only used in that mode.

mirrord/agent/src/runtime.rs
```diff
@@ -269,13 +270,117 @@ impl ContainerRuntime for ContainerdContainer {
...
+        let mut dir_entries = read_dir("/proc").await?;
+
+        loop {
+            match dir_entries.next_entry().await {
+                Ok(Some(dir)) => {
+                    match dir
+                        .path()
+                        .to_string_lossy()
+                        .split_terminator("/")
+                        .last()
```
DmitryDodzin, line 315: This looks a bit strange, why not `dir.file_name()`?

DmitryDodzin (changes requested):

DmitryDodzin (approved):
LGTM

PR #3097, opened 2025-02-17
Prevent reading a remote directory from crashing

gememma:


Razz4780 (changes requested):
Looks good, could add a unit test?

Razz4780 (approved):
:fire:

PR #3101, opened 2025-02-18
Fix mirrord CLI logs

Razz4780:


mirrord/cli/src/logging.rs
```diff
@@ -17,44 +17,43 @@ use crate::{
...
+                .unwrap_or(false)
+        }
+
+        _ => true,
+    };
 
-    if force_log || init_ext_error_handler(command) {
+    if do_init {
         tracing_subscriber::registry()
             .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
```
meowjesty, line 58: Take this opportunity to add `.pretty().with_file_whatever(true).with_line_number(true)`.

meowjesty (approved):
:spiral_notepad:

PR #3102, opened 2025-02-18
Fixed unknownhostexception with basic java application

Razz4780:


t4lz (approved):
If the operator is fine with it, I'm fine with it

PR #3104, opened 2025-02-18
Remove unnecessary deps from mirrord-layer

Razz4780:


mirrord/layer/src/socket/ops.rs
```diff
@@ -338,7 +341,7 @@ pub(super) fn bind(
 
     // node reads errno to check if bind was successful and doesn't care about the return value
     // (???)
```
meowjesty, line 343: ???

meowjesty (approved):
:+1:

mirrord/layer/src/socket/ops.rs
Razz4780: We can bet on who wrote this :cat2:

PR #3107, opened 2025-02-19
Fix different fs mapping / path check behaviour in different hooks

Razz4780:


mirrord/layer/src/file/ops.rs
```diff
@@ -32,37 +43,35 @@ use crate::{
...
-macro_rules! check_relative_paths {
-    ($path:expr) => {
-        if $path.is_relative() {
-            Detour::Bypass(Bypass::relative_path($path.to_str().unwrap_or_default()))?
-        };
-    };
+impl PathExt for Path {
+    fn ensure_not_relative(&self) -> Detour<()> {
+        if self.is_relative() {
+            Detour::Bypass(Bypass::relative_path(self.to_str().unwrap_or_default()))
```
meowjesty, line 55: Wonder if for these `Path` errors we should use the `Path::display` to reduce the noise of `to_str.unwrap_or_default`.

In the error hold a `path::Display`. I'm _meh_ on this suggestion, but wdyt?

mirrord/layer/src/file/filter.rs
```diff
@@ -125,30 +125,34 @@ impl FileFilter {
...
                 if write {
-                    Detour::Bypass(op())
+                    Detour::Bypass(Bypass::IgnoredFile(CString::new(text).unwrap_or_default()))
                 } else {
                     Detour::Success(())
                 }
             }
-            _ if self.local.is_match(text) => Detour::Bypass(op()),
+            _ if self.local.is_match(text) => {
+                Detour::Bypass(Bypass::IgnoredFile(CString::new(text).unwrap_or_default()))
```
meowjesty, line 146: Why not use the `ignored_file` fn, and if you don't want to keep the `expect` there, wouldn't it be safe to change it to `unwrap_or_default`?

Maybe add a `debug_assert!` in helper function there?

meowjesty (approved):
`common_pr_check`

Only have 2 suggestions that are up to you.

mirrord/layer/src/file/ops.rs
Razz4780: The error needs to hold a `CString` for `update_ptr_from_bypass` ;/

mirrord/layer/src/file/filter.rs
Razz4780: The `expect`s have been working fine for a while now, so I'd keep them for now. I'm thinking about doing a small cleanup in another PR

PR #3111, opened 2025-02-21
Add arm64 code shim logic to x86 builds

gememma:


.github/workflows/release.yaml
```diff
@@ -100,6 +100,7 @@ jobs:
       - name: build macOS x86-64 cli with universal dylib
         env:
           MIRRORD_LAYER_FILE: /tmp/target/universal-apple-darwin/release/libmirrord_layer.dylib
+          MIRRORD_LAYER_FILE_MACOS_ARM64: ../../../target/aarch64-apple-darwin/release/libmirrord_layer.dylib
```
DmitryDodzin, line 103:
```suggestion
          MIRRORD_LAYER_FILE_MACOS_ARM64: /tmp/target/aarch64-apple-darwin/release/libmirrord_layer.dylib
```

Maybe it's better it being absolute like line above if it works

DmitryDodzin (approved):
LGTM

.github/workflows/release.yaml
gememma: I checked and this doesn't work - the build is very sensitive to changing this file path for some reason :/

PR #3116, opened 2025-02-24
Some mirrord-kube improvements

Razz4780:
* Improved logging in `wrap_raw_connection`
* Namespaces of the agent pod and the target pod are now non-optional (they were always `Some` anyway)
* Removed unnecessary "agent_version" field - has not been used for a long time
* Incluster version of `KubernetesAPI::create_connection` now takes connect info via reference


gememma (approved):
lgtm 🐭

PR #3129, opened 2025-02-28
Improve intproxy tracing

Razz4780:
1. Currently intproxy logs make sense only on `TRACE` level
2. It's very painful to read them

This PR:
1. Changes some log levels, so that higher levels than `TRACE` produce meaningful output
2. Adds some spans for more context
3. Changes log output format to JSON, so that we can use json log explorers, like [hl](https://github.com/pamburus/hl)

Attaching an example log file (`mirrord=trace`). I find it way easier to read with `hl`, e.g I can do `hl -l warn intproxy.log` to see all failed HTTP requests or `hl -q '(span.request contain "connection_id: 0" or span.self contain "connection_id: 0")' intproxy.log` to track a specific request

[intproxy.log](https://github.com/user-attachments/files/19027910/intproxy.log)


mirrord/cli/src/internal_proxy.rs
```diff
@@ -111,14 +117,11 @@ pub(crate) async fn proxy(
...
-    .map_err(InternalProxyError::from)
-    .inspect_err(|error| {
-        tracing::error!(%error, "Internal proxy encountered an error, exiting");
-    })
+    .map_err(From::from)
 }
 
 /// Creates a connection with the agent and handles one round of ping pong.
-#[tracing::instrument(level = Level::TRACE)]
+#[tracing::instrument(level = Level::TRACE, skip(config, analytics))]
```
gememma, line 124: `skip(config)` 🙌 🙇 ☀️ 🙏

mirrord/cli/src/logging.rs
```diff
@@ -77,9 +73,19 @@ fn default_logfile_path(prefix: &str) -> PathBuf {
         .expect("now must have some delta from UNIX_EPOCH, it isn't 1970 anymore")
         .as_secs();
 
-    PathBuf::from(format!("/tmp/{prefix}-{timestamp}-{random_name}.log"))
+    let mut path = std::env::temp_dir();
+    path.push(format!("/tmp/{prefix}-{timestamp}-{random_name}.log"));
```
gememma, line 77: does this get logged/ printed out anywhere? I know it has a timestamp but might be nice to put it in the progress printout or something?

mirrord/intproxy/src/lib.rs
```diff
@@ -165,29 +164,37 @@ impl IntProxy {
...
-                    }
+                    return Err(IntProxyError::ConnectionAcceptTimeout);
                 },
 
                 _ = time::sleep(idle_timeout), if self.any_connection_accepted && self.task_txs.layers.is_empty() => {
-                    if self.task_txs.layers.is_empty() {
-                        tracing::trace!("intproxy timeout, no active connections. Exiting.");
-                        break;
-                    }
+                    tracing::info!("Reached the timeout on no active layer connections");
```
gememma, line 181:
```suggestion
                    tracing::info!("Reached the idle timeout with no active layer connections");
```

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -1092,21 +1089,29 @@ impl FilesProxy {
         Ok(())
     }
 
+    #[tracing::instrument(level = Level::INFO, skip(message_bus), ret)]
     async fn handle_reconnect(&mut self, message_bus: &mut MessageBus<Self>) {
         for (_, fds) in self.remote_files.drain() {
             for fd in fds {
+                tracing::debug!(fd, "Dropping remote file");
```
gememma, line 1096: it might be less spammy to list the files in one log (unless the drop can fail) before the `for` loop? like:
```rust
tracing::debug!(fds, "Dropping remote files");
for fd in fds {
    self.blah blah blah ...
}
``` 
you could even condense the `for` loop into `fds.for_each(...)`

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -1092,21 +1089,29 @@ impl FilesProxy {
...
                 self.buffered_files.remove(&fd);
             }
         }
 
         for (_, fds) in self.remote_dirs.drain() {
             for fd in fds {
+                tracing::debug!(fd, "Dropping remote directory");
```
gememma, line 1103: same as above 👀

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -1092,21 +1089,29 @@ impl FilesProxy {
...
                 self.buffered_dirs.remove(&fd);
             }
         }
 
+        let responses = self.reconnect_tracker.agent_lost();
+        tracing::debug!(
+            responses = responses.len(),
```
gememma, line 1110:
```suggestion
            num_responses = responses.len(),
```

mirrord/intproxy/src/proxies/files.rs
```diff
@@ -1116,10 +1121,9 @@ impl BackgroundTask for FilesProxy {
     type MessageOut = ProxyMessage;
     type Error = FilesProxyError;
 
+    #[tracing::instrument(level = Level::INFO, name = "files_proxy_main_loop", skip_all, ret, err)]
     async fn run(&mut self, message_bus: &mut MessageBus<Self>) -> Result<(), Self::Error> {
         while let Some(message) = message_bus.recv().await {
-            tracing::trace!(?message, "new message in message_bus");
```
gememma, line 1121: why no trace on message?

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -633,7 +638,7 @@ impl IncomingProxy {
                 self.tasks.clear();
 
                 for subscription in self.subscriptions.iter_mut() {
-                    tracing::debug!(?subscription, "resubscribing");
+                    tracing::info!(?subscription, "Resubscribing after connection refresh");
```
gememma, line 641: is this a subscription_id or like a port number or something?

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -38,7 +44,7 @@ pub struct LocalHttpClient {
 
 impl LocalHttpClient {
     /// Send the given `request` to the user application's HTTP server.
-    #[tracing::instrument(level = Level::DEBUG, err(level = Level::WARN), ret)]
+    #[tracing::instrument(level = Level::TRACE, err(level = Level::TRACE), ret)]
```
gememma, line 47: `err(level = Level::TRACE)` and not DEBUG? por que?

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -99,10 +99,15 @@ impl SimpleProxy {
             .is_some_and(|version| ADDRINFO_V2_VERSION.matches(version))
     }
 
+    #[tracing::instrument(level = Level::INFO, skip_all, ret, err)]
     async fn handle_connection_refresh(
         &mut self,
         message_bus: &mut MessageBus<Self>,
     ) -> Result<(), SimpleProxyError> {
+        tracing::debug!(
+            responses = self.addr_info_reqs.len(),
```
gememma, line 108:
```suggestion
            num_responses = self.addr_info_reqs.len(),
```

mirrord/intproxy/src/proxies/simple.rs
```diff
@@ -111,6 +116,10 @@ impl SimpleProxy {
                 .await;
         }
 
+        tracing::debug!(
+            responses = self.get_env_reqs.len(),
```
gememma, line 120:
```suggestion
            num_responses = self.get_env_reqs.len(),
```

gememma (approved):
so nice to clean out the logs 🧹 🌸 I don't feel strongly about any of my comments, but I make them regardless 🫡

mirrord/intproxy/src/lib.rs
```diff
@@ -372,9 +383,18 @@ impl IntProxy {
...
+                    "Received a log message from the agent"
+                ),
+                LogLevel::Warn => tracing::warn!(
+                    message = log.message,
+                    "Received a log message from the agent"
+                ),
+                LogLevel::Info => tracing::info!(
+                    message = log.message,
+                    "Received a log message from the agent"
+                ),
```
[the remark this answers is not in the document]
Razz4780: Nope, tried this before. Level must be const :<

gememma:
It might also be nice to add something to the CONTRIBUTING.md doc, maybe
1. how to navigate logs easily (as you already described in the PR description using `hl`)
2. the kind of info (roughly) that you would expect to see at each log level 
3. any other best practices like "always log the connection ID because otherwise i will be sad" or "never log the connection ID because otherwise i will be sad"

mirrord/cli/src/logging.rs
Razz4780: I does not, because intproxy/extproxy stderr/stdout is yeeted (it runs in the background). But I think I can pass it to the CLI / generate it in the CLI

mirrord/intproxy/src/proxies/incoming/http.rs
Razz4780: Too spammy, I think. Attempts/errors are already logged on `DEBUG` in the `HttpGatewayTask`

mirrord/intproxy/src/proxies/incoming.rs
Razz4780: It's a more complex struct. Contains all active subscription requests made by all layers for a single remote port :V

mirrord/intproxy/src/proxies/files.rs
Razz4780: I think it's just spam. All messages going through intproxy are traced in the main `IntProxy` struct. Right now I'm seeing the same message like 3-4 times in the logs :sweat_smile:

mirrord/cli/src/logging.rs
Razz4780: Done ^^

gememma (commented):
Just a docs suggestion 👀

Issue #2907, opened 2024-11-07
support docker compose


aviramha:
similar to `mirrord container`, we'd like to have `mirrord compose up` that would inject mirrord to one of the containers (atleast initially, perhaps later on many?)

PR #3132, opened 2025-02-28
Support docker compose

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/2907

(wip)


meowjesty:
Closing this as not-planned right now.

# For future reference

Should contain all the steps required to run `mirrord docker compose up` for a single service in a `compose.yaml` file. It supports having multiple services, but they might not be able to reach each other (have not tested this fully).

At the very least, it has code for reading a `compose config` yaml, and turn it into a modifiable compose that we can add our stuff to.

PR #3133, opened 2025-03-01
Improved agent dns errors, fixed metric

Razz4780:
Some E2E tests keep failing locally (minikube) on remote DNS. I tried to fix it, but failed. Hickory resolver keeps getting responses from a different IP than configured in `/etc/resolv.conf` in the target.

Noticed some space for improvements in the process:
1. Updated conversion of hickory errors to our `mirrord-protocol` representation
2. Agent no longer overrides resolver's default timeout and attempts, unless explicitly instructed via env (it used to override defaults with 1s timeout and 1 max attempt, dunno why)
3. Fixed the prometheus metric for inflight DNS requests
4. Moved conversion from hickory to mirrord-protocol types to the agent, which allowed for removing hickory dependencies from mirrord-protocol


mirrord/agent/src/dns.rs
```diff
@@ -242,22 +286,121 @@ impl DnsApi {
...
+impl ProtocolConversion<LookupIpStrategy> for AddressFamily {
+    fn convert(self) -> LookupIpStrategy {
+        match self {
+            AddressFamily::Ipv4Only => LookupIpStrategy::Ipv4Only,
+            AddressFamily::Ipv6Only => LookupIpStrategy::Ipv6Only,
+            AddressFamily::Both => LookupIpStrategy::Ipv4AndIpv6,
+            AddressFamily::Any => LookupIpStrategy::Ipv4thenIpv6,
+
+            AddressFamily::UnknownAddressFamilyFromNewerClient => {
+                tracing::error!("Unknown address family in addrinfo request. Using IPv4 and IPv6.");
```
meowjesty, line 398: Hmm I think this should be `warn` level, since we just pick a default and proceed.

meowjesty (approved):
Only have 1 nit, so approving.

Nice one removing some of this stuff from the protocol.

mirrord/agent/src/dns.rs
Razz4780: This can only happen when we have a bug in the layer tho

Razz4780: Kind of weird that it makes it to the agent, but oh well

PR #3134, opened 2025-03-01
Fix/improve outgoing e2e tests

Razz4780:
1. Fixed flaky `traffic::traffic_tests::outgoing_traffic_make_request_localhost`. The node.js test app was simply evil and kept failing even without mirrord (it should work just fine after changing the server's port number).
2. Merged `traffic::traffic_tests::test_outgoing_traffic_many_requests_disabled` and `traffic::traffic_tests::test_outgoing_traffic_many_requests_enabled` into one test using `#[case]`. Since this test relies on availability of some publicly accessible servers, I extended the host list and added a sanity check to the test. Now the test itself tries to make requests to the servers, before starting the test app. The app gets a list of hosts to make requests to. Test still fails if it can't find at least 10 available servers.


tests/node-e2e/outgoing/test_outgoing_traffic_many_requests.mjs
```diff
@@ -1,62 +1,41 @@
...
-      response.on("data", (data) => {
-        process.stdout.write(`>> received ${data.slice(0, 4)}`);
-      });
-
-      response.on("error", (fail) => {
-        process.stderr.write(`>> response from ${host} failed with ${fail}`);
-        throw fail;
-      });
-    });
+const HOSTS = process.env.AVAILABLE_HOSTS;
```
DmitryDodzin, line 3:
```suggestion
const AVAILABLE_HOSTS = process.env.AVAILABLE_HOSTS;
```

I would keep the var name the same or similar (ie with `_ENV` suffix) name for easy find and replaces

tests/src/traffic.rs
```diff
@@ -627,45 +627,95 @@ mod traffic_tests {
...
-            None,
-            None,
-        )
-        .await;
 
-        let res = process.child.wait().await.unwrap();
-        assert!(res.success());
-        process.assert_no_error_in_stderr().await;
-    }
+        let client = reqwest::Client::new();
```
DmitryDodzin, line 673: I think changing this to client builder and setting connect and read timeouts needed here (maybe also a concurrency if we somehow trigger some anti scrape thing and that is the reason for issues from the first place) otherwise it won't fix the issue because the host may become unavailable (also a global timeout can be helpful but connect and read should be fine)

DmitryDodzin (changes requested):
Small change to the the rust checking code before the test

tests/src/traffic.rs
Razz4780: Added 5s timeouts for connect and read, and the requests are now sequential ^^

DmitryDodzin: Sequential can be just as problematic because there are more than 2 or 3 endpoints so I would just consider using the relevant tower_layer on the [client builder itself](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connector_layer)

DmitryDodzin: Like in worst case senario the distance from first and last request is 5s * all the timeouts in between

DmitryDodzin: Also a sidenote nitpick maybe we should add a user-agent header something like
```rust
concat!(
    env!("CARGO_PKG_NAME"),
    "/",
    env!("CARGO_PKG_VERSION"),
);
```

DmitryDodzin: Like the parallel nature is fine just afraid of some issue due to sudden multy-dns + multy addr requests originating from a single addr will trip up some sort of protection (I can be also a bit paranoid about it)

tests/src/traffic.rs
```diff
@@ -627,45 +628,108 @@ mod traffic_tests {
...
+            .connect_timeout(Duration::from_secs(5))
+            .read_timeout(Duration::from_secs(5))
+            .default_headers(HeaderMap::from_iter([(
+                USER_AGENT,
+                HeaderValue::from_static(concat!(
+                    env!("CARGO_PKG_NAME"),
+                    "/",
+                    env!("CARGO_PKG_VERSION"),
+                )),
+            )]))
```
DmitryDodzin, line 684:
```suggestion
            .user_agent(concat!(
                  env!("CARGO_PKG_NAME"),
                  "/",
                  env!("CARGO_PKG_VERSION"),
             ))
```

Request has a shorthand for it

tests/src/traffic.rs
```diff
@@ -627,45 +628,108 @@ mod traffic_tests {
...
-            None,
-        )
-        .await;
 
-        let res = process.child.wait().await.unwrap();
-        assert!(res.success());
-        process.assert_no_error_in_stderr().await;
-    }
+        let client = reqwest::Client::builder()
+            .connect_timeout(Duration::from_secs(5))
```
DmitryDodzin, line 675: I think connect can be shorter like 2 seconds, it should respond at most `~500ms`

DmitryDodzin (approved):
I think it should be fine, hopefully there won't be much timeouts anyway

Issue #3120, opened 2025-02-25
Issue a warning to the user when their HTTP filter is preempted by another user


Razz4780:
HTTP filtering is a feature of mirrord that allows the user to steal only a subset of HTTP requests coming to the remote target (see reference [here](https://mirrord.dev/docs/using-mirrord/steal/#stealing-only-a-subset-of-the-remote-targets-traffic) and [here](https://mirrord.dev/docs/reference/configuration/#feature-network-incoming-http-filter).

When the mirrord-agent captures a request, it compares it against HTTP filters from all connected clients [here](https://github.com/metalbear-co/mirrord/blob/592c598a31e0238ae22d80f11c4528dceaea3feb/mirrord/agent/src/steal/connections/filtered.rs#L513). It may happen that the request matches more than one filter. When this happens, only one of the clients receives the request, and the other clients don't even know it happened. We want to notify them about this event to improve the UX. For now, a simple [warning](https://github.com/metalbear-co/mirrord/blob/592c598a31e0238ae22d80f11c4528dceaea3feb/mirrord/protocol/src/codec.rs#L184) should be enough.

Also, we'd like to have a new unit test for this logic (unless it turns out it's already covered by existing `FilteredStealTask` unit tests).

PR #3135, opened 2025-03-02
Warn clients of matching HTTP requests that went to another client (#3120)

ivonindza:
* Steal API now sends type StealerMessage to the agent, which can contain either DaemonTcp or LogMessage.
* If an HTTP request matches more than one filter, the request is forwarded to the first matching client, while the other matching clients recevive a warning via LogMessage.

### DIFF
--- a/changelog.d/3120.changed.md
+++ b/changelog.d/3120.changed.md
@@ -1,0 +1,6 @@
+Warn clients of matching HTTP requests that went to another client
+* Steal API now sends type StealerMessage to the agent, which can contain
+  either DaemonTcp or LogMessage.
+* If an HTTP request matches more than one filter, the request is forwarded to
+  the first matching client, while the other matching clients recevive a
+  warning via LogMessage.
--- a/changelog.d/3120.changed.md
+++ b/changelog.d/3120.changed.md
@@ -1,6 +1,2 @@
-Warn clients of matching HTTP requests that went to another client
-* Steal API now sends type StealerMessage to the agent, which can contain
-  either DaemonTcp or LogMessage.
-* If an HTTP request matches more than one filter, the request is forwarded to
-  the first matching client, while the other matching clients recevive a
-  warning via LogMessage.
+If a stolen HTTP request matches filters of multiple users,
+the users who don't get the request are now notified with a log message.
--- a/mirrord/agent/src/steal/connection.rs
+++ b/mirrord/agent/src/steal/connection.rs
@@ -30,7 +30,10 @@ use tokio::{
 use tokio_util::sync::CancellationToken;
 use tracing::Level;
 
-use super::{http::HttpResponseFallback, subscriptions::PortRedirector, tls::StealTlsHandlerStore};
+use super::{
+    api::StealerMessage, http::HttpResponseFallback, subscriptions::PortRedirector,
+    tls::StealTlsHandlerStore,
+};
 use crate::{
     error::{AgentError, AgentResult},
     metrics::HTTP_REQUEST_IN_PROGRESS_COUNT,
@@ -91,11 +94,17 @@ impl From<SendError<DaemonTcp>> for PassRequestError {
     }
 }
 
+impl From<SendError<StealerMessage>> for PassRequestError {
+    fn from(_: SendError<StealerMessage>) -> Self {
+        Self::ClientDisconnected
+    }
+}
+
 /// A stealer client.
 struct Client {
     /// For sending messages to client's [`TcpStealerApi`](super::api::TcpStealerApi).
     /// Comes to [`TcpConnectionStealer`] in [`Command::NewClient`].
-    tx: Sender<DaemonTcp>,
+    tx: Sender<StealerMessage>,
     /// Client's [`mirrord_protocol`] version.
     ///
     /// Determines which variant of [`DaemonTcp`] we use to send stolen HTTP requests.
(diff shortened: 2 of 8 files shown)


changelog.d/3120.changed.md
```diff
@@ -0,0 +1,6 @@
+Warn clients of matching HTTP requests that went to another client
+* Steal API now sends type StealerMessage to the agent, which can contain
+  either DaemonTcp or LogMessage.
+* If an HTTP request matches more than one filter, the request is forwarded to
+  the first matching client, while the other matching clients recevive a
+  warning via LogMessage.
```
Razz4780, line 6:
```suggestion
If a stolen HTTP request matches filters of multiple users,
the users who don't get the request are now notified with a log message.
```
Changelog entries (except `internal`) are meant for the users, so I'd keep them concise and purely functional

mirrord/agent/src/steal/connection.rs
```diff
@@ -91,11 +94,17 @@ impl From<SendError<DaemonTcp>> for PassRequestError {
     }
 }
 
+impl From<SendError<StealerMessage>> for PassRequestError {
```
Razz4780, line 97: The other `From` impl looks like a dead code now, so could be removed

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -508,7 +508,7 @@ where
         )
         ret,
     )]
-    fn match_request<B>(&self, request: &mut Request<B>) -> Option<ClientId> {
+    fn match_request<B>(&self, request: &mut Request<B>) -> (Option<ClientId>, Vec<ClientId>) {
```
Razz4780, line 511: This could return just one `Vec` of matched clients. The caller can handle it with a nice [`slice::split_first`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_first)

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -672,6 +680,15 @@ where
         self.blocked_requests
             .insert((client_id, id), request.response_tx);
 
+        for client_id in other_client_ids {
+            tx.send(ConnectionMessageOut::LogMessage {
+                client_id,
+                connection_id: self.connection_id,
+                message: LogMessage::warn("Matching HTTP request stolen by another client".into()),
```
Razz4780, line 687:
```suggestion
                message: LogMessage::warn(format!(
                    "An HTTP request was stolen by another user. URI=({:?}), HEADERS=({:?})",
                    request.request.uri(),
                    request.request.headers(),
                )),
```
I'd add request uri and headers for more context, since that's what we filter on

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -1241,6 +1284,92 @@ mod test {
...
+                                LogMessage {
+                                    message: _,
+                                    level: LogLevel::Warn,
+                                },
+                        } => {
+                            assert_ne!(received_client_id, subscribed_client_id);
+                        }
+                        other => unreachable!("unexpected message: {other:?}"),
+                    }
+                }
```
Razz4780, line 1341: This code assumes an exact order of messages produced by the filtered steal task (namely - `SubscribedHttp` and `Request` are sent before `LogMessage`s). Should the order change in the future, we'll have to adjust the test. Since we don't really care about the order here, this could be sth like:
```rust
let mut subscribe_received = false;
let mut request_received = false;
let mut logs_received = false;
for _ in 0..4 {
    match setup.task_out_rx.recv().await.unwrap() {
        ...
    }
}
```

The suggestion is optional though, it's good as it is ^^

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -1241,6 +1284,92 @@ mod test {
...
+                        ..
+                    } => {
+                        assert_eq!(received_client_id, subscribed_client_id);
+                        assert_eq!(destination, setup.original_address);
+                        id
+                    }
+                    other => unreachable!("unexpected message: {other:?}"),
+                };
+
+                // The remaining two cleints receive the warning
```
Razz4780, line 1325:
```suggestion
                // The remaining two clients receive the warning
```

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -1135,6 +1153,31 @@ mod test {
                 .unwrap()
         }
 
+        /// Prepares a new [`Request`] to be sent via [`Self::request_sender`]. Inserts the same
+        /// header filter for multiple clients.
+        ///
+        /// # Params
+        ///
+        /// 1. `client_ids` - For each client_id, insert a header filter. The request will contain a
+        ///    corresponding header such that the [`FilteredStealTask`] will steal it..
```
Razz4780, line 1162:
```suggestion
        ///    corresponding header such that the [`FilteredStealTask`] will steal it.
```

Razz4780 (changes requested):
Very nice and clean, thanks ^^
Left some nitpicks

mirrord/agent/src/steal/connections/filtered.rs
ivonindza: Makes sense, wasn't sure what info would be useful to log

mirrord/agent/src/steal/connections/filtered.rs
ivonindza: I agree that we don't care about the order in which the three clients receive messages and that it could change. I'd leave this as is, because I return `subscribed_client_id` from the first match to use for the unsubscribe message, so I think this way it's a bit more readable.

mirrord/agent/src/steal/connections/filtered.rs
ivonindza: That's true. I did it this way on purpose to signify that the first match is different from the others, because it gets processed differently. If you prefer to return to return just the `Vec`, just let me know and I'll change it.

Razz4780 (approved):
Nice, thank you! ^^

PR #3139, opened 2025-03-03
Try portforwarding for sqs e2e

Razz4780:


tests/src/http.rs
```diff
@@ -8,17 +8,15 @@ mod http_tests {
...
-        kube_client, send_requests, service, service_addr::TestServiceAddr, Application,
+        kube_client, port_forwarder::PortForwarder, send_requests, service, Application,
         KubeService,
     };
 
     /// ## Warning
     ///
-    /// These tests are marked with `ignore` due to flakyness!
+    /// This test is marked with `ignore` due to flakyness!
     #[ignore]
```
meowjesty, line 18: Still flaky even after your mirror fixes?

meowjesty (approved):
:+1:

tests/src/http.rs
Razz4780: Ah, haven't checked that. I'll open an issue for this

Issue #3092, opened 2025-02-13
Fix `HTTP_REQUEST_IN_PROGRESS_COUNT` metric


Razz4780:
### Bug Description

Gets out of sync pretty quickly

### Steps to Reproduce

---

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### Your operating system and version

---

### Local process

---

PR #3140, opened 2025-03-04
Decrement `MIRRORD_AGENT_HTTP_REQUEST_IN_PROGRESS_COUNT` metric whenever a request gets completed.

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/3092

Improves the situation, now my metrics are looking like this:

image

Which looks correct to me, considering that the app I'm using to test is sending requests from 4 different threads.

Q: Do we want to keep these:

```rs
HTTP_REQUEST_IN_PROGRESS_COUNT.store(self.response_body_txs.len() as i64, Ordering::Relaxed);
```

when the response did not end in an error?


mirrord/agent/src/steal/connection.rs
```diff
@@ -837,6 +837,7 @@ where
 
             Command::HttpResponse(response) => {
                 self.send_http_response(client_id, response).await;
+                HTTP_REQUEST_IN_PROGRESS_COUNT.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
```
Razz4780, line 840: The response is still be in progress. We're still getting body chunks

mirrord/agent/src/steal/api.rs
```diff
@@ -262,20 +260,16 @@ impl TcpStealerApi {
                     }
                     if send_err || body.is_last {
                         self.response_body_txs.remove(key);
-                        HTTP_REQUEST_IN_PROGRESS_COUNT.store(
-                            self.response_body_txs.len() as i64,
-                            std::sync::atomic::Ordering::Relaxed,
-                        );
+                        HTTP_REQUEST_IN_PROGRESS_COUNT
+                            .store(self.response_body_txs.len() as i64, Ordering::Relaxed);
```
Razz4780, line 264: There are multiple `TcpStealerApi`s

mirrord/agent/src/steal/api.rs
```diff
@@ -185,15 +185,15 @@ impl TcpStealerApi {
...
             LayerTcpSteal::PortSubscribe(port_steal) => self.port_subscribe(port_steal).await,
             LayerTcpSteal::ConnectionUnsubscribe(connection_id) => {
                 self.response_body_txs
                     .retain(|(key_id, _), _| *key_id != connection_id);
-                HTTP_REQUEST_IN_PROGRESS_COUNT.store(
-                    self.response_body_txs.len() as i64,
-                    std::sync::atomic::Ordering::Relaxed,
-                );
+                HTTP_REQUEST_IN_PROGRESS_COUNT
+                    .store(self.response_body_txs.len() as i64, Ordering::Relaxed);
```
Razz4780, line 196: There are multiple `TcpStealerApi`s

mirrord/agent/src/steal/api.rs
```diff
@@ -262,20 +260,16 @@ impl TcpStealerApi {
...
                 }
                 ChunkedResponse::Error(err) => {
                     self.response_body_txs
                         .remove(&(err.connection_id, err.request_id));
-                    HTTP_REQUEST_IN_PROGRESS_COUNT.store(
-                        self.response_body_txs.len() as i64,
-                        std::sync::atomic::Ordering::Relaxed,
-                    );
+                    HTTP_REQUEST_IN_PROGRESS_COUNT
+                        .store(self.response_body_txs.len() as i64, Ordering::Relaxed);
```
Razz4780, line 272: There are multiple `TcpStealerApi`s

Razz4780 (changes requested):
Nope :crying_cat_face:

mirrord/agent/src/steal/connection.rs
meowjesty: Alright, now we check if it's the end of the stream. Is that good enough? From looking at the graph, it looks correct.

Razz4780 (approved):

PR #3142, opened 2025-03-05
Add E2E tests for unlink and unlinkat (+ fix unlink/unlinkat)

gememma:


mirrord/layer/src/file/ops.rs
```diff
@@ -407,7 +407,7 @@ pub(crate) fn rmdir(path: Detour<PathBuf>) -> Detour<()> {
 pub(crate) fn unlink(path: Detour<PathBuf>) -> Detour<()> {
     let path = common_path_check(path?, true)?;
 
-    let unlink = RemoveDirRequest { pathname: path };
+    let unlink = UnlinkRequest { pathname: path };
```
DmitryDodzin, line 410: Huh, wonder why did it work before hand

DmitryDodzin (approved):
LGTM

mirrord/layer/src/file/ops.rs
gememma: 🤷 no idea, probably just (un)lucky

PR #3144, opened 2025-03-06
Filter the resolved config out from env var in debug log

gememma:


mirrord/layer/src/lib.rs
```diff
@@ -370,13 +370,18 @@ fn layer_start(mut config: LayerConfig) {
     enable_hooks(state);
 
     let _detour_guard = DetourGuard::new();
+
+    // remove resolved encoded config from env vars when logging them
+    let env_vars_print_only: Vec<_> = std::env::vars()
+        .filter(|(k, _v)| k != MIRRORD_RESOLVED_CONFIG_ENV)
+        .collect();
```
Razz4780, line 377: No need to collect, I think. `Debug` impl should print entries of the iterator anyway

Razz4780 (approved):
Yay

mirrord/layer/src/lib.rs
```diff
@@ -370,13 +370,18 @@ fn layer_start(mut config: LayerConfig) {
...
     tracing::info!("Initializing mirrord-layer!");
     tracing::debug!(
         executable = ?EXECUTABLE_PATH.get(),
         args = ?EXECUTABLE_ARGS.get(),
         pid = std::process::id(),
         parent_pid = parent_id(),
-        env_vars = ?std::env::vars(),
+        env_vars = ?env_vars_print_only,
```
Razz4780, line 384: On a second thought - this might be confusing in the future. I think this log should state that the config was omitted. Also, can you add variable value hash here? Just in case

gememma: > Also, can you add variable value hash here?

wdym? a hash of the config value?

PR #3146, opened 2025-03-06
Suspend layer requests until protocol is negotiated with the agent

Razz4780:
`go_statfs` E2E test was flaking, because `StatFsV2` request requires knowing mirrord-protocol version used by the agent.


meowjesty (approved):
:+1: 

Got me thinking, do we even need mirrord/intproxy to do version negotiation at all? Let's have 2 cases here:

1. `mirrord exec` with operator: the operator is giving us an agent to connect to, so the operator could inspect it and fill in the version before mirrord even started anything, the cli could ask the operator for all the versions we care about, then we start mirrord/intproxy/whatever with these versions set;
2. without the operator: while we're creating the agent, we could make the protocol version part of the agent spec, then just check it before starting things.

Wouldn't that be simpler than having to pause/suspend stuff until we have version negotiation in the middle of execution?

Razz4780:
> 👍
> 
> Got me thinking, do we even need mirrord/intproxy to do version negotiation at all? Let's have 2 cases here:
> 
>     1. `mirrord exec` with operator: the operator is giving us an agent to connect to, so the operator could inspect it and fill in the version before mirrord even started anything, the cli could ask the operator for all the versions we care about, then we start mirrord/intproxy/whatever with these versions set;
> 
>     2. without the operator: while we're creating the agent, we could make the protocol version part of the agent spec, then just check it before starting things.
> 
> 
> Wouldn't that be simpler than having to pause/suspend stuff until we have version negotiation in the middle of execution?

Yeah, this could be done more smoothly, but there is one issue - agent config in the operator can be changed at runtime :V

What we have works, so... :man_shrugging:

meowjesty:
>agent config in the operator can be changed at runtime

Sure, but won't we get the latest agent config when mirrord cli is doing it's operator talking shenanigans? Or do you mean that mirrord is supposed to recover from a session where it lost its connection, the agent config gets changed, mirrord reconnects and now we might be unmatched in version?

PR #3148, opened 2025-03-06
show a warning from the CLI when mirrord config allows for stealing health checks from the target

### LINKED ISSUE #3121: Issue a warning from the CLI when mirrord config allows for stealing health checks from the target
[Stealing incoming traffic](https://mirrord.dev/docs/using-mirrord/steal/) is a feature of mirrord that allows the user to intercept TCP connections or HTTP requests coming to the remote target. This is known to interfere with container [health check probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/), which are a standard part of Kubernetes workloads. When the user application steals health check probes from the target container, it often fails the probes, which results in Kubernetes terminating the session target in the cluster. This can be avoided using an [HTTP filter](https://mirrord.dev/docs/reference/configuration/#feature-network-incoming-http-filter).

Because health check probes are part of the resource spec, we can detect potential issues early, when we [fetch the target](https://github.com/metalbear-co/mirrord/blob/main/mirrord/cli/src/connection.rs#L113). We should issue a warning (and suggest using an HTTP filter) when we know for sure that the health check will be captured.

For know, it's sufficient to warn only when the used mirrord config allows to steal the port without the filter and the probe is either `httpGet` or `grpc`. We should inspect both the readiness and the liveness probe.

We don't have to worry about:
1. Whether the HTTP filter allows for stealing the probe (out of scope)
2. Startup probes (the container must be ready to start the session anyway)
3. `tcpSocket` probes (the agent will accept the connection anyway)
4. `exec` probes (can be an arbitrary command, no way to verify this)
5. Stealing probes when the copy target feature (reference [here](https://mirrord.dev/docs/using-mirrord/copy-target/) and [here](https://mirrord.dev/docs/reference/configuration/#feature-copy_target)) is used (probes will be disabled anyway)

TheYahya:
Fix #3121


mirrord/kube/src/resolved.rs
```diff
@@ -532,3 +532,265 @@ impl ResolvedTarget<false> {
...
+    /// This method determines whether the specified Kubernetes resource has any probe
+    /// configurations in its container specifications.
+    ///
+    /// # Returns
+    /// - `Ok(true)`: If the resource has at least one probe defined in the target container.
+    /// - `Ok(false)`: If no probes are found.
+    /// - `Err(KubeApiError)`: If an invalid state is encountered (e.g., no available replicas,
+    ///   missing fields, etc.)
+    #[tracing::instrument(level = Level::DEBUG, skip(client), ret, err)]
+    pub async fn has_probes(self, client: &Client) -> Result<bool, KubeApiError> {
```
Razz4780, line 548: This should not consume `self`, it means a lot of unnecessary cloning

Razz4780, line 548: Now that I see it in the code, I know I was wrong when preparing the issue description :sweat_smile: 

We need to check probes in all containers in the target pod, not only that one we target. All containers in the same pod run in a common network namespace, so our stealing affects all.
Sorry :sweat_smile:

mirrord/operator/src/client.rs
```diff
@@ -610,6 +610,24 @@ impl OperatorApi<PreparedClientCert> {
...
+                    .incoming
+                    .http_filter
+                    .is_filter_set()
+                && target
+                    .clone()
+                    .has_probes(self.client())
+                    .await
+                    // Default to `false` on error since this value is only used
+                    // for displaying a warning.
+                    .unwrap_or(false)
```
Razz4780, line 626: It might be that `feature.network.incoming` config explicitly specifies ports to steal from and there is no clash. E.g we have a TCP probe on port 3000 and the incoming config specifies stealing only from port 80

mirrord/kube/src/resolved.rs
```diff
@@ -532,3 +532,265 @@ impl ResolvedTarget<false> {
         }
     }
 }
+
+impl ResolvedTarget<true> {
```
Razz4780, line 536: This can be implemented on the `ResolvedTarget` regardless of the `CHECKED` param

mirrord/operator/src/client.rs
```diff
@@ -610,6 +610,24 @@ impl OperatorApi<PreparedClientCert> {
         } else {
             let target = target.assert_valid_mirrord_target(self.client()).await?;
 
+            if layer_config.feature.network.incoming.is_steal()
```
Razz4780, line 613: This code runs only when mirrord is used with the mirrord Operator (mfT, paid version). The OSS flow is different, the code paths split [here](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/cli/src/connection.rs#L121)

Razz4780 (changes requested):
I suggest a bit different approach - add all probed ports to the [`RuntimeData`](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L76) struct. We use it in both the operator and the oss flow, e.g to check if we've guessed the target container out of many, and warn the user.

TheYahya:
> I suggest a bit different approach - add all probed ports to the [RuntimeData](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L76) struct. We use it in both the operator and the oss flow, e.g to check if we've guessed the target container out of many, and warn the user.

Hi @Razz4780, So this means to do this work (inspecting the spec) in [from_pod](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L100) and then validating and printing the warning [here](https://github.com/metalbear-co/mirrord/blob/dd0c92c1edfc40f5095f72d703955181f9635226/mirrord/kube/src/api/kubernetes.rs#L219)? Or am I missing something?

mirrord/operator/src/client.rs
TheYahya: Yeah, this makes sense.

mirrord/kube/src/resolved.rs
TheYahya: Yes, unless we move it to [RuntimeData](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L76)

Razz4780:
> > I suggest a bit different approach - add all probed ports to the [RuntimeData](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L76) struct. We use it in both the operator and the oss flow, e.g to check if we've guessed the target container out of many, and warn the user.
> 
> Hi @Razz4780, So this means to do this work (inspecting the spec) in [from_pod](https://github.com/metalbear-co/mirrord/blob/10ceb6583c80bbcac20424c4d99db7228462d9c9/mirrord/kube/src/api/runtime.rs#L100) and then validating and printing the warning [here](https://github.com/metalbear-co/mirrord/blob/dd0c92c1edfc40f5095f72d703955181f9635226/mirrord/kube/src/api/kubernetes.rs#L219)? Or am I missing something?

Yeah, `from_pod` is a good place to inspect the spec and gather relevant info. `KubernetesAPI::create_agent` is a good place to print the warning as well, but mind that it's called only in the OSS flow ^^
So we still need the same warning in the `OperatorApi`

TheYahya:
hey @Razz4780, 
I've moved most of the logic to `RuntimeData`. And now we cover both OSS and Operator flows. I moved it to draft to make sure the approach is correct and I'm not missing something? I'll clean it up if it make sense.

mirrord/kube/src/api/runtime.rs
```diff
@@ -89,6 +90,8 @@ pub struct RuntimeData {
     pub mesh: Option<MeshVendor>,
 
     pub share_process_namespace: bool,
+    // container name to probe ports
+    pub containers_probe_ports: Vec<i32>,
```
Razz4780, line 94: I don't why why it's `i32` in the kube.rs struct, maybe because the original jsonschema allows it. But I'd transform this early into the correct `u16` and ignore invalid port numbers

mirrord/kube/src/api/runtime.rs
```diff
@@ -269,6 +282,43 @@ impl RuntimeData {
...
+    fn probe_ports(container: &Container) -> Vec<i32> {
+        fn port_to_int(port: &IntOrString) -> Option<i32> {
+            match port {
+                IntOrString::Int(p) => Some(*p),
+                IntOrString::String(s) => match s.as_str() {
+                    "http" => Some(80),
+                    "https" => Some(443),
+                    "grpc" => Some(50051),
+                    _ => s.parse().ok(),
+                },
```
Razz4780, line 295: This is the name of a named port inside the container (see [this](https://docs.rs/k8s-openapi/latest/k8s_openapi/api/core/v1/struct.ContainerPort.html) and [this](https://docs.rs/k8s-openapi/latest/k8s_openapi/api/core/v1/struct.Container.html#structfield.ports)), not the protocol name

mirrord/kube/src/api/runtime.rs
```diff
@@ -269,6 +282,43 @@ impl RuntimeData {
             NodeCheck::Success
         }
     }
+
+    fn probe_ports(container: &Container) -> Vec<i32> {
```
Razz4780, line 286: This could use a doc. Why ignore the TCP probes inside, so it's worth mentioning why

mirrord/operator/src/client.rs
```diff
@@ -628,6 +628,43 @@ impl OperatorApi<PreparedClientCert> {
...
+                            .network
+                            .incoming
+                            .http_filter
+                            .is_filter_set()
+                            && layer_config
+                                .feature
+                                .network
+                                .incoming
+                                .port_mapping
+                                .is_empty())
```
Razz4780, line 663: This check is present both here and in `create_agent`, I think it could be extracted to a method like `IncomingConfig::steals_port_without_filter(&self, port: u16)`

Razz4780, line 663: Our user-side configuration is bit complex, I probably should've given the exact check logic in the issue description :sweat_smile: 
Should be sth like:
```
incoming.is_steal().not() => false
incoming.http_filter.is_filter_set() && incoming.http.filter.ports.contains(port) => false
incoming.ignore_ports.contains(port) => false
incoming.ports.contains(port).not() => false
true
```

Razz4780 (changes requested):

Razz4780:
> hey @Razz4780, I've moved most of the logic to `RuntimeData`. And now we cover both OSS and Operator flows. I moved it to draft to make sure the approach is correct and I'm not missing something? I'll clean it up if it make sense.

Hey! Yes, the new approach is correct, thanks. I've left some last comments

mirrord/operator/src/client.rs
TheYahya: Hopefully I've got it correct in the last commit 😄

mirrord/kube/src/api/runtime.rs
TheYahya: Wow! I've make this mistake before! Should've read the docs!

Razz4780 (approved):
Nice! Thanks ^^

PR #3170, opened 2025-03-17
Added missing permissions for `PodTemplate`s

Razz4780:
We need to fetch the workload ref when preparing the session


DmitryDodzin (approved):
LGTM

PR #3173, opened 2025-03-17
strip mirrord dir from exec path in applev

### LINKED ISSUE #3149: VSCode + Go debugger regression in 3.133.1
### Bug Description

3.133.0 works, 3.133.1 start breaking.
When clicking debug in VSCode for a go app it crashes with
```
Build Error: go build -o /Users/aviramhassan/Code/playground/ip-visit-counter/__debug_bin2648580981 -gcflags all=-N -l ./main.go
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set (exit status 2)
```

### Steps to Reproduce

1. Download go 1.23.3 (it might happen with other versions but not sure). It doesn't happen with the brewed version I had so use the site
2. Enable mirrord in the IDE in a go project and start debugging

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### mirrord CLI version

_No response_

### mirrord-agent version

_No response_

_No response_

_No response_

15.3.1

### Local process

go

### Local process version

_No response_

### Additional Info

_No response_

t4lz:
Fix #3149

WIP - for some reason the patched binary is still resistant to DYLD_INSERT_LIBRARIES.


t4lz:
We probably don't need this, at least at this point. When manually signing the patched binary, there is no problem, even without this change. So to solve the issue we probably only need to fix the signing.

We should keep this branch in mind if we ever have problems with go's `os.Executable` returning the patched path.

PR #3179, opened 2025-03-18
Fix code signing in SIP-patching

### LINKED ISSUE #3149: VSCode + Go debugger regression in 3.133.1

t4lz:
Fix #3149 

The new signing method was not overwriting the existing flags of a binary, so binaries like the `go` binary, that have the "runtime" flag set, were running in a hardened runtime even after the patch, and were therefore resistant to dynamic library injection.

### DIFF
--- a/changelog.d/3149.fixed.md
+++ b/changelog.d/3149.fixed.md
@@ -1,0 +1,1 @@
+Regression in running some SIP-protected binaries with mirrord.
--- a/mirrord/sip/src/codesign.rs
+++ b/mirrord/sip/src/codesign.rs
@@ -13,6 +13,9 @@ pub(crate) fn sign<PI: AsRef<Path>, PO: AsRef<Path>>(input: PI, output: PO) -> R
     // but if we sign in place we get permission error, probably because someone is holding the
     // handle to the named temp file.
     let mut settings = SigningSettings::default();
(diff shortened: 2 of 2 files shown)


Razz4780:
> > Can this break something or is it essentially a revert?
> 
> I don't think this will break anything, only mend

Ok :pray:

Razz4780 (approved):

PR #3181, opened 2025-03-19
MirrordProfile

Razz4780:
Example profile:
```yaml
apiVersion: profiles.mirrord.metalbear.co/v1alpha
kind: MirrordProfile # clusterwide resource
metadata:
  name: mirror-and-outgoing-local
spec:
  # Changes to be made in the feature config.
  featureAdjustments:
  - kind: incoming
    change: mirror
  - kind: outgoing
    change: off
  - kind: dns
    change: off
```

### DIFF
--- a/mirrord/operator/src/crd/profile.rs
+++ b/mirrord/operator/src/crd/profile.rs
@@ -32,6 +32,7 @@ pub struct MirrordProfileSpec {
 
 /// A single adjustment to the mirrord config's `feature` section.
 #[derive(Deserialize, Serialize, JsonSchema, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
 pub struct FeatureAdjustment {
     /// The kind of feature to which this adjustment applies.
     pub kind: FeatureKind,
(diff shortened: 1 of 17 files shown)


aviramha:
Perhaps we should use same values as in mirrord.json? it's quite weird in one place it's off/on and now it's disabled

mirrord/operator/src/crd/profile.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// The CLI should error when the profile contains unknown fields.
+    #[schemars(skip)]
+    #[serde(flatten, skip_serializing)]
+    pub unknown_fields: HashMap<String, Value>,
+}
+
+/// A single adjustment to the mirrord config's `feature` section.
+#[derive(Deserialize, Serialize, JsonSchema, Debug, Clone)]
+pub struct FeatureAdjustment {
```
DmitryDodzin, line 35:
```suggestion
#[serde(rename_all = "camelCase")]
pub struct FeatureAdjustment {
```

mirrord/operator/src/crd/profile.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// The CLI should error when the profile contains unknown fields.
+    #[schemars(skip)]
+    #[serde(flatten, skip_serializing)]
+    pub unknown_fields: HashMap<String, Value>,
+}
+
+/// A kind of mirrord feature that can be adjusted in a mirrord policy.
+#[derive(Deserialize, Serialize, JsonSchema, Debug, Clone, Copy)]
+#[serde(rename_all = "camelCase")]
```
DmitryDodzin, line 51: I like kebab-case for flags but to your descression

Razz4780: Yeah, makes more sense. Changed :+1:

PR #3183, opened 2025-03-19
Add some sanity checks e2e tests for copy-target feature.

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/799


tests/node-e2e/copy_target_starts_a_working_copy.mjs
```diff
@@ -0,0 +1,25 @@
+import http from 'node:http';
```
Razz4780, line 1: Can you use one of the existing apps? :eyes:

tests/configs/copy_target_starts_a_working_copy.json
```diff
@@ -0,0 +1,7 @@
+{
```
Razz4780, line 1: Can you create the config in the test code? (`serde_json::json!` and `tempfile`)

Razz4780 (changes requested):

tests/node-e2e/copy_target_starts_a_working_copy.mjs
meowjesty: Oh I removed it thinking "why the heck have I added this?", but now I remember!

Calling `kill` was not stopping the app correctly, we need an app that starts and just ends, otherwise some pod keeps running forever (I don't remember if it's the agent or the copy right now).

tests/src/operator/copy_target.rs
```diff
@@ -0,0 +1,62 @@
+#![cfg(test)]
+// #![cfg(feature = "operator")]
```
Razz4780, line 2: Hello there

tests/src/operator/copy_target.rs
```diff
@@ -0,0 +1,62 @@
...
+    let config = serde_json::json!({
+        "feature": {
+            "copy_target": {
+                "scale_down": false
+            }
+        }
+    });
+    // Write the config to the temporary file
+    let config_path = config_file.path();
+    std::fs::write(&config_path, serde_json::to_string_pretty(&config).unwrap()).unwrap();
```
Razz4780, line 39:
```suggestion
    // Create a temporary file for the config
    let mut config_file = NamedTempFile::with_suffix(".json").unwrap();

    // Create the config JSON
    let config = serde_json::json!({
        "feature": {
            "copy_target": {
                "scale_down": false
            }
        }
    });
    // Write the config to the temporary file
    serde_json::to_writer(config_file.as_file_mut(), &config).unwrap();
```

tests/node-e2e/copy_target_starts_a_working_copy.mjs
Razz4780: I think you can safely use on of the `EnvApp` apps, example [here](https://github.com/metalbear-co/mirrord/blob/2f5eb130475470ab94f72d5fe230c2fd2a39fcd8/tests/src/env.rs#L25). This would also verify that we get the env correctly ^^

Razz4780 (changes requested):

tests/node-e2e/copy_target_starts_a_working_copy.mjs
meowjesty: Changed.

tests/src/operator/copy_target.rs
```diff
@@ -0,0 +1,66 @@
...
+pub async fn copy_target_starts_a_working_copy(
+    #[future]
+    #[notrace]
+    service: KubeService,
+    #[values(
+        EnvApp::Go21,
+        EnvApp::Go22,
+        EnvApp::Go23,
+        EnvApp::NodeInclude,
+        EnvApp::NodeExclude
```
Razz4780, line 26: No need to kill the cluster, I'd use just one

Razz4780 (approved):
:+1:

Issue #3184, opened 2025-03-20
pgrep crash on latest version (macOS)


aviramha:
### Bug Description

running pgrep from mirrord context leads to crash, I suspect it's due to our new signing method. 

### Steps to Reproduce

`mirrord exec pgrep`

### Backtrace

```shell
Binary Images:
       0x2050dc000 -        0x205167fff dyld_path_missing (*) <f2913392-361a-304f-b30d-486be5639e2d> /dyld_path_missing
    0x7ff7ffc91000 -     0x7ff7ffcc0fff main_executable_path_missing (*) <11b18945-107e-3e0d-91fa-b5d5ada3c0c4> /main_executable_path_missing
               0x0 - 0xffffffffffffffff ??? (*) <00000000-0000-0000-0000-000000000000> ???

Error Formulating Crash Report:
dyld_process_snapshot_get_shared_cache failed
PC register does not match crashing frame (0x0 vs 0x7FF7FFC93F7C)

EOF
```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### Your operating system and version

15.3.1 (24D70)

### Local process

pgrep

aviramha:
if I codesign it using `codesign -f -s - ` (old way) pgrep binary launches successfuly.

PR #3191, opened 2025-03-20
mirrord SIP crate docs

gememma:


mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
...
+//!
+//! ### Why?
+//!
+//! SIP makes it such that we are unable to load the `mirrord-layer` lib dynamically into the user's
+//! process onto MacOS machines. This is ostensibly a security measure to prevent the malicious or
+//! unintended modification of processes, but mirrord's dynamic library loading is neither. To get
+//! around this, we patch the user's process by creating a new unprotected version and signing it
+//! ourselves. Then we can load the layer dynamically into our unprotected copy and continue as
+//! usual. If the process is a script, the shebang must be pointed to the patched version of the
+//! interpreter.
```
meowjesty: Why do we need to load mirrord-layer lib? What does that even mean? I think this would be a good candidate point to link this doc to the place where we put an explanation of what has to happen (probably the crate doc for mirrord-layer itself). So this is not feedback saying that you should change this text per say, instead it's more of a future note.

I would also add some link that explains macos SIP around here, so you can look at more details. Think you can shorten the SIP explanation to just say that it blocks modification of processes. The focus here is how it impacts mirrord, and what the crate is doing to get us around this fact (which you are doing in this paragraph).

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
+//! This crate concerns MacOS' SIP (System Integrity Protection) mechanism and how to sidestep it.
+//!
+//! ### Why?
+//!
+//! SIP makes it such that we are unable to load the `mirrord-layer` lib dynamically into the user's
+//! process onto MacOS machines. This is ostensibly a security measure to prevent the malicious or
+//! unintended modification of processes, but mirrord's dynamic library loading is neither. To get
+//! around this, we patch the user's process by creating a new unprotected version and signing it
+//! ourselves. Then we can load the layer dynamically into our unprotected copy and continue as
+//! usual. If the process is a script, the shebang must be pointed to the patched version of the
```
meowjesty: Special cases (like the script process in this case) should probably get a `-` or be a bit separated from the general explanation, like those info-panels you see in books (maybe just a `-` list here, but I can see it being a whole sub-header in some cases, so let's not be too rigid).

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
...
+//! interpreter.
+//!
+//! ### Usage
+//!
+//! This crate is used by `mirrord-cli` and `mirrord-layer` and is used before the user process is
```
meowjesty: Instead of "is used by x and y", I think we should say **how** it's being used, so the crates that are using it will come naturally.

Like: Due to macos SIP, we can't simply dynamically load the `mirrord-layer` lib with `LD_PRELOAD=libmirrord.so`...

Which is kinda the explanation that you put on the previous paragraph.

The `mirrord-cli` mention will come in the explanation of how mirrord is started on macos, that the `cli` will perform some binary patching (whatever it does), and so on.

Something along those lines, wdyt?

gememma: hmm i dont know, i think its useful - maybe if you know what the crate does roughly, youre just checking how it fits into the code and you want a brief skim-able list of places where it gets used - that way, the `mirrord-layer` mention can stay in the `Why?` section and the docs for the `mirrord-cli` crate can mention the flow of starting mirrord on macos 🤷 Do you not think it could be useful at some point?

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
+//! Concerns MacOS' SIP (System Integrity Protection) mechanism and how to sidestep it.
```
Razz4780, line 1: Afaik this machinery requires something called Rosetta :eyes: 
Might be worth to mention in the doc

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
...
+//!
+//! ### Why?
+//!
+//! SIP makes it such that the `mirrord-layer` lib cannot be dynamically loaded into the user's
+//! process on MacOS machines. This is ostensibly a security measure to prevent the malicious or
+//! unintended modification of processes, but mirrord's dynamic library loading is neither. To get
```
t4lz, line 7: Maybe this is a nitpick - take it if you find it helpful:
The SIP-sidestepping is not ok and possible because mirrord is not malicious, it's ok and possible because the patched binary does not anymore have the properties that make the original binary need protection. For example, the patched binary does not have any entitlements the original binary had.
If we were able to inject a library into a binary that does have entitlements, that would be a vulnerability, because now our library would have entitlements it was never intentionally granted.
If we don't strip all of the entitlements off of the patched binary, the signature will be invalid, and the process will be killed immediately by `taskgated`, which is exactly what happened in #3184 (the signing method was changed, the old method used to strip entitlements by default, the new one did not).
Another example I could imagine (but don't know if it actually exists) for properties that make a file need protection, and are stripped in our patch: Apple might make the OS give some special privileges to binaries signed by apple.
As soon we resign the copied binary, it isn't signed by apple anymore, and therefore won't enjoy any special privileges, and it wouldn't be dangerous for us to load into it.

My point is: our SIP-sidestepping doesn't require extra trust, on a security level, in the layer not being malicious or buggy.

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
...
+//! usual. If the process is a script, the shebang must be changed to point to the patched version
+//! of the interpreter.
+//!
+//! ### Usage
+//!
+//! Used by `mirrord-cli` and `mirrord-layer`. Called into before the user process is
+//! run, during setup on MacOS machines - for example, see mirrord_layer::exec_utils::patch_if_sip
+//! and mirrord::execution::MirrordExecution::start_internal.
+//!
+//! The entry point when calling this crate is [`sip_patch`](main::sip_patch).
```
t4lz, line 19: It's also used in the exec hooks, for the case a program executes a protected binary.

mirrord/sip/src/lib.rs
t4lz, line 1: You could consider including links to apple's documentation about The Hardened Runtime, SIP, entitlements, codesigning.

mirrord/sip/src/lib.rs
```diff
@@ -1,3 +1,31 @@
+//! Concerns MacOS' SIP (System Integrity Protection) mechanism and how to sidestep it.
+//!
+//! ### Why?
+//!
+//! SIP makes it such that the `mirrord-layer` lib cannot be dynamically loaded into the user's
```
t4lz, line 5: We could mention it does that by just removing the relevant env vars when starting the process.

t4lz: Also, if we want to really tell the whole truth - it's not always SIP that we need to bypass - it's sometimes the hardened runtime.

t4lz (approved):
Looks good, need to fix the check-rust-docs errors.

mirrord/sip/src/lib.rs
gememma: That's fair - I think I can ditch some of the details here are just link to some docs (as you suggested) but it's good to know 👀

mirrord/sip/src/lib.rs
```diff
@@ -3,27 +3,32 @@
...
 //! of the interpreter.
 //!
 //! ### Usage
 //!
 //! Used by `mirrord-cli` and `mirrord-layer`. Called into before the user process is
-//! run, during setup on MacOS machines - for example, see mirrord_layer::exec_utils::patch_if_sip
-//! and mirrord::execution::MirrordExecution::start_internal.
+//! run, during setup on MacOS machines and in the exec hooks in case the program
+//! runs a protected binary - see mirrord::execution::MirrordExecution::start_internal,
+//! mirrord_layer::exec_hooks::hooks::execve_detour and mirrord_layer::exec_utils::patch_if_sip.
```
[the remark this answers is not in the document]
gememma: I would love to, but it creates circular dependencies - mirrord_layer depends on mirrord_sip, which then depends on mirrord_layer in order to link to it, etc.

PR #3192, opened 2025-03-20
Fix running binaries with entitlements on macOS

### LINKED ISSUE #3184: pgrep crash on latest version (macOS)

t4lz:
Fix #3184 

Recently the method of signing SIP-patched binaries on macOS was changed, and apparently the library that is now used has a different default behaviour than the cli tool used before.

When re-signing with the CLI, any existing entitlements are cleared. When signing with the library, they remain, unless overwritten with a plist passed in xml format.

Since the patched binary was no longer signed by the entity that was granted those entitlements, the signature was invalid and the process was killed by `taskgated`.

So in order to clear any entitlements the binary might have, we set a constant xml plist that contains an empty entitlement dict.

In addition, we also change the identifier, as the codesign man page says we should, to my understanding.

I'll also add a small unit test for the case of a binary with entitlements, but since this is urgent, it will be in a different PR, to not delay the release of the fix.

### DIFF
--- a/mirrord/sip/src/codesign.rs
+++ b/mirrord/sip/src/codesign.rs
@@ -1,4 +1,7 @@
-use std::{fmt::Display, path::Path};
+use std::{
+    fmt::{Display, Write},
+    path::Path,
+};
 
 use apple_codesign::{CodeSignatureFlags, SettingsScope, SigningSettings, UnifiedSigner};
 use rand::Rng;
@@ -11,8 +14,11 @@ const EMPTY_ENTITLEMENTS_PLIST: &str = r#"<?xml version="1.0" encoding="UTF-8"?>
 fn generate_hex_string<const N: usize>() -> String {
     let mut rng = rand::rng();
     (0..N / 2) // N/2 bytes = N hexadecimal digits.
-        .map(|_| format!("{:02x}", rng.random::<u8>()))
-        .collect()
+        .fold(String::new(), |mut acc, _next_in_range| {
+            // ignoring result: writing to String can't fail.
+            let _ = write!(acc, "{:02x}", rng.random::<u8>());
+            acc
+        })
 }
 
 pub(crate) fn sign<PI: AsRef<Path>, PO: AsRef<Path>, PR: Display>(
--- a/mirrord/sip/src/codesign.rs
+++ b/mirrord/sip/src/codesign.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Display, path::Path};
+use std::path::Path;
 
 use apple_codesign::{CodeSignatureFlags, SettingsScope, SigningSettings, UnifiedSigner};
 use rand::RngCore;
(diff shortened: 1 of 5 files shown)


mirrord/sip/src/codesign.rs
```diff
@@ -1,10 +1,31 @@
...
+/// N is the length of the string to generate. If N is odd, the length will be N-1.
+fn generate_hex_string<const N: usize>() -> String {
+    let mut rng = rand::rng();
+    (0..N / 2) // N/2 bytes = N hexadecimal digits.
+        .fold(String::new(), |mut acc, _next_in_range| {
+            // ignoring result: writing to String can't fail.
+            let _ = write!(acc, "{:02x}", rng.random::<u8>());
+            acc
+        })
+}
```
Razz4780, line 22: This is kinda weird, we have `hex` :eyes:

mirrord/sip/src/codesign.rs
```diff
@@ -13,10 +34,25 @@ pub(crate) fn sign<PI: AsRef<Path>, PO: AsRef<Path>>(input: PI, output: PO) -> R
...
     // opting into the hardened runtime, which strips away DYLD_INSERT_LIBRARIES etc.
     settings.set_code_signature_flags(SettingsScope::Main, CodeSignatureFlags::ADHOC);
+
+    // `man codesign`:
+    // > It is a **very bad idea** to sign different programs with the same identifier.
+    settings.set_binary_identifier(
+        SettingsScope::Main,
+        bin_id_prefix.map_or_else(
+            || format!("mirrord-patched-bin-{}", generate_hex_string::<40>()),
+            |prefix| format!("{}-{}", prefix, generate_hex_string::<40>()),
```
Razz4780, line 49: Kinda worried about [this](https://docs.rs/apple-codesign/0.29.0/apple_codesign/struct.SigningSettings.html#method.set_binary_identifier) - `The binary identifier should be a DNS-like name and should uniquely identify the binary. e.g. com.example.my_program`. The file name might contain invalid characters. Can we always use `mirrord.patched.bin.{hex}`?

t4lz: Changed in https://github.com/metalbear-co/mirrord/pull/3192/commits/a18d819e82a3f659c2a868788b02b583337fd7b1

Razz4780 (approved):
🚀 🌔

PR #3198, opened 2025-03-24
Started docs for mirrord-cli crate.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/168

(wip)

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -1,3 +1,103 @@
+//! # mirrord-cli
+//!
+//! CLI tool for everything mirrord!
+//!
+//! Most of the users interact with it through the mirrord IDE plugins (which mostly just runs
+//! `mirrord` commands behind the scenes).
+//!
+//! ## Features overview
+//!
+//! The main command to be aware of is [`mirrord exec`](link), and a couple of notable ones
+//! are [`mirrord operator`](link), and [`mirrord container`](link).
+//!
+//! ### `mirrord exec [OPTIONS] <BINARY> [BINARY_ARGS]`
+//!
+//! - [`exec`]
+//!
+//! > The bread and butter of mirrord.
+//!
+//! The `exec` command runs the user application with mirrord. You can pass pretty much any
+//! binary to it that you want to run in the target context (or just in the cluster context
+//! when running targetless) and it should work.
+//!
+//! - Tip: `mirrord exec -- sh` is a quick way of trying out changes.
+//!
+//! - Tip: You might want to use a `--` when invoking the command (e.g.`-- <BINARY> [BINARY_ARGS]`),
+//!   for some binaries the command will fail to get the proper args without it.
+//!
+//! An `exec` run starts the [`Progress`] logging and the [`AnalyticsReporter`], prepares
+//! the [`LayerConfig`], and checks it with [`LayerConfig::verify`] (which is similar to what's done
+//! in the `mirrord verify-config` command).
+//!
+//! - Tip: [`Progress`] logging might be incovenient sometimes when you want to see normal Rust
+//!   logs, you can disable it with the `MIRRORD_PROGRESS_MODE=off` env var.
+//!
+//! Next, we start the target resolution, and how the target is resolved depends if the
+//! mirrord-operator is available and enabled (see the section below). After [`create_and_connect`],
+//! we now have the mirrord-agent information that the mirrord-internal-proxy needs. We then run
+//! `mirrord intproxy` to start it, patch the user binary (macos only), and finally run the
+//! user binary with the mirrord lib loaded, but this time we use `execve`, instead of
+//! [`tokio::process::Command`].
+//!
+//! #### operator vs no operator `exec`
+//!
+//! Target resolution is performed the same, regardless of operator usage, but
+//! `exec` starts an agent when there is **no** operator, or the operator was explicitly
+//! **disabled** in [`LayerConfig::operator`]. Otherwise, the agent creation is handled by the
+//! operator, so in this case the `AgentConnectInfo` we get comes from the
+//! `OperatorSessionConnection` that was assigned for this run.
+//!
+//! Some mirrord features and targets are only supported when the operator is being used. `exec`
+//! usually stops when one of these is detected, logging an error to the user, be it in the terminal
+//! or in the IDE. [`Progress`] will take care of logging using the appropriate mechanism (stderr
+//! or IDE notification box with nice little buttons). For targets that might have multiple pods
+//! (`deployment` and `rollout`), the user is just warned that mirrord won't impersonate all the
+//! pods without the operator.
+//!
+//! ### `mirrord operator`
+//!
+//! - [`operator_command`]
+//!
+//! > How we pay the bills.
+//!
+//! A family of commands that help managing the mirrord-operator.
+//!
+//! #### `mirrord operator setup`
+//!
+//! - `operator_setup`
+//!
+//! Creates a kubernetes spec `yaml` file that can be used to install the mirrord-operator in the
+//! cluster. No validations are done here! The user may pass an invalid license, operator version
+//! that doesn't exist, whatever. The command will work as long as we can parse the args and produce
+//! a file from it.
+//!
+//! - Tip: The [operator helm chart](https://github.com/metalbear-co/charts/tree/main/mirrord-operator)
+//!   is an alternative way of setting up the mirrord-operator.
+//!
+//! #### `mirrord operator status`
+//!
+//! - `StatusCommandHandler`
+//!
+//! Uses the `OperatorApi` to access the `/status` route in the mirrord-operator and report it to
+//! the user.
+//!
+//! Prints a bunch of information about the mirrord operator `Session`s that are retrieved via the
+//! kubernetes API in the form of the `MirrordOperatorStatus` CRD(-ish, since most of this
+//! information is actually stored in the mirrord-operator itself, and not as a kubernetes
+//! resource).
+//!
+//! Does not interact with the IDE plugins, it's a terminal only command that pretty prints this
+//! information to stdout.
+//!
+//! #### `mirrord operator session`
+//!
+//! - `SessionCommandHandler`
+//!
+//! Uses the `OperatorApi` to manage (kill) mirrord-operator sessions (`SessionSpec` CRD). It makes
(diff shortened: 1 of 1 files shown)


mirrord/cli/src/main.rs
```diff
@@ -1,3 +1,103 @@
...
+//! - `operator_setup`
+//!
+//! Creates a kubernetes spec `yaml` file that can be used to install the mirrord-operator in the
+//! cluster. No validations are done here! The user may pass an invalid license, operator version
+//! that doesn't exist, whatever. The command will work as long as we can parse the args and produce
+//! a file from it.
+//!
+//! - Tip: The [operator helm chart](https://github.com/metalbear-co/charts/tree/main/mirrord-operator)
+//!   is an alternative way of setting up the mirrord-operator.
+//!
```
gememma, line 76: could add a note here that the easiest way to install the operator is to pass the output straight into `kubectl apply -f` like: `mirrord operator setup --accept-tos --license-path $LICENCE_PATH | kubectl apply -f -`
Also, for devs we have a ci-license.pem to use for the operator (i dont know where it exists?)

gememma: honestly, some/all of what i suggested might be better in the operator repo, idk

mirrord/cli/src/main.rs
```diff
@@ -1,3 +1,103 @@
...
+//! `mirrord` commands behind the scenes).
+//!
+//! ## Features overview
+//!
+//! The main command to be aware of is [`mirrord exec`](link), and a couple of notable ones
+//! are [`mirrord operator`](link), and [`mirrord container`](link).
+//!
+//! ### `mirrord exec [OPTIONS] <BINARY> [BINARY_ARGS]`
+//!
+//! - [`exec`]
```
gememma, line 15: i would like some kind of surrounding sentence on these, something like "See: [thing]" or "Handled by [thing]" or "Entrypoint: [thing]", thoughts?

mirrord/cli/src/main.rs
```diff
@@ -1,3 +1,103 @@
...
+//! the [`LayerConfig`], and checks it with [`LayerConfig::verify`] (which is similar to what's done
+//! in the `mirrord verify-config` command).
+//!
+//! - Tip: [`Progress`] logging might be incovenient sometimes when you want to see normal Rust
+//!   logs, you can disable it with the `MIRRORD_PROGRESS_MODE=off` env var.
+//!
+//! Next, we start the target resolution, and how the target is resolved depends if the
+//! mirrord-operator is available and enabled (see the section below). After [`create_and_connect`],
+//! we now have the mirrord-agent information that the mirrord-internal-proxy needs. We then run
+//! `mirrord intproxy` to start it, patch the user binary (macos only), and finally run the
```
gememma, line 38: You could link to the sip crate here (or whichever function in the cli crate does sip patching)

gememma (commented):
Real nice! My comments are more like musings, I think this stuff is really good - I agree that you hit on the right level of detail.

Mildly related, I think at some point (possibly after the crate level docs) we should make a cheat sheet/ glossary of terms for things like target, agent, layer (it took me months to internalise what target was without looking it up) and also a glossary of all the helpful env vars like `MIRRORD_PROGRESS_MODE` and `MIRRORD_OPERATOR_ENABLE`.

mirrord/cli/src/main.rs
meowjesty: > honestly, some/all of what i suggested might be better in the operator repo, idk

That's the hard part of documenting some of this stuff, I'm never sure where to put it.

Maybe in this case we put this doc (with more examples like setting it up with sqs) in the doc closer to the command? Maybe on the fn?

mirrord/cli/src/main.rs
meowjesty: Good point.

Initially I thought about having these as just sort of "tags" marking important fns or points to jump to. But it felt a bit too much, especially since `exec` does a lot of stuff (plus it would be annoying if we refactored some function names).

This is the second attempt at tagging, pointing out either the entrypoint (as is the case for `exec`), or the thing you should look for (in the case of commands that follow the other pattern, where it's `Thing`, `Thing::new`, and `Thing::handle`, as I feel these can be more easily followed, you know to start at `Thing::new`).

I wonder how much a text like "Entrypoint: `exec`" would add in value for a dev coming to look at these docs. With the curse of prior knowledge, I feel like just the "thing" is enough and helpful, you see this `[exec]` so you "Go-to-definition" on it and see what's there, but someone that has no clue may not get this at a glance.

Maybe we could also lay out what this 1st `-` list thingy means in the glossary, kinda like how books explain the different side-panels that they have?

mirrord/cli/src/main.rs
```diff
@@ -20,6 +20,9 @@
 //! binary to it that you want to run in the target context (or just in the cluster context
 //! when running targetless) and it should work.
 //!
+//! - **Attention**: The IDEs use the `mirrord ext` command, which is very similar to `mirrord
```
gememma, line 23: i would probably use something less like a hazard warning, maybe "NB: The IDEs use ..." ?

mirrord/cli/src/main.rs
```diff
@@ -54,15 +57,110 @@
...
+//!
+//! > Makes a neat `libmirrord_layer.so` file.
+//!
+//! The command itself is not really used anywhere (and I -alex- think nobody uses this). Other
+//! commands that are related to starting a mirrord instance use the [`extract_library`]
+//! function directly
+//!
+//! ### `mirrord verify-config [OPTIONS] <PATH>`
+//!
+//! - [`verify_config`]
```
gememma, line 144:
```suggestion
//! - [`verify_config()`]
```

mirrord/cli/src/main.rs
```diff
@@ -54,15 +57,110 @@
...
+//!
+//! It reads a previously resolved [`LayerConfig`] that has already been verified as valid, then
+//! intializes logging, either to a file in `/tmp`, or to stderr when it's being started from
+//! `mirrord container`.
+//!
+//! ### `mirrord container [OPTIONS] [EXEC]`
+//!
+//! - [`container_command`]
+//!
+//! > Runs an equivalent thing of `mirrord exec -- docker run {image}`.
```
gememma, line 91:
```suggestion
//! > Runs the equivalent of `mirrord exec -- docker run {image}`.
```

mirrord/cli/src/main.rs
```diff
@@ -54,15 +57,110 @@
...
+//!
+//! ### `mirrord extract <PATH>`
+//!
+//! - [`extract_library`]
+//!
+//! > Makes a neat `libmirrord_layer.so` file.
+//!
+//! The command itself is not really used anywhere (and I -alex- think nobody uses this). Other
+//! commands that are related to starting a mirrord instance use the [`extract_library`]
+//! function directly
```
gememma, line 140: no more "I, Alex" comments 🗞️ 💥 
```suggestion
//! The command itself is rarely used anywhere. Other commands that are related to starting a 
//! mirrord instance use the [`extract_library`] function directly.
```

mirrord/cli/src/main.rs
```diff
@@ -54,15 +57,110 @@
...
+//! Can be used directly from the terminal, or from an IDE plugin, but in this case we have a
+//! special handling that allows the omission of a target, since in the IDE, a pop-up is shown
+//! for target selection if it was missing from the [`LayerConfig`].
+//!
+//! ### `mirrord operator <COMMAND>`
 //!
 //! - [`operator_command`]
 //!
-//! > How we pay the bills.
+//! > Setup and management of the mirrord-operator.
```
gememma, line 159:
```suggestion
//! > Setup and management of the mirrord-operator, which forms mirrord's paid offering.
```

mirrord/cli/src/main.rs
```diff
@@ -98,6 +196,44 @@
...
+//! ### `mirrord ls [OPTIONS]`
+//!
+//! - [`list::print_targets`]
+//!
+//! > Like `ls`, but for mirrord kubernetes' targets.
+//!
+//! Fetches the list of supported targets from the cluster, using the `OperatorApi` if the
+//! mirrord-operator is available (and has **not** been disabled in the [`LayerConfig`]), and prints
+//! it back to the user. The output is used by the IDE plugins to show a nice selection box to the
+//! user, when they started mirrrod and have not set a target in their [`LayerConfig`].
```
gememma, line 217:
```suggestion
//! user, when they started mirrod and have not set a target in their [`LayerConfig`].
```

mirrord/cli/src/main.rs
```diff
@@ -98,6 +196,44 @@
...
+//!
+//! > Completions for your shell.
+//!
+//! Uses [`clap`] to generate completions for the mirrord CLI.
+//!
+//! ### `mirrord teams`
+//!
+//! - [`teams::navigate_to_intro`]
+//!
+//! > For users interested in getting mirrord for teams.
```
gememma, line 233:
```suggestion
//! > For users interested in getting mirrord for teams, which is a paid feature.
```

gememma (changes requested):
nice! just the link warnings from rust doc and my few suggestions

mirrord/cli/src/main.rs
meowjesty: Sure, but "rarely used" is not really reality though, nobody uses this.

mirrord/cli/src/main.rs
meowjesty: Fixed, `mirrodee`

gememma (commented):
nice, just a small adjustment on a graph 🤏

CONTRIBUTING.md
```diff
@@ -616,6 +617,100 @@ To test the release workflow:
...
-### Changing the release on MacOS
+## Changing the release on MacOS
+
+If you're making changes to the release and/or CI workflows for MacOS specifically - for example changing how the universal binary is created, you need to ensure that [the script for building the universal binary](/scripts/build_fat_mac.sh) that is run manually when developing has also been updated if necessary.
+
+## Architecture
+
+A high level view of mirrord.
+
+```mermaid
```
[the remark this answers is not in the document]
meowjesty: Weird, I checked it in vscode preview and it looks fine, same for in mermaid.live. Is this something to do with resolution, maybe? I have 4k and 2k.

image

image

gememma (approved):
tres bien 🇫🇷 🫡

PR #3199, opened 2025-03-24
Remove schemars renames

DmitryDodzin:
Add titles for shcemars so the rename will not affect the title. (behind `namespaced-schemars` feature flag)


.github/workflows/ci.yaml
```diff
@@ -536,7 +536,7 @@ jobs:
       - name: Run cli E2E tests
         if: ${{ matrix.container-runtime == 'docker' }}
         run: |
-          cargo test --target=x86_64-unknown-linux-gnu -p tests --no-default-features --features cli -- --test-threads=6 cli
```
Razz4780, line 539: Why no `cli`? :eyes: 
This step is called `Run cli E2E tests`

DmitryDodzin: I found we skip a bunch of test that have ht cli feature flag but are not part of the cli module =] like we skipped the operator setup tests because they are under the operator module but do have the cli feature flag

PR #3200, opened 2025-03-24
Profiles E2E fixed

Razz4780:


meowjesty (approved):
Weird one, but it's a test so :+1: 

Don't you want to assert on the status too?

PR #3204, opened 2025-03-25
Add debounced ping-pong interval reset.

DmitryDodzin:


Razz4780 (commented):
What does this do? :eyes:

DmitryDodzin:
> What does this do? 👀

Reset ping-pong timeout if receiving messages from agent (but debounce the resets so it won't be too much)

Razz4780:
> > What does this do? 👀
> 
> Reset ping-pong timeout if receiving messages from agent (but debounce the resets so it won't be too much)

I'm a bit worried that this will lead to session termination. The operator has a timeout on receiving client messages from the internal proxy. We need to make sure that we keep sending the pings on time, regardless of the messages we receive.
Receiving a message from the operator could reset our receive-pong timeout, but it should not affect the send-ping timout (this would also require handling multiple ping-pongs at the same time). Sth like this (weird pseudo code):
```
sent_pings = 0
select {
    time_to_send_ping -> {
        send_ping()
        sent_pings += 1
    }
    pong_timeout -> error()
    agent_send_pong -> {
        if sent_pings == 0 {
            error()
        } else {
            sent_pings -= 1
        }
    }
    agent_sent_not_pong -> {
        ping_timeout.reset()
    }
}
```

Razz4780 (approved):

PR #3207, opened 2025-03-26
Set upper limit of config.feature.fs.readonly_file_buffer to 1 MB

gememma:


mirrord/config/src/lib.rs
```diff
@@ -605,6 +605,13 @@ impl LayerConfig {
             ));
         }
 
+        if self.feature.fs.readonly_file_buffer > 1_000_000 {
```
Razz4780, line 608: I'd use 1024 * 1024 :v

mirrord/config/src/lib.rs
```diff
@@ -605,6 +605,13 @@ impl LayerConfig {
...
+            return Err(ConfigError::Conflict(
```
Razz4780, line 609: Should be `ConfigError::InvalidValue`

mirrord/config/src/lib.rs
```diff
@@ -605,6 +605,16 @@ impl LayerConfig {
             ));
         }
 
+        if self.feature.fs.readonly_file_buffer > 1024 * 1024 {
+            return Err(ConfigError::InvalidValue {
+                name: "feature.fs.readonly_file_buffer",
+                provided: self.feature.fs.readonly_file_buffer.to_string(),
+                error: "the value of feature.fs.readonly_file_buffer must be 1 Megabyte or less."
+                    .to_string()
```
Razz4780, line 613:
```suggestion
```
`.into()` should work on static str

Razz4780 (approved):

PR #3209, opened 2025-03-26
mirrord action that triggers operator e2e workflow on release PR.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/146

### DIFF
--- a/.github/workflows/operator-e2e.yaml
+++ b/.github/workflows/operator-e2e.yaml
@@ -1,0 +1,73 @@
+name: Operator E2E Test
+
+on:
+  # Run when PR is opened or has new commits
+  pull_request:
+    types: [opened, synchronize]
+  # Allow manual triggering
+  workflow_dispatch:
+
+jobs:
+  trigger-operator-e2e:
+    runs-on: ubuntu-latest
+    # Only run if:
+    # 1. PR is opened and not draft
+    # 2. PR has new code and not draft
+    # 3. Manually triggered
+    # 4. PR title contains a version number (e.g. "Release v3.132.0")
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      ((github.event.action == 'opened' && github.event.pull_request.draft == false) ||
+      (github.event.action == 'synchronize' && github.event.pull_request.draft == false)) &&
+       contains(github.event.pull_request.title, 'Release v')
+
+    steps:
+      - name: Trigger operator E2E workflow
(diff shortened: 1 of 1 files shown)


.github/workflows/operator-e2e.yaml
```diff
@@ -0,0 +1,43 @@
+name: Operator E2E Test
+
+on:
+  pull_request:
+    types: [closed]
```
0x00A5: This got changed from PR approved to PR merged in a later commit, any reason we don't want to run this before merging?

.github/workflows/operator-e2e.yaml
```diff
@@ -0,0 +1,43 @@
...
+  workflow_dispatch:
```
0x00A5, line 4: Maybe we can make this triggered by `workflow_dispatch` only first so we can test it, and then add other triggers like `pull_request` or `pull_request_review`. WDYT?

.github/workflows/operator-e2e.yaml
meowjesty: https://github.com/metalbear-co/private/issues/146

See the security part. I think someone could change the PR after it's been approved and trigger the e2e to run? I'm not sure if this is actually a problem, but I'm being conservative here.

Is it an actual problem or am I just imagining things?

.github/workflows/operator-e2e.yaml
meowjesty: Changed to only `workflow_dispatch`, seems like a good idea to test it out manually first.

.github/workflows/operator-e2e.yaml
0x00A5: I think we can use a combination of settings to prevent that:
* we run actions only for PR that targets protected branches
* for protected branches, we require member's approval
* enable "dismiss stale approval when new commit is pushed", this way the job won't get triggered.
  * [Github doc](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches)
  > Optionally, you can choose to dismiss stale pull request approvals when commits are pushed that affect the diff in the pull request

We can address this after testing with manual dispatch works properly though.

0x00A5 (approved):
Looks good, just need towncrier entry.

.github/workflows/operator-e2e.yaml
meowjesty: Alright.

PR #3221, opened 2025-03-28
Fixed test HTTP servers

Razz4780:
3 test HTTP servers (node, python fastapi, python flask) were implemented to exit after receiving a `DELETE` request. This was introducing flakes - in the test code we were expecting a correct response, but reading the response body was often failing with IO errors.

I don't think there's any reason to implement graceful exit logic in these apps, since the tests do verify what they're meant to verify anyway.


gememma (approved):
seems sensible 🫡 as long as everything gets cleaned up in the end, no need to cause flakies 🥐

Issue #3194, opened 2025-03-20
Add an option to enforce policies on copied targets


Razz4780:

eyalb181:
@Razz4780 Could you add an estimate?

PR #3228, opened 2025-04-01
Adds applies_to_copy_targets to the Policy CRDs, so we can skip policies when using copy-targets.

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/3194


mirrord/operator/src/crd/policy.rs
```diff
@@ -93,6 +93,12 @@ pub struct MirrordPolicySpec {
     ///
     /// Optional.
     pub profile_allowlist: Option<Vec<String>>,
+
+    /// Determines if this policy should be applied to copied targets.
```
Razz4780, line 97:
```suggestion
    /// Whether this policy applies also to sessions using the copy target feature.
    ///
    /// By default, policies don't apply to copy target sessions.
```

PR #3236, opened 2025-04-04
Update contributing guide

0x00A5:
### Tweak go app build script
~~Make go app build script continue the loop if any app fail to build. For example, on macOS, `go-e2e-statfs` doesn't build. By skipping it, the script will build the rest of the test apps~~
Replaced with a note in CONTRIBUTING.md

### Correct a log message
Correct a log message for incoming traffic in `mirrord` CLI. For example, it was `incoming traffic will be be stolen`, and it should be `incoming traffic will be stolen`.

### Misc update in CONTRIBUTING.md
* mention installing uvicorn
* different command on macOS for `minikube service`
* typo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Make sure to take a look at the project's [style guide](STYLE.md).
 - [Getting Started](#getting-started)
 - [Debugging mirrord](#debugging-mirrord)
 - [New Hook Guidelines](#new-hook-guidelines)
-- [Compiling on MacOs](#compliling-on-macos)
+- [Compiling on MacOs](#compiling-on-macos)
 - [Adding new target types](#adding-new-target-types)
 - [Testing the release workflow](#testing-the-release-workflow)
 - [Architecture](#architecture)
@@ -27,7 +27,7 @@ The following guide details the steps to setup a local development environment f
 - [GCC](https://gcc.gnu.org/) - only on Linux, GCC is needed for Go dynamic linking
 - [Rust](https://www.rust-lang.org/)
 - [NodeJS](https://nodejs.org/en/), [ExpressJS](https://expressjs.com/)
-- [Python](https://www.python.org/), [Flask](https://flask.palletsprojects.com/en/2.1.x/), [FastAPI](https://fastapi.tiangolo.com/)
+- [Python](https://www.python.org/), [Flask](https://flask.palletsprojects.com/en/2.1.x/), [FastAPI](https://fastapi.tiangolo.com/), [Uvicorn](https://www.uvicorn.org/)
 - [Go](https://go.dev/)
 - Kubernetes Cluster (local/remote)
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -407,12 +407,20 @@ RUST_LOG=debug target/debug/mirrord exec -i test -l debug -c --target pod/py-ser
...
 ```
 
-Send traffic to the Kubernetes Pod through the service
+Send traffic to the Kubernetes Pod through the service. On Linux, run the following command:
 
 ```bash
 curl $(minikube service py-serv --url)
 ```
 
+On macOS, run the command below and keep the terminal open.
```
meowjesty, line 416: Oh wow, do you really have to keep the terminal open in macos to see the service url? That's so sad :disappointed:

mirrord/cli/src/main.rs
```diff
@@ -461,8 +461,8 @@ fn print_config<P>(
     progress.info(&format!("fs: file operations will default to {}", fs_info));
 
     let incoming_info = match config.feature.network.incoming.mode {
-        IncomingMode::Mirror => "be mirrored",
-        IncomingMode::Steal => "be stolen",
+        IncomingMode::Mirror => "mirrored",
```
meowjesty, line 464: Nice catch, were we really displaying `be be mirrored` lol.

scripts/build_go_apps.sh
```diff
@@ -19,7 +19,7 @@ do
     cd $directory
 
     1>&2 echo "Building test app $directory/$1.go_test_app"
-    go build -o "$1.go_test_app"
+    go build -o "$1.go_test_app" || echo "Cannot build test app $directory/$1.go_test_app"
```
meowjesty, line 22: I'm not sure we want this to not fail, as it's used in CI to build our tests, and there we don't want it to pass by skipping things (unless it's explicitly skipped/ignored, like we have a couple of tests).

Why did it fail to build `go-e2e-statfs` on macos for you? Doesn't the `//go:build linux` prevent it from being built?

meowjesty (changes requested):
Welcome aboard!

I'm not sure we want to not hard fail when some go app fails to build, and as we use the `build_go_apps.sh` in CI, the skipping (if a platform is not supported) should be left to the go package.

scripts/build_go_apps.sh
0x00A5: When I run the script from the repository root, it tries to build both the layer integration test apps and the e2e test apps, and it fails to build `go-e2e-statfs` and quit the do loop early leaving the rest of e2e test apps not built. I just noticed that macos-test CI action works just fine because it `cd` into `mirrord/layer/tests` then run the script from there.
This is an easy tweak that can be done locally, so I removed this change.

meowjesty: Since this is surprising, do you see any place to add a note about this?

CONTRIBUTING.md
0x00A5: > ❗  Because you are using a Docker driver on darwin, the terminal needs to be open to run it
Yeah, need to keep it open or the port forwarding will stop working

scripts/build_go_apps.sh
0x00A5: Indeed, I can add a note under the section `E2E tests`

0x00A5: Added some note [here](https://github.com/metalbear-co/mirrord/pull/3236/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R97-R102)

meowjesty (approved):
:+1:

PR #3239, opened 2025-04-08
fix: warn clients below required version

0x00A5:
Warn clients that are below minimum protocol version required.

### DIFF
--- a/mirrord/agent/src/steal/connections/filtered.rs
+++ b/mirrord/agent/src/steal/connections/filtered.rs
@@ -88,6 +88,18 @@ pub struct UpgradedConnection {
     pub http_server_io: UpgradedServerSide,
 }
 
+/// Result of matching a [`Request`] against [`FilteredStealTask::filters`] and state of
+/// [`FilteredStealTask::subscribed`].
+#[derive(Debug)]
+struct RequestMatchResult {
+    /// The minimum protocol version request for handling the [`Request`].
+    protocol_version_req: semver::VersionReq,
+    /// Ids of matched clients.
+    clients_matched: Vec<u32>,
+    /// Ids of clients that do not meet the minimum protocol version requirement
+    clients_below_version_req: Vec<u32>,
+}
+
 /// Errors that can occur in the [`FilteringService`] when serving a stolen request.
 ///
 /// Has a nice [`fmt::Display`] implementation that shows all sources of [`hyper::Error`]s.
@@ -535,7 +547,7 @@ where
         )
         ret,
     )]
-    fn match_request<B>(&self, request: &mut Request<B>) -> Vec<ClientId> {
+    fn match_request<B>(&self, request: &mut Request<B>) -> RequestMatchResult {
         let protocol_version_req = self
             .original_destination
             .connector()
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -547,22 +559,36 @@ where
...
                 let Some(req) = &protocol_version_req else {
                     return true;
                 };
 
-                entry.value().1.as_ref().is_some_and(|v| req.matches(v))
+                entry.value().1.as_ref().is_some_and(|v| {
+                    if req.matches(v) {
+                        true
+                    } else {
+                        clients_below_version_req.push((*entry.key(), v.clone()));
```
Razz4780, line 576: This has to happen after checking whether the client's filter matches the request. Unmatched requests are of no interest to the clients. We can just switch `.filter(...)` calls order, I think

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -716,6 +749,17 @@ where
             .await?;
         }
 
+        for (client_id, client_version) in clients_below_version_req {
+            tx.send(ConnectionMessageOut::LogMessage {
+                client_id,
+                connection_id: self.connection_id,
+                message: LogMessage::warn(format!(
+                    "Client version {client_version} is below the minimum requirement: {protocol_version_req}",
```
Razz4780, line 757:
```suggestion
                    "An HTTP request was not stolen due to mirrord-protocol version requirement: {protocol_version_req}",
```

mirrord/agent/src/steal/connections/filtered.rs
```diff
@@ -88,6 +88,18 @@ pub struct UpgradedConnection {
     pub http_server_io: UpgradedServerSide,
 }
 
+/// Result of matching a [`Request`] against [`FilteredStealTask::filters`] and state of
+/// [`FilteredStealTask::subscribed`].
+#[derive(Debug)]
+struct RequestMatchResult<'a> {
+    /// The minimum protocol version request for handling the [`Request`].
+    protocol_version_req: Option<&'a semver::VersionReq>,
```
Razz4780, line 96: I'd drop the `Option` and use `VersionReq::STAR` as the fallback value. You wouldn't need to call `toString` and `"none"` in `handle_request`

Razz4780 (changes requested):

mirrord/agent/src/steal/connections/filtered.rs
0x00A5: Do we want to log the client's version? If not we can drop that field from `RequestMatchRestult` and only keep the list of client ids that's below min version requirement.

Razz4780: The client version is already logged locally, so no need for that. We can drop the field ^^

mirrord/agent/src/steal/connections/filtered.rs
0x00A5: Good catch. I switched the order and put version requirement check the last.

mirrord/agent/src/steal/connections/filtered.rs
0x00A5: Removed client version, and replaced the message with what you suggested.

mirrord/agent/src/steal/connections/filtered.rs
0x00A5: Good catch, missed the `Display` impl for `VersionReq`. Dropped `Option` and the `ToString` call 👍

Razz4780 (approved):
Noice :+1:

PR #3249, opened 2025-04-11
Make agent's IP table chain names static and detect leftover chain from previous runs

gememma:


changelog.d/3159.changed.md
```diff
@@ -0,0 +1 @@
+The IP table chain names used by the agent are now stable. The agent will detect if another agent is already running or if a previous cleanup failed.
```
Razz4780, line 1:
```suggestion
The IP table chain names used by the agent are no longer randomized. The agent detects if another agent is already running or if a previous cleanup failed.
```

mirrord/agent/iptables/src/lib.rs
```diff
@@ -79,6 +51,8 @@ pub trait IPTables {
     fn add_rule(&self, chain: &str, rule: &str) -> IPTablesResult<()>;
     fn insert_rule(&self, chain: &str, rule: &str, index: i32) -> IPTablesResult<()>;
     fn list_rules(&self, chain: &str) -> IPTablesResult<Vec<String>>;
+
+    fn list_table(&self, table: &str) -> IPTablesResult<Vec<String>>;
```
Razz4780, line 55:
```suggestion
    fn list_table(&self) -> IPTablesResult<Vec<String>>;
```
This trait kind of assumes that all operations are done on the same table. See that its implementor holds the table name in a field

mirrord/agent/iptables/src/lib.rs
```diff
@@ -238,6 +217,28 @@ where
...
+        let ipt = Arc::new(ipt);
+        tracing::trace!(
+            "list rules in {IPTABLES_TABLE_NAME}: {:?}",
+            ipt.list_table(IPTABLES_TABLE_NAME)?
+        );
+        let rules = ipt.list_table(IPTABLES_TABLE_NAME)?;
+        if rules.iter().any(|rule| {
+            [IPTABLE_PREROUTING, IPTABLE_MESH, IPTABLE_STANDARD]
+                .iter()
+                .any(|&chain| rule.contains(&chain.to_string()))
```
Razz4780, line 235:
```suggestion
                .any(|chain| rule.contains(*chain))
```
Sth like this should work

mirrord/agent/src/cli.rs
```diff
@@ -63,6 +63,10 @@ pub struct Args {
     /// cluster.
     #[arg(long, default_value_t = false, env = envs::IPV6_SUPPORT.name)]
     pub ipv6: bool,
+
+    /// Internal flag to indicate this is the second process spawned by the guard
+    #[arg(long, default_value_t = false, hide = true)]
+    pub second_process: bool,
```
Razz4780, line 69: I wouldn't put this in `Args`. It's already cluttered, weird and awful.

I suggest using env variable, with a const name that is private in `entrypoint.rs`

mirrord/agent/src/entrypoint.rs
```diff
@@ -618,7 +645,29 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
     )
     .await;
+
     match first_connection {
+        any if !clean_iptables => {
+            match any {
+                Ok(Ok((stream, _addr))) => {
+                    // Exit after response to the first connection with DaemonMessage::Close
+                    let connection =
+                        ClientConnection::new(stream, 0, state.tls_connector.clone()).await?;
```
Razz4780, line 655: `?` introduces a bad early return

mirrord/agent/iptables/src/lib.rs
```diff
@@ -238,6 +217,28 @@ where
         Ok(Self { redirect })
     }
 
+    /// detect if the IP tables on the target are dirty: this is required since we use static
+    /// IP tables chain names, and they are in danger of conflicting with each-other
+    ///
+    /// returns `true` if iptables are clean and the agent can proceed, other returns `false`
+    #[tracing::instrument(level = Level::TRACE, skip(ipt) ret, err)]
+    pub async fn ensure_iptables_clean(ipt: IPT) -> IPTablesResult<bool> {
```
Razz4780, line 225:
```suggestion
    pub async fn list_mirrord_rules(ipt: IPT) -> IPTablesResult<Vec<String>> {
```
I'd make it return the list of rules that mention mirrord chains. Then, in the caller function, you can do a nice error log:
```rust
let leftover_rules = SafeIpTables::list_mirrord_rules(ipt).await?;
if existing_mirrord_rules.is_empty().not() {
    tracing::error!(
        leftover_rules = ?leftover_rules,
        "Detected dirty iptables. Either some other mirrord agent is running \
        or the previous agent failed to clean up before exit. \
        If no other mirrord agent is targeting this pod, please delete the pod."
    );
    let _ = notify_client_about_dirty_iptables(args).await; // Try to create the tcp listener, accept one connection, and send DaemonMessage::Close.
    return Err(AgentError::DirtyIpTables);
}
```

mirrord/agent/src/entrypoint.rs
```diff
@@ -564,6 +573,24 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
+            } else {
+                new_iptables()
+            };
+            SafeIpTables::ensure_iptables_clean(IPTablesWrapper::from(iptables))
+                .await
+                .map_err(|error| AgentError::IPTablesSetupError(error.into()))
+        })
+        .await
+        .map_err(|error| AgentError::IPTablesSetupError(error.into()))?
+        .map_err(|error| AgentError::IPTablesSetupError(error.into()))?;
```
Razz4780, line 592: If we detected leftover rules, I'd go straight to creating the listener and notifying the first client (skipping background tasks setup). Following the suggestion for `ensure_iptables_clean`

mirrord/agent/src/entrypoint.rs
```diff
@@ -775,7 +817,14 @@ async fn start_iptable_guard(args: Args) -> AgentResult<()> {
             Ok(())
         }
 
-        result = run_child_agent() => result,
+        result = run_child_agent() => match result {
+            Err(AgentError::AgentFailed(status)) if status.code() == Some(99) => {
```
Razz4780, line 821: I'd use a const for this special code

```rust
/// [`ExitCode`](std::process::ExitCode) returned from the child agent process
/// when dirty iptables are detected.
const IPTABLES_DIRTY_EXIT_CODE: u8 = 99;
```

mirrord/agent/src/error.rs
```diff
@@ -54,6 +54,9 @@ pub(crate) enum AgentError {
     #[error("IP tables setup failed: {0}")]
     IPTablesSetupError(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
 
+    #[error("IP tables unclean")]
```
Razz4780, line 57:
```suggestion
    #[error("IP tables dirty")]
```

sample/kubernetes/app_bad_cleanup.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: apps/v1
```
Razz4780, line 1: :wastebasket:

Razz4780 (changes requested):

sample/kubernetes/app_bad_cleanup.yaml
gememma: oop 😳

mirrord/agent/src/entrypoint.rs
gememma: im eliminating this entire terrible block of `match` and fail earlier when i list leftover_rules as you suggested 👀

mirrord/agent/src/main.rs
```diff
@@ -28,6 +32,10 @@ mod vpn;
 
 #[cfg(target_os = "linux")]
 #[tokio::main(flavor = "current_thread")]
-async fn main() -> crate::error::AgentResult<()> {
-    crate::entrypoint::main().await
+async fn main() -> ExitCode {
+    match crate::entrypoint::main().await {
+        Ok(_) => ExitCode::SUCCESS,
+        Err(AgentError::IPTablesDirty) => ExitCode::from(99),
```
Razz4780, line 38: Use the consttt

mirrord/cli/src/execution.rs
```diff
@@ -462,6 +462,7 @@ impl MirrordExecution {
 
     /// Construct filter and retrieve remote environment from the connected agent using
     /// `MirrordExecution::get_remote_env`.
+    #[tracing::instrument(level = Level::ERROR, skip_all)]
```
Razz4780, line 465: :wastebasket:

mirrord/agent/iptables/src/lib.rs
```diff
@@ -535,4 +531,52 @@ mod tests {
 
         assert!(ipt.cleanup().await.is_ok());
     }
+
+    /// Check that dirty ip tables fail the ['SafeIpTables::ensure_iptables_clean'] check
+    /// If there are any chains in the IP table with names used by the agent, the check should fail.
+    /// A fresh IP table, or one with only non-agent names, should pass the check
+    #[tokio::test]
+    async fn fail_on_dirty() {
```
Razz4780, line 539: This could be split into two tests

mirrord/agent/src/entrypoint.rs
```diff
@@ -62,6 +59,11 @@ mod setup;
 /// task.
 const CHANNEL_SIZE: usize = 1024;
 
+/// [`ExitCode`](std::process::ExitCode) returned from the child agent process
+/// when dirty iptables are detected.
+const IPTABLES_DIRTY_EXIT_CODE: i32 = 99;
+const SECOND_PROCESS_ENV: &str = "MIRRORD_AGENT_SECOND_PROCESS";
```
Razz4780, line 65: Doccc. Also, I'd use `CHILD_PROCESS` instead of `SECOND_PROCESS`

mirrord/agent/src/entrypoint.rs
```diff
@@ -62,6 +59,11 @@ mod setup;
 /// task.
 const CHANNEL_SIZE: usize = 1024;
 
+/// [`ExitCode`](std::process::ExitCode) returned from the child agent process
+/// when dirty iptables are detected.
+const IPTABLES_DIRTY_EXIT_CODE: i32 = 99;
```
Razz4780, line 64: `ExitCode` can be created from `u8`, so this should probably be `u8`

mirrord/agent/src/entrypoint.rs
```diff
@@ -564,6 +618,39 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
+        .network_runtime
+        .spawn(async move {
+            let ipt = if args.ipv6 {
+                new_ip6tables()
+            } else {
+                new_iptables()
+            };
+            SafeIpTables::list_mirrord_rules(IPTablesWrapper::from(ipt))
+                .await
+                .map_err(|error| AgentError::IPTablesSetupError(error.into()))
```
Razz4780, line 632: This line can probably be removed, since you do `.map_err(|error| AgentError::IPTablesSetupError(error.into()))?` twice below

mirrord/agent/src/entrypoint.rs
```diff
@@ -857,12 +946,9 @@ pub async fn main() -> AgentResult<()> {
     );
 
     let args = cli::parse_args();
+    let second_process = matches!(std::env::var(SECOND_PROCESS_ENV), Ok(value) if value == "true");
```
Razz4780, line 949:
```suggestion
    let second_process = std::env::var(SECOND_PROCESS_ENV).is_ok();
```
Can be simpler, I think

tests/src/utils/watch.rs
```diff
@@ -101,10 +101,6 @@ where
 /// Waits until the given [`Service`] has at least `min` ready [`Pod`].
 ///
 /// Returns ready pods.
-///
-/// A [`Pod`] is considered to be ready when:
-/// 1. It's in the `Running` phase
-/// 2. All of its containers are ready
```
Razz4780, line 107: Why delete tho :<

tests/src/utils/watch.rs
```diff
@@ -124,27 +120,52 @@ pub async fn wait_until_pods_ready(service: &Service, min: usize, client: Client
...
 
-        containers.iter().all(|status| status.ready)
-    }
+/// Waits until the given [`Pod`] is ready.
+#[cfg(test)]
+#[cfg(all(not(feature = "operator"), feature = "job"))]
+pub async fn wait_until_pod_ready(pod_name: &str, namespace: &str, client: Client) {
+    let api = Api::<Pod>::namespaced(client, namespace);
+
+    let config = Config::default();
```
Razz4780, line 142: Should use `field_selector`: `metadata.name={pod_name}`

tests/src/dirty_iptables.rs
```diff
@@ -0,0 +1,150 @@
...
+    .await;
+
+    // check for expected error message in output
+    process.wait_assert_fail().await;
+    process
+        .assert_stderr_contains("Detected dirty iptables.")
+        .await;
+
+    drop(namespace_guard);
+    drop(pod_guard);
```
Razz4780, line 149: Unnecessary drops

tests/src/dirty_iptables.rs
```diff
@@ -0,0 +1,150 @@
...
+        }))
+        .unwrap(),
+        delete_after_fail,
+    )
+    .await
+    .expect(format!("Should be able to create namespace {namespace}").as_str());
+
+    // Create a single pod as the target and wrap it in ResourceGuard
+    let pod_api: Api<Pod> = Api::namespaced(kube_client.clone(), &namespace);
+    let pod_guard = ResourceGuard::create::<Pod>(
```
Razz4780, line 68: The pod does not need the guard, since the whole namespace is guarded

Razz4780 (changes requested):
Only nits, GJ ^^

tests/src/utils/watch.rs
gememma: i just moved it to `fn pod_is_ready` which is called by this function 👀

Razz4780: Yeah, but now the doc is not complete. The caller has to look into the code to understand what's going on

mirrord/agent/src/entrypoint.rs
gememma: i think this just moves the conversion elsewhere 👀 
when i check `if status.code() == Some(IPTABLES_DIRTY_EXIT_CODE)`, i would have to cast to i32 anyway

Screenshot 2025-04-16 at 09 56 55

Razz4780: Though `i32` has `From<u8>`, not the other way around

gememma: ah, right 🧠 good point

mirrord/agent/iptables/src/lib.rs
```diff
@@ -238,6 +217,23 @@ where
...
+        let rules = ipt.list_table()?;
+
+        Ok(rules
+            .iter()
+            .filter(|rule| {
+                [IPTABLE_PREROUTING, IPTABLE_MESH, IPTABLE_STANDARD]
+                    .iter()
+                    .any(|chain| rule.contains(*chain))
+            })
+            .map(|s| s.as_str().to_string())
```
Razz4780, line 233:
```suggestion
            .into_iter()
            .filter(|rule| {
                [IPTABLE_PREROUTING, IPTABLE_MESH, IPTABLE_STANDARD]
                    .iter()
                    .any(|chain| rule.contains(*chain))
            })
```

Razz4780 (approved):
One unnecessary cloning

mirrord/agent/iptables/src/lib.rs
gememma: this line prevents `error[E0277]: a value of type `Vec<String>` cannot be built from an iterator over elements of type '&String\'` 👀

PR #3253, opened 2025-04-15
Refactor e2e huge `utils.rs` into mods.

meowjesty:
- Issue: n/a

The `utils.rs` file was huge (and getting huger) which made it not so great to work with copilot. So here's a refactor that just moves things around, no functionality has been changed, but a couple of things have been renamed to avoid conflicts.

The initial commits each contain a piece of `utils.rs` that turned into a module. So you can review them to check if all the code has been moved properly. Commits that come afterwards are just fixes to imports and renaming to fix conflicts.


tests/src/utils/run_command.rs
gememma, line 1: picky, but i kind of dislike the order these functions go in - i think `run_mirrord` should be last to preserve the order of X calls Y, Y calls Z, etc., or rather like this:

Screenshot 2025-04-17 at 13 39 56

gememma (approved):
very very nice! one nitlet, and possibly a vague feeling that it would be a good opportunity to add docs to un-documented structs, like `TestProcess` ? i leave it up to you 👀

tests/src/utils/run_command.rs
meowjesty: Changed the order.

PR #3259, opened 2025-04-17
Verify queue deletion after clients exit in SQS E2E test

t4lz:
The SQS operator creates SQS queues for SQS splitting. When the clients exit those queues should be deleted. Now this is verified in the E2E test.

### DIFF
--- a/changelog.d/+verify-queue-deletion.internal.md
+++ b/changelog.d/+verify-queue-deletion.internal.md
@@ -1,0 +1,1 @@
+Verify temp queue deletion in SQS E2E test.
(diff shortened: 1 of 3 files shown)


Razz4780 (approved):
Looks good, but I'd add some comments on the fact that it will break when we have more SQS tests

t4lz:
> Looks good, but I'd add some comments on the fact that it will break when we have more SQS tests

Why would it break?

Razz4780:
> > > > Looks good, but I'd add some comments on the fact that it will break when we have more SQS tests
> > > 
> > > 
> > > Why would it break?
> > 
> > 
> > Because you expect that tmp queues originate only from this test
> 
> I do not. I filter only the temp queues from this test and ignore any other ones
> 
> https://github.com/metalbear-co/mirrord/blob/e9d5a7aa4692112c60ac7eb5984a72ebcc23a7f6/tests/src/utils/sqs_resources.rs#L104

Ah 😅 
Sorry, all good then

PR #3269, opened 2025-04-19
Support skipping SIP patch in layer config

0x00A5:


mirrord/sip/src/lib.rs
```diff
@@ -487,7 +487,17 @@ mod main {
     }
 
     /// SIP check for binaries.
-    fn is_binary_sip(path: &Path, data: &[u8], patch_binaries: &[String]) -> Result<bool> {
+    fn is_binary_sip(
+        path: &Path,
+        data: &[u8],
+        patch_binaries: &[String],
```
DmitryDodzin, line 493: I would concider making a custom struct for patch and skip because it's easy to accidentaly pass them the wrong way around, if they were of different types it would be ok but they are of the same type

DmitryDodzin (changes requested):

DmitryDodzin (changes requested):
Overall looks good, I would just do a custom struct for the 2 string arrays so they won't be swapped by accident

mirrord/sip/src/lib.rs
0x00A5: Thanks for taking a look!
Something like?
```rust
struct SipOptions<'a>  {
    patch: &'a [String],
    skip: &'a [String],
}
```
Let me know if I misunderstand your suggestion.

DmitryDodzin: Yeah exactly, just to give names to the arrays

0x00A5: Added an option type which wraps `patch` and `skip`, thanks!

DmitryDodzin (approved):

PR #3270, opened 2025-04-20
Use ss for killing existing connections to allow stealing to begin

aviramha:


mirrord/agent/iptables/src/flush_connections.rs
```diff
@@ -72,6 +72,18 @@ where
             warn!("`conntrack` output is {conntrack:#?}");
         }
 
+        // ss is available from kernel 5.1 and should work way better than conntrack
+        let formatted_port = format!(":{}", redirected_port);
+
+        let ss = Command::new("ss")
+            .args(["-K", "sport", "=", &formatted_port])
```
Razz4780, line 79: Shouldn't it be `dport`?

Razz4780: `sport` sounds like source port

aviramha: yes, it's source port - when you're running on the server the sport is the port you're listening on.

aviramha: if 1.1.1.1 connects to 1.1.1.2 on port 80, from 1.1.1.2 perspective port 80 is the sport of the connection.

Razz4780: Oh, kk

Razz4780 (approved):

PR #3275, opened 2025-04-22
Add config value to override extproxy listen addr

DmitryDodzin:


mirrord/config/src/external_proxy.rs
```diff
@@ -97,4 +97,11 @@ pub struct ExternalProxyConfig {
     /// Defaults to true.
     #[config(default = true)]
     pub json_log: bool,
+
+    /// ### external_proxy.host_ip {#external_proxy-host_ip}
+    ///
+    /// Specify a custom host ip addr to listen on.
+    ///
+    /// Will try and resolve local addr if not specified.
```
Razz4780, line 105:
```suggestion
    /// This address must be accessible from within the container.
    /// If not specified, mirrord will try and resolve a local address to use.
```

Razz4780 (approved):

PR #3279, opened 2025-04-23
Passthrough traffic mirroring

Razz4780:
Oh well


t4lz:
One general thing: does `passthrough_mirroring` make sense as a user facing name?
Where does this name come from? Is that a well known term for what we're doing here?
If not, we should try to name it in a way that helps users understand what it does, especially what effect it has on them (which I am still not sure about). Like `reset_mirrored_connections` or `iptables_mirroring` (though this one is also a bit TMI).

mirrord/agent/src/entrypoint.rs
```diff
@@ -368,31 +402,12 @@ impl ClientConnectionHandler {
                 // poll the sniffer API only when it's available
```
t4lz, line 402:
```suggestion
                // poll the mirror API only when it's available
```

mirrord/agent/src/entrypoint.rs
```diff
@@ -425,6 +436,10 @@ impl ClientConnectionHandler {
     /// Sends a [`DaemonMessage`] response to the connected client (`mirrord-layer`).
     #[tracing::instrument(level = "trace", skip(self))]
     async fn respond(&mut self, response: DaemonMessage) -> AgentResult<()> {
+        if self.ready_for_logs.not() && matches!(response, DaemonMessage::LogMessage(..)) {
+            return Ok(());
+        }
```
t4lz, line 441: We could start a log buffer. Just an idea for the future.

mirrord/agent/src/entrypoint.rs
```diff
@@ -479,7 +494,7 @@ impl ClientConnectionHandler {
             }
             ClientMessage::Ping => self.respond(DaemonMessage::Pong).await?,
             ClientMessage::Tcp(message) => {
-                if let Some(sniffer_api) = &mut self.tcp_sniffer_api {
+                if let Some(sniffer_api) = &mut self.tcp_mirror_api {
                     sniffer_api.handle_client_message(message).await?
```
t4lz, line 498:
```suggestion
                if let Some(mirror_api) = &mut self.tcp_mirror_api {
                    mirror_api.handle_client_message(message).await?
```

mirrord/agent/src/entrypoint.rs
```diff
@@ -490,8 +505,13 @@ impl ClientConnectionHandler {
...
             ClientMessage::TcpSteal(message) => {
-                if let Some(tcp_stealer_api) = self.tcp_stealer_api.as_mut() {
-                    tcp_stealer_api.handle_client_message(message).await?
+                if let Some(tcp_stealer_api) = self.tcp_steal_api.as_mut() {
+                    if let Err(error) = tcp_stealer_api.handle_client_message(message).await {
+                        self.respond(DaemonMessage::Close(format!(
+                            "invalid HTTP filter: {error}"
+                        )))
+                        .await?;
+                    }
```
t4lz, line 514: "invalid HTTP filter" looks pretty arbitrary in here. Even if it's the only possible reason for an error (is it?) wouldn't it be better to determine that with a `match`?

mirrord/agent/src/incoming/connection/http.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// Handle to the [`tokio::runtime`] in which this struct was created.
+    ///
+    /// Used to spawn the connection task.
+    ///
+    /// Thanks to this handle, this struct can be freely moved across runtimes.
+    runtime_handle: Handle,
+}
+
+impl RedirectedHttp {
+    /// Should be called in the target's Linux network namespace.
```
t4lz, line 45: maybe also explain why

mirrord/agent/src/entrypoint.rs
Razz4780: The error type of `handle_client_message` is `fancy_regex::Error`. Not sure what you mean by `match`. Do you want sth like this?
```rust
                    if let Err(error) = tcp_stealer_api.handle_client_message(message).await {
                        let error: fancy_regex::Error = error;
                        self.respond(DaemonMessage::Close(format!(
                            "invalid HTTP filter: {error}"
                        )))
                        .await?;
                    }
```
This would work as a reminder to update the error message, when someone changes the error type.
Or should the method return just `Result<_, String>` (message for `DaemonMessage::Close`) :thinking:

mirrord/agent/src/entrypoint.rs
```diff
@@ -219,17 +221,19 @@ struct BackgroundTasks {
...
 
 struct ClientConnectionHandler {
     id: ClientId,
+    protocol_version: SharedProtocolVersion,
     /// Handles mirrord's file operations, see [`FileManager`].
     file_manager: FileManager,
     connection: ClientConnection,
     /// [`None`] when targetless.
-    tcp_sniffer_api: Option<TcpSnifferApi>,
+    tcp_mirror_api: Option<Box<dyn TcpMirrorApi>>,
```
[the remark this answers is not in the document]
Razz4780: I would have needed to change the way `ClientConnectionHandler` is created, and it was a bit messy :V
Are you against dynamic dispatch? We can do an enum as well

Razz4780:
> One general thing: does `passthrough_mirroring` make sense as a user facing name? Where does this name come from? Is that a well known term for what we're doing here? If not, we should try to name it in a way that helps users understand what it does, especially what effect it has on them (which I am still not sure about). Like `reset_mirrored_connections` or `iptables_mirroring` (though this one is also a bit TMI).

I didn't think about it much tbh. I believe I came up with this name when presenting the idea internally. I agree that `iptables_mirroring` makes more sense. I wouldn't worry about it being too specific, as this config option is temporary. Eventually we'll drop packet sniffing and make iptables the only implementation.

Razz4780:
Ok, there's something wrong here, some kind of a race.
I'm splitting this PR further, cause it's too big to find it ;_;

PR #3287, opened 2025-04-24
Add support for argo workflow targets

DmitryDodzin:


mirrord/kube/src/resolved.rs
```diff
@@ -522,6 +566,21 @@ impl ResolvedTarget<false> {
...
+                    template,
+                }))
+            }
+
+            ResolvedTarget::Workflow(ResolvedResource {
+                resource,
+                container,
+                template,
+            }) => {
+                // TODO: this
```
t4lz, line 578: ?

mirrord/cli/src/verify_config.rs
```diff
@@ -157,7 +165,7 @@ impl TargetType {
             Self::Pod => !(config.copy_target.enabled && config.copy_target.scale_down),
             Self::Job | Self::CronJob => config.copy_target.enabled,
             Self::Service => !config.copy_target.enabled,
-            Self::Deployment | Self::StatefulSet | Self::ReplicaSet => true,
+            Self::Deployment | Self::StatefulSet | Self::ReplicaSet | Self::Workflow => true,
```
t4lz, line 168: How come workflow targets don't require a copy_target? Since they actually target a specific node of the workflow graph that ends, like a job?

mirrord/config/src/target.rs
t4lz, line 1: Missing user-facing docs for workflow targets.

Also, we might want to clarify the `container/...` part of the target in those docs, because people might assume that part means - choose a template with that container name, but it doesn't, with the current code what it means is - once you found the pod that was created for the target template, choose this container of that pod.

t4lz (commented):
You're changing it to enabling specifying a specific "node" in the workflow graph, right?

Don't forget to run medschool.

mirrord/config/src/target.rs
```diff
@@ -391,6 +399,13 @@ macro_rules! impl_target_display {
                 self.container.as_ref()
             }
         }
+    };
+}
+
+/// Implements the [`TargetDisplay`] and [`fmt::Display`] traits for a target type.
+macro_rules! impl_target_display {
```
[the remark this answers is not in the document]
DmitryDodzin: I was thinking and didn't like `impl_target_display_and_display` but `impl_display_and_target_display` is better

mirrord/cli/src/verify_config.rs
DmitryDodzin: I'm a bit hesitant in requiring it being copy-target just yet because of the amount of customization included in workflows, Right now we can add support for "catching" a running workflow

PR #3290, opened 2025-04-24
Add configuration to sperate host_ip in extproxy and intproxy connection ip

DmitryDodzin:


Razz4780 (changes requested):
I think you're changing the wrong thing :eyes: 
Intproxy shoud listen on localhost. The user container and the intproxy sidecar share the same network, so the layer can talk to intproxy.
Don't you want to change the address used by the intproxy when connecting to the extproxy?

mirrord/config/src/container.rs
```diff
@@ -51,4 +53,21 @@ pub struct ContainerConfig {
...
+    /// ```json5
+    /// {
+    ///   "container": {
+    ///     "override_host_ip": "172.17.0.1" // usual resolution of value from `host.docker.internal`
+    ///   }
+    /// }
+    /// ```
+    ///
+    /// This should be useful if your host machine is exposed via different ip addr than one bound
+    /// as host
```
Razz4780, line 71:
```suggestion
    /// ### container.override_host_ip {#container-override_host_ip}
    ///
    /// Allows to override the IP address for the internal proxy to use
    /// when connecting to the host machine from within the container.
    ///
    /// ```json5
    /// {
    ///   "container": {
    ///     "override_host_ip": "172.17.0.1" // usual resolution of value from `host.docker.internal`
    ///   }
    /// }
    /// ```
    ///
    /// This should be useful if your host machine is exposed with a different IP address than the one
    /// bound as host.
```

Razz4780 (commented):
Small doc request

Razz4780 (approved):

Issue #3244, opened 2025-04-09
Automatically add health probe ports to http_filter ports (if a filter is set)


t4lz:
The idea is that it's a hint for us for what's the application's HTTP port, because it will usually be the same, so we can use that to automatically run the HTTP filter on the correct ports.

A warning was added in https://github.com/metalbear-co/mirrord/pull/3148, so make sure we don't show that warning when there is no need.

Watch out for `feature.network.incoming.ports` and `feature.network.incoming.ignore_ports` and make sure we don't create conflicts.

We should not add the port if the user has set a `feature.network.incoming.http_filter.ports` themselves - otherwise users don't have any way of telling mirrord not to filter that port.

PR #3291, opened 2025-04-24
Automatically insert probes ports into http filter ports.

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/3244

Introduces a new function in `IncomingConfig`: `add_probe_ports_to_http_filter_ports` which automatically adds the health probes ports to the `HttpFilter::ports`, if an HTTP filter is set and some other conditions are in place (avoiding conflicts with other _port_ related configs).


mirrord/cli/src/port_forward.rs
```diff
@@ -891,14 +891,22 @@ enum IncomingMode {
...
-        let http_filter_config = &config.http_filter;
+        let ports = {
+            config
+                .http_filter
+                .ports
+                .get_or_insert_default()
+                .iter()
+                .copied()
+                .collect()
+        };
```
Razz4780, line 907: This does not need `&mut` - `config.http_filter.ports.clone().unwrap_or_default()`

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -154,9 +154,13 @@ pub struct HttpFilterConfig {
     /// Other ports will *not* be stolen, unless listed in
     /// [`feature.network.incoming.ports`](#feature-network-incoming-ports).
     ///
+    /// We check the pod's health probe ports and automatically add them here, as they're
+    /// usually the same ports your app might be listening on. If your app ports and the
+    /// health probe ports don't match, then setting this option will override this behavior.
+    ///
     /// Set to [80, 8080] by default.
     #[config(env = "MIRRORD_HTTP_FILTER_PORTS", default)]
```
Razz4780, line 162:
```suggestion
    #[config(env = "MIRRORD_HTTP_FILTER_PORTS")]
```
I think :eyes:

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -224,10 +234,7 @@ impl MirrordToggleableConfig for HttpFilterFileConfig {
         let all_of = None;
         let any_of = None;
 
-        let ports = FromEnv::new("MIRRORD_HTTP_FILTER_PORTS")
-            .source_value(context)
-            .transpose()?
-            .unwrap_or_default();
+        let ports = None;
```
Razz4780, line 237: Lost reading from env here

mirrord/layer/src/setup.rs
```diff
@@ -216,14 +220,23 @@ pub enum IncomingMode {
...
+                .ports
+                .get_or_insert_default()
+                .clone()
+                .iter()
+                .copied()
+                .collect()
+        };
 
-        let ports = { http_filter_config.ports.iter().copied().collect() };
+        let http_filter_config = &config.http_filter;
```
Razz4780, line 239: This does not need `&mut` - `config.http_filter.ports.clone().unwrap_or_default()`

Razz4780 (changes requested):
We need to issue an info notification as well

mirrord/layer/src/setup.rs
meowjesty: If we don't insert the default config here, won't the config be different than what's being set? I mean in the case:

```rs
filter.ports = None
let ports = filter.ports.unwrap_or_default();
// now `StealHttpSettings` has `ports = [80, 8080]`, but mirrord config thinks it's `None`
```

Won't this cause trouble in the future?

meowjesty: It's the same for `port_forward`, that's what I was afraid of when I changed it to `mut` here and there.

meowjesty:
Added a generic info log.

mirrord/layer/src/setup.rs
Razz4780: Well, I think the problem lies more in the fact that `None` actually means `Some([80, 8080])`. But I'm fine with it, config code is dirty :shrug: 
I just found it weird that a method reading config also injects a value into it.

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -172,7 +176,13 @@ impl HttpFilterConfig {
...
     pub fn get_filtered_ports(&self) -> Option<&[u16]> {
-        self.is_filter_set().then(|| &*self.ports.0)
+        if let Some(ports) = self.ports.as_ref()
+            && self.is_filter_set()
+        {
+            Some(&*ports.0)
+        } else {
+            None
+        }
     }
```
Razz4780, line 186: Ok, I think this will be confusing in the future. This returns `None` if the user did not supply the port list, but actually the port list is `[80, 8080]`. It might be too easy to misuse. Maybe we should have two methods: `get_filtered_ports` and `ports_set_by_user`? `get_filtered_ports` would always return the correct list

mirrord/layer/src/setup.rs
meowjesty: Yeah, this is not great. But to not have the config being `mut` would be quite a refactor I think (and it didn't feel worth it when the actual reason is so small).

mirrord/cli/src/port_forward.rs
meowjesty: I reverted it back to `mut`.

mirrord/progress/src/messages.rs
```diff
@@ -24,3 +24,7 @@ pub const HTTP_FILTER_WARNING: (&str, &str) = (
 pub const EXEC_CONTAINER_BINARY: &str = "`mirrord exec <docker|podman|nerdctl> ...` detected! \
     If you try to run a container with mirrord, please use \
     `mirrord container [options] -- <docker|podman|nerdctl> ...` instead.";
+
+/// Info when we modify the user's `network.incoming.http_filter.ports` config.
+pub const MODIFIED_HTTP_FILTER_PORTS: &str =
+    "`network.incoming.http_filter.ports` has been set to use ports";
```
Razz4780: You use it in just one place and it's weird without the rest of the message. I don't think it's necessary for it to be a separate const

Razz4780 (approved):
Oke, let it be

mirrord/progress/src/messages.rs
meowjesty: Removed the const.

PR #3297, opened 2025-04-25
Fix layer test on macOS aarch64e

### LINKED ISSUE #3267: Investigate layer integration test failures on macOS
On macOS (aarch64e), layer integration tests fail when using the universal binary `libmirrord_layer.dylib`. To reproduce, run the command below:
```sh
./scripts/build_fat_mac.sh

MIRRORD_TEST_USE_EXISTING_LIB=../../target/universal-apple-darwin/debug/libmirrord_layer.dylib cargo test -p mirrord-layer --test spawn
```

However, when pointing `MIRRORD_TEST_USE_EXISTING_LIB` to the X86_64-apple-darwin target, the test passes.

```sh
cargo build -p mirrord-layer --target=x86_64-apple-darwin
cargo build -p mirrord-layer
MIRRORD_TEST_USE_EXISTING_LIB=../../target/X86_64-apple-darwin/debug/libmirrord_layer.dylib cargo test -p mirrord-layer --test spawn
```

0x00A5:
Fix #3267


mirrord/layer/tests/common/mod.rs
```diff
@@ -1383,20 +1383,44 @@ pub fn get_env(
...
+            "MIRRORD_MACOS_ARM64_LIBRARY".to_string(),
+            arm64_dylib_path().to_str().unwrap().to_string(),
         ),
     ]
     .into_iter()
     .chain(extra_vars_owned)
     .collect()
 }
+
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
```
t4lz, line 1411: why not make it a `once` fixture like the existing `dylib_path`?
```suggestion
#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
#[fixture]
#[once]
```

CONTRIBUTING.md
```diff
@@ -201,6 +201,13 @@ And then in order to use that dylib in the tests, run the tests like this:
 MIRRORD_TEST_USE_EXISTING_LIB=../../target/universal-apple-darwin/debug/libmirrord_layer.dylib cargo test -p mirrord-layer
 ```
 
+On Apple Silicon, set `MIRRORD_MACOS_ARM64_LIBRARY` additionally to use the arm64 layer lib as well:
+```bash
+MIRRORD_TEST_USE_EXISTING_LIB=../../target/universal-apple-darwin/debug/libmirrord_layer.dylib \
+MIRRORD_MACOS_ARM64_LIBRARY=../../target/aarch64-apple-darwin/debug/libmirrord_layer.dylib \
+cargo test -p mirrord-layer
+```
```
t4lz, line 209: What do you think about deriving one var from the other if only one of them was specified?

mirrord/layer/tests/common/mod.rs
```diff
@@ -1383,20 +1383,44 @@ pub fn get_env(
...
             "LD_PRELOAD".to_string(),
             dylib_path.to_str().unwrap().to_string(),
         ),
+        #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
         (
-            LayerConfig::RESOLVED_CONFIG_ENV.to_string(),
-            config.encode().unwrap(),
+            "MIRRORD_MACOS_ARM64_LIBRARY".to_string(),
+            arm64_dylib_path().to_str().unwrap().to_string(),
         ),
```
t4lz, line 1404: I think we should document somewhere why we need this to run the tests (especially as it is not needed for CI until we get arm64 darwin runners, so could be easily removed by mistake). I think here could be a place to document it. The contributing guide might also make sense, up to you which one to choose.

CONTRIBUTING.md
0x00A5: That would be a fairly reliable fallback, giving the `universal-apple-darwin` directory exists, the build fat mac script has most likely been run at this point. We can try deriving it and use it if the path exists, otherwise build the aarch64 target.

mirrord/layer/tests/common/mod.rs
0x00A5: I considered making it a fixture but didn't do so because it will require adding an new parameter for each tests (behind the same compilation condition here `macos` + `aarch64`). But if it's a cleaner option, I don't mind making it a fixture, since it is very similar to the `dylib_path` fixture.

mirrord/layer/tests/common/mod.rs
0x00A5: Cool, I will add some note here

CONTRIBUTING.md
0x00A5: Added deriving arm64 lib path from universal lib path as a fallback

mirrord/layer/tests/common/mod.rs
t4lz: Ok, it's not that important. A `OnceLock` is fine, it also wouldn't be the end of the world if we didn't have a "once" mechanism, I mostly just wanted it to be consistent with the other dylib path we have.

changelog.d/3267.fixed.md
t4lz, line 1: should be `internal`, not `fixed`. Everything users shouldn't care about (like how we run our tests) should be under internal.

t4lz (approved):
Please do mention in the contribution guide the option to directly specify the arm64 lib.
It was nice to have it documented there.

Just need to change the changelog entry file name and document in the contribution guide, so approving already.

changelog.d/3267.fixed.md
0x00A5: Renamed, thanks!

PR #3302, opened 2025-04-28
mirrord-protocol update (part of passthrough mirroring prep)

Razz4780:
1. Internal proxy handles HTTP requests from the agent, even if the port subscriptions is unfiltered or mirror
2. New type of message for TLS connections stolen in whole

We need both for passthrough mirroring, as the agent will need to try and run an HTTP server for **every** redirected connection.

This PR should not change anything in the current functionality (new message variants are not being actually produced yet). There a small bug fix included though, highlighted in a comment


mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -257,57 +272,66 @@ impl IncomingProxy {
...
             );
 
             let message = if is_steal {
-                ClientMessage::Tcp(LayerTcp::ConnectionUnsubscribe(connection_id))
-            } else {
                 ClientMessage::TcpSteal(LayerTcpSteal::ConnectionUnsubscribe(connection_id))
+            } else {
+                ClientMessage::Tcp(LayerTcp::ConnectionUnsubscribe(connection_id))
             };
             message_bus.send(message).await;
```
Razz4780, line 336: Small fix here, we were producing bad `ClientMessage` variant

mirrord/intproxy/src/proxies/incoming/tcp_proxy.rs
```diff
@@ -25,11 +27,64 @@ pub enum LocalTcpConnection {
...
+        transport: IncomingTrafficTransportType,
+        tls_setup: Option<Arc<LocalTlsSetup>>,
     },
     /// Upgraded HTTP connection from a previously stolen HTTP request.
     AfterUpgrade(OnUpgrade),
 }
 
+impl LocalTcpConnection {
+    /// Makes the connection, returning the IO stream and data ready to be sent to the agent.
+    async fn make(self) -> Result<(MaybeTls, Vec<u8>), InProxyTaskError> {
```
meowjesty, line 39: nit: isn't this more like `connect`, rather than `make`?

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -49,6 +50,31 @@ mod port_subscription_ext;
 mod subscriptions;
 mod tasks;
 mod tcp_proxy;
+mod tls;
+
+#[derive(Default)]
+struct MirrorOrSteal<T> {
```
meowjesty, line 56: Any reason to keep this so generic, since the only thing that changes is the value type of the hashmap?

Also, missing docs.

meowjesty (approved):
Just a couple of nits that I'll leave up to you.

PR #3312, opened 2025-05-06
Allow targets to be listed by specific resource type

gememma:
https://linear.app/metalbear/issue/MBE-739/intellij-stuck-forever-listing

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -12,12 +12,12 @@ use std::{
 
 use clap::{ArgGroup, Args, Parser, Subcommand, ValueEnum, ValueHint};
 use clap_complete::Shell;
-use mirrord_config::target::TargetType;
 use mirrord_config::{
     feature::env::{
         MIRRORD_OVERRIDE_ENV_FILE_ENV, MIRRORD_OVERRIDE_ENV_VARS_EXCLUDE_ENV,
         MIRRORD_OVERRIDE_ENV_VARS_INCLUDE_ENV,
     },
+    target::TargetType,
     LayerConfig,
 };
 use mirrord_operator::setup::OperatorNamespace;
@@ -802,7 +802,7 @@ pub(super) struct ListTargetArgs {
 
     /// Specify the type of target to be retrieved. If `None`, all types are retrieved.
     #[arg(short = 't', long)]
-    pub target_type: Option<TargetType>
+    pub target_type: Option<TargetType>,
 }
 
 impl ListTargetArgs {
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -12,7 +12,7 @@ use std::{
 
 use clap::{ArgGroup, Args, Parser, Subcommand, ValueEnum, ValueHint};
 use clap_complete::Shell;
-use mirrord_config::target::Target;
+use mirrord_config::target::TargetType;
 use mirrord_config::{
     feature::env::{
         MIRRORD_OVERRIDE_ENV_FILE_ENV, MIRRORD_OVERRIDE_ENV_VARS_EXCLUDE_ENV,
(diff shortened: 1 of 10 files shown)


mirrord/cli/src/list.rs
```diff
@@ -219,7 +244,25 @@ pub(super) async fn print_targets(args: ListTargetArgs, rich_output: bool) -> Cl
         util::remove_proxy_env();
     }
 
-    let targets = FoundTargets::resolve(layer_config, rich_output).await?;
+    let target_types = if let Some(target_type) = args.target_type {
+        Some(vec![target_type])
+    } else {
+        match std::env::var(LS_TARGET_TYPES_ENV)
+            .ok()
+            .map(|val| serde_json::from_str::<Vec<Value>>(val.as_ref()))
```
meowjesty, line 252: Why do we need to first convert it to `Value`? Can't we deserialize it into `Vec<TargetType>` here?

meowjesty (changes requested):
I would rather have separate error variants than adding different strings to the same variant.

Also I think the error logging can be done with tracing on the fn, rather, they should all be logged as `error` anyway, and it's not like we're using progress here I guess.

mirrord/kube/src/api/kubernetes/seeker.rs
```diff
@@ -72,6 +73,71 @@ impl KubeResourceSeeker<'_> {
...
+                    .await
+            }
+            TargetType::Service if operator_active => {
+                self.simple_list_resource::<Service>("service").await
+            }
+            TargetType::ReplicaSet if operator_active => {
+                self.simple_list_resource::<ReplicaSet>("replicaset").await
+            }
+            TargetType::Targetless => {
+                tracing::error!("Cannot list targets with resource type 'targetless'");
```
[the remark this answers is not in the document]
gememma: I think i might remove all tracing stuff from this PR, since `mirrord ls` doesn't show any tracing output (i assume because the plugin consumes the output for target listing) - I can put all the information here into the specific error variants instead 👍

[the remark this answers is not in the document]
meowjesty: Oh makes sense (about not having tracing here).

I would still have it with our usual `tracing::instrument` bit on functions, because I think it's nice (if it doesn't break anything), but it's up to you.

mirrord/cli/src/config.rs
```diff
@@ -798,6 +799,10 @@ pub(super) struct ListTargetArgs {
     /// Specify config file to use.
     #[arg(short = 'f', long, value_hint = ValueHint::FilePath)]
     pub config_file: Option<PathBuf>,
+
+    /// Specify the type of target to be retrieved. If `None`, all types are retrieved.
+    #[arg(short = 't', long)]
+    pub target_type: Option<Vec<TargetType>>,
```
meowjesty, line 805: Does the help command show how to specify multiple targets? Like is it `mirrord ls --target_type "deployments;pods"`? Can you add this as docs for this field?

mirrord/cli/src/list.rs
```diff
@@ -16,6 +16,8 @@ use tracing::Level;
 
 use crate::{util, CliError, CliResult, Format, ListTargetArgs};
 
+const LS_TARGET_TYPES_ENV: &str = "MIRRORD_LS_TARGET_TYPES";
```
meowjesty, line 19: Missing docs.

mirrord/kube/src/api/kubernetes/seeker.rs
```diff
@@ -72,6 +73,57 @@ impl KubeResourceSeeker<'_> {
...
+        Ok(futures::future::try_join_all(
+            resource_types
+                .into_iter()
+                .map(|resource_type| self.filtered_single(resource_type, operator_active)),
+        )
+        .await?
+        .concat())
+    }
+
+    async fn filtered_single(
```
meowjesty, line 93: Missing docs.

meowjesty (approved):
Just some docs requests, so approving.

PR #3313, opened 2025-05-06
Rename mirrord profile to cluster profile

0x00A5:
`MirrordProfile` was introduced in #3181. The linked PR
* adds the `MirrordProfile`  **cluster-wise** CRD
* writes the yaml in set up output
* fetch the resource and apply it to the configuration if profile identifier is specified in user's config

We later realized that we want to add a namespaced profile CRD as well. Thus, we will have to rename `MirrordProfile` to `MirrordClusterProfile` to follow the same kube resource naming convention.
In this PR, we add the `MirrordClusterProfile` which is identical to the existing one but has the desired name/kind. We stop exposing the old mirrord profile CRD in setup (charts need to be updated as well for this). In the operator, we will throw a warning to the user if any old profile CRD or CR is detected and ask the user to migrate. During the migration period, we must support both profile CR mirrord. Once all user's have migrated, we will re-introduce the `MirrordProfile` for namespaced profiles.

When user specify a profile in their config that matches the legacy CR, we throw a warning like shown below:
image

### DIFF
--- a/changelog.d/mirorrdprofile.changed.md
+++ b/changelog.d/mirorrdprofile.changed.md
@@ -1,0 +1,1 @@
+Renamed `MirorrdProfile` to `MirrordClusterProfile` and support both.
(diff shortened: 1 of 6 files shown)


t4lz (approved):
What do you think about checking if there are any legacy MirrordProfiles at all, and warning the user, even if it's not the profile they're using that's deprecated?

PR #3314, opened 2025-05-08
support for MIRRORD_SKIP_EXTRA_BUILD_TOOLS

fordneild:
## What does this do
Adds a new config option enabling users to supply custom build tools they would like to "skip," e.g. not load into the layer. This can be helpful if you build and execute your binary in a single command. There was an existing set of well known build tools, so we simply decided to extend that static list with config.

### DIFF
--- a/mirrord/layer/src/load.rs
+++ b/mirrord/layer/src/load.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::HashSet,
-    env,
     ffi::OsString,
     fmt::{self, Display},
     sync::LazyLock,
@@ -37,17 +36,6 @@ static BUILD_TOOL_PROCESSES: LazyLock<HashSet<&'static str>> = LazyLock::new(|| 
         "jspawnhelper",
         "bazel-real",
     ])
-});
-
-static COMBINED_BUILD_TOOLS: LazyLock<HashSet<String>> = LazyLock::new(|| {
-    let mut combined = BUILD_TOOL_PROCESSES
-        .iter()
-        .map(|&s| s.to_string())
-        .collect::<HashSet<String>>();
-    if let Ok(extra_tools) = env::var("MIRRORD_SKIP_EXTRA_BUILD_TOOLS") {
-        combined.extend(extra_tools.split(';').map(|s| s.trim().to_string()));
-    }
-    combined
 });
 
 /// Credentials of the process the layer is injected into.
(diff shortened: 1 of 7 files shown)


mirrord/layer/src/load.rs
```diff
@@ -38,6 +39,17 @@ static BUILD_TOOL_PROCESSES: LazyLock<HashSet<&str>> = LazyLock::new(|| {
     ])
 });
 
+static COMBINED_BUILD_TOOLS: LazyLock<HashSet<String>> = LazyLock::new(|| {
```
meowjesty, line 42: I think we might not need this to be a global.

[`should_load`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/layer/src/load.rs#L97) could take the `layer_config.skip_extra_build_tools` as an argument, since we have the `LayerConfig` in [`load_type`](https://github.com/metalbear-co/mirrord/blob/8b1d9e1eeb2cb37acaadfcc563f558466d738dbc/mirrord/layer/src/load.rs#L123C9-L123C19). And pass it down to [`is_build_tool`](https://github.com/metalbear-co/mirrord/blob/8b1d9e1eeb2cb37acaadfcc563f558466d738dbc/mirrord/layer/src/load.rs#L86).

You'll also need to modify [`is_sip_only`](https://github.com/metalbear-co/mirrord/blob/8b1d9e1eeb2cb37acaadfcc563f558466d738dbc/mirrord/layer/src/load.rs#L171) to take this config as a function argument, since it calls `is_build_tool`.

mirrord/layer/src/load.rs
```diff
@@ -38,6 +39,17 @@ static BUILD_TOOL_PROCESSES: LazyLock<HashSet<&str>> = LazyLock::new(|| {
...
+    let mut combined = BUILD_TOOL_PROCESSES
+        .iter()
+        .map(|&s| s.to_string())
+        .collect::<HashSet<String>>();
+    if let Ok(extra_tools) = env::var("MIRRORD_SKIP_EXTRA_BUILD_TOOLS") {
+        combined.extend(extra_tools.split(';').map(|s| s.trim().to_string()));
```
meowjesty, line 48: When you make the changes I'm suggesting, you can use `deref().map([_])::to_vec()` on the config (since it's a `VecOrSingle`) and not have to manually `split` it.

meowjesty (changes requested):
Thanks for the PR :heart: !

I believe this config doesn't need to be a global, since it could have access to a `LayerConfig`, I've left a comment on what I think should be doable to turn this into a reality.

mirrord/layer/src/load.rs
```diff
@@ -11,7 +12,7 @@ use tracing::trace;
 
 use crate::error::LayerError;
 
-static BUILD_TOOL_PROCESSES: LazyLock<HashSet<&str>> = LazyLock::new(|| {
+static BUILD_TOOL_PROCESSES: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
```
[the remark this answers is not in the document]
meowjesty: Changed! Thanks for the suggestion alex.

mirrord/layer/src/load.rs
meowjesty: Neat! Thanks alex.

0x00A5 (approved):

meowjesty (approved):
:+1:

PR #3317, opened 2025-05-13
Default skip sip list

0x00A5:


mirrord/sip/src/lib.rs
```diff
@@ -104,6 +104,9 @@ mod main {
     pub static MIRRORD_TEMP_BIN_DIR_STRING: Lazy<String> =
         Lazy::new(|| get_temp_bin_str_prefix(&MIRRORD_TEMP_BIN_DIR_PATH_BUF));
 
+    /// A default list of processes to be skipped for SIP patching.
+    pub static MIRRORD_SIP_DEFAULT_SKIP_LIST: Lazy<&[&str]> = Lazy::new(|| &["git"]);
```
t4lz, line 108: Any reason for this not to be a `const` `[&str; 1]`?

```suggestion
    pub const MIRRORD_SIP_DEFAULT_SKIP_LIST: [&str; 1] = ["git"];
```

mirrord/sip/src/lib.rs
```diff
@@ -503,7 +506,19 @@ mod main {
     /// SIP check for binaries.
     fn is_binary_sip(path: &Path, data: &[u8], opts: SipPatchOptions) -> Result<bool> {
         // Skip patching binary if it is in the list of binaries to skip.
-        if opts.skip.iter().any(|x| path.ends_with(x)) {
+        if opts
+            .skip
+            .as_ref()
+            .unwrap_or(
+                &MIRRORD_SIP_DEFAULT_SKIP_LIST
```
t4lz, line 513: maybe this is a nitpick: maybe it's better to add the default value directly in the configuration, instead of where using that configuration. There could be more usages of the `skip_sip` config field, in the future, and it's easier to stay consistent if we define it at the source. What do you think?

mirrord/sip/src/lib.rs
```diff
@@ -503,7 +506,19 @@ mod main {
     /// SIP check for binaries.
     fn is_binary_sip(path: &Path, data: &[u8], opts: SipPatchOptions) -> Result<bool> {
         // Skip patching binary if it is in the list of binaries to skip.
-        if opts.skip.iter().any(|x| path.ends_with(x)) {
+        if opts
+            .skip
+            .as_ref()
+            .unwrap_or(
```
t4lz, line 512: I like that the user-specified list replaces and is not added to the default list (because otherwise the user wouldn't be able to choose to not skip executables from the default list), but it also means it's a potential footgun for users that as we anticipate need the defaults, but also need additional values, because it would be easy to just add the additional values, and then lose the defaults. So we should definitely document the default behaviour and the effect of specifying values in the config docs.
And then this adds another reason to set the default at the config definition and not at the usage - the docs and the defaults are more likely to go out of sync if they are defined in different places.

t4lz (changes requested):
Documenting the default is a must, let's discuss the other stuff.
Don't forget to recreate the json schema and rerun medschool after adding docs to the config.

mirrord/sip/src/lib.rs
0x00A5: The only concern I had was that I want to keep the default list in the sip crate. Add it as the default value in the config crate will make it depend on the sip crate. But, yeah, keeping it in one place is less error-prone. I can move the `const` list to the config crate.

mirrord/sip/src/lib.rs
0x00A5: Yeah, agreed, I will move it to the config crate and mention to append to the default value instead of replacing it unless to intentionally patch some of the default list of processes.

t4lz (approved):
Nice, thanks!

PR #3318, opened 2025-05-15
Fix iptables check and cleanup

Razz4780:
With IPv6 enabled, the agent is only using the ip6tables. It should use both iptables and ip6tables


gememma (approved):
🦺

PR #3323, opened 2025-05-17
namespaced mirrrod profile

0x00A5:
Before merging, confirm that all users' `MirrordProfile` CRs are migrated to `MirrordClusterProfile` and the old `MirrordProfile` CRD is removed.
Depend on https://github.com/metalbear-co/charts/pull/196 being released and applied by users first.

### DIFF
--- a/mirrord/cli/src/profile.rs
+++ b/mirrord/cli/src/profile.rs
@@ -45,15 +45,8 @@ pub enum ProfileError {
     ))]
     ProfileFetchError(KubeApiError),
 
-    #[error("failed to parse mirrord profile name: {0}")]
-    #[diagnostic(help(
-        "Use `<namespace>/<profile>` for namespaced profile. \
-        Or the field will be used as a single value for fetchting cluster-wise profile."
-    ))]
-    InvalidProfileName(String),
-
-    #[error("selected profile is in {profile:?}, but target is in {target:?}.")]
-    NamespaceConflict { target: String, profile: String },
+    #[error("mirrord profile must be cluster-wide or in the same namespace as the target ({0})")]
+    NamespaceConflict(String),
 }
 
 /// Identifier of [`MirrordClusterProfile`] and [`MirrordProfile`].
@@ -77,24 +70,12 @@ impl Display for ProfileIdentifier {
 impl FromStr for ProfileIdentifier {
     type Err = ProfileError;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let vals: Vec<_> = s.split('/').collect();
-        match vals.len() {
-            1 => Ok(Self::Cluster(
-                vals.first()
-                    .ok_or(ProfileError::InvalidProfileName(s.to_string()))?
-                    .to_string(),
-            )),
-            2 => Ok(Self::Namespaced {
-                namespace: vals
-                    .first()
-                    .ok_or(ProfileError::InvalidProfileName(s.to_string()))?
-                    .to_string(),
-                profile: vals
-                    .get(1)
-                    .ok_or(ProfileError::InvalidProfileName(s.to_string()))?
-                    .to_string(),
+        match s.split_once('/') {
+            Some((namespace, profile)) => Ok(Self::Namespaced {
+                namespace: namespace.to_string(),
+                profile: profile.to_string(),
             }),
-            _ => Err(ProfileError::InvalidProfileName(s.to_string())),
+            None => Ok(Self::Cluster(s.to_string())),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


mirrord/cli/src/profile.rs
```diff
@@ -42,23 +44,104 @@ pub enum ProfileError {
...
+                    .to_string(),
+                profile: vals
+                    .get(1)
+                    .ok_or(ProfileError::InvalidProfileName(s.to_string()))?
+                    .to_string(),
+            }),
+            _ => Err(ProfileError::InvalidProfileName(s.to_string())),
+        }
+    }
+}
```
Razz4780, line 100: I suggest using [`std::split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) for simplicity:

```rust
impl From<&str> for ProfileIdentifier {
    fn from(s: &str) -> {
        match s.split_once('/') {
            Some((namespace, name)) => todo!("namespaced"),
            None => todo!("clusterwide"),
    }
}
```

This way we also wouldn't need the `ProfileError::InvalidProfileName` variant.

mirrord/cli/src/profile.rs
```diff
@@ -42,23 +44,104 @@ pub enum ProfileError {
...
     ProfileFetchError(KubeApiError),
+
+    #[error("failed to parse mirrord profile name: {0}")]
+    #[diagnostic(help(
+        "Use `<namespace>/<profile>` for namespaced profile. \
+        Or the field will be used as a single value for fetchting cluster-wise profile."
+    ))]
+    InvalidProfileName(String),
+
+    #[error("selected profile is in {profile:?}, but target is in {target:?}.")]
```
Razz4780, line 55:
```suggestion
    #[error("mirrord profile must be cluster-wide or in the same namespace as the target ({target})")]
```
wdyt?

mirrord/cli/src/profile.rs
```diff
@@ -42,23 +44,104 @@ pub enum ProfileError {
...
+    Cluster(MirrordClusterProfile),
+    Namespaced(MirrordProfile),
+}
+
+/// Extension trait for [`MirrordProfile`] and [`MirrordClusterProfile`].
+trait ProfileView {
+    fn feature_adjustments(&self) -> impl Iterator<Item = &FeatureAdjustment>;
+
+    fn unknown_fields(&self) -> impl Iterator<Item = (&String, &serde_json::Value)>;
+}
```
Razz4780, line 112: I think you ended up implementing it only for the `ProfileFetchResult` enum, so you don't need it to be a trait

mirrord/cli/src/profile.rs
```diff
@@ -150,67 +228,75 @@ pub async fn apply_profile_if_configured<P: Progress>(
...
-            apply_profile(&mut config.feature, profile)?;
-            subtask.success(Some(&format!("mirrord profile `{name}` applied")));
-            Ok(())
-        }
-        (_, Ok(legacy_profile)) => {
-            subtask.success(Some(&format!("legacy mirrord profile `{name}` fetched")));
+    match fetch_profile(config, &profile_identifier).await {
+        Ok(profile) => {
+            subtask.success(Some(&format!(
+                "mirrord cluster profile `{profile_identifier}` fetched"
```
Razz4780, line 237:
```suggestion
                "mirrord profile `{profile_identifier}` fetched"
```

tests/src/operator/profiles.rs
```diff
@@ -137,3 +138,154 @@ pub async fn mirrord_profile_enforces_stealing(
...
     );
 }
+
+#[rstest]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[timeout(Duration::from_secs(120))]
+pub async fn mirrord_namespaced_profile_enforces_stealing(
+    #[future] kube_client: kube::Client,
+    #[future] basic_service: KubeService,
+) {
```
Razz4780, line 148: This test is basically a clone of the other one, right? I suggest merging like this:

```rust
#[rstest]
#[timeout(Duration::from_secs(120))]
pub async fn mirrord_profile_enforces_stealing(
    #[future] kube_client: kube::Client,
    #[future] basic_service: KubeService,
    #[values(true, false)] clusterwide: bool,
) {
    // create clusterwide or namespaced
    ...
}
```

Razz4780 (changes requested):
What would the UX be if the user simply upgraded the operator via chart? Without cleaning the profiles

mirrord/cli/src/profile.rs
0x00A5: Yep, had it implemented for both CRD at first, but moved them under the enum. Will remove the trait.

mirrord/cli/src/profile.rs
0x00A5: Updated with your suggestion

mirrord/cli/src/profile.rs
0x00A5: Updated, looks much cleaner now 👍

mirrord/cli/src/profile.rs
0x00A5: Removed the trait

mirrord/cli/src/profile.rs
0x00A5: Removed `cluster`, thanks!

tests/src/operator/profiles.rs
```diff
@@ -26,7 +27,7 @@ pub async fn mirrord_profile_enforces_stealing(
     let deployed_at = Instant::now();
     let target_path = service.pod_container_target();
 
-    let (_profile_guard, profile_name) = {
+    let (_profile_guard, cluster_profile_name) = {
```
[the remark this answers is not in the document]
0x00A5: merged the two tests, thanks!

PR #3326, opened 2025-05-21
Better BAD_GATEWAY error responses

Razz4780:
1. When mirrord responds with 502, the body indicates whether the error comes from the agent or from the intproxy
2. Used `std::error::Report` instead of our own wrapper (for 502 bodies)
3. Removed some unnecessary unstable features

Extracted from #3279


gememma (approved):
🌊 1 spelling nit only

PR #3327, opened 2025-05-21
Introduced a shared wrapper for client's protocol version

Razz4780:
Extracted from #3279 (it actually saves lines of code there)

### DIFF
--- a/mirrord/agent/src/util/protocol_version.rs
+++ b/mirrord/agent/src/util/protocol_version.rs
@@ -1,0 +1,52 @@
+use std::sync::{Arc, Mutex};
+
+/// Shared and cloneable [`mirrord_protocol`] version of an agent client.
+///
+/// Client's [`mirrord_protocol`] is used in multiple places.
+/// Storing it in behind a shared wrapper allows us to simplify the code
+/// and avoid passing it around in messages.
+///
+/// Its [`Default`] implementation sets the version to `1.2.2`.
+/// This is the last version that did not support version negotiation.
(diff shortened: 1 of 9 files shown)


mirrord/agent/src/util/protocol_version.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Its [`Default`] implementation sets the version to `1.2.2`.
+/// This is the last version that did not support version negotiation.
+/// See [reference](https://github.com/metalbear-co/mirrord/commit/56da9828ad2553e6c5a11124c5d65f4d4b00e6e6#diff-3c754d6cce3c1b7f4856fc34e66df5e6d8850138dd1273be60f87420bc064f73).
+///
+/// Thanks to having a default value, we can still safely match agains [`semver::VersionReq::STAR`].
+///
+/// # Note
+///
+/// This could be implemeted nicely with [arc-swap](https://docs.rs/arc-swap/latest/arc_swap/),
+/// but this struct alone it is not worth the extra dependency.
```
gememma, line 18:
```suggestion
/// but this struct alone is not worth the extra dependency.
```

mirrord/agent/src/util/protocol_version.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Clone, Debug)]
+pub struct ClientProtocolVersion(Arc<Mutex<semver::Version>>);
+
+impl Default for ClientProtocolVersion {
+    fn default() -> Self {
+        Self(Arc::new(Mutex::new(semver::Version::new(1, 0, 0))))
```
gememma, line 24: not 1.2.2 ?

gememma (commented):
SPAG nits, will approve if `ClientProtocolVersion::default()` is as intended and I just misunderstood 👀

mirrord/agent/src/util/protocol_version.rs
Razz4780: right

gememma (approved):
🥪

PR #3330, opened 2025-05-22
Eager http detection

Razz4780:


mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
+/// Represents a stolen connection in the [`TcpStealerApi`].
+struct IncomingConnection {
+    /// For sending data to the connection source.
+    ///
+    /// If this connection is an HTTP request, this is for sending data after the HTTP upgrade.
+    data_tx: Option<mpsc::Sender<Vec<u8>>>,
+    /// If the client's [`mirrord_protocol`] version does not support
+    /// [`DaemonTcp::HttpRequestChunked`], we must gather the incoming body frames here.
+    ///
+    /// When we've read the full bobdy, we send the full request in one of the legacy types.
```
0x00A5:
```suggestion
    /// When we've read the full body, we send the full request in one of the legacy types.
```

mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
+    /// (only for HTTP requests)
+    request_in_progress: Option<HttpRequest<InternalHttpBody>>,
+    /// For sending the HTTP response.
+    ///
+    /// (only for HTTP requests)
+    response_tx: Option<ResponseProvider>,
+    /// For sending the HTTP response body frames.
+    ///
+    /// (only for HTTP requests)
+    response_body_tx: Option<ResponseBodyProvider>,
```
0x00A5: nit: correct me if I am wrong, only one of the three (data / response / resposne body tx) can be not none at a time. I wonder if there is a better way (enum and/or with typestate) to present a `IncmoingConnection`'s state so that, for example, if header is not sent yet, one can only call the `send_response` method, or after header is sent, one can only send frames.

mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
+                        is_last: true,
+                        connection_id,
+                        request_id: Self::REQUEST_ID,
+                    }),
+                ));
+                self.queued_messages.push_back(message);
+            }
+        } else if request_head.body_finished.not() {
+            // Request body has not finished, and the client does not support chunked.
+            // We need to wait for the body to finish.
```
0x00A5: When handling this variant, does the stealer buffer the complete body and always send the entirety of the body within each stealer message? Asking because later we swap the body stored in requests in progress instead of appending.

mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
 
-                    let key = (response.connection_id, response.request_id);
-                    self.response_body_txs.insert(key, tx.clone());
+            LayerTcpSteal::HttpResponseChunked(ChunkedResponse::Start(response)) => {
+                self.send_response(response, false).await;
+            }
 
-                    self.http_response(HttpResponseFallback::Streamed(http_response))
-                        .await?;
+            LayerTcpSteal::HttpResponseChunked(ChunkedResponse::Body(body)) => {
```
0x00A5: nit: how about moving the handling of this variant to a method like `TcpStealerApi::send_chunk` as comparing to `TcpStealerApi::send_response`?

mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
+            connection_id,
+            remote_address: info.peer_addr.ip(),
+            destination_port: info.original_destination.port(),
+            source_port: info.peer_addr.port(),
+            local_address: info.local_addr.ip(),
+        };
+
+        let message = if self.protocol_version.matches(&MODE_AGNOSTIC_HTTP_REQUESTS) {
+            DaemonTcp::NewConnectionV2(NewTcpConnectionV2 {
+                connection: new_connection,
```
0x00A5, line 332: nit: why does v2 wraps v1 instead of having fields flattened and have a `V2::from_v1`?

mirrord/agent/src/steal/api.rs
```diff
@@ -82,186 +113,564 @@ impl TcpStealerApi {
...
-                self.connection_unsubscribe(connection_id).await
+                self.connections.remove(&connection_id);
+                self.incoming_streams.remove(&connection_id);
+            }
+
+            LayerTcpSteal::Data(data) => {
+                let connection = self.connections.get_mut(&data.connection_id);
+
+                let Some(connection) = connection else {
+                    return Ok(());
```
0x00A5, line 516: should we log a warning or something if this happens?

mirrord/agent/src/steal/api.rs
Razz4780: Yeah, that is the idea. This allows us to handle old CLIs/operators (though these http request variants are very old, so we'd like to silently drop them sooner or later, and no one would notice).
We consume the body when all request frames have been received. If the client's protocol version is *really* old, we transform it to vector of bytes

mirrord/agent/src/steal/api.rs
Razz4780: Oh, this is a pretty standard scenario:
1. User app sends data, intproxy sends to the agent
2. Peer closes the connection, agent sends TcpClose and removes the entry from the map
3. Data from the user is received by the agent

mirrord/agent/src/steal/api.rs
Razz4780: Hmm, yeah. Would be nicer. I'll fix it

mirrord/agent/src/steal/api.rs
Razz4780: Dunno, that's what I came up with. V2 is essentially V1 + tls metadata

mirrord/agent/src/steal/api.rs
0x00A5: Thanks for explaining!

mirrord/agent/src/steal/api.rs
Razz4780: Moved handling client messages to a separate struct that represents the connection from the client perspective ^^

mirrord/agent/src/incoming/connection.rs
```diff
@@ -1,67 +1,121 @@
...
-    /// Returns the address of the TCP socket that accepted this connection.
-    pub fn local_addr(&self) -> io::Result<SocketAddr> {
-        self.stream.local_addr()
-    }
+/// Supertrait for incoming IO streams.
+///
+/// [`MaybeHttp::detect`] transforms the incoming [`TcpStream`](tokio::net::TcpStream)
+/// into one of multiple types due to possible TLS handshake and HTTP detection.
+///
+/// Having a super trait allows us to return [`Box<dyn IncomingStream>`]
```
0x00A5:
```suggestion
/// Having a super trait allows us to return [`Box<dyn IncomingIO>`]
```

mirrord/agent/src/incoming/connection.rs
```diff
@@ -79,6 +133,170 @@ impl AsyncWrite for RedirectedConnection {
...
+            .acceptor()
+            .accept(redirected.stream)
+            .await
+            .map_err(HttpDetectError::TlsAccept)?;
+        let tls_connector = tls_handler.connector(stream.get_ref().1);
+
+        let (stream, http_version): (Box<dyn IncomingIO>, _) = match tls_connector.alpn_protocol() {
+            Some(tls::HTTP_2_ALPN_NAME) => (Box::new(stream), Some(HttpVersion::V2)),
+            Some(tls::HTTP_1_1_ALPN_NAME) => (Box::new(stream), Some(HttpVersion::V1)),
+            Some(tls::HTTP_1_0_ALPN_NAME) => (Box::new(stream), Some(HttpVersion::V1)),
```
0x00A5: Should these three variants get matched to `Box::new(IncomingIoWrapper { ... })` too so that the metric guard gets dropped when the stream is dropped.

mirrord/agent/src/incoming/connection.rs
```diff
@@ -79,6 +133,170 @@ impl AsyncWrite for RedirectedConnection {
...
+        let this = self.get_mut();
+
+        let Some(rx) = this.rx.as_mut() else {
+            return Poll::Ready(None);
+        };
+
+        let item = match std::task::ready!(rx.poll_recv(cx)) {
+            Some(item) => item,
+            None => IncomingStreamItem::Finished(Err(ConnError::Dropped)),
+        };
```
0x00A5: nit: unwrap_or_else()?

mirrord/agent/src/incoming/task.rs
```diff
@@ -97,61 +113,161 @@ where
...
+                %source,
+                %destination,
+                "Redirected connection port is no longer stolen, dropping",
+            );
+            return;
+        };
+
+        let tx = self.internal_tx.clone();
+        let tls_store = self.tls_store.clone();
+        tokio::spawn(async move {
```
0x00A5, line 141: Just curious, do we have to spawn this as a helper task then sending the response back for the main select loop to handle? Or can we make this function async again and call maybe http detect, and then call what's in handle initialized with the conn? Saving channeling the conn back to the select loop.

mirrord/agent/src/incoming/connection.rs
Razz4780: Yup, thanks

mirrord/agent/src/incoming/task.rs
Razz4780: This is because this function is called from the main `RedirectorTask` loop, and we have one global instance running in the agent. At all costs we must avoid blocking there on IO, and `MaybeHttp::detect` does read data from the incoming connection. If we don't do it in the background, we risk that one "sleepy" peer blocks processing of all new connections.
Added a doc to `RedirectorTask::run_inner`

mirrord/agent/src/steal/subscriptions.rs
```diff
@@ -61,50 +69,59 @@ impl PortSubscriptions {
...
+                        STEAL_FILTERED_PORT_SUBSCRIPTION.fetch_sub(1, Ordering::Relaxed);
+                    }
+                    filters.insert(client_id, filter);
+                }
+
+                (PortSubscription::Filtered(filters), None) => {
+                    STEAL_FILTERED_PORT_SUBSCRIPTION.fetch_sub(filters.len(), Ordering::Relaxed);
+                    e.insert(PortSubscription::Unfiltered(client_id));
+                }
+            },
```
0x00A5: `Client`s here are basically layer instances spawn by the same mirrord exec process right? It's probably uncommon for this to happen, but in three of the cases, a client which already subscribes to a port could lose their subscription silently:
* client A subscribed to a port without filter
  * client B takes over with a filtered subscription
  * client B takes over the unfiltered subscription
* clients subscribed to the port with their filter
  * client A takes the port over with a unfiltered subscription

It would be nice if we can inform the clients who lose their subscription (by sending a stealer log message). Or if it is a bit cumbersome to implement, the agent can log this at debug/trace level, so we have a clue when debugging. WDYT?

mirrord/agent/src/steal/subscriptions.rs
```diff
@@ -61,50 +69,59 @@ impl PortSubscriptions {
...
+        port: u16,
         filter: Option<HttpFilter>,
-    ) -> Result<RemoteResult<Port>, RedirectorTaskError> {
-        let filtered = filter.is_some();
+    ) -> Result<(), RedirectorTaskError> {
+        let metric = if filter.is_some() {
+            &STEAL_FILTERED_PORT_SUBSCRIPTION
+        } else {
+            &STEAL_UNFILTERED_PORT_SUBSCRIPTION
+        };
```
0x00A5: Just a nitpick, please feel free to ignore. I think this make the code a little bit harder to read. How about handle the metric add/sub together in each match variant, so it becomes something like
```rust
(PortSubscription::Unfiltered(..), Some(filter)) => {
    ...
    STEAL_UNFILTERED_PORT_SUBSCRIPTION.fetch_sub(1, Ordering::Relaxed);
    STEAL_FILTERED_PORT_SUBSCRIPTION.fetch_add(1, Ordering::Relaxed);
}
```
In two cases, `(PortSubscription::Unfiltered(..), None)` and `(PortSubscription::Filtered(filters), Some(filter))`, we also avoid doing a `fetch_add` right after a `fetch_sub`.

mirrord/agent/src/steal/task.rs
```diff
@@ -0,0 +1,290 @@
...
+
+            (PortSubscription::Filtered(..), StolenTraffic::Tcp(tcp)) => {
+                tcp.pass_through();
+                return;
+            }
+
+            (PortSubscription::Unfiltered(client_id), StolenTraffic::Tcp(tcp)) => {
+                let client = clients
+                    .get(client_id)
+                    .expect("subscription should be removed on client disconnect");
```
0x00A5: Are these `expects` intentional here? Or would it be better to log an error and `return` early?

mirrord/agent/src/incoming/connection/tcp_task.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        if let Destination::StealingClient { data_tx, .. } = &self.destination {
+            let _ = data_tx.send(IncomingStreamItem::Finished(result)).await;
+        }
+    }
+
+    async fn run_inner(&mut self) -> Result<(), ConnError> {
+        let mut source_writes = true;
+        let mut destination_writes = true;
+        let mut read_buf = BytesMut::with_capacity(64 * 1024);
```
0x00A5, line 38: The `self.inconming_io` implements aysnc read and write and uses this buffer when reading. I wonder if we can do the same for `Destination`, i.e. getting rid of the `BytesMut` inside `Destination`, so we can turn the while loop below into using tokio copy io bidirectional.

mirrord/agent/src/steal/subscriptions.rs
Razz4780: Oke, changed

mirrord/agent/src/steal/subscriptions.rs
Razz4780: Not quite, clients here are mirrord sessions (so more like intproxy instances). The only case where we overwrite is when one mirrord session makes a subscription that evicts a subscription from another mirrord session. And this is being managed in the operator, which will notify the user ^^

I added debug logs though, why not

mirrord/agent/src/steal/task.rs
Razz4780: This should never happen - when we delete the client from `self.clients`, we always delete their subscription from `self.subscriptions`. I changed the `expects` to error logs though. In case this contract is broken in the future

mirrord/agent/src/incoming/connection/tcp_task.rs
Razz4780: I tried that, yeah. But with `copy_bidirectional` we loose information about which side of the connection failed, we just get an io error

0x00A5 (approved):

PR #3337, opened 2025-05-25
Add fallback_name to MirrordWorkloadQueueRegistry CRD

aviramha:


mirrord/operator/src/crd.rs
```diff
@@ -357,6 +357,9 @@ pub struct SqsQueueDetails {
     /// this queue, applications will get a modified name from that source.
     pub name_source: QueueNameSource,
 
+    /// Fallback queue name, if name source isn't set or not available.
```
t4lz, line 360:
```suggestion
    /// Fallback queue name, if the source specified in `nameSource` is not present on the target.
```

mirrord/operator/src/crd.rs
```diff
@@ -357,6 +357,9 @@ pub struct SqsQueueDetails {
...
+    pub fallback_name: Option<String>,
```
t4lz, line 361:
```suggestion
    /// If the configured source is not present and the fallback is used - the configured source will be used
    /// to make the target use the temporary queue.
    /// For example, if `nameSource` is `envVar: MEME_QUEUE_NAME`, but `MEME_QUEUE_NAME` is not
    /// present in the target, and `IncomingMemeQueue.fifo` was set as a fallback queue name, then
    /// the target will be modified to include the environment variable `MEME_QUEUE_NAME`, with the name
    /// of the temporary queue as a value.
    /// Setting a fallback name only makes sense if the target application indeed uses the defined queue
    /// name source to override the source it uses on its absence.
    pub fallback_name: Option<String>,
```
We didn't discuss this, but if this is how it's going to work, we should document it. Or did you have something different in mind?

mirrord/operator/src/crd.rs
```diff
@@ -357,6 +357,17 @@ pub struct SqsQueueDetails {
...
+    /// Fallback queue name, if the source specified in `nameSource` is not present on the target.
+    /// If the configured source is not present and the fallback is used - the configured source will be used
+    /// to make the target use the temporary queue.
+    /// For example, if `nameSource` is `envVar: MEME_QUEUE_NAME`, but `MEME_QUEUE_NAME` is not
+    /// present in the target, and `IncomingMemeQueue.fifo` was set as a fallback queue name, then
+    /// the target will be modified to include the environment variable `MEME_QUEUE_NAME`, with the name
+    /// of the temporary queue as a value.
+    /// Setting a fallback name only makes sense if the target application indeed uses the defined queue
+    /// name source to override the source it uses on its absence.
+    pub fallback_name: Option<String>,
```
t4lz, line 369: Will upgrading work on clusters with existing registry CRs?

aviramha: It's Option, so should, no?

PR #3346, opened 2025-05-29
Fixed big packets crashing mirrord agent 

aviramha:
Fixed big packets crashing mirrord agent  - skip and warn, allow user to override max packet size

### DIFF
--- a/mirrord/agent/src/entrypoint/setup.rs
+++ b/mirrord/agent/src/entrypoint/setup.rs
@@ -47,10 +47,15 @@ pub(super) async fn start_sniffer(
 ) -> BackgroundTask<SnifferCommand> {
     let (command_tx, command_rx) = mpsc::channel::<SnifferCommand>(1000);
 
-    let packet_buffer_size = envs::MIRROR_BUFFER_SIZE.try_from_env().unwrap_or_else(| err | {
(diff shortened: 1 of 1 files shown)


mirrord/agent/src/entrypoint/setup.rs
```diff
@@ -47,11 +47,13 @@ pub(super) async fn start_sniffer(
 ) -> BackgroundTask<SnifferCommand> {
     let (command_tx, command_rx) = mpsc::channel::<SnifferCommand>(1000);
 
+    let packet_buffer_size = envs::MIRROR_BUFFER_SIZE.from_env_or_default();
     let sniffer = runtime
         .spawn(TcpConnectionSniffer::new(
             command_rx,
             args.network_interface.clone(),
             args.is_mesh(),
+            packet_buffer_size,
```
gememma, line 56: This needs to be `Some`, or you could have it as not an option and use the default value here instead 👀

gememma: actually, i think `from_env_or_default` might return 0 is the env var isnt set here 😬

gememma (changes requested):
the behaviour of the size buffer when the env var is not set needs double checking 👀 I'm owrried it will default to 0

gememma (approved):
nice 🥳 📦 💥

PR #3351, opened 2025-06-04
WIP: Add SQS splitting test case with fallback value with json object

t4lz:
closes MBE-983

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6243,6 +6243,7 @@ version = "0.1.0"
 dependencies = [
  "aws-config",
  "aws-sdk-sqs",
+ "serde_json",
  "tokio",
 ]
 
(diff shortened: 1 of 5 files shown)


tests/local-test-operator-patch.yaml
```diff
@@ -0,0 +1,22 @@
+# Watch out, this also sets the image.
+# To apply:
+# kubectl patch deployment mirrord-operator -n mirrord --patch-file tests/local-test-operator-patch.yaml
```
Razz4780: What's this :eyes: 
Should rather be a guide in contributing.md on how to run sqs tests with localstack

Razz4780 (approved):

tests/local-test-operator-patch.yaml
t4lz: Ok, deleted from here and [added to the operator's contributing guide](https://github.com/metalbear-co/operator/pull/920) instead.

PR #3353, opened 2025-06-05
Fix sqs tests

Razz4780:


tests/src/utils/sqs_resources.rs
```diff
@@ -108,50 +100,64 @@ impl SqsTestResources {
...
+        }
     }
 }
 
+const AWS_ENDPOINT_ENV: &str = "AWS_ENDPOINT_URL";
+
+/// Required to build an SQS client.
+const AWS_CREDS_ENVS: &[&str] = &[
+    AWS_ENDPOINT_ENV,
+    "AWS_REGION",
```
DmitryDodzin: Small nitpick can also be done like `AWS_ENDPOINT_ENV` because it is used once later

DmitryDodzin (approved):

tests/src/utils/sqs_resources.rs
Razz4780: Done :+1:

PR #3354, opened 2025-06-05
Browser extension initiation

0x00A5:


mirrord/cli/src/browser.rs
```diff
@@ -0,0 +1,94 @@
+use std::process::Command;
+
+use base64::engine::{general_purpose::STANDARD, Engine};
+use mirrord_config::feature::network::NetworkConfig;
+use mirrord_progress::Progress;
+use serde::Serialize;
+
+// Permanent ID once extension is published
+const MIRRORD_CHROME_EXTENSION_ID: &str = "ejpgopblebjoiepjnhngaffejlaggogm";
```
0x00A5: Once our browser extension is released to Chrome Web Store, we will get a permanent static ID. We need to update this with that ID before merging.

DmitryDodzin: I would do `option_env!` to allow compiling with another extension id if needed for local development

DmitryDodzin (approved):
LGTM, just a nitpick

PR #3355, opened 2025-06-07
Exclude agent from mesh (WIP)

0x00A5:

PR #3356, opened 2025-06-09
Exclude agent from mesh

DmitryDodzin:
Building on top of #3355

### DIFF
--- a/mirrord/agent/iptables/src/lib.rs
+++ b/mirrord/agent/iptables/src/lib.rs
@@ -10,7 +10,11 @@ use tracing::{warn, Level};
 use crate::{
     error::IPTablesResult,
     flush_connections::FlushConnections,
-    mesh::{exclusion::WithMeshExclusion, istio::AmbientRedirect, MeshRedirect, MeshVendorExt},
+    mesh::{
+        exclusion::{MeshExclusion, WithMeshExclusion},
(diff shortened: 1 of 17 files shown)


mirrord/agent/iptables/src/lib.rs
```diff
@@ -212,6 +218,11 @@ where
             redirect = Redirects::FlushConnections(FlushConnections::create(Box::new(redirect))?)
         }
 
+        if with_mesh_exclusion {
+            redirect =
+                Redirects::WithMeshExclusion(WithMeshExclusion::create(ipt, Box::new(redirect))?)
+        }
+
```
0x00A5: Maybe worth leaving a comment that this order must not change, i.e. `WithMeshExclusion` stays in the outer layer, or else `Redirect::exclusion` will return `None`. WDYT? Same with `Redirect::load`.

mirrord/agent/iptables/src/lib.rs
```diff
@@ -25,12 +25,16 @@ mod prerouting;
 mod redirect;
 mod standard;
 
+pub use mesh::exclusion::MeshExclusion;
+
```
0x00A5: No longer need to expose this struct since now we use `SafeIpTables` 👍

Razz4780 (approved):

PR #3361, opened 2025-06-13
Pass user git branch to operator for Jira app integration

gememma:


mirrord/cli/src/util.rs
```diff
@@ -75,3 +78,26 @@ pub fn intproxy_container_mode() -> bool {
...
+    if let Ok(branch_name) = std::env::var(MIRRORD_BRANCH_NAME_ENV) {
+        return Some(branch_name);
+    }
+
+    match Command::new("git")
+        .args(["branch", "--show-current"])
+        .output()
+        .await
+        .ok()
+        .map(|output| String::from_utf8(output.stdout).unwrap_or_default())
```
Razz4780: Gotta check `output.status.success()` first. Also, We should probably log on debug:
1. Error from the `output()` call <- we failed to call the command, git probably not installed
2. `.stderr` and `.status` from the output, if the command itself failed

mirrord/cli/src/util.rs
```diff
@@ -75,3 +78,26 @@ pub fn intproxy_container_mode() -> bool {
...
+    {
+        Some(output) if !output.is_empty() => Some(output),
+        _ => None,
+    }
```
Razz4780: This looks like it could be chained: `map(|output| ...).filter(|name| name.is_empty().not())`

mirrord/operator/src/client.rs
```diff
@@ -801,6 +802,7 @@ impl OperatorApi<PreparedClientCert> {
             // Kafka and SQS splits are passed in the request body.
             kafka_splits: Default::default(),
             sqs_splits: Default::default(),
+            branch_name: None
```
Razz4780: Looks like we omitted the copy target flow? :V

gememma: just to double check, this wont cause extra sessions to be recorded right? one session from the POV of the user corresponds exactly to one `OperatorSession` instance in the operator?

Razz4780: Yeah, it should

mirrord/cli/src/execution.rs
```diff
@@ -387,7 +390,7 @@ impl MirrordExecution {
             remove_proxy_env();
         }
 
-        let (connect_info, mut connection) = create_and_connect(config, progress, analytics)
+        let (connect_info, mut connection) = create_and_connect(config, progress, analytics, None)
```
Razz4780: Oh wait, we gotta fetch the branch name here as well

mirrord/cli/src/util.rs
```diff
@@ -75,3 +78,40 @@ pub fn intproxy_container_mode() -> bool {
...
+        .output()
+        .await
+    {
+        Ok(output) if output.status.success() => Some(output)
+            .map(|output| String::from_utf8(output.stdout).unwrap_or_default())
+            .filter(|string| !string.is_empty()),
+        Ok(output) => {
+            tracing::debug!(
+                status = %output.status,
+                stderr = String::from_utf8(output.stderr).unwrap_or_default(),
```
Razz4780: there's `from_utf8_lossy`, should be nicer

Razz4780 (changes requested):

mirrord/cli/src/util.rs
```diff
@@ -75,3 +78,40 @@ pub fn intproxy_container_mode() -> bool {
...
+    }
+
+    match Command::new("git")
+        .args(["branch", "--show-current"])
+        .output()
+        .await
+    {
+        Ok(output) if output.status.success() => Some(output)
+            .map(|output| String::from_utf8(output.stdout).unwrap_or_default())
+            .filter(|string| !string.is_empty()),
```
Razz4780:
```suggestion
        Ok(output) if output.status.success() => String::from_utf8(output.stdout)
            .ok()
            .map(|output| output.trim().to_string())
            .filter(|string| !string.is_empty()),
```

Razz4780: Better trim it

Issue #3285, opened 2025-04-23
Trying to use container feature on WSL


BradleyMcCallionCOT:
### Bug Description

Unable to run mirrord in container mode via `mirrord container -f .devops/mirrord/mirrord.container.json -- docker run dotnet-api:latest`

With error on STDOUT:

```
Error:   × An error ocurred when spawning internal proxy: proxy did not print port number to stdout
```

### Steps to Reproduce

1. Run on WSL within Windows 
2. Build any container image and make sure it runs with `docker run...`
3. Find your `external_proxy.host_ip` by resolving `host.docker.internal` from within a container running on your docker network.
4. Use this IP to set this config value
5. Run mirrord in container mode using the built image with this config

### Backtrace

```shell
mirrord container is currently an unstable feature                                                                                                       * Operator license is valid for 3 more days.
! internal proxy stderr: Error:   × Failed to set up TCP listener for accepting intproxy connections: Cannot
! internal proxy stderr:   │ assign requested address (os error 99)
! internal proxy stderr:   help: This is a bug. Please report it in our Discord or GitHub repository.
! internal proxy stderr:         
! internal proxy stderr:         >> Please open a new bug report at https://github.com/metalbear-co/
! internal proxy stderr:         mirrord/issues/new/choose
! internal proxy stderr:         
! internal proxy stderr:         >> Or join our Discord https://discord.gg/metalbear and request help
! internal proxy stderr:         in #mirrord-help
! internal proxy stderr:         
! internal proxy stderr:         >> Or email us at hi@metalbear.co
! internal proxy stderr:         
! internal proxy stderr:         
! internal proxy stderr: 
  x ! mirrord container is currently an unstable feature
    x ! internal proxy stderr: 
      ✓ using operator
        ✓ operator version compatible
        ✓ operator license valid
        ✓ user credentials prepared
        ✓ session started
          ✓ connected to the target                                                                                                                        Error:   × An error ocurred when spawning internal proxy: proxy did not print port number to stdout
  help: This is a bug. Please report it in our Discord or GitHub repository.
        
        >> Please open a new bug report at https://github.com/metalbear-co/mirrord/issues/new/choose
        
        >> Or join our Discord https://discord.gg/metalbear and request help in #mirrord-help
        
        >> Or email us at hi@metalbear.co
```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell
[...]

PR #3362, opened 2025-06-13
Automatically add config for user when we detect it's `mirrord container docker` in wsl.

meowjesty:
- Issue: https://github.com/metalbear-co/mirrord/issues/3285

Sometimes defaulting to `localhost` and random address may cause problems. We have the configs that allows the user to set the correct addresses, so this PR just sets it for them by default (using default docker values).

- Another approach would be to resolve the container address from within a container;

### DIFF
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -77,7 +77,11 @@ pub(crate) enum ExternalProxyError {
     Intproxy(#[from] IntProxyError),
 
     #[error("Failed to set up TCP listener for accepting intproxy connections: {0}")]
-    #[diagnostic(help("{GENERAL_BUG}"))]
+    #[diagnostic(help(
+        "If you're trying to run `mirrord container` in WSL, try setting the\
+        `container.override_host_ip` to the internal container runtime address.\
+        {GENERAL_BUG}"
+    ))]
     ListenerSetup(std::io::Error),
 
     #[error("Failed to open log file at `{0}`: {1}")]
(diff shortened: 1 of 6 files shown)


mirrord/cli/src/wsl.rs
```diff
@@ -0,0 +1,45 @@
...
+use mirrord_config::LayerConfig;
+
+use crate::ContainerRuntime;
+
+/// Modifies the [`LayerConfig`] for WSL environments when running the `mirrord container`.
+///
+/// - Only when the runtime is Docker, and the user has not specified neither the
+///   `external_proxy.host_ip` nor the `container.override_host_ip`, so we use the default docker
+///   values.
+pub(super) fn container_config_for_wsl(runtime: ContainerRuntime, config: &mut LayerConfig) {
```
Razz4780:
```suggestion
pub(super) fn adjust_container_config_for_wsl(runtime: ContainerRuntime, config: &mut LayerConfig) {
```

mirrord/cli/src/error.rs
```diff
@@ -77,7 +77,11 @@ pub(crate) enum ExternalProxyError {
     Intproxy(#[from] IntProxyError),
 
     #[error("Failed to set up TCP listener for accepting intproxy connections: {0}")]
-    #[diagnostic(help("{GENERAL_BUG}"))]
+    #[diagnostic(help(
+        "If you're trying to run `mirrord container` in WSL, try setting the\
+        `container.override_host_ip` to the internal container runtime address.\
```
Razz4780:
```suggestion
        "If you're trying to run `mirrord container` in WSL, try setting \
        `container.override_host_ip` to the internal container runtime address. \
```

PR #3363, opened 2025-06-15
Release script

0x00A5:


scripts/release.sh
```diff
@@ -0,0 +1,112 @@
...
+        echo "Error: You are on branch '$CURRENT_BRANCH' with uncommitted or unstaged changes. Please commit or stash them before switching to 'main'."
+        exit 1
+    else
+        echo "You are on branch '$CURRENT_BRANCH'. Switching to 'main'..."
+        git checkout main
+    fi
+fi
+
+# Ensure main is up to date
+git pull origin main
```
t4lz: Some devs might call the remote something else, like e.g. "upstream", and I think that's valid (even though mine is called "origin" like yours). Is there a reason not to change this to
```suggestion
git pull
```
since we're already on `main`?

scripts/release.sh
```diff
@@ -0,0 +1,112 @@
...
+# Bump version in Cargo.toml
+cargo set workspace.package.version "$VERSION"
+# Refresh Cargo.lock to reflect version change
+cargo update -w -q
+
+# Build changelog
+towncrier build --yes --version "${VERSION}"
+
+# Stage all changes
+git add .
```
t4lz: I think it's possible that there are still untracked files in the working tree, right? And then we'll stage and later commit them. How about just adding the files that should change - Cargo.lock, Cargo.toml, changelog.d/, CHANGELOG.md, and I think that's it, right?

0x00A5: To prevent checking out main while having untracked files, the script runs `git status --porcelain` first, and if there is untracked files, the script `exit 1`.

scripts/release.sh
0x00A5: That's a good point about different remote. I will make the script supports user specified remote. In that case, I'd need to use `git pull $specified_remote main` in case the specified remote is not the same as the main branch tracks.

0x00A5: Changed this to
1. prompt the user for their choice of remote by providing a list of available remotes in the local repo
2. default to origin
3. `git pull "$selected_remote" main` in case the selected remote is different from the remote that main is tracking.

t4lz: Why not just `git pull` though?

t4lz: and then we don't need to prompt the user or anything, we just use whatever remote their `main` tracks...

0x00A5: I was thinking if the script user has `upstream` (metalbear) and `origin` (their fork), this can allow them to choose where to pull the latest main from. But we can keep it simple and rely on the user to update their main before running the script.

0x00A5: Removed the "select remote" bit, and simply `git pull`.

scripts/release.sh
aviramha: but there's very specific paths that are affected by the previous commands:
cargo.toml, cargo.lock, changelog.d/ directory and the changelog file
I never have my git in porcelain :D

0x00A5: Changed `git add .` to adding specific paths

PR #3365, opened 2025-06-17
Mbe 892 make httptcp data cheaply cloneable in mirrord protocol

BigLullu:
Replaced Vec<u8> with newly implemented struct Payload for zero copy message handling


changelog.d/3365.added.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Introduced a new struct Payload that enables to cheap clone messages
+  and enables almost zero copy message handling
```
Razz4780: Files in changelog.d should contain just the message.
 
```suggestion
Introduced a new struct `Payload`, that makes mirrord-protocol messages cheaply cloneable.
```

changelog.d/3365.added.md
```diff
@@ -0,0 +1,4 @@
+### Added
```
Razz4780: This is an internal change, the file should be named `3365.internal.md`

mirrord/protocol/src/payload.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Payload is a wrapper for bytes, is used as message body in the protocol and by agent and
+/// intproxy as zero copy message payload
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(transparent)]
+pub struct Payload(
+    #[serde(
+        serialize_with = "serialize_payload_inner",
+        deserialize_with = "deserialize_payload_inner"
+    )]
```
Razz4780: I'd implement `Serialize` and `Deserialize` manually for `Payload` instead

mirrord/protocol/src/payload.rs
```diff
@@ -0,0 +1,139 @@
+use std::{borrow::Borrow, slice::SliceIndex};
+
+use bincode::{Decode, Encode};
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+
+/// Payload is a wrapper for bytes, is used as message body in the protocol and by agent and
+/// intproxy as zero copy message payload
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
```
Razz4780:
```suggestion
#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
```

mirrord/protocol/src/payload.rs
```diff
@@ -0,0 +1,139 @@
...
+#[serde(transparent)]
+pub struct Payload(
+    #[serde(
+        serialize_with = "serialize_payload_inner",
+        deserialize_with = "deserialize_payload_inner"
+    )]
+    Bytes,
```
Razz4780:
```suggestion
    pub Bytes,
```

mirrord/protocol/src/payload.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.0.as_ptr()
+    }
+
+    pub fn get<I>(&self, index: I) -> Option<&[u8]>
+    where
+        I: SliceIndex<[u8], Output = [u8]>,
+    {
+        self.0.as_ref().get(index)
+    }
+}
```
Razz4780: All of this could be just

```rust
impl Deref for Payload {
    type Target = Bytes;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

File: mirrord/protocol/src/payload.rs [hunk 7]
```diff
@@ -0,0 +1,139 @@
...
+    fn from(bytes: B) -> Self {
+        Payload(bytes.into())
+    }
+}
+
+impl Borrow<[u8]> for Payload {
+    fn borrow(&self) -> &[u8] {
+        &self.0
+    }
+}
```
Razz4780: yagni, I think. We'd be ok with just

```rust
impl<B: Into<Bytes>> From<B> for Payload {
    ...
}

impl From<Payload> for Bytes {
    ...
}
```

mirrord/protocol/src/payload.rs
Razz4780: Just a suggestion tho

mirrord/protocol/src/payload.rs
BigLullu: I totally agree. The only doubt I still have is about the ToPayload trait — it's very clean and clear when we need to build a value from a reference, like in:

`let payload = b"payload here".to_payload()`

What would you suggest as an alternative in such cases?

mirrord/vpn/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(concat_idents)]
 #![feature(try_blocks)]
 #![deny(unused_crate_dependencies)]
+extern crate core;
```
Razz4780: Who's this little fella? :eyes:

mirrord/intproxy/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(error_reporter)]
 #![warn(clippy::indexing_slicing)]
 #![deny(unused_crate_dependencies)]
+extern crate core;
```
Razz4780: What's this about? :eyes:

mirrord/protocol/src/payload.rs
Razz4780: Yeah, it's ok ^^

PR #3374, opened 2025-06-18
Priority class name for targetless agent pod

0x00A5:


mirrord/config/src/agent.rs
```diff
@@ -391,6 +391,12 @@ pub struct AgentConfig {
     #[config(env = "MIRRORD_AGENT_EXCLUDE_FROM_MESH", default = false)]
     pub exclude_from_mesh: bool,
 
+    /// ### agent.priority_class {#agent-priority_class}
+    ///
+    /// Allows setting the priority class name on the agent pod when using targetless mode.
+    #[config(env = "MIRRORD_AGENT_PRIORITY_CLASS", default = None)]
```
aviramha: can drop the env

0x00A5: Removed

mirrord/config/src/agent.rs
```diff
@@ -391,6 +391,11 @@ pub struct AgentConfig {
     #[config(env = "MIRRORD_AGENT_EXCLUDE_FROM_MESH", default = false)]
     pub exclude_from_mesh: bool,
 
+    /// ### agent.priority_class {#agent-priority_class}
+    ///
+    /// Allows setting the priority class name on the agent pod when using targetless mode.
+    pub priority_class: Option<String>,
```
meowjesty, line 410: Could you add an example? And maybe a brief note on why you would need to set this up when using `targetless` (since it's specific to this case, I wonder why it's a thing here, and not used for other targets)?

changelog.d/agent_priority_class.added.md
meowjesty, line 1: Doesn't this have an issue? I think for when it doesn't, we have to add an `+` to the file name? smh we should have written docs about the changelog somewhere...

mirrord/kube/src/api/container/pod.rs
```diff
@@ -266,6 +267,73 @@ impl ContainerVariant for PodTargetedVariant<'_> {
 
         let mut pod = self.inner.as_update();
         pod.merge_from(update);
+
+        // do not configure priority class if targeted
```
meowjesty: Isn't this more like?

```suggestion
        // Remove `priority_class` from spec if it's targeted.
```

changelog.d/agent_priority_class.added.md
0x00A5: Nice catch, you are right. According the doc: https://towncrier.readthedocs.io/en/stable/markdown.html, I need to either start the file name with the issue number or put a + sign. I will name it `1007.added.md` then.

mirrord/kube/src/api/container/pod.rs
0x00A5: updated with your suggestion

changelog.d/agent_priority_class.added.md
0x00A5: Renamed

mirrord/config/src/agent.rs
0x00A5: Provided an example and added more explanation about the setting.

Issue #3371, opened 2025-06-17
Improve error message when mirrord-config extension is not supported.


josephmate:
# Problem

When using a file that does not end in json/yml/yaml/toml, you get the below error:

All supported extensions are hardcoded here: [https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/lib.rs#L705](https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/lib.rs#L705)

```
mirrord exec \
  --config-file valid \
  -- ls
  x mirrord exec
    ✓ running on latest (3.144.0)!
Error:   × Failed to verify mirrord config: mirrord-config: Unsupported configuration file format!
  ╰─▶ mirrord-config: Unsupported configuration file format!
  help: Inspect your config file and arguments provided.    let mut cfg_context = ConfigContext::default();
```

# Solution

Error message should say

```
mirrord-config: you provided file with extension {extension} but mirrord supports only json, toml, yaml, and yaml
```

PR #3375, opened 2025-06-19
feat: improve error message for unsupported config file extensions

abhinav-1305:
resolves: #3371

### DIFF
--- a/mirrord/config/src/config.rs
+++ b/mirrord/config/src/config.rs
@@ -46,6 +46,9 @@ pub enum ConfigError {
     #[error("mirrord-config: Unsupported configuration file format!")]
     UnsupportedFormat,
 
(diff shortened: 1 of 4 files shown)


mirrord/config/src/config.rs
```diff
@@ -46,6 +46,9 @@ pub enum ConfigError {
     #[error("mirrord-config: Unsupported configuration file format!")]
     UnsupportedFormat,
```
josephmate: This became dead code with change. You could re-use the existing one or delete it.

josephmate (approved):

mirrord/config/src/config.rs
abhinav-1305: removed it

abhinav-1305:
please review - @0x00A5.

PR #3384, opened 2025-06-24
Async copy target creation

Razz4780:


mirrord/operator/src/client.rs
```diff
@@ -584,18 +584,21 @@ impl OperatorApi<PreparedClientCert> {
         };
 
         let (session, reused_copy) = if do_copy_target {
-            let mut copy_subtask = progress.subtask("copying target");
+            let mut copy_subtask = progress.subtask("preparing target copy");
             if let Some(reason) = reason {
                 copy_subtask.info(&format!(
-                    "Creating a target copy for {reason} (even though `copy_target` was not explicitly set)."
+                    "The copy target feature is used for {reason} (even though `copy_target` was not explicitly set)."
```
meowjesty, line 590: Nit: I wonder if these messages need to be changed.

mirrord/operator/src/crd/copy_target.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct CopyTargetStatus {
+    /// The session object of the original operator session that created this CopyTarget.
+    pub creator_session: Session,
+    /// Current phase of the copy.
+    ///
+    /// Either `InProgress`, `Ready`, or `Failed`.
+    /// Stored as a string for some future compatibility.
+    ///
+    /// Not filled by older operator versions.
+    pub phase: Option<String>,
```
meowjesty, line 43: Don't want to deal with?

```rs
enum {
   ...
   Unknown
}
```

meowjesty (approved):
Looks ok to me, except for the javascriptification of using strings rather than enum.

mirrord/operator/src/crd/copy_target.rs
Razz4780: Yeah but then with `serde(other)` you loose the value :<

mirrord/operator/src/client.rs
Razz4780: Well, at this point we don't know if we will be creating or reusing the copy :shrug:

mirrord/operator/src/crd/copy_target.rs
meowjesty: I'm ok with the strings, I don't see them becoming a problem in this case.

mirrord/operator/src/client.rs
meowjesty: Good point. I came at it from the angle of like, I don't care if it's being created reused, but I guess the new messages are more on point on what's going on.

PR #3387, opened 2025-06-24
Skip priority class test

0x00A5:


tests/src/targetless.rs
```diff
@@ -38,10 +38,12 @@ mod targetless_tests {
     }
 
     /// Test spawning a targetless agent pod with a given priority class.
-    #[ignore]
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     pub async fn targetless_agent_with_priority_class(#[future] kube_client: Client) {
+        if !matches!(std::env::var("MIRRORD_E2E").as_deref(), Ok("true")) {
+            return;
+        }
```
Razz4780: Oooh, I have an idea. You can use the kube client to check if the operator is installed in the cluster ^^
We don't need the env var. Wdyt?

tests/src/targetless.rs
```diff
@@ -38,10 +38,12 @@ mod targetless_tests {
     }
 
     /// Test spawning a targetless agent pod with a given priority class.
-    #[ignore]
```
Razz4780, line 41: This test should have `#[cfg(not(feature = "targetless"), ignore)]`. Right now it runs a couple times in the CI

tests/src/targetless.rs
0x00A5: Yeah, I thought about parsing our operator API group `operator.metalbear.co`, but liked your idea of using env var (cuz why not it is so simple :D). I can switch to checking through kube client.

Razz4780: You can reuse (I mean copy-paste) the code we have in the operator client for this purpose ^^

```rust
pub async fn operator_installed(client: &Client) -> kube::Result<bool> {
    let gvk = GroupVersionKind {
        group: MirrordOperatorCrd::group(&()).into_owned(),
        version: MirrordOperatorCrd::version(&()).into_owned(),
        kind: MirrordOperatorCrd::kind(&()).into_owned(),
    };

    match discovery::oneshot::pinned_kind(client, &gvk).await {
        Ok(..) => Ok(true),
        Err(kube::Error::Api(response)) if response.code == 404 => Ok(false),
        Err(error) => Err(error),
    }
}
```

0x00A5: Switched to checking operator group.

tests/src/targetless.rs
0x00A5: Updated

Razz4780 (approved):

PR #3388, opened 2025-06-25
Mbe 776 decouple credentialstore from kubers and add more complex

BigLullu:
decoupled Client from kube.rs in CredentialsStore and Credentials and 
added unit test leveraging on achieved decoupling

### DIFF
--- a/mirrord/auth/src/cluster_api.rs
+++ b/mirrord/auth/src/cluster_api.rs
@@ -16,7 +16,6 @@ pub trait AuthClient<R> {
         key_pair: &KeyPair,
         common_name: &str,
     ) -> impl Future<Output = Result<Certificate, CredentialStoreError>> + Send;
-
 }
 
 impl<R> AuthClient<R> for Client
--- a/mirrord/auth/src/cluster_api.rs
+++ b/mirrord/auth/src/cluster_api.rs
@@ -63,3 +63,46 @@ fn certificate_request(
 
     builder.create_certificate_signing_request(key_pair)
 }
+
+#[cfg(test)]
+pub mod client_mock {
+    use std::fmt::Debug;
+
+    use kube::Resource;
+    use serde::Deserialize;
+
+    use crate::{
+        certificate::Certificate, error::CredentialStoreError, key_pair::KeyPair, AuthClient,
+    };
+
+    #[derive(Clone)]
+    pub struct ClientMock {
+        pub return_error: bool,
+    }
+
+    impl<R> AuthClient<R> for ClientMock
+    where
+        R: for<'de> Deserialize<'de> + Resource + Clone + Debug,
+        <R as Resource>::DynamicType: Default,
+    {
+        async fn obtain_certificate(
+            &self,
+            _key_pair: &KeyPair,
+            _common_name: &str,
+        ) -> Result<Certificate, CredentialStoreError> {
+            if self.return_error {
+                Err(CredentialStoreError::FileAccess(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "Mock Error",
+                )))
+            } else {
+                Ok(certificate_mock())
+            }
+        }
+    }
+
+    pub fn certificate_mock() -> Certificate {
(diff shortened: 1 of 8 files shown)


mirrord/auth/src/lib.rs
```diff
@@ -27,3 +27,5 @@ pub mod credentials;
 pub mod error;
 /// Public/Private key abstraction for serialization and deserialization
 pub mod key_pair;
+
+pub use cluster_api::AuthClient;
```
meowjesty, line 32: nit: both `cluster_api` and `AuthClient` are already `pub` right? so why re-export?

mirrord/auth/src/credentials.rs
```diff
@@ -194,20 +175,22 @@ impl DateValidityExt for rfc5280::Validity {
 /// Extenstion of Credentials for functions that accesses Operator
 #[cfg(feature = "client")]
 pub mod client {
-    use kube::{api::PostParams, Api, Client, Resource};
+    use kube::Resource;
 
+    //use kube::{api::PostParams, Api, Client, Resource};
```
meowjesty: Can be removed.

mirrord/auth/src/cluster_api.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+
+use crate::{certificate::Certificate, error::CredentialStoreError, key_pair::KeyPair};
+
+pub trait AuthClient<R> {
+    fn obtain_certificate(
+        &self,
+        key_pair: &KeyPair,
+        common_name: &str,
+    ) -> impl Future<Output = Result<Certificate, CredentialStoreError>> + Send;
```
meowjesty, line 19: Don't we have `async fn` in traits?

mirrord/auth/src/cluster_api.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl<R> AuthClient<R> for Client
+where
+    R: for<'de> Deserialize<'de> + Resource + Clone + Debug,
+    <R as Resource>::DynamicType: Default,
+{
+    fn obtain_certificate(
```
meowjesty, line 28: I think `async` would simplify this impl, or am I misunderstanding something here?

mirrord/auth/src/cluster_api.rs
```diff
@@ -0,0 +1,108 @@
...
+use kube::{api::PostParams, Api, Client, Resource};
+use serde::Deserialize;
+#[cfg(feature = "client")]
+use x509_certificate::{
+    rfc2986, InMemorySigningKeyPair, X509CertificateBuilder, X509CertificateError,
+};
+
+use crate::{certificate::Certificate, error::CredentialStoreError, key_pair::KeyPair};
+
+pub trait AuthClient<R> {
```
meowjesty, line 14: This trait is part of our `default-features`, while the `Client` impl is part of the `client` feature, right? If I'm understanding this correctly, then I think `AuthClient` should be moved up (to some upper mod, which I guess it's `lib.rs`), and we could feature gate this whole `mod` with `client`?

meowjesty: I don't want to complicate things though (not a big fan of having `features` like this), so do we care @Razz4780 ?

mirrord/auth/src/cluster_api.rs
BigLullu: Here based on what I've seen in [intproxy](https://github.com/metalbear-co/mirrord/blob/c3ef9a915c34d7790849a8750196a7f22166a3a7/mirrord/intproxy/src/background_tasks.rs#L107-L123) the code seems to be aligned with the project, that part at least (though I personally have some reservations about async in traits).
I completely agree with the point about complexity. In this case, I'd say the complexity isn’t particularly high, but we don’t really have a way to measure it objectively, so… xD.
It does tend to hide complexity, though: one for all, the fact that two futures cannot share the same type to the compiler, can only belong to the same family(like closures).
Just sharing thoughts, of course :)

mirrord/auth/src/lib.rs
BigLullu: I'd hide cluster_api here n reexport AuthClient since is a pub trait, do you agree ?

mirrord/auth/src/cluster_api.rs
meowjesty: I think it's like that in the `intproxy` mostly due to it being a more unstable thing at the time, and we've been trying to avoid adding features so one day we can get back on stable rust (a bit of mirrord project background/backstory here).

About complexity, yeah it's not high, but the implementation we have wouldn't it look more like a normal function with `async`? Like the error handling could be done with just `?` instead of `match`, and the `async move` part be dropped completely?

```rs
let request = certificate_request(common_name, key_pair)?;
let certificate = request.encode_pem()?;
let api: Api<R> = Api::all(self.clone());

Ok(api.create_subresource("certificate", "operator", &PostParams::default(), certificate_request.into()).await?)
```

meowjesty: I wouldn't say this is a true blocker, we have not talked about it yet, so if you think `impl Future` would be better, then maybe we can bring it up in the next sync and come to an agreement with the team?

mirrord/auth/src/cluster_api.rs
Razz4780: If it's used only by code hidden behind the `client` feature, it should also be behind this feature

mirrord/auth/src/cluster_api.rs
Razz4780: This is the correct syntax. You could write
```rust
trait Whatever {
    async fn something();
}
```
But then you can implement this trait it in a way that makes the future not `Send`, and clippy complains (because the trait is mostly unusable in async code).
While when you use the `fn name() -> impl Future<...> + ...` syntax, you can add trait bounds on the returned future.

meowjesty: Can't we use https://crates.io/crates/async-trait for this then?

mirrord/auth/src/lib.rs
meowjesty: Sorry, somehow missed this comment.

Sure.

PR #3395, opened 2025-06-26
3.147.0

aviramha:


meowjesty (approved):
:+1: 

Gotta run the config schema updating too I think?

aviramha:
> 👍
> 
> Gotta run the config schema updating too I think?

nah it was by accident.

PR #3396, opened 2025-06-27
Make outgoing network docs clearer

gememma:


mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -70,7 +70,9 @@ pub enum OutgoingFilterConfig {
 /// See the outgoing [reference](https://metalbear.co/mirrord/docs/reference/traffic/#outgoing) for more
 /// details.
 ///
-/// The `remote` and `local` config for this feature are **mutually** exclusive.
+/// You can use either the `remote` or `local` value alone to tunnel all outgoing traffic at once,
```
eyalb181:
```suggestion
/// You can use either the `remote` or `local` value alone to turn outgoing traffic tunneling on or off,
```

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -70,7 +70,9 @@ pub enum OutgoingFilterConfig {
...
+/// or specify individual filters for each. The `remote` and `local` config for this feature are
```
eyalb181:
```suggestion
/// or use more fine-grained configuration. The `remote` and `local` config for this feature are
```

mirrord/config/src/feature/network/outgoing.rs
eyalb181: Lets change the structure a bit: the remote/local text, then a small example json, then "you can also use more fine-grained configuration", then the json we have now

eyalb181 (changes requested):

mirrord/config/src/feature/network/outgoing.rs
```diff
@@ -70,7 +70,20 @@ pub enum OutgoingFilterConfig {
...
+/// {
+///   "feature": {
+///     "network": {
+///       "outgoing": "remote"
+///     }
+///   }
+/// }
+/// ```
+///
+/// Alternatively, you can use more fine-grained configuration. The `remote` and `local` config for
```
eyalb181, line 85: The sentence about remote/local being mutually exclusive belongs to the previous paragraph

eyalb181 (changes requested):

mirrord/config/src/feature/network/outgoing.rs
gememma: does it? surely its obvious you can only use one or the other in the top example, but its the lists of filters in the complex one that are mutually exclusive

PR #3397, opened 2025-06-27
Added startup probe to E2E HTTP server containers

Razz4780:


tests/src/utils/cluster_resource.rs
```diff
@@ -48,6 +48,9 @@ pub(super) fn deployment_from_json(name: &str, image: &str, env: Value) -> Deplo
                                 }
                             ],
                             "env": env,
+                            "tcpSocket": {
```
t4lz: Isn't this missing `readinessProbe` or something like that?

Razz4780: ah shiet, right

tests/src/utils/cluster_resource.rs
```diff
@@ -48,6 +48,9 @@ pub(super) fn deployment_from_json(name: &str, image: &str, env: Value) -> Deplo
...
+                                "port": 80,
+                            }
```
t4lz: I think you meant this?
```suggestion
                            "startupProbe": {
                                "tcpSocket": {
                                    "port": 80,
                                }
                            }
```

tests/src/utils/cluster_resource.rs
Razz4780: Fixed

t4lz (approved):

t4lz (approved):

PR #3406, opened 2025-07-03
Create MirrordSession CRD

DmitryDodzin:


mirrord/operator/Cargo.toml
```diff
@@ -48,6 +48,7 @@ crd = [
   "dep:thiserror",
 ]
 setup = ["crd", "dep:serde_yaml"]
+experimental = []
```
Razz4780: Why do we need a feature for this? It's going to be enabled in the operator anyway. I'd rather not create a staging branch for this :V

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,37 @@
...
+    version = "v1alpha",
+    kind = "MirrordSession"
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordSessionSpec {
+    /// Owner of this session
+    pub owner: MirrordSessionOwner,
+
+    /// Start time of this session when actual websocket connection is first created.
+    pub start_time: DateTime<Utc>,
```
Razz4780: This is essentially [`ObjectMeta::creation_timestamp`](https://docs.rs/k8s-openapi/latest/k8s_openapi/apimachinery/pkg/apis/meta/v1/struct.ObjectMeta.html#structfield.creation_timestamp), no?

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// Max duration for this session
+    pub max_time: Option<Duration>,
```
Razz4780: I don't think this is necessary here, operator api servers could enforce this based on the start time and their config

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,37 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct MirrordSessionOwner {
+    /// Unique ID.
+    pub user_id: String,
+    /// Creator local username.
+    pub username: String,
+    /// Creator hostname.
+    pub hostname: String,
+    /// Creator Kubernetes username.
+    pub k8s_username: String,
```
Razz4780, line 34: I think we should store full [`UserInfo`](https://docs.rs/k8s-openapi/latest/k8s_openapi/api/authentication/v1/struct.UserInfo.html) here. For subject access reviews and such

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,37 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[kube(
+    group = "mirrord.metalbear.co",
+    version = "v1alpha",
+    kind = "MirrordSession"
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordSessionSpec {
```
Razz4780: We'll need more stuff here for sure:
1. Session namespace (always present)
2. Session target (none when targetless)
3. Copy target info
4. Queue splitting info
5. Optional branch name for the jira integration

For the target, I suggest:
```rust
struct SessionTarget {
    api_version: String,
    kind: String,
    name: String,
    container: String,
}
```
for easy forward compatibility

mirrord/operator/Cargo.toml
DmitryDodzin: I want this to be compile flag and not a staging branch, Though I started by entirely omitting it and making it only available when the same flag is specified in operator, after a bit if trial and error the wholsale omitment was a bit too much duplicated code so I changed the operator to always include the new code but not create the actual crd in etcd but just to use the definition in memory like the old object. I can change this to `option_env!` for the setup only but I though about keeping the same as operator just for consistency, WDYT?

mirrord/operator/src/crd/session.rs
DmitryDodzin: Yes pretty much, though I wasn't sure I want to use it for logic because it is optional (and will be none if we don't actually create the instance in etcd)

mirrord/operator/src/crd/session.rs
DmitryDodzin: I wasn't sure about it, main concern is just not to "leak" information (even if it should be internal). We still should be getting the headers to create the `RemoteUser` object on each api call as before and take that as "source of truth" and maybe invalidate the session if it's not correct information there

mirrord/operator/src/crd/session.rs
DmitryDodzin: I added most except queue splitting and copy-target but I think those should be via status subresource and ownerReferance

mirrord/operator/Cargo.toml
Razz4780: I mean, we can just not use any feature/compile-time flag at all. Right now on the operator side it's just creating/deleting one extra CR, it shouldn't affect anything :shrug: 

And when we have some statefuls session feature (like agent reuse), we'll probably need a runtime switch for it, so again - compile-time switch seems unnecessary

mirrord/operator/src/crd/session.rs
Razz4780: I'd remove it and use the creation timestamp tbh. If the resource is in etcd, the timestamp will be there. If it's just backed by our API, we can fill it as well

mirrord/operator/src/crd/session.rs
Razz4780: Ok, yeah. I was thinking we might need to check access when reconciling this CR, but I'm not sure. Maybe we only need to reconcile with the initial user request, so we can use headers

mirrord/operator/src/crd/session.rs
Razz4780: Using owner references is going to be a bit problematic, because we'll need configurable TTL for both splits and copied pods

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,71 @@
...
```
Razz4780:
```suggestion
#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
#[kube(
    group = "mirrord.metalbear.co",
    version = "v1alpha",
    kind = "MirrordClusterSession"
)]
#[serde(rename_all = "camelCase")]
pub struct MirrordClusterSessionSpec {
```

Following someone else's request for consistency :sweat_smile:

mirrord/operator/src/crd/session.rs
```diff
@@ -0,0 +1,71 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct JiraMetricsResources {
+    /// The Jira webhook URL, used to update total session time in the mirrord Jira app
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub jira_webhook_url: Option<String>,
+
+    /// The user's current git branch, used for sending session metrics to mirrord Jira app
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub branch_name: Option<String>,
+}
```
Razz4780:
```suggestion
/// Resources needed to report session metrics to the mirrord Jira app
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
#[serde(rename_all = "camelCase")]
pub struct JiraMetricsResources {
    /// The Jira webhook URL, used to update total session time in the mirrord Jira app.
    pub jira_webhook_url: String,

    /// The user's current git branch, used for sending session metrics to mirrord Jira app.
    pub branch: String,
}
```

We can make the field on `MirrordSesssionSpec` optional instead, would be cleaner imho

mirrord/cli/src/config.rs
```diff
@@ -750,6 +750,9 @@ pub(super) struct OperatorSetupParams {
     /// are targeted with `scale_down` feature enabled.
     #[arg(long, default_value_t = false)]
     pub(super) application_auto_pause: bool,
+
+    #[arg(long = "experimental-statefull-sessions", default_value_t = false)]
+    pub(super) statefull_sessions: bool,
```
Razz4780:
```suggestion
    pub(super) stateful_sessions: bool,
```
typo

Razz4780 (changes requested):
I think lost ones

mirrord/operator/src/crd/session.rs
Razz4780: Ping on this :eyes:

PR #3415, opened 2025-07-10
Removed sleeps, sending messages in batches

Razz4780:


tests/rust-sqs-printer/src/main.rs
```diff
@@ -19,16 +22,39 @@ const QUEUE_JSON_ENV_VAR: &str = "SQS_TEST_Q_JSON";
 const QUEUE1_NAME_KEY: &str = "queue1_name";
 const QUEUE2_URL_KEY: &str = "queue2_url";
 
+/// In a loop, retries the future produced by the given function,
+/// until it succeeds or returns a fatal error.
+///
+/// Some errors are retried with backoff. These are errors that can be triggered by agent loss,
+/// which can happen when the agent is restarting the target workload.
+async fn retry_with_backoff<F, Fut, T, E, R>(mut f: F) -> Result<T, SdkError<E, R>>
```
gememma, line 30: no maximum number of retries? you're confident this wont loop forever?

Razz4780: It surely will. But this app is only used in a test, and the test has timeouts. And the app is killed when `process::Child` struct is dropped.

PR #3419, opened 2025-07-11
Automatically detect iptables backend in the agent

Razz4780:


mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+/// Whether [`get_iptables`] should use nftables when no backend is explicitly configured.
+///
+/// Initialized with the first [`get_iptables`] call when the `nftables` argument is not provided.
+const IPTABLES_BACKEND_NFTABLES: OnceLock<bool> = OnceLock::new();
+
+/// Returns correct [`IPTablesWrapper`] to use for traffic redirection.
+///
+/// If `nftables` argument is not provided, this function will choose between legacy and nftables:
+/// 1. First, kernel support will be checked by adding a dummy rule that will not affect traffic.
+/// 2. Second, if multiple backeds are supported, both will be checked for existing mesh rules. If a
```
gememma:
```suggestion
/// 2. Second, if multiple backends are supported, both will be checked for existing mesh rules. If a
```

mirrord/config/configuration.md
```diff
@@ -513,6 +514,12 @@ one bound as host.
 mirrord Experimental features.
 This shouldn't be used unless someone from MetalBear/mirrord tells you to.
 
+### _experimental_ browser_extension_config {#experimental-browser_extension_config}
+
+mirrord will either open a URL for initiating mirrord browser extension to
```
gememma: i dont think this sentence makes sense 👀 mirrord will either [open URL] or... what?

gememma (commented):
SPAG 🍝

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
     }
 }
 
+/// Whether [`get_iptables`] should use nftables when no backend is explicitly configured.
+///
+/// Initialized with the first [`get_iptables`] call when the `nftables` argument is not provided.
+static IPTABLES_BACKEND_NFTABLES: OnceLock<bool> = OnceLock::new();
```
meowjesty:
```suggestion
static MIRRORD_AGENT_IPTABLES_BACKEND_NFTABLES: OnceLock<bool> = OnceLock::new();
```

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+                        %error,
+                        command = wrapper.cmd,
+                        "Failed to add a dummy rule using one of the iptables binaries, \
+                        assuming no kernel support for it."
+                    )
+                })
+                .is_err();
+            if failed {
+                return None;
+            }
```
meowjesty:
```suggestion
                })
                .ok()?;
```

Can't we do this since `filter_map` returns `Option`?

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+
+            let _ = wrapper.delete("filter", "INPUT", "-p 255 -j DROP")
+                .inspect_err(|error| {
+                    tracing::error!(
+                        %error,
+                        command = wrapper.cmd,
+                        "Failed to delete a dummy rule added when checking kernel support for one of the iptables binaries. \
+                        The rule should not any affect traffic."
```
meowjesty:
```suggestion
                        The rule should not affect any traffic."
```

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+                    )
+                });
+
+            Some(IPTablesWrapper::from(wrapper))
+        })
+        .collect::<Vec<_>>();
+
+    if with_kernel_support.len() > 1 {
```
meowjesty, line 378: Why check `len()`, wouldn't the loop be skipped if this had `0` elements?

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+                        assuming no mesh rules.",
+                    );
+                })
+                .ok()
+                .flatten();
+
+            if mesh.is_some() {
+                let _ = IPTABLES_BACKEND_NFTABLES.set(wrapper.tables.cmd.ends_with("-nft"));
+                return Ok(wrapper.clone());
+            }
```
meowjesty:
```suggestion
            if let Some(_mesh) = MeshVendor::detect(wrapper)
                .inspect_err(|error| {
                    tracing::warn!(
                        %error,
                        "Failed to detect mesh rules with one of the iptables binaries, \
                        assuming no mesh rules.",
                    );
                })
                .ok()
                .flatten() {
                let _ = IPTABLES_BACKEND_NFTABLES.set(wrapper.tables.cmd.ends_with("-nft"));
                return Ok(wrapper.clone());
            }
```

mirrord/agent/iptables/src/lib.rs
```diff
@@ -325,6 +305,111 @@ where
...
+        .pop()
+        .inspect(|wrapper| {
+            let _ = IPTABLES_BACKEND_NFTABLES.set(wrapper.tables.cmd.ends_with("-nft"));
+        })
+        .ok_or(
+            "neither iptables-legacy nor iptables-nft seems to be supported by the kernel, \
+            agent will not be able to redirect incoming traffic",
+        )
+        .map_err(From::from)
+        .map_err(IPTablesError)
```
meowjesty, line 408: Weird that you need both `From` and then convert to `IPTablesError` 👀 .

mirrord/config/configuration.md
Razz4780: :shrug: I just ran medschool

mirrord/agent/iptables/src/lib.rs
Razz4780: Yep, wonders of using `Box<dyn ...>`

mirrord/agent/iptables/src/lib.rs
Razz4780: Because I want to run mesh detection only when we have more than 1 available backend

mirrord/config/configuration.md
Razz4780: Fixed

mirrord/agent/iptables/src/lib.rs
Razz4780: Why? :eyes: 
It's a very local static. I moved it to inside the function, so there's no confusion

mirrord/agent/iptables/src/lib.rs
Razz4780: Changed to a more readable `match`

DmitryDodzin (approved):
LGTM

PR #3425, opened 2025-07-15
Passthrough traffic mirroring

Razz4780:


mirrord/agent/src/incoming/connection/optional_broadcast.rs
```diff
@@ -0,0 +1,64 @@
...
+use hyper::body::Frame;
+use mirrord_protocol::tcp::InternalHttpBodyFrame;
+use tokio::sync::broadcast;
+
+use crate::incoming::{connection::copy_bidirectional::CowBytes, IncomingStreamItem};
+
+/// Utility wrapper over an optiona [`broadcast::Sender`].
+///
+/// 1. The sender is dropped as soon as we detect that there are no receivers left. This allows for
+///    dropping the whole channel as soon as its no longer needed.
```
0x00A5:
```suggestion
/// Utility wrapper over an optional [`broadcast::Sender`].
///
/// 1. The sender is dropped as soon as we detect that there are no receivers left. This allows for
///    dropping the whole channel as soon as it's no longer needed.
```

mirrord/agent/src/incoming/connection/copy_bidirectional.rs
```diff
@@ -0,0 +1,179 @@
...
+/// # Params
+///
+/// * `incoming` - an incoming data stream
+/// * `outgoing` - an outgoing data destination (e.g. a stealing client or a passthrough connection)
+/// * `ready_incoming_data` - incoming data that was already received and is ready to be sent to the
+///   outgoing destination
+pub async fn copy_bidirectional<I, O>(
+    incoming: &mut I,
+    outgoing: &mut O,
+    ready_incoming_data: Bytes,
```
0x00A5: nit: I think it would be neat if we can drop this additional already read incoming data, and use a rolled back stream as the `incoming` from where this function is called.

mirrord/agent/src/incoming/mirror_handle.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        self.mirrored_ports
+            .insert(port, StreamNotifyClose::new(ReceiverStream::new(rx)));
+
+        Ok(())
+    }
+
+    /// Stops mirroring the given port.
+    ///
+    /// If this port is not mirrored, does nothing.
```
0x00A5, line 75: nit: can you add a short doc that mentions dropping the rx leads to the tx closed which then leads to that spawned task sending a dead channel internal message that gets handled by the redirector task. Same as in the stealer handle. These two functions look simple but a lot happen when they are called :D

0x00A5 (approved):

mirrord/agent/src/incoming/mirror_handle.rs
Razz4780: I added a couple docs here and there. I wouldn't document the whole flow here, because I suspect the doc would go out of date with the first change to `RedirectorTask` (separate module, separate file, easy to miss)

PR #3427, opened 2025-07-17
Add e2e test for passthrough mirroring and stealing

Razz4780:


tests/src/http.rs
```diff
@@ -1,69 +1,282 @@
...
+use crate::utils::{
+    application::Application, kube_client, port_forwarder::PortForwarder, services::basic_service,
+};
+
+async fn make_http_conn(portforwarder: &PortForwarder) -> SendRequest<String> {
+    let tcp_conn = TcpStream::connect(portforwarder.address()).await.unwrap();
+    let (sender, conn) = hyper::client::conn::http1::Builder::new()
+        .handshake::<_, String>(TokioIo::new(tcp_conn))
+        .await
+        .unwrap();
```
gememma: might be nice to replace some of these `.unwrap()` with `.expect("super duper helpful msg")`, wdyt? 👀

gememma (approved):
🪞✨

tests/src/http.rs
Razz4780: right, done

PR #3428, opened 2025-07-17
Add logging to a file while patching SIP

gememma:
### sample output

```
[1752748605] (pid 22090, binary: /Users/gem/sdk/go1.24.1/pkg/tool/darwin_arm64/compile, args: Some(["/Users/gem/sdk/go1.24.1/bin/go", "build", "-o", "./application", "./main.go"])) SIP Status: Ok(SipBinary("/Users/gem/sdk/go1.24.1/pkg/tool/darwin_arm64/compile")), layer load type: None
[1752748605] (pid 22090) SIP patch result: patched successfully
```

### DIFF
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -318,6 +318,10 @@ impl MirrordExecution {
                             .map(|x| x.to_vec())
                             .unwrap_or_default(),
                         skip: &config.skip_sip,
+                        log_destination: config
+                            .experimental
+                            .sip_log_destination
+                            .as_ref(),
                     },
                 )
                 .transpose() // We transpose twice to propagate a possible error out of this
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -309,14 +309,15 @@ impl MirrordExecution {
         );
 
         #[cfg(target_os = "macos")]
-        let log_info = match config.experimental.sip_log_destination.as_ref() {
-            None => None,
-            Some(log_destination) => Some(mirrord_sip::SipLogInfo {
+        let log_info = config
+            .experimental
+            .sip_log_destination
+            .as_ref()
+            .map(|log_destination| mirrord_sip::SipLogInfo {
                 log_destination,
                 args,
                 load_type: None,
-            }),
-        };
+            });
 
         #[cfg(target_os = "macos")]
         let patched_path = executable
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
+    ffi::OsString,
     net::SocketAddr,
     time::Duration,
 };
@@ -180,8 +181,9 @@ impl MirrordExecution {
     #[tracing::instrument(level = Level::DEBUG, skip_all, ret, err(level = Level::DEBUG))]
     pub(crate) async fn start_internal<P>(
         config: &mut LayerConfig,
-        // We only need the executable on macos, for SIP handling.
+        // We only need the executable and args on macos, for SIP handling.
         #[cfg(target_os = "macos")] executable: Option<&str>,
+        #[cfg(target_os = "macos")] args: Option<&Vec<OsString>>,
         progress: &mut P,
         analytics: &mut AnalyticsReporter,
     ) -> CliResult<Self>
@@ -311,9 +313,9 @@ impl MirrordExecution {
             None => None,
             Some(log_destination) => Some(mirrord_sip::SipLogInfo {
                 log_destination,
-                args: &args,
-                load_type
-            })
+                args,
+                load_type: None,
+            }),
         };
 
         #[cfg(target_os = "macos")]
@@ -328,12 +330,8 @@ impl MirrordExecution {
                             .map(|x| x.to_vec())
                             .unwrap_or_default(),
                         skip: &config.skip_sip,
-                        // log_destination: config
-                        //     .experimental
-                        //     .sip_log_destination
-                        //     .as_ref(),
                     },
-                    log_info
+                    log_info,
                 )
                 .transpose() // We transpose twice to propagate a possible error out of this
                              // closure.
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -1,6 +1,7 @@
+#[cfg(target_os = "macos")]
+use std::ffi::OsString;
 use std::{
     collections::{HashMap, HashSet},
-    ffi::OsString,
     net::SocketAddr,
     time::Duration,
 };
@@ -183,7 +184,7 @@ impl MirrordExecution {
         config: &mut LayerConfig,
         // We only need the executable and args on macos, for SIP handling.
         #[cfg(target_os = "macos")] executable: Option<&str>,
-        #[cfg(target_os = "macos")] args: Option<&Vec<OsString>>,
+        #[cfg(target_os = "macos")] args: Option<&[OsString]>,
         progress: &mut P,
         analytics: &mut AnalyticsReporter,
     ) -> CliResult<Self>
--- a/mirrord/cli/src/execution.rs
+++ b/mirrord/cli/src/execution.rs
@@ -307,6 +307,16 @@ impl MirrordExecution {
         );
 
         #[cfg(target_os = "macos")]
+        let log_info = match config.experimental.sip_log_destination.as_ref() {
+            None => None,
+            Some(log_destination) => Some(mirrord_sip::SipLogInfo {
+                log_destination,
+                args: &args,
+                load_type
+            })
+        };
+
+        #[cfg(target_os = "macos")]
         let patched_path = executable
             .and_then(|exe| {
                 sip_patch(
@@ -318,11 +328,12 @@ impl MirrordExecution {
                             .map(|x| x.to_vec())
                             .unwrap_or_default(),
                         skip: &config.skip_sip,
-                        log_destination: config
-                            .experimental
-                            .sip_log_destination
-                            .as_ref(),
+                        // log_destination: config
+                        //     .experimental
+                        //     .sip_log_destination
+                        //     .as_ref(),
                     },
+                    log_info
                 )
                 .transpose() // We transpose twice to propagate a possible error out of this
                              // closure.
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -239,13 +239,11 @@
 #![deny(unused_crate_dependencies)]
 
 use std::{
-    collections::HashMap,
-    env::vars,
-    ffi::{CString, OsString},
-    net::SocketAddr,
-    os::unix::ffi::OsStrExt,
+    collections::HashMap, env::vars, ffi::CString, net::SocketAddr, os::unix::ffi::OsStrExt,
     time::Duration,
 };
+#[cfg(target_os = "macos")]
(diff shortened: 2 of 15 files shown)


mirrord/sip/src/lib.rs
```diff
@@ -138,6 +142,16 @@ mod main {
...
+    /// Info for logging to `config.experimental.sip_log_destination`
+    pub struct SipLogInfo<'a> {
+        /// The file destination to write logs to
+        pub log_destination: &'a PathBuf,
+        /// Args to the binary, if available
+        pub args: Option<&'a Vec<OsString>>,
+        /// The load type, if available
+        pub load_type: Option<&'a str>,
+    }
+
```
Razz4780:
```suggestion
    /// Info for logging to `config.experimental.sip_log_destination`
    pub struct SipLogInfo<'a> {
        /// The file destination to write logs to
        pub log_destination: &'a Path,
        /// Args to the binary, if available
        pub args: Option<&'a [&'a str]>,
        /// The load type, if available
        pub load_type: Option<&'a str>,
    }

```

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +648,79 @@ mod main {
         Ok(output)
     }
 
+    /// Write a log to the log file, if the file exists, in a thread-safe way
+    fn write_start_log_to_file(
+        file: &mut Option<File>,
+        binary_path: &str,
+        status: &Result<SipStatus>,
+        log_info: &SipLogInfo,
+    ) -> Result<()> {
```
Razz4780:
```suggestion
    fn try_write_start_log_to_file(
        file: &mut File,
        binary_path: &str,
        status: &Result<SipStatus>,
        log_info: &SipLogInfo,
    ) {
```
We don't want these new logs to introduce additional failure paths to the layer, better swallow the errors early (and write them to stderr)

mirrord/sip/src/lib.rs
```diff
@@ -671,7 +750,16 @@ mod main {
                 // just its valid flow.
                 Ok(None)
             }
-        }
+        };
+        let result_as_str = match &patch_result {
+            Ok(Some(_)) => "patched successfully",
+            Ok(None) => "no patching occurred due to lack of SIP or error getting SIP status",
+            Err(_) => "error while patching",
+        };
```
Razz4780: You can write the result's debug to file, I think. We're loosing the error here

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +648,79 @@ mod main {
...
+                log_info.args,
+                log_info.load_type
+            )
+            .map_err(|error| SipError::IO(error))?;
+            file.unlock().map_err(|error| SipError::IO(error))?;
+        };
+        Ok(())
+    }
+
+    fn write_result_to_file(file: &mut Option<File>, result: &str) -> Result<()> {
```
Razz4780:
```suggestion
    fn try_write_result_to_file(file: &mut File, result: &str) {
```
Same as above, also you can pass the original `Result` via reference

mirrord/cli/src/execution.rs
```diff
@@ -180,8 +181,9 @@ impl MirrordExecution {
     #[tracing::instrument(level = Level::DEBUG, skip_all, ret, err(level = Level::DEBUG))]
     pub(crate) async fn start_internal<P>(
         config: &mut LayerConfig,
-        // We only need the executable on macos, for SIP handling.
+        // We only need the executable and args on macos, for SIP handling.
         #[cfg(target_os = "macos")] executable: Option<&str>,
+        #[cfg(target_os = "macos")] args: Option<&Vec<OsString>>,
```
Razz4780:
```suggestion
        #[cfg(target_os = "macos")] args: Option<&[OsString]>,
```

mirrord/cli/src/main.rs
```diff
@@ -322,6 +326,14 @@ where
         &mut config,
         #[cfg(target_os = "macos")]
         Some(&args.binary),
+        #[cfg(target_os = "macos")]
+        Some(
+            &args
+                .binary_args
+                .iter()
+                .map(|str| str.parse::<OsString>().unwrap())
```
Razz4780: Scary unwrap. I think you can just convert the `str` to bytes and then make `OsString` from bytes
https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStringExt.html#tymethod.from_vec

mirrord/layer/src/exec_utils.rs
```diff
@@ -70,12 +70,31 @@ pub(super) fn patch_if_sip(path: &str) -> Detour<String> {
     let skip_patch_binaries = SKIP_PATCH_BINARIES
         .get()
         .expect("skip patch binaries not set");
+    // use args for logging only, so no need to fail if they're not set
+    let args = EXECUTABLE_ARGS
+        .get()
+        .map(|exec_args| exec_args.args.clone())
+        .unwrap_or_default();
```
Razz4780:
```suggestion
    let args = EXECUTABLE_ARGS
        .get()
        .map(|exec_args| exec_args.args.clone());
```
args are optional in `SipLogInfo`

mirrord/sip/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(iter_intersperse)]
+#![feature(file_lock)]
```
Razz4780: What's this? :eyes:

mirrord/sip/src/lib.rs
gememma: why change the `File` to not `Option`al?

mirrord/sip/src/lib.rs
gememma: so that i can use `file.unlock()`- from the compiler:
```
error: a method with this name may be added to the standard library in the future
   --> mirrord/sip/src/lib.rs:671:18
    |
671 |             file.unlock().map_err(SipError::IO)?;
    |                  ^^^^^^
    |
    = warning: once this associated item is added to the standard library, the ambiguity may cause an error or change in behavior!
    = note: for more information, see issue #48919 <https://github.com/rust-lang/rust/issues/48919>
    = help: call with fully qualified syntax `fs4::fs_std::FileExt::unlock(...)` to keep using the current method
    = note: `-D unstable-name-collisions` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unstable_name_collisions)]`
help: add `#![feature(file_lock)]` to the crate attributes to enable `std::fs::File::unlock`
    |
6   + #![feature(file_lock)]
    |
```

mirrord/sip/src/lib.rs
Razz4780: Dunno, I looks kinda weird. It's essentially `if let Some(...) = file` somewhere, I'd instinctively do it outside of the function, to make the function itself simpler

mirrord/sip/src/lib.rs
Razz4780: Ah, so you're no longer using `fs4` this way. I think there's some pushback on using more unstable std features.
So I suggest using the qualified syntax `FileExt::unlock(file)`

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+        opts: SipPatchOptions,
+        log_info: Option<SipLogInfo>,
+    ) -> Result<Option<String>> {
+        // set up logging to a file if present in config
+        let mut log_file = if let Some(log_info) = &log_info {
+            let output_file = OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(log_info.log_destination)?;
+            Some(output_file)
```
Razz4780:
```suggestion
            OpenOptions::new()
                .create(true)
                .append(true)
                .inspect_err(|error| eprintln!("Fail to open SIP logfile: {error}"))
                .ok()
```

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+
+        let status = get_sip_status(binary_path, opts);
+        // if log_info.is_some() && log_file.is_some() {
+        //     try_write_start_log_to_file(
+        //         log_file.as_mut().unwrap(),
+        //         binary_path,
+        //         &status,
+        //         &log_info.unwrap(),
+        //     );
+        // }
```
Razz4780: :wastebasket:

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+        });
+        let _ = writeln!(
+            file,
+            "[{}] (pid {}) SIP patch result: {result:?}",
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("")
+                .as_secs(),
+            std::process::id()
+        )
```
Razz4780:
```suggestion
        let _ = writeln!(
            file,
            "[{}] (pid {}) SIP patch result: {result:?}",
            SystemTime::now()
                .duration_since(UNIX_EPOCH)
                .expect("")
                .as_secs(),
            std::process::id()
        )
```
It should be ok to `write!` `Debug` impl of the result, it should not allocate

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+    /// Write a log to the log file, if the file exists, in a thread-safe way
+    fn try_write_start_log_to_file(
+        file: &mut File,
+        binary_path: &str,
+        status: &Result<SipStatus>,
+        log_info: &SipLogInfo,
+    ) {
+        let _ = file
+            .lock_exclusive()
+            .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
```
Razz4780:
```suggestion
            .map_err(|error| eprintln!("Failed to lock SIP logfile: {error}"));
```

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+                .duration_since(UNIX_EPOCH)
+                .expect("")
+                .as_secs(),
+            std::process::id(),
+            log_info.args,
+            log_info.load_type
+        )
+            .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
+        let _ = FileExt::unlock(file)
+            .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
```
Razz4780:
```suggestion
            .map_err(|error| eprintln!("Failed to unlock SIP logfile: {error}"));
```

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+    }
+
+    fn try_write_result_to_file(file: &mut File, result: &Result<Option<String>>) {
+        let _ = file
+            .lock_exclusive()
+            .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
```
Razz4780:
```suggestion
            .map_err(|error| eprintln!("Failed to lock SIP logfile: {error}"));
```

mirrord/sip/src/lib.rs
```diff
@@ -634,14 +647,93 @@ mod main {
...
+            "[{}] (pid {}) SIP patch result: {result:?}",
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("")
+                .as_secs(),
+            std::process::id()
+        )
+        .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
+        let _ = FileExt::unlock(file)
+            .map_err(|error| eprintln!("Couldn't log SIP status to file: {error}"));
```
Razz4780:
```suggestion
            .map_err(|error| eprintln!("Failed to unlock SIP logfile: {error}"));
```

PR #3429, opened 2025-07-18
Agent communication port now uses SO_REUSEADDR flag

aviramha:
Agent communication port now uses SO_REUSEADDR flag, fixing cases where agent port is reused in a fast consecutive manner and fails

### DIFF
--- a/mirrord/agent/src/entrypoint.rs
+++ b/mirrord/agent/src/entrypoint.rs
@@ -17,6 +17,7 @@ use metrics::{start_metrics, CLIENT_COUNT};
 use mirrord_agent_env::envs;
 use mirrord_agent_iptables::{error::IPTablesError, SafeIpTables};
 use mirrord_protocol::{ClientMessage, DaemonMessage, GetEnvVarsRequest, LogMessage};
+use socket2::{Domain, SockAddr, Socket, Type};
 use tokio::{
     net::{TcpListener, TcpStream},
     process::Command,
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/entrypoint.rs
```diff
@@ -620,22 +620,33 @@ pub async fn notify_client_about_dirty_iptables(
...
-    let ipv4_listener_result = TcpListener::bind(SocketAddrV4::new(
-        Ipv4Addr::UNSPECIFIED,
-        args.communicate_port,
-    ))
-    .await;
+    // listen for client connections with SO_REUSEADDR
+    use std::net::{TcpListener as StdTcpListener, SocketAddr as StdSocketAddr};
+    use std::os::fd::FromRawFd;
+    use std::os::unix::io::AsRawFd;
+    use socket2::{Socket, Domain, Type, SockAddr};
```
Razz4780: Imports at the top of the file pls

mirrord/agent/src/entrypoint.rs
```diff
@@ -620,22 +620,33 @@ pub async fn notify_client_about_dirty_iptables(
...
+
+    let ipv4_listener_result = (|| {
```
Razz4780: Would be nicer with `tokio::net::TcpSocket`:

```rust
    // Prepares a TCP listener for accepting client connections.
    let setup_listener = |ipv6: bool| -> AgentResult<TcpListener> {
        let (socket, addr) = if ipv6 {
            (
                TcpSocket::new_v6()?,
                SocketAddr::new(Ipv6Addr::UNSPECIFIED.into(), args.communicate_port),
            )
        } else {
            (
                TcpSocket::new_v4()?,
                SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), args.communicate_port),
            )
        };
        // SO_REUSEADDR is required to handle rapid agent restarts.
        socket.set_reuseaddr(true)?;
        socket.bind(addr)?;
        socket.listen(1024).map_err(From::from)
    };

    // listen for client connections
    let ipv4_listener_result = setup_listener(false);
    let listener = if args.ipv6 && ipv4_listener_result.is_err() {
        debug!("IPv6 Support enabled, and IPv4 bind failed, binding IPv6 listener");
        setup_listener(true)
    } else {
        ipv4_listener_result
    }?;
```

Razz4780 (changes requested):

mirrord/agent/src/entrypoint.rs
aviramha: thanks!

Razz4780 (approved):

PR #3434, opened 2025-07-21
Add mirrord newsletter command and periodic invitations to the newsletter

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1034/oss-redirect-to-newsletter-via-cli-mirrord-newsletter

### DIFF
--- a/tests/src/utils/run_command.rs
+++ b/tests/src/utils/run_command.rs
@@ -70,6 +70,7 @@ pub async fn run_exec(
     // The packets arrive to loopback interface.
     base_env.insert("MIRRORD_AGENT_NETWORK_INTERFACE", "lo");
     base_env.insert("RUST_LOG", "warn,mirrord=debug");
+    base_env.insert("MIRRORD_PROGRESS_MODE", "off");
 
     if let Some(env) = env {
         for (key, value) in env {


mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Attempts to open the mirrord newsletter sign-up page in the default browser.
+/// In case of failure, prints the link.
+pub async fn newsletter_command() {
+    // open URL with param utm_source=newslttrcmd
+    let url = format!("{NEWSLETTER_SIGNUP_URL}cmd");
+    match get_open_command().arg(&url).output().await {
+        Ok(output) if output.status.success() => {}
+        other => {
+            tracing::trace!("failed to open browser, command result: {other:?}");
```
meowjesty:
```suggestion
            tracing::trace!(?other, "failed to open browser");
```

`tracing` also supports this thing, which I think looks nicer (up to you though).

mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+            .await
+            .map_err(DataStoreError::FileAccess)?;
+        Ok(())
+    }
+}
+
+/// Errors from [`UserData`] operations.
+#[derive(Debug, Error)]
+pub enum DataStoreError {
+    #[error("failed to parent directory for data store file: {0}")]
```
meowjesty: Failed to make a 👶 📁 ?

mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+        (
+            NEWSLETTER_INVITE_THIRD,
+            "Looks like you're doing some serious work with mirrord!".to_string(),
+        ),
+    ]);
+
+    let current_sessions = bump_session_count().await;
+    if let Some(message) = newsletter_invites.get(&current_sessions) {
+        // print the chosen invite to the user
+        println!("{}", get_full_message(message, current_sessions));
```
meowjesty: Nit, but why a separate function for formatting this?

mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+    ParentDir(std::io::Error),
+
+    #[error("IO on data store file failed: {0}")]
+    FileAccess(std::io::Error),
+
+    #[error("failed to serialize/deserialize data: {0}")]
+    Json(serde_json::Error),
+}
+
+#[cfg(not(target_os = "macos"))]
```
meowjesty: will this work on windows?

Cargo.toml
```diff
@@ -183,6 +183,9 @@ x509-parser = "0.17"
 # Used by `agent`, `auth`, `tls-util`, `tests`
 pem = "3"
 
+# Used by `cli`, `auth`
+home = "0.5"
```
meowjesty, line 187: 🏡

mirrord/cli/src/newsletter.rs
meowjesty: Why `print`? Shouldn't we use `progress`? The user should not be seeing this if they have turned `progress=off`, right?

mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+            "Failed to update number of previous mirrord runs."
+        );
+        return 0;
+    }
+    new_data.prev_runs
+}
+
+#[derive(Default, Debug, Serialize, Deserialize)]
+struct UserData {
+    prev_runs: u32,
```
meowjesty:
```suggestion
    session_count: u32,
```

mirrord/cli/src/newsletter.rs
```diff
@@ -0,0 +1,193 @@
...
+            .write_all(contents.as_slice())
+            .await
+            .map_err(DataStoreError::FileAccess)?;
+        Ok(())
+    }
+}
+
+/// Errors from [`UserData`] operations.
+#[derive(Debug, Error)]
+pub enum DataStoreError {
```
meowjesty: Can't believe I'll be the one saying this but, we won't ever handle these errors right? They're just ignored/logged.

I wonder if we even need them, I would just `Box<dyn Error>` and not bother giving them a better context. If you really want to give them some extra context, then maybe use `anyhow` (seems like overkill to me though)?

meowjesty (changes requested):
I think the main request here is the question of: shouldn't we use `progress` instead of `print` when this is being run from `mirrord`? The standalone command is fine with `print`.

mirrord/cli/src/newsletter.rs
gememma: didn't even occur to me 👀 nice, and you're right about `progress=off` UX

mirrord/cli/src/newsletter.rs
gememma: 🔥🪟🔨🔥

mirrord/cli/src/newsletter.rs
gememma: good point 👀 the rusty instincts possessed me to make an Error enum but just boxing the error is better

mirrord/cli/src/newsletter.rs
gememma: leftover from when I was handling these messages in a far more overkill-ish way 😅

Cargo.toml
gememma: 🌉

mirrord/cli/src/newsletter.rs
gememma: maybe they can use IVF? (in-vitro filesystem)

mirrord/cli/src/newsletter.rs
gememma: i forget Every Time 🤦

mirrord/cli/src/newsletter.rs
gememma: ah. the problem with using progress is that it doesn't print the message at the end of all the other progress stuff, it prints like this: 
```
* Running command: curl py-serv:80
* [other config details...]

Join thousands of devs using mirrord!
>> To subscribe to the mirrord newsletter, run:
>> mirrord newsletter
>> or sign up here: https://metalbear.co/newsletter?utm_medium=cli&utm_source=newsletter5

⠓ mirrord exec
    ✓ running on latest (3.152.0)!
    ✓ ready to launch process
      ✓ layer extracted
      ✓ [etc. etc.]
    ✓ config summary 
```
I want the newsletter invite to be as visible as possible, so it's beter at the end. I'll change it to only print if `progress=on` though 👍

meowjesty: Hmm, maybe we could implement in the progress some sort of buffer where you can queue up messages? And in the `Drop` impl we could print the messages stored in this buffer.

wdyt?

gememma: i mean i _could_, but is it necessary? seems dramatic 🎭

meowjesty: It's nicer than having to remember to check the progress env var. Plus we could use it in the future to queue up other types of messages.

Implementing this would be just like, adding a new method to the trait, right? Since you ended up implementing the message and logic in the plugins instead of relying on the cli, then you would need to implement this new method only for the cli printer.

gememma: after some investigation, `Progress` does not seem to support this as it exists now (printing on `Drop`, printing during `success()`) - I'm going to say this is out of scope and open an issue to add a message buffer to `Progress` which will probably involve some serious refactoring

PR #3437, opened 2025-07-22
Fix lingering agent

Razz4780:
`TcpStealerTask` was lingering forever


gememma (commented):
Looks good, but I think having a test category called "misc" is a slippery slope - it could definitely because a dumping ground for random stuff. Maybe that test could go under a category like "ephemeral" or "cleanup" or something?

PR #3439, opened 2025-07-22
Fix typo in istio ambient message

scottillogical:
Fix typo in "! mirrord detected an ambient Istio service mesh butthe agent is not configured"


0x00A5:
Thanks for the PR! Could you please add a news fragment file under `changelog.d/`. The file could be named as something like `+istio-msg-typo.fixed.md`. This is unfortunately needed for the CI to pass.

scottillogical:
test failure but it has to be intermittent?

PR #3440, opened 2025-07-23
Enable passthrough mirroring by default in OSS

Razz4780:


mirrord/config/src/agent.rs
```diff
@@ -421,9 +421,9 @@ pub struct AgentConfig {
     ///
     /// When this is set, `network_interface` setting is ignored.
     ///
-    /// Defaults to `false`.
-    #[config(default = false)]
-    pub passthrough_mirroring: bool,
+    /// Defaults to `false` in mirrord for Teams.
```
gememma, line 423: nit: I might be tempted to change the language of the start of this doc to _avoid_ calling it "an alternative implementation", I think it's confusing to say something is alternative but also enabled by default

mirrord/config/src/lib.rs
```diff
@@ -430,12 +430,17 @@ impl LayerConfig {
...
+        let mut config = if let Ok(path) = context.get_env(Self::FILE_PATH_ENV) {
             LayerFileConfig::from_path(path)?.generate_config(context)
         } else {
             LayerFileConfig::default().generate_config(context)
         }?;
 
+        // OSS-specific adjustment
+        //
+        // `agent.passthrough_mirroring` is enabled by default in OSS.
+        config.agent.passthrough_mirroring = config.agent.passthrough_mirroring.or(Some(true));
```
gememma, line 442: does this not get called in mfT ?

Razz4780: Nope, `LayerConfig` is only for the mirrord user. In mfT we generate only the agent config

mirrord/config/src/agent.rs
Razz4780: Done

mirrord/config/src/lib.rs
gememma: huh, ok 🫡

PR #3442, opened 2025-07-23
Trigger operator e2e on PR approval, and wait for its result.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/146
- Related PRs:
  - https://github.com/metalbear-co/operator/pull/983

### DIFF
--- a/changelog.d/+146.changed.md
+++ b/changelog.d/+146.changed.md
@@ -1,0 +1,1 @@
+Use return-dispatch action to trigger operator e2e from mirrord.


.github/workflows/operator-e2e.yaml
```diff
@@ -4,32 +4,60 @@ on:
...
-      - name: Trigger operator E2E workflow
-        uses: metalbear-co/repository-dispatch@main
+      - name: Generate unique run ID
+        id: run-id
+        run: echo "run_id=mirrord-e2e-$(date +%s)-${{ github.run_id }}" >> $GITHUB_OUTPUT
+
+      - name: Trigger operator E2E workflow and wait for completion
+        id: trigger-e2e
+        uses: metalbear-co/return-dispatch@main
+        continue-on-error: true
```
Razz4780, line 41: What's this for?

Razz4780 (commented):
What makes it wait for the result now?

.github/workflows/operator-e2e.yaml
meowjesty: The `return-dispatch` action https://github.com/marketplace/actions/return-dispatch should keep waiting until the triggered action in the other repo is either done, or the timeout explodes.

We `continue-on-error` in this PR cause we want to see if it actually works first, and not have it show red Xs to our other pals.

PR #3443, opened 2025-07-23
Add queue filter policy in policy CRD

0x00A5:


mirrord/operator/src/crd/policy.rs
```diff
@@ -248,6 +254,58 @@ pub struct HttpFilterPolicy {
     pub header_filter: Option<String>,
 }
 
+/// Split queues policy that defines requirements for mirrord `feature.split_queues` config.
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct SplitQueuesPolicy {
+    /// When set to true, require message filter being used in split queues.
```
t4lz:
```suggestion
    /// When set to true, require a message filter to be used in split queues.
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -248,6 +254,58 @@ pub struct HttpFilterPolicy {
...
+    #[serde(rename = "Kafka")]
+    Kafka,
+}
+
+/// Defines either one or all of the queue filter rules need to be met.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum QueueFilterRules {
+    AllOf(Vec<QueueFilterRule>),
+    AnyOf(Vec<QueueFilterRule>),
```
t4lz: Since changing CRDs can be a bit of a pain, let's have support for recursive conditions from the get-go
Something like:
```Rust
pub enum Condition {
    AllOf(Vec<Condition>),
    AnyOf(Vec<Condition>),
    Single(QueueFilterRule),
}
```
This is just to explain what I mean, the enum can look different, the point is, we should be able to have complex conditions that combine AllOf and AnyOf.

After this is implemented, please also share some simple examples for review of what rules defined by users will look like in yaml.

PR #3445, opened 2025-07-24
MBE-1302 intproxy compiles on windows

itsamegraf:

### DIFF
--- a/changelog.d/+too-many-files-error.fixed.md
+++ b/changelog.d/+too-many-files-error.fixed.md
@@ -1,0 +1,1 @@
+Fixed a bug where SIP patching would discard the error when too many files are open and cause problems with layer injection.
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -39,12 +39,14 @@ mod rpath;
 /// - A shebang is added to scripts without one in order to point it to the patched binary.
 /// - When checking a script, only the first line of the file is checked for a shebang, in case the
 ///   script is encoded unusually.
+/// - Some errors may prevent logs from printing out, making it harder to diagnose the issue. In
+///   this case, use the `experimental.sip_log_destination` config to print SIP logs to a file.
 mod main {
     use std::{
         env,
         ffi::{OsStr, OsString},
         fs::{File, OpenOptions},
-        io::{self, BufRead, Write},
+        io::{self, BufRead, ErrorKind, Write},
         os::{macos::fs::MetadataExt, unix::fs::PermissionsExt},
         path::{Path, PathBuf},
         str::from_utf8,
@@ -62,7 +64,7 @@ mod main {
         Architecture, Endianness, FileKind,
     };
     use once_cell::sync::Lazy;
-    use tracing::{trace, warn};
+    use tracing::{debug, error, trace, warn};
     use which::which;
 
     use super::*;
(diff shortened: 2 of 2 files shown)


mirrord/agent/env/src/checked_env.rs
```diff
@@ -4,10 +4,16 @@ use std::{
     fmt,
     marker::PhantomData,
     net::{AddrParseError, IpAddr, SocketAddr},
-    os::unix::ffi::OsStrExt,
-    str::{FromStr, Utf8Error},
+    str::{FromStr, Utf8Error}
```
itsamegraf:
```suggestion
    str::{FromStr, Utf8Error},
```

mirrord/agent/env/src/checked_env.rs
```diff
@@ -4,10 +4,16 @@ use std::{
...
-    os::unix::ffi::OsStrExt,
     str::{FromStr, Utf8Error},
 };
 
+#[cfg(not(windows))]
+use std::os::unix::ffi::OsStrExt;
+
+#[cfg(windows)]
+#[allow(unused_imports)]
+use std::os::windows::ffi::OsStrExt;
```
cristeigabriela: is it even needed? it's common for some extensions to be needed on unix but not on windows.

[windows::ffi::OsStrExt](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStrExt.html) only implements `encode_wide`

mirrord/config/src/config/context.rs
```diff
@@ -3,9 +3,16 @@ use std::{
...
     ops::Not,
-    os::unix::ffi::OsStrExt,
 };
 
+#[cfg(not(windows))]
+use std::os::unix::ffi::OsStrExt;
+
+#[cfg(windows)]
+#[allow(unused_imports)]
+use std::os::windows::ffi::OsStrExt;
```
cristeigabriela: ditto

mirrord/config/src/config/context.rs
```diff
@@ -94,17 +101,23 @@ impl ConfigContext {
     ///
     /// This is the only way we should read environment when generating or verifying configuration.
     pub fn get_env(&self, name: &str) -> Result<String, VarError> {
+        #[cfg(not(windows))]
```
cristeigabriela, line 98:
```
gabrielaelae@attolia:~/dev/review$ cat test.rs
use std::ffi::OsStr;

fn get_os_str(name: &str) -> &OsStr {
    OsStr::new(name)
}

fn main() {
    const TEST: &str = "Hello, world!";

    println!("{:?}", get_os_str(TEST));
}
gabrielaelae@attolia:~/dev/review$ rustc test.rs
gabrielaelae@attolia:~/dev/review$ ./test
"Hello, world!"
```

seems like we don't need any cfg-ing here, could be the same? tested on ubuntu

mirrord/intproxy/Cargo.toml
```diff
@@ -38,11 +38,15 @@ hyper = { workspace = true, features = ["client", "http1", "http2"] }
 hyper-util.workspace = true
 http-body-util.workspace = true
 bytes.workspace = true
-rand.workspace = true
 tokio-rustls.workspace = true
 rustls.workspace = true
 tokio-retry.workspace = true
 
+[target.'cfg(target_os = "linux")'.dependencies]
```
cristeigabriela: could be `cfg(all(target_os = "linux", target_os = "macos"))` if i'm not mistaken?

mirrord/protocol/src/file.rs
```diff
@@ -96,6 +103,31 @@ impl From<Metadata> for MetadataInternal {
     }
 }
 
+
+#[cfg(windows)]
+impl From<Metadata> for MetadataInternal {
+
```
cristeigabriela: please fix formatting

mirrord/protocol/src/file.rs
```diff
@@ -96,6 +103,31 @@ impl From<Metadata> for MetadataInternal {
...
 
+
+#[cfg(windows)]
+impl From<Metadata> for MetadataInternal {
+    
+    fn from(metadata: Metadata) -> Self {
+        // let file_id = file_id::get_low_res_file_id(metadata.path()).unwrap();
+        Self {
+            device_id: metadata.volume_serial_number().unwrap() as u64,
+            inode: metadata.file_index().unwrap(), // On Windows, true inode is not exposed directly from std
```
cristeigabriela: please add comments above the code line

```
// a
// b
// c
d: ...,
```

not

```
a = ..., // a
// b
//c
```

mirrord/intproxy/Cargo.toml
itsamegraf: Image
In chatgpt I trusted 🫤

cristeigabriela: here's an example from me produced with 0 AI overlords involvement
```
[target.'cfg(not(any(target_os = "linux", target_os = "macos")))'.dependencies]
winapi = "0.3.9"
```

(of course it's a bad example, but it shows what i mean)

itsamegraf: apparently I could use unix
image
WDYT?

cristeigabriela: i guess it depends on whether others write it as linux | macos, because there's the BSD boomers, but either is ok with me

mirrord/config/src/config/context.rs
itsamegraf: your example does compile but I would rather change as minimum as possible code flows relevant for linux. the usage of `from_bytes` relies on the fact that it reads the `name` str as if it was utf-8.
On windows, OsStr is utf-16, so I don't mind having OsStr::new for windows, but I don't like implicitly relying on it performing utf-8 initialization on linux. I would rather keep the cfg if youre ok with it

mirrord/agent/env/src/checked_env.rs
itsamegraf: removed 👍

mirrord/intproxy/Cargo.toml
itsamegraf: hehe, I'll go with any(macos, linux), thanks

PR #3446, opened 2025-07-24
Cause too many open files error to stop execution and fail SIP patching

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1070/mirrord-exec-air-mirrord-layer-not-injected-into-air-process-binary


gememma:
> BTW, should we consider automatically increasing ulimit ourselves?

We shouldn't have to, I think Air is a special case - [see discussion here](https://metalbear.slack.com/archives/C08TYT624SV/p1753135259565539) - and I don't know if messing with system resource allocation is a good idea (but I'm willing to be convinced)

PR #3449, opened 2025-07-25
Add machine_id to UserData so it can be used in analytics.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/317
- Related PRs:
  - https://github.com/metalbear-co/analytics-server/pull/41

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3846,8 +3846,7 @@ dependencies = [
 [[package]]
 name = "kube"
 version = "0.99.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a4eb20010536b48abe97fec37d23d43069bcbe9686adcf9932202327bc5ca6e"
+source = "git+https://github.com/metalbear-co/kube.git?rev=c095706801fd0fc287e82bcbc314b458097fe9e1#c095706801fd0fc287e82bcbc314b458097fe9e1"
 dependencies = [
  "k8s-openapi",
  "kube-client",
@@ -3859,8 +3858,7 @@ dependencies = [
 [[package]]
 name = "kube-client"
 version = "0.99.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7fc2ed952042df20d15ac2fe9614d0ec14b6118eab89633985d4b36e688dccf1"
+source = "git+https://github.com/metalbear-co/kube.git?rev=c095706801fd0fc287e82bcbc314b458097fe9e1#c095706801fd0fc287e82bcbc314b458097fe9e1"
 dependencies = [
  "base64 0.22.1",
  "bytes",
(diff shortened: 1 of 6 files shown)


mirrord/cli/src/user_data.rs
```diff
@@ -0,0 +1,136 @@
+use std::{path::PathBuf, sync::LazyLock};
```
Razz4780, line 1: This whole logic needs file locking. We're reading and updating the file, probably from multiple concurrent mirrord invocations

mirrord/cli/src/user_data.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// Overwrite the JSON contents at the default file path (`DATA_STORE_PATH`) with `UserData`
+    pub(crate) async fn overwrite_to_file(&self) -> Result<(), io::Error> {
+        // DATA_STORE_DIR and DATA_STORE_PATH are already known to exist
+        let mut store_file = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .truncate(false)
```
Razz4780:
```suggestion
            .truncate(true)
```
This would be a fun bug

mirrord/analytics/src/lib.rs
```diff
@@ -316,15 +335,13 @@ struct AnalyticsReport {
     error: Option<AnalyticsError>,
 }
 
+const ANALYTICS_ADDRESS: &str = "https://analytics.metalbear.co/api/v1/event";
```
Razz4780:
```suggestion
const ANALYTICS_ENDPOINT: &str = "https://analytics.metalbear.co/api/v1/event";
```

.vscode/launch.json
```diff
@@ -4,6 +4,22 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
+    {
```
Razz4780: Let's just remove this file from the repo and add it to `.gitignore`, shall we? :wastebasket:

Razz4780 (changes requested):

.vscode/launch.json
meowjesty: finally

mirrord/cli/src/user_data.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct UserData {
+    /// Amount of times this user has run mirrord.
+    #[serde(default)]
+    session_count: u32,
+
+    /// Helps us keep track of unique users for analytics when telemetry is enabled.
+    ///
+    /// Must use custom `default =`, since the default is [`Uuid::nil`].
+    #[serde(default = "default_uuid")]
```
Razz4780:
```suggestion
    #[serde(default = "Uuid::new_v4")]
```
This should work, I think

mirrord/cli/src/user_data.rs
```diff
@@ -0,0 +1,141 @@
...
+            .create(true)
+            .truncate(true)
+            .open(DATA_STORE_PATH.as_path())
+            .await?;
+
+        if store_file.try_lock_exclusive()? {
+            let contents = serde_json::to_vec(&self)?;
+            store_file.write_all(contents.as_slice()).await?;
+
+            store_file.unlock_async().await?;
```
Razz4780, line 122: Why do we lock synchronously but unlock asynchronously? :eyes:

Razz4780 (approved):

mirrord/cli/src/user_data.rs
meowjesty: There's no async lock: https://docs.rs/fs4/latest/fs4/tokio/trait.AsyncFileExt.html

The async unlock is just smoke and mirrors (it's just a sync unlock that we can `.await`).

PR #3453, opened 2025-07-28
Update Rust to fresh nightly, and change edition to 2024.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/547

I tried to keep the most annoying crates as separate commits, before running `cargo fix --edition` on the others.

- Will run `cargo fmt` as late as possible, since it changes a lot of files (226 to be precise), which would make this pretty much un-reviewable.

### DIFF
--- a/mirrord/config/derive/src/config/field.rs
+++ b/mirrord/config/derive/src/config/field.rs
@@ -25,11 +25,11 @@ pub struct ConfigField {
 impl ConfigField {
     /// Check if field is `Option<T>` and if so return type of `T`
     fn is_option(field: &Field) -> Option<Type> {
-        let seg = if let Type::Path(ty) = &field.ty {
+        let seg = match &field.ty { Type::Path(ty) => {
             ty.path.segments.first()
-        } else {
+        } _ => {
             None
-        }?;
+        }}?;
 
         (seg.ident == "Option").then(|| match &seg.arguments {
             PathArguments::AngleBracketed(generics) => match generics.args.first() {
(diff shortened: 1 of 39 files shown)


mirrord/config/derive/src/config/field.rs
```diff
@@ -124,7 +124,7 @@ impl ConfigField {
     ///           .source_value().transpose()?
     ///           .ok_or(crate::config::ConfigError::ValueNotProvided("MyConfig", "test", Some("TEST")))?
     /// ```
-    pub fn implmentation(&self, parent: &Ident) -> impl ToTokens {
+    pub fn implmentation(&self, parent: &Ident) -> impl ToTokens + use<> {
```
gememma: is `implementation` supposed to be spelled wrong? 🤔

mirrord/kube/src/api/kubernetes/seeker.rs
```diff
@@ -150,7 +151,7 @@ impl KubeResourceSeeker<'_> {
                 .as_ref()?
                 .containers
                 .iter()
-                .filter(|&container| (!SKIP_NAMES.contains(container.name.as_str())))
+                .filter(|&container| SKIP_NAMES.contains(container.name.as_str()).not())
```
gememma, line 153: 💢

mirrord/config/derive/src/config/field.rs
meowjesty: It's the british spelling.

mirrord/kube/src/api/kubernetes/seeker.rs
meowjesty: Yeah, I also don't think that `(` was needed.

PR #3465, opened 2025-07-31
Added vfork emulation

Razz4780:


mirrord/config/src/experimental.rs
```diff
@@ -105,6 +105,16 @@ pub struct ExperimentalConfig {
...
     #[config(default = None)]
     pub sip_log_destination: Option<PathBuf>,
+
+    /// ### _experimental_ vfork_emulation {#experimental-vfork_emulation}
+    ///
+    /// Enables vfork emulation within the mirrord-layer.
+    /// Might solve rare stack corruption issues.  
+    ///
+    /// Note that for Go applications on ARM64, this feature is not yet supported,
+    /// and this settings is ignored.
```
gememma:
```suggestion
    /// and this setting is ignored.
```

mirrord/layer/src/go/linux_aarch64.rs
```diff
@@ -145,5 +146,9 @@ pub(crate) fn enable_hooks(hook_manager: &mut HookManager) {
         } else {
             trace!("found version < 1.19, arm64 not supported - not hooking");
         }
+
+        if experimental.vfork_emulation {
+            tracing::warn!("vfork emulation is not yet supported for Go programs on ARM64.");
```
gememma, line 151: Do we have an issue for this? 👀

mirrord/layer/src/go/linux_x64.rs
```diff
@@ -512,6 +513,30 @@ unsafe extern "C" fn go_syscall_new_detour() {
     )
 }
 
+/// Detour for `syscall.rawVforkSyscall.abi0` function.
+#[unsafe(naked)]
+unsafe extern "C" fn raw_vfork_detour() {
+    naked_asm!(
```
gememma, line 519: new least favourite macro unlocked! 😀

mirrord/layer/src/lib.rs
```diff
@@ -745,6 +755,111 @@ pub(crate) unsafe extern "C" fn fork_detour() -> pid_t {
...
 }
 
+/// Transparently replaces `vfork` call with a safer `fork`.
+///
+/// `vfork` and `fork` calls are very similar (except performance),
+/// except that with `vfork` the parent process is blocked until the child exits or execs.
+/// This hook uses a dummy pipe to handle this difference:
+/// 1. Parent process creates a pipe with O_CLOEXEC flag.
+/// 2. Parent process forks.
+/// 3. Parent process closes the writing end blocks on reading from the reading end.
```
gememma:
```suggestion
/// 3. Parent process closes the writing end and blocks on reading from the reading end.
```

mirrord/layer/src/lib.rs
```diff
@@ -745,6 +755,111 @@ pub(crate) unsafe extern "C" fn fork_detour() -> pid_t {
...
+                        This is a bug, please report it",
+                    );
+                }
+            }
+        }
+        // fork failed, fallback to vfork.
+        Ordering::Less => {
+            std::mem::drop(pipe_read);
+            std::mem::drop(pipe_write);
+            tracing::error!(
```
gememma, line 851: does this have to be an `error!`? or it could be `warn!` if its probably not going to wreck everything 👀

mirrord/layer/src/go/linux_aarch64.rs
Razz4780: Nope. I didn't want to create a zombie :laughing:

mirrord/layer/src/lib.rs
Razz4780: It's probably never going to happen. And if it does (extreme machine resources issue?), it might break things :v
Like, we might hit the original vfork problem

mirrord/layer/src/go/linux_aarch64.rs
Razz4780: I guess we'll open one when someone complains

gememma: lmao 🧟

gememma (approved):
okay ✨ yay 🍴

PR #3467, opened 2025-07-31
Add rtt metrics to the PingPong (so we can check mirrord<->operator latency).

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/550
- Related PRs:
  - https://github.com/metalbear-co/operator/pull/996

### DIFF
--- a/changelog.d/+996.added.md
+++ b/changelog.d/+996.added.md
@@ -1,0 +1,1 @@
+Add rtt metrics to the PingPong (so we can check mirrord<->operator latency).


mirrord/protocol/src/codec.rs
```diff
@@ -142,6 +145,14 @@ pub enum ClientMessage {
...
     Vpn(ClientVpn),
     GetAddrInfoRequestV2(GetAddrInfoRequestV2),
+    PingRtt(PingRtt),
+}
+
+/// Used by `mirrord` to reply to a [`DaemonMessage::Pong`] so we can calculate RTT between
+/// `mirrord` and the `operator` (the `agent` currently just ignores this).
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct PingRtt {
+    pub elapsed: Duration,
```
Razz4780: Why would you send elapsed time on the wire? :eyes:

meowjesty: Figured it could be used by the agent too in the future. But might be too annoying after looking into the integration tests.

Razz4780: Maybe let's not introduce a new ad hoc message for this?
When the session is targeted, the operator already handles client pings internally, so you can use the existing messages.
You'd only need to adjust the targetless case in the operator (handle pings the same way, client<->operator and operator<->agent separately)

meowjesty: > the operator already handles client pings internally

So this metric is supposed to be operator agent latency? Or do you mean I should do something like:

mirrord sends ping, operator receives it, stores the time, send pong to mirrord, mirrord has to immediately respond to a pong with a new ping, operator gets the new ping and calculates the rtt (this avoids adding a new message, but imo, is pretty dumb).

Or do you have something else in mind?

The nice thing about handling it in mirrord is that we could use it to calculate latency of operator, agent, and a maybe even intproxy too (thanks to the integration tests).

Razz4780: Ah, soz. I didn't read the original issue... we want to measure latency from the operator.
In that case, it should be the operator that sends the ping, not intproxy.
I'd add `DaemonMessage::PingV2(u128)` and `ClientMessage::PongV2(u128)`:
1. If client's mirrord-protocol version supports it, operator periodically sends `DaemonMessage::PingV2` with random id inside
2. intproxy responds with `ClientMessage::PongV2`, using the same id
3. operator remembers time when ping was sent, so it can calculate rtt when receiving pongs

meowjesty: Alright changed! Named the messages to be more operator specific.

Whenever we send a `DaemonMessage::Pong` from the operator to mirrord, we also send (if supported) the `DaemonMessage::OperatorPing`.

mirrord/intproxy/src/lib.rs
```diff
@@ -414,7 +414,11 @@ impl IntProxy {
     }
 
     /// Routes most messages from the agent to the correct background task.
+    ///
     /// Some messages are handled here.
+    ///
+    /// - When adding a new message, remember to add it here, otherwise it'll fall under `other` and
+    ///   error out.
```
Razz4780: Can you remove the catch-all branch instead? As in replace `other` with exact variants. Pretty sure this comment won't help :laughing: Compilation error is safer

mirrord/protocol/src/codec.rs
```diff
@@ -191,6 +192,7 @@ pub enum DaemonMessage {
     PauseTarget(crate::pause::DaemonPauseTarget),
     SwitchProtocolVersionResponse(#[bincode(with_serde)] semver::Version),
     Vpn(ServerVpn),
+    OperatorPing(u128),
```
Razz4780, line 203:
```suggestion
    OperatorPing(
        /// Unique ID of this ping.
        u128,
    ),
```

mirrord/protocol/src/codec.rs
```diff
@@ -142,6 +142,7 @@ pub enum ClientMessage {
     ReadyForLogs,
     Vpn(ClientVpn),
     GetAddrInfoRequestV2(GetAddrInfoRequestV2),
+    OperatorPong(u128),
```
Razz4780, line 148:
```suggestion
    OperatorPong(
        /// ID taken from the [`DaemonMessage::OperatorPing`] message.
        u128,
    ),
```

mirrord/protocol/src/lib.rs
```diff
@@ -89,6 +90,9 @@ pub static VERSION: LazyLock<semver::Version> = LazyLock::new(|| {
         .expect("Bad version parsing")
 });
 
+pub static MIRRORD_OPERATOR_LATENCY_PING_PONG: LazyLock<VersionReq> =
```
Razz4780, line 94:
```suggestion
/// Minimal protocol version that allows for sending [`DaemonMessage::OperatorPing`].
pub static MIRRORD_OPERATOR_LATENCY_PING_PONG: LazyLock<VersionReq> =
```

Razz4780 (changes requested):
Pretty sure that now we also need to handle [`DaemonMessage::OperatorPing`] in the CLI (when fetching env vars and such, before spawning the intproxy)

meowjesty:
> Pretty sure that now we also need to handle [DaemonMessage::OperatorPing] in the CLI (when fetching env vars and such, before spawning the intproxy)

Changed the `match` we have to be explicit everywhere, so I think wherever we get a `OperatorPing` we send a `OperatorPong` now.

Razz4780 (approved):
Yessss

PR #3468, opened 2025-08-01
Fixed `mirrord dump` with passthrough mirroring

Razz4780:


mirrord/cli/src/dump.rs
```diff
@@ -47,76 +64,440 @@ pub async fn dump_command(
...
+        ping_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+        Self {
+            connection,
+            ports,
+            confirmations: Default::default(),
+            ping_interval,
+            queued_messages: Default::default(),
+            conn_id_to_req_id: Default::default(),
         }
```
gememma, line 123: aesthetic nit: can you do some kind of `.. Default::default()` here? 👀 smth like this
```
Self {
    connection,
    ports,
    ping_interval,
    .. Default::default()
}
```

mirrord/cli/src/dump.rs
```diff
@@ -47,76 +64,440 @@ pub async fn dump_command(
...
+        for port in &self.ports {
+            let message = ClientMessage::Tcp(LayerTcp::PortSubscribe(*port));
+            self.connection.sender.send(message).await?;
+            info!("Issued subscription to port {} for mirroring", port);
+        }
+
+        Ok(())
+    }
+
+    fn handle_tcp_message(
```
gememma, line 171: no doc comment? 🥺

mirrord/cli/src/dump.rs
```diff
@@ -47,76 +64,440 @@ pub async fn dump_command(
...
+                    message,
+                }) => {
+                    tracing::warn!("Received log: {message}");
+                }
+                DaemonMessage::LogMessage(LogMessage {
+                    level: LogLevel::Info,
+                    message,
+                }) => {
+                    tracing::info!("Received log: {message}");
+                }
```
gememma: it miiiight be possible to condense these by matching on `DaemonMessage::LogMessage(msg)` and then separately on `msg.level`, but i think it would only save a few newlines 🤷 up to you

gememma (changes requested):
Just aesthetics, logic looks good 🧀 plus a missing doc comment

mirrord/cli/src/dump.rs
Razz4780: This only works when the whole struct implements `Default`. Like, this syntax means `produce Self::default(), and use it to assign values for the rest of the fields`. Not `to each remaining field, assign its default value`

gememma: blegh ok

mirrord/cli/src/dump.rs
Razz4780: Done

gememma (approved):
yay

mirrord/cli/src/dump.rs
Razz4780: //    ______________________________
        //  / \                             \
        // |   |  ADDED ONE!                |
        //  \_ |                            |
        //     |  JUST FOR                  |
        //     |                            |
        //     |  YOU                       |
        //     |   _________________________|___
        //     |  /                            /
        //     \_/____________________________/

PR #3469, opened 2025-08-01
Fixed UserData access

Razz4780:


mirrord/cli/src/user_data.rs
```diff
@@ -88,25 +74,32 @@ impl UserData {
...
+
                 Ok(user_data)
             }
         }
     }
 
     /// Overwrite the JSON contents at the default file path (`DATA_STORE_PATH`) with `UserData`
     pub(crate) async fn overwrite_to_file(&self) -> Result<(), io::Error> {
-        // DATA_STORE_DIR and DATA_STORE_PATH are already known to exist
+        if DATA_STORE_DIR.exists().not() {
```
gememma, line 99: et tu, Brute? 😔

gememma (approved):
🗻

mirrord/cli/src/user_data.rs
Razz4780: image

PR #3470, opened 2025-08-01
Improve mirrord progress (interactive auth, message buffer)

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1308/improve-mirrord-progress


gememma:
Screenshot 2025-08-01 at 15 25 52

Got a timeout flake on go23_dns_lookup test 👀 might have to increase the timeout...

mirrord/progress/src/lib.rs
```diff
@@ -86,6 +103,14 @@ impl Progress for NullProgress {
     fn ide(&self, _: serde_json::Value) {}
 
     fn print(&self, _: &str) {}
+
+    fn add_to_print_buffer(&mut self, _: &str) {}
+
+    fn set_fail_on_drop(&mut self, _: bool) {}
+
+    fn suspend<F: FnOnce() -> R, R>(&self, f: F) -> R {
+        f()
```
meowjesty: Why don't we take this opportunity to implement these `{}` whatever functions in the trait itself, so the `impl` for the progresses (progressi?) can have only what's relevant to them?

mirrord/cli/src/main.rs
```diff
@@ -415,6 +415,12 @@ where
         .map(|(k, v)| CString::new(format!("{k}={v}")))
         .collect::<CliResult<Vec<_>, _>>()?;
 
+    // upon success, print all messages in the progress buffer
+    let buffer = progress.success_and_return_buffer(Some("Ready!"));
```
meowjesty: Only the `SpinnerProgress` actually implements this right?

So why not do something like, in the `success` impl of the `SpinnerProgress`, you add this after the `finish_with_message`? You could remove this function (`success_and_return_buffer`), and just handle this stuff in `success`. wdyt?

mirrord/cli/src/newsletter.rs
```diff
@@ -53,22 +53,17 @@ pub async fn suggest_newsletter_signup(user_data: &mut UserData) {
...
-    // the future.
-    match std::env::var(MIRRORD_PROGRESS_ENV).as_deref().ok() {
-        None | Some("std") | Some("standard") => {
-            if let Some(message) = newsletter_invites.get(&current_sessions) {
-                // print the chosen invite to the user if progress mode is on
-                println!(
-                    "\n\n{}\n>> To subscribe to the mirrord newsletter, run:\n\
+    if let Some(message) = newsletter_invites.get(&current_sessions) {
+        // print the chosen invite to the user if progress mode is on
+        progress.add_to_print_buffer(
```
meowjesty, line 58: yay, no more random env var checking here!

meowjesty (changes requested):
Couple of suggestions.

mirrord/cli/src/main.rs
gememma: Returning something from success would mean returning an empty vec (or None) from success in all the other types of `Progress`, which I like less - our `success` function is a wrapper around the upstream `ProgressBar::success` function, which doesn't return anything. Honestly I quite like that you have to handle the buffer explicitly here, and I think we only want to do buffer printouts with exec anyway (for the moment).

mirrord/progress/src/lib.rs
gememma: I think it would be so easy to accidentally not use functions like that if you didn't notice they were there, plus every time we connect to the cluster in a new way, like using a different method on `Client`, we would have to write more impl in `Progress` - that feels like more wasted lines to me...

mirrord/cli/src/main.rs
meowjesty: My point is, I don't think you need to return anything from `success`? Like, just have the `success` method print whatever is in the buffer, instead of having to remember to call this other function, then "manually" print the buffer. Something like this:

```rs
    fn success(&mut self, msg: Option<&str>) {
        self.done = true;
        if let Some(msg) = msg {
            self.progress.finish_with_message(format!("✓ {msg}"));
        } else {
            self.progress
                .finish_with_message(format!("✓ {}", self.progress.message()));
        }
        if let Some(buffer_vec) = buffer {
            buffer_vec.iter().for_each(|msg| eprintln!("{msg}"))
        }
    }

```

mirrord/progress/src/lib.rs
meowjesty: > I think it would be so easy to accidentally not use functions like that if you didn't notice they were there

We do this in the operator, we also do it in some mirrord crates, it's not an uncommon pattern to impl a trait with some "default" implementation, especially since a bunch of those are like, `{}`.

Take `NullProgress` as an example, it goes from this:

```rs
impl Progress for NullProgress {
    fn subtask(&self, _: &str) -> NullProgress {
        NullProgress
    }

    fn set_fail_on_drop(&mut self, _: bool) {}

    fn success(&mut self, _: Option<&str>) {}

    fn failure(&mut self, _: Option<&str>) {}

    fn warning(&self, _: &str) {}

    fn info(&self, _: &str) {}

    fn ide(&self, _: serde_json::Value) {}

    fn print(&self, _: &str) {}
}
```

To this:

```
impl Progress for NullProgress {
    fn subtask(&self, _: &str) -> NullProgress {
        NullProgress
    }
}
```

> plus every time we connect to the cluster in a new way (...)

I don't get what you mean by this. You mean if we added a new method to `Progress`, like `Progress::print_with_emojis`? Wouldn't that be true for the current implementation we have as well?

mirrord/cli/src/main.rs
gememma: i think i tried that last time 👀 maybe I only tried in Drop - youre right tho this would be much better, ill check

mirrord/progress/src/lib.rs
gememma: Aaah, i understand - i completely misinterpreted the suggestion. Yeah, that should be a nice change

mirrord/progress/src/lib.rs
```diff
@@ -346,11 +292,9 @@ impl Progress for SpinnerProgress {
...
                 .finish_with_message(format!("✓ {}", self.progress.message()));
         }
-    }
-
-    fn success_and_return_buffer(&mut self, msg: Option<&str>) -> Option<Vec<String>> {
-        self.success(msg);
-        Some(self.message_buffer.clone())
+        self.message_buffer
+            .iter()
+            .for_each(|msg| eprintln!("{msg}"));
```
meowjesty: We're using `eprintln` to avoid sending these messages to the IDE, or is there another reason?

Please add a comment on why `eprintln` instead of `println`.

meowjesty (approved):
yay for less empty functions scattered around!

mirrord/progress/src/lib.rs
gememma: the reason was typo 🫥 oops

PR #3478, opened 2025-08-05
Use default container annotation when picking a target container

Razz4780:


mirrord/kube/src/api/runtime.rs
```diff
@@ -167,16 +167,21 @@ impl RuntimeData {
             .as_ref()
             .and_then(|status| status.container_statuses.as_ref())
             .ok_or_else(|| KubeApiError::missing_field(pod, ".status.containerStatuses"))?;
+        let default_container_name = pod
+            .metadata
+            .annotations
+            .as_ref()
+            .and_then(|annotations| {
+                // See https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container.
+                annotations.get("kubectl.kubernetes.io/default-container")
```
BigLullu: nitpick here, I'll put the value inside a constant in case you need to use it multiple times

BigLullu (approved):
Just a point raised bout a nitpick but seems all solid to me ^^

PR #3479, opened 2025-08-05
Fix incorrect parsing of PATH-like `kubeconfig` config file values

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1324/mirrord-treats-kubeconfig-as-single-file-instead-of-list

### DIFF
--- a/mirrord/kube/src/api/kubernetes.rs
+++ b/mirrord/kube/src/api/kubernetes.rs
@@ -334,7 +334,7 @@ pub async fn create_kube_config<P>(
     kube_context: Option<String>,
 ) -> Result<Config>
 where
-    P: AsRef<str> + AsRef<OsStr>,
+    P: AsRef<OsStr>,
 {
     let kube_config_opts = KubeConfigOptions {
         context: kube_context,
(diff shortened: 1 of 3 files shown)


mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,38 @@ pub async fn create_kube_config<P>(
...
         ..Default::default()
     };
 
-    let mut config = match kubeconfig {
-        Some(kubeconfig) => {
-            let kubeconfig = shellexpand::full(&kubeconfig)
+    // parse kubeconfig the same way as KUBECONFIG is parsed by `kube-client`, supporting
+    // colon-separated lists of paths. Borrowed affectionately & with love from
+    // https://docs.rs/kube/latest/kube/config/struct.Kubeconfig.html#method.from_env
+    let mut config = if let Some(kubeconfig) = kubeconfig
```
meowjesty, line 347: omg this `kubeconfig` where everything else is `kube_config`, or `kube_*` is triggering me.

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,38 @@ pub async fn create_kube_config<P>(
     kube_context: Option<String>,
 ) -> Result<Config>
 where
-    P: AsRef<str>,
+    P: AsRef<str> + AsRef<OsStr>,
```
meowjesty:
```suggestion
    P: AsRef<OsStr>,
```

Don't see the point of having both?

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,38 @@ pub async fn create_kube_config<P>(
...
 
-    let mut config = match kubeconfig {
-        Some(kubeconfig) => {
-            let kubeconfig = shellexpand::full(&kubeconfig)
+    // parse kubeconfig the same way as KUBECONFIG is parsed by `kube-client`, supporting
+    // colon-separated lists of paths. Borrowed affectionately & with love from
+    // https://docs.rs/kube/latest/kube/config/struct.Kubeconfig.html#method.from_env
+    let mut config = if let Some(kubeconfig) = kubeconfig
+        && let paths = std::env::split_paths(&kubeconfig)
+            .filter(|p| p.as_os_str().is_empty().not())
```
meowjesty: Since you're going to be using these `p`s as string only, you could use `filter_map` here to convert them here, and not in the `try_fold`.

Suggesting this cause you don't really use `p` as `PathBuf` in the `try_fold`

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,38 @@ pub async fn create_kube_config<P>(
...
+            .collect::<Vec<_>>()
+        && paths.is_empty().not()
+    {
+        let parsed_kube_config = paths.iter().try_fold(Kubeconfig::default(), |m, p| {
```
meowjesty:
```suggestion
        let parsed_kube_config = paths.iter().try_fold(Kubeconfig::default(), |final_kubeconfig, p| {
```

Or something like that, it's not very nice seeing just `m` when the thing is not a one-liner.

mirrord/kube/src/api/kubernetes.rs
gememma: Screenshot 2025-08-05 at 16 22 35
lol

meowjesty: I almost want you to add `kube_config` as an additional config, just so we can deprecate `kubeconfig` in the future.

gememma: *sweating* its uhhhhhhhhhhh out of scope 😬

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,43 @@ pub async fn create_kube_config<P>(
...
-            }
-        }
+    // parse kubeconfig the same way as KUBECONFIG is parsed by `kube-client`, supporting
+    // colon-separated lists of paths. Borrowed affectionately & with love from
+    // https://docs.rs/kube/latest/kube/config/struct.Kubeconfig.html#method.from_env
+    let mut config = if let Some(kubeconfig) = kubeconfig
+        && let paths = std::env::split_paths(&kubeconfig)
+            .filter_map(|p| {
+                let string = p.as_os_str().to_string_lossy().into_owned();
+                string.is_empty().not().then(|| string)
```
meowjesty:
```suggestion
                string.is_empty().not().then_some(string)
```

`then_some` if you're not putting the conversion function inside `then` I guess.

I would suggest the name `path_str` instead of `string`.

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -331,30 +334,43 @@ pub async fn create_kube_config<P>(
...
+            })
+            .collect::<Vec<_>>()
+        && paths.is_empty().not()
+    {
+        let parsed_kube_config =
+            paths
+                .iter()
+                .try_fold(Kubeconfig::default(), |merged_kubeconfig, path| {
```
meowjesty: And maybe `path_str` here too?

PR #3482, opened 2025-08-05
Detect exec into a statically linked binary and issue a warning

Razz4780:


mirrord/cli/src/is_static.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use xmas_elf::{ElfFile, program::Type};
+
+/// Returns true if we can be certain that the binary under the given path is statically linked.
+///
+/// Might yield false negatives.
+///
+/// Note that mirrord *can* possibly be used with static binaries - when the actual application is a
+/// child process of the target binary, and it is dynamically linked.
+/// Therefore, static linking is *not* and indication for a hard fail.
```
meowjesty:
```suggestion
/// Therefore, static linking is *not* an indication for a hard fail.
```

mirrord/cli/src/is_static.rs
```diff
@@ -0,0 +1,44 @@
...
+            );
+            return false;
+        }
+    };
+
+    for header in elf.program_iter() {
+        if let Ok(Type::Dynamic) = header.get_type() {
+            return false;
+        }
+    }
```
meowjesty:
```suggestion
    elf.program_iter().any(|header| matches!(header.get_type(), Ok(Type::Dynamic))).not()
```

mirrord/cli/src/extension.rs
```diff
@@ -63,6 +63,22 @@ pub(crate) async fn extension_exec(
...
+
+        use crate::is_static;
+
+        let is_static = args
+            .executable
+            .as_deref()
+            .map(Path::new)
+            .is_some_and(is_static::is_binary_static);
+        if is_static {
+            progress.warning("Target binary is not dynamically linked. mirrord might not work!");
```
meowjesty:
```suggestion
            progress.warning("Target binary might not be dynamically linked. mirrord might not work!");
```

Since we can have false positives, let's use "might" everywhere.

mirrord/cli/src/main.rs
```diff
@@ -328,6 +330,21 @@ where
...
 
+    #[cfg(target_os = "linux")]
+    {
+        use std::path::Path;
+
+        let mut sub_progress =
+            sub_progress.subtask("checking if target binary is dynamically linked");
+        if is_static::is_binary_static(Path::new(&args.binary)) {
+            sub_progress.failure(Some(
+                "target binary is not dynamically linked, mirrord might not work!",
```
meowjesty:
```suggestion
                "target binary might not be dynamically linked, mirrord might not work!",
```

meowjesty (approved):
I think it would be nice to move the failure messages to some `const` and have them be the same.

PR #3488, opened 2025-08-06
Re export kube

t4lz:
[For the operator to use](https://github.com/metalbear-co/operator/pull/1000) instead of manually keeping the same version over there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -566,6 +566,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compression"
+version = "0.4.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


mirrord/kube/Cargo.toml
```diff
@@ -24,10 +24,39 @@ incluster = []
 portforward = [
     "dep:tokio-retry"
 ]
+# re-export kube. Useful for consumers of this library, as we use
+# types from kube in our public API, therefore the versions must
+# match. So callers can use `mirrord_kube::kube` instead of kube
+# directly, to make sure they use the same kube version.
+re-export-kube = []
```
Razz4780, line 31: We probably don't need this, we can always just have that `pub use kube`

PR #3489, opened 2025-08-06
Hide `mirrord_progress::Progress` implementations behind a crate feature

Razz4780:
Right now dependents unnecessarily pull in `indicatif`, `enum_dispatch` and `serde`

(changed loc come from the fact that code from `lib.rs` was moved to a separate file)


gememma (approved):
🧹 ✨

PR #3492, opened 2025-08-07
Allow users to ignore containers and init containers in copy target config

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1330/allow-user-to-remove-init-containers-in-copy-target

### DIFF
--- a/mirrord/config/src/feature/copy_target.rs
+++ b/mirrord/config/src/feature/copy_target.rs
@@ -20,6 +20,8 @@ pub enum CopyTargetFileConfig {
     Advanced {
         enabled: Option<bool>,
         scale_down: Option<bool>,
+        exclude_containers: Option<Vec<String>>,
+        exclude_init_containers: Option<Vec<String>>,
     },
 }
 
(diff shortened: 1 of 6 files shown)


mirrord/operator/src/crd/copy_target.rs
```diff
@@ -27,6 +27,10 @@ pub struct CopyTargetSpec {
     pub scale_down: bool,
     /// Split queues client side configuration.
     pub split_queues: Option<SplitQueuesConfig>,
+    /// Containers that are ignored by copy target.
```
Razz4780, line 30: Both fields need `#[serde(default)]`, otherwise CLI won't be able to deserialize responses from an old operator

mirrord/config/src/feature/copy_target.rs
```diff
@@ -91,6 +101,16 @@ pub struct CopyTargetConfig {
     ///     }
     /// ```
     pub scale_down: bool,
+
+    /// ### feature.copy_target.exclude_containers {#feature-copy_target-exclude_containers}
```
Razz4780, line 105: Can you update configuration.md with `cargo run --bin medschool -- -i ./mirrord/config/src -o ./mirrord/config/configuration.md`?

Razz4780 (changes requested):

Razz4780 (approved):

PR #3496, opened 2025-08-11
Add MysqlBranchDatabase CRD

0x00A5:
https://linear.app/metalbear/issue/MBE-1336/branch-db-cr-creation


mirrord/operator/src/crd/mysql_branching.rs
```diff
@@ -0,0 +1,74 @@
...
+    Url(ConnectionSourceKind),
+    /// A group of connection parameters.
+    Parameters {
+        host: ConnectionSourceKind,
+        port: Option<ConnectionSourceKind>,
+        username: ConnectionSourceKind,
+        password: ConnectionSourceKind,
+        database: ConnectionSourceKind,
+    },
+}
```
t4lz: I think we also want to support the user specifying a `usernameSource`, a `passwordSource` and a `connectionEnvVar` as per the issue, where the connection url does not contain credentials.

mirrord/operator/src/client.rs
```diff
@@ -1122,9 +1141,53 @@ impl OperatorApi<PreparedClientCert> {
...
+        let mysql_branch_names = create_mysql_params
+            .values()
+            .map(|p| p.name.clone())
+            .collect::<Vec<_>>();
+
+        let reusable_mysql_branches =
+            list_reusable_mysql_branches(&mysql_branch_api, &create_mysql_params, &subtask).await?;
+
+        create_mysql_params.retain(|id, _| !reusable_mysql_branches.contains_key(id));
+        create_mysql_branches(&mysql_branch_api, create_mysql_params, &subtask).await?;
```
t4lz: Is there anything that prevents this scenario:

1. user wants to use branch x.
2. client lists existing branches, x exists.
3. branch x gets deleted (TTL over, or whatever).
4. client doesn't create branch x.
5. client expects x to be ready but it isn't, and won't be.

mirrord/operator/src/client/database_branches.rs
```diff
@@ -0,0 +1,268 @@
...
+/// 1. it has a user specified unique ID, and
+/// 2. it is in the "Ready" phase.
+pub(crate) async fn list_reusable_mysql_branches<P: Progress>(
+    api: &Api<MysqlBranchDatabase>,
+    params: &HashMap<BranchDatabaseId, MysqlBranchParams>,
+    progress: &P,
+) -> Result<HashMap<BranchDatabaseId, MysqlBranchDatabase>, OperatorApiError> {
+    let mut subtask = progress.subtask("listing reusable MySQL branch databases");
+    let label_selector = if params.is_empty() {
+        None
```
t4lz: Doesn't an empty `params` mean at this point that the user doesn't want any branches?
We could consider returning early without making the request.

mirrord/operator/src/setup.rs
t4lz, line 1: still need to add permissions to create and list dbs for the mirrord-user role, right?

mirrord/operator/src/client.rs
0x00A5: It is NOT impossible, but the user needs to aim really well for it to happen, and the controllers need to reconcile right after the user lists and right before the user connects. So the branches are ready here, but later when making the connection they are dead. The result will be connection failure, as the operator tries to patch the branches and encounter k8s api error 404.

mirrord/operator/src/client/database_branches.rs
0x00A5: Yes, should return there. And the else block is wrong too :D. It should check again if the list/iterator is empty after filtering by whether the IDs are specified or not. So we don't end up having a label selector like `label in ()`. I will fix it.

mirrord/operator/src/setup.rs
0x00A5: Yep, I added them in charts: https://github.com/metalbear-co/charts/pull/255
But let me double check if there are places to add them in `setup.rs`.

0x00A5: Added

mirrord/operator/src/crd/mysql_branching.rs
0x00A5: I ended up even removing the `Parameters` variant, since the user only wants everything in one URL env var.

mirrord/config/src/feature/database_branches.rs
```diff
@@ -0,0 +1,143 @@
...
+///       "variable": "DB_CONNECTION_URL"
+///     }
+///   }
+/// }
+/// ```
+#[derive(Clone, Debug, Eq, PartialEq, JsonSchema, Serialize, Deserialize)]
+pub struct DatabaseBranchConfig {
+    /// A unique identifier chosen by the user. This is useful when reusing branch database
+    /// sharing among different Kubernetes users.
+    pub id: Option<String>,
```
t4lz, line 73: In the issue this is called `db_branch_id`. @0x00A5, did you discuss a change with @liron-sel?

mirrord/config/src/feature/database_branches.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    /// Name of the database
+    pub name: String,
```
t4lz: this was not discussed in the issue, right @0x00A5?
Please discuss this with @liron-sel and update the issue accordingly. We should be synced with her about the user interface (config) so that we meet the requirements and specs, and so that she can write the documentation correctly.

mirrord/config/src/feature/database_branches.rs
0x00A5: This one is surfaced here: https://metalbear.slack.com/archives/C0959M0306L/p1755229930240839
I will change what's in the ticket to this.

mirrord/config/src/feature/database_branches.rs
0x00A5: Cool, I will message Liron about this.

0x00A5: Updated in the ticket

mirrord/config/src/feature/database_branches.rs
0x00A5: Updated the ticket

t4lz (approved):

PR #3498, opened 2025-08-11
Fixed incoming HTTP connection shutdown (mirrord-agent)

Razz4780:
Just saw 35MB of "Gracefully shutting down a redirected HTTP connection" logs :skull: 

`CancellationToken::cancelled` immediately returns `Poll::Ready` when the token has already been cancelled, so the task was never returning control to the runtime

### DIFF
--- a/mirrord/agent/src/incoming/task.rs
+++ b/mirrord/agent/src/incoming/task.rs
@@ -218,14 +218,18 @@ where
         let token = port_state.http_shutdown.clone();
         let mut requests = ExtractedRequests::new(TokioIo::new(conn.stream), http_version);
         tokio::spawn(async move {
+            let mut token_cancelled = false;
             loop {
                 let result = tokio::select! {
                     result = requests.next() => result,
-                    _ = token.cancelled() => {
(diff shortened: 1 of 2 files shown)


mirrord/agent/src/incoming/task.rs
```diff
@@ -218,15 +218,19 @@ where
...
                 let result = tokio::select! {
                     result = requests.next() => result,
-                    _ = token.cancelled() => {
+                    _ = token.cancelled(), if shutting_down.not() => {
                         tracing::debug!(
                             connection = ?conn.info,
                             "Gracefully shutting down a redirected HTTP connection",
                         );
+                        // After starting the graceful shutdown,
+                        // `requests` iterator will eventually finish on its own.
```
DmitryDodzin, line 231: Shouldn't we just `break` and drop the iterator as well?

Razz4780: This would abruptly terminate the connection, possibly triggering nasty errors on the client side. Graceful shutdown should be nicer

DmitryDodzin (approved):

PR #3505, opened 2025-08-14
add logfile to the intproxy commandline for easy debugging

aviramha:


mirrord/cli/src/container/sidecar.rs
```diff
@@ -154,14 +154,20 @@ impl IntproxySidecar {
 
         let cleanup = config.container.cli_prevent_cleanup.not().then_some("--rm");
 
+        let mut intproxy_args = vec![&config.container.cli_image, "mirrord", "intproxy"];
+        // Add logfile argument if specified in config
```
gememma: I would get rid of this, it's pretty obvious from the code :D

```suggestion
```

mirrord/cli/src/execution.rs
```diff
@@ -275,6 +275,11 @@ impl MirrordExecution {
             )
             .env(LayerConfig::RESOLVED_CONFIG_ENV, &encoded_config);
 
+        // Add logfile argument if specified in config
```
gememma: Same as above, I don't think you need a comment here
```suggestion
```

changelog.d/+intproxy-debug.added.md
```diff
@@ -0,0 +1 @@
+add logfile to the intproxy commandline for easy debugging
```
gememma:
```suggestion
Added the logfile location to the intproxy command-line so it can be seen in the process output.
```

gememma (approved):
Small things 👌

PR #3506, opened 2025-08-14
Add periodic dump of connected processes for diagnostic purposes

aviramha:


mirrord/intproxy/src/lib.rs
```diff
@@ -200,6 +212,26 @@ impl IntProxy {
...
+
+        tracing::info!("Connected processes ({}): ", count);
+        for (layer_id, process_info) in &self.connected_processes {
+            tracing::info!(
+                layer_id = layer_id.0,
+                pid = process_info.pid,
+                name = %process_info.name,
+                cmdline = ?process_info.cmdline,
+                loaded = process_info.loaded,
+                "Process connected"
```
t4lz: Could be a bit confusing, cause "Process connected" sounds like we're announcing that event just happened. How about:
```suggestion
                "connected processes info"
```

PR #3510, opened 2025-08-19
Fix faulty iptables backend detection in the agent

Razz4780:


mirrord/agent/iptables/src/lib.rs
```diff
@@ -307,105 +307,67 @@ where
...
+    /// backends for the protocols, at the same time).
     ///
-    /// Initialized with the first call of this function, when the `nftables` argument is not
-    /// provided.
-    static DETECTED_NFTABLES: OnceLock<bool> = OnceLock::new();
+    /// Respective [`OnceLock`] is initialized with the first call of this function, if the
+    /// `nftables` argument is not provided.
+    static DETECTED_NFTABLES: (OnceLock<bool>, OnceLock<bool>) = (OnceLock::new(), OnceLock::new());
+    let detected_nftables = if ip6 {
+        &DETECTED_NFTABLES.1
```
DmitryDodzin: I would maybe do 2 static vars instead of .0 and .1 but up to you I just don't like indexed member access

PR #3511, opened 2025-08-19
Added debug args to intproxy and extproxy

Razz4780:


mirrord/cli/src/execution.rs
```diff
@@ -265,6 +265,10 @@ impl MirrordExecution {
             Command::new(std::env::current_exe().map_err(CliError::CliPathError)?);
         proxy_command
             .arg("intproxy")
+            // Start of debug args. Don't add real args after this point,
```
meowjesty, line 268: Ah, so `trailing_var_arg` here will eat up everything, like `mirrord exec --debug-args meow -cats -blabla`?

I don't get how it does it without adding it here though, like, don't we need `.arg("--debug-args")` before we start eating everything?

Razz4780: `_debug_args` is just the name of the field in clap parser. The directives tell clap to eat everything and put it in the vector

Razz4780: The intproxy is called like `mirrord intproxy --log-destination whatever --another-arg-why-not whatever2`

PR #3512, opened 2025-08-19
Improved internal proxy debug logs

Razz4780:


mirrord/intproxy/src/lib.rs
```diff
@@ -280,30 +241,28 @@ impl IntProxy {
...
-                            Err(ProxyFailure::Startup(err))?
-                        }
-                        Err(InternalProxyError::Runtime(err)) => {
-                            tracing::error!(%err, "Manageable error in background task update, proxy is entering in failover state...");
-                            return Err(ProxyFailure::Runtime(FailoverStrategy::from_failed_proxy(proxy, err)))
-                        }
-                        _ => ()
+                    if let Err(error) = proxy.handle_task_update(task_id, task_update).await {
+                        tracing::error!(%error, "Proxy encountered a critical error, and is entering the failover state...");
+                        return ControlFlow::Continue(FailoverStrategy::from_failed_proxy(proxy, error));
```
meowjesty, line 254: Think it would be nice to add some docs here (or at the function declaration) saying why this is `Continue` (it's so we go into the `failover_strategy.run`, right?)

meowjesty (approved):
Just a comment about docs.

PR #3513, opened 2025-08-19
Add method to HTTP filter.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/543

### DIFF
--- a/mirrord/config/configuration.md
+++ b/mirrord/config/configuration.md
@@ -1358,7 +1358,7 @@ case-insensitive.
 ##### feature.network.incoming.http_filter.method_filter {#feature-network-incoming-http-method-filter}
 
 
-Supports standard [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods).
+Supports standard [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods), and non-standard HTTP methods.
 
 Case-insensitive. If the request method matches the filter, the request is stolen.
 
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -1256,7 +1256,7 @@
         },
         "method_filter": {
           "title": "feature.network.incoming.http_filter.method_filter {#feature-network-incoming-http-method-filter}",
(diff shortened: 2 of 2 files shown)


mirrord/protocol/src/tcp.rs
```diff
@@ -430,6 +462,11 @@ pub static HTTP_FILTERED_UPGRADE_VERSION: LazyLock<VersionReq> =
 pub static HTTP_COMPOSITE_FILTER_VERSION: LazyLock<VersionReq> =
     LazyLock::new(|| ">=1.11.0".parse().expect("Bad Identifier"));
 
+// TODO(alex) [high]: Remember to change this!
+/// Minimal mirrord-protocol version that allows [`HttpFilter::Method`]
+pub static HTTP_METHOD_FILTER_VERSION: LazyLock<VersionReq> =
+    LazyLock::new(|| ">=1.11.0".parse().expect("Bad Identifier"));
```
t4lz: should probably bump the protocol version and use the new one here, right?

meowjesty: `1.19.9` or `1.20.0`?

mirrord/config/configuration.md
```diff
@@ -1353,6 +1355,13 @@ Supports regexes validated by the
 The HTTP traffic feature converts the HTTP headers to `HeaderKey: HeaderValue`,
 case-insensitive.
 
+##### feature.network.incoming.http_filter.method_filter {#feature-network-incoming-http-method-filter}
+
+
+Supports standard [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods).
```
t4lz: doesn't it also support non-standard ones because of that `Other` variant?

mirrord/protocol/src/tcp.rs
t4lz: 1.20.0

mirrord/config/configuration.md
meowjesty: Good point. Will add it to the docs.

mirrord/protocol/src/tcp.rs
meowjesty: Bumped.

mirrord/config/configuration.md
meowjesty: Changed.

PR #3518, opened 2025-08-20
3.157.3

meowjesty:


CHANGELOG.md
```diff
@@ -8,6 +8,31 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
 
 <!-- towncrier release notes start -->
 
+## [3.157.3](https://github.com/metalbear-co/mirrord/tree/3.157.3) - 2025-08-20
+
+
+### Added
+
+- Add rtt metrics to the PingPong (so we can check mirrord<->operator latency).
```
gememma: this might be better under internal?

gememma (approved):
🧄

CHANGELOG.md
meowjesty: These metrics are for the people though, not for us.

meowjesty: I'll change the description to something better.

eyalb181: So maybe change the release note to read "users" instead of "we"

eyalb181: Also, since you added new stuff, it should be a minor release, not a patch

meowjesty: Does it make more sense now?

gememma: should be "Added" 🔍 otherwise yeah that's better :)

meowjesty:
https://github.com/metalbear-co/mirrord/pull/3519

PR #3525, opened 2025-08-24
mirrord container: change to always use native arch for intproxy

aviramha:
mirrord container: change to always use native arch for intproxy, improving performance signficantly

### DIFF
--- a/changelog.d/+copy-target-docs.changed.md
+++ b/changelog.d/+copy-target-docs.changed.md
@@ -1,0 +1,1 @@
+Improve copy target schema docs
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -671,20 +671,26 @@
       "additionalProperties": false
     },
     "CopyTargetFileConfig": {
+      "title": "feature.copy_target {#copy_target}",
+      "description": "Allows the user to target a pod created dynamically from the original [`target`](#target). The new pod inherits most of the original target's specification, e.g. labels.\n\nSee the [copy target reference](https://metalbear.co/mirrord/docs/reference/copy-target/) for more details.\n\n### Minimal `copy_target` config {#copy_target-minimal}\n\n```json { \"feature\": { \"copy_target\": true } } ```\n\n### Advanced `copy_target` config {#copy_target-advanced}\n\n```json { \"feature\": { \"copy_target\": { \"enabled\": true, \"scale_down\": true, \"exclude_containers\": [\"my-container\"], \"exclude_init_containers\": [\"my-init-container\"] } } } ```",
       "anyOf": [
         {
+          "description": "Basic configuration that controls whether copy target is enabled (default false).",
           "type": "boolean"
         },
         {
+          "description": "Allows the user to specify both enabling copy target and additional configuration options.",
           "type": "object",
           "properties": {
             "enabled": {
+              "description": "Whether copy target is enabled",
               "type": [
                 "boolean",
                 "null"
               ]
             },
             "exclude_containers": {
+              "description": "List of containers to be ignored by copy_target",
               "type": [
                 "array",
                 "null"
@@ -694,6 +700,7 @@
               }
             },
             "exclude_init_containers": {
+              "description": "List of init containers to be ignored by copy_target",
               "type": [
                 "array",
                 "null"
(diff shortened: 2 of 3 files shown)


mirrord/cli/src/container.rs
```diff
@@ -59,6 +59,29 @@ fn get_mirrord_console_addr() -> Option<String> {
...
+        // Extract architecture from platform string (e.g., "linux/amd64" -> "amd64")
+        let arch = platform.split('/').last().unwrap_or("amd64");
+        let arch_suffix = match arch {
+            "amd64" | "x86_64" => "x86_64",
+            "arm64" | "aarch64" => "aarch64",
+            _ => arch, // fallback to the original arch string
+        };
+
+        // Return architecture-specific path
+        format!("/opt/mirrord/lib/{}/libmirrord_layer.so", arch_suffix)
```
DmitryDodzin: I think it would be better to use `config.container.cli_image_lib_path` and insert the arch_suffix before the last element and not hardcode it

mirrord/cli/src/container.rs
```diff
@@ -59,6 +59,29 @@ fn get_mirrord_console_addr() -> Option<String> {
...
 
+/// Resolves the architecture-specific library path based on the platform configuration.
+///
+/// When a platform is specified (e.g., "linux/amd64"), this function returns the
+/// architecture-specific path within the multi-architecture container.
+/// If no platform is specified, returns the default path.
+fn resolve_library_path(config: &LayerConfig) -> String {
+    if let Some(platform) = &config.container.platform {
+        // Extract architecture from platform string (e.g., "linux/amd64" -> "amd64")
+        let arch = platform.split('/').next_back().unwrap_or("amd64");
```
DmitryDodzin: should we default to `amd64`?

mirrord/cli/Dockerfile
```diff
@@ -26,12 +26,34 @@ RUN cargo +nightly-2025-08-01 zigbuild -p mirrord -p mirrord-layer -Z bindeps --
 RUN cp /build/target/$(cat /.platform)/release/mirrord /mirrord
 RUN cp /build/target/$(cat /.platform)/release/libmirrord_layer.so /libmirrord_layer.so
 
+# For ARM64 builds, also build x86_64 version to support cross-platform containers
+RUN if [ "$TARGETARCH" = "arm64" ]; then \
+    cargo +nightly-2025-08-01 zigbuild -p mirrord-layer -Z bindeps --target x86_64-unknown-linux-gnu --release --locked && \
```
DmitryDodzin, line 31: Also maybe make this symetrical, we should maybe bundle `arm64` with `amd64` target too no? just to cover if x86 computer runs an arm container

aviramha: I think this is too uncommon to care about for now

mirrord/cli/src/container.rs
aviramha: Something feels wrong having such file name, and if we anyway give a way to override, why do we care?

mirrord/cli/src/container.rs
aviramha: what would you recommend instead?

mirrord/cli/src/container.rs
DmitryDodzin: To not be stuck and have a manual override, if for some reason we have a path colistion like we did with the tls certificate

aviramha: But you have a manual override?

mirrord/cli/src/container.rs
DmitryDodzin: I would fallback to `platform` without the split and let the `--platfrom` command print out the error (tough if it's empty I would omit the changes and fallback to just `config.container.cli_image_lib_path`)

mirrord/cli/src/container.rs
aviramha: nvm got it

aviramha: see now

mirrord/cli/src/container.rs
aviramha: good catch!

mirrord/config/src/container.rs
```diff
@@ -51,8 +51,7 @@ pub struct ContainerConfig {
     /// Path of the mirrord-layer lib inside the specified mirrord-cli image.
     ///
     /// Defaults to `"/opt/mirrord/lib/libmirrord_layer.so"`.
```
DmitryDodzin: I would remove the defaults to maybe, but that's optional

PR #3546, opened 2025-09-01
CRD for workload patch (wip)

Razz4780:


mirrord/operator/src/crd/patch.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Represents operator patch state of some workload.
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema, Hash)]
+#[kube(
+    group = "mirrord.metalbear.co",
+    version = "v1alpha",
+    kind = "MirrordClusterWorkloadPatch",
+    printcolumn = r#"{"name":"WORKLOAD-API-VERSION", "type":"string", "description":"API version of the workload.", "jsonPath":".spec.workloadRef.apiVersion"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-KIND", "type":"string", "description":"Kind of the workload.", "jsonPath":".spec.workloadRef.kind"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAME", "type":"string", "description":"Name of the workload.", "jsonPath":".spec.workloadRef.name"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAMESPACE", "type":"string", "description":"API version of the workload.", "jsonPath":".spec.workloadRef.namespace"}"#
```
t4lz:
```suggestion
    printcolumn = r#"{"name":"WORKLOAD-NAMESPACE", "type":"string", "description":"Namespace of the workload.", "jsonPath":".spec.workloadRef.namespace"}"#
```

mirrord/operator/src/crd/patch.rs
```diff
@@ -0,0 +1,45 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct PatchRequest {
+    pub owner: String,
+    pub extra_inline_env: Vec<ExtraInlineEnv>,
+    pub replicas: Option<i32>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct ExtraInlineEnv {
```
t4lz: I find the name `ExtraInlineEnv` a bit confusing. Maybe I'm misunderstanding it?
"Extra" seems a bit confusing because we don't necessarily add a new env var, we replace an existing one if there.
"Inline" - I guess it means "not `valueFrom`"?

I think it could just be called `EnvVar` and if we add support for fancier env vars in the future, we can call them `EnvVarFrom` or whatever.
If you don't want to change the name, then please add some docs so that people can quickly understand an `ExtraInlineEnv` is just a normal env var.

mirrord/operator/src/crd/patch.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub namespace: String,
+    pub name: String,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct PatchRequest {
+    pub owner: String,
+    pub extra_inline_env: Vec<ExtraInlineEnv>,
+    pub replicas: Option<i32>,
```
t4lz: please add some short docs for these fields, I they're not self-explanatory for people who don't already know all our patching stuff.

mirrord/operator/src/crd/patch.rs
Razz4780: Oke, renamed to `EnvVar`

mirrord/operator/src/crd/patch.rs
Razz4780: Added

t4lz (approved):

Razz4780:
> The operator's cluster role should get permissions for the new CRDs in operator setup.

Added :+1:

PR #3550, opened 2025-09-02
Fix iptables detection on kernel 6.12+

Razz4780:


mirrord/agent/iptables/src/lib.rs
```diff
@@ -346,33 +368,94 @@ pub fn get_iptables(nftables: Option<bool>, ip6: bool) -> IPTablesWrapper {
...
     );
 
     wrapper
 }
 
+/// Drops [`Capability::CAP_SYS_MODULE`] from the current thread.
+///
+/// This will prevent the thread from loading kernel modules.
+fn try_drop_cap_sys_module() {
+    let has_cap = caps::has_cap(None, CapSet::Effective, Capability::CAP_SYS_MODULE)
```
DmitryDodzin, line 444: I also noticed that the cap check is thread specific, the agent can be multithreaded no? so do we care only when we are setting up iptables?

DmitryDodzin (approved):
LGTM, I have one question though =]

mirrord/agent/iptables/src/lib.rs
Razz4780: We really care only when doing the check. After that, we should be using the correct backend

PR #3553, opened 2025-09-03
Make db branch default TTL 5 min

0x00A5:


t4lz (changes requested):
Why is that config field an `Option` if it has a default actually?
Can't we make it a `u64` and it would still be optional for users because we specify a default?

Also, please document the default value in the field's docs, and then regenerate the schema and rerun medschool.

mirrord/config/configuration.md
```diff
@@ -828,6 +828,33 @@ This option is compatible only with deployment targets.
     }
 ```
 
+# feature.db_branches {#feature-db_branches}
+
+Configuration for the database branching feature.
+
+A list of configuration for database branches.
```
t4lz: This should probably say "A list of configuration**s**..." plural, right?
Could you please fix that at the source and regenerate?

mirrord/config/src/feature/database_branches.rs
```diff
@@ -75,8 +75,7 @@ pub struct DatabaseBranchConfig {
     /// Name of the database
     pub name: Option<String>,
 
-    /// The time-to-live (TTL) for the branch database, in seconds.
-    #[serde(default = "default_ttl_secs")]
+    /// The time-to-live (TTL) for the branch database, in seconds. Defaults to 300 seconds.
     pub ttl_secs: Option<u64>,
```
t4lz: IMO it's better to have the default here at the config and also eliminate this `Option`, but I guess it's a matter of taste, so up to you.

mirrord/config/configuration.md
```diff
@@ -828,6 +828,33 @@ This option is compatible only with deployment targets.
     }
 ```
 
+# feature.db_branches {#feature-db_branches}
```
t4lz: Seems like the fields' docs aren't being included here. I think it's because `DatabaseBranchConfig` doesn't implement `MirrordConfig`. Did you try just deriving it?

0x00A5: Added the derive, but doesn't include the struct's doc still. My guess is our config generation doesn't support tuple struct?

mirrord/config/src/feature/database_branches.rs
0x00A5: Updated

mirrord/config/configuration.md
0x00A5: Addressed

mirrord/config/configuration.md
0x00A5: In medschool crate, we have [this line](https://github.com/metalbear-co/mirrord/blob/main/medschool/src/types.rs#L78). It tries to get the field name and if it is None, it errors out. The error gets mapped into None again and got ignored by a filter map :sad-pepe:. Tuple struct fields have no ident so the field.ident is None, thus, the inner Vec<DatabaseBranchConfig> docs got ignored by medschool.

0x00A5: Here I hacked up a quick fix for `medschool` so that unamed struct field are included in the type tree: https://github.com/metalbear-co/mirrord/pull/3557. And you can see the generated configuration.md now has the unamed type included.
Just for root causing the issue here, and the issue requires a separate proper fix I think.

PR #3555, opened 2025-09-03
Fixed gRPC error getting transformed into general failure in gRPC

aviramha:
Fixed gRPC error getting transformed into general failure in gRPC because of not passing trailers

### DIFF
--- a/mirrord/intproxy/src/proxies/incoming/http_gateway.rs
+++ b/mirrord/intproxy/src/proxies/incoming/http_gateway.rs
@@ -4,12 +4,17 @@ use std::{
     error::Report,
     fmt,
     net::SocketAddr,
-    ops::ControlFlow,
+    ops::{ControlFlow, Not},
     time::{Duration, Instant},
 };
 
+use bytes::Bytes;
 use http_body_util::BodyExt;
-use hyper::{StatusCode, body::Incoming, http::response::Parts};
+use hyper::{
+    StatusCode,
+    body::{Frame, Incoming},
+    http::response::Parts,
+};
 use mirrord_protocol::{
     Payload,
     batched_body::BatchedBody,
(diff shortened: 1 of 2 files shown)


mirrord/intproxy/src/proxies/incoming/http_gateway.rs
```diff
@@ -78,6 +78,58 @@ impl HttpGatewayTask {
         }
     }
 
+    /// Handles gRPC error responses by ensuring grpc-status and grpc-message are available as
+    /// trailers. This is necessary because gRPC clients expect error information in trailers
+    /// for proper error handling.
```
Razz4780:
```suggestion
    /// trailers.
    ///
    /// This is necessary because gRPC clients expect error information in trailers
    /// for proper error handling, and hyper seems to prefer merging the trailers
    /// into response parts.
```

mirrord/intproxy/src/proxies/incoming/http_gateway.rs
```diff
@@ -1068,4 +1141,104 @@ mod test {
...
+                    response.headers_mut().insert(
+                        "grpc-message",
+                        HeaderValue::from_static("resource not found"),
+                    );
+                    response
+                }))
+            });
+
+            let (connection, _) = listener.accept().await.unwrap();
+            http1::Builder::new()
```
Razz4780: gRPC is HTTP/2

mirrord/intproxy/src/proxies/incoming/http_gateway.rs
```diff
@@ -78,6 +78,58 @@ impl HttpGatewayTask {
         }
     }
 
+    /// Handles gRPC error responses by ensuring grpc-status and grpc-message are available as
+    /// trailers. This is necessary because gRPC clients expect error information in trailers
+    /// for proper error handling.
+    fn ensure_grpc_trailers(
+        &self,
+        headers: &hyper::HeaderMap,
+        mut body_frames: VecDeque<InternalHttpBodyFrame>,
```
Razz4780: This could be a `Vec`, I see some needless conversions at callsites

Razz4780 (changes requested):
Same thing has to be done in the agent's passthrough logic. There, we also use hyper to read an HTTP response that might be a gRPC error

PR #3557, opened 2025-09-04
Support tuple struct for medschool

0x00A5:


mirrord/config/src/feature/database_branches.rs
t4lz, line 1: Why have these internal? I think they should have headers like all other configuration fields and be indexed etc.

0x00A5: I was being lazy :D
The old doc when packed together generates an odd paragraph so I wanted to disable them. But now I have reworked the doc and the paragraph reads much more natural.

t4lz: I think they should have the headers like all other config fields, e.g.:
```
    /// ### feature.copy_target.scale_down {#feature-copy_target-scale_down}
```
Is there any reason to do that differently in here?

0x00A5: Added sub headers for fields

mirrord/config/src/feature/database_branches.rs
```diff
@@ -70,37 +68,69 @@ impl DatabaseBranchesConfig {
...
-    /// Name of the database
+    /// ### feature.db_branches.name {#feature-db_branches-name}
+    ///
+    /// When source database connection detail is not accessible to mirrord operator, users
+    /// can specify the database `name` so it is included in the connection options mirrord
+    /// uses as the override.
     pub name: Option<String>,
 
-    /// The time-to-live (TTL) for the branch database, in seconds. Defaults to 300 seconds.
+    /// ### feature.db_branches.ttl_secs {#feature-db_branches-ttl-secs}
```
t4lz:
```suggestion
    /// ### feature.db_branches.ttl_secs {#feature-db_branches-ttl_secs}
```

PR #3562, opened 2025-09-04
3.161.0

meowjesty:


Cargo.toml
```diff
@@ -31,7 +31,7 @@ resolver = "2"
 # latest commits on rustls suppress certificate verification
 [workspace.package]
 edition = "2024"
-version = "3.160.0"
+version = "3.161.0"
```
DmitryDodzin, line 34: I think the ci will fail because the lock file isn't up to date

meowjesty: Oops, good catch

meowjesty: Should be good now?

PR #3564, opened 2025-09-04
Retries kube requests with a tower `Policy`.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/625
- Related PRs:
  - https://github.com/metalbear-co/kube/pull/3

~~Adds a wrapper around `kube::Api` that retries stuff.~~

Adds a [tower retry policy](https://docs.rs/tower/latest/tower/retry/trait.Policy.html) to the kube `Client`, so kube requests can be retried during mirrord startup.

### DIFF
--- a/changelog.d/+better-ping-pong-restart-error.changed.md
+++ b/changelog.d/+better-ping-pong-restart-error.changed.md
@@ -1,0 +1,1 @@
+Updated so if the PingPong will not attempt to request restart if not connected to operator.
--- a/mirrord/intproxy/src/agent_conn.rs
+++ b/mirrord/intproxy/src/agent_conn.rs
@@ -94,6 +94,12 @@ impl ReconnectFlow {
     }
 }
 
+impl ReconnectFlow {
+    const fn reconnectable(&self) -> bool {
+        matches!(self, ReconnectFlow::ConnectInfo { .. })
+    }
+}
+
 impl fmt::Debug for ReconnectFlow {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
@@ -211,6 +217,10 @@ impl AgentConnection {
             .send(msg)
             .await
             .map_err(|_| AgentConnectionTaskError::ChannelError(self.reconnect.kind()))
+    }
+
+    pub fn reconnectable(&self) -> bool {
+        self.reconnect.reconnectable()
     }
 }
 
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -124,6 +124,8 @@ impl IntProxy {
         let mut background_tasks: BackgroundTasks<MainTaskId, ProxyMessage, ProxyRuntimeError> =
             Default::default();
 
+        let agent_conn_reconnectable = agent_conn.reconnectable();
+
         let agent = background_tasks.register_restartable(
             agent_conn,
             MainTaskId::AgentConnection,
@@ -141,7 +143,14 @@ impl IntProxy {
         // to requests that have a requirement on the mirrord-protocol version.
         background_tasks.suspend_messages(MainTaskId::LayerInitializer);
         let ping_pong = background_tasks.register_restartable(
-            PingPong::new(Self::PING_INTERVAL, Self::PING_PONG_MAX_RECONNECTS),
+            PingPong::new(
+                Self::PING_INTERVAL,
+                if agent_conn_reconnectable {
+                    Self::PING_PONG_MAX_RECONNECTS
+                } else {
+                    0
+                },
+            ),
             MainTaskId::PingPong,
             Self::CHANNEL_SIZE,
         );
--- a/mirrord/intproxy/src/ping_pong.rs
+++ b/mirrord/intproxy/src/ping_pong.rs
@@ -163,7 +163,17 @@ impl RestartableBackgroundTask for PingPong {
 
                 ControlFlow::Continue(())
             }
-            _ => ControlFlow::Break(error),
+            _ => {
+                if self.max_reconnects == 0 {
+                    tracing::warn!(
+                        "no reconnect will be performed because agent is not connected via operator"
+                    );
+                } else {
+                    tracing::warn!(max_reconnects = %self.max_reconnects, "reached maximum amount of reconnects due to PongTimeout");
+                }
+
+                ControlFlow::Break(error)
+            }
         }
     }
 }


meowjesty:
> Retry implementation (...)

Ok, I tried making this work, but am running into some difficulties here.

`Policy` requires the `Request` to be some sort of `Clone` for the impl of `clone_request`:

```rs
    fn clone_request(&mut self, req: &Req) -> Option<Req> {
        todo!()
    }
```

And the kube request stuff just isn't? Even if I try passing completely generic stuff to the impl of `Policy`, like:

```rs
impl<Req, Res, Err> Policy<Req, Res, Err> for RetryKube { ... }
```

We end up with some nasty type errors in the building part:

```rs
        let client_builder = ClientBuilder::try_from(base_config.clone())
            .map_err(KubeApiError::from)?
            .with_layer(&RetryLayer::new(RetryKube::new()))
            .build(); <- trait bound not satisfied
```

```rs
error[E0277]: the trait bound `BoxService<Request<Body>, Response<Box<dyn Body<Data = Bytes, Error = Box<dyn Error + Send + Sync>> + Send + Unpin>>, Box<dyn Error + Send + Sync>>: Clone` is not satisfied
   --> mirrord/operator/src/client.rs:261:14
    |
261 |             .build();
    |              ^^^^^ the trait `Clone` is not implemented for `BoxService<Request<Body>, Response<Box<dyn Body<Data = Bytes, Error = Box<dyn Error + Send + Sync>> + Send + Unpin>>, Box<dyn Error + Send + Sync>>`
    |
    = help: the trait `tower::Service<Request>` is implemented for `tower::retry::Retry<P, S>`
    = note: required for `Retry<RetryKube, BoxService<Request<Body>, Response<Box<dyn Body<Data = Bytes, Error = Box<dyn Error + Send + Sync>> + Send + Unpin>>, Box<dyn Error + Send + Sync>>>` to implement `tower::Service<http::Request<kube::client::Body>>`
    = note: the full name for the type has been written to '/home/meowjesty/dev/metalbear/mirrord/target/debug/deps/mirrord_operator-dc1a994d7726934e.long-type-5153615794418096965.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

I've tried changing some types around, but could not really get this working, but even if it compiles, the `Body` won't be `Clone` (I managed to get it compiling, but `clone_request(&mut self, req: &Req<Body>) -> Option<Req<Body>>` which doesn't work, because `req.clone()` would return me a `&`, and there was no way of cloning the `Body` ...).

The docs say this for `clone_request`:

> If the request cannot be cloned, return [None](https://doc.rust-lang.org/nightly/core/option/enum.Option.html#variant.None). Moreover, the retry function will not be called if the [None](https://doc.rust-lang.org/nightly/core/option/enum.Option.html#variant.None) is returned.

So this means that if we return `None` there, then we don't get a `retry` anyway?

Halp @Razz4780

Cargo.toml
```diff
@@ -74,15 +74,15 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 futures = "0.3"
 thiserror = "2"
-k8s-openapi = { version = "0.24", features = ["earliest"] }
+k8s-openapi = { version = "0.24.0", features = ["earliest"] }
```
Razz4780: Why? :eyes:

mirrord/config/src/retry.rs
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// Sets the max amount of retries that mirrord will try to perform during its startup for
+    /// cluster operations, such as: searching for the target pod, connecting to the
+    /// mirrord-operator, creating the mirrord-agent, etc.
+    ///
+    /// If you want to **disable** mirrord startup retry, set this value to `0`.
+    ///
+    /// Defaults to `1` (retries each operation **once**).
+    #[config(default = 1)]
+    pub max_attempts: u32,
```
Razz4780: Name is confusing, this should be `max_retries`. As it sets the max amount of retries :cat2:

mirrord/config/src/retry.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// ## startup_retry.max_attempts {#startup_retry-max_attempts}
+    ///
+    /// Sets the max amount of retries that mirrord will try to perform during its startup for
+    /// cluster operations, such as: searching for the target pod, connecting to the
+    /// mirrord-operator, creating the mirrord-agent, etc.
+    ///
+    /// If you want to **disable** mirrord startup retry, set this value to `0`.
+    ///
+    /// Defaults to `1` (retries each operation **once**).
+    #[config(default = 1)]
```
Razz4780: I'd set the default retry limit to 2

mirrord/config/src/retry.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// ## startup_retry.max_attempts {#startup_retry-max_attempts}
+    ///
+    /// Sets the max amount of retries that mirrord will try to perform during its startup for
+    /// cluster operations, such as: searching for the target pod, connecting to the
+    /// mirrord-operator, creating the mirrord-agent, etc.
+    ///
+    /// If you want to **disable** mirrord startup retry, set this value to `0`.
+    ///
+    /// Defaults to `1` (retries each operation **once**).
```
Razz4780:
```suggestion
    /// Sets the max amount of retries for Kubernetes API requests made by mirrord during startup
    /// (e.g. for resolving the target or connecting to the mirrord Operator).
    ///
    /// If you want to **disable** request retries, set this value to `0`.
    ///
    /// Defaults to `2`.
```

Also, please rename "cluster operations" to "Kubernetes API requests" in the remaining docs, to be consistent

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -57,7 +60,13 @@ impl KubernetesAPI {
         )
         .await?;
 
-        let client = progress.suspend(|| client_config.try_into())?;
+        let client = progress
+            .suspend(|| ClientBuilder::try_from(client_config.clone()))?
+            .with_layer(&BufferLayer::new(1024))
```
Razz4780, line 65: What is buffer layer? :eyes:

mirrord/config/src/lib.rs
```diff
@@ -715,6 +721,27 @@ impl LayerConfig {
             );
         }
 
+        if self.startup_retry.min_ms > self.startup_retry.max_ms {
+            return Err(ConfigError::InvalidValue {
+                name: "startup_retry.min_ms",
+                provided: self.startup_retry.min_ms.to_string(),
+                error: format!(
+                    "the value of startup_retry.min_ms must be smaller than {}\
```
Razz4780:
```suggestion
        if self.startup_retry.min_ms > self.startup_retry.max_ms {
            return Err(ConfigError::InvalidValue {
                name: "startup_retry.min_ms",
                provided: self.startup_retry.min_ms.to_string(),
                error: format!(
                    "the value of startup_retry.min_ms must not be greater than {}\
```

changelog.d/+625.added.md
```diff
@@ -0,0 +1 @@
+Adds two new options for retrying mirrord startup issues that may happen due to cluster connectivity problems.
```
Razz4780: mirrord now retries failed Kubernetes API requests made during startup. This should improve stability when facing transient errors, like cluster connectivity issues.

The retry policy can be adjusted in `startup_retry` mirrord config section.

mirrord/kube/src/retry.rs
```diff
@@ -0,0 +1,131 @@
...
+        match result {
+            Ok(response)
+                if matches!(
+                    response.status(),
+                    StatusCode::TOO_MANY_REQUESTS
+                        | StatusCode::SERVICE_UNAVAILABLE
+                        | StatusCode::GATEWAY_TIMEOUT
+                ) && self.current_attempt < self.max_attempts =>
+            {
+                self.current_attempt += 1;
```
Razz4780, line 112: Uhm, are you sure this is right? Should the number of attempts be tracked in the `Policy` implementor? :eyes: 
It would have to be cloned by `tower` for each request, I don't know if this is the case. Code for [`RetryLayer`](https://docs.rs/tower/latest/src/tower/retry/layer.rs.html#17-27) looks weird.

Intuitively, I'd store attempt count in a request extension, as it should be tracked separately for each request

mirrord/kube/src/lib.rs
```diff
@@ -16,6 +16,15 @@
...
+pub use kube;
+
+use crate::retry::RetryKube;
+
 pub mod api;
 pub mod error;
 pub mod resolved;
+pub mod retry;
+
+pub static RETRY_KUBE_OPERATIONS_POLICY: OnceLock<RetryKube> = OnceLock::new();
```
Razz4780: I think this variable is never actually read? Only initialized in multiple places. I'd remove it

Cargo.toml
meowjesty: Ops, left over from when I tried to plug our kube fork from main.

mirrord/config/src/retry.rs
meowjesty: Changed.

mirrord/kube/src/api/kubernetes.rs
meowjesty:
```
/// - We need the `BufferLayer` so the service implements `Clone`.
```

It says that the service has to be `BoxService: Clone`, but our thing is not `Clone`, so `BufferLayer` basically turns one of the generic crap into `BoxService<whatever>: Clone`.

mirrord/config/src/retry.rs
meowjesty: Changed.

mirrord/kube/src/lib.rs
meowjesty: Yup, removed.

mirrord/kube/src/retry.rs
meowjesty: Sure? Nope.

Changed it to extension, it's added and incremented in `try_clone`.

Should I wrap the `u32` into some type to avoid a potential (unlikely?) future problem if kube requests decide to add their own `u32` extension?

mirrord/config/src/retry.rs
meowjesty: Changed.

mirrord/config/src/lib.rs
meowjesty: Changed. Also changed where we print the number.

changelog.d/+625.added.md
meowjesty: Changed.

mirrord/kube/src/retry.rs
Razz4780: Ah, I think nvm :sweat_smile: 

```
The Policy must also implement Clone. This middleware will clone the policy
for each request session. This means a new clone of the policy will be created
for each initial request and any subsequent retries of that request.
Therefore, any state stored in the Policy instance is for that
request session only. In order to share data across request sessions,
that shared state may be stored in an Arc,
so that all clones of the Policy type reference the same instance of the shared state.
```

Also checked `tower` source code. You can revert this change, having the state in `KubeRetry` is fine :V

meowjesty: Changed it back.

changelog.d/+625.added.md
```diff
@@ -0,0 +1,4 @@
+mirrord now retries failed Kubernetes API requests made during startup. This should improve stability when facing transient errors, like cluster connectivity issues.
+
+The retry policy can be adjusted in startup_retry mirrord config section.
+`
```
Razz4780:
```suggestion

```
:wastebasket:

Issue #2538, opened 2024-06-20
Support HTTP filtering when mirroring


eyalb181:
Support using `feature.network.incoming.http_filter` in mirror mode.

Given the recent passthrough mirroring overhaul, this should not be hard. Things to do:

1. Add a new mirrord-protocol message for filtered mirror subscription.
    * See crate-level docs to learn how not to break stuff.
    * See [steal subscription request](https://github.com/metalbear-co/mirrord/blob/9c15bd11ecb091de3c1ae0fd75af0ae5f1f3047f/mirrord/protocol/src/tcp.rs#L340) for reference.
2. Use the filter in the mirrord-agent when receiving mirrored HTTP traffic. Looks like most of the changes will happen in the [`TcpMirrorApi`](https://github.com/metalbear-co/mirrord/blob/9c15bd11ecb091de3c1ae0fd75af0ae5f1f3047f/mirrord/agent/src/mirror.rs#L24) struct. In there, we can:
    * Reject the subscription if passthrough mirroring mode is not enabled.
    * Discard TCP traffic, and HTTP traffic that does not match the filter. See how filtering is done [when stealing](https://github.com/metalbear-co/mirrord/blob/9c15bd11ecb091de3c1ae0fd75af0ae5f1f3047f/mirrord/agent/src/steal/task.rs#L187).
3. Make sure that nothing breaks in the CLI and mirrord-layer. Right now the http filter is ignored in mirror mode, and this change might affect logic (e.g. when handling `bind` or `listen` in mirrord-layer, due to `feature.network.incoming.http_filter.ports`)
4. Add an E2E test.

PR #3567, opened 2025-09-09
Support HTTP filtering when mirroring 

vrrashkov:
I saw that the `LayerTcpSteal` already provides http_filter so we can use similar logic for the mirrord type. 
To keep it consistent I think changing `PortSubscribe(Port)` to `PortSubscribe(MirrorType)` should be best, so I have created a new type same as Steal that will split the logic filters/no filters. 

Added incoming feature I added `MirrorFiltered` which is similar to how Steal works, maybe this can be just called Mirror and also provide filter settings to it but make them optional because currently Mirror is returned as default if no filters are provided. But I was not sure which approach is best os thats why I split it into `MirroFiltered` and keeping the old Mirror here.

For the tests I saw there was only steal ones in traffic so created a new mirror mod and added couple of tests similar to the steal ones. 

Also changes to the parts of the code that need to use the `MirrorType`. 

Let me know if this is the desired approach to the task. Also not sure if my setup is working fully correct because when trying to run all the tests some are failing this is without any changes only with the main branch I am on **Mac** so there might be some stuff I have not configured correctly, but the project builds and the newly added tests are working, need to investigate more on that. 

Issue: #2538


mirrord/protocol/src/tcp.rs
```diff
@@ -65,7 +65,7 @@ pub struct TcpClose {
 pub enum LayerTcp {
     /// User is interested in mirroring traffic on this `Port`, so add it to the list of
     /// ports that the sniffer is filtering.
-    PortSubscribe(Port),
+    PortSubscribe(MirrorType),
```
Razz4780: We can't change existing variants due to bincode constraints :<
This has to be a new variant of `LayerTcp`, added at the end of the enum.
See crate level docs for reference (also please check the bit on required version negotiation)

mirrord/protocol/src/tcp.rs
```diff
@@ -315,6 +315,28 @@ impl StealType {
...
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+#[protocol_break(2)]
+pub enum MirrorType {
+    /// Mirror all traffic to this port.
+    All(Port),
+    /// Mirror HTTP traffic matching a given filter (header based). - REMOVE THIS WHEN BREAKING
+    /// PROTOCOL
+    FilteredHttp(Port, Filter),
+    /// Mirror HTTP traffic matching a given filter - supporting more than once kind of filter
+    FilteredHttpEx(Port, HttpFilter),
```
Razz4780:
```suggestion
    /// Mirror HTTP traffic matching a given filter.
    FilteredHttp(Port, HttpFilter),
```
No need to add the deprecated variant in the new messages ^^

changelog.d/2538.added.md
```diff
@@ -0,0 +1,8 @@
+`LayerTcpSteal` already provides `http_filter` so we can use similar logic for the mirrord type. 
+To keep it changed `PortSubscribe(Port)` to `PortSubscribe(MirrorType)` should be best, so I have created a new type same as Steal that will split the logic filters/no filters. 
+
+Added incoming feature I added `MirrorFiltered` which is similar to how Steal works, maybe this can be just called `Mirror` and also provide filter settings to it but make them optional because currently Mirror is returned as default if no filters are provided. 
+
+For the tests created a new mirror mod and added couple of tests similar to the steal ones. 
+
+Also changes to the parts of the code that need to use the `MirrorType`.
```
Razz4780:
```suggestion
Mirroring incoming TCP traffic now supports using an HTTP filter.
```

This changelog is for the users, we try to keep the entries short.

mirrord/layer/src/setup.rs
```diff
@@ -212,13 +212,25 @@ pub struct StealHttpSettings {
...
     pub ports: HashSet<Port>,
 }
+/// Settings for handling HTTP with the `mirror` feature.
+#[derive(Debug)]
+pub struct MirrorHttpSettings {
+    /// The HTTP filter to use.
+    pub filter: HttpFilter,
+    /// Ports to filter HTTP on.
+    pub ports: HashSet<Port>,
+}
```
Razz4780: Instead of adding a new struct, I'd change `IncomingMode` to reuse existing types (and rename them):

```rust
struct IncomingMode {
    pub steal: bool,
    pub http_settings: Option<HttpSettings>,
}
```

tests/src/traffic/mirror.rs
```diff
@@ -0,0 +1,230 @@
...
+        req_builder,
+        Some("OK - DELETE: Request completed\n"),
+        HeaderMap::default(),
+    )
+    .await;
+
+    application.assert(&mirror_process).await;
+}
+
+/// Test that mirror mode rejects HTTP filters when not using passthrough mirroring.
```
Razz4780: This doesn't have to be an E2E test, can be a unit test in the agent ^^

tests/src/traffic/mirror.rs
```diff
@@ -0,0 +1,230 @@
...
+        &service.namespace,
+        80,
+    )
+    .await;
+    let url = format!("http://{}", portforwarder.address());
+
+    // Use mirror mode with HTTP header filter config
+    let mut config_path = config_dir.to_path_buf();
+    config_path.push("http_filter_header_mirror.json");
+
```
Razz4780: Instead of adding more test configs in the repo, please use [`NamedTempFile`](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html) to create the config file from within the test.
Like this:

```rust
fn my_test() {
    let config = serde_json!({
        // config json
    });
    let mut config_file = NamedTempFile::with_suffix(".json").unwrap();
    serde_json::to_writer(config_file.as_file_mut(), &config).unwrap();
}
```

(we plan to remove the existing test configs as well)

Razz4780 (changes requested):
Looking good! :smile: 
Left some comments for u

mirrord/layer/src/setup.rs
vrrashkov: The idea here was to change the whole enum to struct like you suggested right? I did that and also it reduced quite a bit of the repeated logic.

mirrord/protocol/src/tcp.rs
vrrashkov: Reverted the change and added the new on the bottom which now only takes Port and Filters, not sure if the name is okay or you want me to change that. `PortSubscribeFilteredHttp(Port, HttpFilter)`. I bumped the minor version as I saw its mentioned to do that when extending the protocol or there was no need for that. Not sure if I need to do more on this end ?

mirrord/layer/src/setup.rs
vrrashkov: I did similar logic in `port_forward` I noticed I forgot that part on the initial commit.

mirrord/protocol/src/tcp.rs
vrrashkov: Looked into it more and implemented the protocol_version check so the older versions don't interact with filters logic.

mirrord/agent/src/mirror.rs
```diff
@@ -76,23 +83,47 @@ impl TcpMirrorApi {
...
                     incoming_streams.remove(&id);
                 }
-
                 LayerTcp::PortSubscribe(port) => {
+                    tracing::debug!(port = %port, "Received PortSubscribe (no filter)");
                     mirror_handle.mirror(port).await?;
                     queued_messages.push_back(DaemonTcp::SubscribeResult(Ok(port)));
                 }
+                LayerTcp::PortSubscribeFilteredHttp(port, filter) => {
+                    if protocol_version.matches(&MIRROR_HTTP_FILTER_VERSION) {
```
Razz4780: The version requirement is for sending the message, not receiving :sweat_smile: 
We use it to confirm that sending the message will not break the decoding logic on the other side of the `mirrord-protocol` connection.
So instead, this check should be done whenever we want to send this message.

mirrord/agent/src/mirror.rs
```diff
@@ -202,7 +244,18 @@ impl TcpMirrorApi {
...
+                        MirroredTraffic::Http(mut http) if protocol_version.matches(&MODE_AGNOSTIC_HTTP_REQUESTS) => {
+                            // Check HTTP filter if one exists for this port
+                            if let Some(filter) = port_filters.get(&http.info.original_destination.port()) &&  !filter.matches(http.parts_mut()) {
+                                // Request doesn't match filter - don't mirror it to the layer
+                                // The original request will still be passed through to its destination
+                                return Ok(DaemonMessage::LogMessage(LogMessage::warn(format!(
+                                    "HTTP request filtered out from mirror: {} {}",
+                                    http.parts().method,
+                                    http.parts().uri
+                                ))));
```
Razz4780: No need to issue a warning in this case. It's normal that some requests will not be captured

tests/src/traffic/mirror.rs
```diff
@@ -0,0 +1,116 @@
...
+    services::basic_service,
+};
+
+/// Test mirror mode with HTTP header filter.
+/// Only requests with matching headers should be mirrored.
+/// All requests (matching and non-matching) should still reach the original server.
+#[rstest]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[timeout(Duration::from_secs(120))]
+async fn mirror_with_http_header_filter(
```
Razz4780, line 23: This test doesn't actually check that the local app received exactly the requests we expect, only that it received the last `DELETE`. It could as well receive all of the requests.

I suggest doing the same thing as we have in the mirror tests [here](https://github.com/metalbear-co/mirrord/blob/main/tests/src/http.rs)

mirrord/agent/src/mirror.rs
vrrashkov: Ah makes sense, I changed it  but not sure if there was easier way to access the protocol_version so currently I had to pass it down to the agent_subscribe to use it.

tests/src/traffic/mirror.rs
```diff
@@ -0,0 +1,134 @@
...
+            "network": {
+                "incoming": {
+                    "mode": "mirror",
+                    "http_filter": {
+                        "header_filter": "x-filter: yes"
+                    }
+                }
+            }
+        }
+    });
```
Razz4780:
```suggestion
    let config = json!({
        "feature": {
            "network": {
                "incoming": {
                    "mode": "mirror",
                    "http_filter": {
                        "header_filter": "x-filter: yes"
                    }
                }
            }
        },
        "agent": {
            "passthrough_mirroring": true,
        },
    });
```

Razz4780: I think this test needs passthrough mirroring enabled

vrrashkov: Changed it, also trying to see if I can reproduce this locally.

vrrashkov: Still trying to reproduce but wanted to mention, not sure if you noticed but its happening only for FastAPI on the CI , the other 2 are passing, not sure if you have encountered similar behaviour before 🤔? But the strange thing is the previous commit passed all of them. 


```sh
test traffic::mirror::mirror_with_http_header_filter::application_3_Application__NodeHTTP ... ok
test traffic::mirror::mirror_with_http_header_filter::application_2_Application__PythonFastApiHTTP ... FAILED
test traffic::mirror::mirror_with_http_header_filter::application_1_Application__PythonFlaskHTTP ... ok

vrrashkov: So it turns out that `get_stdout` returns some info data from the FastAPI so when I'm checking for GET/DELETE strings it gets from the FastAPI logs also. So I changed it to just check the "Request completed" string. This can probably be improved further so get_stdout only prints the responses maybe without the logs from the FastAPI.  I think that was it at least, because here is what we have from the `get_stdout`. Those INFO lines are counted as a request but they should not be here I think. 

This issue doesn't exist in the `http.rs` mod because there is no FastAPI usage there. 

```sh
stdout lines:
GET: Request completed
INFO:     127.0.0.1:64052 - "GET / HTTP/1.1" 200 OK
DELETE: Request completed
INFO:     127.0.0.1:64070 - "DELETE / HTTP/1.1" 200 OK`

Razz4780: Huh, okay. All good than ^^

Issue #3563, opened 2025-09-04
Forked child process signal handling


owen-c:
### Bug Description

I am using mirrord to run a Rails app that uses Puma app server in cluster mode. I have tested this across many versions of puma and ruby and this behavior is present only when using mirrord. I haven't had time to prove it but my hunch is something is going wrong with child -> parent signals.

Puma in cluster mode has a forks child worker processes. When it receives a SIGINT it forwards this to its children and then waits for all children to exit.

The behavior I'm seeing with mirrord is that it properly catches and forwards SIGINT, the child worker process shutdown as expected, however the master process never exits because it is stuck waiting for the children to exit even though they have.

I profiled the master process and it get stuck in the Process.waitall function of ruby
https://rubydoc.info/stdlib/core/Process:waitall

I attached the flamegraph for reference.

![Image](https://github.com/user-attachments/assets/8ef5b650-4a69-4cf3-be19-35b12c74e41c)

This is mostly an annoyance b/c I can kill the process, but would love for SIGINT to work properly with puma in cluster mode.

### Steps to Reproduce

Repo:
https://github.com/owen-c/puma-repro/tree/main

Install ruby
Run `bundle install`
Run `mirrord exec bundle exec puma`
Use CTRL-C to kill, see logs that show worker boot and worker exit but master process hangs

Run `bundle exec puma`
Use CTRL-C to kill, see all processes exit.

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### Your operating system and version

Mac OS 15.6.1

### Local process

/Users/owen/.local/share/mise/installs/ruby/2.7.8/bin/puma: Ruby script text executable, ASCII text

PR #3570, opened 2025-09-11
Move intproxy out of user process's children

MintSoup:
Fix for #3563

### DIFF
--- a/tests/intproxy_child/intproxy_child.c
+++ b/tests/intproxy_child/intproxy_child.c
@@ -1,4 +1,3 @@
-#include <asm-generic/errno-base.h>
 #include <stdio.h>
 #include <string.h>
 #include <sys/wait.h>


changelog.d/3563.fixed.md
```diff
@@ -0,0 +1 @@
+Intproxy now no longer runs as a child of the user process so it doesn't get reaped by wait(2) calls. ([#3536](https://github.com/metalbear-co/mirrord/issues/3563))
```
Razz4780:
```suggestion
Intproxy now no longer runs as a child of the user process so it doesn't get reaped by wait/waitall calls.
```

When you use the issue number in the filename, towncrier adds these links automatically ^^

mirrord/cli/src/util.rs
```diff
@@ -53,6 +60,23 @@ pub(crate) unsafe fn detach_io() -> Result<(), nix::Error> {
...
 }
 
+/// "Reparent" this process to init by forking and then exiting on the
+/// parent. The child process will get reparented to init.
+///
+/// # Safety
+///
+/// Since this function calls `nix::unistd::fork`, the safety section
+/// from that function applies. Namely, we can only call it in a
+/// single-threaded context.
```
Razz4780:
```suggestion
/// "Reparent" this process to init by forking and then exiting in the
/// parent. The child process will get reparented to init,
/// and return from this function.
///
/// # Safety
///
/// This function forks the current process.
/// If the current process uses multiple threads,
/// it will be bound by [`signal-safety`](https://man7.org/linux/man-pages/man7/signal-safety.7.html) rules after returning from this function.
```

mirrord/cli/src/execution.rs
```diff
@@ -310,6 +314,12 @@ impl MirrordExecution {
 
         let stdout = proxy_process.stdout.take().expect("stdout was piped");
 
+        // The pre_exec(reparent_to_init) causes the process to fork
+        // and our immediate child promptly exits (which is what we
+        // wait for here), reparenting our (now former) grandchild to
+        // init.
+        proxy_process.wait().await.unwrap();
```
Razz4780:
```suggestion
        // The pre_exec(reparent_to_init) causes the process to fork
        // and our immediate child promptly exits (which is what we
        // wait for here), reparenting our (now former) grandchild to
        // init.
        // This should *never* fail, see https://man7.org/linux/man-pages/man2/wait.2.html
        // for reference.
        proxy_process.wait().await.unwrap();
```

tests/intproxy_child/intproxy_child.c
```diff
@@ -0,0 +1,6 @@
+#include <sys/wait.h>
+
+int main() {
+	wait(0);
```
Razz4780: I'd verify that the call failed with `ECHILD`

tests/src/utils/application.rs
```diff
@@ -47,6 +47,7 @@ pub enum Application {
     ///
     /// - `node-e2e/fspolicy/test_operator_fs_policy.mjs`
     NodeFsPolicy,
+    IntproxyChild,
```
Razz4780, line 52:
```suggestion
    /// Waits for any child process to finish, and verifies that the `wait` call fails with `ECHILD`.
    IntproxyChild,
```

tests/intproxy_child/intproxy_child.c
Razz4780:
```suggestion
    // Verify that this process has no children.
    // This tests that mirrord-intproxy is *not* a child of this process.
	wait(0);
```

PR #3571, opened 2025-09-11
Prevent potentially sensitive envvars from being logged

MintSoup:
Self explanatory. The Debug impl is identical to the one generated by `derive` except in the `GetEnvVarsResponse` variant we print `<REDACTED>` instead of the actual envvars.

Ideally we'd use something like [SecretBox](https://docs.rs/secrecy/latest/secrecy/struct.SecretBox.html) but that'd make us unable to use many `derive`s, most notably `bincode::Encode`. `Decode` and `Clone` would *technically* be doable but not very ergonomical.

### DIFF
--- a/mirrord/agent/src/entrypoint.rs
+++ b/mirrord/agent/src/entrypoint.rs
@@ -506,10 +506,8 @@ impl ClientConnectionHandler {
                 let env_vars_result =
                     env::select_env_vars(&self.state.env, env_vars_filter, env_vars_select);
 
-                self.respond(DaemonMessage::GetEnvVarsResponse(
-                    env_vars_result.map(Into::into),
-                ))
-                .await?
+                self.respond(DaemonMessage::GetEnvVarsResponse(env_vars_result))
+                    .await?
             }
             ClientMessage::GetAddrInfoRequest(request) => {
                 self.dns_api
--- a/mirrord/agent/src/env.rs
+++ b/mirrord/agent/src/env.rs
@@ -3,7 +3,7 @@ use std::{
     path::PathBuf,
 };
 
-use mirrord_protocol::RemoteResult;
+use mirrord_protocol::{RemoteEnvVars, RemoteResult};
 use tokio::io::AsyncReadExt;
 use wildmatch::WildMatch;
 
@@ -119,7 +119,7 @@ pub(crate) fn select_env_vars(
     full_env: &HashMap<String, String>,
     filter_env_vars: HashSet<String>,
     select_env_vars: HashSet<String>,
-) -> RemoteResult<HashMap<String, String>> {
+) -> RemoteResult<RemoteEnvVars> {
     let env_filter = EnvFilter::new(select_env_vars, filter_env_vars);
     let env_vars = full_env
         .iter()
@@ -128,7 +128,7 @@ pub(crate) fn select_env_vars(
         // [("DB", "foo.db")]
         .collect::<HashMap<_, _>>();
 
-    Ok(env_vars)
+    Ok(env_vars.into())
 }
 
 #[cfg(test)]


mirrord/protocol/src/codec.rs
```diff
@@ -205,6 +205,42 @@ pub enum DaemonMessage {
     OperatorPing(u128),
 }
 
+// We use a custom Debug impl to redact the payload from
```
Razz4780: Instead of implementing `Debug` for the whole `DaemonMessage`, you can safely apply the newtype pattern to the `HashMap` inside `GetEnvVarsResponse`, it won't affect bincode representation.

```rust
#[derive(Encode, Decode, PartialEq, Eq, Clone)]
struct GetEnvVarsResponse(pub HashMap<String, String>);

impl fmt::Debug for GetEnvVarsResponse { ... }
```

And the compliation errors will help you make sure that the map are not unnecessarily logged anywhere ;)

And also, we'll want to log the full env vars at least once after receiving them (in the CLI or intproxy). It's ok to do it on the local side, since the user's process will see them anyway. What we don't want is logging them in the cluster (agent or operator)

MintSoup: What about making this generic, something like `Sensitive<T>`? (could probably use a better name) Basically `SecretBox` but with relaxed restrictions (and without the box).
Is there anywhere else in the codebase this could be useful?

MintSoup: And what about creating something like `struct EnvVars` with `Deref<HashMap<String, String>>` (and a debug impl with redacted output) and using it everywhere? I realize removing all envvar logs is bit extreme but if its used for secrets it might be better to err on the side of caution and make the logging behavior opt-in rather than opt-out.

Razz4780: I don't think we redact anything else atm, so I wouldn't add unnecessary generics.

The `EnvVars` struct sounds nice though, I'd call it `RemoteEnvVars` for clarity :+1:

MintSoup: Alright, the basic impl is ready now. Where else might `RemoteEnvVars` be useful? We should probably refactor those in the same PR.

BTW, we could do something like this as well:
```rust
#[derive(derive_more::Debug, ...)]
pub enum DaemonMessage {
	...
    GetEnvVarsResponse(#[debug("<REDACTED>")] RemoteResult<HashMap<String, String>>),
	...
}
``` 
Pretty much the same thing as the original solution except a little more terse.

Razz4780: Nah, I think having a wrapper is better. It will be harder to log the env vars by mistake

mirrord/protocol/Cargo.toml
```diff
@@ -30,6 +30,7 @@ http-body-util = { workspace = true }
 fancy-regex = { workspace = true }
 socket2.workspace = true
 semver = { workspace = true, features = ["serde"] }
+derive_more = { workspace = true, features = ["from", "into", "deref"] }
```
Razz4780, line 33: :wastebasket:

Cargo.toml
```diff
@@ -63,6 +63,9 @@ actix-codec = "0.5"
 bincode = { version = "2", features = ["serde"] }
 bytes = "1"
 
+# Used by `protocol`
+derive_more = "2.0.1"
```
Razz4780, line 67: :wastebasket:

mirrord/agent/src/entrypoint.rs
```diff
@@ -506,8 +506,10 @@ impl ClientConnectionHandler {
                 let env_vars_result =
                     env::select_env_vars(&self.state.env, env_vars_filter, env_vars_select);
 
-                self.respond(DaemonMessage::GetEnvVarsResponse(env_vars_result))
-                    .await?
+                self.respond(DaemonMessage::GetEnvVarsResponse(
+                    env_vars_result.map(Into::into),
```
Razz4780: Instead of converting the plain hashmap at the last moment, I'd make sure that the envs are *always* stored and passed around in the wrapper on the agent side.
Right now I think no `Debug` impl contains them, but it's possible that this will change in the future by accident

Cargo.toml
MintSoup: ?

PR #3573, opened 2025-09-11
Adds a timeout for agent exit, and runs `BackgroundTask`s in a separate runtime always.

meowjesty:
- Issue: https://github.com/metalbear-co/private/issues/592

### DIFF
--- a/changelog.d/+workload-patch-crd.internal.md
+++ b/changelog.d/+workload-patch-crd.internal.md
@@ -1,0 +1,1 @@
+Added a new CRD for persisting workload patches.
--- a/mirrord/operator/src/crd.rs
+++ b/mirrord/operator/src/crd.rs
@@ -23,6 +23,7 @@ pub mod kafka;
 pub mod kube_target;
 pub mod label_selector;
 pub mod mysql_branching;
+pub mod patch;
 pub mod policy;
 pub mod profile;
 pub mod session;
--- a/mirrord/operator/src/crd/patch.rs
+++ b/mirrord/operator/src/crd/patch.rs
@@ -1,0 +1,88 @@
+use kube::CustomResource;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Internal resource, representing operator patch state of some workload.
+///
+/// This is an aggregated state of accepted [`MirrordClusterWorkloadPatchRequest`]s.
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[kube(
+    group = "mirrord.metalbear.co",
+    version = "v1alpha",
+    kind = "MirrordClusterWorkloadPatch",
+    printcolumn = r#"{"name":"WORKLOAD-API-VERSION", "type":"string", "description":"API version of the workload.", "jsonPath":".spec.workloadRef.apiVersion"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-KIND", "type":"string", "description":"Kind of the workload.", "jsonPath":".spec.workloadRef.kind"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAME", "type":"string", "description":"Name of the workload.", "jsonPath":".spec.workloadRef.name"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAMESPACE", "type":"string", "description":"Namespace of the workload.", "jsonPath":".spec.workloadRef.namespace"}"#
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordClusterWorkloadPatchSpec {
+    /// Reference to the workload being patched.
+    pub workload_ref: PatchedWorkloadRef,
+    /// Collection of accepted [`MirrordClusterWorkloadPatchRequest`]s.
+    ///
+    /// These requests will be enforced on the workload.
+    pub requests: Vec<AcceptedPatchRequest>,
+}
+
+/// Reference to a Kubernetes workload patched with [`MirrordClusterWorkloadPatch`] and
+/// [`MirrordClusterWorkloadPatchRequest`].
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct PatchedWorkloadRef {
+    pub api_version: String,
+    pub kind: String,
+    pub namespace: String,
+    pub name: String,
+}
+
+/// An accepted [`MirrordClusterWorkloadPatchRequest`], part of [`MirrordClusterWorkloadPatch`].
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct AcceptedPatchRequest {
+    /// UID of the origin [`MirrordClusterWorkloadPatchRequest`] resource.
+    pub uid: String,
+    /// Environment variables to be set in the pod template.
+    pub env_vars: Vec<EnvVar>,
+    /// Number of replicas to be set in the workload spec.
+    ///
+    /// [`None`] means that replicas should not be patched.
+    pub replicas: Option<i32>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct EnvVar {
+    pub container: String,
+    pub variable: String,
+    pub value: String,
+}
+
+/// Internal resource, representing an operator request to patch some workload.
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+#[kube(
+    group = "mirrord.metalbear.co",
+    version = "v1alpha",
+    kind = "MirrordClusterWorkloadPatchRequest",
+    printcolumn = r#"{"name":"WORKLOAD-API-VERSION", "type":"string", "description":"API version of the workload.", "jsonPath":".spec.workloadRef.apiVersion"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-KIND", "type":"string", "description":"Kind of the workload.", "jsonPath":".spec.workloadRef.kind"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAME", "type":"string", "description":"Name of the workload.", "jsonPath":".spec.workloadRef.name"}"#,
+    printcolumn = r#"{"name":"WORKLOAD-NAMESPACE", "type":"string", "description":"Namespace of the workload.", "jsonPath":".spec.workloadRef.namespace"}"#,
+    status = "MirrordClusterWorkloadPatchRequestStatus"
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordClusterWorkloadPatchRequestSpec {
+    /// Reference to the workload being patched.
+    pub workload_ref: PatchedWorkloadRef,
+    /// Environment variables to be set in the pod template.
+    pub env_vars: Vec<EnvVar>,
+    /// Number of replicas to be set in the workload spec.
+    ///
+    /// [`None`] means that replicas should not be patched.
+    pub replicas: Option<i32>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Hash, PartialEq, Eq)]
+pub struct MirrordClusterWorkloadPatchRequestStatus {
+    pub accepted: bool,
+}
--- a/mirrord/operator/src/setup.rs
+++ b/mirrord/operator/src/setup.rs
@@ -31,6 +31,7 @@ use crate::crd::{
     MirrordOperatorCrd, MirrordSqsSession, MirrordWorkloadQueueRegistry, TargetCrd,
     kafka::{MirrordKafkaClientConfig, MirrordKafkaEphemeralTopic, MirrordKafkaTopicsConsumer},
     mysql_branching::MysqlBranchDatabase,
+    patch::{MirrordClusterWorkloadPatch, MirrordClusterWorkloadPatchRequest},
     policy::{MirrordClusterPolicy, MirrordPolicy},
     profile::{MirrordClusterProfile, MirrordProfile},
     session::MirrordClusterSession,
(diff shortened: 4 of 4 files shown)


changelog.d/+592.changed.md
```diff
@@ -0,0 +1 @@
+Improves the agent handling of background tasks by having them run in a separate tokio runtime. Also adds a timeout for waiting on these tasks to finish. The timeout can be configured via the agent arg `exit_timeout` or the agent env var `MIRRORD_AGENT_EXIT_TIMEOUT` (defaults to `5` seconds).
```
Razz4780: I don't think we need the timeout to be configurable. The background tasks should always finish instantly, more or less. We can use hardcoded 5s

mirrord/agent/src/entrypoint.rs
```diff
@@ -241,6 +242,8 @@ struct BackgroundTasks {
     stealer: BackgroundTask<StealerCommand>,
     dns: BackgroundTask<DnsCommand>,
     mirror_handle: Option<MirrorHandle>,
+    // forever: BackgroundTask<i32>,
+    // dies: BackgroundTask<i32>,
```
Razz4780: :wastebasket:

mirrord/agent/src/entrypoint.rs
```diff
@@ -844,24 +849,24 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
-            error!(%error, "start_agent -> Sniffer task failed");
-        }),
-        stealer.wait().inspect_err(|error| {
-            error!(%error, "start_agent -> Stealer task failed");
-        }),
-        dns.wait().inspect_err(|error| {
-            error!(%error, "start_agent -> DNS task failed");
-        }),
-    );
+    let _ = timeout(
```
Razz4780: This won't tell us which task is lingering. Maybe you can implement the timeout logic in `BackgroundTask::wait`?
Also, the timeout error is only logged. Would be nice if it was returned as well.

mirrord/agent/src/entrypoint/setup.rs
```diff
@@ -26,6 +24,8 @@ pub(super) async fn start_traffic_redirector(
     target_pid: u64,
     with_mesh_exclusion: Option<u16>,
 ) -> AgentResult<(StealHandle, MirrorHandle)> {
+    let _rt = runtime.handle().enter();
```
Razz4780, line 29: Not sure if I like this pattern :V
I feel like `runtime.handle().spawn(async move { ... })` is more explicit.

If you want this, please add a comment like:
```rust
// IMPORTANT: this makes tokio tasks spawn on `runtime`.
// Do not remove this.
let _rt = runtime.handle().enter();
```

mirrord/agent/src/sniffer.rs
```diff
@@ -345,6 +345,20 @@ where
...
+                            client_id,
+                            destination_port = identifier.dest_port,
+                            source_port = identifier.source_port,
+                            tcp_flags = tcp_packet.flags,
+                            bytes = tcp_packet.bytes.len(),
+                            "TEST-ONLY: Client queue of new sniffed TCP connections is full, dropping",
+                        );
+
+                        continue;
+                    }
```
Razz4780: Wut :eyes: 
Pls no

mirrord/agent/src/sniffer.rs
```diff
@@ -426,6 +443,9 @@ mod test {
         packet_tx: Sender<(TcpSessionDirectionId, TcpPacketData)>,
         times_filter_changed: Arc<AtomicUsize>,
         next_client_id: ClientId,
+        cancellation_token: CancellationToken,
+        #[allow(dead_code)]
+        runtime: BgTaskRuntime,
```
Razz4780:
```suggestion
        _runtime: BgTaskRuntime,
```

mirrord/agent/src/sniffer.rs
```diff
@@ -733,9 +763,12 @@ mod test {
 
     /// Simulates scenario where client does not read notifications about new connections fast
     /// enough. Client should miss new connections.
+    ///
+    /// There's a special `cfg(test)` case in [`TcpConnectionSniffer::handle_packet`] that skips
+    /// over a message if the `source_port: 3128` to simulate a full channel of sniffed connections.
```
Razz4780: Yeah, this test just doesn't work with multiple threads. Give it a 100ms sleep after `std::mem::drop(permit);`, and it will be fine

Razz4780: Also, please add a similar sleep to `client_lagging_on_data`, after:
```rust
        // Wait until sniffer consumes all messages.
        setup
            .packet_tx
            .reserve_many(setup.packet_tx.max_capacity())
            .await
            .unwrap();
```
That one seems to be passing, but there is a race condition for sure

mirrord/agent/src/entrypoint/setup.rs
meowjesty: I'm not sure how to feel about it either.

I started with the `runtime.handle`, and changed cause it made it easier for me to search where the runtime is being used (I just `/_rt`, but if I `/runtime` there are a bunch more results).

I've added the warning. I think this one should be self-explanatory (like a lock guard or whatever), but it's nice to make sure people see it.

mirrord/agent/src/sniffer.rs
meowjesty: Added.

mirrord/agent/src/sniffer.rs
meowjesty: Changed.

mirrord/agent/src/sniffer.rs
meowjesty: I was scared of `sleep` being flaky :crying_cat_face: 

Changed.

mirrord/agent/src/entrypoint.rs
meowjesty: Changed.

mirrord/agent/src/entrypoint.rs
meowjesty: Deleted.

changelog.d/+592.changed.md
meowjesty: Removed.

mirrord/agent/env/src/envs.rs
```diff
@@ -38,6 +38,9 @@ pub const JSON_LOG: CheckedEnv<bool> = CheckedEnv::new("MIRRORD_AGENT_JSON_LOG")
 /// Enables IPv6 support in the agent.
 pub const IPV6_SUPPORT: CheckedEnv<bool> = CheckedEnv::new("AGENT_IPV6_ENV");
 
+/// Timeout for waiting on `BackgroundTask`s to finish.
+pub const EXIT_TIMEOUT: CheckedEnv<bool> = CheckedEnv::new("MIRRORD_AGENT_EXIT_TIMEOUT");
+
```
Razz4780: :wastebasket:

mirrord/agent/src/task/status.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            let _ = result_tx.send(result);
+        });
+
+        BgTaskStatus {
+            task_name,
+            result: result_rx.shared(),
+        }
+    }
+}
```
Razz4780, line 111: Wow. I did really implement mapping results by spawning extra tasks. My bad. The `BgTaskStatus` should look like this:

```rust
    pub struct BgTaskStatus {
        task_name: &'static str,
        result: Shared<BoxFuture<'static, Result<(), Arc<dyn Error + Send + Sync>>>>,
    }

    impl BgTaskStatus {
        pub async fn wait(&self) -> Result<(), AgentError> {
            match self.result.clone().await {
                Ok(()) => Ok(()),
                Err(error) => Err(AgentError::BackgroundTaskFailed {
                    task: self.task_name,
                    error,
                }),
            }
        }

        pub async fn wait_assert_running(&self) -> AgentError {
            match self.result.clone().await {
                Ok(()) => AgentError::BackgroundTaskFailed {
                    task: self.task_name,
                    error: Box::<dyn Error + Send + Sync>::from("task finished unexpectedly")
                        .into(),
                },
                Err(error) => AgentError::BackgroundTaskFailed {
                    task: self.task_name,
                    error,
                },
            }
        }
    }

    pub trait IntoStatus {
        fn into_status(self, task_name: &'static str) -> BgTaskStatus;
    }

    impl<E> IntoStatus for JoinHandle<Result<(), E>>
    where
        E: Error + Send + Sync + 'static,
    {
        fn into_status(self, task_name: &'static str) -> BgTaskStatus {
            BgTaskStatus {
                task_name,
                result: self
                    .map_ok_or_else(
                        |join_error| Err(Arc::new(join_error) as Arc<dyn Error + Send + Sync>),
                        |result| {
                            result.map_err(|error| Arc::new(error) as Arc<dyn Error + Send + Sync>)
                        },
                    )
                    .boxed()
                    .shared(),
            }
        }
    }

    impl IntoStatus for JoinHandle<()> {
        fn into_status(self, task_name: &'static str) -> BgTaskStatus {
            BgTaskStatus {
                task_name,
                result: self
                    .map_err(|join_error| Arc::new(join_error) as Arc<dyn Error + Send + Sync>)
                    .boxed()
                    .shared(),
            }
        }
    }
```

mirrord/agent/src/task.rs
```diff
@@ -0,0 +1,146 @@
...
+    #[tracing::instrument(level = Level::TRACE, ret, err)]
+    pub(crate) async fn spawn(
+        namespace: Option<RuntimeNamespace>,
+    ) -> Result<Self, AgentRuntimeError> {
+        let (result_tx, result_rx) = oneshot::channel();
+        let notify_drop = Arc::new(Notify::new());
+        let notify_cloned = notify_drop.clone();
+
+        let thread_logic = move || {
+            tracing::info!("thread_logic ->");
```
Razz4780: Pls remove all info logs from here

mirrord/agent/src/task.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.target_pid
+    }
+}
+
+impl Drop for BgTaskRuntime {
+    #[tracing::instrument(level = Level::DEBUG)]
+    fn drop(&mut self) {
+        if self.handle.metrics().num_alive_tasks() == 0 {
+            self.notify_drop.notify_one();
+        }
```
Razz4780: Why? The runtime should always be aborted when the handle is dropped.

mirrord/agent/src/task/status.rs
```diff
@@ -0,0 +1,137 @@
...
+impl BgTaskStatus {
+    /// Waits for the future to finish and returns its result.
+    ///
+    /// Should the future fail or panic, this function will return
+    /// [`AgentError::BackgroundTaskFailed`].
+    #[tracing::instrument(level = Level::DEBUG, fields(rt, rt_metrics), err)]
+    pub(crate) async fn wait(&self) -> Result<(), AgentError> {
+        let handle = tokio::runtime::Handle::current();
+        tracing::Span::current().record("rt", format!("{handle:?}"));
+        tracing::Span::current().record("rt_metrics", format!("{:?}", handle.metrics()));
```
Razz4780: Why are you logging metrics from the current runtime? 👀

mirrord/agent/src/task/status.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Waits for the future to finish and returns its result.
+    ///
+    /// This function always returns [`AgentError::BackgroundTaskFailed`]. Use it when the task is
+    /// not expected to finish yet, e.g. when we send a message to the `BackgroundTask` through its
+    /// channel, and `send` returns an error.
+    #[tracing::instrument(level = Level::DEBUG, fields(rt, rt_metrics), ret)]
+    pub(crate) async fn wait_assert_running(&self) -> AgentError {
+        let handle = tokio::runtime::Handle::current();
+        tracing::Span::current().record("rt", format!("{handle:?}"));
+        tracing::Span::current().record("rt_metrics", format!("{:?}", handle.metrics()));
```
Razz4780: Why are you logging metrics from the current runtime? 👀

mirrord/agent/src/entrypoint.rs
```diff
@@ -214,12 +215,15 @@ impl<Command> BackgroundTask<Command> {
...
     /// If the task is [`BackgroundTask::Disabled`], returns [`Ok`].
+    #[tracing::instrument(level = Level::TRACE, skip(self), fields(status), err(level = Level::DEBUG))]
     async fn wait(self) -> Result<(), AgentError> {
         let Self::Running(status, channel) = self else {
             return Ok(());
         };
         std::mem::drop(channel);
-        status.wait().await
+
+        tracing::Span::current().record("status", format!("{status:?}"));
```
Razz4780:
```suggestion
        let current_span = tracing::Span::current();
        if current_span.is_disabled().not() {
            current_span.record("status", format!("{status:?}"));
        }
```

mirrord/agent/src/entrypoint.rs
```diff
@@ -389,7 +393,7 @@ impl ClientConnectionHandler {
     /// Starts a loop that handles client connection and state.
     ///
     /// Breaks upon receiver/sender drop.
-    #[tracing::instrument(level = "trace", skip(self))]
+    #[tracing::instrument(level = Level::TRACE, skip(self), ret, err)]
```
Razz4780: This will log IO errors from the client connections. Revert.

mirrord/agent/src/task/status.rs
meowjesty: It was being dropped too soon, so these helped me.

mirrord/agent/src/task.rs
meowjesty: Changed the runtime in `state` to `Arc`.

mirrord/agent/src/task/status.rs
meowjesty: Removed.

mirrord/agent/src/task.rs
```diff
@@ -0,0 +1,133 @@
...
+/// **Attention**: Keep the runtime alive! In the [`Drop`] impl we call
+/// [`Notify::notify_one`] which will end the `BackgroundTask`. If you see an error like
+/// `BackgroundTaskFailed { task: "TcpSnifferTask", error: BgTaskPanicked }`, this means that
+/// the runtime was dropped before the `BackgroundTask` could start.
+///
+/// We keep the client connection (main loop) runtime separate from the background tasks to avoid
+/// potentially blocking the agent from exiting if some background task hangs. When the agent
+/// doesn't exit properly (i.e. it stays alive in the cluster, and the user has to manually kill
+/// it), we cannot start another agent due to the dirty iptables check.
+#[derive(Clone)]
```
Razz4780:
```suggestion
/// **Attention**: Keep the runtime alive! In the [`Drop`] impl we call
/// [`Notify::notify_one`] which will end the `BackgroundTask`. If you see an error like
/// `BackgroundTaskFailed { task: "TcpSnifferTask", error: BgTaskPanicked }`, this means that
/// the runtime was dropped before the `BackgroundTask` could start.
///
/// Because of the custom [`Drop`] implementation, this struct should not implement [`Clone`].
///
/// We keep the client connection (main loop) runtime separate from the background tasks to avoid
/// potentially blocking the agent from exiting if some background task hangs. When the agent
/// doesn't exit properly (i.e. it stays alive in the cluster, and the user has to manually kill
/// it), we cannot start another agent due to the dirty iptables check.
```

PR #3574, opened 2025-09-14
Update PingPong task reconnect attempts

DmitryDodzin:
Updated so if the PingPong will not attempt to request restart if not connected to operator.

### DIFF
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -124,6 +124,8 @@ impl IntProxy {
         let mut background_tasks: BackgroundTasks<MainTaskId, ProxyMessage, ProxyRuntimeError> =
             Default::default();
 
+        let agent_conn_reconnectable = agent_conn.reconnectable();
+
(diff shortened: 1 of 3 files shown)


mirrord/intproxy/src/lib.rs
```diff
@@ -141,7 +143,12 @@ impl IntProxy {
         // to requests that have a requirement on the mirrord-protocol version.
         background_tasks.suspend_messages(MainTaskId::LayerInitializer);
         let ping_pong = background_tasks.register_restartable(
-            PingPong::new(Self::PING_INTERVAL, Self::PING_PONG_MAX_RECONNECTS),
+            PingPong::new(
```
aviramha, line 146: Can we add a specific error? like ReconnectNotAvailable because operator isn't being used?

DmitryDodzin: It's just the ping-pong timeout error essentially, this error would not happen if operator is used (I can add some text about reconnecting in the message)

PR #3577, opened 2025-09-16
Aregh/cor 168 inject headers into responses that went through mirrord

MintSoup:
Inject `Mirrord-Agent` headers into responses that went through the agent. 

- `passed-through`: set when the request was not sent to the local app (perhaps because it didn't match the filters)
- `forwarded-to-client`: set when the request was sent to the local app

Please see `mirrord_agent::steal::test::utils::TestRequest::send`, an optional argument was needed and I used a wrapper function to avoid making changes to other parts of the code. Let me know if I should change it to Option<T>.

### DIFF
--- a/mirrord/agent/src/http.rs
+++ b/mirrord/agent/src/http.rs
@@ -19,6 +19,8 @@ pub mod extract_requests;
 pub mod filter;
 pub mod sender;
 
+pub(crate) const MIRRORD_AGENT_HTTP_HEADER_NAME: &str = "Mirrord-Agent";
+
 /// [`Response`] type with a boxed body.
 pub type BoxResponse = Response<BoxBody<Bytes, hyper::Error>>;
 
--- a/mirrord/agent/src/http.rs
+++ b/mirrord/agent/src/http.rs
@@ -19,6 +19,9 @@ pub mod extract_requests;
 pub mod filter;
 pub mod sender;
 
+/// When the corresponding config flag is enabled, a header with this
+/// name is injected into http responses. See
+/// `mirrord_config::agent::AgentConfig::inject_headers` for details.
 pub(crate) const MIRRORD_AGENT_HTTP_HEADER_NAME: &str = "Mirrord-Agent";
 
 /// [`Response`] type with a boxed body.
--- a/mirrord/agent/src/incoming/task.rs
+++ b/mirrord/agent/src/incoming/task.rs
@@ -8,6 +8,7 @@ use std::{
 
 use futures::{FutureExt, StreamExt, future::Shared};
 use hyper_util::rt::TokioIo;
+use mirrord_agent_env::envs;
 use tokio::sync::{
     mpsc::{self, error::TrySendError},
     oneshot,
@@ -58,6 +59,8 @@ pub struct RedirectorTask<R> {
     internal_tx: mpsc::Sender<InternalMessage>,
     /// For accepting redirected TLS connections.
     tls_store: StealTlsHandlerStore,
+    /// Configuration
+    config: Arc<RedirectorTaskConfig>,
 }
 
 impl<R> RedirectorTask<R>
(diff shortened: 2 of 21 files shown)


mirrord/config/src/agent.rs
```diff
@@ -424,6 +425,20 @@ pub struct AgentConfig {
     #[config(default = true)]
     pub passthrough_mirroring: bool,
 
+    /// #### agent.inject_headers {#agent-inject_headers}
```
Razz4780:
```suggestion
    /// ### agent.inject_headers {#agent-inject_headers}
```

mirrord/agent/src/steal/test.rs
```diff
@@ -425,6 +425,96 @@ async fn multiple_filtered_subscriptions(
     );
 }
 
+#[rstest]
+#[tokio::test(flavor = "current_thread")]
+#[timeout(Duration::from_secs(10))]
+async fn header_injection(
```
Razz4780, line 433:
```suggestion
/// Verifies that...
#[rstest]
#[tokio::test(flavor = "current_thread")]
#[timeout(Duration::from_secs(10))]
async fn header_injection(
```

mirrord/agent/src/http.rs
```diff
@@ -19,6 +19,8 @@ pub mod extract_requests;
 pub mod filter;
 pub mod sender;
 
+pub(crate) const MIRRORD_AGENT_HTTP_HEADER_NAME: &str = "Mirrord-Agent";
```
Razz4780, line 25: Could use a doc

mirrord/agent/src/steal/test.rs
```diff
@@ -425,6 +425,96 @@ async fn multiple_filtered_subscriptions(
...
+    http_kind: TestHttpKind,
+) {
+    use crate::util::ClientId;
+
+    let mut setup = TestSetup::new_http(http_kind).await;
+
+    // Enable header injection.
+    unsafe {
+        std::env::set_var("MIRRORD_AGENT_INJECT_HEADERS", "true");
+    }
```
Razz4780: This is `std::env::set_var` is unsafe not without a reason. And, since `cargo test` runs tests in multi-threaded environment, I think it is not safe to use it here :V

mirrord/agent/src/steal/api.rs
```diff
@@ -683,6 +688,18 @@ impl ClientConnectionState {
             *self = Self::HttpResponseReceived { body_provider };
         }
     }
+
+    /// Used for applying transformations on the response returned
+    /// from the client. Currently just inserts the mirrord agent
+    /// header.
+    fn modify_response<T>(&self, response: &mut Response<T>) {
```
Razz4780:
```suggestion
    fn modify_response<T>(response: &mut Response<T>) {
```
Does not need `&self` right now

Razz4780 (changes requested):
Only nits

mirrord/agent/src/steal/test.rs
Razz4780: Could use a doc

mirrord/agent/src/steal/api.rs
MintSoup: I intentionally made it that way no signature changes would be needed the signature if more transformations got added. Isn't it better to keep it like this? If not, `inject_headers` or something like that might be a better name.

Razz4780: I don't see any reason to accept unused arguments :eyes: 
If someone wants to add more transformations later, they can add the argument themselves

mirrord/agent/src/incoming/task.rs
```diff
@@ -436,6 +441,19 @@ impl<R> fmt::Debug for RedirectorTask<R> {
     }
 }
 
+pub struct RedirectorTaskConfig {
```
Razz4780, line 445:
```suggestion
#[derive(Debug, Clone, Copy)]
pub struct RedirectorTaskConfig {
```

mirrord/agent/src/incoming/task.rs
```diff
@@ -58,6 +59,8 @@ pub struct RedirectorTask<R> {
     internal_tx: mpsc::Sender<InternalMessage>,
     /// For accepting redirected TLS connections.
     tls_store: StealTlsHandlerStore,
+    /// Configuration
+    config: Arc<RedirectorTaskConfig>,
```
Razz4780: The config is just 1 byte right now, I think we can get rid of the `Arc`

Issue #3578, opened 2025-09-16
Improve UX when spawning the agent pod fails


Razz4780:
Part of the OSS mirrord flow is creating a k8s Job running a Pod with the mirrord-agent container ([here](https://github.com/metalbear-co/mirrord/blob/81e715586e3ff04307bf4ce50ea4d31e6a674a4e/mirrord/kube/src/api/container/job.rs#L31)).

Starting that pod might fail for multiple reasons. Our logic there is not the best, we basically only wait until `status.phase == "Running"`. When the agent pod cannot be spawned, most of the time this results in a generic timeout error (the timeout is enforced somewhere up the call stack).

We should:
1. Fail early if the agent pod moves to `Failed` phase. This can happen due to cluster conditions. We should extract `status.reason` and `status.message` and include them in the error message presented to the user.
2. Fail early if the agent pod moves to `Succeeded` phase. This should never happen, and should be reported to the user as a bug.
3. Fail early if the agent pod is deleted while in `Pending` phase. This usually means that the user does not have sufficient permissions to spawn the agent pod in the cluster. The error message presented to the user should mention [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) as a probable cause of the failure, and suggest trying out mirrord for Teams (similar to [this](https://github.com/metalbear-co/mirrord/blob/81e715586e3ff04307bf4ce50ea4d31e6a674a4e/mirrord/cli/src/error.rs#L317)).
4. For every 10s while the agent pod is stuck in the `Pending` phase, we should issue a `Progress::warning`. The warning should state that the agent pod startup takes longer than expected, and contain info about `status.containerStatuses.[].state` of the agent container. See [container states](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states) for reference.

iniw:
I did this for my take home already!!

PR #3581, opened 2025-09-17
Improve UX when spawning the agent pod fails

iniw:
Hello ＼(٥⁀▽⁀ )／! 

This fixes issue #3578.

I will now go through each of subtasks as described in the issue and explain where they are in this patch and why I did things the way I did,,,

> 1. Fail early if the agent pod moves to Failed phase. This can happen due to cluster conditions. We should extract status.reason and status.message and include them in the error message presented to the user.

This is one of the more straightforward changes, it corresponds to lines 104..=114 in the diff:
```rust
"Failed" => {
    pod_progress.failure(Some(&format!(
        "agent pod failed to initialize - '{}' ({})",
        status.message.as_deref().unwrap_or("<no message>"),
        status.reason.as_deref().unwrap_or("<unknown reason>")
    )));
    return Err(KubeApiError::PodFailedToInitialize(
        status.reason.clone(),
        status.message.clone(),
    ));
}
```
I decided to add a new error to `KubeApiError` to better describe what went wrong since I thought none of the existing errors quite captured this particular error path.

> 2. Fail early if the agent pod moves to Succeeded phase. This should never happen, and should be reported to the user as a bug.

This is handled right below the last step:
```rust
"Succeeded" => {
    pod_progress.failure(Some(
        "agent pod terminated prematurely - this is a bug, please report it!",
    ));
    return Err(KubeApiError::PodTerminatedPrematurely);
}
```
Again, decided to add a new error variant - these could be changed to the more generic `AgentPodNotRunning`. 

This branch also features a nice little message to the user.

Initially I implemented this with an `unreachable!` block, since this branch kinda implies that a logic bug somewhere else happened, but crashing mirrord in the middle of pod initialization seems scary so I decided against that.

> 3. Fail early if the agent pod is deleted while in Pending phase. This usually means that the user does not have sufficient permissions to spawn the agent pod in the cluster. The error message presented to the user should mention [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) as a probable cause of the failure, and suggest trying out mirrord for Teams (similar to [this](https://github.com/metalbear-co/mirrord/blob/81e715586e3ff04307bf4ce50ea4d31e6a674a4e/mirrord/cli/src/error.rs#L317)).

For context, here's the impl:
```rust
Ok(None) | Ok(Some(Err(_))) => {
    if is_pending {
        pod_progress.failure(Some("agent pod unexpectedly closed, \
        this likely means that you don't have the required permissions to spawn it. \
        look into your namespace's Pod Security admission controllers and try mirrord for Teams if the issue persists."));
        return Err(KubeApiError::AgentPodNotRunning);
    } else {
        break;
    }
}
```

This one I have some questions about. I'm not sure how to detect that the pod is deleted. I interpreted either getting a `None` event or a `Some(Err(err))` as the way to detect that, but I'm not fully convinced of this. The declaration of `agent_pod` on lines 62..=66 does seem to reinforce this theory:
```rust
let agent_pod = stream
    .next()
    .await
    .ok_or(KubeApiError::AgentPodNotRunning)?
    .map_err(|_| KubeApiError::AgentPodNotRunning)?;
```
This statement is also why I chose to return `AgentPodNotRunning` in this case.

This shows how I decided to check whether the pod is in the "Pending" phase or not: a mutable variable declared outside the loop that gets set to `phase == "Pending"` when a new event is received.

> 4. For every 10s while the agent pod is stuck in the Pending phase, we should issue a Progress::warning. The warning should state that the agent pod startup takes longer than expected, and contain info about status.containerStatuses.[].state of the agent container. See [container states](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states) for reference.

For this I added wrapped the `stream.next()` in a `tokio::time::timeout` call and shoved the whole thing into a `loop`:

```rust
loop {
    match timeout(Duration::from_secs(10), stream.next()).await {
    	// ...
    	
    	// Timeout elapsed
        Err(_) => {
            if is_pending {
                pod_progress.warning("agent pod initialization is taking longer than expected");
            }
        }
    }
}    
```
 
> and contain info about status.containerStatuses.[].state of the agent container

I didn't implement this part since I don't know which entry in the `containerStatuses` list belongs to the container that we care about.

It's also a bit annoying to implement because inside of this branch we don't have access to `status` so, much like `is_pending`, it'd have to be some sort of `Option<String>` declared outside of the loop that gets set to the appropriate status string when a new event arrives. This made the implementation a bit harder to read and I wasn't even sure that it was the right approach, so I ended up leaving it out for now, but do plan on adding it back as soon as I know which index to use for `containerStatuses`.

---------------------

Opened as a draft to share my progress, ask questions and get feedback on the current direction. Please let me know what you think :)

The biggest issue I faced while writing this was *testing* it. I don't really know how to consistently reach the error path in this case. I tried several things, but setting a breakpoint on where the pod gets created then pausing the `minikube` session with `minikube pause` and resuming from the breakpoint seemed to be the best strategy, even though it barely ever worked. When it did though I did manage to reach the timeout branch (yay) but I never got the `Ok(None) | Ok(Some(Err(_)))` branch, so I don't really know if that's *really* what happens when the pod closes.

### DIFF
--- a/mirrord/kube/src/api/container/job.rs
+++ b/mirrord/kube/src/api/container/job.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, time::Duration};
 
 use futures::StreamExt;
 use k8s_openapi::api::{
@@ -12,8 +12,8 @@ use kube::{
 };
 use mirrord_config::agent::AgentConfig;
 use mirrord_progress::Progress;
-use tokio::pin;
-use tracing::debug;
+use tokio::{pin, time::timeout};
+use tracing::warn;
 
 use crate::{
     api::{
@@ -83,15 +83,62 @@ where
 
     let mut pod_progress = progress.subtask("waiting for pod to be ready...");
 
-    while let Some(Ok(pod)) = stream.next().await {
-        let Some(phase) = pod.status.as_ref().and_then(|status| status.phase.as_ref()) else {
-            continue;
-        };
+    let mut is_pending = false;
+    loop {
+        match timeout(Duration::from_secs(10), stream.next()).await {
+            Ok(Some(Ok(pod))) => {
+                let Some(ref status) = pod.status else {
+                    continue;
+                };
 
-        debug!(?phase, "Agent pod changed");
+                let Some(ref phase) = status.phase else {
+                    continue;
+                };
 
-        if phase == "Running" {
-            break;
+                is_pending = phase == "Pending";
+
+                // Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+                match phase.as_str() {
+                    "Pending" => continue,
+                    "Running" => break,
+                    "Failed" => {
+                        pod_progress.failure(Some(&format!(
+                            "agent pod failed to initialize - '{}' ({})",
+                            status.message.as_deref().unwrap_or("<no message>"),
+                            status.reason.as_deref().unwrap_or("<unknown reason>")
+                        )));
+                        return Err(KubeApiError::PodFailedToInitialize(
+                            status.reason.clone(),
+                            status.message.clone(),
+                        ));
+                    }
+                    "Succeeded" => {
+                        pod_progress.failure(Some(
+                            "agent pod terminated prematurely - this is a bug, please report it!",
+                        ));
+                        return Err(KubeApiError::PodTerminatedPrematurely);
+                    }
+                    phase => warn!("Unhandled pod phase '{}'", phase),
+                }
+            }
+
+            Ok(None) | Ok(Some(Err(_))) => {
+                if is_pending {
+                    pod_progress.failure(Some("agent pod unexpectedly closed, \
+                    this likely means that you don't have the required permissions to spawn it. \
+                    look into your namespace's Pod Security admission controllers and try mirrord for Teams if the issue persists."));
+                    return Err(KubeApiError::AgentPodNotRunning);
+                } else {
+                    break;
+                }
+            }
+
+            // Timeout elapsed
+            Err(_) => {
+                if is_pending {
+                    pod_progress.warning("agent pod initialization is taking longer than expected");
+                }
+            }
         }
     }
 
--- a/mirrord/kube/src/error.rs
+++ b/mirrord/kube/src/error.rs
@@ -79,6 +79,16 @@ pub enum KubeApiError {
 
     #[error(transparent)]
     InvalidBackoff(#[from] InvalidBackoff),
+
+    /// Failed to initialize the job agent's pod.
+    /// This will happen when the pod moves to the `"Failed"` stage while initializing.
+    #[error("Pod failed to spawn with reason `{0:?}` and message `{1:?}`.")]
+    PodFailedToInitialize(Option<String>, Option<String>),
+
+    /// The job agent's pod finished running all of it's containers without notifying us of the
+    /// `"Running"` stage. This is a bug and should never happen.
+    #[error("Pod terminated before running")]
+    PodTerminatedPrematurely,
 }
 
 impl KubeApiError {
--- a/changelog.d/3578.added.md
+++ b/changelog.d/3578.added.md
@@ -1,0 +1,1 @@
+Improved UX when spawning the agent pod fails or takes a long time


iniw:
I have pushed a new commit reimplementing step 4.

Previously it was a simple `timeout` against the event stream, which didn't actually produce the desired behavior: instead of warning the user after 10s of being stuck on the "Pending" state, it warned the user after 10s _without an event_, which are two very different things.

Now it's implement as a proper timer that gets `select`ed against the event stream and warns the user when expired.

The diff is unfortunately a bit overwhelming because everything got indented to the right but I can assure that everything else stayed the same :)

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+    let mut warned_user = false;
+    let pending_timer = sleep(Duration::from_secs(10));
+    tokio::pin!(pending_timer);
+
+    loop {
+        tokio::select! {
+            _ = &mut pending_timer, if is_pending && !warned_user => {
+                pod_progress.warning("agent pod initialization is taking longer than expected");
+                warned_user = true;
+            }
```
Razz4780: This will warn the user only once, after 10s. I'd like it to keep issuing the warning. I suggest using `tokio::time::interval`, like this:

```rust
let mut warn_interval = tokio::time::interval(Duration::from_secs(20));
warn_interval.tick().await; // first tick is instant

loop {
    tokio::select! {
        _ = warn_interval.tick() => {
            // issue warning
        },
        event = stream.next() => {
            // ...
        },
    }
}
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+    // We want to warn the user if the pod has been in the "Pending" phase for 10+ seconds
+    let mut is_pending = false;
+    let mut warned_user = false;
+    let pending_timer = sleep(Duration::from_secs(10));
+    tokio::pin!(pending_timer);
+
+    loop {
+        tokio::select! {
+            _ = &mut pending_timer, if is_pending && !warned_user => {
+                pod_progress.warning("agent pod initialization is taking longer than expected");
```
Razz4780: The agent's entry in `status.container_statuses` is the one with `name: "mirrord-agent"`.
You'll probably need to store the last observed pod state in an `Option` outside of this loop to inspect the statuses here

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+                warned_user = true;
+            }
+            event = stream.next() => {
```
Razz4780, line 112: I think to detect the deleted case, you'll need to change how the stream is created in line 59. Namely, get rid of the `.applied_objects()` thingy. Then, the stream will yield [`Events`](https://docs.rs/kube/latest/kube/runtime/watcher/enum.Event.html). Without going into kube.rs details:
1. `InitApply(pod)` and `Apply(pod)` -> pod changed state
2. `Delete(pod)` -> pod was deleted
3. `Init` and `InitDone` -> ignore

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+                                    status.message.clone(),
+                                ));
+                            }
+                            "Succeeded" => {
+                                pod_progress.failure(Some(
+                                    "agent pod terminated prematurely - this is a bug, please report it!",
+                                ));
+                                return Err(KubeApiError::PodTerminatedPrematurely);
+                            }
+                            phase => warn!("Unhandled pod phase '{}'", phase),
```
Razz4780: This should be a fatal error as well

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+                        }
+                    }
+                    Some(Err(_)) | None => {
```
Razz4780, line 159: This should always be a fatal error, either "watch stream failed: {error}" or "watch stream unexpectedly finished"

mirrord/kube/src/error.rs
```diff
@@ -79,6 +79,16 @@ pub enum KubeApiError {
...
+
+    /// Failed to initialize the job agent's pod.
+    /// This will happen when the pod moves to the `"Failed"` stage while initializing.
+    #[error("Pod failed to spawn with reason `{0:?}` and message `{1:?}`.")]
+    PodFailedToInitialize(Option<String>, Option<String>),
+
+    /// The job agent's pod finished running all of it's containers without notifying us of the
+    /// `"Running"` stage. This is a bug and should never happen.
+    #[error("Pod terminated before running")]
+    PodTerminatedPrematurely,
```
Razz4780: In terms of errors, I'd do one generic with a string message, and one for the case where agent was deleted, like:

```rust
enum KubeApiError {
    ... // previous variants
    /// Generic failure of the agent pod startup routine.
    #[error("Failed to wait for the agent pod to start: {0}")]
    AgentPodStartError(String),

    /// Spawned agent pod was deleted during startup.
    #[error("Agent pod was unexpectedly deleted")]
    AgentPodDeleted,
}
```

And remove the `AgentPodNotRunning` variant from the enum entirely - it doesn't contain any message, so it kinda drops the exact failure reason.

The case where the agent pod is deleted should be moved to it's own variant, because then we'll treat it differently when converting `KubeApiError` to `CliError` in the `mirrord` crate (see `CliError::friendler_error_or_else`). It should have it's own variant in the `CliError` enum, with an extra diagnostic message (like `CliError::FeatureRequiresOperatorError`). `miette`'s diagnostics look nice in the output  : D

Error handling here is cursed, I know

mirrord/kube/src/api/container/job.rs
Razz4780: I'd also include the total elapsed time in the message

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
+                        if is_pending {
+                            pod_progress.failure(Some("agent pod unexpectedly closed, \
+                            this likely means that you don't have the required permissions to spawn it. \
+                            look into your namespace's Pod Security admission controllers and try mirrord for Teams if the issue persists."));
```
Razz4780: mfT should be mentioned in `miette`'s diagnostics, see my comment for the new `KubeApiError` variants

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +83,67 @@ where
...
-            break;
+    // We want to warn the user if the pod has been in the "Pending" phase for 10+ seconds
+    let mut is_pending = false;
+    let mut warned_user = false;
+    let pending_timer = sleep(Duration::from_secs(10));
+    tokio::pin!(pending_timer);
+
+    loop {
+        tokio::select! {
+            _ = &mut pending_timer, if is_pending && !warned_user => {
```
Razz4780:
```suggestion
            _ = &mut pending_timer, if is_pending && !warned_user => {
```
I think in the final version, this loop will aways exit when the pod exits the `Pending` phase. So no need to store `is_pending`

mirrord/kube/src/api/container/job.rs
Razz4780: Also, I reconsidered - 10s will probably be too annoying. Let's do 20s

Razz4780 (changes requested):
Nice progress, left some comments ^^

mirrord/kube/src/api/container/job.rs
iniw: Previously the `"Unknown"` phase was explicitly not handled, but because it's now a fatal error path I've handled it alongside `"Pending"`, so it just `continue`s the loop. This path should now only be hit if more phases get added.

mirrord/kube/src/api/container/job.rs
```diff
@@ -56,14 +61,18 @@ where
...
 
     let agent_pod = stream
+        .as_mut()
+        .applied_objects()
         .next()
         .await
-        .ok_or(KubeApiError::AgentPodNotRunning)?
-        .map_err(|_| KubeApiError::AgentPodNotRunning)?;
+        .ok_or(KubeApiError::AgentPodStartError(
+            "Watch stream failed".to_owned(),
```
Razz4780:
```suggestion
            "watch stream unexpectedly finished".to_owned(),
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -56,14 +61,18 @@ where
...
+        ))?
+        .map_err(|err| KubeApiError::AgentPodStartError(err.to_string()))?;
```
Razz4780:
```suggestion
        .map_err(|err| KubeApiError::AgentPodStartError(format!("watch stream failed: {err})))?;
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -116,6 +186,37 @@ where
...
+            } else if state.waiting.is_some() {
+                Some("waiting")
+            } else {
+                // This should be unreachable - the docs say that if the state is unknown then the
+                // active state will be 'waiting', but it doesn't sound nice to
+                // crash if another state ever gets added.
+                None
+            }
+        })
+}
```
Razz4780:
```suggestion
/// Tries finding mirrord-agent's container in `status.container_statuses` and returns a string
/// representing it's state.
fn find_agent_container_state(status: &Option<PodStatus>) -> String {
    status
        .as_ref()
        .into_iter()
        .flat_map(|status| status.container_statuses.as_deref().unwrap_or_default())
        .find(|status| status.name == "mirrord-agent")
        .and_then(|status| status.state.as_ref())
        .and_then(|state| {
            if state.running.is_some() {
                Some("running".into())
            } else if let Some(terminated) = &state.terminated {
                Some(format!(
                    "terminated ({})",
                    terminated.reason.as_deref().unwrap_or("no reason found")
                ))
            } else {
                state.waiting.as_ref().map(|waiting| {
                    format!(
                        "waiting ({})",
                        waiting.reason.as_deref().unwrap_or("no reason found")
                    )
                })
            }
        })
        .unwrap_or_else(|| "not found".into())
}
```
Less cloning, and reason included ^^

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
-            break;
+                    Some(Ok(Event::Delete(_))) => {
+                        pod_progress.failure(Some("agent pod got unexpectedly deleted"));
+                        return Err(KubeApiError::AgentPodDeleted);
+                    }
+
+                    Some(Ok(Event::Init | Event::InitDone)) => continue,
+
+                    Some(Err(_)) | None => {
+                            pod_progress.failure(Some("watch stream has unexpectedly failed"));
```
Razz4780:
```suggestion
                            pod_progress.failure(Some("watch stream has failed"));
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
+                                )));
+                                return Err(KubeApiError::AgentPodStartError(
+                                    status.message.clone().unwrap_or_else(|| "<no message>".to_owned()),
+                                ));
+                            }
+                            "Succeeded" => {
+                                pod_progress.failure(Some(
+                                    "agent pod terminated prematurely - this is a bug, please report it!",
+                                ));
+                                return Err(KubeApiError::AgentPodStartError("Skipped 'Running' phase".to_owned()));
```
Razz4780:
```suggestion
                                return Err(KubeApiError::AgentPodStartError("agent pod unexpectedly finished".to_owned()));
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
+                            "Running" => break,
+                            "Failed" => {
+                                pod_progress.failure(Some(&format!(
+                                    "agent pod failed to initialize - '{}' ({})",
+                                    status.message.as_deref().unwrap_or("<no message>"),
+                                    status.reason.as_deref().unwrap_or("<unknown reason>")
+                                )));
+                                return Err(KubeApiError::AgentPodStartError(
+                                    status.message.clone().unwrap_or_else(|| "<no message>".to_owned()),
+                                ));
```
Razz4780:
```suggestion
                                return Err(KubeApiError::AgentPodStartError(format!(
                                    "agent pod failed with reason {}: {}",
                                    status.reason.as_deref().unwrap_or("<unknown reason>"),
                                    status.message.as_deref().unwrap_or("<no message>"),
                                )));
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
+                            }
+                            "Succeeded" => {
+                                pod_progress.failure(Some(
+                                    "agent pod terminated prematurely - this is a bug, please report it!",
+                                ));
+                                return Err(KubeApiError::AgentPodStartError("Skipped 'Running' phase".to_owned()));
+                            }
+                            phase => {
+                                pod_progress.failure(Some(&format!("agent pod moved to an unhandled phase - '{phase}'")));
+                                return Err(KubeApiError::AgentPodStartError(format!("Unhandled pod phase '{phase}'")));
```
Razz4780:
```suggestion
                                return Err(KubeApiError::AgentPodStartError(format!("agent pod entered an unexpected phase '{phase}'")));
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
+                            },
+                        }
+                    }
 
-        if phase == "Running" {
-            break;
+                    Some(Ok(Event::Delete(_))) => {
+                        pod_progress.failure(Some("agent pod got unexpectedly deleted"));
```
Razz4780:
```suggestion
                        pod_progress.failure(Some("agent pod was unexpectedly deleted"));
```

mirrord/kube/src/api/container/job.rs
```diff
@@ -83,15 +92,76 @@ where
...
+                        return Err(KubeApiError::AgentPodDeleted);
+                    }
+
+                    Some(Ok(Event::Init | Event::InitDone)) => continue,
+
+                    Some(Err(_)) | None => {
+                            pod_progress.failure(Some("watch stream has unexpectedly failed"));
+                            return Err(KubeApiError::AgentPodStartError("Stream failed".to_owned()));
```
Razz4780:
```suggestion
                            return Err(KubeApiError::AgentPodStartError("watch stream failed".to_owned()));
```

Razz4780 (changes requested):
Some more nits, and we're good

mirrord/kube/src/api/container/job.rs
iniw: Cool `into_iter` + `flat_map` trick!

mirrord/kube/src/api/container/job.rs
iniw: Also moved this to an `ok_or_else` call to avoid allocating the string.

mirrord/kube/src/api/container/job.rs
iniw: Factored this out into a variable to reuse the same `format!` allocations as the user-facing message since they ended up being so similar

mirrord/kube/src/api/container/job.rs
iniw: Same thing here, moved to a shared variable since the strings were nearly identical, which avoids a `format!` alloc.

Razz4780 (approved):

PR #3582, opened 2025-09-17
Add a nix devshell

iniw:
This adds a nix devshell containing all the required packages to run mirrord and the e2e tests.

Contains both aarc64 and x86_64 toolchains on darwin to build universal binaries


aviramha:
Can you add changelog entry/ (should be marked internal)

PR #3587, opened 2025-09-19
Verify correct Prometheus metrics in SQS E2E tests

t4lz:
Closes https://linear.app/metalbear/issue/INT-201/test-sqs-prometheus-metrics-in-e2e

https://github.com/metalbear-co/operator/pull/1063 has to be merged first, so that if the operator is updated to a new mirrord crate version, CI doesn't fail.

And I also want to verify the test passes on CI before merging. Testing in this manual workflow with a test branch https://github.com/metalbear-co/operator/actions/runs/17865079044

### DIFF
--- a/tests/src/utils/sqs_resources.rs
+++ b/tests/src/utils/sqs_resources.rs
@@ -18,6 +18,7 @@ use aws_sdk_sqs::types::{
     QueueAttributeName::{FifoQueue, MessageRetentionPeriod},
 };
 use futures_util::FutureExt;
+use http_body_util::BodyExt;
 use k8s_openapi::{
     api::{
         apps::v1::Deployment,
@@ -28,7 +29,10 @@ use k8s_openapi::{
     },
     apimachinery::pkg::apis::meta::v1::ObjectMeta,
 };
-use kube::{api::PostParams, Api, Client, Resource};
+use kube::{
+    api::{ListParams, PostParams},
+    Api, Client, Resource, ResourceExt,
+};
 use mirrord_operator::{
     crd::{
         MirrordSqsSession, MirrordWorkloadQueueRegistry, MirrordWorkloadQueueRegistrySpec,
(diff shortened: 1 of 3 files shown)


tests/src/utils/sqs_resources.rs
```diff
@@ -144,6 +149,97 @@ impl SqsTestResources {
             tokio::time::sleep(Duration::from_secs(2)).await;
         }
     }
+
+    fn get_required_metrics_strings(&self) -> Vec<String> {
+        vec![
+            format!(
+                r#"mirrord_read_sqs_messages_count{{original_queue="{}"}} 6"#,
```
meowjesty, line 156: Where are these numbers coming from?

And won't they change if we add more tests that are SQS related?

tests/src/utils/sqs_resources.rs
```diff
@@ -144,6 +149,97 @@ impl SqsTestResources {
...
+            }
+        });
+
+        let http_req = hyper::Request::builder()
+            .uri("/metrics")
+            .header("Connection", "close")
+            .header("Host", "localhost")
+            .method("GET")
+            .body(http_body_util::Empty::<hyper::body::Bytes>::new())
+            .unwrap();
```
meowjesty, line 222: Why not use `reqwest`?

tests/src/utils/sqs_resources.rs
t4lz: The test sends 6 messages to each queue. Each test run gets its own queues, and this metric is per queue, so adding more tests won't affect these numbers.

tests/src/utils/sqs_resources.rs
t4lz: Just copied from kube-rs's example for sending an HTTP request over port forwarding. Is it better to use reqwest?

meowjesty (approved):
👍

tests/src/utils/sqs_resources.rs
meowjesty: Nah it's fine, was wondering if there was any special reason.

PR #3588, opened 2025-09-19
mirrord ci api key command

0x00A5:


mirrord/auth/src/credentials.rs
```diff
@@ -105,6 +110,67 @@ impl LicenseValidity for NaiveDate {
...
+
+impl CiApiKey {
+    const V1_PREFIX: &str = "mci-v1:";
+
+    /// V1 encoding:
+    /// 1. Serialize [`Credentials`] with `bincode` (standard config).
+    /// 2. Compress with `zlib` with highest compression level (9).
+    /// 3. Encode with URL-safe base64 without padding.
+    /// 4. Prefix with `mci-v1:`, mirrord CI version 1.
+    pub fn encode(&self) -> Result<String, ApiKeyError> {
```
Razz4780: This seems unnecessarily complex 👀

Razz4780 (commented):
Wdyt about implementing certification with:
1. a new resource for user credentials
2. a `NewOperatorFeature` variant for checking whether the operator serves endpoints for the new resource

The resource would work just like `TargetCrd`/`CopyTargetCrd`, so k8s api requests would be handled entirely in the operator (no etcd).

```rust
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
pub struct MirrordOperatorUserCredentialsStatus {
    // `Certificate` type moved from `mirrord-auth`, serialized/deserialized in PEM
    pub certificate: Certificate,
    // `KeyPair` type moved from `mirrord-auth`, serialized/deserialized in PEM
    pub key_pair: KeyPair,
}

#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
#[kube(
    group = "operator.metalbear.co",
    version = "v1",
    kind = "MirrordOperatorUserCredentials",
    status = "MirrordOperatorUserCredentialsStatus",
    namespaced
)]
pub struct MirrordOperatorUserCredentialsSpec {
    // New wrapper over x509 certification request, serialized/deserialized in PEM
    pub csr: CertificationRequest,
    pub kind: String,
}
```

mirrord/auth/src/credentials.rs
0x00A5: I chose bincode + compression in order to minimize the length of the api key. The compression step shrinks the over all size by about 40% and was done very fast. Any better options?

0x00A5:
@Razz4780 we had a discussion [here](https://metalbear.slack.com/archives/C08TYT624SV/p1758601345146959?thread_ts=1758567040.433619&cid=C08TYT624SV) on Slack and decided to have a new endpoint that handles cases for both CI and regular. I have opened a PR for it [here](https://github.com/metalbear-co/operator/pull/1066).
I didn't follow what's the advantage of having a new resource for this, does the current approach store anything in etcd?
And I guess my main concern is whether we can have endpoint input that indicates if the request is for CI or regular session. The endpoint needs to perform some license type check (type=enterprise) if it is CI.

mirrord/auth/src/credentials.rs
Razz4780: Ah nvm. I was confused, sorry. Since this is printed to the user, the size is kinda important. All good

mirrord/auth/Cargo.toml
```diff
@@ -34,14 +35,18 @@ client = [
 ]
 
 [dependencies]
+base64.workspace = true
+bincode.workspace = true
 chrono.workspace = true
 whoami = { version = "1", optional = true }
 home = { workspace = true, optional = true }
 pem.workspace = true
+flate2 = { version = "1", default-features = false, features = ["rust_backend"] }
```
Razz4780: This should be optional and hidden behind the `client` feature

mirrord/auth/src/credentials.rs
```diff
@@ -105,6 +110,67 @@ impl LicenseValidity for NaiveDate {
...
+        match self {
+            CiApiKey::V1(credentials) => {
+                let bytes =
+                    bincode::serde::encode_to_vec(credentials, bincode::config::standard())?;
```
Razz4780: `Serialize` implementation uses PEM format for both the cert and the key pair. And since PEM is base64, we already get 33% size increase.

Maybe we should implement `Encode`/`Decode` for the `Credentials` struct? The crypto primitives inside have methods for encoding/decoding flat byte arrays (DER). Wdyt?

The rest would stay the same, so:

1. `bincode::encode_to_vec` on credentials (inside: `cert.encode_der().encode(bincode_encoder);, key_pair.key_pair.to_pkcs8_one_asymmetric_key_der().as_ref().encode(bincode_encoder)`)
2. compression (might actually not be that efficient anymore, need to verify if still needed)
3. base64
4. prefix

mirrord/cli/src/config.rs
```diff
@@ -153,6 +153,9 @@ pub(super) enum Commands {
 
     /// Subscribe to the mirrord newsletter
     Newsletter,
+
+    /// Execute a command related to mirrord CI.
+    Ci(Box<CiArgs>),
```
Razz4780, line 159: This should be probably be hidden for now, as the key is not used anywhere

Razz4780 (changes requested):

mirrord/auth/src/credentials.rs
0x00A5: Ah, I missed the custom serde for certificate. Yeah, skipping pem and directly go to bincode sounds nicer.

mirrord/auth/Cargo.toml
0x00A5: Hmm, encoding and decoding between credentials and ci api key doesn't rely on the `client` feature though.

Razz4780: But it could, right? The operator will not be doing the conversion

mirrord/auth/src/credentials.rs
0x00A5: With custom encode decode impl, the api key length shrinked from ~1049 to ~773.

0x00A5: With custom encode decode impl, but with no compression, the length is ~943. Still another 20% shorter, I am gonna keep the compression then.

mirrord/auth/src/credential_store.rs
```diff
@@ -115,38 +120,54 @@ impl CredentialStore {
...
     where
-        R: Resource + Clone + Debug,
-        R: for<'de> Deserialize<'de>,
-        R::DynamicType: Default,
+        Old: Resource + Clone + Debug,
+        Old: for<'de> Deserialize<'de>,
+        Old::DynamicType: Default,
+        New: Clone + Debug + SigningRequest + SigningResponse + Serialize,
+        New: for<'de> Deserialize<'de>,
+        New::DynamicType: Default,
```
0x00A5, line 136: This is a bit ugly until we sack the old subresource endpoint. But I can't think of a way to work around it, as both are needed to abstract away `Resource` definition. @Razz4780

mirrord/auth/Cargo.toml
0x00A5: Moved the encode decode method behind client feature.

mirrord/cli/src/config.rs
0x00A5: Fixed

mirrord/operator/src/setup.rs
```diff
@@ -705,6 +706,16 @@ impl OperatorClusterRole {
...
+            PolicyRule {
+                api_groups: Some(vec![
+                    MirrordClusterOperatorUserCredential::group(&()).into_owned(),
+                ]),
+                resources: Some(vec![
+                    MirrordClusterOperatorUserCredential::plural(&()).into_owned(),
+                ]),
+                verbs: vec!["create".to_owned()],
+                ..Default::default()
+            },
```
Razz4780: This role describes the operator's capabilities. The new rule be added to `OperatorClusterUserRole` instead.

mirrord/operator/src/crd.rs
```diff
@@ -405,6 +406,7 @@ impl Display for NewOperatorFeature {
                 "SQS queue splitting without copy target"
             }
             NewOperatorFeature::MySqlBranching => "MySQL branching",
+            NewOperatorFeature::CertificateSigning => "client certificate signing",
```
Razz4780: This might be a bit confusing, since the operator supports certificate signing right now. Wdyt about "extendable user credentials"?

mirrord/operator/src/crd.rs
```diff
@@ -728,3 +730,33 @@ pub struct MirrordSqsSessionSpec {
...
+pub struct MirrordClusterOperatorUserCredentialStatus {
+    pub certificate: String,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Default)]
+#[serde(rename_all = "camelCase")]
+pub enum UserCredentialKind {
+    #[default]
+    Regular,
+    Ci,
```
Razz4780, line 761:
```suggestion
    Ci,
    #[schemars(skip)]
    #[serde(other)]
    Unknown,
```

For nicer error when we add some more, and the installed operator version does not recognize them

mirrord/auth/src/credential_store.rs
Razz4780: This is fine for know, I wouldn't worry about it.
There's a task to tidy up this crate, and hide all IO behind traits anyway

mirrord/operator/src/crd.rs
0x00A5: Yeah, sounds good

mirrord/operator/src/setup.rs
0x00A5: My bad, moved to user role.

mirrord/operator/src/crd.rs
0x00A5: Changed to the suggested name.

mirrord/operator/src/crd.rs
0x00A5: Added the unknown variant.

PR #3589, opened 2025-09-21
Detect and abort nested mirrord calls

### LINKED ISSUE #2806: Detect mirrord running mirrord
Some of our users integrate mirrord into their build/run scripts, for example in package.json they do 'mirrord exec ..'
Problem is when they also use the IDE they sometime enable it there causing mirrord to run within mirrord, so second mirrord usually fails because remote service account doesn't have permissions, and if it works then it's probably causing slowness.
We should be able to detect such scenario from cli and warn/abort
I'm leaning towards about.

MintSoup:
Fixes #2806

### DIFF
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -453,6 +453,10 @@ pub(crate) enum CliError {
         You can get started with mirrord for Teams at this link: https://metalbear.com/mirrord/docs/overview/teams/?utm_source=errreqop&utm_medium=cli"
     ))]
     AgentPodDeleted,
+
+    #[error("Nested mirrord is not supported")]
+    #[diagnostic(help("Check your IDE and build settings and ensure you are not running mirrord within mirrord."))]
+    NestedExec,
 }
 
 impl CliError {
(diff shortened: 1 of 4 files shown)


mirrord/cli/src/error.rs
```diff
@@ -453,6 +453,12 @@ pub(crate) enum CliError {
         You can get started with mirrord for Teams at this link: https://metalbear.com/mirrord/docs/overview/teams/?utm_source=errreqop&utm_medium=cli"
     ))]
     AgentPodDeleted,
+
+    #[error("Nested mirrord is not supported")]
+    #[diagnostic(help(
+        "Check your IDE and build settings and ensure you are not running mirrord within mirrord."
+    ))]
```
Razz4780:
```suggestion
    #[error("Detected mirrord being run within mirrord")]
    #[diagnostic(help(
        "Running mirrord within mirrord is likely to fail or introduce severe slowness. \
        If you are using a mirrord IDE plugin, please check that your build/run script \
        does not invoke mirrord."
    ))]
```

Razz4780 (approved):
Nice. Left a suggestion for the error message.

PR #3605, opened 2025-09-30
Improve MirrordClusterSession crd

Razz4780:


mirrord/operator/src/crd/session.rs
```diff
@@ -6,63 +7,80 @@ use serde::{Deserialize, Serialize};
...
 }
 
-/// Resources needed to report session metrics to the mirrord Jira app
-#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+/// Describes the reason for with a mirrord session was closed.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
 #[serde(rename_all = "camelCase")]
-pub struct JiraMetricsResources {
-    /// The Jira webhook URL, used to update total session time in the mirrord Jira app
-    pub jira_webhook_url: String,
```
0x00A5, line 64: This url is dropped in the new `SessionJiraMetrics` struct, is it no longer needed?

Razz4780: It should not be stored in the session resource, it should be taken from current operator configuration when needed

mirrord/cli/src/config.rs
```diff
@@ -777,15 +777,6 @@ pub(super) struct OperatorSetupParams {
...
     pub(super) application_auto_pause: bool,
 
-    /// Enable MirrordClusterSession CRD's (curretly experimental and requires operator compiled
-    /// with experimental flag).
-    #[arg(
-        long = "experimental-statefull-sessions",
-        default_value_t = false,
-        hide = true
-    )]
-    pub(super) stateful_sessions: bool,
```
0x00A5, line 821: This is moved to operator config I suppose? Or is it always on after it gets merged.

Razz4780: Yeah, it's always on

PR #3613, opened 2025-10-04
Add methods for getting env var from target

0x00A5:


mirrord/config/src/feature/database_branches.rs
```diff
@@ -79,6 +79,71 @@ pub enum DatabaseBranchConfig {
...
+pub enum MysqlBranchCopyConfig {
+    Empty {
+        tables: Option<HashMap<String, MysqlBranchTableCopyConfig>>,
+    },
+
+    Schema {
+        tables: Option<HashMap<String, MysqlBranchTableCopyConfig>>,
+    },
+
+    Data,
```
t4lz: IDK if `data` is clear enough about what this option means. Would go with `all` or `everything`.

mirrord/operator/src/crd/mysql_branching.rs
```diff
@@ -85,3 +94,77 @@ pub struct SessionInfo {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct BranchCopyConfig {
+    /// The default copy mode for the branch.
+    pub mode: BranchCopyMode,
+
+    /// An optional list of tables whose schema and data will be copied based on their
+    /// table level copy config. Only compatible with `Empty` and `Schema` copy mode.
+    pub tables: BTreeMap<String, TableCopyConfig>,
```
t4lz: Shouldn't this be an Option?

mirrord/operator/src/crd/mysql_branching.rs
```diff
@@ -85,3 +94,77 @@ pub struct SessionInfo {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum BranchCopyMode {
+    /// Create an empty database only.
+    Empty,
+    /// Create a database with all tables' schema copied from the source database.
+    Schema,
+    /// Create a database and copy all tables' schema and data fro mthe source database.
+    Data,
```
t4lz: maybe document that specifying `tables` is not compatible with this mode, and say what happens when specified (ignored? error?).

mirrord/config/src/feature/database_branches.rs
0x00A5: OK, changing it back to all.

mirrord/operator/src/crd/mysql_branching.rs
0x00A5: Yeah, I can make it Option<T> and skip serialization if none

mirrord/operator/src/crd/mysql_branching.rs
0x00A5: Sounds good, it is important to highlight this is not compatible with `tables`.

PR #3631, opened 2025-10-11
Improve DB branching config

0x00A5:


mirrord/config/src/feature/database_branches.rs
```diff
@@ -65,23 +68,35 @@ impl DatabaseBranchesConfig {
...
-    /// ### feature.db_branches.name {#feature-db_branches-name}
+    /// ### feature.db_branches.base.name {#feature-db_branches-name}
     ///
     /// When source database connection detail is not accessible to mirrord operator, users
     /// can specify the database `name` so it is included in the connection options mirrord
     /// uses as the override.
     pub name: Option<String>,
 
-    /// ### feature.db_branches.ttl_secs {#feature-db_branches-ttl_secs}
+    /// ### feature.db_branches.base.ttl_secs {#feature-db_branches-ttl_secs}
```
t4lz: Did you leave the link anchors unchanged on purpose?
```suggestion
    /// ### feature.db_branches.base.id {#feature-db_branches-base-id}
   ///
   /// Users can choose to specify a unique `id`. This is useful for reusing or sharing
   /// the same database branch among Kubernetes users.
   pub id: Option<String>,

    /// ### feature.db_branches.base.name {#feature-db_branches-base-name}
   ///
   /// When source database connection detail is not accessible to mirrord operator, users
   /// can specify the database `name` so it is included in the connection options mirrord
   /// uses as the override.
   pub name: Option<String>,

    /// ### feature.db_branches.base.ttl_secs {#feature-db_branches-base-ttl_secs}
```

0x00A5: Changed to suggested

PR #3636, opened 2025-10-13
INT-169 Kafka operator status simplified without session crd

vladrbg:
image

### DIFF
--- a/mirrord/cli/src/operator/status.rs
+++ b/mirrord/cli/src/operator/status.rs
@@ -374,7 +374,7 @@ Operator License
         }
 
         // The Kafka topic statuses are grouped by consumer (target).
-        for (kafka_consumer, kafka_row) in kafka_rows {
+        for (_, kafka_row) in kafka_rows {
             let mut kafka_table = Table::new();
             kafka_table.add_row(row![
                 "Session ID",


changelog.d/+kafka-status.added.md
Razz4780, line 1: Should be called sth like `+kafka-status.added.md`:
1. The number prefix corresponds to the github issue number in this repo (I don't think we have any). If there is no issue, you should use `+<some-title-doesnt-really-matter` prefix
2. Looks like `added` category fits better

mirrord/cli/src/operator/status.rs
```diff
@@ -281,6 +336,25 @@ Operator License
...
+            }) {
+                // Merge each session Kafka topics into our map keyed by consumer (target).
+                for (consumer, rows) in kafka_in_status {
+                    match kafka_rows.entry(consumer) {
+                        Entry::Occupied(mut consumer_rows) => consumer_rows.get_mut().extend(rows),
+                        Entry::Vacant(consumer_rows) => {
+                            consumer_rows.insert(rows);
+                        }
+                    }
+                }
```
Razz4780: Isn't `kafka_in_status` already rows grouped by consumer? What's going on here? 👀

mirrord/operator/src/crd.rs
```diff
@@ -335,6 +336,7 @@ pub struct Session {
     pub locked_ports: Option<Vec<LockedPortCompat>>,
     pub user_id: Option<String>,
     pub sqs: Option<Vec<MirrordSqsSession>>,
+    pub kafka: Option<Vec<MirrordKafkaEphemeralTopic>>,
```
Razz4780: This could be just `MirrordKafkaEphemeralTopicSpec`, right? Without the whole metadata, looks like we're not using it

mirrord/cli/src/operator/status.rs
vladrbg: Oh make sense no need to loop again, I was just mostly replicating how SQS does it.

mirrord/operator/src/crd.rs
vladrbg: Yep good point,  we only use the Spec so no need the full Topic metadata.

mirrord/cli/src/operator/status.rs
```diff
@@ -307,6 +373,26 @@ Operator License
...
+            let mut kafka_table = Table::new();
+            kafka_table.add_row(row![
+                "Session ID",
+                "Topic Name",
+                "User",
+                "Client Config",
+                "Type",
+            ]);
+
+            println!("Kafka Topics for {kafka_consumer}");
```
Razz4780: 🗑️ I think?

Razz4780 (approved):

PR #3637, opened 2025-10-14
Medschool enum variants

0x00A5:


mirrord/config/configuration.md
```diff
@@ -1305,6 +1336,12 @@ Takes a list of values, such as:
 
 Valid values follow this pattern: `[name|address|subnet/mask][:port]`.
 
+DNS queries matching what is specified here will go through the local app, everything else
```
meowjesty: Hmm it seems that this change doesn't play well with our `Remote|Local` enum variants? Since the user will never actually see the variants.

0x00A5: Updated the docs here and the one in outgoing filter (plus queue splitting ones), and they should be more readable.

PR #3638, opened 2025-10-14
Don't block user app's thread when making outgoing connections

Razz4780:
Extracted from #3629

### DIFF
--- a/changelog.d/+db-branching.config.internal.md
+++ b/changelog.d/+db-branching.config.internal.md
@@ -1,0 +1,1 @@
+Updated Rust types for DB branching configs in order to support more types of DBs.
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -768,62 +768,64 @@
     },
     "DatabaseBranchConfig": {
       "description": "Configuration for a database branch.\n\nExample:\n\n```json { \"id\": \"my-branch-db\", \"name\": \"my-database-name\", \"ttl_secs\": 120, \"type\": \"mysql\", \"version\": \"8.0\", \"connection\": { \"url\": { \"type\": \"env\", \"variable\": \"DB_CONNECTION_URL\" } } } ```",
-      "type": "object",
-      "required": [
-        "connection",
-        "type"
-      ],
-      "properties": {
-        "connection": {
-          "title": "feature.db_branches.connection {#feature-db_branches-connection}",
-          "description": "`connection` describes how to get the connection information to the source database. When the branch database is ready for use, Mirrord operator will replace the connection information with the branch database's.",
-          "allOf": [
-            {
-              "$ref": "#/definitions/DbBranchingConnectionSource"
+      "oneOf": [
+        {
+          "type": "object",
+          "required": [
+            "connection",
+            "type"
+          ],
+          "properties": {
+            "connection": {
+              "title": "feature.db_branches.base.connection {#feature-db_branches-base-connection}",
+              "description": "`connection` describes how to get the connection information to the source database. When the branch database is ready for use, Mirrord operator will replace the connection information with the branch database's.",
+              "allOf": [
+                {
+                  "$ref": "#/definitions/DbBranchingConnectionSource"
+                }
+              ]
+            },
+            "id": {
+              "title": "feature.db_branches.base.id {#feature-db_branches-base-id}",
+              "description": "Users can choose to specify a unique `id`. This is useful for reusing or sharing the same database branch among Kubernetes users.",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "name": {
+              "title": "feature.db_branches.base.name {#feature-db_branches-base-name}",
+              "description": "When source database connection detail is not accessible to mirrord operator, users can specify the database `name` so it is included in the connection options mirrord uses as the override.",
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "ttl_secs": {
+              "title": "feature.db_branches.base.ttl_secs {#feature-db_branches-base-ttl_secs}",
+              "description": "Mirrord operator starts counting the TTL when a branch is no longer used by any session. The time-to-live (TTL) for the branch database is set to 300 seconds by default. Users can set `ttl_secs` to customize this value according to their need. Please note that longer TTL paired with frequent mirrord session turnover can result in increased resource usage. For this reason, branch database TTL caps out at 15 min.",
+              "default": 300,
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            },
+            "type": {
+              "type": "string",
+              "enum": [
+                "mysql"
+              ]
+            },
+            "version": {
+              "title": "feature.db_branches.base.version {#feature-db_branches-base-version}",
+              "description": "Mirrord operator uses a default version of the database image unless `version` is given.",
+              "type": [
+                "string",
+                "null"
+              ]
             }
-          ]
-        },
-        "id": {
-          "title": "feature.db_branches.id {#feature-db_branches-id}",
-          "description": "Users can choose to specify a unique `id`. This is useful for reusing or sharing the same database branch among Kubernetes users.",
-          "type": [
-            "string",
-            "null"
-          ]
-        },
-        "name": {
-          "title": "feature.db_branches.name {#feature-db_branches-name}",
-          "description": "When source database connection detail is not accessible to mirrord operator, users can specify the database `name` so it is included in the connection options mirrord uses as the override.",
-          "type": [
-            "string",
-            "null"
-          ]
-        },
-        "ttl_secs": {
-          "title": "feature.db_branches.ttl_secs {#feature-db_branches-ttl_secs}",
-          "description": "Mirrord operator starts counting the TTL when a branch is no longer used by any session. The time-to-live (TTL) for the branch database is set to 300 seconds by default. Users can set `ttl_secs` to customize this value according to their need. Please note that longer TTL paired with frequent mirrord session turnover can result in increased resource usage. For this reason, branch database TTL caps out at 15 min.",
-          "default": 300,
-          "type": "integer",
-          "format": "uint64",
-          "minimum": 0.0
-        },
-        "type": {
-          "title": "feature.db_branches.type {#feature-db_branches-type}",
-          "allOf": [
-            {
-              "$ref": "#/definitions/DbBranchingDatabaseType"
-            }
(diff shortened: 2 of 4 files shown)


mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
+//! Implementation of a hack that allows for handling user app's outgoing TCP connects with
+//! non-blocking sockets.
```
meowjesty: Would be helpful to have an explanation on what this is solving, if we don't do this async single-threaded blocks yadda yadda.

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+    sync::OnceCell,
+};
+
+static WORKING_METHOD: OnceCell<Option<HackMethod>> = OnceCell::const_new();
+
+/// Returns a [`HackMethod`] that works on this system (if any).
+///
+/// On the first call, this function might take some time to complete.
+/// Subsequent calls should return instantly.
+pub async fn working_method() -> Option<HackMethod> {
```
meowjesty: nit:

```suggestion
pub async fn find_working_method() -> Option<HackMethod> {
```

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+impl HackMethod {
+    /// Verifies if this hack method works.
+    async fn test(self) -> io::Result<()> {
+        let prepared = self.prepare_socket(false).await?;
+        let addr = prepared.local_addr()?;
+
+        let mut conn_fut = Box::pin(TcpStream::connect(addr));
+
+        match tokio::time::timeout(Duration::from_millis(300), &mut conn_fut).await {
+            Err(_elapsed) => {}
```
meowjesty: Wait, we want it to timeout here why, exactly?

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+/// Returns a [`HackMethod`] that works on this system (if any).
+///
+/// On the first call, this function might take some time to complete.
+/// Subsequent calls should return instantly.
+pub async fn working_method() -> Option<HackMethod> {
+    *WORKING_METHOD
+        .get_or_init(|| async {
+            let futs = HackMethod::iter().map(|method| {
+                tokio::time::timeout(Duration::from_secs(2), method.test())
+                    .unwrap_or_else(|_elapsed| Err(io::Error::other("test timed out")))
```
meowjesty:
```suggestion
                    .unwrap_or_else(|_elapsed| Err(io::Error::other("HackMethod test timed out")))
```

meowjesty: Otherwise I can see someone (me) reading this and thinking that this is some test failure.

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Accept the first connection made from an address different than the one used for the dummy
+    /// connection.
+    Backlog0,
+    /// Same as [`HackMethod::Backlog0`], but with backlog size 1.
+    Backlog1,
+}
+
+impl HackMethod {
+    /// Verifies if this hack method works.
+    async fn test(self) -> io::Result<()> {
```
meowjesty: nit: Maybe rename it to some sort of "check connection doesn't block".

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+                    listener,
+                    dummy_self_conn,
+                })
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum PreparedTcpSocket {
```
meowjesty: docs please

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: The timeout here confirms that `TcpStream::connect` call will not complete until we call `.accept()`. 300ms is kinda arbitrary, but :shrug: 
We can't do `let hack_works = TcpStream::connect(addr).now_or_never().is_none()`, because the future will never resolve instantly. But if it doesn't resolve within 300ms, the hack probably works

meowjesty: Truly deserving of the name hack. An I was wondering why you were calling this a hack, even thought it generally makes sense...

meowjesty: docs for it pls

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+    Socket(TcpSocket),
```
meowjesty:
```suggestion
pub(super) enum PreparedTcpSocket {
    Socket(TcpSocket),
```

I think we don't need any of this to be `pub`?

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -162,66 +229,101 @@ impl OutgoingProxy {
...
+                UnexpectedAgentMessage(message)
+            })?;
+
+        let DaemonConnect {
+            connection_id,
+            remote_address,
+            local_address,
+        } = match connect {
+            Ok(connect) => {
+                tracing::info!(
```
meowjesty, line 272: `info` and not `debug`?

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -233,12 +335,59 @@ impl OutgoingProxy {
...
+                    remote_address = %request.remote_address,
+                    "Cannot emulate non-blocking TCP connect, no working hack detected."
+                );
+                None
+            }
+        } else {
+            None
+        };
+
+        let connection_id = rand::random::<u128>();
```
meowjesty, line 378: Doesn't random here have a tiny chance of collision? If so, maybe add a comment here?

mirrord/layer/tests/issue2744.rs
```diff
@@ -0,0 +1,127 @@
...
+        tcp::{DaemonTcpOutgoing, LayerTcpOutgoing},
+    },
+};
+use rstest::rstest;
+use tempfile::NamedTempFile;
+
+#[rstest]
+#[tokio::test]
+#[timeout(Duration::from_secs(15))]
+async fn issue_2744_non_blocking_outgoing_tcp(dylib_path: &Path) {
```
meowjesty, line 24: Small description of the issue as docs? I think this would be nice for every test we add (especially those that are super specific).

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
```diff
@@ -0,0 +1,189 @@
...
+/// socket.
+///
+/// Expected flow:
+/// 1. Create a socket with [`HackMethod::prepare_socket`]
+/// 2. Send sockket's local address to the layer
+/// 3. User app initiates connection to the socket, and is blocked on EINPROGRESS
+/// 4. Accept the connection with [`PreparedTcpSocket::accept`]
+/// 5. User app's socket becomes writable, which notifies the app's async runtime
+#[derive(Clone, Copy, Debug, EnumIter)]
+pub enum HackMethod {
```
meowjesty: Isn't this more like, `NonBlockingConnect`? Or if you want to keep the hack `NonBlockingConnectHack`.

meowjesty (changes requested):
Some questions and nits.

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: This whole module is private anyway

mirrord/intproxy/src/proxies/outgoing.rs
Razz4780: Yeah, I figured why not. We log incoming HTTP requests on info, outgoing connections are just as important

mirrord/layer/tests/issue2744.rs
Razz4780: Added

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Added

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Improved the whole module a bit, wdyt now?

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Added

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Improved the whole module a bit, wdyt now?

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Changed the whole module to be more self-contained. Now it does not mention any non-blocking connect stuff, it only implements a concept of a "busy" tcp listener. I documented how it's used in the non-blocking flow in the `OutgoingProxy` doc

mirrord/intproxy/src/proxies/outgoing/non_blocking_hack.rs
Razz4780: Improved the whole module a bit, wdyt now?

meowjesty (approved):
Nice 👍

PR #3639, opened 2025-10-14
INT-164 mirrord db-branches status / destroy

vladrbg:
CLI for db branches to show and destroy:

Currently it calls the kube API on the MysqlBranchDatabase, so if we need to add more database types will need to add it in a similar way. Also currently the type is just a static "mysql" .

image

### DIFF
--- a/changelog.d/+cargo-deny-ignore-rust-unic.internal.md
+++ b/changelog.d/+cargo-deny-ignore-rust-unic.internal.md
@@ -1,0 +1,1 @@
+Add ignore to `RUSTSEC-2025-0074`, `RUSTSEC-2025-0075`, `RUSTSEC-2025-0080`, `RUSTSEC-2025-0081`, `RUSTSEC-2025-0098`and `RUSTSEC-2025-0104` becuase open-i18n/rust-unic is unmaintained but we should not have any issue with it.
--- a/deny.toml
+++ b/deny.toml
@@ -12,8 +12,14 @@ version = 2
 ignore = [
     "RUSTSEC-2020-0095", # difference is unmaintained. (part of apple-codesign)
     "RUSTSEC-2023-0071", # rsa crate, potential key recovery through timing sidechannels, though it sounds bad
-                         # we only use rsa as part of apple-codesign for re-sign binaries, so this should not matter.
+    # we only use rsa as part of apple-codesign for re-sign binaries, so this should not matter.
     "RUSTSEC-2024-0436", # paste is unmaintained but we should not have an issue with it.
+    "RUSTSEC-2025-0074", # unic-segment (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
+    "RUSTSEC-2025-0075", # unic-char-range (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
+    "RUSTSEC-2025-0080", # unic-common (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
+    "RUSTSEC-2025-0081", # unic-char-property (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
+    "RUSTSEC-2025-0098", # unic-ucd-version (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
+    "RUSTSEC-2025-0104", # unic-ucd-segment (tera -> mirrord-config) is unmaintained, but we should not have an issue with it.
 ]
 
 [licenses]
@@ -40,9 +46,7 @@ allow = [
 [[licenses.clarify]]
 crate = "ring"
 expression = "MIT AND ISC AND OpenSSL"
-license-files = [
-    { path = "LICENSE", hash = 0xbd0eed23 }
-]
+license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
 
 [bans]
 # enable this when dependencies are a bit more stable
@@ -57,5 +61,5 @@ allow-git = [
     "https://github.com/metalbear-co/rawsocket",
     "https://github.com/metalbear-co/rust-iptables",
     "https://github.com/metalbear-co/kube",
-    "https://github.com/metalbear-co/minhook-detours-rs"
+    "https://github.com/metalbear-co/minhook-detours-rs",
 ]


mirrord/cli/src/config.rs
```diff
@@ -1093,6 +1096,39 @@ pub(super) enum CiCommand {
...
+    #[command(subcommand)]
+    pub command: DbBranchesCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub(super) enum DbBranchesCommand {
+    /// Show the status of database branches
+    Status {
+        /// Names of specific branches to show status for (all branches if none specified)
+        #[arg()]
```
0x00A5, line 1143: Is this a left over?

mirrord/cli/src/config.rs
```diff
@@ -156,6 +156,9 @@ pub(super) enum Commands {
     #[command(name = "port-forward")]
     PortForward(Box<PortForwardArgs>),
 
+    /// Manage database branching.
+    #[command(name = "db-branches")]
+    DbBranches(Box<DbBranchesArgs>),
```
0x00A5, line 169: super nit: add a newline after this to make it consistent with other fields of the struct.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+    CliResult,
+    config::{DbBranchesArgs, DbBranchesCommand},
+};
+
+pub async fn db_branches_command(args: DbBranchesArgs) -> CliResult<()> {
+    match &args.command {
+        DbBranchesCommand::Status { names } => status_command(&args, names.clone()).await,
+        DbBranchesCommand::Destroy { all, names } => {
+            destroy_command(&args, *all, names.clone()).await
+        }
```
0x00A5: Can `status_command()` and `destroy_command()` take a reference of `names` so we don't need to clone? It seems they are only used for resource filtering.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+async fn status_command(args: &DbBranchesArgs, names: Vec<String>) -> CliResult<()> {
```
0x00A5: I think a set works better than `Vec` here. Shall we use something like a `HashSet` or `BTreeSet`?

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        table.add_row(row![name, db_type, phase, ttl, db_name, users]);
+    }
+
+    table.printstd();
+
+    Ok(())
+}
+
+async fn destroy_command(args: &DbBranchesArgs, all: bool, names: Vec<String>) -> CliResult<()> {
```
0x00A5: Have we tested this command? I think the user will miss the permission to delete this resource. The current permissions are defined here: https://github.com/metalbear-co/charts/blob/main/mirrord-operator/templates/_helpers.tpl#L51-L61.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+                    .metadata
+                    .name
+                    .as_ref()
+                    .is_some_and(|name| names.contains(name))
+            })
+            .collect()
+    };
+
+    if filtered_branches.is_empty() {
+        println!("No active DB branch found");
```
0x00A5: Let's not use `println` and use `ProgressTracker`. You can use the `operator status` as an example.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::{
+    CliResult,
+    config::{DbBranchesArgs, DbBranchesCommand},
+};
+
+pub async fn db_branches_command(args: DbBranchesArgs) -> CliResult<()> {
+    match &args.command {
+        DbBranchesCommand::Status { names } => status_command(&args, names.clone()).await,
+        DbBranchesCommand::Destroy { all, names } => {
```
0x00A5: We can add `ProgressTracker`  to both commands to update the user with active progress. Check operator status command for example.

mirrord/cli/src/db_branches.rs
0x00A5: I saw the operator status command does `println!` too, but if we add a `ProgressTracker`, this can be `progress.success(Some("No active DB branch found."));`

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let mut table = Table::new();
+    table.add_row(row![
+        "Name",
+        "DB Type",
+        "Phase",
+        "TTL (sec)",
+        "Database",
+        "Users"
+    ]);
```
0x00A5, line 105: I think the user might want to see the expire time too, can we add it here?

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+            .status
+            .as_ref()
+            .map(|s| {
+                if s.session_info.is_empty() {
+                    "none".to_string()
+                } else {
+                    s.session_info
+                        .values()
+                        .map(|session| {
+                            format!("{}@{}", session.owner.username, session.owner.hostname)
```
0x00A5: Does operator status command format the users column this way? Let's make it consistent if not already.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+            ));
+        }
+
+        // Delete all branches
+        let delete_params = DeleteParams::default();
+        for branch in branches {
+            if let Some(name) = branch.metadata.name {
+                match api.delete(&name, &delete_params).await {
+                    Ok(_) => println!("Destroyed branch: {}", name),
+                    Err(e) => eprintln!("Failed to destroy branch {}: {}", name, e),
```
0x00A5: Let's use `ProgressTracker` for organizing these print statements. Same to other places.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+                    ));
+                }
+            }
+        }
+
+        if !found_any {
+            return Err(crate::CliError::DbBranchNotFound(
+                "No active DB branch found".to_string(),
+            ));
+        }
```
0x00A5: This seems redundant, since we already print a `"Branch not found: {name}"` per given name.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,176 @@
...
+    if all {
+        // List all branches first to check if any exist
+        let branches = api.list(&ListParams::default()).await.map_err(|e| {
+            crate::CliError::ListTargetsFailed(mirrord_kube::error::KubeApiError::KubeError(e))
+        })?;
+
+        if branches.items.is_empty() {
+            return Err(crate::CliError::DbBranchNotFound(
+                "No active DB branch found".to_string(),
+            ));
```
0x00A5: nit: I think this shouldn't be an error, but `progress.success(Some("No active DB branch found."));`

mirrord/cli/src/error.rs
```diff
@@ -472,6 +472,12 @@ pub(crate) enum CliError {
     ))]
     NestedExec,
 
+    #[error("Database branch error: {0}")]
+    #[diagnostic(help(
+        "Make sure the database branch exists and the operator is running correctly."
+    ))]
+    DbBranchNotFound(String),
```
0x00A5: Mentioned in the commands methods, I think we don't need this error, i.e. destroy command should be idempotent.

mirrord/cli/src/db_branches.rs
vladrbg: Hmm  I tested it seems to work, but maybe there is a case where it wont without the permission, not sure? 

image

mirrord/cli/src/db_branches.rs
vladrbg: Got it, yep it's much better like that.

mirrord/cli/src/db_branches.rs
vladrbg: Oh I see, the `k8s_username` actually displays the full data no need to do it like this. Nice! 

image

mirrord/cli/src/db_branches.rs
vladrbg: Do you mean just convert to it, because clap doesn't support HashSet right? 

https://github.com/metalbear-co/mirrord/blob/b9be125922437a52606237a2ffa77d22d52d1156/mirrord/cli/src/db_branches.rs#L23-L24

mirrord/cli/src/config.rs
vladrbg: The empty arg works like this:  

```bash
# select all
mirrord db-branches status
# select one
mirrord db-branches status mysql-app-mysql-branch-77thg
# select many
mirrord db-branches status mysql-app-mysql-branch-77thg mysql-app-mysql-branch-fqvjk 
```

image

Or we want something different here?

mirrord/cli/src/db_branches.rs
0x00A5: Oh I see, yeah, that looks good 👍

mirrord/cli/src/db_branches.rs
0x00A5: What's the test setup here, is it docker desktop? I usually create a service account and give it only the basic permission + mirrord operator roles. That should catch issues of missing permission.

vladrbg: Ohh yep yep, thats it I had full permissions, will improve my setup.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,190 @@
...
+
+async fn status_command(args: &DbBranchesArgs, names: &[String]) -> CliResult<()> {
+    let names: HashSet<_> = names.iter().collect();
+
+    let mut progress = ProgressTracker::from_env("DB Branches Status");
+    let mut status_progress = progress.subtask("fetching branches");
+
+    let client = Client::try_default().await.map_err(|e| {
+        crate::CliError::CreateKubeApiFailed(mirrord_kube::error::KubeApiError::KubeError(e))
+    })?;
```
0x00A5: I think we shouldn't use `try_default` here. In `LayerConfig` (it is really mirrord config not just layer), the user could have specified either kube config or kube context, we should use those values when constructing the client in order to be consistent. I recommend adding an optional field for mirrord config like the following:
```
    /// Load config from config file
    /// When using -f flag without a value, defaults to "./.mirrord/mirrord.json"
    #[arg(short = 'f', long, value_hint = ValueHint::FilePath, default_missing_value = "./.mirrord/mirrord.json", num_args = 0..=1)]
    pub config_file: Option<PathBuf>,
```
Then use the set of functions we already have to resolve the mirrord config and construct kube client. Here is an example: https://github.com/metalbear-co/mirrord/blob/main/mirrord/cli/src/vpn.rs#L22-L43

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,190 @@
...
+    Ok(())
+}
+
+async fn destroy_command(args: &DbBranchesArgs, all: bool, names: &Vec<String>) -> CliResult<()> {
+    let mut progress = ProgressTracker::from_env("DB Branches Destroy");
+    let mut destroy_progress = progress.subtask("deleting branches");
+
+    let client = Client::try_default().await.map_err(|e| {
+        crate::CliError::CreateKubeApiFailed(mirrord_kube::error::KubeApiError::KubeError(e))
+    })?;
```
0x00A5: Same here please

0x00A5 (changes requested):

mirrord/cli/src/db_branches.rs
vladrbg: Ah I see, changing it!

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    let mut progress = ProgressTracker::from_env("DB Branches Status");
+    let mut status_progress = progress.subtask("fetching branches");
+
+    let mut cfg_context = ConfigContext::default()
+        .override_env_opt(LayerConfig::FILE_PATH_ENV, args.config_file.clone())
+        .override_env_opt("MIRRORD_TARGET_NAMESPACE", args.namespace.clone());
+
+    let mut layer_config = LayerConfig::resolve(&mut cfg_context)?;
+    layer_config.agent.privileged = true;
```
0x00A5: I think this one, `layer_config.agent.privileged`,  is VPN command specific. We just need to resolve the layer config and build a client with it.

mirrord/cli/src/db_branches.rs
```diff
@@ -0,0 +1,236 @@
...
+async fn destroy_command(args: &DbBranchesArgs, all: bool, names: &Vec<String>) -> CliResult<()> {
+    let mut progress = ProgressTracker::from_env("DB Branches Destroy");
+    let mut destroy_progress = progress.subtask("deleting branches");
+
+    let mut cfg_context = ConfigContext::default()
+        .override_env_opt(LayerConfig::FILE_PATH_ENV, args.config_file.clone())
+        .override_env_opt("MIRRORD_TARGET_NAMESPACE", args.namespace.clone());
+
+    let mut layer_config = LayerConfig::resolve(&mut cfg_context)?;
+    layer_config.agent.privileged = true;
```
0x00A5: Same here

0x00A5 (approved):

PR #3641, opened 2025-10-15
add MIRRORD_DONT_LOAD to crash mirrord if exists when mirrord loads

aviramha:
add MIRRORD_DONT_LOAD to crash mirrord if exists when mirrord loads to enable users running destructive actions with mirrord

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -395,7 +395,6 @@ fn layer_start(mut config: LayerConfig) {
     let proxy_connection_timeout = *PROXY_CONNECTION_TIMEOUT
         .get_or_init(|| Duration::from_secs(config.internal_proxy.socket_timeout));
 
-
     let debugger_ports = DebuggerPorts::from_env();
     let local_hostname = trace_only || !config.feature.hostname;
     let process_info = EXECUTABLE_ARGS


mirrord/layer/src/lib.rs
```diff
@@ -385,6 +389,11 @@ fn layer_start(mut config: LayerConfig) {
     let proxy_connection_timeout = *PROXY_CONNECTION_TIMEOUT
         .get_or_init(|| Duration::from_secs(config.internal_proxy.socket_timeout));
 
+    // we don't care about value, just that this env exists
+    let dont_start = std::env::var(FAILSAFE_ENV).is_ok();
+    if dont_start {
+        panic!("{FAILSAFE_ENV} environment variable found, stopping execution.")
+    }
```
Razz4780: Can you move this to the top of `layer_pre_initialization`?

aviramha: done

mirrord/layer/src/lib.rs
```diff
@@ -187,6 +191,12 @@ static PROXY_CONNECTION_TIMEOUT: OnceLock<Duration> = OnceLock::new();
 
 /// Loads mirrord configuration and does some patching (SIP, dotnet, etc)
 fn layer_pre_initialization() -> Result<(), LayerError> {
+    // we don't care about value, just that this env exists
+    let dont_start = std::env::var(FAILSAFE_ENV).is_ok();
```
DmitryDodzin, line 195: I would try and parse boolean and not just that it exists maybe

aviramha: It's intended - I don't want someone doing "false" or "0" or some other idea and leading to hazard. no reason for the env exists beyond making mirrord fail.

Razz4780 (approved):

PR #3642, opened 2025-10-15
Handle half-shutdown of TCP stream (port forward)

gememma:
Issue: https://linear.app/metalbear/issue/MBE-1469/traffic-from-browser-lost-during-port-forwarding

### DIFF
--- a/mirrord/cli/src/port_forward.rs
+++ b/mirrord/cli/src/port_forward.rs
@@ -881,7 +881,7 @@ impl LocalConnectionTask {
                         // ignore repeat empty messages
                         if let Some(write) = self.write.take() {
                             drop(write);
-                            tracing::info!(
+                            tracing::debug!(
                                 port_mapping = ?self.port_mapping,
                                 "remote half closed the connection",
                             );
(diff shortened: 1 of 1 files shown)


mirrord/cli/src/port_forward.rs
```diff
@@ -877,18 +877,31 @@ impl LocalConnectionTask {
...
-                            Ok(_) => continue,
-                            Err(error) => {
-                                tracing::error!(
-                                    %error,
-                                    port_mapping = ?self.port_mapping,
-                                    "local connection failed",
-                                );
-                                break Ok(());
-                            },
+                    Some(message) if message.is_empty() => {
```
meowjesty, line 880: Empty messages means the stream has closed?

mirrord/cli/src/port_forward.rs
```diff
@@ -877,18 +877,31 @@ impl LocalConnectionTask {
...
+                        // ignore repeat empty messages
+                        if let Some(write) = self.write.take() {
+                            drop(write);
+                            tracing::info!(
```
meowjesty: Really want `info` here or maybe `debug`?

meowjesty (changes requested):
Can we get a test for the thing that was going wrong?

mirrord/cli/src/port_forward.rs
gememma: "half closed", yeah 👀 means the write half is closed, even if the read half is still sending data

PR #3645, opened 2025-10-16
Outgoing connect v2

Razz4780:


mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -156,14 +161,20 @@ pub struct OutgoingProxy {
     datagrams_reqs: RequestQueue<ConnectInProgress>,
     /// For [`OutgoingConnectRequest`]s related to [`NetProtocol::Stream`].
     stream_reqs: RequestQueue<ConnectInProgress>,
+    v2_reqs: HashMap<(Uid, NetProtocol), ConnectInProgress>,
```
meowjesty, line 176: Docs for these pls.

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -440,6 +498,19 @@ impl OutgoingProxy {
                 )))
                 .await;
         }
+
+        tracing::debug!(
+            responses = self.v2_reqs.len(),
+            "Flushing error responses to V2 connect requests"
+        );
+        for in_progress in std::mem::take(&mut self.v2_reqs).into_values() {
```
meowjesty, line 518: Isn't this just `Drain` iter?

mirrord/layer/tests/issue1458.rs
```diff
@@ -47,13 +50,16 @@ async fn test_issue1458(
     println!("connecting to address {addr:#?}");
 
     intproxy
```
meowjesty, line 40: Make one of those helper functions that replies with this? It's looks pretty much the same in the other tests.

Cargo.toml
```diff
@@ -195,7 +195,7 @@ pem = "3"
 home = "0.5"
 
 # Used by `cli`, `analytics`
```
meowjesty:
```suggestion
# Used by `cli`, `analytics`, `protocol`
```

I don't think we're keeping (or that we need) this up-to-date, but the suggestion is here.

mirrord/intproxy/src/proxies/outgoing.rs
Razz4780: Yeah, but `Drain` leaves the memory allocated. Recently I figured we never shrink collections, while we probably should :skull: 
Started taking this into consideration

mirrord/intproxy/src/proxies/outgoing.rs
Razz4780: Added

PR #3652, opened 2025-10-17
Improve UX when configuring intproxy/extproxy logs

Razz4780:


mirrord/config/src/logfile_path.rs
```diff
@@ -0,0 +1,307 @@
...
+
+impl LogSource for Extproxy {
+    fn generate_file_name() -> String {
+        let random_name: String = Alphanumeric.sample_string(&mut rand::rng(), 7);
+        let timestamp = SystemTime::UNIX_EPOCH
+            .elapsed()
+            .expect("system time should not be earlier than UNIX EPOCH")
+            .as_secs();
+        format!("mirrord-extproxy-{timestamp}-{random_name}.log")
+    }
```
BigLullu, line 47:
```suggestion
    fn generate_file_name() -> String {
        let timestamp = SystemTime::UNIX_EPOCH
            .elapsed()
            .expect("system time should not be earlier than UNIX EPOCH")
            .as_micros();
        format!("mirrord-extproxy-{timestamp}.log")
    }
```
Timestamp is not enough here ?

Razz4780: Micros would be enough, but I wouldn't want to change the current format

PR #3654, opened 2025-10-20
Run macos CI on aarch64 runners

iniw:


changelog.d/3305.internal.md
```diff
@@ -0,0 +1 @@
+The macOS CI now uses github's aarch64 runners
```
meowjesty, line 1: Changelog should be `internal`. I think when there's an issue for it the name is `3305.internal.md`

0x00A5 (approved):
So we load a x86_64-only layer dylib into the test process on an aarch64 runner, and it works :panik:

PR #3655, opened 2025-10-20
Independent send handles for intproxy tasks

MintSoup:
Rework the intproxy to have independent tx handles for all tasks. Allows us to buffer/queue outgoing messages and block the tasks when they enqueue too many messages, and lays the groundwork for the chunked protocol encoding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4861,7 +4861,7 @@ dependencies = [
 
 [[package]]
 name = "mirrord-protocol-io"
-version = "3.168.1"
+version = "3.170.0"
 dependencies = [
  "actix-codec",
  "bincode",
@@ -9152,11 +9152,10 @@ dependencies = [
 
 [[package]]
 name = "yoke"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f41bb01b8226ef4bfd589436a297c53d118f65921786300e427be8d487695cc"
+checksum = "72d6e5c6afb84d73944e5cedb052c4680d5657337201555f9f2a16b7406d4954"
 dependencies = [
- "serde",
  "stable_deref_trait",
  "yoke-derive",
  "zerofrom",
@@ -9254,10 +9253,11 @@ dependencies = [
 
 [[package]]
 name = "zerovec"
-version = "0.11.4"
+version = "0.11.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7aa2bd55086f1ab526693ecbe444205da57e25f4489879da80635a46d90e73b"
+checksum = "6c28719294829477f525be0186d13efa9a3c602f7ec202ca9e353d310fb9a002"
 dependencies = [
+ "serde",
  "yoke",
  "zerofrom",
  "zerovec-derive",


mirrord/agent/src/client_connection.rs
```diff
@@ -97,65 +96,56 @@ impl ClientConnection {
...
-    #[tracing::instrument(level = "trace", err)]
+    #[tracing::instrument(level = "trace")]
     pub async fn send(&mut self, message: DaemonMessage) -> io::Result<()> {
-        match &mut self.framed {
-            ConnectionFramed::Tcp(framed) => framed.send(message).await?,
-            ConnectionFramed::Tls(framed) => framed.send(message).await?,
-        }
+        self.connection.send(message).await;
 
+        // REVIEW Fix up the signatures
```
Razz4780: Yup

mirrord/operator/src/client/conn_wrapper.rs
```diff
@@ -1,126 +1,97 @@
...
-                },
-            }
-        }
-
-        let _ = self.connection.send(Message::Close(None)).await;
-
-        Ok(())
-    }
-}
+// use futures::{Sink, SinkExt, Stream, StreamExt};
```
Razz4780: 🗑️

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        Ok(Self {
+            rx: inbound_rx,
+            tx_handle,
+        })
+    }
+
+    pub async fn from_channel<C, Filter>(
+        channel: C,
+        filter: Option<Filter>,
```
Razz4780, line 143: What's this for?

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,388 @@
+use std::{
```
Razz4780, line 3: Docs for everything here please  : D

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        self.nonempty.notify_one();
+
+        Ok(())
+    }
+
+    fn poll_next(&self) -> Option<Vec<u8>> {
+        let mut lock = self.queues.lock().unwrap();
+
+        let (key, queue) = lock
```
Razz4780: This is O(n), kinda meh

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+struct OutQueue {
+    messages: VecDeque<Vec<u8>>,
+    used_bytes: usize,
+
+    free: Arc<Notify>,
+}
+
+type FilterFn<I, O> = dyn Fn(O) -> Either<O, I> + Send + Sync;
+
+struct OutQueues<Type: ProtocolEndpoint> {
```
Razz4780: This needs UTs

mirrord/operator/src/client.rs
```diff
@@ -1225,34 +1132,68 @@ impl OperatorApi<PreparedClientCert> {
...
+                        Either::Left(ClientMessage::SwitchProtocolVersion(
+                            operator_protocol_version.min(&version).clone(),
+                        ))
+                    }
+                    _ => Either::Right(DaemonMessage::SwitchProtocolVersionResponse(
+                        semver::Version::new(1, 2, 1),
+                    )),
+                },
+                other => Either::Left(other),
+            }),
```
Razz4780, line 1278: Right now all client code simply assumes that the protocol negotiation is supported, right? We just send `SwitchProtocolVersion`. If so, this is unnecessary. No need to add support for very old operators now.

mirrord/protocol/src/io.rs
```diff
@@ -0,0 +1 @@
+
```
Razz4780: 🗑️

mirrord/protocol/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mirrord-protocol"
-version = "1.22.0"
+version = "1.21.3"
```
Razz4780: We shouldn't need to add any new code to `mirrord-protocol`

mirrord/protocol/src/queueing.rs
```diff
@@ -0,0 +1,69 @@
...
+    fmt,
+    hash::Hash,
+    mem::{Discriminant, discriminant},
+};
+
+use crate::ConnectionId;
+
+/// A trait implemented on message types that dictates how queueing
+/// should work for different message types.
+pub trait Queueable: Sized {
```
Razz4780: A thing occurred to me when reviewing. I believe we need to go with the other approach, where each send handle has its own independent queue (instead of having queues with ids derived from messages).
This is because the agent sends logs that are related to other messages, like:

1. Outgoing connection failed
2. Send `LogMessage` with error
3. Send `DaemonTcp::Close`

And we might have more cases like this in the future 🤔

mirrord/protocol-io/src/lib.rs
MintSoup: The operator client uses it for running over websockets

mirrord/protocol/Cargo.toml
MintSoup: I added this to get the CI to stop complaining, is there another way?

mirrord/operator/src/client.rs
MintSoup: tbh I'm not too sure what this code is for, but currently we have equivalent logic. See https://github.com/metalbear-co/mirrord/blob/c1796bf04808f6323fdd988368fa2eda296805d5/mirrord/operator/src/client/conn_wrapper.rs#L93

Razz4780: Ah right, sorry (lol)

mirrord/operator/src/client.rs
```diff
@@ -1225,34 +1131,68 @@ impl OperatorApi<PreparedClientCert> {
...
+            .map(|i| match i.map_err(OperatorClientError::from)? {
+                tungstenite::Message::Binary(pl) => Ok(pl),
+                other => Err(OperatorClientError::InvalidMessage(other)),
+            });
+
+        let operator_protocol_version = session.operator_protocol_version.clone();
+
+        let conn = Connection::<ProtocolClient>::from_channel(
+            ws,
+            Some(move |msg| match msg {
```
Razz4780, line 1266:
```suggestion
            // Mock protocol version negotiation if the operator does not support it.
            Some(move |msg| match msg {
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
+/// This module implements mirrord-protocol's wire-level IO.
```
Razz4780:
```suggestion
//! This module implements mirrord-protocol's wire-level IO.

```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+{
+}
+
+impl<T, I, O> Transport<I, O> for T where
+    T: Sink<O> + Stream<Item = Result<I, <Self as Sink<O>>::Error>> + Send + 'static
+{
+}
+
+/// "Phony" trait, implemented by `Client` and `Agent`, used for
+/// determining input and output message types.
```
Razz4780:
```suggestion
/// "Phony" trait, used for determining input and output message types.
///
/// Implemented by [`Client`] and [`Agent`].
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    #[error("IO error: {0}")]
+    IO(#[from] io::Error),
+    #[error("unexpected peer message")]
+    UnexpectedPeerMessage,
+}
+
+#[derive(Clone)]
+pub struct Client;
+#[derive(Clone)]
+pub struct Agent;
```
Razz4780, line 62:
```suggestion
/// Marker type, representing the client side of a mirrord-protocol connection.
#[derive(Clone)]
pub struct Client;

/// Marker type, representing the server side of a mirrord-protocol connection (agent or operator).
#[derive(Clone)]
pub struct Agent;
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+/// `Self::tx_handle` to obtain a handle for sending messages.
+/// Messages sent with different send handles are *not* guaranteed to
+/// be delivered in order. To preserve message delivery order, send
+/// all relevant messages through a single `TxHandle` or clones
+/// derived from it.
+///
+/// The impl works by randomly picking a message from one of the
+/// nonempty queues and sending it over the wire. Future versions will
+/// use message chunking to fairly split the available bandwidth
+/// between all queues.
```
Razz4780:
```suggestion
/// The main handle to our end of the mirrord-protocol connection.
/// `Type` is either [`Client`] or [`Agent`].
///
/// Implements message queueing and almost-fair scheduling. Use
/// [`Self::tx_handle`] to obtain a handle for sending messages.
/// Messages sent with different send handles are *not* guaranteed to
/// be delivered in order. To preserve message delivery order, send
/// all relevant messages through a single [`TxHandle`], or clones
/// derived from it.
///
/// The impl works by randomly picking a message from one of the
/// nonempty queues and sending it over the wire. Future versions will
/// use message chunking to fairly split the available bandwidth
/// between all queues (as messages can vary in size).
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    /// Send a message. Note that all messages sent with this function
+    /// will be put in a single queue, and thus be delivered
+    /// sequentially. Use `Self::tx_handle` for queueing messages
+    /// independently.
+    #[inline]
+    pub async fn send(&self, msg: Type::OutMsg) {
+        self.shared_state.push(QueueId(0), msg).await
+    }
```
Razz4780:
```suggestion
    /// Send a message.
    ///
    /// Note that all messages sent with this function
    /// will be put in a single queue, and thus be delivered
    /// sequentially. Use [`Self::tx_handle`] for queueing messages
    /// independently.
    #[inline]
    pub async fn send(&self, msg: Type::OutMsg) {
        self.shared_state.push(QueueId(0), msg).await
    }
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    Type::OutMsg: bincode::Decode<()>,
+{
+    pub async fn next(&self) -> Option<Type::OutMsg> {
+        bincode::decode_from_slice(&self.0.next().await, bincode::config::standard())
+            .ok()
+            .map(|e| e.0)
+    }
+}
+
+#[instrument(skip_all)]
```
Razz4780:
```suggestion
#[instrument(level = Level::TRACE, name = "mirrord_protocol_io_task", skip_all)]
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    Type: ProtocolEndpoint,
+    Channel: Transport<Type::InMsg, Vec<u8>>,
+    Channel::Error: std::error::Error + Send,
+{
+    pin!(framed);
+    loop {
+        select! {
+            to_send = queues.next() => {
+                if let Err(err) = framed.send(to_send).await {
+                    tracing::error!(?err, "failed to send message");
```
Razz4780:
```suggestion
                if let Err(error) = framed.send(to_send).await {
                    tracing::error!(?error, "failed to send message");
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+                }
+            }
+            received = framed.next() => {
+                match received {
+                    None => {
+                        tracing::info!("no more messages, exiting task");
+                        break;
+                    }
+                    Some(Err(err)) => {
+                        tracing::error!(?err, "failed to receive message");
```
Razz4780:
```suggestion
                    Some(Err(error)) => {
                        tracing::error!(?error, "failed to receive message");
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+                        break;
+                    }
+                    Some(Ok(e)) => {
+                        if let Err(e) = tx.send(e).await {
+                            tracing::info!(?e, "io task channel closed");
```
Razz4780:
```suggestion
                    Some(Ok(msg)) => {
                        if let Err(error) = tx.send(msg).await {
                            tracing::info!(?error, "io task channel closed");
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+            connection,
+            inbound_tx,
+            ConnectionOutput::<Type>(shared_state),
+        )
+    }
+
+    #[inline]
+    pub async fn recv(&mut self) -> Option<Type::InMsg> {
+        self.rx.recv().await
+    }
```
Razz4780, line 201:
```suggestion
    /// Receive the next message.
    ///
    /// Cancel safe.
    #[inline]
    pub async fn recv(&mut self) -> Option<Type::InMsg> {
        self.rx.recv().await
    }
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    #[inline]
+    pub fn poll_recv(&mut self, cx: &mut Context) -> Poll<Option<Type::InMsg>> {
```
Razz4780, line 205:
```suggestion
    /// Poll the next message.
    #[inline]
    pub fn poll_recv(&mut self, cx: &mut Context) -> Poll<Option<Type::InMsg>> {
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+        self.rx.poll_recv(cx)
+    }
+
+    /// Send a message. Note that all messages sent with this function
+    /// will be put in a single queue, and thus be delivered
+    /// sequentially. Use `Self::tx_handle` for queueing messages
```
Razz4780:
```suggestion
    /// sequentially. Use [`Self::tx_handle`] for queueing messages
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+            in_tx,
+            out_filter,
+            // 0 is reserved for the Connection struct
+            next_queue_id: 1.into(),
+        }
+    }
+
+    /// Try to push a new message into the queue with id `queue_id`,
+    /// creating it if it doesn't exist. If the queue is full, return
+    /// the message and an `OwnedNotified` that will resolve then the
```
Razz4780:
```suggestion
    /// the message and an `OwnedNotified` that will resolve when the
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+        &self,
+        queue_id: QueueId,
+        encoded: Vec<u8>,
+    ) -> Result<(), (Vec<u8>, OwnedNotified)> {
+        let mut lock = self.queues.lock().unwrap();
+
+        // Garbage-collect unused queues
+        if lock.queues.len() > 8 && lock.ready.len() < lock.queues.len() / 3 {
+            // We can remove all empty ones because we know they wont be in `ready`
+            lock.queues.retain(|_, v| !v.messages.is_empty());
```
Razz4780, line 378:
```suggestion
            lock.queues.retain(|_, v| !v.messages.is_empty());
            lock.queues.shrink_to_fit();
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+
+        queue.used_bytes += encoded.len();
+        queue.messages.push_back(encoded);
+
+        if queue.messages.len() == 1 {
+            // .len() > 1 implies that it was already in `ready`
+            lock.ready.push(queue_id);
+        }
+
+        tracing::info!(queues = ?lock.queues.keys() ,"added queue");
```
Razz4780:
```suggestion
        tracing::trace!(queues = ?lock.queues.keys(), "added queue");
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    /// Return a new `TxHandle` to a new independent queue.
+    ///
+    /// The queue is actually created lazily and may be
+    /// garbage-collected and recreated later on, but that process is
+    /// transparent to users of this function.
+    fn new_queue(self: Arc<Self>) -> TxHandle<Type> {
+        let next = self.next_queue_id.fetch_add(1, Ordering::Relaxed);
+        if next > usize::MAX / 2 {
+            panic!("Too many QueueIds generated");
+        }
```
Razz4780, line 485: Why? 👀

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+            shared_state: self,
+            id: QueueId(next),
+        }
+    }
+}
+
+/// A handle to a queue for outbound messages.
+///
+/// Cloning this struct returns a handle to the same queue. To create
+/// a new queue, use `Self::another`.
```
Razz4780:
```suggestion
/// a new queue, use [`Self::another`].
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+            match self.poll_next() {
+                Some(msg) => break msg,
+                None => {
+                    self.nonempty.notified().await;
+                }
+            }
+        }
+    }
+
+    /// Return a new `TxHandle` to a new independent queue.
```
Razz4780:
```suggestion
    /// Return a new [`TxHandle`] to a new independent queue.
```

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+    type InMsg: bincode::Decode<()> + Send + fmt::Debug;
+    type OutMsg: bincode::Encode + Send + fmt::Debug;
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ProtocolError {
+    #[error("IO error: {0}")]
+    IO(#[from] io::Error),
+    #[error("unexpected peer message")]
+    UnexpectedPeerMessage,
```
Razz4780: Is this variant still used?

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -189,7 +190,7 @@ pub struct IncomingProxy {
     /// Each entry here maps to a request that is in progress both locally and remotely.
     http_gateways: ConnectionMap<HashMap<RequestId, HttpGatewayHandle>>,
     /// Running [`BackgroundTask`]s utilized by this proxy.
-    tasks: BackgroundTasks<InProxyTask, InProxyTaskMessage, InProxyTaskError>,
+    tasks: Option<BackgroundTasks<InProxyTask, InProxyTaskMessage, InProxyTaskError>>,
```
Razz4780, line 193: A lot of unwraps in the code, here and in `OutgoingProxy`. Does this field still need to be optional?

mirrord/protocol-io/src/lib.rs
MintSoup: Guarding against overflow. Essentially impossible to hit in normal code but maybe later down the line we accidentally create them in a loop or something like that and this should help contain the damage. Tbh I got the idea from Arcs impl (it does something similar for the refcounts) and thought it won't hurt to have here as well ¯_(ツ)_/¯

mirrord/intproxy/src/proxies/incoming.rs
MintSoup: The problem is that creating this field requires a `TxHandle`, which we do not receive until `run`. We could move it out of the struct and make it a local variable but then it would need to be passed down the callstack everywhere, and to me this option seems less bad.

MintSoup: We could also make the `TxHandle` field in `BackgroundTasks` optional but that's just moving the problem around, although it will result in fewer `unwrap`s ...

mirrord/protocol-io/src/lib.rs
Razz4780: Yeah but why divided by 2? :sweat_smile: 
I'm not gonna complain tho, comment can be resolved.

mirrord/intproxy/src/proxies/incoming.rs
Razz4780: Oke, let it be

Razz4780 (approved):

PR #3656, opened 2025-10-21
Add TypedHeader like extractor to mirrord-kube

DmitryDodzin:
`axum` like extraction from kubernetes resources

### DIFF
--- a/changelog.d/+reconnect-check-410.internal.md
+++ b/changelog.d/+reconnect-check-410.internal.md
@@ -1,0 +1,1 @@
+When reconnecting to an operator session, the CLI now checks for a fatal 410 error.
--- a/mirrord/intproxy/src/agent_conn.rs
+++ b/mirrord/intproxy/src/agent_conn.rs
@@ -10,8 +10,15 @@ use mirrord_config::LayerConfig;
 use mirrord_kube::{
     api::{kubernetes::AgentKubernetesConnectInfo, wrap_raw_connection},
     error::KubeApiError,
+    kube,
 };
-use mirrord_operator::client::{OperatorApi, OperatorSession, error::OperatorApiError};
+use mirrord_operator::{
+    client::{
+        OperatorApi, OperatorSession,
+        error::{OperatorApiError, OperatorOperation},
+    },
+    types::{RECONNECT_NOT_POSSIBLE_CODE, RECONNECT_NOT_POSSIBLE_REASON},
+};
 use mirrord_protocol::{ClientMessage, DaemonMessage};
 use serde::{Deserialize, Serialize};
 use strum::IntoDiscriminant;
@@ -22,7 +29,7 @@ use tokio::{
     net::{TcpSocket, TcpStream},
     sync::mpsc::{Receiver, Sender},
 };
-use tokio_retry::{Retry, strategy::ExponentialBackoff};
+use tokio_retry::{RetryIf, strategy::ExponentialBackoff};
 use tracing::Level;
 
 use crate::{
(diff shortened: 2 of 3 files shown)


mirrord/kube/src/extract.rs
```diff
@@ -0,0 +1,126 @@
+use std::sync::Arc;
+
+pub mod metadata;
+
+/// Types that can be created from kubernetes resource
+pub trait FromResource<'r, R, C>: Sized {
+    /// "rejection" type that will be used if extraction is failed.
+    type Rejection;
+
+    fn from_resource(resource: &'r R, context: Arc<C>) -> Result<Self, Self::Rejection>;
```
Razz4780: Why do we take `Arc`? We could be taking just `&C`

mirrord/kube/src/extract/metadata.rs
```diff
@@ -0,0 +1,165 @@
+use std::{borrow::Cow, sync::Arc};
+
+use kube::Resource;
+
+use super::{FromResource, OptionalFromResource};
+use crate::error::KubeApiError;
+
+/// Resource's `metadata.name` value, cannot be empty.
+#[derive(Debug)]
+pub struct Name<'r>(pub Cow<'r, str>);
```
Razz4780: I don't think we'll pass these wrappers around, so maybe we can do just:

```rust
pub struct Name<'r>(&'r str);
```

wdyt?

mirrord/kube/src/extract/metadata.rs
```diff
@@ -0,0 +1,165 @@
...
+impl<'r, R, C> OptionalFromResource<'r, R, C> for Name<'r>
+where
+    R: Resource,
+{
+    type Rejection = KubeApiError;
+
+    fn from_resource(resource: &'r R, _: Arc<C>) -> Result<Option<Self>, Self::Rejection> {
+        Ok(resource.meta().name.as_deref().map(Cow::Borrowed).map(Name))
+    }
+}
```
Razz4780: I wouldn't implement `OptionalFromResource` for the name. The name should always be there

mirrord/kube/src/extract/metadata.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    fn from_resource(resource: &'r R, _: Arc<C>) -> Result<Option<Self>, Self::Rejection> {
+        Ok(resource
+            .meta()
+            .namespace
+            .as_deref()
+            .map(Cow::Borrowed)
+            .map(Namespace))
+    }
+}
```
Razz4780: I wouldn't implement `OptionalFromResource` for namespace. We have the concrete resource type, so the namespace should be there either always or never

mirrord/kube/src/extract/metadata.rs
```diff
@@ -0,0 +1,165 @@
...
+            .meta()
+            .uid
+            .as_deref()
+            .map(Cow::Borrowed)
+            .map(Uid)
+            .ok_or_else(|| KubeApiError::missing_field(resource, ".metadata.namespace"))
+    }
+}
+
+impl<'r, R, C> OptionalFromResource<'r, R, C> for Uid<'r>
```
Razz4780: Same as name, should always be there

mirrord/kube/src/extract/metadata.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn into_inner(self) -> Cow<'r, str> {
+        self.0
+    }
+}
+
+impl<'r, R, C> FromResource<'r, R, C> for Name<'r>
+where
+    R: Resource<DynamicType = ()>,
+{
+    type Rejection = KubeApiError;
```
Razz4780, line 20: I'd use a simpler and more precise error type when possible.
Maybe move https://github.com/metalbear-co/operator/blob/2bdb220662d24f92669cec9a28487035dcc39cfa/operator/context/src/resource.rs#L44 here?

Razz4780 (approved):

mirrord/kube/src/extract/metadata.rs
DmitryDodzin: Also fine, we don't have `IntoResource` yet and not sure if needed at all

mirrord/kube/src/extract/metadata.rs
DmitryDodzin: Ah the annoying part is the error conversion then later, I do think for now using the KubeApiError is the simplest solution but after we have more prolific use of we can switch out the error to a more appropriate one

PR #3662, opened 2025-10-22
Send layer close

Razz4780:
Right now it's quite impossible for the operator to determine whether a mirrord-session has really finished.

This PR:
1. Adds a new meaning to the unused `ClientMessage::Close` message - now the message means that the whole mirrord session is finished.
2. Makes mirrord-intproxy always send the close message on exit

With this, the HA operator will be able to do the cleanup more quickly (before hitting session's idle timeout), at least in most cases.

Note: this change does not cover all mirrord-protocol connections (e.g. mirrord port-forward), but it's just a quick improvement - until we introduce the protocol utils (like [here](https://github.com/metalbear-co/mirrord/pull/3655)) and unify the way we use mirrord-protocol

### DIFF
--- a/changelog.d/+close-session-message.internal.md
+++ b/changelog.d/+close-session-message.internal.md
@@ -1,0 +1,1 @@
+mirrord-intproxy now always finishes the agent connection with `ClientMessage::Close`.


mirrord/intproxy/src/agent_conn.rs
```diff
@@ -254,6 +254,7 @@ impl BackgroundTask for AgentConnection {
                 msg = message_bus.recv() => match msg {
                     None => {
                         tracing::trace!("message bus closed, exiting");
+                        let _ = self.send(ClientMessage::Close).await;
```
DmitryDodzin, line 257: Will this be actually sent? when the message bus is dropped shouldn't it imply that the connection is closed?

Razz4780: The bus should be closed whenever the intproxy exists, e.g. after all layers have disconnected

PR #3667, opened 2025-10-23
Update mirrord cluster sessions for HA agents

DmitryDodzin:


mirrord/operator/src/crd/session.rs
```diff
@@ -90,15 +251,37 @@ pub struct SessionCopyTarget {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub connection_info: Option<AgentKubernetesConnectInfo>,
+    /// If the agent is of the ephemeral variaty
+    pub ephemeral: bool,
+    /// Agent spawn error if there is one.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub error: Option<ErrorMessage>,
+    /// The phase of agent's pod.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub phase: Option<String>,
```
t4lz, line 260: We could have a pod phase enum I guess. Up to you though, not important.

t4lz (approved):

mirrord/operator/src/crd/session.rs
DmitryDodzin: I would have wanted to make one but since the `Pod`'s status stores it as String I think I should keep it the same

mirrord/kube/src/api/kubernetes.rs
Razz4780, line 332: I wouldn't use this type for API that should be backwards compatible. Too easy to miss and change accidentally.

I think we should have a separate thing in the `mirrord-operator` crd module.

mirrord/kube/src/api/container.rs
```diff
@@ -33,7 +34,7 @@ pub static SKIP_NAMES: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
 });
 
 /// Configuration of the mirrord-agent container.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
```
Razz4780: Where do we serialize/deserialize it? If the representation is supposed to be backwards compatible, I wouldn't reuse this type. I'd make a new one in place where we usually put persistent stuff

mirrord/operator/src/crd/session.rs
```diff
@@ -14,6 +34,7 @@ use serde::{Deserialize, Serialize};
     printcolumn = r#"{"name":"HOSTNAME", "type":"string", "description":"User hostname.", "jsonPath":".spec.owner.hostname"}"#,
     printcolumn = r#"{"name":"K8S USER", "type":"string", "description":"User Kubernetes name.", "jsonPath":".spec.owner.k8sUsername"}"#,
     printcolumn = r#"{"name":"NAMESPACE", "type":"string", "description":"Namespace of the session.", "jsonPath":".spec.namespace"}"#,
+    printcolumn = r#"{"name":"AGENT_LIMIT", "type":"string", "description":"Relative or absolute limit of agents spawns.", "jsonPath":".spec.agentLimit"}"#,
```
Razz4780: 🗑️

mirrord/operator/src/crd/session.rs
```diff
@@ -28,6 +49,9 @@ pub struct MirrordClusterSessionSpec {
     pub owner: SessionOwner,
     /// Kubernetes namespace of the session.
     pub namespace: String,
+    /// State of concurrent steal
+    #[serde(default)]
+    pub on_concurrent_steal: ConcurrentSteal,
```
Razz4780: I don't think we need this here, the operator needs it only in the task that actually handles the websocket connection. It can read it from connect params, we don't need to put it here eagerly (not to mention that this setting is not really respected in the operator 💀 )

mirrord/operator/src/crd/session.rs
```diff
@@ -45,6 +69,56 @@ pub struct MirrordClusterSessionSpec {
     pub copy_target: Option<SessionCopyTarget>,
 }
 
+impl MirrordClusterSessionSpec {
+    pub fn as_target(&self) -> Result<Target, SessionTarget> {
```
Razz4780, line 66: Maybe `pub fn as_target(&self) -> Option<Target>` would be better?

mirrord/operator/src/crd/session.rs
```diff
@@ -90,15 +251,37 @@ pub struct SessionCopyTarget {
...
+    pub target: AgentTarget,
+}
+
+/// Describes an owner of a mirrord session.
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct MirrordClusterSessionStatus {
+    /// Running agents status
+    #[serde(default)]
+    pub agents: Vec<MirrordClusterSessionAgent>,
```
Razz4780, line 271: I'd consider putting this list in a wrapper struct, so that we can add more fields that don't relate to any specific agent. Like dunno, `.status.agents.something`

mirrord/operator/src/crd/session.rs
```diff
@@ -90,15 +251,37 @@ pub struct SessionCopyTarget {
 }
 
 /// Status of a mirrord session.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordClusterSessionAgent {
```
Razz4780, line 249: Do we need all of this info right now? 👀 
A minimal working implementation should need only a list of socket addresses - agents ready to accept client connections

mirrord/kube/src/api/container.rs
DmitryDodzin: Ops I don't need it to be serde

PR #3668, opened 2025-10-23
Remove a leftover file

0x00A5:
The whole `operator` folder is gone after https://github.com/metalbear-co/mirrord/pull/3619. This module is no longer needed, is it?


meowjesty (approved):
🫡

PR #3669, opened 2025-10-23
Adds a config for setting stdio pipe to a file for mirrord ci commands.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1059/mirrord-ci-start-support-sending-process-logs-to-file

### DIFF
--- a/changelog.d/+mirrord-kube-extract.internal.md
+++ b/changelog.d/+mirrord-kube-extract.internal.md
@@ -1,0 +1,1 @@
+Add TypedHeader like extractor to mirrord-kube for axum like extraction from kubernetes resources.
--- a/mirrord/kube/src/api/container/job.rs
+++ b/mirrord/kube/src/api/container/job.rs
@@ -31,6 +31,10 @@ use crate::{
         runtime::RuntimeData,
     },
     error::{KubeApiError, Result},
+    extract::{
+        FromResource,
+        metadata::{Name, Namespace},
+    },
 };
 
 pub async fn create_job_agent<P, V>(
@@ -74,18 +78,8 @@ where
         })?
         .map_err(|err| KubeApiError::AgentPodStartError(format!("watch stream failed: {err}")))?;
 
-    let pod_name = agent_pod
-        .metadata
-        .name
-        .as_ref()
-        .ok_or_else(|| KubeApiError::missing_field(&agent_pod, ".metadata.name"))?
-        .clone();
-    let pod_namespace = agent_pod
-        .metadata
-        .namespace
-        .as_ref()
-        .ok_or_else(|| KubeApiError::missing_field(&agent_pod, ".metadata.namespace"))?
-        .clone();
+    let (Name(pod_name), Namespace(pod_namespace)) = FromResource::from_resource(&agent_pod, &())?;
+
     pod_progress.success(Some(&format!(
         "agent pod {pod_namespace}/{pod_name} created"
     )));
@@ -169,7 +163,7 @@ where
         }
     }
 
-    let version = wait_for_agent_startup(&pod_api, &pod_name, "mirrord-agent".to_string()).await?;
+    let version = wait_for_agent_startup(&pod_api, pod_name, "mirrord-agent".to_string()).await?;
     match version.as_ref() {
         Some(version) if version != env!("CARGO_PKG_VERSION") => {
             let message = format!(
@@ -184,8 +178,8 @@ where
     pod_progress.success(Some("pod is ready"));
 
     Ok(AgentKubernetesConnectInfo {
-        pod_name,
-        pod_namespace,
+        pod_name: pod_name.to_owned(),
+        pod_namespace: pod_namespace.to_owned(),
         agent_port: params.port,
     })
 }
--- a/mirrord/kube/src/api/runtime.rs
+++ b/mirrord/kube/src/api/runtime.rs
@@ -27,6 +27,10 @@ use crate::{
         kubernetes::get_k8s_resource_api,
     },
     error::{KubeApiError, Result},
+    extract::{
+        FromResource,
+        metadata::{Name, Namespace},
+    },
     resolved::ResolvedTarget,
 };
 
@@ -108,18 +112,7 @@ impl RuntimeData {
     /// 2. pod is not in deletion,
     /// 3. target container is ready.
     pub fn from_pod(pod: &Pod, container_name: Option<&str>) -> Result<Self> {
-        let pod_name = pod
-            .metadata
-            .name
-            .as_ref()
-            .ok_or_else(|| KubeApiError::missing_field(pod, ".metadata.name"))?
-            .to_owned();
-        let pod_namespace = pod
-            .metadata
-            .namespace
-            .as_ref()
-            .ok_or_else(|| KubeApiError::missing_field(pod, ".metadata.namespace"))?
-            .to_owned();
+        let (Name(pod_name), Namespace(pod_namespace)) = FromResource::from_resource(pod, &())?;
 
         let phase = pod
             .status
@@ -231,8 +224,8 @@ impl RuntimeData {
 
         Ok(RuntimeData {
             pod_ips,
-            pod_name,
-            pod_namespace,
+            pod_name: pod_name.to_owned(),
+            pod_namespace: pod_namespace.to_owned(),
             node_name,
             container_id,
             container_runtime,
--- a/mirrord/kube/src/error.rs
+++ b/mirrord/kube/src/error.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{convert::Infallible, fmt};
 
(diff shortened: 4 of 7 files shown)


mirrord/config/src/ci.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct CiConfig {
+    /// ### ci.stdout_file {#ci-stdout_file}
+    ///
+    /// Pipe `stdout` to the file specified here.
+    pub stdout_file: Option<PathBuf>,
+
+    /// ### ci.stdin_file {#ci-stdin_file}
+    ///
+    /// Pipe `stdin` to the file specified here.
+    pub stdin_file: Option<PathBuf>,
```
0x00A5: We don't need a stdin file right?

mirrord/config/src/ci.rs
```diff
@@ -0,0 +1,54 @@
...
+///     "stdout_file": "/tmp/mirrord-ci-logs/stdout",
+///     "stdin_file": "/tmp/mirrord-ci-logs/stdin",
+///     "stderr_file": "/tmp/mirrord-ci-logs/stderr"
+///   }
+/// }
+/// ```
+#[derive(MirrordConfig, Clone, Debug, Serialize, Deserialize, PartialEq, Default)]
+#[config(map_to = "CiFileConfig", derive = "JsonSchema")]
+#[cfg_attr(test, config(derive = "PartialEq, Eq"))]
+pub struct CiConfig {
```
0x00A5, line 22: I am not sure this is future-proof enough. Maybe we just need an optional output directory path, and default to a tmp dir if none. The directory is not limited to logs, we might need it for other stuff like tracking process status etc. in the future.
For each client app process (run by ci start), we create a subfolder (named by the binary + unique suffix) for now and writes output and error to `stdout.log` and `stderr.log`.
WDYT?

mirrord/cli/src/ci.rs
```diff
@@ -197,18 +200,34 @@ impl MirrordCi {
...
+                    std::io::ErrorKind::NotFound,
+                    format!("Could not find parent directory for {full_path:?}!"),
+                ))?;
+
+                create_dir_all(parent).await?;
+                let stdio_file = File::create(full_path).await?;
+
+                Ok::<_, io::Error>(stdio_file.into_std().await.into())
+            }
+            None => Ok(Stdio::null()),
```
0x00A5: I wonder if we should just writes it to tmp dir if a destination is not specified by the user. And always print the location similar to what we do for int proxy log.

Razz4780:
Why do we need this? The child process should be inheriting stdout/stderr from the `mirrord ci` command. So if you want to pipe output to a file, you should be able to do just
`mirrord ci ... > file`.
Or no?

Razz4780:
Ah I see, we're currently using null streams. I think we should be inheriting instead

mirrord/config/src/ci.rs
meowjesty: Probably not. Will remove it.

meowjesty:
> I think we should be inheriting instead

So scrap this and just inherit? The user will always have to pipe it though, right, even if it's to `/dev/null`?

Razz4780:
> > I think we should be inheriting instead
> 
> So scrap this and just inherit? The user will always have to pipe it though, right, even if it's to `/dev/null`?

Yeah, that's what I'd do

aviramha:
I don't think that inheriting by default makes sense tbh. It runs in background by default hence it's easy to get messy

I would write to a temp file by default if user doesn't provide value and would also support getting this setting from env/command line

mirrord/config/src/ci.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/2168ae199bb9f07ed7d0be5254a60fc17aa2ee2c

mirrord/config/src/ci.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/e5d7274f170ea9c2020171c31c41b42f661f2c1f

Good point, now we create a dir and dump whatever files into it (stdio stuff for now).

mirrord/cli/src/ci.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/e5d7274f170ea9c2020171c31c41b42f661f2c1f

Prints it:

```
* internal proxy: logs will be written to /tmp/mirrord-intproxy-1761318550-fgi1ngJ.log
* mirrord ci files will be stored in /tmp/mirrord/node-1761318553-fig7sqJ
```

changelog.d/+config-for-stdio-mirrord-ci.added.md
```diff
@@ -0,0 +1 @@
+Adds 3 new mirrord configs `ci.stdout_file`, `ci.stdin_file`, and `ci.stderr_file`, so the user can pipe `stdio` to files specified in these configs.
```
0x00A5: This needs an update.

meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/09e7bf392b79fde021f1d9834475a091d7f9342a

changelog.d/+config-for-stdio-mirrord-ci.added.md
```diff
@@ -0,0 +1 @@
+`mirrord ci start` now pipes stdio output to a file under `/tmp/mirrord/{binary-name}-{timestamp}-{random}`. This can be configured with the `ci.info_dir` option in your mirrord config.
```
Razz4780:
```suggestion
`mirrord ci start` now pipes application's stdout and stderr to files in a temporary directory (`/tmp/mirrord/{binary-name}-{timestamp}-{random}` by default). This can be configured with the `ci.output_dir` option in your mirrord config.
```

mirrord/config/src/ci.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// ### ci.info_dir{#ci-info_dir}
+    ///
+    /// `mirrord ci` commands creates some temporary files (e.g. a file for the output of `stdio`),
+    /// and you can specify the directory where these files will be created here.
+    ///
+    /// Defaults to `/tmp/mirrord/{binary-name}-{timestamp}-{random-name}`
+    pub info_dir: Option<PathBuf>,
```
Razz4780:
```suggestion
    pub output_dir: Option<PathBuf>,
```

mirrord/config/src/ci.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(MirrordConfig, Clone, Debug, Serialize, Deserialize, PartialEq, Default)]
+#[config(map_to = "CiFileConfig", derive = "JsonSchema")]
+#[cfg_attr(test, config(derive = "PartialEq, Eq"))]
+pub struct CiConfig {
+    /// ### ci.info_dir{#ci-info_dir}
+    ///
+    /// `mirrord ci` commands creates some temporary files (e.g. a file for the output of `stdio`),
+    /// and you can specify the directory where these files will be created here.
+    ///
+    /// Defaults to `/tmp/mirrord/{binary-name}-{timestamp}-{random-name}`
```
Razz4780:
```suggestion
    /// Path to a directory where `mirrord ci` will flush application's stdout and stderr.
    ///
    /// Defaults to `/tmp/mirrord/`.
```

mirrord/cli/src/ci.rs
```diff
@@ -197,18 +202,54 @@ impl MirrordCi {
         binary_path: &Path,
         binary_args: &[String],
         env_vars: &HashMap<String, String>,
+        LayerConfig { ci: ci_config, .. }: &LayerConfig,
```
Razz4780: You could accept just the ci config as param 👀

mirrord/cli/src/ci.rs
```diff
@@ -197,18 +202,54 @@ impl MirrordCi {
...
+                        .await?
+                        .into_std()
+                        .await,
+                )
+                .stderr(
+                    File::create(ci_run_info_dir.join("stderr"))
+                        .await?
+                        .into_std()
+                        .await,
+                )
```
Razz4780: I'm not sure if this will work. The descriptor will be non blocking :/
I'd use `std::fs::File` instead of the tokio version

meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/490065a2232cd4a64e60cd186a747ae699385619

mirrord/cli/src/ci.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/3669/commits/490065a2232cd4a64e60cd186a747ae699385619

PR #3674, opened 2025-10-27
Check for 410 when reconnecting to the operator

Razz4780:


mirrord/intproxy/src/agent_conn.rs
```diff
@@ -313,25 +317,38 @@ impl RestartableBackgroundTask for AgentConnection {
                     .factor(500)
                     .max_delay(Duration::from_secs(8))
                     .take(10);
+                // Unless the operator responded with explicit 410 (meaning that the session is
+                // permanently gone), we can still retry.
+                let can_retry = |error: &AgentConnectionError| match error {
+                    AgentConnectionError::Operator(OperatorApiError::KubeError {
+                        error: kube::Error::Api(error),
+                        operation: OperatorOperation::WebsocketConnection,
+                    }) => error.code != 410,
```
DmitryDodzin: I would also check for 409 since it's basically the same error but for earlier versions of the operator

Razz4780: The thing that kinda worries me is that 409 is quite common. The operator might be returning it for some failed k8s api calls as well.
Now that I think of it, maybe we should use a const `error.reason` for this? Wdyt?

```rust
    AgentConnectionError::Operator(OperatorApiError::KubeError {
        error: kube::Error::Api(error),
        operation: OperatorOperation::WebsocketConnection,
    }) => error.code != 410 || error.reason != "ReconnectNotPossible",
```

DmitryDodzin: Yep, I was thinking about adding the earlier message too, like the previous error (before the status 410 update)

Razz4780: Oke, added

PR #3676, opened 2025-10-27
mirrord-sip: Make running arm64e binaries possible (COR-1025)

iniw:
Read [Aviram's backstory](https://linear.app/metalbear/issue/COR-1025/drop-the-need-for-rosettastop-running-x64-when-patching-sip) for this before reviewing the code, it contextualizes the problem and introduces the solution we're using

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -421,8 +421,8 @@ jobs:
     needs: changed_files
     if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
     env:
-      MIRRORD_TEST_USE_EXISTING_LIB: ../../target/x86_64-apple-darwin/debug/libmirrord_layer.dylib
-      MIRRORD_LAYER_FILE_MACOS_ARM64: ../../../target/debug/libmirrord_layer.dylib
+      MIRRORD_TEST_USE_EXISTING_LIB: ../../target/universal-apple-darwin/debug/libmirrord_layer.dylib
+      MIRRORD_LAYER_FILE_MACOS_ARM64: ../../../target/aarch64-apple-darwin/debug/libmirrord_layer.dylib
     steps:
       - uses: actions/checkout@v4 # Checkout the mirrord repo.
       # the setup rust toolchain action ignores the input if file exists.. so remove it
@@ -439,10 +439,8 @@ jobs:
       - name: verify SIP status
         run: csrutil status
       # Build layer before linting because the `mirrord` crate expects `MIRRORD_LAYER_FILE_MACOS_ARM64` to point to an existing file.
-      - name: build layer
-        run: cargo build -p mirrord-layer
-      - name: build layer x86
-        run: cargo build -p mirrord-layer --target=x86_64-apple-darwin
+      - name: build universal layer
+        run: ./scripts/build_fat_mac.sh
       - name: lint
         run: cargo clippy -p mirrord -p mirrord-layer -p mirrord-sip --tests -- -Wclippy::indexing_slicing -D warnings
       - name: mirrord-sip UT


mirrord/sip/src/lib.rs
```diff
@@ -345,18 +320,52 @@ mod main {
...
+        // https://github.com/llvm/llvm-project/blob/cbe7c49e93b630d3388dba2663b08a3c5c1bc8b6/llvm/include/llvm/BinaryFormat/MachO.h#L1642-L1678
+        // https://github.com/llvm/llvm-project/blob/7d1538cd3db3e228459e483ce9cdeb7fa4ae5e00/llvm/tools/llvm-objdump/MachODump.cpp#L8407-L8421
+        if (binary_info.cpu_subtype & !macho::CPU_SUBTYPE_MASK) == macho::CPU_SUBTYPE_ARM64E {
+            let versioned_ptrauth = (binary_info.cpu_subtype & macho::CPU_SUBTYPE_PTRAUTH_ABI) != 0;
+            if versioned_ptrauth {
+                const PTRAUTH_VERSION_MASK: u32 = 0x0f00_0000;
+                let ptrauth_version = (binary_info.cpu_subtype & PTRAUTH_VERSION_MASK) >> 24;
+                if ptrauth_version == 0 {
+                    // Override the ptrauth version to '1', allowing us to execute the file
+                    let new_subtype = binary_info.cpu_subtype | (1 << 24);
```
0x00A5, line 362: Oh the trick is telling the OS to run binary built against a preview ABI as if it is against the stable ABI :panik:

iniw: I'm not 100% sure if that's what the versions mean, I couldn't find any documentation on them, but yeah, I think that's the trick. We're leaning on the hope that v1 is backwards compatible.

changelog.d/+arm64e.added.md
```diff
@@ -0,0 +1 @@
+Mirrord can now run arm64e binaries, both when inside a universal binary and standalone. In practice, this means running arm64e+x86_64 universal binaries no longer has to go through the Rosetta emulation layer.
```
aviramha: *mirrord

iniw:
Green CI with the universal layer 🥳

.github/workflows/ci.yaml
```diff
@@ -439,10 +439,8 @@ jobs:
       - name: verify SIP status
         run: csrutil status
       # Build layer before linting because the `mirrord` crate expects `MIRRORD_LAYER_FILE_MACOS_ARM64` to point to an existing file.
-      - name: build layer
-        run: cargo build -p mirrord-layer
-      - name: build layer x86
-        run: cargo build -p mirrord-layer --target=x86_64-apple-darwin
+      - name: build universal layer
+        run: ./scripts/build_fat_mac.sh
```
0x00A5, line 443: nit: 👀 this builds two `mirrord` executable too, does it? We only need the fat layer for layer tests right?

iniw: Yes, it does, I even [talked about it on slack](https://metalbear.slack.com/archives/C08TYT624SV/p1761670694546309?thread_ts=1761669209.926679&cid=C08TYT624SV).

The problem is that we need an arm64+arm64e fat binary, which this scripts handles for us. Ideally we'd split this script into a "build arm fat binaries", that doesn't build for x86, and "build universal binaries", which does, with some shared logic between them, but this seems too much of a hassle rn.

The nice thing about doing it like this is that we'll be running the tests using the same binaries we use for our releases, but yeah, we could make `macos_tests` ~2x faster if we dropped x86 on CI

mirrord/sip/src/lib.rs
```diff
@@ -158,30 +130,67 @@ mod main {
...
+                    .iter()
+                    .find(|fat| fat.architecture() == arch)
+            };
+
+            #[cfg(target_arch = "aarch64")]
+            let found_arch =
+                find_arch(Architecture::Aarch64).or_else(|| find_arch(Architecture::X86_64));
+
+            #[cfg(target_arch = "x86_64")]
+            let found_arch = find_arch(Architecture::X86_64);
```
t4lz, line 181: I know the logic here didn't change, but I think we can improve it.
I think it's possible for an x86_64 executable to e.g. `exec` an aarch64 executable.
In that case, our layer, running loaded into the x86_64 executable would only look for an x86_64, and in the weird case where it's a fat MachO, and there is an aarch64 executable but no x86_64, we would fail unnecessarily.

If we add it as a fallback, all we have to lose is the nice error. Cause if it's an intel mac and there's no x86 - we fail anyways. Just, with the fallback the error might be really unhelpful (an error about wrong architecture). So it's probably good to try and determine if we're on a system that could run a aarch64 binary before trying that here.
And if we are, we actually might as well choose that binary first and only look for x86 as a fallback.

Does that make sense?

Probably not a common case, so you can do that in a separate PR if you prefer to not block this one for that.

t4lz (approved):

mirrord/sip/src/lib.rs
iniw: It does make sense, but on the other hand I think this case will be way less commmon now that the built-in MacOS binaries, which are commonly the first thing to run because of shebangs, will pick the arm64e executable instead of x86, so we're much more likely to never take the x86 path that locks us out of running arm64 code, as you described.

Will take a look at that in a separate PR though, mostly because this will involve runtime detection of the system's arch which sounds mildly annoying.

PR #3682, opened 2025-10-29
Added a github release

pavelzag:
This PR adds an idempotent “Create/Update Release + Upload” step so Linux/macOS artifacts are published to the same GitHub Release page as Windows (keyed by github.ref_name).

**What changed:**
New step uses getReleaseByTag → createRelease (on 404) → getReleaseByTag (on 422 race) and uploads all files from /tmp/release.
If an asset with the same filename exists, it’s deleted and re-uploaded; unrelated assets (e.g., mirrord.exe, signed DLLs) are untouched.
Build/sign/test steps remain unchanged.

**The change will allow:**
Single canonical release per tag with all platforms.
Race-safe, repeatable publishing (order doesn’t matter).
New tags: Unix + Windows land on the same page automatically.
Existing tags: Re-run the Unix Release for that tag to backfill Unix assets.


.github/workflows/release.yaml
```diff
@@ -203,21 +204,24 @@ jobs:
...
       - name: Build and push (final/release)
         if: github.event_name != 'workflow_dispatch'
         uses: docker/build-push-action@v6
         with:
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: |
+            type=registry,ref=ghcr.io/metalbear-co/mirrord-cli:buildcache
+          cache-to: |
+            type=registry,ref=ghcr.io/metalbear-co/mirrord-cli:buildcache,mode=max
```
t4lz, line 224: maybe this would be faster if the ci workflow would also use the same cache registry, and the branch would be used as a tag with a fallback to main (`/mirrord-cli:main` in a second cache-from line)?

That way on workflow_dispatch there is probably already a pretty good cache in the branch tag, and on final/release we can just directly use `main` as a tag, which would probably have very relevant cache already.

t4lz: Could be optimized in a separate PR though

PR #3685, opened 2025-10-29
INT-223 Configurable DB branching timeout

vladrbg:


mirrord/config/src/feature/database_branches.rs
```diff
@@ -243,3 +251,7 @@ impl CollectAnalytics for &DatabaseBranchesConfig {
 fn default_ttl_secs() -> u64 {
     300
 }
+
+pub fn default_creation_timeout_secs() -> u64 {
```
0x00A5, line 255: Do we need `pub`?

mirrord/operator/src/client.rs
```diff
@@ -1288,8 +1290,24 @@ impl OperatorApi<PreparedClientCert> {
...
+        let timeout_secs = layer_config
+            .feature
+            .db_branches
+            .iter()
+            .map(|branch_config| match branch_config {
+                mirrord_config::feature::database_branches::DatabaseBranchConfig::Mysql(
+                    mysql_config,
+                ) => mysql_config.base.creation_timeout_secs,
+            })
+            .max()
```
0x00A5, line 1304: Hmm, I was thinking to let each branch's `await_condition` wrapped by its own timeout since it's configured separately. But this makes sense too and simplifies the logic.

mirrord/config/src/feature/database_branches.rs
0x00A5: Oh, it is used as a fallback also in the other crate.

PR #3687, opened 2025-10-30
𓂃𓏲࣪  win-47 °˖ file filters + file remapper for windows

cristeigabriela:

### DIFF
--- a/mirrord/cli/Cargo.toml
+++ b/mirrord/cli/Cargo.toml
@@ -70,11 +70,12 @@ fs4.workspace = true
 hex.workspace = true
 tower = { workspace = true, features = ["retry"] }
 is_ci = "1.2.0"
+
+[target.'cfg(unix)'.dependencies]
 rand.workspace = true
 
 [target.'cfg(target_os = "windows")'.dependencies]
 mirrord-layer-lib = { path = "../layer-lib", features = ["cli-execution"] }
-
 
 [target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
 nix = { workspace = true, features = ["process", "resource", "signal"] }
--- a/mirrord/cli/src/ci.rs
+++ b/mirrord/cli/src/ci.rs
@@ -3,11 +3,10 @@ use std::os::unix::process::ExitStatusExt;
 use std::{
     collections::HashMap,
     env::{self, temp_dir},
-    fs::File,
     path::{Path, PathBuf},
-    process::Stdio,
-    time::SystemTime,
 };
+#[cfg(unix)]
+use std::{fs::File, process::Stdio, time::SystemTime};
 
 use drain::Watch;
 use fs4::tokio::AsyncFileExt;
@@ -16,12 +15,12 @@ use mirrord_auth::credentials::CiApiKey;
 use mirrord_config::{LayerConfig, ci::CiConfig, config::ConfigContext};
 use mirrord_operator::client::OperatorApi;
 use mirrord_progress::{Progress, ProgressTracker};
+#[cfg(unix)]
 use rand::distr::{Alphanumeric, SampleString};
 use serde::{Deserialize, Serialize};
-use tokio::{
-    fs::{self, create_dir_all},
-    io::AsyncWriteExt,
-};
+#[cfg(unix)]
+use tokio::fs::create_dir_all;
+use tokio::{fs, io::AsyncWriteExt};
 use tracing::Level;
 
 use crate::{
@@ -157,6 +156,7 @@ impl MirrordCiStore {
     }
 
     /// Removes the [`MirrordCiStore`] file at [`Self::MIRRORD_FOR_CI_TMP_FILE_PATH`].
+    #[cfg_attr(windows, allow(unused))]
     async fn remove_file() -> CiResult<()> {
         match tokio::fs::remove_file(temp_dir().join(Self::MIRRORD_FOR_CI_TMP_FILE_PATH)).await {
             Ok(_) => Ok(()),
@@ -313,6 +313,7 @@ impl MirrordCi {
     }
 
     #[cfg_attr(windows, allow(unused))]
+    #[cfg(windows)]
     pub(super) async fn prepare_command<P: Progress>(
         self,
         progress: &mut P,
@@ -350,6 +351,7 @@ impl MirrordCi {
 
 #[derive(Debug, Default)]
 struct StartArgs {
+    #[cfg_attr(windows, allow(unused))]
     foreground: bool,
 }
 


mirrord/layer-win/src/hooks/files/mod.rs
```diff
@@ -223,117 +257,232 @@ unsafe extern "system" fn nt_create_file_hook(
...
-
-                    try_insert_handle(HandleContext {
-                        path: linux_path.clone(),
-                        fd: file.fd,
+            Some(FileMode::NotFound(_)) => {
+                *file_handle = std::ptr::null_mut();
+                *io_status_block = _IO_STATUS_BLOCK::default();
+                return STATUS_OBJECT_PATH_NOT_FOUND;
+            }
+            // TODO(gabriela): edit when supported!
```
itsamegraf, line 328: please a comment about this "rest" case, ReadOnly and ReadWrite means remote necessarily? what does the None case mean? how come its the same as "remote"?
additionaly please add an explanation for what does the `desired_access & MODE_CONST != 0` rerpesents - I assume it's if any of them is set, we should explicitly note it.
additionally, please dont ignore the "else" case of that if,atleast log it explicitly so we know we are fallthrough-ing due to missing required access

mirrord/layer-win/src/hooks/files/mod.rs
```diff
@@ -223,117 +257,232 @@ unsafe extern "system" fn nt_create_file_hook(
...
 
-            if managed_handle.is_some() {
-                // Write managed handle at the provided pointer
-                *file_handle = *managed_handle.unwrap();
-                tracing::info!(
-                    "nt_create_file_hook: Succesfully opened remote file handle for {} ({:8x})",
-                    linux_path,
-                    *file_handle as usize
+        // Check if pointer to handle is valid.
+        if file_handle.is_null() {
```
itsamegraf, line 345: can we refrain from copying the origin(...) call so many times? i used to do it either by helper function that returns result where Err calls the original, or use a closure that does not repeat the entire parameter list.
this is cosmetic, but will help with maintenance in the future

mirrord/agent/Dockerfile
```diff
@@ -31,7 +29,7 @@ RUN cargo +nightly-2025-08-01 chef prepare --recipe-path recipe.json
 FROM chef AS builder
 
 COPY --from=planner /app/recipe.json recipe.json
-RUN cargo +nightly-2025-08-01 chef cook --release --zigbuild --target $(cat /.platform) --recipe-path recipe.json
+RUN cargo +nightly-2025-08-01 chef cook --release --zigbuild --recipe-path recipe.json
```
itsamegraf: please revert the changes in this file

mirrord/cli/src/ci/error.rs
```diff
@@ -31,6 +31,7 @@ pub(crate) enum CiError {
         the running mirrord session, and we could not retrieve this pid. You can manually stop mirrord \
         by searching for the pid with `ps | grep mirrord` and calling `kill [pid]`."
     ))]
+    #[cfg(not(target_os = "windows"))]
```
itsamegraf, line 34: these should disapper once you fix conflict

mirrord/layer-win/src/hooks/files/mod.rs
cristeigabriela: > please a comment about this "rest" case, ReadOnly and ReadWrite means remote necessarily? what does the None case mean? how come its the same as "remote"?

read-only and read-write affect the permissions you get for a remote file, if it is also present on the remote, and you try to access it.

> what does the None case mean?

simply that there's no filter. `file_filter.check(...)` returns an `Option<filter::FileMode>`

> additionaly please add an explanation for what does the desired_access & MODE_CONST != 0 rerpesents

sure, i thought it's descriptive however, it just checks for *any* attempt to get a write mode handle to the file

> please dont ignore the "else" case of that if,atleast log it explicitly so we know we are fallthrough-ing due to missing required access

this is a bad idea, because 99% of the cases are that.

mirrord/layer-win/src/hooks/files/mod.rs
cristeigabriela: implemented a closure

changelog.d/+add-gopath-to-exclude.changed.md
```diff
@@ -0,0 +1 @@
+Exclude `GOPATH` and `GOMODCACHE` from env fetched
```
itsamegraf: is this intended?

mirrord/layer-win/src/hooks/files/mod.rs
```diff
@@ -207,23 +206,49 @@ static NT_CLOSE_ORIGINAL: OnceLock<&NtCloseType> = OnceLock::new();
...
+/// 1. We check if the layer is configured to have file-system hooks enbled or not.
+///     * If not, jump to ["Fallback"](#fallback).
+///     * This is the only point we have to check for this configuration. Without
+///       [`managed_handle::MirrordHandle`]s, there is no other logic.
+/// 2. We try to convert the path from `object_attributes` to a Unix path.
+///     * If we fail, jump to ["Fallback"](#fallback).
+/// 3. We run the Unix path through
+///    [`mirrord_layer_lib::file::mapper::FileRemapper::change_path_str`] to account for file-system
+///    configuration, This becomes the new "path".
+///     * If the result of this operation is different from the intiial path, this is logged.
```
itsamegraf: question - initial path is the converted unix path or the original input parameter?

mirrord/layer-win/src/hooks/files/mod.rs
```diff
@@ -233,117 +258,172 @@ unsafe extern "system" fn nt_create_file_hook(
...
+                share_access,
+                create_disposition,
+                create_options,
+                ea_buf,
+                ea_size,
+            )
+        };
+
+        let setup = layer_setup();
+        // If file-system hooks are enabled, don't proceed with logic for mirrord handles.
```
itsamegraf:
```suggestion
        // If file-system hooks are disabled, don't proceed with logic for mirrord handles.
```
?

mirrord/layer-win/src/hooks/files/mod.rs
itsamegraf: i just read the code - it looks like its the unix path, the `initial` naming is a bit confusting :)

mirrord/layer-win/src/hooks/files/mod.rs
itsamegraf: > read-only and read-write affect the permissions you get for a remote file, if it is also present on the remote, and you try to access it.

do we restrict the permissions on the local handle somehow? 

> please dont ignore the "else" case of that if,atleast log it explicitly so we know we are fallthrough-ing due to missing required access
this is a bad idea, because 99% of the cases are that.

now i understand better this match match matched_filter is just a group of check to "exit early on unsupported or errournous usage".  thank you for the answers

mirrord/layer-win/src/hooks/mod.rs
```diff
@@ -20,13 +20,10 @@ pub fn initialize_hooks(guard: &mut DetourGuard<'static>) -> LayerResult<()> {
...
-    if setup.fs_hooks_enabled() {
-        tracing::info!("Enabling file system hooks");
-        files::initialize_hooks(guard)?;
-    } else {
-        tracing::info!("File system hooks disabled by configuration (fs.mode = Local)");
-    }
+    // NOTE(gabriela): currently I believe the ideal way to handle this is
+    // through hook-level checks
+    tracing::info!("Enabling file system hooks");
+    files::initialize_hooks(guard)?;
```
itsamegraf, line 26: If you insist on this change i will not push-back.
My 2 cents:
I follow the existing convention of unix layer as set in `layer\src\lib.rs` to not hooks functions we know from configuration are unneeded for the current execution.
Personally i tend to agree with that approach - If we enable all hooks, we force more flows to run through our code which reduces performance and additionally introduces more opportunities for mirrord to fck-up existing logic flows that might be unrelated to what user is attempting to do. (bugs are inevitable)
Here's an example case for explaining the issue with enabling hooks always: user wants only fs ops to go through mirrord but due to a socket hooks bug he is unable to use mirrord at all - an unrelated code flow causes him frustration due to unnecessary friction.

mirrord/layer-lib/src/process/windows/console.rs
```diff
@@ -62,7 +62,6 @@ pub fn ensure_vt_or_dumb_progress() {
             } else {
                 warn!(error = ?error, "SetConsoleMode probe on stdout failed");
             }
-            return;
```
itsamegraf, line 65: nice, thank you :)

mirrord/layer-lib/src/file/filter.rs
```diff
@@ -0,0 +1,137 @@
+use std::path::Path;
```
itsamegraf, line 1: should be beneath the following comment

mirrord/layer/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+// NOTE(gabriela): prevent compiling lib.rs so layer acts as empty
+// library to allow layer-lib as optional dependency for unix.
```
itsamegraf, line 2: cool i didn't know preventing the lib.rs compilation would be like guarding out the entire crate

mirrord/layer-lib/src/file/windows/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use regex::{RegexSet, RegexSetBuilder};
+
+/// List of files that mirrord should use locally, as they probably exist only in the local user
+/// machine, or are system configuration files (that could break the process if we used the remote
+/// version).
+///
+/// You most likely do **NOT** want to include any of these, but if have a reason to do so, then
+/// setting any of the overrides - `MIRRORD_FILE_X_PATTERN` allows you to override this list.
+pub fn generate_local_set() -> RegexSet {
```
itsamegraf: can we have the equivalent functions (`generate_local_set` and `generate_remote_ro_set`) be defined in mod.rs instead of duplicating them?

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
```diff
@@ -0,0 +1,17 @@
+use regex::RegexSetBuilder;
+
+/// This is the list of path patterns that are read locally by default in all fs modes. If you want
+/// to read or write in the cluster a path covered by those patterns, you need to include it in a
+/// pattern in the `feature.fs.read_only` or `feature.fs.read_write` configuration field,
+/// respectively.
+pub fn regex_set_builder() -> RegexSetBuilder {
+    let patterns: Vec<String> = [
+        // Python folder on Windows.
+        r"^/Users/[^/]+/AppData/Local/Programs/Python/",
```
itsamegraf, line 24: the linux version includes file types that should be implicitly read locally, why not have something similar here? please go over that list and pick the values that can also be useful for our case. just from a first glance - py and pyc

mirrord/layer-lib/src/file/windows/read_remote_by_default.rs
```diff
@@ -0,0 +1,3 @@
+/// These paths will be read remotely by default when `fs.feature.mode` is set to
+/// `localwithoverrides`.
+pub const PATHS: [&str; 0] = [];
```
itsamegraf, line 3: because our remote is linux, I think we should have the same `read_remote_by_default` for unix and windows, it will be useful for programs reading these linux paths explicitly. wdyt?

mirrord/layer/src/file/filter.rs
```diff
@@ -1,479 +0,0 @@
...
-
-    let patterns = not_found_by_default::PATHS
-        .into_iter()
-        .map(|cloud_dir| format!("^{home_clean}/{cloud_dir}"));
-
-    RegexSetBuilder::new(patterns)
-        .case_insensitive(true)
-        .build()
-        .expect("Building not found path regex set failed")
-}
```
itsamegraf, line 66: please mimic this behaviour (of having not_found paths relative to user's home directory), as documented at the top of `not_found_by_default`

mirrord/layer-lib/src/file/windows/not_found_by_default.rs
```diff
@@ -0,0 +1,5 @@
+/// When running in a mode other than `local`, mirrord will make these paths under the running
+/// user's home directory be not found, unless they are covered by patterns in the
+/// `feature.fs.read_only` or `feature.fs.read_write` pattern sets (and not in the
+/// `feature.fs.not_found` set).
+pub const PATHS: [&str; 0] = [];
```
itsamegraf: Hey @t4lz, I saw in [#2017](https://github.com/metalbear-co/mirrord/pull/2017/files#diff-3d5bbb7fa843f9e0cb9b3ce865ed265ad45470c864cf678a9b66263c2b196b11R5)  you introduced a default not_found list with with some cloud provider values. 
I'm not really sure why you added them (why would want them not be found?), could you share your opinion whether we should also have the same values for the windows client? I don't really understand the usecase, mostly for the sake of uniformity between the platforms.

itsamegraf: @cristeigabriela 
well, tal is on PTO, for now - lets assume we will want the same values (which are relative to user homedir). please use the same not_found_by_default on windows and unix.
unless his response say otherwise (after he's back from pto next week), we should follow layer's behaviour as it is now.

mirrord/layer-win/src/hooks/files/util.rs
```diff
@@ -74,7 +81,14 @@ pub fn try_seek(fd: u64, seek: SeekFromInternal) -> Option<u64> {
 
     match seek {
         Ok(Ok(res)) => Some(res.result_offset),
-        _ => None,
+        Ok(Err(e)) => {
+            tracing::error!(?e, "Error trying to seek into file!");
+            None
+        }
+        Err(e) => {
+            tracing::error!(?e, "Error trying to seek into file!");
```
itsamegraf: please differentiate the errors between the Ok(Err()) and Err() branches

itsamegraf (changes requested):
Great work, I left a few comments. additionally, please make sure CI is green.

mirrord/layer-lib/src/file/windows/mod.rs
cristeigabriela: if you look at the Unix logic, you'll see that there is some platform-dependent code that exists now, and some might even exist in the future. I don't think it's a good idea to mix this up as far as it can be avoided, but I would be in support of a trait

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
cristeigabriela: I think it's better to approach this slowly by having us test each case and implement this filtering where need be. In my testing, I used a lot of Python, and I noticed this case, so I added it.

Defaults are not removable by the user so I would prefer to avoid bugs

mirrord/layer-lib/src/file/windows/read_remote_by_default.rs
cristeigabriela: The input paths are the platform paths, we will never have meaningful remote paths besides maybe configs or w/e

mirrord/layer-lib/src/file/windows/mod.rs
itsamegraf: image
i did look at the unix logic, this is a comparison of both mod.rs
please direct me to the platform dependant code you are refering to

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
itsamegraf: according to the comments they can be overriden by explicitly mentioning them in an include filter, am i missing anything? why does it work for unix but wouldnt work for windows? 
I'm in favor of having the same behavior for both platforms, the tests might rely on it.

mirrord/layer-lib/src/file/windows/mod.rs
cristeigabriela: on Windows, the logic needs to look something like
```
    let Ok(Some(home)) = env::var("USERPROFILE").map(path_to_unix_path) else {
        tracing::warn!("Unable to resolve %USERPROFILE% directory, generating empty not-found set");
        return Default::default();
    };
```
on Unix, 
```
    let Ok(home) = env::var("HOME") else {
        tracing::warn!("Unable to resolve $HOME directory, generating empty not-found set");
        return Default::default();
    };
```

I think it's better to avoid joinng the files to avoid mixing Unix and Windows logic in the same file

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
cristeigabriela: oh, that's my mistake! i misread but yes that is supported, i'll comb through to find more cases

mirrord/layer-lib/src/file/windows/not_found_by_default.rs
t4lz: Hi, that PR just moves that path list to its own file.
That config field and its defaults were added in https://github.com/metalbear-co/mirrord/pull/1825 as a solution to https://github.com/metalbear-co/mirrord/issues/1694.
I would set the equivalent windows file paths on windows - I don't see why not.

t4lz: BTW, the reason to move all the default path pattern lists to their own files was so that we can link to the from the docs - so please make sure you update the docs if you move those lists, and add a link to the windows path list if it's different.

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
cristeigabriela: added

mirrord/layer/src/file/filter.rs
cristeigabriela: done

mirrord/layer-lib/src/file/windows/mod.rs
itsamegraf: we dont have to join all of them, we can have the different function logic (generate_not_found_set) separated into the windows/unix files, and the rest can be in mod.rs. if we ever need to differentiate them per-platform, we can always split it back.
alternatively we can always implement `generate_not_found_set` in mod.rs as well with platforms specific guards just for the `USERPROFILE`/`HOME` env vars retrieval portion, then the rest should be generic.

mirrord/layer-lib/src/file/windows/read_local_by_default.rs
itsamegraf: 🙏

mirrord/str-win/src/lib.rs
```diff
@@ -264,3 +266,27 @@ pub unsafe fn find_multi_buffer_safe_len<T: MultiBufferChar>(
 
     None // Double null terminator not found within bounds
 }
+
+// This prefix is a way to explicitly indicate that we're looking in
+// the global namespace for a path.
+const GLOBAL_NAMESPACE_PATH: &str = r#"\??\"#;
+
+/// Responsible for turning global namespace, default volume paths into Linux paths.
```
itsamegraf: please document here the process for translating a "path" to "unix_path" - (optionally remove global namespace, remove root drive, swap slash direction)

mirrord/layer-lib/src/file/windows/not_found_by_default.rs
itsamegraf: Tal, Thank you for taking the time to answer during your PTO, much appreciated 🙏 
@cristeigabriela, wdyt?

mirrord/layer-lib/src/file/windows/mod.rs
cristeigabriela: oops, i accidentally sent in a wrong comment before. if you got it by mail, please ignore it!

i'm willing to concede on sharing the generators if you insist, by implementing a "get_home_dir" function

mirrord/layer-lib/src/file/windows/not_found_by_default.rs
cristeigabriela: noted, implemented, will get to external docs when we next merge to main

mirrord/layer-win/src/hooks/files/util.rs
cristeigabriela: done

mirrord/layer-lib/src/file/windows/mod.rs
cristeigabriela: decided to implement this

mirrord/layer-lib/src/file/filter.rs
itsamegraf: unresolving as its not fixed, please fix

cristeigabriela: cargo fmt decides that it should be there, though

itsamegraf: I worked on resolving the conflicts for the pr so did it on the way - fmt did not resist me moving it to there.

itsamegraf:
I attempted to fix the conflicts but had issues and not enough time to fix them.
reverted the PR to the state it was before i started the work.
Please fix the conflicts and get green CI

PR #3691, opened 2025-10-30
Listen ports always take effect, even when port is ignored

gememma:
Issue: https://linear.app/metalbear/issue/COR-1014/listen-ports-should-affect-non-stolenmirrored-ports

Things in this PR:
- some typo corrections in docs (just in the files I was working in, nothing extensive)
- removed wildcard matches when matching on variants of `SocketState`, now we match on all states explicitly
- Altered `SocketState::Bound` variant to have `is_only_bound: bool` which is true when we do not listen on the socket (in practice this occurs when a port is local only but _is_ specified in `listen_ports`)
- Unified `Bypass` variants `Port` and `IgnoreLocalhost` into `IgnoredInIncoming`, which covers both

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -265,10 +265,10 @@ pub(super) fn bind(
         Err(HookError::AddressAlreadyBound(requested_address))?;
     }
 
-    let is_requested_in_listen = incoming_config
+    let listen_port = incoming_config
         .listen_ports
         .get_by_left(&requested_address.port())
-        .is_some();
+        .copied();
 
     // we don't use `is_localhost` here since unspecified means to listen
     // on all IPs.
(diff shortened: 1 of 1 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -333,10 +336,23 @@ pub(super) fn bind(
         .and_then(|(_, address)| address.as_socket())
         .bypass(Bypass::AddressConversion)?;
 
-    Arc::get_mut(&mut socket).unwrap().state = SocketState::Bound(Bound {
-        requested_address,
-        address,
-    });
+    Arc::get_mut(&mut socket).unwrap().state = if will_not_trigger_subscription {
```
Razz4780: Looks like this `if` can now be collapsed : D

mirrord/layer/src/socket/ops.rs
```diff
@@ -272,14 +257,32 @@ pub(super) fn bind(
...
                 }
             })
     {
         Err(HookError::AddressAlreadyBound(requested_address))?;
     }
 
+    let is_requested_in_listen = incoming_config
+        .listen_ports
+        .get_by_left(&requested_address.port())
+        .is_some();
```
Razz4780: Since we get this port later again, we can get it early here (and use `listen_port.is_some()` in boolean expressions)

Razz4780 (approved):

PR #3694, opened 2025-10-31
Update tera to the newest version, resolving some security advisories

iniw:
CC @cristeigabriela This affects some windows-specific deps on the lockfile, could you double check everything is still fine?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2206,7 +2206,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2462,7 +2462,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -2206,7 +2206,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
```
0x00A5, line 2209: Are these windows-* deps downgrade required for things to compile? `0.61.2` seems still in the deps tree though. Try reverting these downgrade?

iniw: Cargo did this automatically when I ran `cargo update tera` so I assume that they are required to resolve the entire dependency graph but I'll try reverting it locally and see what happens

0x00A5 (approved):

Cargo.lock
0x00A5: Huh, TIL: https://github.com/rust-lang/cargo/issues/5702

iniw:
```console
$ cargo update windows-sys
error: There are multiple `windows-sys` packages in your project, and the specification `windows-sys` is ambiguous.
Please re-run this command with one of the following specifications:
  windows-sys@0.48.0
  windows-sys@0.52.0
  windows-sys@0.59.0
  windows-sys@0.60.2
  windows-sys@0.61.2
```
  
I'm not fully sure what kind of wizardry cargo does to deal with this but considering the Windows build on CI succeeded I assume it Just Works™ (pending Gabriela's test)

cristeigabriela:
@iniw looks good! sorry for the delay \(˚☐˚”)/

PR #3696, opened 2025-11-03
Create ADOPTERS.md and CTA to README.md

Oshratn:
This PR introduces an ADOPTERS.md file to showcase organizations using mirrord. It includes a table of adopters with hyperlinked company names, use cases, and links to detailed user stories hosted on MetalBear.com
In addition, the README.md has been updated in the Help and Community section to include a call-to-action inviting users to add themselves to the adopters list via pull request.

Changes:

Added ADOPTERS.md with initial list of adopters

Updated README.md with CTA linking to ADOPTERS.md

The goal is to make adoption more visible and encourage community participation.

### DIFF
--- a/changelog.d/+add-adopters-md.added.md
+++ b/changelog.d/+add-adopters-md.added.md
@@ -1,0 +1,1 @@
+Added ADOPTERS.md


ADOPTERS.md
```diff
@@ -0,0 +1,23 @@
+# ADOPTERS.md
```
eyalb181, line 1: A. Lets use User Story instead of Case Study just to stay consistent. We use User Story for the OSS user stories and Case Studies for the interviews with paid customers.
B. The Use Case column is confusing IMO. mirrord doesn't actually have that many use cases. A lot of these are either different phrasings for the same thing. I think it should be removed altogether.
C. If I'm a contributor opening a PR, what do I put under More Info?

Oshratn: Good points!
Made additional changes.

PR #3697, opened 2025-11-03
Build layer only for macos test in CI

0x00A5:
We started using `./scripts/build_fat_mac.sh` for building fat binary in CI after https://github.com/metalbear-co/mirrord/pull/3676. To run macos test (layer tests on macos), we don't need a the CLI binary. This PR changes it so only layer dylib is built.

### DIFF
--- a/scripts/build_fat_mac.sh
+++ b/scripts/build_fat_mac.sh
@@ -7,8 +7,7 @@
 
 set -e
 
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
-"$SCRIPT_DIR/build_layer_mac.sh" "$@"
+./scripts/build_layer_mac.sh "$@"
 
 # build mirrord package - both targets require MIRRORD_LAYER_FILE_MACOS_ARM64 var
 MIRRORD_LAYER_FILE_MACOS_ARM64=../../../target/aarch64-apple-darwin/debug/libmirrord_layer.dylib MIRRORD_LAYER_FILE=../../../target/universal-apple-darwin/debug/libmirrord_layer.dylib cargo build "$@" -p mirrord --target=aarch64-apple-darwin


scripts/build_fat_mac.sh
```diff
@@ -7,24 +7,8 @@
...
-
-# create shim to always load arm64 and sign it
-clang -arch arm64e -dynamiclib -o target/universal-apple-darwin/debug/shim.dylib mirrord/layer/shim.c
-codesign -f -s - target/universal-apple-darwin/debug/shim.dylib
-
-# Create universal binary for mirrord-layer from x86 and shim, as well as arm64
-lipo -create -output target/universal-apple-darwin/debug/libmirrord_layer.dylib target/x86_64-apple-darwin/debug/libmirrord_layer.dylib target/universal-apple-darwin/debug/shim.dylib target/aarch64-apple-darwin/debug/libmirrord_layer.dylib
-codesign -f -s - target/universal-apple-darwin/debug/libmirrord_layer.dylib
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+"$SCRIPT_DIR/build_layer_mac.sh" "$@"
```
iniw: This script is already required to be executed from the root of the project so this can be simplified to `./scripts/build_layer_mac.sh`

0x00A5: Yep, replaced with `./scripts/build_layer_mac.sh`.

scripts/build_fat_mac.sh
```diff
@@ -7,24 +7,7 @@
...
-mkdir -p target/universal-apple-darwin/debug
-
-# create shim to always load arm64 and sign it
-clang -arch arm64e -dynamiclib -o target/universal-apple-darwin/debug/shim.dylib mirrord/layer/shim.c
-codesign -f -s - target/universal-apple-darwin/debug/shim.dylib
-
-# Create universal binary for mirrord-layer from x86 and shim, as well as arm64
-lipo -create -output target/universal-apple-darwin/debug/libmirrord_layer.dylib target/x86_64-apple-darwin/debug/libmirrord_layer.dylib target/universal-apple-darwin/debug/shim.dylib target/aarch64-apple-darwin/debug/libmirrord_layer.dylib
-codesign -f -s - target/universal-apple-darwin/debug/libmirrord_layer.dylib
+./scripts/build_layer_mac.sh
```
iniw: Forgot to pass `"$@"` after this change 😅

PR #3698, opened 2025-11-03
Prioritize using the fs filters on relative file ops

iniw:
This will now run *all* fs filters before checking if path is relative, let me know if we only want to specifically run `fs.not_found`

### DIFF
--- a/mirrord/layer/src/file/ops.rs
+++ b/mirrord/layer/src/file/ops.rs
@@ -45,14 +45,19 @@ const MAX_READ_SIZE: u64 = 1024 * 1024;
 
 /// Convenience extension for verifying that a [`Path`] is not relative.
 trait PathExt {
-    /// If this [`Path`] is relative, returns a [`Detour::Bypass`].
-    fn ensure_not_relative(&self) -> Detour<()>;
+    /// If this [`Path`] is relative and is not present in the `fs.not_found` filters, returns a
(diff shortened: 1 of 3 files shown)


mirrord/layer/src/file/ops.rs
```diff
@@ -62,15 +62,21 @@ impl PathExt for Path {
...
-    path.ensure_not_relative()?;
+    // `path` will be moved on the call to `change_path` so we compute this here even though the
+    // result will only be used after `ensure_remote` is called
+    let bail_if_relative = path.ensure_not_relative();
+
     let path = crate::setup().file_remapper().change_path(path);
     crate::setup().file_filter().ensure_remote(&path, write)?;
+
+    bail_if_relative?;
+
```
Razz4780: This can change the bypass returned for relative paths. Can we do this instead?

```rust
if path.is_relative() {
    // Note that `check_not_found` does not exist yet,
    // has to be implemented
    if crate::setup().file_filter().check_not_found(&path) {
        ... // HookError::FileNotFound
    } else {
        ... // Bypass::relative_path
    }
}

... // rest, unchanged
```

Razz4780 (changes requested):

mirrord/layer/src/file/ops.rs
iniw: Should be good now, also handled `unlinkat` which I previously fogor 💀

Razz4780 (approved):

PR #3712, opened 2025-11-09
Add support for HTTP json body filters

MintSoup:


mirrord/protocol/src/tcp.rs
```diff
@@ -257,6 +263,9 @@ pub enum HttpFilter {
     },
     /// Filter by method ("POST")
     Method(HttpMethodFilter),
+
+    /// Filter by body
+    Body(HttpBodyFilter),
```
Razz4780, line 268: I don't see any version req static for sending this variant, there should be one (like [here](https://github.com/metalbear-co/mirrord/blob/683f633561e0d3ca1fd47420e33b1d8d62acef6c/mirrord/protocol/src/tcp.rs#L483))

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -242,6 +249,111 @@ pub enum InnerFilter {
...
+    ///       "path": "/buildings"
+    ///     },
+    ///     {
+    ///       "type": "json",
+    ///       "query": "$.library.books[*]",
+    ///       "matches": "^\\d{3,5}$"
+    ///     }
+    ///   ]
+    /// }
+    /// ```
```
Razz4780, line 420: I'd rename "type" -> "body" for clarity, since all other filter kinds are quite explicit. Like header filters have field "header", and so on.

```json
{
  "body": "json",
  "query": "...",
  "matches": "..."
}
```

mirrord/cli/src/port_forward.rs
```diff
@@ -972,11 +973,13 @@ impl IncomingMode {
         let http_filter_config = &config.http_filter;
 
         // Matching all fields to make this check future-proof.
+        // TODO(areg) unify this with mirrord_layer::setup::parse_http_filter
```
Razz4780, line 976: https://www.youtube.com/watch?v=K6qJcrVmomY

mirrord/cli/src/port_forward.rs
```diff
@@ -1006,6 +1011,26 @@ impl IncomingMode {
                 path_filter: None,
                 header_filter: None,
                 method_filter: None,
+                body_filter: Some(filter),
+                all_of: None,
+                any_of: None,
+                ports: _ports,
+            } => HttpFilter::Body(
+                // TODO(areg) unification
```
Razz4780, line 1019: https://www.youtube.com/watch?v=K6qJcrVmomY

mirrord/agent/env/src/envs.rs
```diff
@@ -83,3 +83,10 @@ pub const IDDLE_TTL: CheckedEnv<u64> = CheckedEnv::new("MIRRORD_AGENT_IDLE_TTL")
 /// Sets whether `Mirrord-Agent` headers are injected into HTTP
 /// responses that went through the agent.
 pub const INJECT_HEADERS: CheckedEnv<bool> = CheckedEnv::new("MIRRORD_AGENT_INJECT_HEADERS");
+
+/// Sets the max size for bodies buffered for body filters.
+pub const MAX_BODY_BUFFER_SIZE: CheckedEnv<u32> = CheckedEnv::new("MIRRORD_MAX_BODY_BUFFER_SIZE");
+
+/// Sets how long to wait to receive the entire body for body filters.
```
Razz4780:
```suggestion
/// Sets how long to wait to receive the entire body for body filters (in milliseconds).
```

mirrord/agent/src/http/filter.rs
```diff
@@ -42,14 +61,88 @@ impl TryFrom<&mirrord_protocol::tcp::HttpFilter> for HttpFilter {
...
+            mirrord_protocol::tcp::HttpBodyFilter::Json { query, matches } => Self::Json {
+                query: JsonPath::parse(query)?,
+                matches: Regex::new(matches)?,
+            },
+        })
+    }
+}
+
+#[serde_json_path::function(name = "typeof")]
+fn type_of(nodes: NodesType) -> ValueType {
```
Razz4780, line 90: Ah, so the `typeof` thing is like our custom extension to jsonpath?

mirrord/agent/src/http/filter.rs
```diff
@@ -42,14 +61,88 @@ impl TryFrom<&mirrord_protocol::tcp::HttpFilter> for HttpFilter {
...
+    };
+
+    tracing::error!(%first_type, "not undefined");
+
+    for v in iter {
+        tracing::error!(?v, "json match");
+        if Types::from(*v) != first_type {
+            return ValueType::Nothing;
+        }
+    }
```
Razz4780: This jsonpath comes from the user. If anything goes wrong when parsing it, a descriptive error should be returned back to the user. We should not log it on error (like, it's not an actual error of the agent)

mirrord/agent/src/steal/subscriptions.rs
```diff
@@ -118,7 +118,9 @@ impl PortSubscriptions {
             },
 
             Entry::Vacant(e) => {
-                self.handle.steal(port).await?;
+                self.handle
+                    .steal(port, filter.as_ref().is_some_and(|f| f.needs_body()))
```
Razz4780:
```suggestion
                    .steal(port, filter.as_ref().is_some_and(HttpFilter::needs_body))
```
Nit

mirrord/agent/src/incoming/connection/http_task.rs
```diff
@@ -69,7 +74,30 @@ where
...
+            for frame in buffered.iter() {
+                let frame = if let Some(data) = frame.data_ref() {
+                    Frame::data(data.clone())
+                } else if let Some(trailers) = frame.trailers_ref() {
+                    Frame::trailers(trailers.clone())
+                } else {
+                    panic!("{frame:?} was neither a data or a trailers frame");
+                };
+
+                self.destination.send_frame(frame).await?;
```
Razz4780: I think we can consume the `BufferedBody` here, right? We don't have to clone the frames. Cloning the data frame is cheap, but trailers is essentially a map of strings

mirrord/agent/src/incoming/connection/http.rs
```diff
@@ -24,6 +35,105 @@ use crate::{
...
+    Empty,
+
+    // Since we only store data frames and one optional trailer frame
+    // at the end, we could save a couple hundred bytes by storing raw
+    // `Bytes` objects instead of `Frame<Bytes>`-es, (32 vs 96 bytes
+    // on 64bit) and having an optional `first_nondata_frame` field.
+    // We go with this approach because it's a little simpler to
+    // implement.
+    Successful(Arc<Vec<Frame<Bytes>>>),
+    Failed(Arc<Vec<Frame<Bytes>>>),
```
Razz4780: Why do we keep the frames in case of failure?

Razz4780 (changes requested):
Not sure how I feel about putting this in the connection task:
1. It makes that module aware of the filters. Right now its responsibility is quite simple (redirect ports, decode HTTP traffic), and the code is already a bit complex.
2. It affects HTTP client -> original dst communication, even if we don't have any steal subscription.

What do you think about moving the body buffering to `mirror` and `steal` modules instead?

Like this:

1. Buffering
    * Add `RedirectedHttp::with_whole_body(self, size_limit: u32, timeout: Duration) -> impl Future<Output = <Option<Self>>` and `RedirectedHttp::body_finished(&self) -> bool`. On error, sends the response and returns `None`. On exceeded size/time limit, returns `Some`.
    *  Add `MirroredHttp::with_whole_body(size_limit: u32) -> impl Future<Output = <Option<Self>>`. On error or exceeded size limit, returns `None`.
3. Mirror (`TcpMirrorApi`)
    * Use the `MirrorHandle` via wrapper responsible for applying filters (including body buffering). If there's a body filter, the wrapper handles body buffering with `MirroredHttp::with_whole_body` and `JoinSet`.
5. Steal (`TcpStealerTask`)
    * At first, apply only non-body filters. If there's a certain match, send the request to the user. If we're certain there is no match, pass it through.
    * If we're not certain, use `RedirectedHttp::with_whole_body` and `JoinSet` to process the request again later.

mirrord/protocol/src/tcp.rs
Razz4780: We should only send the new variant if mirrord-protocol version allows it, and this should be checked right after making the first connection (like we do for other new filters)

mirrord/agent/src/http/filter.rs
MintSoup: Yup, allows us to filter by type right in the query without having to add additional fields to the filter.

mirrord/agent/src/http/filter.rs
MintSoup: oops these are (were) just leftovers from debugging

mirrord/agent/src/incoming/connection/http.rs
MintSoup: The buffering can "fail" (timeout/exceed max size) but we still want to send whatever we have to the client.

mirrord/protocol/src/tcp.rs
MintSoup: What do you think about replacing `HttpFilterConfig::is_composite`, `HttpFilterConfig::has_method_filter` and friends with a `HttpFilterConfig::matches_protocol_version(semver::Version) -> Result<(), ConfigError>`? Those methods are only used for checking protocol version and this would centralize that logic a bit.

MintSoup:
> Not sure how I feel about putting this in the connection task:
> 
> 1. It makes that module aware of the filters. Right now its responsibility is quite simple (redirect ports, decode HTTP traffic), and the code is already a bit complex.

Fair enough, although the extent of being aware of filters is "do we need to buffer body or not".

> 2. It affects HTTP client -> original dst communication, even if we don't have any steal subscription.

Yeah this is something I thought of, ideally we would buffer the body in parallel to streaming it to the original destination /any clients that don't have body filters.

> What do you think about moving the body buffering to `mirror` and `steal` modules instead?

Actually I did something very similar to this initially, the implementation was pretty simple for the steal flow but I got stuck when it came to mirroring, so I switched to this approach. How would `TcpMirrorApi` (1) tell the owner of the `RedirectedHttp` to start buffering (and do so in time, before it transmits and loses data) and (2) get access to the buffered body?

> Add `MirroredHttp::with_whole_body(size_limit: u32) -> impl Future<Output = <Option<Self>>
What would this do though? Buffer frames received from `Self::stream` until we have enough/error out? I steered away from this approach because it'd kinda be duplicating the buffering logic and storing multiple copies of the body (although now that I think of it, `Frame`s are refcounted anyway so the overhead would only be container storing them)

> Use the MirrorHandle via wrapper responsible for applying filters (including body buffering)
Not quite sure what this means?

mirrord/cli/src/port_forward.rs
MintSoup: Should I remove these?

mirrord/config/src/feature/network/incoming/http_filter.rs
MintSoup: What about the single filter case? Then it'd look like this:
```json
"body_filter": {
  "body": "json",
  "query": "$.library.books[*]",
  "matches": "^\\d{3,5}$"
 }
```
The repetition (`"body"`/`"body_filter"`) looks a bit odd IMHO.

mirrord/agent/src/incoming/connection/http.rs
Razz4780: Ah right, ok

mirrord/agent/src/http/filter.rs
Razz4780: If it's not part of the jsonpath RFC, then I'm not sure if we need it right now :thinking: 
Personally, I'd wait with this extension until there's demand from the users

mirrord/protocol/src/tcp.rs
Razz4780: No objection :+1:

mirrord/config/src/feature/network/incoming/http_filter.rs
Razz4780: Yeah, it's a tradeoff. But I think the nested filter needs some "body" mention.

mirrord/agent/src/http/filter.rs
```diff
@@ -42,14 +61,83 @@ impl TryFrom<&mirrord_protocol::tcp::HttpFilter> for HttpFilter {
...
+
+    first_type.into()
+}
+
 impl HttpFilter {
     /// Checks whether the given request [`Parts`] match this filter.
-    #[tracing::instrument(level = Level::DEBUG, skip(parts), ret)]
-    pub fn matches(&self, parts: &mut Parts) -> bool {
+    #[tracing::instrument(level = Level::DEBUG, skip_all, fields(has_body = body.is_some()), ret)]
+    pub fn matches<T: Read + Copy>(&self, parts: &mut Parts, body: Option<T>) -> bool {
```
Razz4780, line 140: Haven't read the rest of the PR yet, but feels like this function should be returning something like:

```
enum HttpFilterOutput {
    NoMatch,
    Match,
    NeedsBody,
}
```

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+            InternalHttpBodyFrame::Data(payload) => Some(payload),
+            InternalHttpBodyFrame::Trailers(_) => None,
+        }
+    }
+}
+
+/// Implements [`Read`] for a &[Frame]
+pub struct FramesReader<'a, T> {
+    remaining: &'a [T],
+    read_until: usize,
```
Razz4780, line 33: Docs pls

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl<T> Copy for FramesReader<'_, T> {}
+impl<T> Clone for FramesReader<'_, T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
```
Razz4780, line 41: What purpose does this serve? 👀 
I believe you're allowing for copying a partially consumed reader.

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl<T: Framelike> Read for FramesReader<'_, T> {
```
Razz4780, line 55: Can we have UTs for this?

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+        .unwrap_or(1000)
+        .into(),
+    )
+});
+
+#[derive(thiserror::Error, Debug)]
+pub enum BufferBodyError {
+    #[error("io error while receiving http body: {0}")]
```
Razz4780:
```suggestion
    #[error("hyper error while receiving http body: {0}")]
```

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+    Hyper(#[from] hyper::Error),
+    #[error(transparent)]
+    Conn(#[from] ConnError),
+    #[error("body size exceeded max configured size")]
```
Razz4780: This error message could mention the limit

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+    BodyTooBig,
+    #[error("receiving body took too long")]
```
Razz4780: This error message could mention the limit

mirrord/agent/src/incoming/connection/body_utils.rs
```diff
@@ -0,0 +1,188 @@
...
+impl<T> BufferedBody<T> {
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        matches!(self, Self::Empty)
+    }
+
+    /// Returns the buffered data if we have anything, regardless of
+    /// whether it's full or partial.
+    #[inline]
+    pub fn buffered_data(self) -> Option<Vec<T>> {
```
Razz4780: Could return an empty Vec in the Empty case (if that makes sense in the rest of the code).
Also, BufferedBody::Empty looks to me like BufferedBody::Full(vec![]). Maybe we could unify? I'm not sure what `BufferedBody::Empty` represents

mirrord/agent/src/incoming/connection/http.rs
```diff
@@ -41,6 +56,8 @@ pub struct RedirectedHttp {
 
     /// Configuration of the RedirectorTask that created this
     redirector_config: RedirectorTaskConfig,
+
+    buffered_body: BufferedBody<Frame<Bytes>>,
```
Razz4780: This struct already stores full body state, in `self.request.body_head` and `self.request.body_tail`. No need for this

mirrord/agent/src/incoming/connection/http.rs
```diff
@@ -300,25 +334,170 @@ impl ResponseBodyProvider {
...
             .field("info", &self.info)
             .field("request_head", &self.request_head)
-            .field("parts", &self.parts)
+            .field("buffered_body", &self.buffered_body)
             .finish()
     }
 }
+
+/// Used to implement `buffer_body` on both [`MirroredHttp`] and [`StolenHttp`]
+pub trait BodyBufferable: Debug {
```
Razz4780: This looks overly complex with these extra traits 👀 

I believe you only need this:

```rust
impl RedirectedHttp {
    /// If we're certain that we have read all of the request's body, returns it.
    ///
    /// Otherwise, returns [`None`].
    pub fn full_body(&self) -> Option<&[Frame<Bytes>]> {
        self.request
            .body_tail
            .is_none()
            .then_some(&self.request.body_head)
    }

    /// Attempts to read the full body of the request, updating the internal state of this struct.
    ///
    /// Broadcasts any frames/updates to [`MirroredHttp`] instances obtained with previous calls to
    /// [`Self::mirror`].
    ///
    /// Returns:
    /// * [`Ok`] if the full body was read, or the limits were exceeded
    /// * [`Err`] if reading the body failed
    pub async fn try_read_full_body(
        &mut self,
        size_limit: usize,
        time_limit: Duration,
    ) -> Result<(), ConnError> {
        let Some(tail) = self.request.body_tail.as_mut() else {
            return Ok(());
        };

        let mut current_size = self
            .request
            .body_head
            .iter()
            .filter_map(Frame::data_ref)
            .map(|data| data.len())
            .sum::<usize>();
        if current_size >= size_limit {
            return Ok(());
        }

        let mut mirror_tx = OptionalBroadcast::from(self.mirror_tx.clone());

        let result = tokio::time::timeout(time_limit, async {
            while current_size < size_limit {
                match tail.frame().await {
                    None => {
                        mirror_tx.send_item(IncomingStreamItem::NoMoreFrames);
                        return Ok(true);
                    }
                    Some(Ok(frame)) => {
                        current_size += frame.data_ref().map(|data| data.len()).unwrap_or_default();
                        mirror_tx.send_frame(&frame);
                        self.request.body_head.push(frame);
                    }
                    Some(Err(error)) => {
                        let error = ConnError::IncomingHttpError(Arc::new(error));
                        mirror_tx.send_item(IncomingStreamItem::Finished(Err(error.clone())));
                        return Err(error);
                    }
                }
            }

            Ok(false)
        })
        .await;

        match result {
            // We successfully read the whole body.
            Ok(Ok(true)) => {
                self.request.body_tail = None;
                Ok(())
            }
            // We hit the size limit when reading the body.
            Ok(Ok(false)) => Ok(()),
            // We failed to ready the body.
            Ok(Err(error)) => Err(error),
            // We hit the time limit when reading the body.
            Err(..) => Ok(()),
        }
    }
}

impl MirroredHttp {
    /// If we're certain that we have read all of the request's body, returns it.
    ///
    /// Otherwise, returns [`None`].
    pub fn full_body(&self) -> Option<&[InternalHttpBodyFrame]> {
        self.request_head
            .body_finished
            .then_some(&self.request_head.body_head)
    }

    /// Attempts to read the full body of the request, updating the internal state of this struct.
    ///
    /// Returns:
    /// * [`Ok`] if the full body was read, or the limit was exceeded
    /// * [`Err`] if reading the body failed
    pub async fn try_read_full_body(&mut self, size_limit: usize) -> Result<(), ConnError> {
        if self.request_head.body_finished {
            return Ok(());
        }

        let mut current_size = self
            .request_head
            .body_head
            .iter()
            .map(|frame| match frame {
                InternalHttpBodyFrame::Data(data) => data.len(),
                InternalHttpBodyFrame::Trailers(..) => 0,
            })
            .sum::<usize>();
        if current_size >= size_limit {
            return Ok(());
        }

        while current_size < size_limit {
            match self.stream.next().await {
                Some(IncomingStreamItem::Frame(frame)) => {
                    if let InternalHttpBodyFrame::Data(data) = &frame {
                        current_size += data.len();
                    }
                    self.request_head.body_head.push(frame);
                }
                Some(IncomingStreamItem::NoMoreFrames) => {
                    self.request_head.body_finished = true;
                    break;
                }
                Some(IncomingStreamItem::Finished(Err(error))) => {
                    return Err(error);
                }
                item @ None
                | item @ Some(
                    IncomingStreamItem::Data(..)
                    | IncomingStreamItem::NoMoreData
                    | IncomingStreamItem::Finished(Ok(())),
                ) => {
                    return Err(ConnError::AgentBug(format!(
                        "received an unexpected IncomingStreamItem when buffering HTTP request body: {item:?}"
                    )));
                }
            }
        }

        Ok(())
    }
}
```

mirrord/agent/src/steal/task.rs
```diff
@@ -184,31 +210,52 @@ impl TcpStealerTask {
             }
         };
 
+        if filters.values().any(HttpFilter::needs_body) {
```
Razz4780, line 213: First, I'd check if there's a filter that does not require body, and matches the request. If there is one, send it to that user.

mirrord/agent/src/steal/task.rs
```diff
@@ -294,6 +341,33 @@ impl TcpStealerTask {
         self.clients.remove(&client_id);
         self.subscriptions.remove_all(client_id);
     }
+
+    #[tracing::instrument(level = Level::TRACE, ret)]
+    async fn handle_buffered_http(&mut self, http: RedirectedHttp) {
```
Razz4780, line 346: I'd do it like this:
1. Make `handle_stolen_traffic` take `ongoing: Option<&mut JoinSet<RedirectedHttp>>`
2. If there is a filter that certainly matches the request, send it to that client
3. If we're not certain, and we have ongoing, spawn buffering task
4. If we're not certain, and we don't have ongoing, consider uncertain filters to be non-matching
5. If the traffic comes from `PortSubscriptions`, pass `Some(ongoing)` to `handle_stolen_traffic`
6. If the traffic comes from `ongoing`, pass `None` to `handle_stolen_traffic`

Razz4780 (changes requested):

mirrord/agent/src/steal/task.rs
MintSoup: I did try to do that, but we also send logs to preempted clients, and we have to preemptively buffer the body unless we're ok with preempted clients with body filters that would have stolen the request to not receive logs.

mirrord/agent/src/incoming/connection/http.rs
MintSoup: I did want to store the buffered chunks in `body_head` initially, but the problem is that later it gets sent to the client with the request start, all in one packet. Not sure how significant/semantically important this behavior is but I didn't want to mess with the protocol logic or potentially send lots of data in one chunk, so here we are. I do agree that it's pretty ugly though, if we could avoid sending all of `request_head` inside the request start packet that would be nice.

mirrord/agent/src/incoming/connection/body_utils.rs
MintSoup: `HttpFilter::matches` is recursive (for the composite filter case) so it needs to be at least `Clone`.

> allowing for copying a partially consumed reader

True, but I don't think there's a nice to prevent that at compile time without extra types.

mirrord/agent/src/steal/task.rs
MintSoup: Why though? Just seems more confusing to me because 
1. It's not immediately clear at first glance that `ongoing = Some(...)` means "this is a fresh request" and `ongoing = None` means "this a request with an already buffered body".
2. When the traffic comes from `ongoing`, the subscription lookup will have to move to the `run` loop (ugly D:) or a separate function that will call `handle_stolen_traffic` (kinda like we have now)
3. We can't short-circuit the body buffering when there's a matching non-body filter (i.e. step 2) (see other comment specifically about that), so we lose most of the simplicity benefits of this flow anyway :(

I guess the 3rd point is the most important one here, if we give up on logging we could do it this way.

mirrord/agent/src/incoming/connection/http.rs
Razz4780: Ah, don't worry about it. We use 64kb limit, right? This is comparable to max size of an outgoing mirrord-protocol message.

mirrord/agent/src/incoming/connection/body_utils.rs
Razz4780: Hmm, you could pass `Option<&[F]>` to `HttpFilter::matches`, and wrap it in `FramesReader` when needed.
But ok, no biggie. Can you add a comment to `HttpFilter::matches`, on the fact that we copy the reader for each body filter, so each time we read from the start?

mirrord/agent/src/incoming/connection/http.rs
Razz4780: Pushing to fields that already exist allows us to do https://github.com/metalbear-co/mirrord/pull/3712#discussion_r2539755987 - saves us a couple new types, including custom error type

mirrord/agent/src/steal/task.rs
Razz4780: Oke, sounds fair

mirrord/agent/src/steal/task.rs
Razz4780: Oke, sounds fair

mirrord/agent/src/http/filter.rs
MintSoup: We have a separate `needs_body` method for that, since we have to buffer the body before running any filters

mirrord/agent/src/incoming/connection/body_utils.rs
MintSoup: okok

mirrord/agent/src/incoming/connection/http.rs
```diff
@@ -43,6 +55,18 @@ pub struct RedirectedHttp {
     redirector_config: RedirectorTaskConfig,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum BufferBodyError {
+    #[error("hyper error while receiving http body: {0}")]
+    Hyper(#[from] hyper::Error),
```
Razz4780: This one is no longer used, I think

mirrord/agent/src/steal/task.rs
```diff
@@ -39,6 +40,8 @@ pub struct TcpStealerTask {
     clients: HashMap<ClientId, Client>,
     /// Futures that resolve when clients disconnect (drop their [`StealerMessage`] receivers).
     disconnected_clients: FuturesUnordered<ChannelClosedFuture>,
+    /// For tracking http requests whose bodes are being buffered
```
Razz4780:
```suggestion
    /// For tracking http requests whose biodes are being buffered
```

Razz4780: ...... bodies

PR #3714, opened 2025-11-12
INT-214 DB branching for Postgres with emtpy DB creation

vladrbg:


mirrord/cli/src/db_branches.rs
```diff
@@ -64,28 +76,119 @@ async fn status_command(args: &DbBranchesArgs, names: &[String]) -> CliResult<()
...
+        name: String,
+        db_type: &'static str,
+        phase: String,
+        ttl: u64,
+        database: String,
+        users: String,
+        expire_time: String,
+    }
+
+    let mut all_branches = Vec::new();
```
0x00A5: nit: we have the length of this `Vec` at this point (pg+mysql), should we `Vec::with_capacity()` instead?

mirrord/cli/src/db_branches.rs
```diff
@@ -64,28 +76,119 @@ async fn status_command(args: &DbBranchesArgs, names: &[String]) -> CliResult<()
...
+
+    // Process MySQL branches
+    for branch in mysql_branches {
+        if let Some(name) = &branch.metadata.name
+            && (names.is_empty() || names.contains(name))
+        {
+            all_branches.push(BranchInfo {
```
0x00A5: nit: maybe cleaner and easier to read if we move these type conversions to `impl From<&Branch> for BranchInfo` from the function.

mirrord/cli/src/db_branches.rs
```diff
@@ -64,28 +76,119 @@ async fn status_command(args: &DbBranchesArgs, names: &[String]) -> CliResult<()
...
-            })
-            .collect()
-    };
+                    .map(|s| s.phase.to_string())
+                    .unwrap_or_else(|| "Unknown".to_string()),
+                ttl: branch.spec.ttl_secs,
+                database: branch
+                    .spec
+                    .database_name
+                    .unwrap_or_else(|| "<none>".to_string()),
```
0x00A5: We have some duplicated `unwrap_or_else(|| "none")` or unknown here. Maybe we can make these fields `Option<T>` under branch info, and convert `None` to their fallbacks altogether when we finally do`add_row()`.

mirrord/cli/src/db_branches.rs
```diff
@@ -184,45 +272,88 @@ async fn destroy_command(args: &DbBranchesArgs, all: bool, names: &Vec<String>)
...
         let delete_params = DeleteParams::default();
 
         for name in names {
             let mut branch_progress = destroy_progress.subtask(&format!("destroying {name}"));
-            match api.get(name).await {
-                Ok(_) => match api.delete(name, &delete_params).await {
-                    Ok(_) => branch_progress.success(Some(&format!("destroyed branch: {name}"))),
+
+            // Check which type this branch is and delete from the correct API
+            if mysql_names.contains(name) {
```
0x00A5, line 320: Currently not possible (since we specify a resource name prefix that include the db type and let k8s add unique suffix), but if a name matches both mysql branch and pg branch, we should delete both, right?

mirrord/operator/src/client/database_branches.rs
```diff
@@ -163,8 +172,142 @@ pub(crate) async fn list_reusable_mysql_branches<P: Progress>(
     Ok(reusable_mysql_branches)
 }
 
+/// Create MySQL branch databases and wait for their readiness.
```
0x00A5: MySQL -> Postgres

mirrord/operator/src/client/database_branches.rs
```diff
@@ -163,8 +172,142 @@ pub(crate) async fn list_reusable_mysql_branches<P: Progress>(
...
+        .map_err(|_| OperatorApiError::OperationTimeout {
+            operation: OperatorOperation::PgBranching,
+        })?;
+    subtask.success(Some("new PostgreSQL branch databases ready"));
+
+    Ok(created_branches)
+}
+/// Given parameters of all MySQL branch databases needed for a session, list reusable ones.
+///
+/// A MySQL branch is considered reusable if
```
0x00A5: MySQL -> Postgres

mirrord/cli/src/db_branches.rs
vladrbg: Ah yep.. I did not consider they can be the same. Yep both should be deleted if they have same name.

mirrord/cli/src/db_branches.rs
vladrbg: Much better like that yep

0x00A5 (approved):

PR #3717, opened 2025-11-14
Remove kubernetes stuff from the devshell

iniw:
The kubernetes stuff should just be installed system-wide


meowjesty (approved):
👍 but I'm not a nix user, maybe you want input from someone that uses it?

Issue #3716, opened 2025-11-14
non_blocking_tcp_connect: true causes ~1 second delay on every connection


fedorkosilov:
### Bug Description

Since version 3.172.0 when `non_blocking_tcp_connect` was enabled by default, every TCP connection has ~1 second delay even though connection is established immediately.

Workaround:
Setting `non_blocking_tcp_connect: false` fixes the issue completely

### Steps to Reproduce

A Python app using httpx/httpcore making outgoing HTTP requests through mirrord 3.172.0+ (with default config) will show this 1s delay per connection.

### Backtrace

```shell

```

### mirrord layer logs

```shell
13:06:47.613 TRACE mirrord_layer::socket::ops: we are connected Connected {
    layer_address: Some(Ip(127.0.0.1:36349)),
    connection_id: Some(310737559328176939962308352904990333721),
}
[~1 second gap]
13:06:48.637 TRACE mirrord_layer::socket::ops: getsockname -> local_address
13:06:48.637 httpcore.connection - connect_tcp.complete
```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json
{
    "target": {
        "path": "deployment/my-service",
        "namespace": "my-namespace"
    },
    "feature": {
        "network": {
            "incoming": "steal",
            "outgoing": {
                "filter": {
                    "local": ["127.0.0.1", "localhost"]
                }
            }
        },
        "env": true
    }
}
```

### Your operating system and version

WSL2 (Windows Subsystem for Linux) - kernel 6.6.87.2-microsoft-standard-WSL2

### Local process

python3.12 running httpx/httpcore for HTTP requests

PR #3718, opened 2025-11-16
Update `non_blocking_tcp_connect` to default to false.

DmitryDodzin:
Set so `experimental.non_blocking_tcp_connect` defaults to `false` regardless if it's OS or mfT.

relevant for #3716.

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -1163,7 +1163,7 @@
         },
         "non_blocking_tcp_connect": {
           "title": "_experimental_ non_blocking_tcp_connect {#experimental-non_blocking_tcp_connect}",
-          "description": "Enables better support for outgoing connections using non-blocking TCP sockets.\n\nDefaults to `true` in OSS. Defaults to `false` in mfT.",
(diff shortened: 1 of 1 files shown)


mirrord/intproxy/src/lib.rs
```diff
@@ -160,7 +160,7 @@ impl IntProxy {
             Self::CHANNEL_SIZE,
         );
         let outgoing = background_tasks.register(
-            OutgoingProxy::new(experimental.non_blocking_tcp_connect.unwrap_or_default()),
```
aviramha, line 163: would be easier to not change schema and just change the default?

DmitryDodzin: It's not much of a scheme change it's still `Option<bool>` for `ExperimentalFileConfig` because of the `MirrordConfig` derive macro

DmitryDodzin: Ah there is a schema change, but it's the comment change that's the difference

aviramha: I just think we'll flip it back so think it'll be easier, but ok.

aviramha (approved):

PR #3727, opened 2025-11-18
config, cli: Allow overlap in incoming.ports and http_filter.ports

gememma:
Reverts some of https://github.com/metalbear-co/mirrord/pull/2486 to allow ports in `incoming.ports` and `incoming.http_filter.ports` to overlap. If the port is present in both, the port is filtered according to `http_filter` filters.

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -243,7 +243,12 @@
 #![cfg_attr(all(windows, feature = "windows_build"), feature(windows_change_time))]
 #![cfg_attr(all(windows, feature = "windows_build"), feature(windows_by_handle))]
 
-use std::{collections::HashMap, env::vars, net::SocketAddr, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    env::vars,
+    net::SocketAddr,
+    time::Duration,
+};
 #[cfg(not(target_os = "windows"))]
 use std::{ffi::CString, os::unix::ffi::OsStrExt};
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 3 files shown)


mirrord/cli/src/main.rs
```diff
@@ -683,21 +683,30 @@ pub(crate) fn print_config<P>(
...
-                    _ => Some(format!(
-                        "ports [{}] (unfiltered)",
-                        ports
-                            .iter()
-                            .copied()
-                            .map(|n| n.to_string())
-                            .collect::<Vec<String>>()
-                            .join(", ")
-                    )),
+                .and_then(|ports| {
```
meowjesty, line 695: Is this iter.filter being used to collect the "difference" between filtered and unfiltered ports?

Add some comments to what's going on here (I could swear that rust had a "difference" iter thingy, but I think it's only for hash collections).

deny.toml
```diff
@@ -14,6 +14,7 @@ ignore = [
     "RUSTSEC-2023-0071", # rsa crate, potential key recovery through timing sidechannels, though it sounds bad
     # we only use rsa as part of apple-codesign for re-sign binaries, so this should not matter.
     "RUSTSEC-2024-0436", # paste is unmaintained but we should not have an issue with it.
+    "RUSTSEC-2025-0119", # number_prefix is unmaintained, dependency of indicatif, used for progress bars in cli
```
meowjesty: Do we still need this? I updated the `indicatif` yesterday, see https://github.com/metalbear-co/mirrord/commit/cbb36f31537ee07b33ce3c17855ddf9e1ca325e9

mirrord/cli/src/main.rs
gememma: yeah it is - the previous printout depended on the lists being disjoint so I had to fix it 👀 
imagine you have filtered ports [1, 2], incoming ports [1, 2, 3, 4] - you want the printout to say
```text
filtered ports: 1, 2
unfiltered ports: 3, 4
```
will add a comment too :)

deny.toml
gememma: ugh the ci failed a couple of times, i probably managed to dodge your commit ☠️ annoyingggg

PR #3734, opened 2025-11-20
Fix DNS regression

0x00A5:


mirrord/layer-lib/src/socket/ops.rs
```diff
@@ -209,7 +209,15 @@ where
             // We need to fill the exact IP here.
             match &user_socket_info.state {
                 SocketState::Bound(Bound { address, .. }) => {
```
0x00A5, line 211: If the user bind their outgoing socket to 0.0.0.0:xxxx, address in `SocketState::Bound(Bound { address, .. })` is `0.0.0.0`. So is the layer address we get from int proxy. We then try to sendto this address, which fails always.

Issue #3477, opened 2025-08-05
Reverse port-forwarding with mirrord


drehelis:
Starting reverse port-forwarding with `mirrord port-forward --target deployment/http-server -R 8080:8080`.

curl https://ingress/mypage.html return:
```
mirrord-agent: failed to pass the request to its original destination: failed to make a passthrough TCP connection: Connection refused (os error 111)

Caused by:
      Connection refused (os error 111)
```

However, locally I can see incoming request hitting the server
```
python -m http.server -b 127.0.0.1 8080
Serving HTTP on 127.0.0.1 port 8080 (http://127.0.0.1:8080/) ...
127.0.0.1 - - [05/Aug/2025 09:26:56] "GET /mypage.html HTTP/1.1" 200 -
```

By default, mirroring traffic is default so the remote deployment must also listen on the same port, which in my case was just a dummy `sleep infinity` command.

- Improved UX can be printing message that explains this part
- add `--steal` flag to the port forward command line

Other than that, it’s a standout feature. really appreciate it!

niladrix719:
Hey! I’d like to work on this

PR #3739, opened 2025-11-22
feat: Added --steal flag to reverse port forwarding and port-forward warnings

niladrix719:
### Description
- Added `--steal` flag to reverse port forwarding
- Added port-forward warnings for UX

Fixes: #3477

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -917,7 +917,8 @@ async fn port_forward(
     result?;
 
     if args.steal && args.reverse_port_mapping.is_empty() {
-        progress.warning("`--steal` has no effect without reverse port mappings; ignoring the flag.");
+        progress
+            .warning("`--steal` has no effect without reverse port mappings; ignoring the flag.");
     }
 
     if !args.reverse_port_mapping.is_empty()


mirrord/cli/src/config.rs
```diff
@@ -614,6 +614,13 @@ pub(super) struct PortForwardArgs {
     /// Can be used multiple times.
     #[arg(short = 'R', long)]
     pub reverse_port_mapping: Vec<PortOnlyMapping>,
+
+    /// Steal remote traffic instead of mirroring it when using reverse port forwarding.
+    ///
+    /// When enabled, requests are handled exclusively by the local machine and are not
+    /// forwarded back to the remote target.
+    #[arg(long = "steal")]
+    pub steal: bool,
```
iniw, line 623: There's no need to make this an extra flag, the user should just set `feature.network.incoming.mode` in the config file

mirrord/cli/src/main.rs
```diff
@@ -898,6 +898,10 @@ async fn port_forward(
     let mut config = LayerConfig::resolve(&mut cfg_context)?;
     crate::profile::apply_profile_if_configured(&mut config, &progress).await?;
 
+    if args.steal && !args.reverse_port_mapping.is_empty() {
+        config.feature.network.incoming.mode = IncomingMode::Steal;
+    }
+
```
iniw, line 904: This can then be removed

mirrord/cli/src/main.rs
```diff
@@ -912,6 +916,19 @@ async fn port_forward(
     }
     result?;
 
+    if args.steal && args.reverse_port_mapping.is_empty() {
+        progress
+            .warning("`--steal` has no effect without reverse port mappings; ignoring the flag.");
+    }
```
iniw, line 922: Then we check the config instead:
```suggestion
    if config.feature.network.incoming.mode.is_steal() && args.reverse_port_mapping.is_empty() {
        progress
            .warning("Traffic stealing has no effect without reverse port mappings");
    }
```

mirrord/cli/src/main.rs
```diff
@@ -1185,4 +1202,19 @@ mod tests {
...
+            Commands::PortForward(params) => assert!(params.steal),
+            other => panic!("unexpected args parsed: {other:?}"),
+        }
+
+        let args = ["mirrord", "port-forward", "-R", "8080:8080"];
+        match Cli::parse_from(args).commands {
+            Commands::PortForward(params) => assert!(!params.steal),
+            other => panic!("unexpected args parsed: {other:?}"),
+        }
+    }
```
iniw, line 1219: Now that we don't have the new argument this test will just validate that the config is being parsed correctly, which we already have tests for, so let's just remove it.

mirrord/cli/src/config.rs
aviramha: I disagree - users complained that they need a mirrord.json just to specify steal so I think making it an argument is a good change :)

aviramha:
hi @niladrix719 we're awaiting your changes to merge - please let us know if we should wait or close this.

PR #3747, opened 2025-11-24
COR-1077: Support dlopen cgo library

0x00A5:
Add a dlopen hook on Linux. The hook replace symbols found in the cgo library module with our detours based on the golang version found in the module.

### DIFF
--- a/mirrord/layer/src/go/mod.rs
+++ b/mirrord/layer/src/go/mod.rs
@@ -235,7 +235,7 @@ fn get_go_runtime_version(hook_manager: &mut HookManager) -> Option<f32> {
     get_version_from_symbol(version_symbol)
 }
 
-/// Extracts version of the Go rutime from all modules.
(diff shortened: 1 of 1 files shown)


mirrord/layer/src/go/mod.rs
```diff
@@ -231,14 +232,28 @@ unsafe extern "C" fn raw_vfork_handler(
 /// Extracts version of the Go runtime in the current process.
 fn get_go_runtime_version(hook_manager: &mut HookManager) -> Option<f32> {
     let version_symbol = hook_manager.resolve_symbol_main_module("runtime.buildVersion.str")?;
+    get_version_from_symbol(version_symbol)
+}
+
+/// Extracts version of the Go rutime from all modules.
```
Razz4780:
```suggestion
/// Extracts version of the Go runtime from the given module.
```

PR #3748, opened 2025-11-25
Small v less tag windows

pavelzag:
Smol v-less tagging for windows builds


mirrord/cli/src/ci.rs
```diff
@@ -313,14 +314,18 @@ impl MirrordCi {
     #[cfg(target_os = "windows")]
```
cristeigabriela, line 314: this will be merged in with a future pr, but i would suggest adding #[cfg_attr(windows, allow(unused)] and then removing the changes between line 317 to 322 inclusively

PR #3753, opened 2025-11-26
Releases tests workflow fix

pavelzag:
Smol fix for Releases monitor slack messages format

### DIFF
--- a/changelog.d/+releases-tests-fix.internal.md
+++ b/changelog.d/+releases-tests-fix.internal.md
@@ -1,1 +1,0 @@
-This is a fix for release tests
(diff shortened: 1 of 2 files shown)


changelog.d/+releases-tests-fix.internal.md
```diff
@@ -0,0 +1 @@
+This is a fix for release tests
```
t4lz: This will land in the CHANGELOG.md file. This sentence doesn't fit as a changelog item.
But since there was no release since the first PR, this PR doesn't introduce a new change for the next version, and therefore doesn't necessitate a changelog entry.

So I'd recommend you delete the new changelog entry, and instead just make some small change to the changelog entry file you added in the first PR - just to pass the CI towncrier check.

t4lz (approved):
Just fix the changelog files please

PR #3764, opened 2025-11-28
Independent send handles for intproxy tasks, with working reconnect

MintSoup:
Quick recap of changes:
- Intproxy tasks now all no longer send agent-bound messages to the main task. Each background intproxy task has a handle to a queue of unsent messages. At all times, background tasks picks a random queue, pops a message and sends it over to the agent. Messages in a single queue are processed sequentially.
- Reconnect logic has been slightly reworked. Now, when a new connection to the agent is established, background tasks receive new send handles to the message queues, which they also propagate to their children. 
- When a reconnect starts, all ongoing connections and fds are immediately dropped.
- During reconnects, all agentbound messages generated by background tasks are discarded.
- Once a reconnect takes place, relevant bg tasks (notably filesproxy and incomingproxy) restore their state (port subscriptions, etc) *after* a protocol version negotiation has been completed with the new agent.

### DIFF
--- a/mirrord/protocol-io/src/lib.rs
+++ b/mirrord/protocol-io/src/lib.rs
@@ -175,7 +175,6 @@ impl<Type: ProtocolEndpoint> Connection<Type> {
     }
 
     /// Create a dummy instance, mainly used for tests.
-    #[cfg(test)]
     pub fn dummy() -> (Self, mpsc::Sender<Type::InMsg>, ConnectionOutput<Type>) {
         let (inbound_tx, inbound_rx) = mpsc::channel(32);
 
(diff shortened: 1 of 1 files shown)


mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,649 @@
...
+        ));
+
+        Ok(Self {
+            rx: inbound_rx,
+            shared_state,
+        })
+    }
+
+    /// Create a dummy instance, mainly used for tests.
+    pub fn dummy() -> (Self, mpsc::Sender<Type::InMsg>, ConnectionOutput<Type>) {
```
0x00A5, line 178: nit: gate this behind cfg test?

mirrord/protocol-io/src/lib.rs
```diff
@@ -0,0 +1,649 @@
...
+
+impl<Type: ProtocolEndpoint> Drop for Connection<Type> {
+    fn drop(&mut self) {
+        self.shared_state.cancel.cancel();
+    }
+}
+
+/// The "other end" of a `Connection`. Used for pulling out messages
+/// sent through it. Used only for testing.
+pub struct ConnectionOutput<Type: ProtocolEndpoint>(Arc<SharedState<Type>>);
```
0x00A5, line 240: nit: similar here, gate it with test?

0x00A5 (approved):

mirrord/protocol-io/src/lib.rs
MintSoup: Can't, because they're used for tests in other crates (i.e. mirrord-intproxy) and the test flag isn't set for other crates when compiling/linting.

mirrord/protocol-io/src/lib.rs
MintSoup: https://github.com/metalbear-co/mirrord/pull/3764#discussion_r2647175880

PR #3765, opened 2025-11-28
Support node's `fs.copyFile`

iniw:
Node's `fs.copyFile` implementation uses `libuv`'s [`uv_fs_copyfile`](https://github.com/libuv/libuv/blob/8e024629fe76c7e4d009fd01ccf8b329c99433d3/src/unix/fs.c#L1229) under the hood, which calls several functions we didn't have hooked:

- `sendfile`
- `ftruncate`
- `futimens`
- `fchown`
- `fchmod`

Implementing these makes it usable! Yay!

Here's a sample program for you to test :)
```js
const fs = require("fs");
const path = require("path");

const SRC = "/app/file";
const DEST = "/app/file.copy";

fs.copyFile(SRC, DEST, (err) => {
  if (err) {
    console.error("Copy failed:", err.message);
  } else {
    console.log(`Copied ${SRC} -> ${DEST}`);
  }
});
```

~Note that unlike other currently implemented libc functions, `sendfile` works with two fds simultaneously, since it acts like an "atomic" `read`+`write`, so the user could invoke it in a way where we try copying a file from the remote to the local machine or vice-versa in a single operation. This is implicitly unsupported and will error. Implementing this in a way that is transparent to the user would add significant complexity so I have left it out of this PR.~

Cross-machine `sendfile`ing now works by implementing it directly in the layer as a combination of `libc::read` + `libc::write` :)

# TODO

- [x] Protocol version bump 
- [x] Add tests
  - Only on MacOS for now, for some reason the Linux CI machine hates that test

### DIFF
--- a/changelog.d/3741.added.md
+++ b/changelog.d/3741.added.md
@@ -1,0 +1,1 @@
+Added support for node.js' [`fs.copyFile`](https://nodejs.org/api/fs.html#fspromisescopyfilesrc-dest-mode) function.


CONTRIBUTING.md
```diff
@@ -556,62 +556,36 @@ In order to have a more structured approach, here's the flow you should follow w
...
-        "aarch64-apple-darwin",
-        "x86_64-apple-darwin",
-        "x86_64-unknown-linux-gnu",
-        "aarch64-unknown-linux-gnu"
-    ]
-}
 ```
 
-You can use `cargo-zigbuild` to run `cargo check` or `clippy` on the agent's code on macOS.
+2. Get a C compiler capable of targetting the `x86_64-unknown-linux-gnu` triplet
```
gememma:
```suggestion
2. Get a C compiler capable of targeting the `x86_64-unknown-linux-gnu` triplet
```

CONTRIBUTING.md
```diff
@@ -556,62 +556,36 @@ In order to have a more structured approach, here's the flow you should follow w
 
 # Compiling on MacOS
 
-The `mirrord-agent` crate makes use of the `#[cfg(target_os = "linux")]` attribute to allow the whole repo to compile on MacOS when you run `cargo build`.
+The `mirrord-agent` feature-gates dependencies, modules and it's entrypoint to allow compilation on MacOS, but produces a binary that does nothing.
```
gememma:
```suggestion
The `mirrord-agent` feature-gates dependencies, modules and its entrypoint to allow compilation on MacOS, but produces a binary that does nothing.
```

mirrord/layer/src/file/hooks.rs
```diff
@@ -181,6 +181,154 @@ pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> u
...
+    len: *mut off_t,
+    _hdtr: *const libc::sf_hdtr,
+    _flags: c_int,
+) -> c_int {
+    unsafe {
+        let Some(count) = len.as_mut() else {
+            return -1;
+        };
+
+        sendfile_impl(fd, s, Some(offset), *count as usize)
```
0x00A5, line 240: I assume this is a decision made to just to read and write on the fds and ignore the logic about whether `s` is a socket or not.

mirrord/layer/src/file/hooks.rs
```diff
@@ -181,6 +181,154 @@ pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> u
...
+        };
+        futimens(fd, times)
+            .map(|()| 0)
+            .unwrap_or_bypass_with(|_| FN_FUTIMENS(fd, raw_times))
+    }
+}
+
+/// Hook for [`libc::fchown`].
+#[hook_guard_fn]
+pub(super) unsafe extern "C" fn fchown_detour(fd: c_int, owner: uid_t, group: gid_t) -> c_int {
```
0x00A5, line 314: how is the `fchown` function used by node's `copyFile()`? I assume it copies the src file's owner and set them for the dst file?

mirrord/layer/src/file/hooks.rs
```diff
@@ -181,6 +181,154 @@ pub(super) unsafe extern "C" fn opendir_detour(raw_filename: *const c_char) -> u
...
+        return None;
+    }
+
+    Some((written, offset.unwrap_or(0) + bytes_read as off_t))
+}
+
+/// Hook for macos's [`libc::sendfile`].
+#[cfg(target_os = "macos")]
+#[hook_fn]
+pub(super) unsafe extern "C" fn sendfile_macos_detour(
```
0x00A5, line 227: nit: the function doesn't need to be named differently right? Each target sees their own version of `sendfile_detour`.

mirrord/layer/src/file/ops.rs
```diff
@@ -850,6 +851,37 @@ pub(crate) fn rename(old_path: Detour<PathBuf>, new_path: Detour<PathBuf>) -> De
...
+}
+
+pub(crate) fn futimens(fd: RawFd, times: Option<[Timespec; 2]>) -> Detour<()> {
+    let fd = get_remote_fd(fd)?;
+    Detour::Success(common::make_proxy_request_with_response(
+        FutimensRequest { fd, times },
+    )??)
+}
+
+pub(crate) fn fchown(fd: RawFd, owner: u32, group: u32) -> Detour<()> {
```
0x00A5, line 868: The only thing I am not sure about is that the source of owner and group ids may not be acquired by calling stat/fstat functions. And we end up setting local owner/group ids on the remote file. It is unlikely to succeed so we fallback to bypass and call it locally. But if it succeeds, :monkey-lookaway:. Is this a valid concern? Correct me if I am wrong please.

mirrord/layer/src/file/hooks.rs
iniw: Yup and also because this handles cross-machine file copying for us - if we implemented `sendfile` "normally" (as a file operation on the agent) then the reading and writing would happen on the kernel, which doesn't go through our hooks, so the remote fd would not actually contain any data (because the fd is managed by us) and, consequently, the read/write would be meaningless

mirrord/layer/src/file/hooks.rs
iniw: Yup https://github.com/libuv/libuv/blob/8e024629fe76c7e4d009fd01ccf8b329c99433d3/src/unix/fs.c#L1334-L1340

mirrord/layer/src/file/hooks.rs
iniw: Already fixed this :)

mirrord/layer/src/file/ops.rs
iniw: Not sure what you mean, they are acquired from `stat` (see the other reply where I linked the implementation)

0x00A5: Discussed on Slack. The correct behaviours are 1) if fd is local, bypass 2) if fd is remote, owner and group need to be remote too. We decided to leave this for now, as for number 2, the remote operation actually succeed with local group and owner IDs should be quite rare.

PR #3766, opened 2025-12-01
win-56 ꫂ᭪  only support on-disk paths (fix agent crash);

cristeigabriela:
thanks to daniel's investigation, we figured out that agent seems to close connection when a OpenFileRequest fails `strip_prefix('/')`. as a result, this implementation of `path_to_unix_path` has been changed to guarantee that

- all paths must have initially pointed to the disk
- most importantly, begin with `/`, and so are *never* empty (""/no components) or fail the strip_prefix check

### DIFF
--- a/mirrord/layer-win/src/hooks/files/util.rs
+++ b/mirrord/layer-win/src/hooks/files/util.rs
@@ -151,32 +151,26 @@ impl PartialEq<u64> for WindowsTime {
 ///
 /// * `path` - The path to check.
 pub fn is_nt_path_disk_path<T: AsRef<str>>(path: T) -> bool {
-    let mut path = String::from(path.as_ref());
+    let mut path = path.as_ref();
 
     if !path.starts_with(GLOBAL_NAMESPACE_PATH) {
         return false;
     }
 
-    // Remove global namespace path.
-    path = path
-        .strip_prefix(GLOBAL_NAMESPACE_PATH)
-        .unwrap()
-        .to_string();
+    // Remove global namespace path to normalize.
+    path = &path[GLOBAL_NAMESPACE_PATH.len()..];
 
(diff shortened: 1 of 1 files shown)


mirrord/str-win/src/lib.rs
```diff
@@ -295,8 +303,24 @@ pub fn path_to_unix_path<T: AsRef<Path>>(path: T) -> Option<String> {
...
     }
 
-    // Skip root dir
-    let new_path: PathBuf = path.components().skip(1).collect();
+    let mut new_path: PathBuf = path.components().collect();
+
+    let maybe_root_dir = new_path.components().next()?;
+
+    // NOTE(gabriela): WIN-56 make sure path access is to disk
+    let maybe_root_dir = maybe_root_dir.as_os_str().to_string_lossy();
```
itsamegraf: this fix should not be part of this function - this is only a translation utility function of one path form to another.
We need to make sure `nt_create_file_hook` exits *explicitly* early if the path given is not an on-disk one. your current approach is implicit - "in the problematic cases, we will fail translation implictly and get a generic translation error for what needs to be unsupported path which should explicitly be handled by windows".
Additionally windows drives are not bound to A-Z, I cant find evidence online but I have seen "AA:" drive in the past, additionaly see [this](https://www.ryanliptak.com/blog/windows-drive-letters-are-not-limited-to-a-z/) recent blog post.

itsamegraf (changes requested):
good job, left a comment that needs to be addressed

mirrord/str-win/src/lib.rs
cristeigabriela: fixed! ty for the reference, it helped come up with a better solution! ^^

itsamegraf (approved):

PR #3767, opened 2025-12-01
win-57 ☆ fix getprocaddress_detour crash, implement module utilities

cristeigabriela:
GetProcAddress supports ordinals as the second argument. Ordinals are defined in the PE format to use the LOWORD. GetProcAddress second argument lpProcName is an ordinal if all bits besides LOWORD bits are 0.

Trying to turn this into a string through str-win naturally causes a crash, breaking C# applications which use ordinal imports, one example:
```
  2025-12-01T07:36:31.293810Z TRACE mirrord_layer_win::hooks::process: GetProcAddress: module=Some("mscoreei.dll") ptr=0x7ffc5c620000 ordinal='24' address=0x7ffc5c65e2d0
    at mirrord\layer-win\src\hooks\process\mod.rs:217 on ThreadId(2)
```

PR also implements some utilities for working with modules (get module base address from pointer within said module, get module name from base address)

**!** PR includes changes from #3766 as it was necessary for this particular repro. Branch is based on it's branch.

### DIFF
--- a/mirrord/layer-win/src/process/mod.rs
+++ b/mirrord/layer-win/src/process/mod.rs
@@ -77,13 +77,8 @@ pub fn get_module_name(module: *mut c_void) -> Option<String> {
 
     let mut name = [0u16; 512];
 
-    let ret = unsafe {
-        GetModuleFileNameW(
-            module as _,
-            name.as_mut_ptr() as _,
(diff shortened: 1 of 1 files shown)


itsamegraf (commented):
Thank you for finding the bug and getting this fixed.
WIN-57 portion is approved, will formally approve once #3766 is approved (this PR relies on it) and when last commit will have green CI

PR #3771, opened 2025-12-02
COR-1098 Agent security context

0x00A5:


mirrord/kube/src/api/container/pod.rs
```diff
@@ -125,7 +125,6 @@ impl ContainerVariant for PodVariant<'_> {
                 image_pull_secrets,
                 tolerations: agent.tolerations.clone(),
                 node_selector: Some(node_selector),
-                service_account_name: agent.service_account.clone(),
```
0x00A5, line 128: Why did I remove this!? Adding it back.

PR #3772, opened 2025-12-02
Get big boxes for large variants

Razz4780:
Passing/returning 1kb+ structs is not the wisest idea

### DIFF
--- a/mirrord/sip/src/error.rs
+++ b/mirrord/sip/src/error.rs
@@ -42,5 +42,15 @@ pub enum SipError {
     BinaryMoveFailed(std::io::Error),
 
     #[error("Code sign error: `{0}`")]
-    CodeSignError(#[from] apple_codesign::AppleCodesignError),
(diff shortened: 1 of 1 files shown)


mirrord/intproxy/src/agent_conn.rs
```diff
@@ -66,7 +66,10 @@ pub enum AgentConnectInfo {
         tls_pem: Option<PathBuf>,
     },
     /// Connect to the agent through the operator.
-    Operator(OperatorSession),
+    Operator(
+        /// Boxed due to large size.
+        Box<OperatorSession>,
```
meowjesty, line 71: If only rust supported emojis as identifiers: https://internals.rust-lang.org/t/supporting-emoji-in-identifiers/16838

```
📦 <OperatorSession>
```

meowjesty (approved):
📦 🪞

PR #3773, opened 2025-12-02
Throttle incoming traffic in the agent (outgoing connections)

Razz4780:


mirrord/agent/src/outgoing/throttle.rs
```diff
@@ -0,0 +1,74 @@
...
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+use bytes::Bytes;
+use futures::{Stream, StreamExt};
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use tokio_util::sync::PollSemaphore;
+
+/// Wrapper over a [`Stream`] of [`Bytes`], that yields items only after only acquiring
```
0x00A5:
```suggestion
/// Wrapper over a [`Stream`] of [`Bytes`], that yields items only after acquiring
```

PR #3776, opened 2025-12-03
WIN-55 Windows E2E tests in CI (+ WIN-52 WIN-60)

itsamegraf:

Please make sure you add a changelog file in `changelog.d/` named `<identifier>.<category>.md`

**Examples:**
- `1054.changed.md` (with GitHub issue)
- `+some-name.added.md` (without issue)

**Identifier:**
- If a GitHub issue exists, use the issue number from the public [mirrord repo](https://github.com/metalbear-co/mirrord)
- Use `+some-name` if no issue exists
- Don't use Linear issues or private repo issue numbers

**Category:**
Check `towncrier.toml` for available categories (`added`, `changed`, `fixed`, etc.) and choose the one that fits your change.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -391,6 +391,7 @@ jobs:
       - run: |
           cd mirrord/layer/tests/apps/rebind0
           cargo build
+      - run: ./mirrord/layer/tests/apps/dlopen_cgo/build_test_app.sh
       - run: ./scripts/build_c_apps.sh
       - run: cargo test --target x86_64-unknown-linux-gnu -p mirrord-layer
       - name: mirrord protocol UT
@@ -542,7 +543,7 @@ jobs:
       - run: python3 -m pip install flask fastapi uvicorn[standard] # For http mirroring test with Flask.
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 24
       - run: npm install express # For http mirroring test with node.
       - run: npm install portfinder # For a node specific test
       - name: mirrord layer tests
--- a/.github/workflows/releases-test.yaml
+++ b/.github/workflows/releases-test.yaml
@@ -1,0 +1,535 @@
+name: Release Monitor
+
+on:
+  schedule:
+    # Every 5 minutes (for testing)
+    - cron: "*/5 * * * *"
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  # ============================================================
+  # 1) Check latest GitHub Release has all required assets
+  # ============================================================
+  check_latest_release:
+    runs-on: ubuntu-24.04
+    outputs:
+      tag: ${{ steps.check.outputs.tag }}
+      html_url: ${{ steps.check.outputs.html_url }}
+    steps:
+      - name: Check latest GitHub Release assets
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+
+            // === REQUIRED ASSETS ===
+            // These must exist on every production release.
+            const requiredAssets = [
+              // Linux x86_64
+              "mirrord_linux_x86_64.zip",
+              "mirrord_linux_x86_64.shasum256",
+              "mirrord_linux_x86_64",
+              "libmirrord_layer_linux_x86_64.so",
+
+              // Linux aarch64
+              "mirrord_linux_aarch64.zip",
+              "mirrord_linux_aarch64.shasum256",
+              "mirrord_linux_aarch64",
+              "libmirrord_layer_linux_aarch64.so",
+
+              // macOS universal
+              "mirrord_mac_universal.zip",
+              "mirrord_mac_universal.shasum256",
+              "mirrord_mac_universal",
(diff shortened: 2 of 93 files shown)


mirrord/layer-lib/src/socket/hostname.rs
```diff
@@ -96,7 +96,7 @@ pub trait HostnameResolver {
     fn is_enabled(check_enabled: bool) -> bool {
         // Check if hostname feature is enabled
         #[cfg(target_os = "windows")]
-        let hostname_enabled = layer_setup().local_hostname();
+        let hostname_enabled = !layer_setup().local_hostname();
```
itsamegraf, line 99: this was a bug that prevent `traffic::traffic_tests::gethostname_remote_result` from passing, not sure how it passed in the first place

mirrord/layer/tests/apps/fileops/go/main.go
```diff
@@ -1,3 +1,5 @@
+//go:build !windows
```
itsamegraf, line 1: restricted to !windows as it uses syscall.Statfs which is only available on unix

mirrord/layer/tests/apps/faccessat_go/main.go
```diff
@@ -1,3 +1,5 @@
+//go:build !windows
```
itsamegraf, line 1: restricted to !windows as it uses syscall.Access which is only available on unix

tests/src/traffic.rs
```diff
@@ -217,6 +217,7 @@ mod traffic_tests {
     #[cfg_attr(not(feature = "job"), ignore)]
     #[rstest]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[timeout(Duration::from_secs(60))]
```
itsamegraf, line 220: if there are issues with the tests i have seen it hanging forever, this should resolve it

tests/src/env.rs
```diff
@@ -12,7 +12,7 @@ mod env_tests {
         services::basic_service,
     };
 
-    #[cfg_attr(not(feature = "job"), ignore)]
+    #[cfg_attr(any(not(feature = "job"), target_os = "windows"), ignore)]
```
itsamegraf, line 15: this is ignored on windows to prevent it from attempting to run bash (which will run in wsl), there are too many issues with wsl to put it in CI

scripts/win_install_test_app_frameworks.ps1
```diff
@@ -1,21 +0,0 @@
-@REM install choco if necessary
```
itsamegraf: removed this file as it seems like its scripts/win_build_test_apps.ps1 but older

tests/src/targetless.rs
```diff
@@ -125,7 +125,7 @@ mod targetless_tests {
         drop(test_process);
     }
 
-    #[cfg_attr(not(feature = "targetless"), ignore)]
+    #[cfg_attr(any(not(feature = "targetless"), target_os = "windows"), ignore)]
```
itsamegraf, line 128: Application::IntproxyChild of this test is a c application that uses `wait` - a POSIX only function. ignoring this test

.gitignore
```diff
@@ -2,7 +2,7 @@ out
 dist
 node_modules
 package-lock.json
-package.json
+/package.json
```
itsamegraf, line 5: this was added to ensure our node-e2e/package.json file is being tracked by git.
coordinated with Michal who wrote this line originally
Image

mirrord/layer-lib/src/error/windows.rs
```diff
@@ -89,25 +89,38 @@ impl WindowsError {
         };
         Self::format_error_code(err)
     }
+
+    /// Returns the underlying error code formatted as a hexadecimal string.
+    pub fn hex_code(&self) -> String {
```
cristeigabriela, line 94: as long as it's a pub, `get_hex_string` instead perhaps?

mirrord/layer-win/src/hooks/files/mod.rs
```diff
@@ -296,6 +297,16 @@ unsafe extern "system" fn nt_create_file_hook(
             return run_original();
         }
 
+        // Some Windows APIs try to open the drive root (e.g. `C:\`) to inspect metadata. We don't
+        // support mirroring root directory handles, so let Windows handle these locally.
+        if is_nt_path_disk_root(&name) {
```
cristeigabriela, line 302: i think this can be replaced by checking if path_to_unix_path is `/` for now

it will always *exclusively* be a disk root, unless i'm somehow mistaken, because of `is_nt_path_disk_path`

mirrord/layer-lib/src/error/windows.rs
itsamegraf: np, ill do it in my next pr 🙏

mirrord/layer-win/src/hooks/files/mod.rs
itsamegraf: this method has a bit more tests but overall I agree with the approach, will be as part of my next PR 🙏

PR #3784, opened 2025-12-08
COR-1097: Outgoing network policy

iniw:
- Related PRs:
    - https://github.com/metalbear-co/operator/pull/1158
    - https://github.com/metalbear-co/charts/pull/301

### DIFF
--- a/mirrord/agent/src/entrypoint.rs
+++ b/mirrord/agent/src/entrypoint.rs
@@ -20,7 +20,8 @@ use mirrord_agent_iptables::{
     error::{IPTablesError, IPTablesResult},
 };
 use mirrord_protocol::{
-    ClientMessage, DaemonMessage, GetEnvVarsRequest, LogMessage, dns::ReverseDnsLookupResponse,
+    ClientMessage, DaemonMessage, GetEnvVarsRequest, LogMessage, ResponseError,
+    dns::ReverseDnsLookupResponse,
 };
 use tokio::{
     net::{TcpListener, TcpSocket, TcpStream},
@@ -549,19 +550,18 @@ impl ClientConnectionHandler {
                     .await?;
             }
             ClientMessage::ReverseDnsLookup(request) => {
-                let hostname = match dns_lookup::lookup_addr(&request.ip_address) {
-                    Ok(hostname) => Some(hostname),
-                    Err(error) => {
-                        debug!(
-                            "Reverse DNS lookup failed for {}: {}",
-                            request.ip_address, error
-                        );
-                        None
-                    }
-                };
+                let hostname = dns_lookup::lookup_addr(&request.ip_address).inspect_err(|error| {
+                    error!(
+                        address = ?request.ip_address,
+                        error = ?error,
+                        "Reverse DNS lookup failed",
+                    )
+                });
 
                 self.respond(DaemonMessage::ReverseDnsLookup(Ok(
-                    ReverseDnsLookupResponse { hostname },
+                    ReverseDnsLookupResponse {
+                        hostname: hostname.map_err(ResponseError::from),
+                    },
                 )))
                 .await?
             }
--- a/mirrord/protocol/src/dns.rs
+++ b/mirrord/protocol/src/dns.rs
@@ -136,6 +136,7 @@ pub struct ReverseDnsLookupRequest {
 /// Response from reverse DNS lookup.
 #[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
 pub struct ReverseDnsLookupResponse {
-    /// The hostname associated with the IP address, if lookup succeeded.
-    pub hostname: Option<String>,
+    /// The hostname associated with the IP address,
+    /// Contains an error if the lookup failed.
+    pub hostname: RemoteResult<String>,
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2330,6 +2330,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "dns-lookup"
+version = "3.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e39034cee21a2f5bbb66ba0e3689819c4bb5d00382a282006e802a7ffa6c41d"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "socket2 0.6.1",
+ "windows-sys 0.60.2",
+]
+
+[[package]]
 name = "dns_resolve"
 version = "0.1.0"
 
@@ -4549,6 +4561,7 @@ dependencies = [
  "bytes",
  "clap",
  "containerd-client",
+ "dns-lookup",
  "enum_dispatch",
  "faccess",
  "fancy-regex",
@@ -4969,7 +4982,7 @@ dependencies = [
 
 [[package]]
 name = "mirrord-protocol"
-version = "1.24.2"
+version = "1.25.0"
 dependencies = [
  "actix-codec",
  "bincode",
--- a/mirrord/protocol/Cargo.toml
+++ b/mirrord/protocol/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mirrord-protocol"
-version = "1.24.2"
+version = "1.25.0"
 authors.workspace = true
 description.workspace = true
 documentation.workspace = true
@@ -30,7 +30,12 @@ http-body-util = { workspace = true }
 fancy-regex = { workspace = true }
 socket2.workspace = true
 semver = { workspace = true, features = ["serde"] }
-derive_more = { workspace = true, features = ["from", "into", "deref", "display"] }
+derive_more = { workspace = true, features = [
+  "from",
+  "into",
+  "deref",
+  "display",
+] }
 uuid = { workspace = true, features = ["v4"] }
 
 mirrord-macros = { path = "../macros" }


flake.nix
```diff
@@ -94,8 +94,9 @@
           env =
             with pkgs;
             lib.optionalAttrs stdenv.isDarwin {
-              # Tells bindgen which C compiler to compile frida with when targetting linux
+              # Tells bindgen which C/C++ compiler to compile frida with when targetting linux
               CC_x86_64_unknown_linux_gnu = lib.getExe pkgsCross.gnu64.stdenv.cc;
+              CXX_x86_64_unknown_linux_gnu = lib.getExe pkgsCross.gnu64.stdenv.cc;
```
MintSoup, line 102: What's this for?

mirrord/intproxy/src/proxies/incoming/subscriptions.rs
```diff
@@ -284,19 +284,28 @@ impl SubscriptionsManager {
...
                 let port = match blocked_action {
-                    BlockedAction::Steal(steal_type) => steal_type.get_port(),
-                    BlockedAction::Mirror(port) => *port,
-                };
-                let Some(subscription) = self.subscriptions.remove(&port) else {
-                    return Ok(vec![]);
+                    BlockedAction::Steal(steal_type) => Some(steal_type.get_port()),
+                    BlockedAction::Mirror(port) => Some(*port),
+                    BlockedAction::OutgoingTcp(addr) | BlockedAction::OutgoingUdp(addr) => {
+                        addr.get_port()
```
MintSoup: Should this be here? This is `IncomingProxy`, so this branch should never run (given the operator doesn't do anything funky).
Any connect errors should be handled by `OutgoingProxy::handle_connect_response`, but that one just logs and exits so I don't think anything extra is necessary.

MintSoup (changes requested):

flake.nix
iniw: You mean the file or the new variable? 

The file is a nix flake, which declares a "devShell" which is like a sandboxed hermetic environment that contains all the packages to work on the project

The new variable is an env var to tell rustc/rust-analyzer which C compiler to use for frida (see https://github.com/metalbear-co/mirrord/blob/main/CONTRIBUTING.md#compiling-on-macos)

mirrord/intproxy/src/proxies/incoming/subscriptions.rs
iniw: You're right - I added this to fix the "unhandled pattern" errors that appeared once I added the two new enum variants, but I'll just turn this branch into an `unreachable!()`.

flake.nix
MintSoup: Yeah but I see we already had a var for the C compiler, why did we add a C++ compiler too?

mirrord/intproxy/src/proxies/incoming/subscriptions.rs
MintSoup: IMO `unreachable!()` is more for branches that are genuinely unreachable but the compiler is unable to prove as such, but since this depends on data coming from the network it's _technically_ reachable (operator may be buggy, etc etc). And this error is recoverable anyway so I think handling it with a `tracing::error!` and moving on would make for a better UX than panicking. Resolving anyway because it's a minor nit

flake.nix
iniw: Oh it's just cause the operator uses `librdkafka` which is C++ and compiles with cmake, so it uses both `CC` and `CXX`, so I added the `CXX` var there and just decided to port it over to the mirrord codebase too.

mirrord/intproxy/src/proxies/incoming/subscriptions.rs
iniw: That makes more sense, updated to just log and early-return instead.

flake.nix
MintSoup: Could be a separate PR but it's not worth bothering with, just mention it somewhere (changelog or whatnot) so it's easier to track later

MintSoup (approved):
:pray:

mirrord/agent/src/entrypoint.rs
```diff
@@ -537,6 +539,28 @@ impl ClientConnectionHandler {
...
+                let hostname = match dns_lookup::lookup_addr(&request.ip_address) {
+                    Ok(hostname) => Some(hostname),
+                    Err(error) => {
+                        debug!(
+                            "Reverse DNS lookup failed for {}: {}",
+                            request.ip_address, error
+                        );
+                        None
+                    }
+                };
```
meowjesty: Why is this just a log + option? If the lookup fails, what do we want to happen?

changelog.d/+outgoing-network-policy.added.md
```diff
@@ -0,0 +1 @@
+Support for limiting outgoing access from mirrord sessions using a kubernetes policy has been added to the operator.
```
meowjesty:
```suggestion
Support for limiting outgoing traffic from mirrord sessions using a kubernetes policy has been added to the operator.
```

I think we mostly use "traffic" everywhere.

mirrord/agent/src/entrypoint.rs
iniw: Well, it depends? This should only be a hard error if we have hostname-based filters I guess, since we won't be able to apply them. It started out as an `Option` because initially I didn't have the reverse DNS implemented at all and so it just stayed like that.

changelog.d/+outgoing-network-policy.added.md
iniw: I'll use the word "connections" instead since "traffic" implies it's about blocking reads/writes after a connection has been made.

mirrord/operator/src/crd/policy.rs
```diff
@@ -232,6 +237,82 @@ pub struct IncomingNetworkPolicy {
...
+/// IP block specification for outgoing connections.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct IpBlock {
+    /// CIDR notation of the IP range (e.g., "10.0.0.0/16").
+    pub cidr: String,
+
+    /// List of CIDR ranges to exclude from the IP range.
+    #[serde(default)]
+    pub except: Vec<String>,
```
meowjesty: Nit: I think a name like `exclude` is better here, considering how in some places this gets called `exception`, that term means something else, so might cause confusion.

mirrord/operator/src/crd/policy.rs
```diff
@@ -232,6 +237,82 @@ pub struct IncomingNetworkPolicy {
...
+/// Hostname rule specification for outgoing connections.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct HostnameRule {
+    /// List of exact hostnames.
+    #[serde(default)]
+    pub exact: Vec<String>,
+
+    /// Regex pattern for matching hostnames.
+    pub expression: Option<String>,
```
meowjesty: Nit 2: Maybe `pattern`? Do we use `expression` for other regexes?

mirrord/operator/src/crd/policy.rs
iniw: The name was taken from the [Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource) that inspired our spec, so I think there's value in keeping it. What do you think?

mirrord/operator/src/crd/policy.rs
iniw: Don't think so. I'm not even sure there is any need for `exact` now that `expression` is a regex. @0x00A5 can I just remove it and make `hostname` a `Option<String>` instead of a struct and always treat is a regex?

The user can always recreate `exact` with a `^string$` regex

mirrord/operator/src/crd/policy.rs
meowjesty: Oh, dang, yeah let's keep whatever kubernetes uses.

mirrord/operator/src/crd/policy.rs
iniw: Talked to Yuan on slack: the `HostnameRule` struct has been removed, the `hostname` field of `OutgoingRule` is now simply a `Option<Regex>`.

mirrord/agent/src/entrypoint.rs
iniw: Errors are now propagated from the agent back to the operator and the connection is blocked (with a special message) if the rDNS lookup fails and we have policy rules that contain hostname filters.

mirrord/agent/src/entrypoint.rs
```diff
@@ -546,6 +549,27 @@ impl ClientConnectionHandler {
                     .make_request(ClientGetAddrInfoRequest::V2(request))
                     .await?;
             }
+            ClientMessage::ReverseDnsLookup(request) => {
+                debug!(
+                    "ClientMessage::ReverseDnsLookup client id {:?} ip {:?}",
+                    self.id, request.ip_address
+                );
```
meowjesty: 🗑️

mirrord/protocol/src/dns.rs
```diff
@@ -124,3 +124,19 @@ impl From<GetAddrInfoRequest> for GetAddrInfoRequestV2 {
...
+pub struct ReverseDnsLookupRequest {
+    pub ip_address: IpAddr,
+}
+
+/// Response from reverse DNS lookup.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct ReverseDnsLookupResponse {
+    /// The hostname associated with the IP address,
+    /// Contains an error if the lookup failed.
+    pub hostname: Result<String, ResponseError>,
```
meowjesty:
```suggestion
    pub hostname: RemoteResult<String>,
```

mirrord/operator/src/crd/policy.rs
```diff
@@ -232,6 +237,70 @@ pub struct IncomingNetworkPolicy {
...
 
+/// Outgoing network operations policy that partialy mimics the mirrord `network.outgoing` config.
+#[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct OutgoingNetworkPolicy {
+    /// List of allowed outgoing connections.
+    /// If specified, only connections matching these rules will be permitted.
+    /// Evaluated before `block` rules, so a connection that matches both will be blocked.
+    #[serde(default)]
+    pub allow: Vec<OutgoingRule>,
```
meowjesty, line 248: I think these `Vec`s should probably be `HashSet`s, we don't want duplicate rules, duplicate ports, duplicate anything in these lists.

PR #3789, opened 2025-12-11
MBE-1585 Add sccache to mirrord

vladrbg:


.github/workflows/ci.yaml
```diff
@@ -255,15 +255,47 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: docker/setup-buildx-action@v3
+      - name: Sanitize branch name for cache prefix
+        id: sanitize
+        run: |
+          BRANCH_NAME="${{ github.ref_name }}"
+          SANITIZED=$(echo "$BRANCH_NAME" | sed 's/\//-/g')
+          echo "cache_tag=$SANITIZED" >> "$GITHUB_OUTPUT"
+      - name: Check sccache secrets
```
pavelzag, line 264: Does these print something when a secret is passed from GitHub?

vladrbg: not exposing any private data, for secrets we just print the bucket name for now, but for the creds its just the length to see if empty or not. Currently this is just for me while debugging but doesn't hurt to leave it also because length and bucket name should not be problematic. and cache tag is just the branch name but without special chars that we cannot use

PR #3800, opened 2025-12-17
Fix db branching

0x00A5:
Fix get env from rollout target.
Add support for getting connection source from `envFrom`.

### DIFF
--- a/mirrord-schema.json
+++ b/mirrord-schema.json
@@ -1034,7 +1034,7 @@
       ]
     },
     "DbBranchingConnectionSourceKind": {
-      "description": "Different ways to source the connection options.",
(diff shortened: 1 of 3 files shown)


mirrord/operator/src/crd/mysql_branching.rs
```diff
@@ -56,6 +56,12 @@ pub enum ConnectionSourceKind {
         container: Option<String>,
         variable: String,
     },
+
+    /// Environment from resource reference in the the pod template.
+    EnvFrom {
+        container: Option<String>,
```
0x00A5, line 62: This field is never used. We should always use the container defined under target 🤷

PR #3810, opened 2025-12-21
Agent flag to cleanup dirty iptables instead of erroring out on startup

t4lz:


mirrord/agent/src/cli.rs
```diff
@@ -63,6 +63,13 @@ pub struct Args {
     /// cluster.
     #[arg(long, default_value_t = false, env = envs::IPV6_SUPPORT.name)]
     pub ipv6: bool,
+
+    /// Enable support for IPv6-only clusters
```
aviramha: woops?

t4lz: Thanks, fixed in https://github.com/metalbear-co/mirrord/pull/3810/changes/7e648199b3ab19b53d1c6627e13b299b30a3eec3

PR #3812, opened 2025-12-22
INT-248 Db branches support for local redis

vladrbg:


mirrord/cli/src/local_redis.rs
```diff
@@ -0,0 +1,304 @@
...
+};
+use mirrord_progress::Progress;
+use tokio_retry::{Retry, strategy::ExponentialBackoff};
+
+use crate::{CliError, CliResult};
+
+/// Represents a local Redis instance managed by mirrord.
+pub enum LocalRedis {
+    /// Docker container (container name).
+    Docker(String),
```
DmitryDodzin: I think we should use the `ContainerRuntime` from the `mirrord container` feature to allow the usage of docker-like runtimes (podman for example), we can default to Docker but allowing an override would be nice

mirrord/cli/src/local_redis.rs
```diff
@@ -0,0 +1,304 @@
...
+    let mut sub = progress.subtask("starting local Redis");
+    sub.print("starting native redis-server...");
+
+    let port = config.port;
+
+    // Build redis-server command with port and any custom args
+    let mut redis_args = vec!["--port".to_string(), port.to_string()];
+    redis_args.extend(config.options.args.iter().cloned());
+
+    let child = Command::new("redis-server")
```
DmitryDodzin: allowing the override of the redis-server command to path in user provided config (this can allow to use specific redis-server versions if needed or if it does not exist in PATH)

mirrord/cli/src/local_redis.rs
vladrbg: Is it okay if I move the `ContainerRuntime` inside the `config` crate ?

mirrord/cli/src/local_redis.rs
```diff
@@ -0,0 +1,331 @@
...
+use mirrord_progress::Progress;
+use tokio_retry::{Retry, strategy::ExponentialBackoff};
+
+use crate::{CliError, CliResult};
+
+/// Represents a local Redis instance managed by mirrord.
+pub enum LocalRedis {
+    /// Container (runtime command, container name).
+    Container {
+        runtime: String,
```
DmitryDodzin:
```suggestion
        runtime: ContainerRuntime,
```

I would keep it as ContainerRuntime and not to_string it yet

mirrord/config/src/container.rs
```diff
@@ -1,11 +1,44 @@
...
 use crate::config::source::MirrordConfigSource;
 
+/// Container runtimes supported by mirrord.
+#[derive(
+    Clone, Copy, Debug, Eq, PartialEq, JsonSchema, Serialize, Deserialize, Default, ValueEnum,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum ContainerRuntime {
+    /// Docker container runtime. (default)
+    #[default]
```
DmitryDodzin, line 17: I just would check it doesn't affect the container command, just that it still works and does a correct error message if you do some unsupported runtime (like that it doesn't auto default because of clap logic or something)

mirrord/cli/src/local_redis.rs
```diff
@@ -0,0 +1,331 @@
...
+    // Remove any existing container with same name
+    let _ = Command::new(runtime_cmd)
+        .args(["rm", "-f", &container_name])
+        .output();
+
+    // Build container run command
+    let mut container_args = vec![
+        "run".to_string(),
+        "-d".to_string(),
+        "--name".to_string(),
```
DmitryDodzin, line 155: I would consider using `--rm` also to not fill the user's engine with exited containers (or make it configurable by user with enabled by default)

DmitryDodzin (approved):
LGTM

PR #3816, opened 2025-12-22
sip: exit when SIP patch fails due to hitting FD limit

gememma:
Issue: https://linear.app/metalbear/issue/COR-181/sip-failure-doesnt-crashprovide-enough-information-for-user


mirrord/cli/src/execution.rs
```diff
@@ -376,7 +376,14 @@ impl MirrordExecution {
                 .transpose() // We transpose twice to propagate a possible error out of this
                 // closure.
             })
-            .transpose()?;
+            .transpose()
+            .map_err(|sip_error| {
```
iniw: Makes more sense to use `inspect_err` for this since we're not actually mapping the error to anything

PR #3817, opened 2025-12-23
Correctly terminate redirected TCP connections

MintSoup:
This PR introduces 2 changes:
1. The agent pod will automatically exit once it detects the main container of the targeted pod has finished executing. In the OSS flow, this prevents redirected TCP connections from staying open after the main pod has exit.
2. Redirected TCP and HTTP connections will now be gracefully shut down *before* iptables rules are flushed. Flushing iptables rules before closing TCP prevents graceful shutdown as FIN packets are sent from the wrong port.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -690,12 +690,12 @@ jobs:
           config: ".markdownlint.json"
           globs: "README.md"
 
-  windows_build:
-    needs: changed_files
-    if: ${{ needs.changed_files.outputs.rs_changed == 'true' ||needs.changed_files.outputs.windows_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' }}
-    uses: ./.github/workflows/windows-build.yaml
+  # windows_build:
+  #   needs: changed_files
+  #   if: ${{ needs.changed_files.outputs.rs_changed == 'true' ||needs.changed_files.outputs.windows_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' }}
+  #   uses: ./.github/workflows/windows-build.yaml
 
-    secrets: inherit
+  #   secrets: inherit
 
   # we build mirrord to run ide tests, while it can take time for e2e/integration to finish,
   # building concurrently can run faster the release ide tests
@@ -761,7 +761,7 @@ jobs:
         lint_markdown,
         check-rust-docs,
         cargo-deny,
-        windows_build,
+        # windows_build,
       ]
     runs-on: ubuntu-24.04
     steps:
@@ -782,6 +782,6 @@ jobs:
             (needs.intellij_e2e_on_release_branch.result == 'success' || needs.intellij_e2e_on_release_branch.result == 'skipped') &&
             (needs.vscode_e2e_on_release_branch.result == 'success' || needs.vscode_e2e_on_release_branch.result == 'skipped') &&
             (needs.check-rust-docs.result == 'success' || needs.check-rust-docs.result == 'skipped') &&
-            (needs.cargo-deny.result == 'success' || needs.cargo-deny.result == 'skipped') &&
-            (needs.windows_build.result == 'success' || needs.windows_build.result == 'skipped') }}
+            (needs.cargo-deny.result == 'success' || needs.cargo-deny.result == 'skipped') }} # &&
+            # (needs.windows_build.result == 'success' || needs.windows_build.result == 'skipped') 
         run: echo $CI_SUCCESS && if [ "$CI_SUCCESS" == "true" ]; then echo "SUCCESS" && exit 0; else echo "Failure" && exit 1; fi
--- a/changelog.d/+ci-improvement-windows.internal.md
+++ b/changelog.d/+ci-improvement-windows.internal.md
@@ -1,0 +1,1 @@
+Don't force install existing software on Windows CI
--- a/scripts/win_build_test_apps.ps1
+++ b/scripts/win_build_test_apps.ps1
@@ -34,8 +34,12 @@ function Install-NasmPortable {
 }
 
 # Install everything except nasm via Chocolatey to reduce privilege needs; fetch nasm manually.
-choco install -y --force nodejs go python3 curl mingw cmake llvm
-Install-NasmPortable
+choco install -y nodejs go python3 curl mingw cmake llvm
+if (-not (Get-Command nasm -ErrorAction SilentlyContinue)) {
+    Install-NasmPortable
+} else {
+    Write-Host 'nasm already installed.'
+}
 
 Write-Host 'Preparing Python virtual environment for test dependencies'
 $venvRoot = Join-Path $env:TEMP "mirrord-test-venv"
@@ -52,7 +56,7 @@ try {
 
     Write-Host 'Installing Python packages required by tests'
     python -m pip install --upgrade pip
-    python -m pip install uvicorn fastapi flask --ignore-installed
+    python -m pip install uvicorn fastapi flask
     Write-Host 'Installed Python packages:'
     python -c "import pkgutil, sys; [print(module.module_finder.path) for module in pkgutil.iter_modules()]"
 } finally {


mirrord/agent/src/entrypoint.rs
```diff
@@ -726,9 +729,41 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
+        // As soon as the main container exits, we want to exit as well.
+        // This not only makes for a better UX (no unused agent pods
+        // lingering), but is important for terminating redirected
+        // connections gracefully. If we don't exit as soon as the main
+        // container dies, the veth interface will be deleted while
+        // redirected connections are still open, and thus the other ends
+        // will stay open with no way to terminate them correctly.
+        let cancel = cancellation_token.clone();
+        let pidfd = AsyncPidFd::from_pid(target_pid.try_into().unwrap())
+            .expect("could not acquire pidfd to target process");
```
Razz4780: I'd make everything related to this new thing fallible.
If anything goes wrong, log an error and continue without it

mirrord/agent/src/incoming/connection/tcp.rs
```diff
@@ -131,10 +148,19 @@ impl RedirectedTcp {
                 buffer: BytesMut::with_capacity(64 * 1024),
                 mirror_data_tx,
             };
-            let result = copy_bidirectional::copy_bidirectional(&mut self.io, &mut outgoing).await;
-            outgoing
-                .mirror_data_tx
-                .send_item(IncomingStreamItem::Finished(result));
+            tokio::select! {
```
Razz4780: Call to `self.make_pass_through_connection` above needs the same logic

mirrord/agent/src/incoming/task.rs
```diff
@@ -208,22 +211,35 @@ where
...
+                        .await;
+
+                    rx.await.expect("TcpStealerTask dropped oneshot tx for returning JoinHandle to IO task for TCP connection")
                 }
-                None => {
-                    redirected.pass_through();
+                None => redirected.pass_through(port_state.shutdown.child_token()),
+            };
+
+            port_state.connections.spawn(async {
```
Razz4780: Looks like `PortState::connections` is never polled for task results until `PortState::shutdown`. If we have a long port subscription, and a constant stream of incoming connections, the agent will eventually go OOM

Razz4780 (changes requested):
Left some comments. Overall looks good. For graceful shutdown of redirected connections, I'd also consider implementing it in RAII manner, where `RedirectorTask` alone keeps track of live connections using the redirected port. Since it wraps the IO stream in `Box<dyn ...>`, it could also store a guard in there. And then wait for all guards to be dropped in order to drop the redirect rule. Does this make sense?

mirrord/agent/src/incoming/connection/tcp.rs
MintSoup: You mean the `select!` with the `.cancelled()` stuff? Why? It's connecting to the main container process so it'll shut down anyway, and it's not even a long lived future (resolves as soon as the connection is done).

Razz4780: Whelp, I think from our perspective it's kinda the same thing – we exchange TCP packets with the target app

mirrord/agent/src/incoming/error.rs
```diff
@@ -53,4 +53,6 @@ pub enum ConnError {
     BroadcastLag,
     #[error("bug in the mirrord-agent, please report it: {0}")]
     AgentBug(String),
+    #[error("connection cancelled")]
+    Cancelled,
```
Razz4780:
```suggestion
    #[error("connection cancelled, mirrord-agent is exiting")]
    AgentExiting,
```
This is should be the only case, so we can make the log friendlier

mirrord/agent/src/entrypoint.rs
```diff
@@ -658,6 +659,41 @@ pub async fn notify_client_about_dirty_iptables(
...
+                        "wait() on target container did not return expected value"
+                    );
+                }
+            }
+        }
+        tracing::debug!("Main target process exited, shutting down...");
+        cancel.cancel();
+    });
+    Ok(())
+}
```
Razz4780, line 688:
```suggestion
/// Monitors the main container process and cancels `cancel` when it
/// exits.
/// 
/// This not only makes for a better UX (no unused agent pods
/// lingering), but is important for terminating redirected
/// connections gracefully. If we don't exit as soon as the main
/// container dies, the veth interface will be deleted while
/// redirected connections are still open, and thus the other ends
/// will stay open with no way to terminate them correctly.
fn monitor_main_container(cancel: CancellationToken, pid: libc::pid_t) {
    let fd = match AsyncPidFd::from_pid(pid) {
        Ok(fd) => fd,
        Err(error) => {
            tracing::error!(
                %error,
                "Failed to get the target container process descriptor",
            );
            return;
        },
    };

    tokio::spawn(async move {
        tracing::trace!("Starting watcher task for main target process");
        match fd.wait().await {
            Ok(info) => {
                tracing::warn!(
                    status = %info.status(),
                    "Target container process exited, shutting down...",
                );
                cancel.cancel();
            }
            Err(error) if error.raw_os_error() == Some(libc::ECHILD) => {
                tracing::warn!("Target container process exited, shutting down...");
                cancel.cancel();
            }
            Err(error) => {
                tracing::error!(%error, "Failed to wait() on target container");
            }
        }
    });
}
```
Mostly style. One important change - don't cancel the token if `wait()` fails with an unexpected error code

mirrord/agent/src/incoming/task.rs
```diff
@@ -148,44 +151,101 @@ where
...
+    /// unconditionally passed through. Otherwise, the connection is
+    /// dropped. We consider this to be an unlikely race condition.
     #[tracing::instrument(level = Level::TRACE, ret)]
-    fn handle_connection(&self, conn: Redirected) {
+    fn handle_connection(&mut self, conn: Redirected) {
         let source = conn.source;
         let destination = conn.destination;
 
-        if self.ports.contains_key(&conn.destination.port()).not() {
+        let Entry::Occupied(mut e) = self.ports.entry(destination.port()) else {
```
Razz4780: You're calling `get_mut()` on it right away, might as well go with

```rust
let Some(state) = self.ports.get_mut(&destination.port()) else {
    // ...
};
```

mirrord/agent/src/incoming/task.rs
```diff
@@ -403,12 +491,37 @@ where
...
+        while let Some(joined) = port_state.connections.try_join_next() {
+            if let Err(err) = joined {
+                tracing::warn!(?err, "IO task returned JoinError");
+            }
+        }
+
+        port_state.connections.spawn(async move {
+            f.await;
+
+            tx.send(InternalMessage::DeadChannel(port))
```
Razz4780: `DeadChannel` means that port subscription was dropped. Now you're using it for something else, so its meaning changed and its doc is outdated.

mirrord/agent/src/incoming/task.rs
```diff
@@ -559,9 +674,17 @@ impl fmt::Debug for PortState {
...
 }
 
-impl Drop for PortState {
-    fn drop(&mut self) {
-        self.http_shutdown.cancel();
+impl PortState {
+    /// Tell and wait for all connections to gracefully shut down.
+    /// This function is essentially `AsyncDrop`, and it should always
+    /// be called before removing the redirection.
+    async fn graceful_shutdown(mut self) {
```
Razz4780, line 687: Wait. Looks like this is only called in:
* `RedirectorTask::handle_dead_channel`, if `PortState::connections` is empty - doesn't really do anything there.
* `RedirectorTask::cleanup`, which is only called when the task is exiting. In this case, we can just drop `JoinSet`, which will immediately abort all tasks 👀 

Am I missing something? What do we get from these extra cancellation tokens?

I'm getting lost in this change 👀

mirrord/agent/src/entrypoint.rs
MintSoup: Also, I'm pretty sure we **don't** want to cancel if it returns `Ok(_)` - we're supposed to get an `ECHILD` because the target process isn't our child. That being said, my idea was to fail fast to catch any potential bugs ASAP, although that probably makes for worse UX.

mirrord/agent/src/incoming/task.rs
MintSoup: :skull: This is basically some cruft left over from (many) previous design iterations, which I forgot to clean up. The idea was to extend the graceful-http-shutdown-when-all-clients-unsubscribe logic to all connections, AND to ensure that all existing connections are terminated correctly before removing the port redirection, AND to correctly handle the case where the application unsubscribes but we have existing connections, which we also want to keep alive and terminate only it's correct to do so.

You're correct, this needs to change.
* In `RedirectorTask::handle_dead_channel`, it's effectively a noop.
* In `RedirectorTask::cleanup`, it triggers a *graceful* shutdown of both redirected http and tcp connections. This is good in the normal flow, but when the target shuts down, we should abort all tasks because we want to exit ASAP. I suggest adding some mechanism for telling `RedirectorTask` how to shut down, gracefully or ASAP, wdyt?

Also while looking through the code, I discovered a tangentially related bug on `main`, and I think it'd be good to fix it here because we're touching relevant machinery anyway. The following flow breaks over mirrord:

1. Start HTTP1.1 or HTTP/2 server, anything that allows serving multiple request over a single TCP connection
2. Accept **ONE** connection.
3. Close the server socket.
4. Serve requests from the accepted connection.

On step 3, we unsubscribe from the port, so we cannot accept any more requests. But without mirrord, the existing TCP connections would keep functioning.

I think the issue stems from the fact that we don't do 1:1 mapping of TCP connections when HTTP is used, and in our mental model the requests arrive on the server socket, not on some client connection accepted by it.

mirrord/agent/src/entrypoint.rs
Razz4780: Ah right, we should not get `Ok(_)` ever.

Re failing fast, I do think it would make better UX. The only reason I asked for continuing on failure is stability. If there was a bug somewhere here (not even in this code, something like a missing Linux capability for the agent process, or selinux/apparmor config), it would kill the whole mirrord flow for the current users.

Razz4780 (approved):

PR #3822, opened 2025-12-25
linux to windows release build link

pavelzag:
This PR allows to trigger Windows build after a successful linux release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -468,7 +468,7 @@ jobs:
     needs: update_latest
     if: ${{ !cancelled() && !failure() }}
     permissions:
-      contents: write
+      contents: read
     uses: ./.github/workflows/windows-build.yaml
     with:
 


aviramha:
We can still split it - just use reusable workflow. I don't like that it's completely detached from the workflow.

pavelzag:
@aviramha Changed some stuff. Let me know what you think

.github/workflows/README.md
```diff
@@ -0,0 +1,81 @@
+# Workflow Structure & Options
+
+This document outlines the three main ways the Windows build workflow operates in the new structure.
```
aviramha: Why do we have a README just for the windows build?

.github/workflows/release.yaml
```diff
@@ -461,3 +466,17 @@ jobs:
           update-latest: true
           # Don't update the vX.Y tags
           update-minor: false
+
+  trigger_windows_build:
+    needs: update_latest
+    if: ${{ !cancelled() && !failure() && (inputs.build_windows || inputs.build_windows == null) }}
```
aviramha: why do we have still build_windows flag?

.github/workflows/ci.yaml
```diff
@@ -654,6 +654,14 @@ jobs:
           config: ".markdownlint.json"
           globs: "README.md"
 
+  windows_build:
+    needs: changed_files
+    if: ${{ needs.changed_files.outputs.windows_changed == 'true' }}
+    uses: ./.github/workflows/windows-build.yaml
+    with:
+      mode: build_sign
```
aviramha: why are we signing a ci/test build?

.github/workflows/ci.yaml
aviramha: we don't need a windows_changed anymore.

aviramha: just add `.ps` to the general "rs_changed"

.github/workflows/ci.yaml
```diff
@@ -654,6 +654,14 @@ jobs:
...
+    secrets: inherit
+
   # we build mirrord to run ide tests, while it can take time for e2e/integration to finish,
   # building concurrently can run faster the release ide tests
   build_mirrord_on_release_branch:
```
aviramha, line 702: > Workflow does not contain permissions
Check warning: Workflow does not contain permissions
Actions job or workflow does not limit the permissions of the GITHUB_TOKEN. Consider setting an explicit permissions block, using the following as a minimal starting point: {{}}

.github/workflows/ci.yaml
aviramha: nvm, I realize name is just confusing. Let's change name to build. sign_artiffacts is a modifier of a build..

.github/workflows/release.yaml
```diff
@@ -6,6 +6,11 @@ on:
   # Running from workflow dispatch (AKA manual) will not publish anything.
   # This is intended for testing changes to this flow.
   workflow_dispatch:
+    inputs:
+      build_windows:
+        description: "Trigger Windows build after release?"
```
aviramha: why do we need this input?

.github/workflows/ci.yaml
```diff
@@ -656,13 +650,22 @@ jobs:
           config: ".markdownlint.json"
           globs: "README.md"
 
+  windows_build:
+    needs: changed_files
+    if: ${{ needs.changed_files.outputs.rs_changed == 'true' }}
```
aviramha: if ci changed windows needs to run too.

.github/workflows/ci.yaml
```diff
@@ -656,13 +650,22 @@ jobs:
...
+    uses: ./.github/workflows/windows-build.yaml
+
+    secrets: inherit
+
   # we build mirrord to run ide tests, while it can take time for e2e/integration to finish,
   # building concurrently can run faster the release ide tests
   build_mirrord_on_release_branch:
     runs-on: ubuntu-24.04
```
aviramha, line 703: seems tehre's still CodeQL warning?

.github/workflows/README.md
```diff
@@ -0,0 +1,85 @@
+# Workflow Structure & Options
```
aviramha, line 1: I don't have hard opinion on this file but will ask @Razz4780 / @0x00A5  / @t4lz for their opinion before merging.

.github/workflows/ci.yaml
pavelzag: I don't see it any more..

.github/workflows/ci.yaml
```diff
@@ -98,7 +100,16 @@ jobs:
             rust-toolchain.toml
             rustfmt.toml
             deny.toml
+            deny.toml
```
t4lz: second one is for good luck?

.github/workflows/ci.yaml
```diff
@@ -161,14 +159,13 @@ jobs:
...
-          echo ${{ steps.changed-windows.outputs.any_changed }};
-          echo ${{ steps.changed-windows.outputs.all_changed_files }};
+
           echo ${{ steps.changed-wizard-frontend.outputs.any_changed }};
           
   lint:
     runs-on: ubuntu-24.04
     needs: changed_files
-    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' || needs.changed_files.outputs.windows_changed == 'true'}}
+    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
```
t4lz, line 174: I don't get why, also before this PR, we were letting windows-specific changes trigger the rerunning of linux-only linting.
Instead of making `rs_changed` include both the general changes and the windows-specific changes, we should keep the separate `windows_changed` output, and keep this change, deleting the `windows_changed` check from this step, so that windows-specific changes can't on their own trigger running linux-specific checks.

.github/workflows/ci.yaml
```diff
@@ -124,20 +135,7 @@ jobs:
...
-        with:
-          files: |
-            mirrord/layer-win/**
-            mirrord/str-win/**
-            mirrord/**/windows/**
-            mirrord/**/win32/**
-            mirrord/**/win/**
-            scripts/**/*.ps1
-            **/*.ps1
-            .github/workflows/windows-build.yaml
```
t4lz, line 140: Let's keep the separate `changed-windows` output, so that we can avoid unnecessarily running non-windows checks, if only windows-specific files changed.

.github/workflows/ci.yaml
```diff
@@ -72,7 +75,6 @@ jobs:
       ci_changed: ${{ steps.changed-ci.outputs.any_changed }}
       protocol_changed: ${{ steps.changed-protocol.outputs.any_changed }}
       dockerfile_changed: ${{ steps.changed-dockerfile.outputs.any_changed }}
-      windows_changed: ${{ steps.changed-windows.outputs.any_changed }}
```
t4lz, line 75: Let's keep this

.github/workflows/ci.yaml
```diff
@@ -227,7 +224,7 @@ jobs:
 
   cargo-deny:
     needs: changed_files
-    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' || needs.changed_files.outputs.windows_changed == 'true'}}
+    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
```
t4lz: After reintroducing the `windows_changed` output, we'd need to undo this change, to rerun cargo-deny also on window-specific changes.

.github/workflows/ci.yaml
```diff
@@ -238,7 +235,7 @@ jobs:
   check-rust-docs:
     runs-on: ubuntu-24.04
     needs: changed_files
-    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' || needs.changed_files.outputs.windows_changed == 'true'}}
+    if: ${{needs.changed_files.outputs.rs_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true'}}
```
t4lz: After reintroducing the windows_changed output, we'd need to undo this change, to rerun cargo-deny also on window-specific changes.

.github/workflows/ci.yaml
```diff
@@ -99,6 +102,22 @@ jobs:
             rustfmt.toml
             deny.toml
             .cargo/**
+            mirrord/**/win/**
+            scripts/**/*.ps1
+            **/*.ps1
```
t4lz: Why add this at both changed-rs and changed-windows?

.github/workflows/ci.yaml
t4lz: @itsamegraf @cristeigabriela FYI, we're not rerunning this step on windows-specific changes anymore, since this step currently doesn't check windows-only code anyways.
You might want to add all the rustfmt, clippy, etc. checks for windows as well. It's probably best to do it in a separate PR though, not to hold back Pavel's change.

.github/workflows/ci.yaml
```diff
@@ -161,14 +167,13 @@ jobs:
           echo ${{ steps.changed-protocol.outputs.any_changed }};
           echo ${{ steps.changed-protocol-toml.outputs.any_changed }};
           echo ${{ steps.changed-dockerfile.outputs.any_changed }};
-          echo ${{ steps.changed-windows.outputs.any_changed }};
```
t4lz, line 164: Since we're keeping this output, let's not delete this line.

.github/workflows/ci.yaml
pavelzag: Not an expert on Rust build quirks yet, but Gemini told me that:

changed-rs: This triggers our standard Rust CI pipeline (likely checking formatting, clippy, and running unit tests on Linux).
Even if the code is Windows-specific (mirrord/**/win/**), we still want to ensure it is formatted correctly and doesn't contain syntax errors that might leak into shared code or break tooling.
For 
.ps1
 scripts, if they utilize any Rust components or are part of the broader repo integrity, having them tracked here might be a catch-all, though less critical than for .rs files.
 
 
changed-windows: This specifically triggers the Windows Build & Test workflow (
windows-build.yaml
).
This is the only workflow that actually runs the code on a Windows machine to verify compilation and functionality.

.github/workflows/ci.yaml
```diff
@@ -687,13 +693,22 @@ jobs:
           config: ".markdownlint.json"
           globs: "README.md"
 
+  windows_build:
+    needs: changed_files
+    if: ${{ needs.changed_files.outputs.windows_changed == 'true' || needs.changed_files.outputs.ci_changed == 'true' }}
```
t4lz: We also need to run this when rs_changed!

.github/workflows/ci.yaml
t4lz: Tell Gemini I said hi.
You can ignore its advice in this case though.
Currently all steps that don't depend on `windows_changed` don't do any windows checks, so there's no point in running them if there were only windows specific checks, and no general rust changes.
So just

```suggestion
```

t4lz (approved):

PR #3823, opened 2025-12-25
`CiApiKey` doesn't break on operator updates (temp solution).

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1088/persistent-ci-api-key

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,49 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
 
 <!-- towncrier release notes start -->
 
+## [3.180.0](https://github.com/metalbear-co/mirrord/tree/3.180.0) - 2026-01-06
+
+
+### Added
+
+- Add support local redis in db branches.
+- Added an experimental configuration for logging all Apple variables before
+  process start.
+- Support for limiting outgoing connections from mirrord sessions using a
+  kubernetes policy has been added to the operator.
+
+
+### Changed
+
+- Changed `experimental.force_hook_connect` default value to `true`.
+- Changed changelog instructions
+- Make file not found logs more detailed
+
+
+### Fixed
+
+- Fixed file not found Windows build regression
+- iptables cleanup would fail if some rules were already deleted.
+
+
+### Internal
+
+- CLI and Intproxy now queue and (almost) fairly schedule agent-bound messages.
+  Intproxy now also has proper flow control when receiving network traffic from
+  user apps.
+- Change from Prepare artifacts to build artifacts
+- Improve CI build times with cache
+- Improvements to the `outgoing_udp` test to make it less flaky.
+- JSONPath queries for JSON body filters are now also validated client-side, in
+  the CLI.
+- Move is_ci analytics reporting to analytics crate (and add it when building
+  the AnalyticsReporter).
+- New E2E test that verifies that when the cleanup on start setting is enabled,
+  the session is successful even when there are leftover rules on the target
+  pod.
+- Releases fix for windows builds
+- This PR allows to trigger Windows build after a successful linux release.
+
 ## [3.179.0](https://github.com/metalbear-co/mirrord/tree/3.179.0) - 2025-12-24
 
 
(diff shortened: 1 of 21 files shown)


mirrord/auth/src/credentials.rs
```diff
@@ -415,49 +436,69 @@ pub mod client {
...
 
         /// Decoding a CI API key by first checking the prefix to determine the version.
         /// Then follow the encoding process in reverse for each version.
         pub fn decode(encoded: &str) -> Result<Self, ApiKeyError> {
-            if let Some(rest) = encoded.strip_prefix(Self::V1_PREFIX) {
-                let compressed_bytes = BASE64_URL_SAFE_NO_PAD.decode(rest)?;
+            let (rest, version) = encoded
+                .strip_prefix(Self::V1_PREFIX)
+                .zip(Some(Self::V1_PREFIX))
+                .ok_or(ApiKeyError::InvalidFormat)?;
```
0x00A5: The original code I wrote wasn't pretty :D. The plan was to just 
```rust
if let Some(rest) = encoded.strip_prefix(v1_prefix) {
    decode_v1(rest);
} else if let Some(rest) = encoded.strip_prefix(v2_prefix) {
    decode_v2(rest);
} else {
    return Err(ApiKeyError::InvlaidFormat);
}
```

But line 482 to 485 makes it only work with v1 prefix.

Maybe the original code can be improved by some dispatching like this:
```rust
let decoders = [
    (v1_prefix, decode_v1),
    (v2_prefix, decode_v2),
];

if let Some((rest, handler)) = decoders.iter().find_map(|(prefix, decode)| {
    value.strip_prefix(prefix).map(|rest| (rest, decode))
}) {
    decode(rest);
} else {
    return Err(ApiKeyError::InvlaidFormat);
}
```

mirrord/operator/src/client.rs
```diff
@@ -295,9 +295,7 @@ impl OperatorApi<NoClientCert> {
         R: Reporter,
         P: Progress,
     {
-        let certificate = match ci_api_key {
-            CiApiKey::V1(credentials) => credentials.as_ref(),
-        };
+        let certificate = ci_api_key.credentials().as_ref();
```
0x00A5, line 298: nit: credentials() already returns ref, no need to `as_ref()` again?

mirrord/auth/src/key_pair.rs
```diff
@@ -175,6 +175,24 @@ impl<'de> Deserialize<'de> for KeyPair {
...
+
+impl<Context> bincode::Decode<Context> for KeyPair {
+    fn decode<D: bincode::de::Decoder<Context = Context>>(
+        decoder: &mut D,
+    ) -> Result<Self, bincode::error::DecodeError> {
+        let pem = String::decode(decoder)?;
+        Self::try_from(pem).map_err(bincode::error::DecodeError::custom)
+    }
+}
+
```
0x00A5: Are these used anywhere?

mirrord/auth/src/credentials.rs
meowjesty: Ah, you can kinda keep `zip`ping with different prefixes, did this while I was implementing it as the v1.5 (the other key idea), and kinda left it.

Can just revert it back tbh, when we come back to actually implementing the next version of the key, we can do your suggestion.

mirrord/operator/src/client.rs
meowjesty: Yup, but we want the `Certificate`, so we need to call `AsRef<Credentials>` to get it.

Before this change, it was the same: `CiApiKey::V1(credentials) => credentials.as_ref(),`

mirrord/auth/src/key_pair.rs
meowjesty: bef7dfd9587e3950c0cd378e36cb48bb79416ba2

Removed (they were, but not anymore).

mirrord/auth/src/credentials.rs
meowjesty: bef7dfd9587e3950c0cd378e36cb48bb79416ba2

Reverted back.

0x00A5 (approved):

PR #3824, opened 2025-12-26
Fix partial iptables cleanup, add E2E test for the fix and for a related feature from another PR

t4lz:
Fix: iptables cleanup would fail if some rules were already deleted.
Added E2E test that tests both that bugfix and the new feature from https://github.com/metalbear-co/mirrord/pull/3810

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,7 @@
 # Contributing
 
 Before submitting pull request features, please discuss them with us first by opening an issue or a discussion.
-We welcome new/junior/starting developers. Feel free to join to our [Slack](https://metalbear.co/slack) for help and
-guidance.
+We welcome new/junior/starting developers. Feel free to join to our [Slack](https://metalbear.co/slack) for help and guidance.
 
 If you would like to start working on an issue, please comment on the issue on GitHub, so that we can assign you to that
 issue.
@@ -35,8 +34,7 @@ The following guide details the steps to setup a local development environment f
 
 ### Setup a Kubernetes cluster
 
-For E2E tests and testing mirrord manually you will need a working Kubernetes cluster. A minimal cluster can be easily
-setup locally using either of the following:
+For E2E tests and testing mirrord manually you will need a working Kubernetes cluster. A minimal cluster can be easily setup locally using either of the following:
 
 - [Minikube](https://minikube.sigs.k8s.io/)
 - [Docker Desktop](https://www.docker.com/products/docker-desktop/)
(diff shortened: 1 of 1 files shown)


mirrord/agent/iptables/src/standard.rs
```diff
@@ -46,10 +46,9 @@ where
     }
 
     async fn unmount_entrypoint(&self) -> IPTablesResult<()> {
-        self.prerouting.unmount_entrypoint().await?;
-        self.output.unmount_entrypoint().await?;
-
-        Ok(())
+        let first_res = self.prerouting.unmount_entrypoint().await;
+        let second_res = self.output.unmount_entrypoint().await;
+        first_res.and(second_res)
```
0x00A5: What is the purpose of always try applying everything instead of failing early?

mirrord/agent/iptables/src/mesh.rs
```diff
@@ -97,10 +97,9 @@ where
     }
 
     async fn unmount_entrypoint(&self) -> IPTablesResult<()> {
-        self.prerouting.unmount_entrypoint().await?;
-        self.output.unmount_entrypoint().await?;
-
-        Ok(())
+        let first_res = self.prerouting.unmount_entrypoint().await;
+        let second_res = self.output.unmount_entrypoint().await;
+        first_res.and(second_res)
```
DmitryDodzin:
```suggestion
        let prerouting_unmount = self.prerouting.unmount_entrypoint().await;
        let output_unmount = self.output.unmount_entrypoint().await;
        prerouting_unmount.and(output_unmount)
```

Just because I think it looks nicer than enumerating them

mirrord/agent/iptables/src/mesh/exclusion.rs
```diff
@@ -108,16 +105,19 @@ where
     async fn mount_entrypoint(&self) -> IPTablesResult<()> {
         // Should mount the internal iptables because exclusion inserts itself as first rule and
         // this may also happen here so exclusion should mount second
-        self.inner.mount_entrypoint().await?;
+        let first_res = self.inner.mount_entrypoint().await;
```
DmitryDodzin: Do we want to mount the second if we failed the first? I agree for doing unmounts on fail but for mounting it's probably unnecessary

DmitryDodzin (commented):
Overall looks good, just maybe the variable names and the early error on one mount function =]

mirrord/agent/iptables/src/standard.rs
t4lz: If the first part does not exist in iptables, deletion will fail.
So if only the second part exists, if we fail early, we don't delete the second part.
By deleting the second part regardless of the first part's result, we handle that case properly - we cleanup whatever is left.

mirrord/agent/iptables/src/mesh/exclusion.rs
t4lz: Oops, you're right, that's a mistake. Thanks!

t4lz: fixed.

mirrord/agent/iptables/src/mesh.rs
t4lz: renamed all variables with those names

mirrord/agent/iptables/src/standard.rs
t4lz: Added a comment to all of these locations.

PR #3825, opened 2025-12-28
Malformed http filter json query fails silently

MintSoup:
Adds client-side JSONPath query validation for JSON body filter queries. We already have agent-side validation but this should improve UX as it will detect errors faster.

Note that the validation is done in the CLI, **not** the layer, unlike other similar checks (e.g. filter regexes). This is due to technical reasons: the JSONPath library we're using allows adding extension functions (`typeof`) solely through a proc macro, which generates code to register the extension at startup, using a *constructor function* (i.e. one that runs before main, using platform-specific linker magic). Problem is, `mirrord-layer` also does something similar, and it's not possible to reliably control the order in which constructor functions execute, so what ends up happening is the layer setup and subsequently the query validation runs *before* we've had a chance to register our extension functions, which means ext fns are not recognized in queries and we get false errors.

I'm also sliding in two (tangentially) related minor changes:
1. Fix body filters getting ignored when ensuring we don't have >1 HTTP filters enabled simultaneously
2. Adjusted levels for some logs related to JSON body filters.


mirrord/cli/src/port_forward.rs
```diff
@@ -979,7 +995,8 @@ impl IncomingMode {
                 // TODO(areg) unification
```
0x00A5, line 981: Unrelated, what does this TODO mean?

mirrord/config/src/lib.rs
```diff
@@ -495,6 +503,41 @@ impl LayerConfig {
             ))?;
         }
 
+        let verify_body_filter = |filter: &BodyFilter| match filter {
+            BodyFilter::Json { query, .. } => {
+                // Only need to verify `query` as `matches` is later
```
0x00A5, line 508: Since we are here already, can we check `matches` as well? Have these failures bubbled up as `ConfigError` seems nicer. WDYT?

mirrord/cli/src/port_forward.rs
MintSoup: Some of these functions are duplicates, there's a few copies of them floating between different crates. Need to refactor those into a separate utils crate at some point, it's one of the tech debt cleanup things.

mirrord/config/src/lib.rs
MintSoup: Since http filter validation is done (almost) entirely in the layer (including `matches`), my idea was to change that as little as possible and do only the bare minimum here. If we check `matches` we might as well move the validation for all filter variants into the CLI, otherwise I don't think it makes too much sense to split it like that. The current split is "whatever can be done in the layer is done in the layer" which is decent IMO.

0x00A5 (approved):

PR #3831, opened 2025-12-31
cli+config: Add "env-key", an identifier for mirrord sessions

iniw:
Part 1/3 of preview env


mirrord/cli/src/container.rs
```diff
@@ -224,7 +231,10 @@ pub async fn container_command(
 
     progress.warning("mirrord container is currently an unstable feature");
 
-    let cfg_context = ConfigContext::default().override_envs(exec_params.as_env_vars());
+    let cfg_context = ConfigContext::default()
+        .override_envs(exec_params.as_env_vars())
+        .with_cli_key(exec_params.key.clone());
```
0x00A5: Why not using the same override with envs pattern? Have an env var for `--key`, for example `MIRRORD_ENV_KEY`, then add it in `ExecParams`'s as_env_vars function.
In `LayerConfig` annotate the `key` field with the same env var name.

iniw: The `port-forward` command doesn't have access to a `ExecParams`, so doing it that way is simpler but it means `--key` wouldn't work for `port-forward`. Is that fine? I think it's fine.

iniw: Aviram said on slack that we can ignore `port-forward` for now.

0x00A5: Yeah, I am fine with it too.

iniw: Rewrote the entire thing (again). We now propagate the key through `as_env_vars`.

mirrord/config/src/env_key.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Generates a new UUID v4 key without any marker prefix.
+    ///
+    /// Used by [`EnvKeyFileConfig::generate_config`] when no key was found in the context at all
+    /// (i.e., when [`LayerFileConfig::from_path`] was never called, such as when using defaults).
+    ///
+    /// [`LayerFileConfig::from_path`]: crate::LayerFileConfig::from_path
+    pub fn autogenerate() -> String {
+        uuid::Uuid::new_v4().to_string()
+    }
```
0x00A5: nit: unlike `autogenerate_with_marker()`, it feels like this can just return `Self::Generated(uuid::Uuid::new_v4().to_string())`. But up to you whether to keep it this way.

0x00A5 (approved):
👍

PR #3833, opened 2025-12-31
NamedTempfiles for housekeeping

parames0:
Just submitting a draft PR for you review. So far, I did the following:

- Added util funcs under test to generate the tempfile and tempfile path
- 1 additional test to show that the tempfile path passes

A few questions on my end:

- Does the scope of this issue only cover config files under tests/ or also mirrord/layer/tests?
- Most of the tests are failing for me locally and differ on number of fails per run with the exact same code (i.e. 50 on first run, 60 on the next) - is this expected behaviour?

### DIFF
--- a/tests/src/cli/sanity.rs
+++ b/tests/src/cli/sanity.rs
@@ -45,12 +45,6 @@ mod cli {
 
         let tempfile = ManagedTempFile::new(CONFIG_JSON.clone());
 
-        println!("Tempfile path is: {:?}", &tempfile.path);
-
-        if tempfile.path.exists() == false {
-            assert!(false);
-        }
-
         let mut process = run_verify_config(Some(vec![
             "--ide",
             tempfile.path.to_str().expect("Valid config path!"),
--- a/mirrord/layer/tests/rebind0.rs
+++ b/mirrord/layer/tests/rebind0.rs
@@ -39,10 +39,6 @@ async fn rebind0(dylib_path: &Path, config_dir: &Path) {
     let tempfile = ManagedTempFile::new(config_json);
     config_path.push(&tempfile.path);
 
-    if tempfile.path.exists() == false {
-        assert!(false);
-    }
-
     let application = Application::RustRebind0;
     let (mut test_process, _intproxy) = application
         .start_process_with_layer(dylib_path, vec![], Some(&config_path))
--- a/tests/src/traffic/steal.rs
+++ b/tests/src/traffic/steal.rs
@@ -468,10 +468,6 @@ mod steal_tests {
         let tempfile = ManagedTempFile::new(config_json);
         config_path.push(&tempfile.path);
 
-        if tempfile.path.exists() == false {
-            assert!(false);
-        }
-
         let client = application
             .run(
                 &service.pod_container_target(),
@@ -532,12 +528,6 @@ mod steal_tests {
         });
         let tempfile = ManagedTempFile::new(config_json); 
         config_path.push(&tempfile.path);
-
-        println!("Tempfile path is: {:?}", &tempfile.path);
-
-        if tempfile.path.exists() == false {
-            assert!(false);
-        }
 
         let client = application
             .run(


t4lz:
Hi @parames0, thank you for your contribution!

> - Does the scope of this issue only cover config files under tests/ or also mirrord/layer/tests?

We should do the same for both places, but it's fine to only do one of them in this PR if you prefer that.

> - Most of the tests are failing for me locally and differ on number of fails per run with the exact same code (i.e. 50 on first run, 60 on the next) - is this expected behaviour?

I don't think so, did you follow the steps from the [contributing guide](https://github.com/metalbear-co/mirrord/blob/main/CONTRIBUTING.md#e2e-tests)?

Would you like to tell us about your plan for your next steps with this PR, so that we can give you some helpful feedback?

parames0:
@t4lz I found out my env setup with the dependencies (i.e. python and node libs) were faulty so I fixed most of those and got more tests passing now (~45). 

I can do all the config files once I sort this out. Thank you!

tests/src/utils.rs
```diff
@@ -59,6 +61,28 @@ fn format_time() -> String {
...
+    }
+  }
+}
+
+/// Returns tempfile path
+pub fn json_to_path(config: Value) -> PathBuf {
+    let tempfile = json_to_tempfile(config);
+    let path = tempfile.path();
+    PathBuf::from(path)
+}
```
t4lz: This is a bug :)
When `tempfile` goes out of scope, the file is deleted from the filesystem.
So this function returns a path that no longer exists.

tests/Cargo.toml
```diff
@@ -2,7 +2,16 @@
...
+authors.workspace = true
+description.workspace = true
+documentation.workspace = true
+readme.workspace = true
+homepage.workspace = true
+repository.workspace = true
 license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+publish.workspace = true
```
t4lz, line 14: Let's not change this right now.

changelog.d/3308.added.md
t4lz, line 1: Since this change is only in the tests, and isn't interesting for users, the category should be `internal`, not `added` - so `3308.internal.md`.

tests/src/cli/sanity.rs
```diff
@@ -36,6 +39,38 @@ mod cli {
...
+        });
+        let config_path = json_to_path(config_json);
+
+        let mut process = run_verify_config(Some(vec![
+            "--ide",
+            config_path.to_str().expect("Valid config path!"),
+        ]))
+        .await;
+
+        assert!(process.wait().await.success());
```
t4lz, line 54: The verify-config command exits with exit code 0 even when the verification failed (I don't know if that's the right way to go, but let's leave that question aside for a moment).
So this test would be more meaningful if we also verified the output of the command. (I know the existing test doesn't do that.)

tests/src/cli/sanity.rs
```diff
@@ -36,6 +39,38 @@ mod cli {
...
+    /// Tests `verify-config` with tempfile and `--ide` args, which should be:
+    ///
+    /// ```sh
+    /// mirrord verify-config --ide /path/to/config.json
+    /// ```
+    #[cfg_attr(target_os = "windows", ignore)]
+    #[rstest]
+    #[tokio::test]
+    #[timeout(Duration::from_secs(30))]
+    pub async fn tempfile_ide_verify_config() {
```
t4lz, line 43: since we want to eliminate config files in the repo, we can change the existing test to use a tempfile, we don't need to add a new test in addition to the old test that uses a config file from the repo.

tests/src/cli/sanity.rs
parames0: @t4lz Thanks for the catch! First time dealing with temp files in Rust and I totally forgot about the Drop and went with the passing test

tests/src/utils.rs
parames0, line 1: @t4lz this persists to the fs but drops the NamedTempfile so I opted to remove the file by path after use in respective test cases. if you think this is good, can continue with the other config files.

also, well aware the test I changed in traffic/steal.rs is failing now... that's a WIP

.vscode/launch.json
t4lz, line 1: Why delete this file?

tests/src/cli/sanity.rs
```diff
@@ -36,6 +42,47 @@ mod cli {
...
+                "env": true
+            }
+        });
+        let config_path = json_to_path(config_json);
+
+        println!("Tempfile path is: {:?}", &config_path);
+
+        if config_path.exists() == false {
+            assert!(false);
+        }
```
t4lz:
```suggestion
        assert!(config_path.exists());
```

tests/src/utils.rs
```diff
@@ -59,6 +61,30 @@ fn format_time() -> String {
...
 }
 
+/// Returns tempfile path
+pub fn json_to_path(config: Value) -> PathBuf {
+    let tempfile = NamedTempFile::new();
+    match tempfile {
+      Ok(mut tf) => {
+        let json_string = serde_json::to_string(&config).unwrap_or_default();
+        let _ = write!(tf, "{}", json_string);
+        let tf_keep = tf.keep();
```
t4lz: Keeping the file beats the purpose of using a tempfile. Requiring the caller to explicitly delete the temporary file is not a very nice interface. How about just returning the `NamedTempFile`? Or if you think that type is not nice enough for the callers to work with, we can wrap it in our own type.

parames0: @t4lz wrapping NamedTempfile is what I started thinking about today too. impl for one method to return pathbuf and the other to cleanup

.vscode/launch.json
parames0: sorry didn't see that, will bring back up

mirrord/layer/tests/bash.rs
```diff
@@ -28,7 +30,16 @@ async fn bash_script(dylib_path: &Path, config_dir: &Path) {
...
+    let config_json = json!({
+        "feature": {
+            "fs": {
+                "mode": "read",
+                "local": ["/var/db/"]
+            }
+        }
+    });
+    let tempfile = ManagedTempFile::new(config_json);
+    config_path.push(&tempfile.path);
```
t4lz, line 42: We don't need the `config_path` `PathBuf` created from `config_dir`. We don't need `config_dir`. That's the path of the directory containing the config files. But we're not using those config files anymore. You can just use `tempfile.path` directly.

t4lz: Same in all of the other places as well.

tests/src/utils.rs
```diff
@@ -59,6 +61,46 @@ fn format_time() -> String {
...
+            }
+          },
+          Err(e) => {
+            panic!("Error generating tempfile: {}", e);
+          }
+        }
+    }
+}
+
+impl Drop for ManagedTempFile {
```
t4lz, line 96: Instead of `keep`ing the temporary file, and implementing `Drop` on our own, we should just keep the `NamedTempFile`, that already has RAII semantics and deletes the temp file when dropped.

tests/Cargo.toml
t4lz: I mean don't add these lines to this file (the comment above was marked as resolved but the lines are still there).

.gitignore
```diff
@@ -26,8 +26,8 @@ build/
 
 ### VSCode ###
 vscode-ext/bin
-.vscode/settings.json
-.vscode/launch.json
+!.vscode/settings.json
+!.vscode/launch.json
```
t4lz, line 30: Let's not change this right now.

tests/src/utils.rs
```diff
@@ -59,6 +61,46 @@ fn format_time() -> String {
...
 }
 
+pub struct ManagedTempFile {
+    pub file: File,
+    pub path: PathBuf
+}
+
+impl ManagedTempFile {
+    /// Returns tempfile path
+    pub fn new(config: Value) -> Self {
```
t4lz, line 71: Taking a `Value` here seems to be inconvenient (it made you use `LazyLock` and made the calls to this functions a bit less simple).
How about we just take a string slice.
We can add `let _: serde::de::IgnoredAny = serde_json::from_str(data).unwrap();` in this function if we want to have verification that the config is a valid json.

PR #3836, opened 2026-01-03
Make `http_filter.ports`/`incoming.ports` more intuitive

MintSoup:
Now we'll no longer have hidden defaults (`[80, 8080]` stuff etc), and both `http_filter.ports` and `incoming.ports` will function by the same simple logic:
* When set, filtering/stealing will only apply to the ports in the list.
* When not set, filtering/stealing will apply to all (eligible) ports, i.e. "default to `[1, 2, 3, ..., 65535]`

Same algorithm in pseudo-rust:
```rust
fn on_bind(port: u16) {
    if let Some(remote_ports) = incoming.ports && remote_ports.contains(port).not() {
        bind_locally(port);
        return;
    }

    let Some(filter) = incoming.http_filter else {
        bind_remote_unfiltered(port);
        return;
    }

    if let Some(filtered_ports) = filter.ports && filtered_ports.contains(port).not() {
        bind_remote_unfiltered(port)
    } else {
        bind_remote_filtered(port, incoming.http_filter)
    }
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5921,28 +5921,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "pnet_sys"
-version = "0.35.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d4643d3d4db6b08741050c2f3afa9a892c4244c085a72fcda93c9c2c9a00f4b"
-dependencies = [
- "libc",
- "winapi",
-]
-
-[[package]]
-name = "pnet_transport"
-version = "0.35.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f604d98bc2a6591cf719b58d3203fd882bdd6bf1db696c4ac97978e9f4776bf"
-dependencies = [
- "libc",
- "pnet_base",
- "pnet_packet",
- "pnet_sys",
-]
-
-[[package]]
 name = "polling"
 version = "3.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/mirrord/layer/tests/filter_ports.rs
+++ b/mirrord/layer/tests/filter_ports.rs
@@ -132,7 +132,7 @@ async fn filter_ports(
         .unwrap();
 
     let mut ctx = ConfigContext::default();
-    let config_parsed = LayerFileConfig::from_path(&config_file)
+    let config_parsed = LayerFileConfig::from_path(&config_file, &mut ctx)
         .unwrap()
         .generate_config(&mut ctx)
         .unwrap();


changelog.d/+incoming-and-filter-ports-unification.changed.md
```diff
@@ -0,0 +1 @@
++ Unified `http_filter.ports` and `incoming.ports` behavior and made it more intuitive. Now, ports will be stolen unfiltered/not stolen when `http_filter.ports`/`incoming.ports` is set AND the port in question is not in the list.
```
gememma:
```suggestion
Unified `http_filter.ports` and `incoming.ports` behavior and made it more intuitive. Now, ports will be stolen unfiltered/not stolen when `http_filter.ports`/`incoming.ports` is set AND the port in question is not in the list.
```

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -175,18 +170,17 @@ pub struct HttpFilterConfig {
 
     /// ##### feature.network.incoming.http_filter.ports {#feature-network-incoming-http_filter-ports}
     ///
-    /// Activate the HTTP traffic filter only for these ports.
+    /// Activate the HTTP traffic filter only for these ports. When
```
gememma, line 173: don't forget to update [config docs](https://github.com/metalbear-co/docs-configuration) after this change is merged :)

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -273,14 +267,16 @@ impl HttpFilterConfig {
...
-            None
-        }
+    /// Returns `None` when we have no HTTP filter set.
+    ///
+    /// Returns `Some(None)` when we have an HTTP filter set, but
+    /// `http_filter.ports` is unset, i.e. we filter all ports.
+    ///
+    /// Returns `Some(&[u16])` when we have an HTTP filter set, AND
+    /// `http_filter.ports` is set, i.e. we filter only the selected
+    /// ports.
```
gememma: might be worth writing a type other than `Option` to use here? just so the behaviour is very clear from the type? may be overkill tho... something with three variants like:
```
enum Thingymajig {
    NoFilter, // instead of None
    FilterAllPorts, // instead of Some(None)
    FilterSomePorts(&[u16]) // instead of Some(Some)
}
```
up to you 🤷 would need a little bit of converting from option stuff

mirrord/config/src/feature/network/incoming/http_filter.rs
MintSoup: I thought it was autogenerated :skull:

mirrord/config/src/feature/network/incoming/http_filter.rs
MintSoup: it's only used for analytics stuff to count # of ports being filtered, I think I can just do that here

mirrord/config/src/feature/network/incoming/http_filter.rs
gememma: 😬

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -175,18 +170,10 @@ pub struct HttpFilterConfig {
...
-    /// We check the pod's health probe ports and automatically add them here, as they're
-    /// usually the same ports your app might be listening on. If your app ports and the
-    /// health probe ports don't match, then setting this option will override this behavior.
-    ///
-    /// Set to [80, 8080] by default.
+    /// Activate the HTTP traffic filter only for these ports. When
+    /// absent, filtering will be done for all ports.
     #[config(env = "MIRRORD_HTTP_FILTER_PORTS")]
-    pub ports: Option<PortList>,
+    pub ports: Option<VecOrSingle<u16>>,
```
0x00A5, line 176: Still need to run `cargo run -p medschool -- --input ./mirrord/config/src --output ./mirrord/config/configuration.md` and check in the updated configuration.md.

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -273,13 +260,16 @@ impl HttpFilterConfig {
...
     }
 
-    pub fn get_filtered_ports(&self) -> Option<&[u16]> {
-        if let Some(ports) = self.ports.as_ref()
-            && self.is_filter_set()
-        {
-            Some(&*ports.0)
+    /// Returns the number of ports that get filtered.
+    pub fn count_filtered_ports(&self) -> u16 {
+        if self.is_filter_set() {
```
0x00A5: Is this wrong? If filter is NOT set, the count=0.

mirrord/config/src/feature/network/incoming.rs
```diff
@@ -532,46 +531,6 @@ impl IncomingConfig {
...
-    /// ports `[80, 8080]`.
-    ///
-    /// Usually the user app will be listening on HTTP on the same ports as these probes, so
-    /// we can insert them in the user config.
-    ///
-    /// If the user has set anything in [`HttpFilterConfig::ports`], then we do nothing, to
-    /// avoid overriding their config. We also take care to not create conflicts with other
-    /// port configs that we have, such as [`IncomingConfig::ignore_ports`]`, and
-    /// [`IncomingConfig::ports`].
-    pub fn add_probe_ports_to_http_filter_ports(
```
0x00A5, line 546: Huh, this is interesting. So we no longer need to deal with probe ports because a) if filter ports is some, we use those ports (same as before), and b) if filter is none, we filter all ports anyways, which include probe ports. Is my understanding right?

mirrord/layer/src/setup.rs
```diff
@@ -344,10 +343,14 @@ impl IncomingMode {
             let steal_type = match &self.http_settings {
                 None => StealType::All(port),
                 Some(settings) => {
-                    if settings.ports.contains(&port) {
-                        StealType::FilteredHttpEx(port, settings.filter.clone())
-                    } else {
+                    if settings
```
0x00A5, line 346: Looks like we don't check if a http filter is set because it is always set at this point. If the user uses steal and do not set any http filter, where do we catch that before this?

mirrord/config/src/util.rs
```diff
@@ -155,6 +155,29 @@ pub enum VecOrSingle<T> {
     Multiple(Vec<T>),
 }
 
+impl<T> VecOrSingle<T> {
+    pub fn push(&mut self, new: T) {
```
0x00A5: Is this method used anywhere?

mirrord/config/src/feature/network/incoming/http_filter.rs
MintSoup: Sure, there's also a PR on the config docs for the same change

mirrord/config/src/feature/network/incoming/http_filter.rs
MintSoup: oops

mirrord/config/src/feature/network/incoming.rs
MintSoup: Yup

mirrord/layer/src/setup.rs
MintSoup: Not sure if I understand the question, but we match on `self.http_settings` and return `StealType::All` if it's `None`.

mirrord/config/src/util.rs
MintSoup: Doesn't look like it, I think I just rewrote it from the previous impl without checking if its gonna be used. Removing.

PR #3837, opened 2026-01-05
INT-253 DB Branching (PostgreSQL) support for IAM DB Authentication

vladrbg:


mirrord/config/src/feature/database_branches/pg.rs
```diff
@@ -13,6 +13,84 @@ pub struct PgBranchConfig {
...
+/// ```
+#[derive(Clone, Debug, Eq, PartialEq, JsonSchema, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum PgIamAuthConfig {
+    /// AWS RDS/Aurora IAM authentication.
+    /// The init container must have AWS credentials (via IRSA, instance profile, or env vars).
+    AwsRds {
+        /// AWS region. If not specified, uses AWS_REGION or AWS_DEFAULT_REGION.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        region: Option<ConnectionSourceKind>,
```
DmitryDodzin: Since we are using `ConnectionSourceKind` for more than just the connection a change of name for the variable might be in order, not critical just would be nice.

mirrord/operator/src/client/database_branches.rs
```diff
@@ -467,6 +467,59 @@ impl PgBranchParams {
                 }),
             },
         };
+        // Helper to convert config ConnectionSourceKind to CRD ConnectionSourceKind
+        fn convert_source_kind(src: &ConnectionSourceKind) -> CrdConnectionSourceKindPg {
```
DmitryDodzin: Instead of doing this I would add a `From<mirrord_config::feature::database_branches::ConnectionSourceKind>` implementation so it would be easy to come by if anyone else needs this conversion

mirrord/operator/src/client/database_branches.rs
```diff
@@ -467,6 +467,59 @@ impl PgBranchParams {
...
+                    "Converting GcpCloudSql config"
+                );
+                CrdIamAuthConfig::GcpCloudSql {
+                    credentials_json: credentials_json.as_ref().map(convert_source_kind),
+                    credentials_path: credentials_path.as_ref().map(convert_source_kind),
+                    project: project.as_ref().map(convert_source_kind),
+                }
+            }
+        });
+        tracing::info!(?iam_auth, "Converted IAM auth for CRD");
```
DmitryDodzin: We don't do a lot of info logs, mainly debug or trace, if we want to say something to the user than using the progress bar is the way

mirrord/operator/src/client/database_branches.rs
```diff
@@ -467,6 +467,59 @@ impl PgBranchParams {
...
+                    variable,
+                } => CrdConnectionSourceKindPg::EnvFrom {
+                    container: container.clone(),
+                    variable: variable.clone(),
+                },
+            }
+        }
+
+        // Convert IAM auth config if present
+        tracing::info!(?config.iam_auth, "Converting IAM auth config from user config to CRD");
```
DmitryDodzin: We don't do a lot of info logs, mainly debug or trace, if we want to say something to the user than using the progress bar is the way (same as bellow)

mirrord/operator/src/client/database_branches.rs
```diff
@@ -467,6 +467,59 @@ impl PgBranchParams {
...
+        let iam_auth = config.iam_auth.as_ref().map(|iam| match iam {
+            PgIamAuthConfig::AwsRds {
```
DmitryDodzin: Same here, would be nicer to add `From<PgIamAuthConfig>` to `CrdIamAuthConfig`

mirrord/operator/src/client/database_branches.rs
```diff
@@ -467,6 +467,59 @@ impl PgBranchParams {
...
+                access_key_id: access_key_id.as_ref().map(convert_source_kind),
+                secret_access_key: secret_access_key.as_ref().map(convert_source_kind),
+                session_token: session_token.as_ref().map(convert_source_kind),
+            },
+            PgIamAuthConfig::GcpCloudSql {
+                credentials_json,
+                credentials_path,
+                project,
+            } => {
+                tracing::info!(
```
DmitryDodzin: this is more of a `debug!` message

mirrord/operator/src/crd/pg_branching.rs
```diff
@@ -39,6 +39,55 @@ pub struct PgBranchDatabaseSpec {
...
+    pub iam_auth: Option<IamAuthConfig>,
+}
+
+/// IAM authentication configuration for connecting to cloud-managed databases.
+/// Environment variable sources follow the same pattern as `connection.url`:
+/// - `Env` - direct env var from pod spec
+/// - `EnvFrom` - from configMapRef/secretRef
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum IamAuthConfig {
```
DmitryDodzin, line 57: Do we have to duplicate? can't we use the `PgIamAuthConfig` directly?

mirrord/config/src/feature/database_branches/pg.rs
```diff
@@ -13,6 +13,84 @@ pub struct PgBranchConfig {
...
+///     "credentials_json": { "type": "env_from", "variable": "GOOGLE_APPLICATION_CREDENTIALS_JSON" }
+///   }
+/// }
+/// ```
+#[derive(Clone, Debug, Eq, PartialEq, JsonSchema, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum PgIamAuthConfig {
+    /// AWS RDS/Aurora IAM authentication.
+    /// The init container must have AWS credentials (via IRSA, instance profile, or env vars).
+    AwsRds {
```
DmitryDodzin, line 55: We might need to also support `AWS_SHARED_CREDENTIALS_FILE` like the GCP one, though I doubt it is needed for first version just FIY

mirrord/config/src/feature/database_branches/pg.rs
vladrbg: Any suggestions ? :D

mirrord/operator/src/crd/pg_branching.rs
vladrbg: I think yes because they are the same but ConnectionSourceKind is different.

mirrord/config/src/feature/database_branches/pg.rs
vladrbg: Yep got it will revisit when needed, I have prepared linear issues for other connected db branching stuff but need to to them a bit in order so this can be added as well.

mirrord/config/src/feature/database_branches/pg.rs
DmitryDodzin: `TargetEnviromentVariableSource` maybe, or something similar

PR #3838, opened 2026-01-05
INT-233 DB Branching - support for Mongo db 

vladrbg:


mirrord/operator/src/crd/mongodb_branching.rs
```diff
@@ -0,0 +1,172 @@
...
+    group = "dbs.mirrord.metalbear.co",
+    version = "v1alpha1",
+    kind = "MongodbBranchDatabase",
+    status = "MongodbBranchDatabaseStatus",
+    namespaced
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MongodbBranchDatabaseSpec {
+    /// ID derived by mirrord CLI.
+    pub id: String,
```
DmitryDodzin, line 28: How much different it is from `metadata.name`?

mirrord/operator/src/crd/mongodb_branching.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Database connection info from the workload.
+    pub connection_source: ConnectionSource,
+    /// MongoDB database name.
+    pub database_name: Option<String>,
+    /// Target k8s resource to extract connection source info from.
+    pub target: Target,
+    /// The duration in seconds this branch database will live idling.
+    pub ttl_secs: u64,
+    /// MongoDB server image version, e.g. "7.0".
+    pub mongodb_version: Option<String>,
```
DmitryDodzin, line 38: An interesting feature we can add is a warning to the user if they are using an affected version of mongo-bleed

mirrord/operator/src/crd/mongodb_branching.rs
```diff
@@ -0,0 +1,172 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum ConnectionSource {
+    /// A complete connection URL.
+    Url(ConnectionSourceKind),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum ConnectionSourceKind {
```
DmitryDodzin, line 53: I think we need to consolidate these types, it's the 3rd one now no?

vladrbg: Oh for sure, there are couple places where I am planning to change, but i want to first have. Mongo Db and then convert Mysql to use binary as well and when all 3 are actually the same i want to see what can be done to reduce duplication and create some common utils between them and so on, that's why I try to keep things similar for so all 3 are working the same and go from there.

mirrord/operator/src/crd/mongodb_branching.rs
vladrbg: But from what I saw pretty much all versions are affected 😂

mirrord/operator/src/crd/mongodb_branching.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MongodbBranchDatabaseStatus {
+    pub phase: BranchDatabasePhase,
+    /// Time when the branch database should be deleted.
+    pub expire_time: MicroTime,
+    /// Information of sessions that are using this branch database.
+    #[serde(default)]
+    pub session_info: HashMap<String, SessionInfo>,
```
DmitryDodzin, line 75: Do you need it or can you use ownership ref to keep what sessions are connected to the split?

vladrbg: Hmm will check, but again similar to some of the other comments, pretty much reusing the same functionality we have for the other branches Pg/MySQL to keep consistent, so maybe we can do this at the stage where we do the improvements for common functionality etc.

DmitryDodzin (approved):
LGTM

PR #3839, opened 2026-01-05
Fix premature port unsubscriptions

MintSoup:
It is possible to create multiple handles (file descriptors/`SOCKET`s) to the same underlying socket. In our `close` hook impl, we delete the port subscription as soon as `close` is called on one of them, but the correct behavior would be to wait for all duplicates to be `close`d before deleting the port subscription.

### DIFF
--- a/changelog.d/+duplicate-socket-handles.fixed.md
+++ b/changelog.d/+duplicate-socket-handles.fixed.md
@@ -1,0 +1,1 @@
+Ports will now be unsubscribed only when `close.2` has been called on all handles referring to the matching socket.


mirrord/layer-lib/src/socket.rs
```diff
@@ -213,7 +213,7 @@ impl UserSocket {
     /// Closes the socket and performs necessary cleanup.
     /// If this socket was listening and bound to a port, notifies agent to stop
     /// mirroring/stealing that port by sending PortUnsubscribe.
-    pub fn close(&self) {
+    pub fn close(&self) -> HookResult<()> {
```
itsamegraf, line 216: great catch and fix, windows side approved

0x00A5 (approved):

PR #3842, opened 2026-01-06
Agent allows metrics to be queried by the operator.

meowjesty:
- Issue: https://github.com/metalbear-co/operator/issues/1014

Adds a protocol message that collects the agent metrics, so we can send them to the operator on a `ClientMetrics` request.


.helix/languages.toml
iniw, line 1: Committed on accident? Also, you can add `.helix/` to your global ignore in `~/.config/git/ignore` - that's what I do

meowjesty: Could swear I had added it to mirrord gitignore actually.

PR #3845, opened 2026-01-08
dlopen many cgo libs

0x00A5:
This PR implements go detours for go v1.24.*. These detours are used only for Linux amd64 dlopen hook.
The go detours use the same approach in our 1.25 detours, but do not use the `gosave_systemstack_switch` function resolved from the go binary. Instead, the whole call stack is implemented with Rust naked functions.
This approach shall handle edge cases when multiple c-shared go libs are opened.


mirrord/layer/src/go/c_shared/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+    unsafe extern "C" fn gosave_systemstack_switch() {
+        naked_asm!(
+            // TODO: In Go's implementation, it loads the address of `runtime.systemstack_switch` + 8 bytes.
+            // [`runtime.systemstack_switch`](https://github.com/golang/go/blob/go1.24.11/src/runtime/asm_amd64.s#L475)
+            // is a dummy marker function. If it is directly called, it will hit a `ud2` trap.
+            // Go only cares that g->sched->pc is set to an address between the prologue and epilogue.
+            //
+            // I don't know if it matters for the address to be between the actual `systemstack_switch` function.
+            // Or a simple dummy function will work?
+            "lea    r9, [rip + {dummy} + 0x8]",
```
0x00A5, line 201: Context from go source code comments:
* https://github.com/golang/go/blob/go1.24.11/src/runtime/asm_amd64.s#L468
* https://github.com/golang/go/blob/go1.24.11/src/runtime/asm_amd64.s#L817

Questions:
1. does go care if the address is actually in the middle of the systemstack_switch function? Or a dummy function with the same stack frame layout will work?
2. if so, how does it determine that?
3. in my test, I can give random address to it, and it doesn't crash. So go must only check it in certain occasions. what are these occasions?
4. what will go do if the address is not actually in systemstack_switch?

mirrord/layer/tests/apps/dlopen_cgo/build_test_app.sh
```diff
@@ -4,15 +4,24 @@ set -euo pipefail
...
 CPP_FILE="$SCRIPT_DIR/main.cpp"
-SO_FILE="$SCRIPT_DIR/libgo_server.so"
 OUT_BIN="$SCRIPT_DIR/out.cpp_dlopen_cgo"
 
 echo "Script directory: $SCRIPT_DIR"
 
 # Ensure files exist
-if [[ ! -f "$GO_FILE" ]]; then
+if [[ ! -f "$SERVER_GO_FILE" ]]; then
+    echo "ERROR: main.go not found in $SCRIPT_DIR"
```
Razz4780:
```suggestion
    echo "ERROR: server/main.go not found in $SCRIPT_DIR"
```

mirrord/layer/tests/apps/dlopen_cgo/build_test_app.sh
```diff
@@ -4,15 +4,24 @@ set -euo pipefail
...
+    exit 1
+fi
+
+if [[ ! -f "$FILEOPS_GO_FILE" ]]; then
     echo "ERROR: main.go not found in $SCRIPT_DIR"
```
Razz4780:
```suggestion
    echo "ERROR: fileops/main.go not found in $SCRIPT_DIR"
```

mirrord/layer/tests/apps/dlopen_cgo/build_test_app.sh
```diff
@@ -22,13 +31,18 @@ if [[ ! -f "$CPP_FILE" ]]; then
...
 fi
 
-echo "Building Go shared library..."
-go build -buildmode=c-shared -o "$SO_FILE" "$GO_FILE"
+go version
+echo "Building Go shared server library..."
+go build -buildmode=c-shared -o "$SERVER_SO_FILE" "$SERVER_GO_FILE"
+
+echo "Building Go shared file ops library..."
+go build -buildmode=c-shared -o "$FILEOPS_SO_FILE" "$FILEOPS_GO_FILE"
```
Razz4780: Shouldn't we test shared libs from different Go versions? Was that the user's case?

mirrord/layer/tests/apps/dlopen_cgo/main.cpp
```diff
@@ -28,34 +30,51 @@ string get_exe_dir() {
...
     string exe_dir = get_exe_dir();
-    string so_path = exe_dir + "/libgo_server.so";
+    string server_so_path = exe_dir + "/server/libgo_server.so";
+    string fileops_so_path = exe_dir + "/fileops/libgo_fileops.so";
 
-    // Load library
-    void* handle = dlopen(so_path.c_str(), RTLD_LAZY);
-    if (!handle) {
+	// These dlopen() flags are provided by the user.
+    void* server_so_handle = dlopen(server_so_path.c_str(), RTLD_LAZY | RTLD_NODELETE | RTLD_DEEPBIND);
```
Razz4780: Huh. `RTLD_NODELETE` makes me thing that maybe they're loading and unloading the same library multiple times 👀 
We should test it probably

mirrord/layer/src/go/c_shared/mod.rs
Razz4780: Huh, it's been a while 💀 

I did *some* research, and looks like there's nothing in the runtime that explicitly compares addresses against systemstack_switch.

AI was hallucinating real hard, but in the end said this:


>Final, precise conclusion
Using only the Go 1.24.12 source you provided:
runtime.systemstack_switch exists
It is not used by:
stack walking
traceback
GC
scheduler
FuncID logic
It serves as a last-resort, non-return safety stub
It is not a sentinel in the logical sense

and

>It simply exists so that undefined behavior becomes predictable behavior if something goes catastrophically wrong.

Can we believe it? No. But imo we can try rolling out this change, it won't get worse than it already is 😆

mirrord/layer/tests/apps/dlopen_cgo/main.cpp
Razz4780: Huh. Also `RTLD_LAZY` looks fishy as hell :thinking: 
Not sure why it even works with `RTLD_LAZY` right now. Not sure how exactly layer "hooks" the symbol either (does it overwrite pointer in the symbol table? does it overwrite instructions at the start of the hooked function itself?). But `RTLD_LAZY` sounds like there are other trampolines and function replacements in the game.

mirrord/layer/tests/apps/dlopen_cgo/build_test_app.sh
0x00A5: The user uses same version v1.24.11. I also found that `(gdb) info functions Syscall6` only returns one `internal/runtime.syscall.Syscall6` despite that this test app load two libs. And it always returns the one from the first lib the app dlopen'd (I flipped the order of my dlopen calls to verify). So, I am not sure what will happen if different go versions are used 💀

mirrord/layer/tests/apps/dlopen_cgo/main.cpp
0x00A5: > RTLD_NODELETE

Yes, worth asking the user about this in my call with them next week. `dlclose()` will unload the go runtime code, maybe they don't want that to happen.

> RTLD_LAZY

The one we hook is from static library, thus the symbols are bind already when the go lib is compiled? This flag affects all functions in dynamic library I think.

mirrord/layer/src/go/c_shared/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+            // Store r9 in g->sched->pc.
+            "mov    QWORD PTR [r14+0x40],r9",
+            "lea    r9, [rsp+0x8]",
+            // Store r9 in g->sched->sp.
+            "mov    QWORD PTR [r14+0x38],r9",
+            // Store r9 in g->sched->ret.
+            "mov    QWORD PTR [r14+0x58],0x0",
+            // Store r9 in g->sched->bp.
+            "mov    QWORD PTR [r14+0x68],rbp",
```
0x00A5: Oops these comments are wrong

mirrord/layer/src/go/c_shared/mod.rs
0x00A5: Thanks for taking a look at this 🙏

PR #3846, opened 2026-01-08
Tiny refactor to make a struct field name not be a lie

t4lz:
The field `empty_target_final` of the `ConfigContext` struct was holding exactly the opposite value of what its name suggest. It was `true` when the target **wasn't** final and `false` when it was.

I just changed all its usages to hold the value that its name says it holds. I changed both the writing to and the reading from that value, so there is no change in behaviour or functionality.


iniw (approved):
:not:

PR #3848, opened 2026-01-08
Allow concurrent ci start commands.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1106/allow-to-run-concurrent-ci-start-commands

Refactors `mirrord ci start` and `mirrord ci stop` to allow running concurrent sessions. Now we keep a list (`HashSet`) of pids (for user processes and intproxies), instead of just 1 pid for each, and when `mirrord ci stop` is called, we kill every pid from that list.

### DIFF
--- a/changelog.d/+duplicate-socket-handles.fixed.md
+++ b/changelog.d/+duplicate-socket-handles.fixed.md
@@ -1,0 +1,1 @@
+Ports will now be unsubscribed only when `close.2` has been called on all handles referring to the matching socket.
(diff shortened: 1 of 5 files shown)


mirrord/cli/src/ci/stop.rs
```diff
@@ -36,9 +42,19 @@ impl CiStopCommandHandler {
             unistd::Pid,
         };
 
+        let Self {
+            store,
+            mut progress,
+        } = self;
```
iniw, line 48: 🤢

PR #3858, opened 2026-01-12
Fix flaky test

MintSoup:
Reason for flakiness is in comments

### DIFF
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -1538,7 +1538,7 @@ mod test {
         // Assert intproxy closes the connection.
         //
         // `read_to_end` reads until EOF, which is the *graceful*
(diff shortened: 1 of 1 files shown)


mirrord/intproxy/src/lib.rs
```diff
@@ -1534,8 +1535,31 @@ mod test {
 
         drop(to_proxy);
 
-        // Assert intproxy closed the connection
-        assert!(conn.read(&mut [0]).await.is_ok_and(|s| s == 0));
+        // Assert intproxy closes the connection.
+        //
+        // `read_to_end` reads until EOF, which is the *graceful*
+        // termination by of TCP connection by peer (a.k.a intproxy).
```
0x00A5:
```suggestion
        // termination of TCP connection by peer (a.k.a intproxy).
```

PR #3859, opened 2026-01-12
local http server is not notified when stolen http request

MintSoup:
In the agent, when channeling HTTP response frames from the intproxy to the http client, we use a `StreamBody<ReceiverStream<...>>` to transfer frames to the IO task. When the client closes the connection, this body object is dropped, but we don't react to it in any way and explicitly ignore any errors when trying to send more frames from the tx.

https://github.com/metalbear-co/mirrord/blob/3cb7e978dbc54b2b8bd654fc7ba8df89550ce5fe/mirrord/agent/src/incoming/connection/http.rs#L399

The result is that the intproxy and thus the client app (as in agent client, which is actually an http **server**) will have no idea that the connection has been terminated and will keep doing its thing.

This PR adds proper error handling to this flow, terminating the intproxy side of the connection when the client closes that of its own.

There is one potential problem with this approach, which is that the connection won't actually be terminated until the http server sends some data. That being said, I don't think this will cause issues since TCP already gives similarly loose guarantees anyway.

### DIFF
--- a/changelog.d/+connection-termination.fixed.md
+++ b/changelog.d/+connection-termination.fixed.md
@@ -1,0 +1,1 @@
+Agent will now terminate the intproxy side of HTTP connections that were closed by the remote HTTP client.


mirrord/agent/src/steal/api.rs
0x00A5, line 1: 👍 is it feasible to add an UT to cover this for stealer api?

MintSoup: Not without also adding a lot of supporting machinery :(

PR #3866, opened 2026-01-14
Add fix kubeconfig command

MintSoup:
Finds non-absolute paths in kubeconfig user exec fields, prompts the user and replaces them with absolute paths. Run with `mirrord fix kubeconfig`

### DIFF
--- a/mirrord/cli/src/fix.rs
+++ b/mirrord/cli/src/fix.rs
@@ -53,7 +53,8 @@ pub enum FixKubeconfigError {
     MissingUser { ctx: String, user: String },
 
     #[error(
-        "user {user}'s auth config uses {cmd_path:?}, which cannot be resolved using the current $PATH ({err})"
+        "user {user}'s auth config uses {cmd_path:?}, which cannot be resolved using the current $PATH ({err}).
+Ensure this command is run from a context where the parent directory of {cmd_path:?} is included in $PATH."
     )]
     CannotResolve {
         user: String,
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -407,10 +407,11 @@ pub(crate) enum CliError {
         mirrord failed to execute Kube authentication command.
         This can happen when the command is not specified using absolute path and cannot be found in $PATH in the context where mirrord is invoked.
         Possible fixes:
-        1. Change global $PATH to include the authentication command and relaunch the IDE/terminal.
-        2. In the kubeconfig, specify the command with an absolute path.{GENERAL_HELP}
+        1. Run `mirrord fix kubeconfig` from a terminal.
+        2. Change global $PATH to include the authentication command and relaunch the IDE/terminal.
+        3. In the kubeconfig, specify the command with an absolute path.{GENERAL_HELP}
     "))]
-    KubeAuthExecFailed(String),
+    KubeAuthExecFailed(io::Error),
 
     #[error("Failed while resolving target while using the mirrord-operator: {0}")]
     #[diagnostic(help(
@@ -508,8 +509,8 @@ impl CliError {
         use kube::{Error, client::AuthError};
 
         match error {
-            KubeApiError::KubeError(Error::Auth(AuthError::AuthExec(error))) => {
-                Self::KubeAuthExecFailed(error.to_owned())
+            KubeApiError::KubeError(Error::Auth(AuthError::AuthExecStart(error))) => {
+                Self::KubeAuthExecFailed(error)
             }
             // UGH(alex): Type-erased errors are messy, and this one is especially bad.
             // See `kube_service_error_dependency_is_in_sync` for a "what's going on here".
@@ -545,9 +546,12 @@ impl From<OperatorApiError> for CliError {
                 operation,
             } if code == StatusCode::FORBIDDEN => Self::OperatorApiForbidden(operation, message),
             OperatorApiError::KubeError {
-                error: Error::Auth(AuthError::AuthExec(error)),
+                error: Error::Auth(AuthError::AuthExecStart(error)),
                 ..
-            } => Self::KubeAuthExecFailed(error),
+            }
+            | OperatorApiError::KubeApi(KubeApiError::KubeError(Error::Auth(
+                AuthError::AuthExecStart(error),
+            ))) => Self::KubeAuthExecFailed(error),
             OperatorApiError::KubeError { error, operation } => {
                 Self::OperatorApiFailed(operation, error)
             }


mirrord/cli/src/config.rs
```diff
@@ -1204,6 +1207,24 @@ pub struct WizardArgs {
...
 }
 
+#[derive(Args, Debug)]
+pub struct FixArgs {
+    #[command(subcommand)]
+    pub command: FixCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum FixCommand {
```
meowjesty, line 1220: Docs pls.

mirrord/cli/src/config.rs
```diff
@@ -1204,6 +1207,24 @@ pub struct WizardArgs {
...
+    /// Look for non-absolute paths in kubeconfig and make them absolute
+    Kubeconfig(FixKubeconfig),
+}
+
+#[derive(Args, Debug)]
+pub struct FixKubeconfig {
```
meowjesty, line 1227: Docs pls.

mirrord/cli/src/config.rs
```diff
@@ -200,6 +200,9 @@ pub(super) enum Commands {
     /// configuration. Like `mirrord exec` it requires a connection to the cluster.
     #[cfg(feature = "wizard")]
     Wizard(Box<WizardArgs>),
+
+    /// Fix something related to mirrord
```
meowjesty:
```suggestion
    /// Fix issues related to mirrord.
```

Just a nit on the docs, "something" feels a bit too generic for `fix`.

mirrord/cli/src/config.rs
```diff
@@ -200,6 +200,9 @@ pub(super) enum Commands {
...
+    Fix(FixArgs),
```
meowjesty, line 205:
```suggestion
    Fix(Box<FixArgs>),
```

Might as well start with it `Box`ed, since we plan on adding more stuff to this command, right?

mirrord/cli/src/config.rs
```diff
@@ -1204,6 +1207,24 @@ pub struct WizardArgs {
     pub telemetry: bool,
 }
 
+#[derive(Args, Debug)]
+pub struct FixArgs {
+    #[command(subcommand)]
+    pub command: FixCommand,
```
meowjesty, line 1215: Docs pls

mirrord/cli/src/main.rs
```diff
@@ -325,6 +325,8 @@ mod wsl;
 #[cfg(feature = "wizard")]
```
meowjesty, line 333: Please also add a little bit of docs for the mod docs of `main.rs`.

Look at the example for `mirrord wizard` on line ~237

Cargo.toml
```diff
@@ -261,6 +261,10 @@ winapi = { version = "0.3.9", features = [
 # Used by `analytics`, `cli`
 ci_info = "0.14"
 
+# Used by `cli`
+yamlpatch = "0.9.0"
```
meowjesty, line 265: 5 years since the last update, do we really need this one?

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    #[error("absolute path {0:?} contains non-unicode characters")]
+    StringConversion(OsString),
+}
+
+async fn fix_kubeconfig(args: FixKubeconfig) -> Result<(), FixKubeconfigError> {
+    let kubeconfig_path = args.file_path.unwrap_or(
+        std::env::home_dir()
+            .ok_or(FixKubeconfigError::Home)?
+            .join(".kube/config"),
```
meowjesty, line 73: Shouldn't we also check for the path specified in the `KUBECONFIG` env var?

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+            continue;
+        };
+
+        let Some(exec) = &auth.exec else {
+            continue;
+        };
+
+        let Some(command) = &exec.command else {
+            continue;
+        };
```
meowjesty:
```suggestion
        if user.auth_info.and(auth.exec).and(exec.command).is_none() { continue ;}
```

Something like that, since I don't see you using `auth`, `exec`, or `command` anywhere? If you need to use them, you could `zip` instead.

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            match &*answer {
+                "y\n" => {
+                    break true;
+                }
+                "n\n" => {
+                    break false;
+                }
+                _ => {}
+            }
```
meowjesty: Isn't there a crate to do this?

Also, this could potentially be an arg, so you run the command once `mirrord fix kubeconfig` which just prints the error, and then in the error it says to actually fix it, you should run `mirrord fix kubeconfig --commit` or something.

Since this command will be run from the IDEs by the looks of it, I don't think we want this `stdin` wait for input approach.

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+            continue;
+        }
+
+        let absolute = match which::which(&cmd_path) {
+            Ok(a) => a
+                .into_os_string()
+                .into_string()
+                .map_err(FixKubeconfigError::StringConversion)?,
+            Err(err) => {
+                println!(
```
meowjesty: Instead of `print`, I think you should use `Progress`, so it can be used by the IDEs, if we want to show a notification box or warning box in those, the `Progress` is implemented to do just that.

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+                    "WARNING: user `{}` has an auth config that uses `exec` with non-absolute binary {cmd_path:?}, which cannot be resolved using the current $PATH ({err})",
+                    user.name
+                );
+                continue;
+            }
+        };
+
+        let replace = loop {
+            print!(
```
meowjesty: Same `Progress` consideration here.

Even though I don't think we want this stdin approach here.

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+        };
+
+        let cmd_path = PathBuf::from(&*command);
+
+        if cmd_path.is_absolute() {
+            continue;
+        }
+
+        let absolute = match which::which(&cmd_path) {
+            Ok(a) => a
```
meowjesty: pls avoid these single letter variable names.

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        let document = Document::new(tokio::fs::read_to_string(&kubeconfig_path).await?)?;
+
+        let patch = Patch {
+            route: route!("users", user_idx, "user", "exec", "command"),
+            operation: Op::Replace(Value::String(absolute)),
+        };
+
+        let patched = apply_yaml_patches(&document, &[patch])?;
+
```
meowjesty, line 144: I'm a bit scared of the 5 years of no update this crate brings.

Can't we index into the values we care about with `serde_yaml`, sort of like `document["users"]` and modify it like that?

mirrord/cli/src/fix.rs
```diff
@@ -0,0 +1,126 @@
...
+        tokio::fs::write(&kubeconfig_path, patched.source()).await?;
```
meowjesty: This command should probably also take a `--dry-run` arg, so we can show the user how we're going to be changing their kubeconfig.

meowjesty (changes requested):
I'm scared of `yamlpatch`, looked at the code and it looks alright-ish, but I wonder if we can't just `serde_yaml` (or whatever is the serde + yaml crate) to index into the document and make the changes we want.

mirrord/cli/src/config.rs
MintSoup: Tbh not sure, I'm not aware of any plans. Besides, we could just add it at any point down the line without major refactors, no?

mirrord/cli/src/fix.rs
MintSoup: `command` is used, can't use `zip` either because they're nested. It's either this or 
```rust
let Some(command) = user.auth_info.and_then(|a| a.exec).and_then(|a| a.command) else {
    continue;
};
```
or 
```rust
let Some(command): Option<_> = (try { user.auth_info?.exec?.command? }) else {
    continue;
};
```

IMO the `let`s are quicker to decipher while reading through the code ¯\_(ツ)_/¯

mirrord/cli/src/fix.rs
MintSoup: Does it? We explicitly tell the user what we're about to change and confirm with a prompt.

mirrord/cli/src/fix.rs
MintSoup: We do, it's just done by `clap` through `#[arg(env = ...)]`. It's `MIRRORD_KUBECONFIG` instead though which I too find weird but it's what the wizard was doing, maybe we should change both.

mirrord/cli/src/fix.rs
meowjesty: Hmm I kinda like the `try {}` more.

mirrord/cli/src/config.rs
meowjesty: Yeah it's fine either way, just might as well do it for every new command args thingy we add.

But if you don't want to, just resolve this comment. It's a nit.

Cargo.toml
meowjesty: Thought about this some more and its hmm, since the code itself looks fine, then add some unit tests to validate the kubeconfig change in a few cases, like, one case where we have to change, another case where we don't have to (so we can see that this doesn't break a valid file).

We should fork it, update the `serde_yaml` dependency of the crate (since `serde_yaml` has been deprecated, at some point we'll have to fork this crate to change it to use whatever serde yaml crate becomes the default for the rust community, so might as well do it now).

I've tried my hand at patching a yaml config in the past, using the `config.get("foo")?.get("baz")` approach and I know how annoying it becomes, so I think it's ok to keep this dep.

mirrord/cli/src/fix.rs
meowjesty: No I mean, `MIRRORD_KUBECONFIG` checks for an env var you can pass to mirrord, if you want to specify the kubeconfig dir, right?

What I was saying is, shouldn't we also check if there's an env var `KUBECONFIG`?

mirrord/cli/src/fix.rs
meowjesty: Alright, I really think that `mirrord fix kubeconfig` and `mirrord fix kubeconfig --dry-run` is a better UX than having it stop in the middle waiting for user input.

With this, you can use `Progress` instead of `print`s.

mirrord/cli/src/fix.rs
MintSoup: Not sure, I don't think other parts of mirrord react to `KUBECONFIG`, so I think to keep things consistent we can keep it the same?  That being said I think that just checking `KUBECONFIG` would be more intuitive.

Cargo.toml
MintSoup: > 5 years since the last update

?
Latest release was yesterday, first release in 2025.

PR #3882, opened 2026-01-19
Accept OTel context propagation carriers and forward to Operator

gememma:
Changes included:
- Added config fields for `traceparent` and `baggage` (alternatively we could add these under experimental or an otel block)
- Inject these values as headers into http requests made to the operator in `connect_target`

### DIFF
--- a/changelog.d/+otel-w3c-context-propagation.added.md
+++ b/changelog.d/+otel-w3c-context-propagation.added.md
@@ -1,1 +1,0 @@
-Accept OTel/W3C context propagation environment variables.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -170,6 +170,7 @@ jobs:
         run: |
           if [ "${{ steps.changed-config-schema.outputs.any_changed }}" == "true" ] && [ "${{ steps.changed-configuration-md.outputs.any_changed }}" != "true" ]; then
             echo "Error: mirrord configuration has changed but `configuration.md` documentation has not."
+            echo "Try running medschool (see `./medschool/readme.md`) and updating `configuration.md`."
             exit 1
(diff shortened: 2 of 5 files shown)


mirrord/operator/src/client.rs
```diff
@@ -156,8 +163,14 @@ impl fmt::Debug for OperatorSession {
...
             .field("operator_version", &self.operator_version)
-            .field("allow_reconnect", &self.allow_reconnect)
-            .finish()
+            .field("allow_reconnect", &self.allow_reconnect);
+        if let Some(traceparent) = &self.traceparent {
+            debug_struct.field("traceparent", traceparent);
+        }
+        if let Some(baggage) = &self.baggage {
+            debug_struct.field("baggage", baggage);
+        }
```
0x00A5: nit: why do we skip them if they are none. `operator_license_fingerprint` is also `Option<String>`, and we add it always. Looks like the whole point of having this manual debug impl is to avoid printing out user's private key in `client_cert`.

wizard-frontend/package-lock.json
0x00A5, line 1: Thank you for this 🙏  I have been git restore wizard for a while loll

0x00A5: Or is this a unintentional check in? since `jj` does it by default?

changelog.d/+otel-w3c-context-propagation.added.md
0x00A5, line 1: There are two change log news fragment in this PR. Can we merge them and make sure they still reflect the change we are doing?

wizard-frontend/package-lock.json
gememma: oops, didnt even realise that got in there lol

PR #3883, opened 2026-01-20
macOS connectx

0x00A5:
Added a bare minimum `connectx(2)` detour implementation.

Example call of the function by `nc`:
```
2026-01-20T02:54:13.454874Z TRACE ThreadId(01) connectx: mirrord_layer::socket::ops: close time.busy=7.73ms time.idle=39.7µs socket=5 endpoints=0x30485c560 associd=0 flags=0 iov=0x0 iovcnt=0 len=0x0 connid=0x30485c59c pid=99345
```
.NET 10 started using this C function in their [`TryStartConnect`](https://github.com/dotnet/runtime/blob/facd688567d1acf67f4264537e8c1917f5b2deeb/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs#L685-L690) API. And they actually send data with this function and save data length `sent`.

Detour logic is simple:
1. If source address is given, bind first.
2. Call connect with destination address next.

`connectx(2)` is supposed to copy data from `iov` to the socket's send buffer, and update bytes copied in `len`. The current implementation ignores `iov`, `iovcnt`, `flags` and simply set `len` to 0.

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -450,13 +450,14 @@ pub(super) fn listen(sockfd: RawFd, backlog: c_int) -> Detour<i32> {
     }
 }
 
-/// Implementation of BSD `connectx(2)` detour:
+/// Implementation of BSD `connectx(2)` detour.
 /// * If source address is present in `endpoints`, call [`bind`].
 /// * Call [`connect`] with the given destination address.
 /// * On success `connectx` is **supposed** to update `len` with the data length enqueued in the
 ///   socket's send buffer copied from `iov`. For simplicity, we always set `len` to 0 on success to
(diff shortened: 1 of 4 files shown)


0x00A5:
> Looks nice ^^
> 
> Have you checked what happens without mirrord if accosid and/or connid are invalid? Are they ignored, like in the implemented detour?

Copied from Slack:
nc  does this:
```
2026-01-20T02:54:13.454874Z TRACE ThreadId(01) connectx: mirrord_layer::socket::ops: close time.busy=7.73ms time.idle=39.7µs socket=5 endpoints=0x30485c560 associd=0 flags=0 iov=0x0 iovcnt=0 len=0x0 connid=0x30485c59c pid=99345
```
And libc allows `connid` to be non-null.

PR #3884, opened 2026-01-20
INT-283 Multi Cluster - Core + Authentication

vladrbg:


mirrord/operator/src/crd.rs
```diff
@@ -274,6 +279,56 @@ pub struct MirrordOperatorStatus {
 
     /// Option because added later.
     pub copy_targets: Option<Vec<CopyTargetEntryCompat>>,
+
+    /// Status of connected remote clusters (only on primary with multi-cluster enabled).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub connected_clusters: Option<Vec<ConnectedClusterStatus>>,
```
Razz4780:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub connected_clusters: Vec<ConnectedClusterStatus>,
```

mirrord/operator/src/crd.rs
```diff
@@ -274,6 +279,56 @@ pub struct MirrordOperatorStatus {
...
+}
+
+/// Status of a connected remote cluster.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ConnectedClusterStatus {
+    /// Logical name of the cluster.
+    pub name: String,
+    /// Timestamp of last health check (RFC3339 format).
+    pub last_check: String,
```
Razz4780: Can use `MicroTime`

mirrord/operator/src/client/error.rs
```diff
@@ -37,6 +38,7 @@ impl fmt::Display for OperatorOperation {
             Self::MongodbBranching => "mongodb branching",
             Self::MysqlBranching => "mysql branching",
             Self::PgBranching => "postgresql branching",
+            Self::MultiClusterSession => "multi-cluster session",
```
Razz4780: This is unused 🗑️

mirrord/operator/src/crd.rs
```diff
@@ -149,6 +149,10 @@ pub struct MirrordOperatorSpec {
     /// this field).
     #[deprecated(note = "use supported_features instead")]
     copy_target_enabled: Option<bool>,
+    /// Multi-cluster configuration (if enabled)
+    /// Contains information about whether this operator is part of a multi-cluster setup
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub multi_cluster: Option<MultiClusterConfig>,
```
Razz4780: This is unnecessary. However, we need two new `NewOperatorFeature`s:
1. Can be used as envoy
2. Can be used by the envoy

Razz4780 (changes requested):

mirrord/operator/src/crd.rs
```diff
@@ -274,6 +275,42 @@ pub struct MirrordOperatorStatus {
...
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum ClusterCheckResult {
+    /// Cluster is connected and responding.
+    Connected {
+        /// Operator version running on the remote cluster.
+        operator_version: String,
+        /// License fingerprint from the remote cluster (for license validation).
+        #[serde(skip_serializing_if = "Option::is_none")]
+        license_fingerprint: Option<String>,
```
Razz4780, line 321: Can you make sure these fields are camelCase as well? Not sure how `rename_all` works on enums with struct variants

PR #3886, opened 2026-01-21
win-38 >> implement mirrord-side features for ide support

cristeigabriela:
didnt check unix build/lint so its probably gonna fail for now. didnt do changes md yet. will open mirrord-vscode pr soon

took tracking down a crazy stupid bug 4-child-processes deep to unblock something mid-dev

implements "ide orchestrated" execution for windows, to do the process execution part, and mirrord ext is used to give the *user code* processes the mirrord target environment variables, and config file

that is also how we identify which process is user code :^)

needs nested execution support... well, that's fine.

starts with progress mode off to look as close to normal debugging as possible

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,7 +61,7 @@ jobs:
       - run: uv python install
       - run: uv venv
       - name: install towncrier
-        run: uv pip install towncrier==23.11.0
+        run: uv pip install towncrier==25.08.0
       - name: verify newsfragment exist
         run: uv run towncrier check
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,38 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
 
 <!-- towncrier release notes start -->
 
+## [3.183.0](https://github.com/metalbear-co/mirrord/tree/3.183.0) - 2026-01-23
+
+
+### Added
+
+- Add bypassed requests metric.
+- Added BSD `connectx(2)` detour.
+- Added `mirrord kubeconfig fix`, which finds non-absolute paths in kubeconfig
+  `exec` fields and interactively replaces them with absolute paths to make
+  them `$PATH`-indepdendent.
+- Db Branching support for MongoDB.
+- Experimental configuration to add artificial latency to outgoing connections
+
+
+### Changed
+
+- Improved error messages produced in case of mirrord-agent pod failure.
+- make /Program Files always local, helps when debugging Python in WSL
+
+
+### Fixed
+
+- Agent will no longer ignore trailing slashes when handling file open requests
+- Fixed bogus error logs printed by the agent on failed reverse DNS lookups.
+- Link to tree-sitter version that doesn't break release build.
+
+
+### Internal
+
+- Added an initial `CONTRIBUTING.md` guide to the mirrord-agent crate.
+- update latest tag on mirrord-cli container image
+
 ## [3.182.0](https://github.com/metalbear-co/mirrord/tree/3.182.0) - 2026-01-16
 
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2673,7 +2673,7 @@ dependencies = [
 
 [[package]]
 name = "fileops"
-version = "3.182.0"
+version = "3.183.0"
 dependencies = [
  "libc",
 ]
@@ -3834,7 +3834,7 @@ checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
 
 [[package]]
 name = "issue1317"
-version = "3.182.0"
+version = "3.183.0"
 dependencies = [
  "actix-web",
  "env_logger",
@@ -3844,7 +3844,7 @@ dependencies = [
 
 [[package]]
 name = "issue1776"
-version = "3.182.0"
+version = "3.183.0"
 dependencies = [
  "errno 0.3.14",
  "libc",
@@ -3853,7 +3853,7 @@ dependencies = [
 
 [[package]]
 name = "issue1776portnot53"
-version = "3.182.0"
+version = "3.183.0"
 dependencies = [
  "libc",
  "socket2 0.5.10",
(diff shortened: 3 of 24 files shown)


mirrord/cli/src/main.rs
```diff
@@ -357,6 +352,13 @@ async fn exec_process<P>(
 where
     P: Progress,
 {
+    // IDE-orchestrated mode: only propagate DLL injection without full mirrord setup.
+    // The IDE manages the actual mirrord environment.
+    #[cfg(target_os = "windows")]
+    if args.ide_orchestrated {
```
itsamegraf, line 363: are there any specific benefits of going through `exec` command? the current implementation introduces a "subvariant" of exec, i.e. `mirrord exec --ide-orchestrated`, but doesnt run any code of exec_process - you immediately call run_ide_orchestrated and return.
I suggest you introduce a new `ide` command (or whatever you choose to name it) in `config.rs` `Commands`, or just take over the `Commands::ExtensionExec` specifically for windows.

mirrord/cli/src/main.rs
itsamegraf, line 541: 👑

mirrord/cli/src/main.rs
```diff
@@ -543,13 +546,13 @@ async fn run_process_with_mirrord<P>(
...
     progress: P,
     analytics: &mut AnalyticsReporter,
-    _config: &LayerConfig,
 ) -> CliResult<()>
 where
     P: Progress,
 {
     // Let Windows handle executable resolution naturally
     // Don't force .exe extension - Windows will try .exe, .bat, .cmd, etc. automatically
+
```
itsamegraf, line 560: wouldnt lint cry about this? 😓

mirrord/layer-lib/src/ide.rs
```diff
@@ -0,0 +1,51 @@
+//! IDE integration detection utilities.
+//!
+//! This module provides functions to detect various IDE-related execution modes
+//! for the mirrord layer.
+
+/// Environment variable set when running in IDE-orchestrated mode.
+/// In this mode, the IDE manages the mirrord environment, so we only need to
```
itsamegraf, line 7: will remote env vars be fetched and presetn for the "mirrord -attached" process?

mirrord/layer-lib/src/ide.rs
```diff
@@ -0,0 +1,51 @@
...
+/// install process hooks for DLL propagation to child processes without full
+/// layer initialization.
+pub const MIRRORD_IDE_ORCHESTRATED_ENV: &str = "__MIRRORD_IDE_ORCHESTRATED";
+
+/// Environment variable that will be present only in the user code processes.
+pub const MIRRORD_EXT_INJECTED_ENV: &str = "__MIRRORD_EXT_INJECTED";
```
itsamegraf, line 18: please document where/who sets this environment variable

mirrord/layer-win/src/lib.rs
```diff
@@ -102,14 +130,51 @@ fn mirrord_start() -> LayerResult<()> {
     // Create layer initialization event first
     let init_event = LayerInitEvent::for_child()?;
 
+    // Check for IDE-orchestrated mode first.
+    // In this mode, we only install process hooks for DLL propagation
+    // without full layer initialization - the IDE manages the mirrord environment.
+    // Skip this mode if we're in the user process (the actual target application).
+    if is_ide_orchestrated() && !is_user_process() {
```
itsamegraf: can we please refrain from this code duplication with the code later in this function?

mirrord/cli/src/main.rs
```diff
@@ -1079,11 +1183,38 @@ fn main() -> miette::Result<()> {
...
+        }
+    }
+
+    #[cfg(windows)]
+    {
+        // Allow nesting when IDE-orchestrated mode with ext_injected
+        // This means the IDE is managing the session and this is the user process launch
+
+        use mirrord_layer_lib::ide::{is_ext_injected, is_ide_orchestrated};
+        if is_ext_injected() || is_ide_orchestrated() {
```
itsamegraf: Codex comment:
```
3) Nesting logic on Windows: is_ext_injected() || is_ide_orchestrated() is probably too permissive

Your Windows block says nesting is OK if either ext-injected OR ide-orchestrated is true.

But the comment says: “Allow nesting when IDE-orchestrated mode with ext_injected”.
That implies it should likely be AND, not OR.

Why it matters

__MIRRORD_IDE_ORCHESTRATED might be set in contexts where you still don’t want arbitrary nested mirrord exec.

__MIRRORD_EXT_INJECTED alone (without orchestrated) might also be set in places you still want to treat as nested misuse.

Recommendation

Make the condition match the intent: if is_ext_injected() && is_ide_orchestrated() { Ok(()) } else { Err(...) } (or an even tighter “this exact workflow” check).
```

mirrord/layer-lib/src/ide.rs
cristeigabriela: yes! mirrord ext retrieves them, and they are later given to the child process. you can assume they are there when `is_ext_injected` == true.

mirrord/layer-win/src/lib.rs
cristeigabriela: I was going to do this as well, I will update configuration to describe our intent for hooks, etc.

mirrord/cli/src/main.rs
cristeigabriela: It's because it follows the same large loop and syntax of exec:
- get process
- get process arguments
- extract layer
- start process
- wait until process dies


`ext` cannot be extended to support this, unless we implement the selfsame thing as below. but this is not the purpose of ext, the purpose of ext is to spin up agent, and serialize the state that the process will need to be given later in the ide extension

i felt it would be extra to implement a second `exec` command, and potentially more annoying to maintain? curious for your further thoughts on this

mirrord/layer-win/src/lib.rs
cristeigabriela: fixed

mirrord/cli/src/main.rs
itsamegraf: Thank you for the explanation, I understand your point better, 
My issue is mostly with the immediate `return run_ide_orchestrated` - I don't think the call to `run_ide_orchestrated` should be in `exec_process` if it doesnt share code with it (the code immediately returns).
`exec_process` is called only from `exec` in main.rs, the code there is the shared code between regular exec and `ide_orchestrated` exec, I suggest to move the `if ide_orchestrated: run_ide_orchestrated` there instead like:
`if ide_orchestrated: run_ide_orchestrated, else: exec_process`.

PR #3887, opened 2026-01-21
Fix agent's reverse DNS lookup logic

Razz4780:
1. Removed `tracing::error!` log. In mfT, the agent is a shared resource, used by multiple sessions. We should only ever log errors in case of unexpected failures / bugs / misconfiguration. Reverse lookup can fail due to transient network errors, or simply because the mirrord session tries to access a random unknown IP.
2. Moved reverse lookups to happen in the target's network namespace, if available.
3. Moved reverse lookups to happen in blocking tasks - each lookup was blocking the tokio runtime.

### DIFF
--- a/changelog.d/+reverse-lookup-agent.fixed.md
+++ b/changelog.d/+reverse-lookup-agent.fixed.md
@@ -1,0 +1,1 @@
+Fixed bogus error logs printed by the agent on failed reverse DNS lookups.


mirrord/agent/src/reverse_dns.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::{
+    error::{AgentError, AgentResult},
+    task::BgTaskRuntime,
+};
+
+/// Handles [`ClientMessage::ReverseDnsLookup`](mirrord_protocol::codec::ClientMessage::ReverseDnsLookup) requests.
+///
+/// Every client connection should use its own instance.
+///
+/// # TODO
```
0x00A5, line 16: 👀

Razz4780: Yeah, not solving this now :laughing:

Razz4780:
> Sorry for the bad code 😅

Yeah, same 👉 👉

PR #3892, opened 2026-01-21
Add CiSession (just counting the concurrent sessions) to mirrord operator status.

meowjesty:
- Issue: [COR-1162](https://linear.app/metalbear/issue/COR-1162/show-concurrent-ci-session-count-in-mirrord-operator-status-output)
- Related PRs:
  - https://github.com/metalbear-co/operator/pull/1216

Adds a print showing the count of active ci sessions (sessions started with `mirrord ci start`).

image

image


mirrord/operator/src/crd.rs
```diff
@@ -345,6 +354,42 @@ pub struct Session {
     pub kafka: Option<Vec<MirrordKafkaEphemeralTopicSpec>>,
 }
 
+/// _CRD-ish_ that represents a mirrord for ci session.
```
0x00A5: The .status.sessions already include these sessions. Do we need to add a new session type under status for only the length? I think we just need to add a count number, maybe under `MirrordOperatorStatusStatistics` which currently includes daily and monthly active users.

meowjesty: aa54dd4359054f09d9d16ef84d6e631f614b318b

Moved it there.

mirrord/cli/src/operator/status.rs
```diff
@@ -274,10 +274,17 @@ Operator License
...
-        if let Some(statistics) = status.statistics.as_ref() {
+        status.statistics.as_ref().and_then(|statistics| {
             println!("Operator Daily Users: {}", statistics.dau);
             println!("Operator Monthly Users: {}", statistics.mau);
-        }
+
+            println!(
+                "Operator Concurrent CI Sessions: {}",
+                statistics.active_ci_sessions_count?
+            );
```
0x00A5, line 284: This will print `Operator Concurrent CI Sessions: Some(42)` right in the output right? Maybe wrap it in a if, and only print if is_some()?

meowjesty: Nope, just the number. And if there's no number it doesn't print anything.

PR #3900, opened 2026-01-27
Added Kubeconfig rotation using GCP Secrets

pavelzag:
This PR allows K3s kubeconfig secrets rotation using GCP secrets manager


.github/workflows/scripts/verify_k3s_connectivity.ps1
aviramha, line 1: let's move this script to be under .github/workflows/scripts since I don't think it's relevant for anything but CI?

pavelzag: done

PR #3905, opened 2026-01-27
CRDs for HA copy target

Razz4780:


mirrord/operator/src/crd/copy_target.rs
```diff
@@ -58,3 +64,68 @@ impl CopyTargetStatus {
...
+
+/// Details of why a copy target request is closed.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordClusterCopyTargetRequestClosed {
+    /// Short reason name in PascalCase.
+    pub reason: String,
+    /// Human-friendly message.
+    pub message: String,
+}
```
t4lz, line 131: Nit: why not an enum that implements Display somehow (thiserror/strum/whatever)?

Razz4780:
Internal CRDs were moved to the other repo

PR #3906, opened 2026-01-27
Add a code review guide

t4lz:
Adds CODE_REVIEW.md documenting code review standards, and expectations from reviewers and PR authors.

### DIFF
--- a/CODE_REVIEW.md
+++ b/CODE_REVIEW.md
@@ -58,7 +58,7 @@ The reviewer can, of course, use AI however they like during their review, and t
 author's responsibility for a comprehensive and comprehensible PR.
 
 Both the author and the reviewer are encouraged to use LLMs as a first superficial review, that can easily find some
-small issues. An automatic solution for that can be incorporated in to the CI pipeline.
+small issues.
 
 The reviewer must still look at the actual code, and not only at AI-generated summaries etc.
 


CODE_REVIEW.md
```diff
@@ -0,0 +1,143 @@
...
+
+If you see something beautiful, smart, helpful, nice or well-crafted, feel free to comment just to acknowledge that.
+
+## By approving a PR, the reviewer confirms that:
+
+- [ ] The goal of this PR is good (we should indeed do this change)
+- [ ] The chosen solution is good
+- [ ] The new code is readable and comprehensible
+- [ ] The code does what it says it does
+- [ ] The code meets non-functional requirements (performance, security, etc.)
```
0x00A5, line 110: Do we want to mention tests?

t4lz: [added](https://github.com/metalbear-co/mirrord/pull/3906/changes/f912f57ebe235b4dbb2bf333bc37921f4ec8e976)

CODE_REVIEW.md
```diff
@@ -0,0 +1,144 @@
...
+their code are correct or not. Therefore, the author is responsible for including the necessary correct information in
+their PR (whether AI-generated or not).
+
+**The author is responsible for the correctness of any AI-generated content in their code and their PR.**
+
+The reviewer can, of course, use AI however they like during their review, and that does not take away from the
+author's responsibility for a comprehensive and comprehensible PR.
+
+Both the author and the reviewer are encouraged to use LLMs as a first superficial review, that can easily find some
+small issues. An automatic solution for that can be incorporated in to the CI pipeline.
```
gememma: > An automatic solution for that can be incorporated in to the CI pipeline.

I think you have to be _extremely_ careful considering this. Annecdotally I know of other companies that have this kind of thing, say for rating the quality of a changelog entry, and it is not worth the pain. I would say it's much easier and safer to let people use their own tools in this way, and let the human be the buffer between AI and CI.

meowjesty: Every time I tried this out, the AI generated a list of issues that really, REALLY, looked like real issues, but were not real issues. So I wonder if we should really bother integrating it into CI, or if this should just be left up to whoever is reviewing.

CODE_REVIEW.md
```diff
@@ -0,0 +1,144 @@
...
+
+The reviewer should actively try to help the author write good code.
+This means keeping an eye out for things that can be improved, and sharing ideas and suggestions.
+Some things are a matter of taste. If they are not clearly settled by any style guide followed by the team, it is ok to
+make suggestions that are not dictated by any guidelines, but make sure it is clear which comments require resolving,
+and which comments are actually just suggestions that the author is free to ignore if they disagree.
+
+### Highlighting Positives
+
+If you see something beautiful, smart, helpful, nice or well-crafted, feel free to comment just to acknowledge that.
```
gememma, line 102: peace and love on planet earth 🌻

CODE_REVIEW.md
t4lz: Ok, two comments agains this line. I think we can remove it. Doesn't really add a lot anyways, even if we did want to add such a solution to CI, this line doesn't add anything to the guide.
```suggestion
small issues.
```

PR #3909, opened 2026-01-28
INT-291 Multi Cluster - Session + Branching

vladrbg:


wizard-frontend/package-lock.json
t4lz, line 1: did you change this file on purpose?

mirrord/operator/src/client/connect_params.rs
```diff
@@ -49,6 +49,12 @@ pub struct ConnectParams<'a> {
         with = "session_ci_info_serde"
     )]
     pub session_ci_info: Option<SessionCiInfo>,
+
+    /// Full feature configuration as JSON for multi-cluster mode.
+    /// When set, the primary operator creates database branches on the default cluster
+    /// instead of the CLI creating them locally.
```
t4lz: "locally" is misleading here

mirrord/operator/src/client/connect_params.rs
```diff
@@ -49,6 +49,12 @@ pub struct ConnectParams<'a> {
...
+    #[serde(skip_serializing_if = "Option::is_none", with = "feature_config_serde")]
```
t4lz: why do we need `with`? Doesn't it do exactly what the default serializer and deserializer of `Option<String>` would do?

wizard-frontend/package-lock.json
vladrbg: Lol, nope I have missed that somehow will fix.

mirrord/operator/src/client.rs
```diff
@@ -1096,6 +1192,40 @@ impl OperatorApi<PreparedClientCert> {
...
+        target: &Target,
+        namespace: &str,
+        connect_params: &ConnectParams<'_>,
+    ) -> String {
+        use mirrord_config::target::TargetDisplay;
+
+        let name = {
+            let mut urlfied_name = target.type_().to_string();
+            urlfied_name.push('.');
+            urlfied_name.push_str(target.name());
```
t4lz: on targetless runs, this will result in `"targetless.targetless"`, which won't be a valid connect url. Idk if it makes sense to do a targetless multi-cluster run, but let's still make this function return correct urls.
The existing `target_connect_url` has unit tests, this function should too.

mirrord/operator/src/client.rs
```diff
@@ -890,6 +895,97 @@ impl OperatorApi<PreparedClientCert> {
...
+        let params = ConnectParams::new(
+            layer_config,
+            branch_name,
+            mongodb_branch_names,
+            mysql_branch_names,
+            pg_branch_names,
+            session_ci_info,
+        );
+
+        let namespace = namespace.unwrap_or("default");
```
t4lz: This shouldn't fallback to a hardcoded `"default"`, but rather fallback to the default namespace from the kubeconfig.
The code in `target_connect_url` is a bit misleading, because it also does `.unwrap_or("default")`, but in reality it will never be `None`, because it calls that on a namespaced resource returned from k8s, which will always the namespace in its metadata. Because in the existing `target_connect_url` the target is a `ResolvedTarget`, what actually happens there is: if the mirrord config specifies a namespace, we use that namespace, otherwise, we use the namespace from the kubeconfig to fetch the target resource from k8s. K8s will return that resource with its namespace in the metadata.

So for example, if I'm working in a kubeconfig context that has the `playground` as a default namespace, and I just specify a target name without a namespace in the mirrord config, mirrord will target that name under the `playground` namespace. But then if the cluster is the primary of a multi-cluster, mirrord would target that name under the `default` namespace, instead of under the `playground` namespace, and won't find the target because it's not there.

Idk if that was clear, lmk if you want me to explain again or go on a call.

mirrord/cli/src/connection.rs
```diff
@@ -91,28 +94,54 @@ where
...
+            .unwrap_or(Target::Targetless);
+        let target_namespace = layer_config.target.namespace.clone();
 
-    let mut session_subtask = operator_subtask.subtask("starting session");
-    let connection = api
-        .connect_in_new_session(
-            target.clone(),
+        api.connect_in_multi_cluster_session(
+            &target_config,
+            target_namespace.as_deref(),
```
t4lz: Why this clone, won't this work:
```suggestion
        api.connect_in_multi_cluster_session(
            &target_config,
            layer_config.target.namespace.as_deref(),
```

mirrord/cli/src/connection.rs
```diff
@@ -91,28 +94,54 @@ where
...
             layer_config,
             &session_subtask,
             branch_name,
             mirrord_for_ci.map(MirrordCi::info),
         )
-        .await?;
+        .await?
+    } else {
+        // Single-cluster: resolve target locally
```
t4lz: What do you mean by "locally" here?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
+use std::collections::HashMap;
+
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::MicroTime;
+use kube::CustomResource;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use super::session::{SessionOwner, SessionTarget};
+
+/// Multi-cluster session coordinated by Envoy
```
t4lz: Please mention on which cluster this resource is created.

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(CustomResource, Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
+#[kube(
+    group = "operator.metalbear.co",
+    version = "v1alpha",
+    kind = "MirrordMultiClusterSession",
+    status = "MirrordMultiClusterSessionStatus",
+    namespaced
```
t4lz, line 17: why is this namespaced?
the normal session CRD is not, is it?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+)]
+#[serde(rename_all = "camelCase")]
+pub struct MirrordMultiClusterSessionSpec {
+    /// Owner of this session
+    pub owner: SessionOwner,
+
+    /// Kubernetes namespace for the session (in primary cluster)
```
t4lz: really, in the primary cluster? Aren't the child session objects in the other clusters?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub namespace: String,
+
+    /// Logical target alias provided by admin (e.g., "x-crm")
```
t4lz: > e.g., "x-crm"
Isn't this a lie - isn't it something like "deployment.x-crm"

> provided by admin
Isn't this controlled by the mirrord config provided by the user?

> Logical target alias
What does "**logical**" target alias mean?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// This resolves to multiple targets across clusters
+    pub target_alias: String,
+
+    /// Clusters to create sessions in
+    /// Key: cluster name (e.g., "us-east-1")
+    /// Value: target details for that cluster
+    pub cluster_targets: HashMap<String, ClusterTarget>,
```
t4lz: Isn't this the same for every session?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// Primary cluster where ephemeral resources will be created
```
t4lz: Isn't whatever operator seeing this always the primary cluster, and aren't resources created on the _default_ cluster, not necessarily the _primary_ cluster?

t4lz: If I misunderstood, please explain what ephemeral resources are.

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub primary_cluster: String,
+
+    /// Default cluster for stateful operations (db_branches, etc.)
```
t4lz, line 40:
```suggestion
    /// Default cluster for stateful operations (db_branches, etc.)
    /// The primary cluster will be used as the default cluster when a default cluster is not specified.
```

mirrord/operator/src/crd/session.rs
```diff
@@ -43,6 +43,54 @@ pub struct MirrordClusterSessionSpec {
...
+    /// The default cluster's operator then uses these branches for the session.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub mysql_branch_names: Vec<String>,
+
+    /// MongoDB branch database names created by the primary for this session.
+    ///
+    /// For multi-cluster: primary creates branches on default cluster and passes names here.
+    /// The default cluster's operator then uses these branches for the session.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub mongodb_branch_names: Vec<String>,
```
t4lz: What do we need those here for?

mirrord/operator/src/crd/multi_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// MySQL branch database names (created by primary before session init).
+    /// These are passed to the default cluster's child session.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub mysql_branch_names: Vec<String>,
+
+    /// MongoDB branch database names (created by primary before session init).
+    /// These are passed to the default cluster's child session.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub mongodb_branch_names: Vec<String>,
```
t4lz: What do we need those here for?

mirrord/operator/src/crd/session.rs
```diff
@@ -43,6 +43,54 @@ pub struct MirrordClusterSessionSpec {
     /// Set when the session uses a copied pod.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub copy_target: Option<SessionCopyTarget>,
+
+    /// Multi-cluster coordination: ID of the parent MirrordMultiClusterSession.
+    ///
+    /// If set, this session is part of a multi-cluster session and was created by Envoy.
```
t4lz:
```suggestion
    /// If set, this session is part of a multi-cluster session and was created by the envoy.
```
Let's say "the envoy", otherwise it sounds like a name, and then it sounds like you mean https://www.envoyproxy.io/

mirrord/operator/src/crd/session.rs
```diff
@@ -43,6 +43,54 @@ pub struct MirrordClusterSessionSpec {
...
+    /// The value is the name of the MirrordMultiClusterSession CRD in the primary cluster.
```
t4lz:
```suggestion
    /// The value is the name of the MirrordMultiClusterSession CR in the primary cluster.
```
CRD = CustomResourceDefinition
CR = Custom Resrouce
In this case - we're talking about the name of a specific instance -> CR.

mirrord/operator/src/crd/session.rs
```diff
@@ -43,6 +43,54 @@ pub struct MirrordClusterSessionSpec {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub multi_cluster_parent_id: Option<String>,
```
t4lz: nit: if it's the name of the parent, then maybe we should call it `multi_cluster_parent_name`, not `multi_cluster_parent_id`.

mirrord/operator/src/client.rs
```diff
@@ -715,24 +715,29 @@ impl OperatorApi<PreparedClientCert> {
             .supported_features()
             .contains(&NewOperatorFeature::ProxyApi);
 
-        let mongodb_branch_names = if layer_config.feature.db_branches.is_empty().not() {
+        // Create branches locally (single-cluster mode).
+        // Branches are created and processed on the same cluster.
```
t4lz:
```suggestion
```

This is outdated, right?

mirrord/operator/src/client.rs
t4lz: Or maybe currently the default namespace comes is defined in the operator? The point is, we should use the same default namespace as in single-cluster.

mirrord/operator/src/client.rs
```diff
@@ -890,6 +895,97 @@ impl OperatorApi<PreparedClientCert> {
...
+            .operator
+            .spec
+            .supported_features()
+            .contains(&NewOperatorFeature::ProxyApi);
+
+        // Create branch CRDs on the cluster CLI has access to:
+        // - Single-cluster: CRDs created directly on target cluster, branching happens locally.
+        // - Multi-cluster: CRDs created on primary cluster. If Primary == Default, branching
+        //   happens locally. If Primary != Default, a sync controller copies CRDs to Default
+        //   where actual branching happens, then syncs status back.
```
t4lz: Ok, you're using "locally" to mean "on the cluster we're directly connected to" when in the context of multicluster. However, that's confusing, because we usually use "locally" to mean "on the user's machine". So please something else instead of "locally", maybe "on the entry cluster" or something like that.

mirrord/cli/src/connection.rs
vladrbg: Yep, but actually it was not needed to be here, we are already passing layer_config so we can use it in the function to get the namespace and not pass separately.

mirrord/cli/src/connection.rs
vladrbg: I updated the comments of both single/multi-cluster, the idea is that when:

1. multi-cluster: we connect to primary and primary connects to 'remote-1' for example that we have the target. 
2. single-cluster: CLI is connecting directly to the cluster that we have the target that we resolve

mirrord/operator/src/crd/multi_cluster.rs
vladrbg: Oh you are right yep.

mirrord/operator/src/client.rs
vladrbg: Yep yep, okay will fix on other places if I see that I mentioned locally.

mirrord/operator/src/client.rs
vladrbg: yep, removed

mirrord/operator/src/crd/multi_cluster.rs
vladrbg: lol, yep thats reversed :D fixed it, it is on workload clusters. not primary.

mirrord/operator/src/crd/session.rs
vladrbg: Took me some time but yep, i was initially storing them in crds, but changed it to work have them passed through ConnectParams where we pass them directly as query params but in the multi cluster case we pass them also not only from CLI -> Primary but also Primary -> Remote. So fixed it removed it from MultiCluster sesison and Mirrord Session.

PR #3910, opened 2026-01-28
Removed operator setup command, use helm chart instead

aviramha:


Razz4780:
I think we need to remove `setup` feature from the `mirrord-operator` crate as well

PR #3911, opened 2026-01-28
Reuse intproxy connections

MintSoup:
When calling exec, the new layer would start a brand new connection to the intproxy, which broke remote resource tracking as the parent layer/child layer relationship was severed. This PR makes it so that the connection is recovered and reused across exec calls, which fixes this issue.

Note that some apps in the test suite (uvicorn) close all fds before execing, which breaks this obviously. To work around that I've added a `close_range` hook (which was used by uvicorn) with appropriate guards that prevent the intproxy connection from getting closed. Same guard has been added to the normal `close` hook as well, because surprise, in addition to calling `close_range(3, i32::MAX)` uvicorn ALSO loops over all open fds and closes them one by one...

We're also hooking (and modifying already hooked functions from) the `posix_spawn` family because there's some additional macos-specific ways to close fds before spawning a new process. Nodejs spawn was the source of issues here.

### DIFF
--- a/mirrord/layer/src/lib.rs
+++ b/mirrord/layer/src/lib.rs
@@ -902,8 +902,8 @@ pub(crate) unsafe extern "C" fn close_range_detour(
         // Set the close-on-exec flag on the specified file
         // descriptors, rather than immediately closing them.
         if flags.bitand(CLOSE_RANGE_CLOEXEC) == 0 {
-            sockets.retain(|k, _| range.contains(k));
-            files.retain(|k, _| range.contains(k));
+            sockets.retain(|k, _| range.contains(k).not());
+            files.retain(|k, _| range.contains(k).not());
         }
 
         res


mirrord/layer/src/lib.rs
```diff
@@ -628,6 +630,13 @@ fn enable_hooks(state: &LayerSetup) {
 
     unsafe {
         replace!(&mut hook_manager, "close", close_detour, FnClose, FN_CLOSE);
+        replace!(
+            &mut hook_manager,
+            "close_range",
+            close_range_detour,
```
0x00A5, line 640: This is the one exposed by our tests, but I am a bit concerned that there are more places where we need to guard the proxy conn fd (e.g. `fcntl`, `closefrom`).

mirrord/layer/src/lib.rs
```diff
@@ -896,7 +966,11 @@ pub(crate) unsafe extern "C" fn vfork_detour() -> pid_t {
         }
     };
 
+    // Drop the `DetourGuard` so we actually fork_detour.
+    drop(__bypass);
```
0x00A5: nit: since we are manually managing the guard, might as well make the function `hook_function` and create the guard in the function? A bit more readable than `__bypass`

mirrord/layer/src/common.rs
```diff
@@ -137,6 +144,7 @@ impl CheckedInto<Argv> for *const *const c_char {
             .filter(|value| !value.is_null())
             .map(|value| unsafe { CStr::from_ptr(*value) }.to_owned())
             .filter(|value| !value.to_string_lossy().starts_with(SHARED_SOCKETS_ENV_VAR))
+            .filter(|value| !value.to_string_lossy().starts_with(FD_ENV_VAR))
```
0x00A5: super nit: `FD_ENV_VAR` is quite generic, maybe `INTPROXY_FD_ENV_VAR`?

mirrord/layer/src/proxy_connection.rs
```diff
@@ -59,7 +84,10 @@ impl ProxyConnection {
         session: NewSessionRequest,
         timeout: Duration,
     ) -> Result<Self> {
-        let connection = TcpStream::connect(proxy_addr)?;
+        let guard = DetourGuard::new();
```
0x00A5, line 87: Do we need this guard here? Proxy connection should be established before detours are hooked.

mirrord/intproxy/src/lib.rs
```diff
@@ -634,7 +635,26 @@ impl IntProxy {
...
                     .await
             }
-            other => Err(ProxyRuntimeError::UnexpectedLayerMessage(other))?,
+            LayerToProxyMessage::NewSession(req) => {
+                let layer_info = self
+                    .connected_layers
+                    .get_mut(&layer_id)
+                    .expect("Received NewSession from layer with an unknown layer id");
+
+                *layer_info = req.process_info;
```
0x00A5, line 644: Will new process receive agent response to requests queued for the old process (same layer id)?

mirrord/layer/src/lib.rs
0x00A5, line 1: > When calling exec, the new layer would start a brand new connection to the intproxy, which broke remote resource tracking as the parent layer/child layer relationship was severed.

What is the resource tracking issue before this change? Can you expand a little more on this bit?

mirrord/layer/src/exec_hooks/hooks.rs
```diff
@@ -41,6 +43,7 @@ pub(crate) fn prepare_execve_envp(env_vars: Detour<Argv>) -> Detour<Argv> {
     drop(lock);
 
     env_vars.insert_env(SHARED_SOCKETS_ENV_VAR, &encoded)?;
+    env_vars.insert_env(FD_ENV_VAR, &proxy_conn_fd().unwrap().to_string())?;
```
0x00A5: This will also set the env for process created by `posix_spawn_detour`. I only found parent layer id set in `fork_detour()`. So spawned process will reuse proxy connection too, how does that work? I may have missed something 🤔

mirrord/layer/src/proxy_connection.rs
MintSoup: > Proxy connection should be established before detours are hooked

Not right after forking, in that case we create a new connection after hooks are on.

mirrord/layer/src/lib.rs
MintSoup: It has to do with everything that uses `RemoteResources`. That struct tracks which layer has access to which resources, and it duplicates them from parent to child during a fork. Problem is when we call `exec`, the new process inherits the resources but the intproxy is not aware of this, as it creates a new connection and registers with a new layer id with no relation to the parent. So effectively doing a fork+exec makes the intproxy lose track of all resources inherited from parent to child.

mirrord/intproxy/src/lib.rs
MintSoup: Yes. Intproxy-side, all that really changes is the `ProcessInfo` for the layer in question, nothing else.

mirrord/layer/src/exec_hooks/hooks.rs
MintSoup: Oh I think I might have goofed up here, I think in `posix_spawn_detour` we should create a new connection (with current process id as parent), serialize it, and recover it in the child. That would correctly replicate the behavior of a fork + exec

mirrord/layer/src/lib.rs
MintSoup: Yeah, there are many syscalls that could potentially mess up our fd, `dup` family being a particularly nasty case which I haven't hooked. My intention was to hook everything that might be commonly used to "close all dangling fds and exec". I've added a hook for `closefrom`, and I think it might make sense to guard `fcntl` against adding `CLOEXEC` on the intproxy connection, but honestly idk if we can/should hook other things like dup.

mirrord/intproxy/src/lib.rs
0x00A5: Resolving this one, discussed in our call. Queued response will reach new process, but stored in response manager's map instead of erroring out.

mirrord/layer/src/lib.rs
0x00A5: >doing a fork+exec makes 

you mean doing a exec without fork right? Fork will clone all from parent to child?

mirrord/layer/src/proxy_connection.rs
```diff
@@ -88,6 +116,57 @@ impl ProxyConnection {
...
+        Ok(conn)
+    }
+
+    fn acquire_connection(have_parent_layer: bool, proxy_addr: SocketAddr) -> Result<TcpStream> {
+        // Parent layer is Some OR we don't have the env var set =>
+        // We're calling this from a fork(), i.e. this is a a brand
+        // new process and we have to establish a new connection to
+        // the intproxy.
+        if have_parent_layer {
+            return Self::make_new_connection(proxy_addr);
```
0x00A5, line 136: Should this be reusing connection now after the change in spawn?

mirrord/layer/src/lib.rs
MintSoup: Yeah we don't need to do a fork to lose everything, just an exec is enough. Fork + exec will copy but immediately discard,

mirrord/layer/src/proxy_connection.rs
MintSoup: Yes, why would it not?

PR #3913, opened 2026-01-29
DEV-192 Automate Everything for mirrord Releases

pavelzag:
# Automated Release Workflow

## Overview

This workflow automates the entire release process for the mirrord repository. Developers can now trigger releases with a single command, eliminating manual version bumping, changelog generation, and GitHub release creation.

## Key Features

**Automatic Version Calculation**
The workflow analyzes changelog fragments to determine the appropriate semantic version bump. Patch releases for fixes, minor releases for features, and the flexibility to force specific versions when needed.

**Changelog Generation**
Release notes are automatically compiled from changelog fragments using towncrier. Each fragment becomes part of a structured, consistent changelog that's generated and committed with every release.

**One-Click Releases**
Trigger a release with a single GitHub Actions workflow dispatch. The system handles version updates in `Cargo.toml`, changelog generation, git tagging, and GitHub release creation automatically.

**Idempotent Design**
The workflow can be safely re-run at any point. It detects existing tags, changelog entries, and commits, skipping unnecessary steps while completing any missing actions.

## Dockerized Tooling

The workflow uses a containerized version calculation script stored in the `metal-ops` repository at `projects/calculate-version-script`. This script analyzes changelog fragments to determine semantic version bumps and is packaged with towncrier in a Docker image published to Google Artifact Registry at `us-central1-docker.pkg.dev/mirrord/release-tools/calculate-version-script`. This centralized approach ensures consistent versioning logic across all repositories.

## Usage

**Standard Release**
```bash
gh workflow run auto-release.yaml --ref main
```

**Forced Version**
```bash
gh workflow run auto-release.yaml --ref main -f force_version="4.0.0"
```

## Benefits

- Eliminates manual release steps and potential human errors
- Ensures consistent versioning across all releases
- Maintains a clean, structured changelog automatically
- Reduces release time from minutes to seconds
- Provides full traceability through git history and GitHub releases


.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Set Docker tags
+        uses: metalbear-co/ci/.github/actions/set-image-tags@main
+        id: set-tags
+        with:
+          base-name: ci/calculate-version-script
+      - name: Calculate Version
+        id: version
+        run: |
+          echo "Using Dockerized tools..."
+          TOOLS_IMAGE=$(echo "${{ steps.set-tags.outputs.tags }}" | cut -d',' -f1)
```
t4lz: What's the purpose of this?
If I understand `set-image-tags` correctly, what's happening here is that if the run is a workload dispatch (which is the only case here cause that's the only `on` defined for this workflow), we use `ghcr.io/metalbear-co/ci/calculate-version-script-staging:${{ github.sha }}` where `github.sha` is going to be the commit sha from this repo, not `metalbear-co/ci`. So how does that even work - will that image ever exist?
And why would we want to use `calculate-version-script-staging`?
Am I completely misreading this?

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,97 @@
...
+               /usr/local/bin/calculate_version.py --file Cargo.toml --changelog-dir changelog.d)
+          fi
+          echo "next_version=$next_version" >> $GITHUB_OUTPUT
+
+      - name: Update Files
+        if: ${{ github.event.inputs.dry_run != 'true' }}
+        run: |
+          v="${{ steps.version.outputs.next_version }}"
+          # Using ^ to match start of line to avoid matching dependencies
+          sed -i "s/^version = \".*\"/version = \"$v\"/" Cargo.toml
```
t4lz: This could still match a version dependency if we define it as a toml sub-table. For example this is less common but valid:

```toml
[workspace.dependencies.tokio]
version = "1"
features = ["fs", "net", "io-util"]
```
It's possible that at some point we define a dependency with so many parameters, that someone decides it's too long for a single line and makes it a toml sub-table.

If we want to really prevent a case where we accidentally replace a dependency version with the new workspace version, we could do that the same way as the current [`release.sh`](https://github.com/metalbear-co/mirrord/blob/eabe775f4751eadce6a0ec49b15a89cc316e643c/scripts/release.sh) script does that:
https://github.com/metalbear-co/mirrord/blob/eabe775f4751eadce6a0ec49b15a89cc316e643c/scripts/release.sh#L86-L87

(for that `cargo-set` needs to be installed first).

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,97 @@
...
+            echo "Changelog entry for $v already exists, skipping creation."
+          else
+            docker run --rm -v $PWD:/workspace -w /workspace \
+              --entrypoint towncrier \
+              $TOOLS_IMAGE \
+              build --yes --version "$v"
+          fi
+          
+          # Extract the new entry for release body (from first header to next header)
+          awk '/^## / { if (p) { exit }; p=1; print; next } p { print }' CHANGELOG.md > RELEASE_NOTES.md
```
t4lz: `internal` entries are present at the changelog but should not be included in the release notes.

changelog.d/+automated-release.internal.md
```diff
@@ -0,0 +1 @@
+Added automated release workflow. Check the PR for details.
```
t4lz:
```suggestion
Added automated release workflow.
```

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          sudo chown -R $(id -u):$(id -g) .
+          v="${{ steps.version.outputs.next_version }}"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add Cargo.toml CHANGELOG.md changelog.d
+          if ! git diff --staged --quiet; then
+            git commit -m "Release $v"
+            git pull --rebase origin main
+            git push origin main
```
t4lz, line 83: Will this succeed, given that main is a protected branch in this repository?

.github/workflows/auto-release.yaml
t4lz: I think after changing the package version in `Cargo.toml` we need to update `Cargo.lock`, otherwise building mirrord will fail in the workflows that are triggered after the release.

Also here you can just copy the existing `release.sh` script:

https://github.com/metalbear-co/mirrord/blob/eabe775f4751eadce6a0ec49b15a89cc316e643c/scripts/release.sh#L88-L89

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,97 @@
...
+          else
+            git tag "$v"
+            git push origin "$v"
+          fi
+          
+      - name: Create Release
+        if: ${{ github.event.inputs.dry_run != 'true' }}
+        uses: metalbear-co/action-gh-release-1@v1
+        with:
+          tag_name: ${{ steps.version.outputs.next_version }}
```
t4lz, line 97: Just to make sure, the commit this creates the release from is not the commit this workflow was triggered for, but the commit at `tag_name`, right?

.github/workflows/auto-release.yaml
pavelzag: It will fail unless we have a PAT with repo scope and replace the old one that probably doesn't have the repo scope. I'll make one
`token: ${{ secrets.MIRRORD_RELEASE_PAT }}`

.github/workflows/auto-release.yaml
pavelzag: The release will be associated with the commit pointed to by the `tag_name`
Not the commit that originally triggered the workflow.


1: In the Commit and Push step, we create a new commit containing the version bump and the cleaned-up changelog.
2: We then run `git tag "$v"` which points that tag to the newly created commit at the tip of your local branch.
3: We run `git push origin "$v"` to push that tag to GitHub.
4: Create Release step is called with `tag_name: ${{ steps.version.outputs.next_version }}`

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install Rust tools
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-get,cargo-edit
```
t4lz: I see we're installing `cargo-get` and `cargo-edit`, but I only see `cargo-set` being used (without being installed?). Is there some reason here or is it just by mistake?

.github/actions/set-image-tags/action.yaml
t4lz, line 1: why duplicate that reusable action here?

.github/workflows/auto-release.yaml
pavelzag: I removed cargo-get installation. We are doing cargo-set so we need cargo-edit installation

.github/workflows/auto-release.yaml
pavelzag: This part was a leftover copy-pasta from the CI repo (where we build the `calculate-version-script` images) and made zero sense here since we're just consuming them. I've nuked the confusing `workflow_dispatch` block so it strictly pulls `:latest` for tools now. Bless you for catching that!

.github/actions/set-image-tags/action.yaml
pavelzag: You mean why am I not using this:
https://github.com/metalbear-co/ci/blob/0a4e182e52fceb77152b0dd409b18b09ec81b847/.github/actions/set-image-tags/action.yaml
?
It's a bit different. Doesn't make sense to call for outside action for such a small thing imho

.github/workflows/auto-release.yaml
t4lz: From what I can see, [cargo-edit](https://crates.io/crates/cargo-edit) gives us `cargo set-version` which we're not currently using.
And [`cargo set`](https://crates.io/crates/cargo-set) is installed by `cargo-set`, not `cargo-edit`.

.github/workflows/auto-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install Rust tools
+        uses: taiki-e/install-action@v2
```
aviramha: is that whitelisted? if so it shouldn't be.. probably gonna fail since not owned by us or trusted vendor.

.github/workflows/auto-release.yaml
```diff
@@ -36,15 +36,15 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install Rust tools
-        uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-get,cargo-edit
+      - name: Set Environment Variables
+        run: |
+          TAGS="${{ steps.set-tags.outputs.tags }}"
+          echo "TOOLS_IMAGE=${TAGS%%,*}" >> $GITHUB_ENV
```
t4lz: When is `TOOLS_IMAGE` ever not going to end up being `ghcr.io/metalbear-co/ci/calculate-version-script:latest`?

If the answer is never, please let's just use that directly instead of going through the unnecessary and confusing `set-tags` step.

.github/actions/set-image-tags/action.yaml
t4lz: I don't get the need for it here at all actually.

.github/workflows/auto-release.yaml
```diff
@@ -56,6 +56,12 @@ jobs:
           fi
           echo "next_version=$next_version" >> $GITHUB_OUTPUT
 
+      - name: Install Rust tools
+        if: ${{ github.event.inputs.dry_run != 'true' }}
+        uses: taiki-e/install-action@v2
```
t4lz: this will probably be blocked, no? I think we block external actions by default.

pavelzag: changed to pull it from our org

.github/workflows/auto-release.yaml
pavelzag: @aviramha You mentioned you want me to use it. 
Do we need it? I personally don't really think we do..

aviramha: That comment was for the image you added in CI.. I'm pretty sure I didn't write it in this automation context

.github/workflows/auto-release.yaml
aviramha: did you check in the test repo you made to have protected branch with same settings as we have to see if that's true?

pavelzag: mirrord-dummy has protected branch enabled
https://github.com/metalbear-co/mirrord-dummy/settings

.github/workflows/auto-release.yaml
pavelzag: I have forked it in our org

aviramha: It was already forked no?

pavelzag: yes

PR #3917, opened 2026-01-31
Preview Environment

iniw:
- Related issues:
  - https://linear.app/metalbear/issue/COR-1086/23-mirrord-preview-env-support-with-key
  - https://linear.app/metalbear/issue/COR-1087/33-mirrord-preview-env-status-and-clean-commands
- Related PRs:
  - https://github.com/metalbear-co/operator/pull/1224
  - https://github.com/metalbear-co/charts/pull/324
  
### Design overview

The CLI's role is pretty narrow: create a `PreviewSessionCrd`, watch its status until it reaches `Ready`, and report the result. All real work — pod creation, agent management, traffic routing — is the operator's responsibility. The CRD is the contract between the two: the CLI writes the spec (image, target, traffic config, TTL), and the operator communicates progress back through the status enum (`Initializing -> Waiting -> Ready / Failed`).

Config follows the standard mirrord pipeline: CLI flags become env var overrides that feed into `LayerConfig`. TTL goes into the CRD spec for the operator to enforce; creation timeout stays CLI-side, deleting the CRD if readiness isn't reached in time, causing the operator's controller to clean everything else up.

### DIFF
--- a/changelog.d/+preview-env.added.md
+++ b/changelog.d/+preview-env.added.md
@@ -1,1 +1,1 @@
-mirrord now supports "preview environments" - a new type of mirrord session that lives directly in the cluster and can be shared with other users. Read more about it [here](PLACEHOLDER-THERE-ARE-NO-DOCS-YET).
+mirrord now supports "Preview Environments" - a new type of mirrord session that lives directly in the cluster and can be shared with other users.


changelog.d/+preview-env.added.md
```diff
@@ -0,0 +1 @@
+mirrord now supports "preview environments" - a new type of mirrord session that lives directly in the cluster and can be shared with other users. Read more about it [here](PLACEHOLDER-THERE-ARE-NO-DOCS-YET).
```
gememma: dont merge without those docs 👀 leaving a comment so you have to resolve it

mirrord/cli/src/config.rs
```diff
@@ -1165,6 +1169,227 @@ pub struct FixKubeconfig {
...
+            envs.insert("MIRRORD_KUBE_CONTEXT".as_ref(), context.as_ref());
+        }
+
+        envs
+    }
+}
+
+/// Arguments for `mirrord preview clean` command.
+#[derive(Args, Debug)]
+pub(super) struct PreviewStopArgs {
```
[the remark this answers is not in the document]
iniw: Note: the command used to be called `clean` but was renamed to `stop` to mirror `start` and to match the existing `ci {start,stop}` commands. This is a leftover :)

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,462 @@
...
+        let name = session
+            .metadata
+            .name
+            .as_deref()
+            .expect("preview session should have a name");
+        let namespace = session
+            .metadata
+            .namespace
+            .as_deref()
+            .unwrap_or("preview session should have a namespace");
```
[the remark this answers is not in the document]
iniw: Real bug! The robots are great at typos...

changelog.d/+preview-env.added.md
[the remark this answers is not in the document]
iniw: Gemma already told me!!

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -51,17 +51,17 @@ use crate::{
...
 mod port_subscription_ext;
 mod subscriptions;
-mod tasks;
-mod tcp_proxy;
+pub mod tasks;
+pub mod tcp_proxy;
 #[cfg(test)]
 mod tests;
-mod tls;
+pub mod tls;
```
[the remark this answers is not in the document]
iniw: I've reduced the number of export modules to just the three the traffic router actually uses: `http`, `tasks` and `tls` - these must stay public. The rest were only used during development and have been made private again.

mirrord/config/configuration.md
```diff
@@ -716,18 +715,20 @@ Disables any system wide proxy configuration for affecting the running applicati
 
 ### _experimental_ latency {#experimental-latency}
 
+Configuration for adding artificial latency to outgoing network operations.
+
 Configuration for adding artificial latency to outgoing network operations.
 Useful for testing application behavior under network delay conditions.
```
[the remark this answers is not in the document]
iniw: This is unrelated to this PR

mirrord/cli/src/port_forward.rs
```diff
@@ -931,123 +924,17 @@ impl IncomingMode {
...
-                any_of: Some(filters),
-                ports: _ports,
-            } => Self::make_composite_filter(false, filters),
-
-            _ => panic!("No HTTP filters specified, this should have been caught earlier"),
-        };
+        let filter = config
+            .http_filter
+            .as_protocol_http_filter()
+            .expect("invalid HTTP filter expression");
```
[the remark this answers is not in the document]
iniw: This already the behavior prior to this change.

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -185,6 +189,11 @@ impl HttpFilterConfig {
             || self.body_filter.is_some()
     }
 
+    pub fn is_filter_not_set(&self) -> bool {
+        // :not:
+        self.is_filter_set().not()
+    }
```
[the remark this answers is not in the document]
iniw: The robot doesn't know 💀 :not:

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -271,6 +280,99 @@ impl HttpFilterConfig {
...
+                    Ok(HttpFilter::Method(HttpMethodFilter::from_str(method)?))
+                }
+                InnerFilter::Body(body_filter) => Ok(HttpFilter::Body(
+                    body_filter.as_protocol_http_body_filter()?,
+                )),
+            })
+            .collect::<Result<Vec<_>, HttpFilterParseError>>()?;
+
+        Ok(HttpFilter::Composite { all, filters })
+    }
```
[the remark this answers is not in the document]
iniw: This is just keeping the old behavior. Bad bot.

mirrord/cli/src/config.rs
```diff
@@ -1165,6 +1169,280 @@ pub struct FixKubeconfig {
...
+    /// - `rollout/{rollout-name}[/container/{container-name}]`
+    /// - `statefulset/{statefulset-name}[/container/{container-name}]`
+    ///
+    /// The preview pod will be a copy of the target's pod spec with your image.
+    #[arg(short = 't', long)]
+    pub target: Option<String>,
+
+    /// Namespace of the target.
+    #[arg(short = 'n', long)]
+    pub target_namespace: Option<String>,
```
0x00A5, line 1265: minor: it is `--namespace` in stop and status, but `--target-namespace` here. Make this one `--namespace` too?

mirrord/config/configuration.md
```diff
@@ -2594,6 +2595,39 @@ of regexes specified here. If there is a match, mirrord will connect your applic
...
+### feature.preview.image {#feature-preview-image}
+
+Container image to run in the preview pod.
+The image must be pre-built and pushed to a registry accessible by the cluster.
+
+### feature.preview.ttl_mins {#feature-preview-ttl_mins}
+
+How long (in minutes) the preview session is allowed to live after creation.
+The operator will terminate the session when this time elapses.
+
```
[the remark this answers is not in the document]
iniw: This is just the style we use....

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    let api = Api::<PreviewSessionCrd>::namespaced(operator_api.client().clone(), namespace);
+
+    let sanitized_target = target.to_string().replace('/', "-");
+    let uuid_short = Uuid::new_v4().simple().to_string();
+    let uuid_short = &uuid_short[..8];
+    let preview = PreviewSessionCrd::new(
+        &format!("preview-session-{sanitized_target}-{uuid_short}"),
+        spec,
+    );
```
iniw: But how would a target like `deployment/My.App_v2` even exist in the first place if the name is invalid...

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+    config::{PreviewArgs, PreviewCommand, PreviewStartArgs, PreviewStatusArgs, PreviewStopArgs},
+    error::{CliError, CliResult},
+    user_data::UserData,
+};
+
+/// Handle commands related to preview environments: `mirrord preview ...`
+pub(crate) async fn preview_command(
+    args: PreviewArgs,
+    _watch: Watch,
+    _user_data: &UserData,
```
0x00A5: Re these unused args, do you plan to add analytics later?

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+async fn preview_start(args: PreviewStartArgs) -> CliResult<()> {
+    let mut progress = ProgressTracker::from_env("mirrord preview start");
+
+    let layer_config = load_preview_config(args.as_env_vars(), &mut progress)?;
+
+    // Connect to the operator — preview environments require an operator deployment.
+
+    let mut subtask = progress.subtask("connecting to operator");
+
+    let operator_api = OperatorApi::try_new(&layer_config, &mut NullReporter::default(), &progress)
```
0x00A5: When we add analytics reporter to the command, call `OperatorAPI::with_client_certificate` so the operator properties are added to the report.

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+        "found {} session{}",
+        sessions.len(),
+        if sessions.len() == 1 { "" } else { "s" }
+    )));
+
+    progress.success(None);
+
+    // Display sessions grouped by key.
+
+    let mut grouped: HashMap<&str, Vec<&PreviewSessionCrd>> = HashMap::new();
```
0x00A5: super minor: switch to `BTreeMap` so the output order is deterministic (always sorted by `key`)?

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+    let sessions = api.list(&ListParams::default()).await.map_err(|e| {
+        subtask.failure(None);
+        CliError::PreviewListFailed(e.to_string())
+    })?;
+
+    let key_filter = layer_config.key.provided();
+    let sessions: Vec<_> = sessions
+        .items
+        .into_iter()
+        .filter(|session| key_filter.is_none_or(|key| session.spec.key == key))
```
0x00A5: Should we add `key` as a label of the preview CR, and use label selector to do server-side filtering?

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+    let sessions = api.list(&ListParams::default()).await.map_err(|e| {
+        subtask.failure(None);
+        CliError::PreviewListFailed(e.to_string())
+    })?;
+
+    let sessions_to_delete: Vec<_> = sessions
+        .items
+        .into_iter()
+        .filter(|session| {
+            session.spec.key == key
```
0x00A5: Same here, use label selector?

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,480 @@
...
+    let mut subtask = progress.subtask("connecting to operator");
+
+    let operator_api = OperatorApi::try_new(&layer_config, &mut NullReporter::default(), &progress)
+        .await?
+        .ok_or_else(|| {
+            subtask.failure(None);
+            CliError::OperatorRequiredForPreview
+        })?;
+
+    operator_api.check_license_validity(&progress)?;
```
0x00A5: I think we should add a new variant in `NewOperatorFeature` for preview env. And check if `self.operator.spec.supported_features` contains preview env. If not, exit with a clean error that the user's operator does not support this command.

0x00A5: The linear issue says this feature is available to enterprise tier clients. So remember to check the license tier somewhere.

mirrord/config/src/feature/preview.rs
```diff
@@ -0,0 +1,49 @@
...
+///       "creation_timeout_secs": 60
+///     }
+///   }
+/// }
+/// ```
+#[derive(MirrordConfig, Default, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
+#[config(map_to = "PreviewFileConfig", derive = "JsonSchema")]
+#[cfg_attr(test, config(derive = "PartialEq, Eq"))]
+pub struct PreviewConfig {
+    /// ### feature.preview.image {#feature-preview-image}
```
0x00A5: These should be `####` instead of `###` I believe. `FeatureConfig` fields are already `###`

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Status of a preview session resource.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
+pub enum PreviewSessionStatus {
+    /// Preview pod hasn't been created yet - operator is still setting things up.
+    Initializing {
+        // kube-rs does not allow mixing unit variants with tuple/struct variants, so this variant
+        // has to be a tuple/struct too. If we leave the tuple empty, k8s complains about an object
+        // without any items, and kube-rs does not support internally tagged enums, so we actually
+        // have to put something in there, even if we don't actually care about that info.
+        start_time_utc: String,
```
0x00A5: Most of the timestamp in our CRDs use the type `MicroTime`.

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Whether to mirror or steal traffic from the target.
+    #[serde(default, skip_serializing_if = "IncomingMode::is_off")]
+    pub mode: IncomingMode,
+
+    /// HTTP filter configuration (header, path, method, body, composite filters).
+    /// Unfiltered stealing is rejected by the operator to avoid disrupting all traffic to the
+    /// target.
+    #[serde(default, skip_serializing_if = "HttpFilterConfig::is_filter_not_set")]
+    pub http_filter: HttpFilterConfig,
```
0x00A5: Just a note and don't need to address. The serde cases mismatch here. Config types are snake case, CRDs are camel case. Don't really matter though.

mirrord/operator/src/crd/preview.rs
iniw: This is actually only here to make the schema generated by this struct valid for k8s. See the comment above the field. It's not actually used for anything... We do the same thing for `SqsSessionStatus` that's where I got the fix/comment from https://github.com/metalbear-co/mirrord/blob/3f68b0c3eccc5b735c16ff25bc4e2841cd3c7fba/mirrord/operator/src/crd.rs#L690-L697

mirrord/cli/src/config.rs
iniw: I wanna keep it like this I think since `status` and `stop` are using the namespace to query, not necessarily to select a "target", which is what `start` uses it for. I don't know, it just feels better. Could I keep it like this?

mirrord/cli/src/preview.rs
iniw: This is a leftover from when I was just starting on the implementation and didn't know which arguments would be necessary. Removed it now.

mirrord/cli/src/preview.rs
iniw: Do we wanna do analytics for v1?

mirrord/cli/src/preview.rs
iniw: Done in https://github.com/metalbear-co/mirrord/pull/3917/changes/3500ced692365cd9c8ff24fbc4bcef31a26a2e72 Note that the feature is only added to `supported_features` when the license is on the enterprise tier, so this also solves https://github.com/metalbear-co/mirrord/pull/3917#discussion_r2775946126

iniw: See https://github.com/metalbear-co/mirrord/pull/3917#discussion_r2778176196

iniw: See https://github.com/metalbear-co/operator/pull/1224/changes/0ba5fe5f9b12b69d1f4beb434f66c8619688e8e3 and https://github.com/metalbear-co/charts/pull/324/changes/ec1e16908aeab4e0bfd45e5d46f52e8fe4699db3

mirrord/cli/src/preview.rs
iniw: Done in https://github.com/metalbear-co/mirrord/pull/3917/changes/57300ad56795da96e024e4cd442696fe0e5cd23c

mirrord/cli/src/preview.rs
iniw: Done in https://github.com/metalbear-co/mirrord/pull/3917/changes/57300ad56795da96e024e4cd442696fe0e5cd23c

mirrord/config/src/feature/preview.rs
iniw: Done in https://github.com/metalbear-co/mirrord/pull/3917/changes/57300ad56795da96e024e4cd442696fe0e5cd23c

mirrord/cli/src/preview.rs
0x00A5: Nothing essential I think. Can be added quickly in a followup PR.

mirrord/cli/src/config.rs
```diff
@@ -1165,6 +1169,280 @@ pub struct FixKubeconfig {
...
+
+    /// Load config from config file.
+    ///
+    /// When using -f flag without a value, defaults to "./.mirrord/mirrord.json"
+    #[arg(short = 'f', long, value_hint = ValueHint::FilePath, default_missing_value = "./.mirrord/mirrord.json", num_args = 0..=1)]
+    pub config_file: Option<PathBuf>,
+
+    /// Kube context to use from Kubeconfig.
+    #[arg(long)]
+    pub context: Option<String>,
```
Razz4780: I feel like these could be moved to the upper level from all three subcommands

mirrord/cli/src/error.rs
```diff
@@ -493,6 +493,78 @@ pub(crate) enum CliError {
...
+        Check the operator logs for more details.{GENERAL_HELP}"
+    ))]
+    PreviewTimeout,
+
+    #[error("Operator is required for preview environments")]
+    #[diagnostic(help(
+        "Preview environments require the mirrord operator. \
+        Please install the operator first: https://metalbear.com/mirrord/docs/overview/teams/"
+    ))]
+    OperatorRequiredForPreview,
```
Razz4780: This is the same as `CliError::OperatorNotInstalled`

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -185,6 +189,11 @@ impl HttpFilterConfig {
             || self.body_filter.is_some()
     }
 
+    pub fn is_filter_not_set(&self) -> bool {
```
Razz4780: Isn't this too much ;-;

mirrord/config/src/feature/network/incoming/http_filter.rs
```diff
@@ -271,6 +280,99 @@ impl HttpFilterConfig {
             }
         }
     }
+
+    /// Converts this config into the protocol-level [`HttpFilter`].
+    ///
+    /// Returns an error if a filter expression is invalid. Panics if no filter is set
+    /// (call [`is_filter_set`](Self::is_filter_set) first).
+    pub fn as_protocol_http_filter(&self) -> Result<HttpFilter, HttpFilterParseError> {
```
Razz4780, line 281: Yaaaay :monke-spin:

mirrord/config/src/feature/preview.rs
```diff
@@ -0,0 +1,49 @@
...
+///       "ttl_mins": 60,
+///       "creation_timeout_secs": 60
+///     }
+///   }
+/// }
+/// ```
+#[derive(MirrordConfig, Default, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
+#[config(map_to = "PreviewFileConfig", derive = "JsonSchema")]
+#[cfg_attr(test, config(derive = "PartialEq, Eq"))]
+pub struct PreviewConfig {
```
Razz4780, line 24: Hmm, do we need an entry in the mirrord config for this feature? Is this a requirement?

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+pub struct PreviewSession {
+    /// User's container image to run in the preview pod.
+    pub image: String,
+
+    /// Environment key used to group related preview pods and for traffic filtering.
+    pub key: String,
+
+    /// Target to copy pod configuration from (deployment, pod, statefulset, etc.).
+    /// The preview pod will be a copy of the target's pod spec with the user's image.
+    pub target: String,
```
Razz4780: Should be `SessionTarget`

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Incoming traffic configuration for the preview environment.
+    ///
+    /// Specifies which ports to steal/mirror traffic from and optional HTTP filters.
+    /// This configuration is extracted from the user's mirrord config.
+    pub incoming: PreviewIncomingConfig,
+
+    /// How long (in minutes) this session is allowed to live.
+    /// The operator will terminate the session when this time elapses.
+    pub ttl_mins: u64,
```
Razz4780: Maybe we can do seconds here? Might be handier in E2E, and it's not a user-facing field anyway

mirrord/operator/src/crd/preview.rs
Razz4780: Still, let's use `MicroTime`  : D

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+impl PreviewSession {
+    /// Parse the target string into a [`Target`].
+    pub fn parsed_target(&self) -> Result<Target, mirrord_config::config::ConfigError> {
+        Target::from_str(&self.target)
+    }
+}
+
+/// Status of a preview session resource.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
+pub enum PreviewSessionStatus {
```
Razz4780: I feel like this could be

```rust
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
pub struct PreviewSessionStatus {
    pub started_at: MicroTime,
    pub pod_name: Option<String>,
    pub failure_message: Option<String>,
}
```
Then, you'd be able to do merge patch on status, which has a nicer api in kube.rs and does not require resourceVersion check.
And the CLI can watch pod status independently

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Extracted from the user's mirrord config by the CLI and included in the CR.
+/// The operator uses this configuration to set up traffic stealing/mirroring from
+/// the original target to the preview pod.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct PreviewIncomingConfig {
+    /// Explicit set of ports to steal/mirror. When `None`, the operator discovers ports from
+    /// the preview pod's container port declarations.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub ports: Option<HashSet<u16>>,
```
Razz4780:
```suggestion
    pub ports: Option<Vec<u16>>,
```
I think it's nice to keep resource layout stable. Hashing collections will provide changing JSON representation, and can't be hashed themselves

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Whether to mirror or steal traffic from the target.
+    #[serde(default, skip_serializing_if = "IncomingMode::is_off")]
+    pub mode: IncomingMode,
```
Razz4780: Not a fan of using mirrord-config types in CRDs. CRDs have a very strict requirement of backwards compatibility (and yes, we sadly use them in some places atm 💀). Stricter than mirrord-config, because:
1. breaking compat can brick the operator and might not be easily fixable
2. k8s validation, unlike serde, does not like unknown fields

This could be just `pub steal: bool`. To handle disabled incoming, I'd make `PreviewSession::incoming` optional

mirrord/operator/src/crd/preview.rs
Razz4780: Another mirrord-config type used. This one's a bit more complex tho, ugh.
I'd consider passing it as string even

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,131 @@
...
+    group = "preview.mirrord.metalbear.co",
+    version = "v1alpha",
+    kind = "PreviewSession",
+    root = "PreviewSessionCrd",
+    status = "PreviewSessionStatus",
+    namespaced
+)]
+#[serde(rename_all = "camelCase")]
+pub struct PreviewSession {
+    /// User's container image to run in the preview pod.
```
Razz4780, line 31:
```suggestion
#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
#[kube(
    group = "preview.mirrord.metalbear.co",
    version = "v1alpha",
    kind = "PreviewSession",
    root = "PreviewSession",
    status = "PreviewSessionStatus",
    namespaced
)]
#[serde(rename_all = "camelCase")]
pub struct PreviewSessionSpec {
    /// User's container image to run in the preview pod.
```

`Crd` suffix should be forgotten, it's a lie and it never happened 😆

mirrord/config/src/feature/network/incoming/http_filter.rs
iniw: This only exists so that we can use it in `skip_serializing_if` without wrapping it in an `Option` (cause then we'd have two "null" states, `None` and `Some(filter_not_set)`). What do you think would be better?

mirrord/config/src/feature/preview.rs
iniw: I think so, it was requested by Liron https://metalbear.slack.com/archives/C0A0J9QEPC1/p1769427641332859?thread_ts=1768754546.323129&cid=C0A0J9QEPC1

mirrord/operator/src/crd/preview.rs
iniw: You mean serializing it into a json string or something using serde then deseriaizing on the operator? 💀

mirrord/operator/src/crd/preview.rs
iniw: But this is a bit more complicated because now the CLI would have to watch both the session resource (to check if `failure_message` is `Some`) and the preview pod (to check if it's ready), no?

mirrord/config/src/feature/network/incoming/http_filter.rs
Razz4780: Ah oke, nvm

mirrord/config/src/feature/preview.rs
Razz4780: Okok

mirrord/operator/src/crd/preview.rs
Razz4780: Hmhmhm. Adding `ready_at: Option<MicroTime>` is also a viable option

mirrord/operator/src/crd/preview.rs
Razz4780: Yeah 💀 
Unholy, but I'm really scared about the future

mirrord/operator/src/crd/preview.rs
iniw: I've decided to do it like this:
```rust
/// Status of a preview session resource.
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
pub struct PreviewSessionStatus {
    /// Current lifecycle phase of the session.
    pub phase: PreviewSessionPhase,

    /// Timestamp of when the operator started processing this session.
    pub started_at: MicroTime,

    /// Name of the preview pod created for this session.
    ///
    /// Set once the pod is created (from the `Waiting` phase onward). `None` during
    /// `Initializing` or if pod creation failed before a name was assigned.
    pub pod_name: Option<String>,

    /// Human-readable description of why the session failed.
    ///
    /// Only set when `phase` is `Failed`.
    pub failure_message: Option<String>,
}

/// Phase of a preview session's lifecycle.
///
/// Progresses through `Initializing` → `Waiting` → `Ready`. Any phase may transition to
/// `Failed` on error.
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Eq, PartialEq)]
pub enum PreviewSessionPhase {
    /// Operator is setting up — the preview pod has not been created yet.
    Initializing,
    /// Preview pod has been created, waiting for it to become ready.
    Waiting,
    /// Preview pod is running and traffic routing is active.
    Ready,
    /// Session has encountered an unrecoverable error.
    Failed,
}
```

It's heavily inspired by the way we did the mysql branching CR https://github.com/metalbear-co/mirrord/blob/5b816eb85586413e17d2d2163269c2c22fd9f449/mirrord/operator/src/crd/mysql_branching.rs#L67-L89

I also copied it's builder patcher thing on the operator, see https://github.com/metalbear-co/operator/blob/preview-env/operator/mysql-branching/src/controller/status.rs and https://github.com/metalbear-co/operator/blob/preview-env/operator/preview-env/src/status.rs

mirrord/operator/src/crd/session.rs
```diff
@@ -81,6 +95,52 @@ pub struct SessionTarget {
...
+        }
+    }
+}
+
+impl SessionTarget {
+    /// Parse into a [`Target`] by reconstructing the canonical target path string and parsing it.
+    pub fn as_target(&self) -> Result<Target, ConfigError> {
+        self.to_string().parse()
+    }
+}
```
Razz4780, line 175: A couple issues here, this should be correct:

```rust
impl SessionTarget {
    /// Returns [`None`] when [`Target`] has no target container.
    pub fn from_config(target: Target) -> Option<Self> {
        match target {
            Target::Deployment(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Deployment>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.deployment,
                    container: target.container?,
                })
            }
            Target::Pod(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Pod>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.pod,
                    container: target.container?,
                })
            }
            Target::Rollout(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Rollout>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.rollout,
                    container: target.container?,
                })
            }
            Target::StatefulSet(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<StatefulSet>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.stateful_set,
                    container: target.container?,
                })
            }
            Target::Service(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Service>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.service,
                    container: target.container?,
                })
            }
            Target::ReplicaSet(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<ReplicaSet>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.replica_set,
                    container: target.container?,
                })
            }
            Target::Job(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Job>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.job,
                    container: target.container?,
                })
            }
            Target::CronJob(target) => {
                let TypeMeta { api_version, kind } = TypeMeta::resource::<Job>();
                Some(Self {
                    api_version,
                    kind,
                    name: target.cron_job,
                    container: target.container?,
                })
            }
            Target::Targetless => None,
        }
    }

    /// Returns [`None`] if the referenced resource does not correspond to any known target type.
    pub fn into_config(self) -> Option<Target> {
        format!(
            "{}/{}/container/{}",
            self.kind.to_ascii_lowercase(),
            self.name,
            self.container
        )
        .parse()
        .ok()
    }
}
```

mirrord/operator/src/crd/session.rs
```diff
@@ -81,6 +95,52 @@ pub struct SessionTarget {
     pub container: String,
 }
 
+impl fmt::Display for SessionTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}/{}", self.kind, self.name)?;
+        if !self.container.is_empty() {
+            write!(f, "/container/{}", self.container)?;
+        }
```
Razz4780, line 101: `SessionTarget::container` should always be valid. If it was optional, it would be an `Option`  : DD 👉 👉 

If the use did not specify the container, you gotta resolve it in the CLI, like we do in `mirrord exec`

mirrord/operator/src/crd/session.rs
```diff
@@ -81,6 +95,52 @@ pub struct SessionTarget {
...
+        let api_version = match target {
+            Target::CronJob(_) => <CronJob as Resource>::API_VERSION,
+            Target::Deployment(_) => <Deployment as Resource>::API_VERSION,
+            Target::Job(_) => <Job as Resource>::API_VERSION,
+            Target::Pod(_) => <Pod as Resource>::API_VERSION,
+            Target::Service(_) => <Service as Resource>::API_VERSION,
+            Target::StatefulSet(_) => <StatefulSet as Resource>::API_VERSION,
+            Target::ReplicaSet(_) => <ReplicaSet as Resource>::API_VERSION,
+            Target::Targetless => "",
+            #[cfg(feature = "client")]
```
Razz4780: You can add the dep, we use mirrord-kube everywhere anyway

mirrord/cli/src/preview.rs
```diff
@@ -0,0 +1,520 @@
...
+        let namespaced_api = Api::<PreviewSession>::namespaced(client.clone(), namespace);
+
+        if let Err(e) = namespaced_api.delete(name, &DeleteParams::default()).await {
+            delete_subtask.failure(None);
+            return Err(CliError::PreviewDeleteFailed {
+                name: name.to_owned(),
+                reason: e.to_string(),
+            });
+        }
+    }
```
Razz4780, line 482:
```suggestion
    let mut result: CliResult<()> = Ok(());
    for session in sessions_to_delete {
        let name = session
            .metadata
            .name
            .as_deref()
            .expect("preview session should have a name");
        let namespace = session
            .metadata
            .namespace
            .as_deref()
            .expect("preview session should have a namespace");

        let namespaced_api = Api::<PreviewSession>::namespaced(client.clone(), namespace);

        if let Err(e) = namespaced_api.delete(name, &DeleteParams::default()).await {
            result = Err(e);
        }
    }
    if result.is_err() {
        delete_subtask.failure(None);
        return result;
    }
```

Nit: we can still try to delete the other sessions

mirrord/operator/src/crd/preview.rs
```diff
@@ -0,0 +1,142 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, Eq, PartialEq)]
+pub enum PreviewSessionPhase {
+    /// Operator is setting up — the preview pod has not been created yet.
+    Initializing,
+    /// Preview pod has been created, waiting for it to become ready.
+    Waiting,
+    /// Preview pod is running and traffic routing is active.
+    Ready,
+    /// Session has encountered an unrecoverable error.
+    Failed,
```
Razz4780, line 98:
```suggestion
    Failed,
    /// For future compatibility.
    #[serde(other)]
    Unknown,
```

Razz4780 (approved):

changelog.d/+preview-env.added.md
iniw: Removed this link entirely since we want to merge this PR before the docs PR :)

PR #3919, opened 2026-02-02
Fix `mirrord db-branches status` command

t4lz:
Closes INT-293

https://linear.app/metalbear/issue/INT-293/mirrord-db-branches-a-status-gives-404

The `mirrord db-branches status` fails when one of the DB branching features (PostgreSQL / MySQL) is not enabled on the cluster, and therefore the CRD isn't defined.

The solution in this PR is not to error out if the list call returns a 404, which is what is returned when that CRD is not defined.

Additionally, I've done some refactoring to eliminate duplicate code (and also some small optimizations just for fun - less memory allocation, less cloning, etc.).

I recommend reviewing this PR commit-by-commit, as it will make it easier to see what was done exactly and to separate the refactoring from the fix.

### DIFF
--- a/changelog.d/+db-branches-status.fixed.md
+++ b/changelog.d/+db-branches-status.fixed.md
@@ -1,0 +1,1 @@
+The db-branches status command would fail when one of the branching features (MySQL/PostgreSQL) was not enabled on the cluster.


mirrord/cli/src/db_branches.rs
```diff
@@ -52,6 +53,37 @@ impl From<&MysqlBranchDatabase> for BranchInfo {
     }
 }
 
+impl From<MysqlBranchDatabase> for BranchInfo {
```
t4lz, line 54: We already have `impl From<&MysqlBranchDatabase> for BranchInfo`, I added this to be able to move resources out of the object instead of cloning them.

Same for `impl From<PgBranchDatabase> for BranchInfo` below

vladrbg (approved):
LGTM, I like that you also removed a lot of the duplicate logic and it's cleaner now 🚀

PR #3921, opened 2026-02-04
Fix double listen making port unsubscribed. Happens in Rails.

aviramha:


mirrord/layer/src/socket/ops.rs
```diff
@@ -443,10 +443,22 @@ pub(super) fn listen(sockfd: RawFd, backlog: c_int) -> Detour<i32> {
...
+        SocketState::Listening(_) => {
+            tracing::debug!("second listen called");
+            let listen_result = unsafe { FN_LISTEN(sockfd, backlog) };
+            if listen_result != 0 {
+                let error = io::Error::last_os_error();
+                error!("listen -> Failed `listen` sockfd {:#?}", sockfd);
+                Err(error)?
+            }
+
+            SOCKETS.lock()?.insert(sockfd, socket);
```
Razz4780: Shouldn't we make PortUnsubscribe + PortSubscribe in this case?

Razz4780: Ah nvm

Razz4780: We should probably return the socket to `SOCKETS` regardless of the listen result. If it failed, the socket is still listening

aviramha: done

PR #3925, opened 2026-02-05
WIN-76 Layer Unification - detour infra

itsamegraf:


mirrord/layer/src/file/ops.rs
```diff
@@ -949,12 +952,16 @@ mod test {
         Success,
     }
 
-    impl<S> Detour<S> {
+    trait DetourKindExt {
```
itsamegraf, line 955: this is a workaround for "error[E0117]: only traits defined in the current crate can be implemented for arbitrary types"

mirrord/layer-lib/src/socket/apple_dnsinfo.rs
itsamegraf, line 1: this needed to move because of `impl From<HookError> for *mut dns_config_t` that must be in the same crate as HookError, if reviewer wants I can change it to an extension trait and keep apple_dnsinfo in layer (more suitable)

PR #3928, opened 2026-02-05
Fix sockets regressions

Razz4780:
See commit history, most of the changes are reverts

### DIFF
--- a/mirrord/layer/tests/apps/dup_listen/src/main.rs
+++ b/mirrord/layer/tests/apps/dup_listen/src/main.rs
@@ -1,9 +1,11 @@
+#[cfg(target_family = "unix")]
 use std::{
     io::Read,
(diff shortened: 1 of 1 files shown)


mirrord/layer/src/lib.rs
```diff
@@ -994,14 +907,7 @@ pub(crate) unsafe extern "C" fn vfork_detour() -> pid_t {
         }
     };
 
-    // Drop the `DetourGuard` so we actually fork_detour.
-    drop(guard);
```
0x00A5, line 998: Maybe we want to do this in a separate PR. But fork detour here is called when the hook is already guarded, so the FN_FORK is called.

Razz4780: Yep, we'll want to restore that :V

PR #3939, opened 2026-02-12
Revert "Revert arm64e sip (#3703)"

aviramha:
Bring back arm64e non-emulation support!

### DIFF
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -333,11 +333,11 @@ mod main {
         // If we're executing an arm64e binary we need to patch the ptr auth ABI version to allow
         // executing third party binaries.
         //
-        // This can be analyzed by first stripping a properly codesigned arm64e binary out of it's
+        // This can be analyzed by first stripping a properly codesigned arm64e binary out of its
         // fat mach-o image:
         // $ lipo /bin/sh -thin arm64e -output hax-sh
         //
-        // Then looking at it's header using something like otool/objdump:
+        // Then looking at its header using something like otool/objdump:
         // $ otool -vh hax-sh
         // Mach header
         //       magic cputype cpusubtype  caps filetype ncmds sizeofcmds flags


mirrord/layer/src/common.rs
```diff
@@ -92,9 +96,12 @@ pub fn strip_mirrord_path(path_str: &str) -> Option<&str> {
     // SAFETY: We only slice after we find the string in the path
     // so it must be valid
     #[allow(clippy::indexing_slicing)]
-    path_str
-        .find(MIRRORD_PATCH_DIR)
-        .map(|index| &path_str[(MIRRORD_PATCH_DIR.len() + index)..])
+    path_str.find(MIRRORD_PATCH_DIR).map(|index| {
+        let remain = &path_str[(MIRRORD_PATCH_DIR.len() + index)..];
```
Razz4780, line 100: What is going on here, why aren't we just doing `strip_prefix`? 😕

mirrord/layer/src/common.rs
```diff
@@ -7,6 +7,10 @@ use mirrord_protocol::file::OpenOptionsInternal;
 use null_terminated::Nul;
 use tracing::warn;
 
+/// Make it const so can be shared and re-used with by reference without fear.
+#[cfg(target_os = "macos")]
+const ROOT_DIR: &str = "/\0";
```
Razz4780, line 12: Why does this need a trailing nul byte?

mirrord/layer/src/common.rs
aviramha: The entire SIP string handling is cursed, ask @gememma :D
I think that we were afraid of having different address because we handle memory. I think strip_prefix is safe but wouldn't want to touch it..

mirrord/layer/src/common.rs
aviramha: SIP/layer uses str/String instead of using CStr/OsPath etc... when I do just "/" &str isn't null terminated by default, so it leads to huge string (ENAMELONG) error.

mirrord/layer/src/common.rs
Razz4780: oke

mirrord/layer/src/common.rs
Razz4780: Oh lord...

mirrord/layer/src/common.rs
gememma: looks like it was this change - https://github.com/aviramha/mirrord/commit/412426e7f5df54af6e4ed8eb834b1cf9e755d402#diff-5610448ebdfac5c7ed69bc14959adddc1d42942290dbe370842e282412ff013d

mirrord/sip/src/lib.rs
```diff
@@ -344,18 +319,52 @@ mod main {
...
         let binary = data
-            .get(binary_info.offset..binary_info.offset + binary_info.size)
+            .get_mut(binary_info.offset..binary_info.offset + binary_info.size)
             .expect("invalid SIP binary");
 
-        std::fs::write(&temp_binary, binary)?;
+        // If we're executing an arm64e binary we need to patch the ptr auth ABI version to allow
+        // executing third party binaries.
+        //
+        // This can be analyzed by first stripping a properly codesigned arm64e binary out of it's
```
gememma:
```suggestion
        // This can be analyzed by first stripping a properly codesigned arm64e binary out of its
```
(sorry)

mirrord/layer/src/common.rs
aviramha: me and @t4lz wrecking havoc on codebase :D

mirrord/sip/src/lib.rs
```diff
@@ -344,18 +319,52 @@ mod main {
...
+        // fat mach-o image:
+        // $ lipo /bin/sh -thin arm64e -output hax-sh
+        //
+        // Then looking at it's header using something like otool/objdump:
```
gememma:
```suggestion
        // Then looking at its header using something like otool/objdump:
```
(sorRY)

mirrord/sip/src/lib.rs
```diff
@@ -344,18 +319,52 @@ mod main {
...
 
         // Just the thin binary - if the file was a thin binary of a supported architecture to
         // begin with then it's the whole file, if its a fat binary then it's just a part of it.
         let binary = data
-            .get(binary_info.offset..binary_info.offset + binary_info.size)
+            .get_mut(binary_info.offset..binary_info.offset + binary_info.size)
             .expect("invalid SIP binary");
 
-        std::fs::write(&temp_binary, binary)?;
+        // If we're executing an arm64e binary we need to patch the ptr auth ABI version to allow
```
gememma, line 333: I like this big verbose comment with references very much 🌻 [peace and love on planet earth](https://github.com/metalbear-co/mirrord/blob/main/CODE_REVIEW.md#highlighting-positives)

gememma (approved):
Yep, I think this is good 🫡 SIP gods be kind 🙏

mirrord/layer/src/common.rs
t4lz: `MIRRORD_PATCH_DIR` is not a path, it's just a dirname, so it's not the prefix.

aviramha: yeah but it's str so strip_prefix would've worked, wouldn't it?

PR #3940, opened 2026-02-12
xtask + release speedup

aviramha:


xtask/src/tasks/release.rs
```diff
@@ -0,0 +1,165 @@
...
+
+impl Default for BuildOptions {
+    fn default() -> Self {
+        Self {
+            platform: Platform::detect().unwrap_or(Platform::MacosUniversal),
+            release: true,
+            with_wizard: true,
+        }
+    }
+}
```
iniw: Defaulting to `MacosUniversal` is super sketchy, but turns out this is entirely unused so we can just remove it :)

iniw (approved):
This could've been a nushell or python script but alright, this works

xtask/src/tasks/release.rs
aviramha: Thanks for the catch :)

iniw (dismissed):

PR #3942, opened 2026-02-12
INT-295 Multi Cluster - mirrord CLI operator status 

vladrbg:
image

Btw not sure if we want the clusters field it might be repetitive.


mirrord/cli/src/operator/status.rs
```diff
@@ -367,6 +367,41 @@ Operator License
         sessions.printstd();
         println!();
 
+        // Multi-cluster sessions are only on Primary cluster.
+        if let Some(mc_sessions) = status.multi_cluster_sessions.as_ref() {
+            if mc_sessions.is_empty() {
+                println!("No active multi-cluster sessions.");
```
DmitryDodzin: I think it's fine to print an empty table when the array is empty instead of writing a different message since we don't do it for the sessions and sqs queues

vladrbg: image

DmitryDodzin (approved):

PR #3805, opened 2025-12-19
WIN-67 Unify LayerSetup + some Socket operations and utils

itsamegraf:
This PR unifies layer crate setup functionality, error handling, and proxy connection management into mirrord_layer_lib. Windows-specific socket and setup infrastructure are moved to platform-agnostic implementations. The layer modules are refactored to depend on centralized exports, consolidating shared logic and reducing duplication across platform-specific implementations.

taken from RabbitAI summary of the PR 🤷

PR #3946, opened 2026-02-13
WIN-84 Layer Unification - detour infra + socket example

itsamegraf:
~~relies on https://github.com/metalbear-co/mirrord/pull/3805, see https://github.com/itsamegraf/mirrord/pull/4/ for clean PR diff~~
___

Make Detour helper traits and ergonomics fully cross-platform so they work cleanly on Windows, removing Unix/Windows-specific forks.

Introduce a shared socket helper in layer-lib and unify Unix + Windows socket detours around it as the first real consumer of the cross-platform Detour API.

This reduces platform divergence, simplifies detour code, and establishes a pattern for future hook unification.

### DIFF
--- a/mirrord/layer-lib/src/detour.rs
+++ b/mirrord/layer-lib/src/detour.rs
@@ -471,7 +471,7 @@ pub trait OptionDetourExt<T>: OptionExt {
     /// use mirrord_layer_lib::detour::{Detour, OptionDetourExt};
     /// let result: Option<Detour<i32>> = Some(Detour::Success(5));
     /// match result.transpose() {
-    ///     Detour::Success(inner) => assert!(inner, Some(5)),
+    ///     Detour::Success(inner) => assert_eq!(inner, Some(5)),
     ///     _ => unreachable!(),
     /// };
     /// ```


.github/workflows/ci.yaml
```diff
@@ -631,7 +631,7 @@ jobs:
       - name: mirrord layer tests
         run: cargo test -p mirrord-layer
       - name: mirrord layer-lib tests
-        run: cargo test -p mirrord-layer-lib
+        run: cargo test -p mirrord-layer-lib --lib --bins --tests
```
itsamegraf: purpose - dont run doctest code.
The reason we need it here is because some doctests' visibility was upgraded from `pub(crate)` in layer, where they are not checked by clippy, to `pub` in layer-lib, where they do. adding these flags makes them be ignored.

0x00A5: why doctests are failing tho?

mirrord/layer-lib/src/socket/sockets.rs
```diff
@@ -265,11 +265,11 @@ pub fn remove_socket(socket: SocketDescriptor) {
 }
 
 /// Get socket info
-pub fn get_socket(socket: SocketDescriptor) -> Option<Arc<UserSocket>> {
+pub fn get_socket(socket: SocketDescriptor) -> Option<UserSocket> {
```
0x00A5: Is this a windows specific pub function?

mirrord/layer-lib/src/socket/sockets.rs
```diff
@@ -265,11 +265,11 @@ pub fn remove_socket(socket: SocketDescriptor) {
...
     SOCKETS
         .lock()
         .ok()
-        .and_then(|sockets| sockets.get(&socket).cloned())
+        .and_then(|sockets| sockets.get(&socket).map(|s| s.as_ref().clone()))
```
0x00A5: Do we need clone here? It seems a arc could be enough where this function is used (I only found one place in layer win)

.github/workflows/ci.yaml
itsamegraf: tl;dr - this change is unneeded (for macos), but we might want to revert an equivalent linux change I introduced in the previous Layersetup PR but it will require to do something with the doctest..

Here's the long story (not) short:
In the [previous LayerSetup PR](https://github.com/metalbear-co/mirrord/pull/3805) we had the [following change](https://github.com/metalbear-co/mirrord/pull/3805/changes#diff-96d21e489d9975c5e8125e66762b96865a218859f9306f431fe0a82b608b176bR470):
image

Before, there was a typo `Detour::Sucess` which was never caught because it never ran -(why?)->the visibility was `pub(crate)` -> only accesible in layer -> so never checked as they were public for cargo/clippy/not-sure-which-component to actually check it. ([source](https://users.rust-lang.org/t/doctest-of-private-functions/60805/2))
This extension is used only by linux so I gated it to `target_os = linux` and in the previous PR I already introduced this `--lib --bin --tests` workaround in linux `ci.yaml` for `layer-lib`, so it doesnt check it as well.
image
So eventually I fixed the typo (in the previous PR) but then it failed because we cannot `assert_eq(Option, Option)`
image
I could fix it by doing `derive(PartialEq)` on Detour, but it becomes a domino chain Detour->HookError->{All enum in error.rs practically}, while trying to do that I eventually hit EncodeError which is missing it and is out of my control
image

because this doctest is guarded with `cfg(target_os = linux)`, which already has this `--lib --bin --tests`, I can actually remove the current addition, but I would like your opinion on whether I should revert those flags from the linux version of `cargo test` - should i revert it and continue down the rabbithole of making this doctest pass? maybe just remove that damned doctest?

mirrord/layer-lib/src/socket/sockets.rs
itsamegraf: yes and it's really just used once in closesocket, I will remove it together with `remove_socket`, residues from the old windows-specific implemenntation, and align the impl with close/close_layer_fd of layer. anyway this will be unified soon as part of the unification of all socketops

mirrord/layer-lib/src/socket/sockets.rs
itsamegraf: removed the fn :)

.github/workflows/ci.yaml
itsamegraf: btw apparently this is the only culprit doctest, removing it makes doctests pass
image

0x00A5: I see, thanks for the explanation. How about replacing the problematic `assert_eq!` with something like:
```
    /// ```no_run
    /// use mirrord_layer_lib::detour::{Detour, OptionDetourExt};
    /// let result: Option<Detour<i32>> = Some(Detour::Success(5));
    /// match result.transpose() {
    ///     Detour::Success(inner) => assert!(inner, Some(5)),
    ///     _ => unreachable!(),
    /// };
    /// ```
```

I'd rather keep running doctest for everything.

itsamegraf: Should've thought of that, thank you 🙏  will incorporate

PR #3947, opened 2026-02-13
rollback loading into arm64e, fixes many bugs on macOS

aviramha:
:(


changelog.d/+rollback-sip.fixed.md
```diff
@@ -0,0 +1 @@
+rollback loading into arm64e, fixes many bugs on macOS
```
gememma:
```suggestion
Rolled back the change for loading into arm64e, fixing many bugs on macOS.
```

PR #3951, opened 2026-02-15
Inject mirrord session key headers into requests

MintSoup:
### Quality Checklist:

- [X] ~I have tested this change and know it succeeds and fails as expected~
- [X] ~I have written unit tests or purposefully omitted them~
- [X] ~I have written e2e tests or purposefully omitted them~
  documentation)

### DIFF
--- a/mirrord/cli/src/main.rs
+++ b/mirrord/cli/src/main.rs
@@ -687,7 +687,13 @@ pub(crate) fn print_config<P>(
         config.internal_proxy.log_destination.display()
     ));
 
-    progress.info(&format!("key: {}", config.key.as_str()));
+    if operator_used {
+        progress.info(&format!(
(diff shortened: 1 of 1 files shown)


MintSoup:
> I don't think you need to add the key to CRDs.
> 
> Regardless, adding mandatory string fields would break backwards compatibility :V

How else would it work then? Make it optional?

Razz4780:
> > I don't think you need to add the key to CRDs.
> > Regardless, adding mandatory string fields would break backwards compatibility :V
> 
> How else would it work then? Make it optional?

Yeah, optional would work

Razz4780 (changes requested):
Ticket says that CLI should print the header:
```The CLI prints the header name + value when the session starts (so the user can search logs/traces).```

PR #3954, opened 2026-02-17
change to lint all platforms from one job

aviramha:
This became longer and bigger than I thought.
It was an emotional roller coaster of trying to understand what is actually needed or not, but in general the idea is to reduce unneeded builds/compilations in different jobs and concurrently lint all platforms together in one place.


MintSoup (dismissed):
LGTM

PR #3968, opened 2026-02-22
Take jq programs to split SQS queues

t4lz:
Closes INT-240

Adds a new config field for jq filters.
Also adds it in `ConnectParams`, where we send the sqs filters to the operator, and in the `MirrordSqsSession`, which is a split request by the main operator to the SQS session operator.
Added new fields and not under the existing `sqs_splits` etc. fields, for backward compatibility.
It has the disadvantage that we have to do 2 hashmap lookups - 1 in the old map, one in the new jq map, but I think it's unavoidable.

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/operator/src/crd.rs
+++ b/mirrord/operator/src/crd.rs
@@ -908,11 +908,12 @@ mod tests {
     use crate::crd::{
         MirrordClusterOperatorUserCredential, MirrordOperatorCrd, MirrordSqsSession,
         MirrordWorkloadQueueRegistry, SessionCrd,
+        db_branching::{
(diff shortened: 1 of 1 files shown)


mirrord/config/src/feature/split_queues.rs
```diff
@@ -53,46 +54,104 @@ impl SplitQueuesConfig {
...
+                    name.clone(),
+                    error.into(),
+                )
+            })?;
+        }
+        Ok(())
+    }
+
+    fn verify_jq_program(
+        queue_id: &str,
```
vladrbg, line 112: Nit: maybe we can use QueueId for both?

PR #3969, opened 2026-02-22
cli+operator/crd: show remaining TTL in `preview status`

iniw:
### Quality Checklist:

  documentation)

- Related PRs:
  - https://github.com/metalbear-co/operator/pull/1302
  - https://github.com/metalbear-co/charts/pull/344

### DIFF
--- a/mirrord/cli/src/preview.rs
+++ b/mirrord/cli/src/preview.rs
@@ -13,6 +13,7 @@ use std::{
 };
 
 use futures::StreamExt;
+use k8s_openapi::chrono::Utc;
 use kube::{
     Api, ResourceExt,
     api::{DeleteParams, ListParams, ObjectMeta, PostParams},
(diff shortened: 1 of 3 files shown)


mirrord/cli/src/preview.rs
```diff
@@ -370,14 +371,25 @@ async fn preview_status(args: PreviewStatusArgs) -> CliResult<()> {
             let status = match session.status.as_ref().map(|s| &s.phase) {
                 Some(PreviewSessionPhase::Initializing) => "initializing".to_owned(),
                 Some(PreviewSessionPhase::Waiting) => "waiting".to_owned(),
-                Some(PreviewSessionPhase::Ready) => "running".to_owned(),
+                Some(PreviewSessionPhase::Ready) => {
+                    let remaining = session
+                        .status
+                        .as_ref()
+                        .and_then(|s| s.expires_at.as_ref())
+                        .and_then(|expires_at| (expires_at.0 - Utc::now()).to_std().ok())
```
0x00A5, line 379: Do we want to use [saturating_sub](https://doc.rust-lang.org/nightly/core/time/struct.Duration.html#method.saturating_sub), so we display 0 instead of negative duration for some weird edge case

mirrord/operator/src/crd/preview.rs
0x00A5, line 1: Update crd yaml in charts

mirrord/cli/src/preview.rs
iniw: These are actually [`DateTime`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html), not `Duration`, so `saturating_sub` doesn't really make sense as an operation.

We do handle the negative duration case naturally though because of the `to_std()` call - `std::time::Duration` doesn't represent negative durations while `chrono::Duration` does, which is why `to_std()` returns `Result<std::time::Duration, OutOfRangeError>`. When we get a negative duration `ok()` will return `None`, making `remaining` be `None`, so we just render `running` and don't show the TTL at all.

mirrord/operator/src/crd/preview.rs
iniw: i fogor 💀

iniw: https://github.com/metalbear-co/charts/pull/344 👍

PR #3974, opened 2026-02-24
Allow port auto-detection on `mirrord dump` cmd

gememma:
### Quality Checklist:

  documentation)

### PR Details
- `mirrord dump` will now subscribe to all auto-detected ports in a target when `--ports` is not provided
- It creates a kube `Client` up-front to do this, and the resolves the target from the cluster
- If no ports were found, the command fails
- The command still respects targets requiring the operator and the target being required as before
- If ports were not provided, progress prints "Using auto-detected ports on target"

### DIFF
--- a/mirrord/cli/src/dump.rs
+++ b/mirrord/cli/src/dump.rs
@@ -110,9 +110,9 @@ pub async fn dump_command(
     };
 
     if ports.is_empty() {
-        Err(DumpSessionError::PortDetectionFailed(
-            "no ports found in the target's resource spec".to_string(),
-        ))?;
+        Err(DumpSessionError::PortDetectionFailed(format!(
+            "no ports found in the resource spec for target {path}"
+        )))?;
     }
 
     // Start the dump session


changelog.d/+autodetect-dump-ports.added.md
```diff
@@ -0,0 +1 @@
+`mirrord dump` no longer requires `--ports` to be specified manually - they will be auto-detected if absent.
```
hank-metalbear, line 1: I like that it's backwards compatible

mirrord/cli/src/dump.rs
```diff
@@ -69,8 +81,42 @@ pub async fn dump_command(
...
+    // If the user didn't specify ports, detect them on the target
+    let ports = if args.ports.is_empty() {
+        progress.info("Using auto-detected ports on target");
+        let client = kube_client_from_layer_config(&config).await?;
+
+        let resolved = ResolvedTarget::new(&client, &path, namespace.as_deref())
+            .await
+            .map_err(|error| {
+                DumpSessionError::PortDetectionFailed(format!("failed to resolve target: {error}"))
+            })?;
```
hank-metalbear, line 93: Nit: consider adding `tracing::warn!` before mapping to `DumpSessionError` on the error paths here and in `resolve_pod_spec` below. Would help with debugging when users hit auto-detection issues:

```rust
.map_err(|error| {
    tracing::warn!(%error, "failed to resolve target for port auto-detection");
    DumpSessionError::PortDetectionFailed(format!("failed to resolve target: {error}"))
})?;
```

Not blocking, just useful for observability.

mirrord/cli/src/dump.rs
```diff
@@ -69,8 +81,42 @@ pub async fn dump_command(
...
+            })
+            .unwrap_or_default()
+    } else {
+        args.ports.clone()
+    };
+
+    if ports.is_empty() {
+        Err(DumpSessionError::PortDetectionFailed(
+            "no ports found in the target's resource spec".to_string(),
+        ))?;
```
hank-metalbear: Same here, a `tracing::warn!` with the target path would help:

```rust
tracing::warn!(?path, "no ports found in target's resource spec");
```

mirrord/cli/src/dump.rs
gememma: since we immediately exit (`.map_err(...)?`), the user will see this error printed in full - no need to print twice

mirrord/cli/src/dump.rs
gememma: again, no need to print twice, but i'll add the target name to the error message 🫡

mirrord/cli/src/dump.rs
```diff
@@ -69,8 +81,42 @@ pub async fn dump_command(
...
+        let resolved = ResolvedTarget::new(&client, &path, namespace.as_deref())
+            .await
+            .map_err(|error| {
+                DumpSessionError::PortDetectionFailed(format!("failed to resolve target: {error}"))
+            })?;
+        let pod_spec = resolved.resolve_pod_spec(&client).await.map_err(|error| {
+            DumpSessionError::PortDetectionFailed(format!(
+                "failed to resolve target pod spec: {error}"
+            ))
+        })?;
```
iniw, line 98: Do we need to treat rollouts especially here? See how we do it in preview env for example: https://github.com/metalbear-co/operator/blob/39f30886330a07dc4a7729a7169f08b16cd7c28a/operator/preview-env/src/task.rs#L142-L152

gememma: Screenshot 2026-02-24 at 18 00 55

tadaaaaa 🎉 magical function that does it for me

iniw: Aw man I gotta get on this IMMEDIATELY 💀 
https://github.com/metalbear-co/mirrord/blob/54feba7028a9a6379af0e2e84042b525ce46c5ec/mirrord/cli/src/preview.rs#L87-L103

I didn't know we could resolve the target so easily from the CLI

gememma: ikr, i wrote so much code that got deleted when Yuan reminded me this existed 🙏 🐸

mirrord/cli/src/dump.rs
```diff
@@ -52,12 +60,16 @@ pub async fn dump_command(
     );
 
     // Ensure a target was specified
-    if matches!(config.target.path, Some(Target::Targetless)) || config.target.path.is_none() {
```
MintSoup, line 55: FWIW you can also do `matches!(..., Some(...) | None)` if you'd like

gememma: 9 lines vs 8 lines... alex has taught you well 🙏 (jk i will remember that for the future 🧠 )

PR #3984, opened 2026-02-25
WIN-92 move run_command to mirrord-test-utils

itsamegraf:
This PR is a preparatory PR for moving run_command utility into `test-utils` crate, without forcing compilation of `mirrord` crate, see [more details in the ticket](https://linear.app/metalbear/issue/WIN-92/move-run-command-utility-to-mirrord-test-utils)


itsamegraf:
> Note that this will require changes in the operator. See [this test](https://github.com/metalbear-co/operator/blob/f0a4f7928618606e2eaf18a3ab573ebe8600c010/tests/src/operator/ha/operator_restart.rs#L334-L341) example

good call, thank you for noticing that, I frankly had no idea 😅 
Opened the [operator PR](https://github.com/metalbear-co/operator/pull/1319), reached out in DM for some assistance with them

PR #3986, opened 2026-02-25
WIN-87 exec-based test harness

itsamegraf:
Adds an exec-based test harness for layer integration tests by allowing mirrord exec to target a pre-spawned internal proxy (`MIRRORD_TEST_INTPROXY_ADDR`), and introduces mirrord-layer-tests as the harness crate. CI now builds the CLI/layer artifacts and runs the unified test suite across Linux/macOS/Windows.

This is an alternative to the existing CLI-mimicking harness, closer to real CLI execution while avoiding full agent startup and reducing harness complexity.

Cross-platform layer integration tests will live in the new `layer-tests` crate. The remaining tests will be migrated as part of [WIN-93](https://linear.app/metalbear/issue/WIN-93); for now, outgoing TCP integration tests are copied there as a proof of concept.

__
~relies on https://github.com/metalbear-co/mirrord/pull/3984~

___

### Quality Checklist:

- ~I have written user-facing website docs for new features, or opened a (sub)issue to do so~
- ~I have checked and updated existing website docs for changed features~
  documentation)

### DIFF
--- a/mirrord/layer-tests/tests/common/mod.rs
+++ b/mirrord/layer-tests/tests/common/mod.rs
@@ -154,12 +154,10 @@ pub fn get_env(
         .map(|(key, value)| (key.to_string(), value.to_string()))
         .collect::<Vec<_>>();
 
-    [
-        (
-            MIRRORD_TEST_INTPROXY_ADDR.to_string(),
-            intproxy_addr.to_string(),
-        ),
-    ]
+    [(
+        MIRRORD_TEST_INTPROXY_ADDR.to_string(),
+        intproxy_addr.to_string(),
+    )]
     .into_iter()
     .chain(extra_vars_owned)
     .collect()


mirrord/layer-lib/src/file/windows/read_local_by_default.rs
```diff
@@ -10,6 +10,7 @@ use str_win::path_to_unix_path;
 pub fn regex_set_builder() -> RegexSetBuilder {
     let mut patterns: Vec<String> = [
         r".\.dll$",
+        r".\.pdb$",
```
itsamegraf, line 13: read_local_by_default additions introduced to not break TestIntProxy tests for getting the wrong messages (Remote file reads) during integration tests

mirrord/layer/tests/common/mod.rs
```diff
@@ -118,795 +90,6 @@ pub fn init_tracing() -> DefaultGuard {
     }
 }
 
-pub struct TestIntProxy {
```
itsamegraf, line 121: moved entirely as-is to `mirrord-layer-tests::intproxy` mod

mirrord/layer-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,170 @@
+use std::{
```
itsamegraf, line 1: this file is the `exec`-based version of layer's test harness in common/mod.rs
Should be almost entirely the same except for:
* having a single outgoing tcp poc test
* no layer mimic operations, only calls run_command with TEST_INTPROXY envvar as test-harness
As part of WIN-72 all layer IT will move to layer-tests crate and use this test harness

.cargo/config.toml
```diff
@@ -18,8 +18,3 @@ debug = 2  # full debug info
 overflow-checks = true
 debug-assertions = true
 incremental = false
-
```
itsamegraf, line 21: this windows-only stack enlargement is not really needed since some cli stuff got Box'd a couple of months ago, tested as working without it

mirrord/layer-tests/src/intproxy.rs
```diff
@@ -0,0 +1,821 @@
+use core::assert_matches;
```
itsamegraf, line 1: this file is entirely a copy of the old TestIntProxy from `layer/tests/common/mod.rs`, no changes introduced.
for yuan, includes the first-connect timeout enlargement introduced to solve bash flakey

mirrord/layer-tests/tests/outgoing.rs
```diff
@@ -0,0 +1,104 @@
+#![warn(clippy::indexing_slicing)]
```
itsamegraf, line 1: note: this file should be excatly the same as `layer/tests/outgoing.rs` (with udp stuff redacted for not being part of the poc scope of this PR)

mirrord/layer/Cargo.toml
```diff
@@ -58,6 +58,7 @@ futures.workspace = true
 rstest.workspace = true
 serde_json.workspace = true
 tempfile.workspace = true
+mirrord-layer-tests = { path = "../layer-tests", default-features = false }
```
itsamegraf, line 61: default features turned off as integration tests of layer should never compile mirrord ci - they do not use the "exec-based" test harness

.cargo/config.toml
itsamegraf: that didnt age well... e2e is oveflowed in BULKS after this change :) reverted

mirrord/cli/build.rs
```diff
@@ -9,31 +7,17 @@ fn recheck_and_setup_layer_file() {
...
-                "cargo:rustc-env=MIRRORD_LAYER_FILE={}",
-                std::env::var("CARGO_CDYLIB_FILE_MIRRORD_LAYER").unwrap()
-            );
-        }
+        let layer_path = if cfg!(windows) {
+            std::env::var("CARGO_CDYLIB_FILE_MIRRORD_LAYER_WIN").unwrap()
+        } else {
+            std::env::var("CARGO_CDYLIB_FILE_MIRRORD_LAYER").unwrap()
+        };
+        unsafe { std::env::set_var("MIRRORD_LAYER_FILE", layer_path) };
```
0x00A5: Why do we add setting `MIRRORD_LAYER_FILE`? Setting it for rustc should be sufficient right?

mirrord/cli/build.rs
```diff
@@ -9,31 +7,17 @@ fn recheck_and_setup_layer_file() {
...
     };
+    println!(
+        "cargo:rustc-env=MIRRORD_LAYER_FILE={}",
+        std::env::var("MIRRORD_LAYER_FILE").unwrap()
+    );
```
0x00A5: Should this be moved to the if block above, when `MIRRORD_LAYER_FILE` is not set.

mirrord/layer/Cargo.toml
0x00A5: > mirrord ci

That was a jumpscare. You mean `mirrord cli` 😄

mirrord/layer-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+            .as_ref()
+            .map(|args| args.iter().map(|s| s.as_str()).collect());
+        let env_pairs: Vec<(String, String)> = env.into_iter().collect();
+        let env_refs: Vec<(&str, &str)> = env_pairs
+            .iter()
+            .map(|(k, v)| (k.as_str(), v.as_str()))
+            .collect();
+
+        let cmdline: Vec<String> = [executable].into_iter().chain(self.get_args()).collect();
+        run_exec_targetless(cmdline, None, cli_args, Some(env_refs)).await
```
0x00A5: We shouldn't use `run_exec_targetless` regardless of what's in config file I think.

mirrord/layer-tests/tests/common/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    pub async fn start_process_with_layer(
```
0x00A5: nit: now we actually run mirrord cmd, maybe rename this and the next function without the `_with_layer` part.

mirrord/layer/Cargo.toml
itsamegraf: yeah, sorry about that 🤭 I should verify my comments more frequently

mirrord/cli/build.rs
itsamegraf: IIUC, rustc-env is only provided to the rustc and used by `env!` compile time usages, which as it seems is only layer extraction. there are a couple of runtime usages that needs the process execution to have this env var set, including all processes (and subprocesses) in windows injection process.
so I think we should keep it - this function makes sure we have both runtime and compiletime env `MIRRORD_LAYER_FILE` set properly.

mirrord/cli/build.rs
itsamegraf: per my previous comment, you're right

mirrord/cli/build.rs
itsamegraf: asked chat, you are right, setting it in build.rs does nothing as it is running in a separate process and does not propagate to the execution later, which was my wrong assumption when writing this. I will remove the set_env, and move `rustc-env` to within the if block

mirrord/layer-tests/tests/common/mod.rs
itsamegraf: I agree with you, but annoyed by the fact this will require renaming all `start_process_with_layer...` usages in the next PRs to `start_process` 🤣

mirrord/layer-tests/tests/common/mod.rs
itsamegraf: Not sure what I should do instead, here's my suggested alternatives:
* call `run_exec(target=None)` (which is exactly what `run_exec_targetless` does)
* refactor run_command to split `run_exec` to `run_exec` and `run_exec_with_agent` which"ll require the target and namespace arguments.

please advise or suggest a different approach 🙏

0x00A5: I see. The name of the function `run_exec_targetless` can be misleading here, but it is doing the right thing, which is do not override target and namespace by setting cli args, and use the config file instead.
I think just call `run_exec` here with target and namespace both set to none?

mirrord/layer-tests/tests/common/mod.rs
```diff
@@ -154,15 +154,10 @@ pub fn get_env(
         .map(|(key, value)| (key.to_string(), value.to_string()))
         .collect::<Vec<_>>();
 
-    [
-        ("RUST_LOG".to_string(), "warn,mirrord=debug".to_string()),
```
itsamegraf, line 158: the removed env-vars are already set by `run_exec`

PR #3993, opened 2026-02-27
Automate configuration.md update on release

gememma:
### Quality Checklist:

Issue: https://linear.app/metalbear/issue/COR-1109/automate-configurationmd-update

Summary: Adds a release step to run medschool and push latest docs to the `docs` branch, which can then be synced with gitbook to replace https://github.com/metalbear-co/docs-configuration. Also, removed CI check that forces PR owner to update `configuration.md` manually (and removed `configuration.md`)

Related change: The `docs` branch has been created and populated with the most recent website config docs, and will need to be synced with gitbook manually.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1067,7 +1067,7 @@ jobs:
           Write-Host "Successfully updated WinGet manifest"
 
   update_latest:
-    needs: release_homebrew
+    needs: [release_homebrew, check_for_new_docs]
     runs-on: ubuntu-24.04
     if: github.event_name != 'workflow_dispatch'
     permissions:
--- a/scripts/residency.sh
+++ b/scripts/residency.sh
@@ -33,11 +33,15 @@ if ! cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE
 then
     # flaked once, rerun
     echo "\nresidency: rerunning medschool once (flake)"
-    cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE_PATH"
+    cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE_PATH" || exit
 fi
 
+# from now on, if any command fails abort the script
+set -e
+
 # now all operations occur from the temp dir
-cd "$TEMP_DIR_NAME" || exit
+cd "$TEMP_DIR_NAME"
+# if csplit fails unexpectedly: read comment above (starting "Thing I learned")
 csplit -s -z -f docs_ -n 1 ./temp.md '/^# /' '{*}'
 rm ./temp.md
 echo "\nresidency: docs split successfully"


.github/workflows/ci.yaml
eyalb181, line 48:
```suggestion
  contents: write
```

I think this might sort it out?

.github/workflows/release.yaml
```diff
@@ -705,6 +705,58 @@ jobs:
...
+        shell: bash
+      - name: Commit to docs branch (syncs with GitBook)
+        uses: metalbear-co/github-pages-deploy-action@d92aa235d04922e8f08b40ce78cc5442fcfbfa2f
+        with:
+          folder: ${{ env.TEMP_DIR_NAME }}
+          target-folder: configuration
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: docs
+          clean: false # prevent SUMMARY.md in target folder from being removed
+          dry-run: $ {{ github.event_name == 'workflow_dispatch' }} # dry run on manual dispatch
```
itsamegraf:
```suggestion
          dry-run: ${{ github.event_name == 'workflow_dispatch' }} # dry run on manual dispatch
```

scripts/residency.sh
```diff
@@ -0,0 +1,90 @@
...
+# you want (new pages, etc.)
+
+TEMP_FILE_PATH=./$TEMP_DIR_NAME/temp.md
+
+# prep and split markdown from medschool
+if ! cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE_PATH"
+then
+    # flaked once, rerun
+    echo "\nresidency: rerunning medschool once (flake)"
+    cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE_PATH"
```
itsamegraf: Do we intend this to quit on second failed run? 
I'm not that experienced with bash so I checked online and it was suggested to use `set -e` to achieve it. lmkwyt
```suggestion
    set -e
    cargo run -p medschool -- --input ./mirrord/config/src --output "$TEMP_FILE_PATH"
```

.github/workflows/release.yaml
```diff
@@ -705,6 +705,58 @@ jobs:
...
+        with:
+          fetch-tags: true
+          fetch-depth: 0
+      - name: get config changes since last release
+        id: changed-config-schema
+        uses: metalbear-co/changed-files@9200e69727eb73eb060652b19946b8a2fdfb654b
+        with:
+          files: |
+            mirrord-schema.json
+          base_sha: latest
```
itsamegraf, line 723: (codex suggested)
super nit: we rely on `latest` tag, therefore we should make sure it is not being updated before this job runs. latest is being updated by `update_latest`... which is at the very end of the queue.
If for some reason this (dependancy-less) job runs after it (when 🐖 🐖  fly), then config changes will not be picked up. we can make `update_latest` need `check_for_new_docs`. but I guess it's gonna be ugly in the diagram, and the chances are sooooooo low.
feel free to disregard this note 🤣

scripts/residency.sh
```diff
@@ -0,0 +1,90 @@
...
+fi
+
+# now all operations occur from the temp dir
+cd "$TEMP_DIR_NAME" || exit
+csplit -s -z -f docs_ -n 1 ./temp.md '/^# /' '{*}'
```
itsamegraf, line 45: nit: worth adding a comment here directing to read the csplit journey above if it doesnt behave as expected.
feel free to disregard

scripts/residency.sh
gememma: yeah good idea, i dont think it would be harmful to keep running the script (?) but no need to waste time on it

itsamegraf (approved):

PR #3995, opened 2026-03-01
win-38 ⊹ CLI/layer: attach support 2: the electric boogaloo

cristeigabriela:
## Linear issue

https://linear.app/metalbear/review/win-38-implement-mirrord-side-features-for-ide-support-168c5d060eee

## Overview

This is a PR that implements functionality meant to later be used in implementations of IDE extension support.

To understand how it is used, you may refer to the first, "authoritative" implementation of Windows, attach-based IDE extension support in the context of the VSCode extension: https://github.com/metalbear-co/mirrord-vscode/pull/210

## The core issue

Unix supports the `LD_PRELOAD` (or the MacOS equivalent, for dylibs, and whatnot) environment variable as a way to tell processes to load a dynamic library.

Windows, unfortunately, does not have any such similar mechanism.
> UNIMPORTANT: The closest similar mechanism would be IFEO, but this allows the substitution of the `.exe` being ran, and that has admin rights as a prerequisite

That means that we cannot facillitate the loading of the mirrord layer through environment variables as the Unix versions of the code do, in the case of IDE extensions.

## The suggested solution

The suggested equivalent is the `attach` command, **being used with `stopOnEntry` to substitute the now-optional signal event**, and then running `attach` upon a **Process Event** ([as defined by the DAP](https://microsoft.github.io/debug-adapter-protocol//specification.html)) with the provided `systemProcessId`.

This will always point to the user-code process, so this does not require explicit handling.

## Notes

`mirrord.exe attach` is:
1. Windows-only (the technique *could* be ported over to Unix, but manual mapping on there is it's own ordeal);
2. Marked as hidden to the CLI, as it should never be manually invoked;

## Important

- The caller context for `mirrord.exe attach` must guarantee that the process at `PID` is set up with an appropriate mirrord environment already, otherwise `layer-win` will fail to initialize.
- As mentioned before, in the scenarios of `mirrord.exe attach`, you are responsible to make sure that `layer-win` loads before relevant user-code runs.

---

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/cli/src/attach.rs
+++ b/mirrord/cli/src/attach.rs
@@ -10,8 +10,28 @@ const ATTACH_SIGNAL_TIMEOUT_MS: u32 = 30_000;
 
 /// Attach the mirrord layer to an already-running process by injecting the layer DLL.
 ///
-/// The target process is expected to already have all mirrord environment variables
-/// configured (by the IDE extension), so no k8s setup or proxy spawning is needed.
+/// This is only the DLL-injection half of the attach flow. By the time this function runs,
+/// the **IDE extension** (mirrord-vscode) has already done all the heavy lifting:
+///
+/// 1. Started the intproxy.
+/// 2. Retrieved the necessary environment variables from the agent/intproxy (proxy address, layer
+///    ID, resolved config, etc.).
+/// 3. Injected those environment variables into the target process (through editing the debug
+///    launch configuration).
+/// 4. Invoked `mirrord attach --pid <pid>` (this function).
+///
+/// Because of this, `attach_command` does **not** spawn an intproxy, resolve a k8s
+/// target, or set up any environment variables itself — all of that state already
+/// exists in the target process's environment before we get here.
+///
+/// Presently, the debugging VSCode instance is waiting for attach to finish.
+///
+/// When we are done, the VSCode instance will do post-attach chores, such as
+/// resuming the user-code process' primary execution thread, making this
+/// whole process invisible to the user.
+///
+/// The extension-side implementation was introduced in the following pull request:
+/// <https://github.com/metalbear-co/mirrord-vscode/pull/210>.
 pub(crate) fn attach_command<P>(args: AttachArgs, progress: &P) -> CliResult<()>
 where
     P: Progress,
@@ -26,9 +46,9 @@ where
         .map_err(|e| CliError::AttachProcessOpenFailed(args.pid, e))?;
     sub_progress.info(&format!("obtained handle to process {}", args.pid));
 
-    // Create the attach event before injection. The layer reads this by name (derived from its
-    // own PID) and signals it when initialization is complete.
-    let init_event = LayerInitEvent::for_attach_parent(args.pid)
+    // Create the event before injection. The layer opens it by deriving the same
+    // name from its own PID and signals it when initialization is complete.
+    let init_event = LayerInitEvent::for_parent(args.pid)
         .map_err(|e| CliError::AttachInjectionFailed(args.pid, e.to_string()))?;
 
     let syringe = Syringe::for_process(process);
--- a/mirrord/layer-lib/src/process/windows/sync.rs
+++ b/mirrord/layer-lib/src/process/windows/sync.rs
@@ -1,11 +1,9 @@
 /// Process synchronization utilities for Windows layer initialization.
 ///
-/// Provides Windows-specific process synchronization functionality for coordinating
-/// layer initialization between parent and child processes using named events.
-use std::{
-    ffi::CString,
-    time::{SystemTime, UNIX_EPOCH},
-};
+/// Provides Windows-specific process synchronization between parent and child processes
+/// using named events. Both sides derive the event name from the child's PID
+/// (`mirrord_layer_init_{child_pid}`), so no environment variable handoff is needed.
+use std::ffi::CString;
 
 use winapi::{
     shared::{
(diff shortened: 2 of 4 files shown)


mirrord/cli/src/attach.rs
```diff
@@ -0,0 +1,33 @@
...
+/// configured (by the IDE extension), so no k8s setup or proxy spawning is needed.
+pub(crate) fn attach_command<P>(args: AttachArgs, progress: &P) -> CliResult<()>
+where
+    P: Progress,
+{
+    let mut sub_progress = progress.subtask("attaching to process");
+
+    let lib_path = extract_library(None, progress, true)?;
+
+    unsafe { std::env::set_var("MIRRORD_LAYER_FILE", &lib_path) };
```
itsamegraf: incorporate MIRRORD_LAYER_FILE existing value similar to how its in `start_internal`, otherwise it's gonna make it difficult to hot-reload layer-win dll during development when using the plugin

mirrord/layer-win/src/lib.rs
```diff
@@ -79,8 +79,19 @@ fn initialize_windows_proxy_connection() -> LayerResult<()> {
 }
 
 fn layer_start() -> LayerResult<()> {
-    // Create layer initialization event first
-    let init_event = LayerInitEvent::for_child()?;
+    // Try to open the parent's init event. In the normal `mirrord exec` flow, the parent
+    // creates this event and waits on it. In the `mirrord attach` flow (IDE extension),
+    // the env var won't be set because the process was spawned independently — that's fine,
+    // we just skip the synchronization.
+    let init_event = match LayerInitEvent::for_child() {
```
itsamegraf: I have worries about this, why dont we need it?

mirrord/cli/src/main.rs
```diff
@@ -995,9 +998,9 @@ fn main() -> miette::Result<()> {
             Commands::Operator(args) => {
                 operator_command(*args).await?;
             }
-            Commands::ExtensionExec(args) => windows_unsupported!(args, "ext", {
+            Commands::ExtensionExec(args) => {
```
itsamegraf, line 1060: is this used by windows extension?

cristeigabriela: Yes

mirrord/cli/src/attach.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use dll_syringe::{Syringe, process::OwnedProcess};
+use mirrord_progress::Progress;
+
+use crate::{CliResult, config::AttachArgs, error::CliError, extract::extract_library};
+
+/// Attach the mirrord layer to an already-running process by injecting the layer DLL.
+///
+/// The target process is expected to already have all mirrord environment variables
+/// configured (by the IDE extension), so no k8s setup or proxy spawning is needed.
```
itsamegraf: Can you please direct me to where the k8s setup being done?

mirrord/layer-win/src/lib.rs
cristeigabriela: Because process ownership is not ours in the scenario of IDE attach, therefore there's no parent to signal after DLL injection in the "process creation loop", because there's no "process creation loop" in the first place, only an existing process.

cristeigabriela: It's addressed by `stopOnEntry` instead in the VSCode extension. See the VSCode PR and diagram.

mirrord/cli/src/main.rs
itsamegraf: can you please redirect me where? I don't see it in your PR

cristeigabriela: It's not implemented by me, but it was used by the extension to begin with. It's a dependency that is necessary for IDE extension.

https://github.com/search?q=repo%3Ametalbear-co%2Fmirrord-vscode+makeMirrordArgs&type=code

mirrord/cli/src/attach.rs
cristeigabriela: This can be seen by looking at how the IDE extension works.

https://github.com/metalbear-co/mirrord-vscode/blob/d201aa7376b3f055f256413a43d7db5d0fedba75/src/debugger.ts#L205

The environment variables come from the invocation to `mirrord.exe ext`.

mirrord/layer-win/src/lib.rs
itsamegraf: How do we currently determine `layer-win` injection success in the extension context? this synchronization is the only way to determine hooks were registered successfully as their registration is done in a background thread to prevent blocking DllMain. any other way risks having the debugee process not use our hooked methods as they might not be registered.

cristeigabriela: Failure should propagate to the debugger as it is already attached to the process, there's no  other current way to check for it and it should be unlikely in the scenario of a correct environment (which is the responsibility of `mirrord.exe ext` to verify before we get there).

I haven't tested edge cases, but I don't think there's many ways to gracefully check them. We could still have an event in attach to await to restart the threads execution, or time out on it, though. But I feel like that might be overengineering.

mirrord/cli/src/attach.rs
itsamegraf: nit: please document here as well that previous `mirrord ext` execution took care of the k8s spawning and injected the context data in env var to this attach (IIUC)

cristeigabriela: If you personally believe that level of detail is warranted, I will introduce it, but I thought that the comment already addressed this summarily enough, and further detail can be easily extracted from IDE extension

mirrord/layer-win/src/lib.rs
cristeigabriela: I decided to implement the synchronization you suggested for attach.

mirrord/cli/src/attach.rs
itsamegraf: it wasn't easy enough for me to extract those details. If it was trivially to answer these details I wouldn't have asked.

mirrord/layer-win/src/lib.rs
itsamegraf: your synchronization implementation that is based on child pid and does not rely on the fact we spawn the child makes alot of sense and is significant better than the existing one, great work.
The code duplication between the 2 approaches is redundant, replace the existing "generate_event_name + env" approach with the new "deterministic child pid name" approach.

cristeigabriela: fixed

mirrord/layer-lib/src/process/windows/sync.rs
```diff
@@ -22,86 +20,44 @@ use winapi::{
...
     Parent,
-    /// This process opened an existing event (child role)
+    /// This process opened an existing event (child role).
     Child,
 }
 
 /// Managed layer initialization event with RAII resource management.
 ///
-/// Provides a clean, type-safe API for Windows event-based process synchronization
+/// Provides a type-safe API for Windows event-based process synchronization
```
Razz4780, line 34: Love this change 😆

cristeigabriela: honesty is valuable!

PR #4146, opened 2026-04-09
INT-371 Support DB credentials defined in mirrord.json

vladrbg:
We are adding a new `value` field to connection parameters. This lets us provide a plain credential value directly in the mirrord config. We need this when the target pod doesn't have a connection parameter for example `password` in its pod spec this can happen when the app gets it from another place in runtime. Because operator can't read those runtime values, we can provide them explicitly with `value` so the operator can use it to connect to the source database when creating the branch.

**How it works**
When `value` is provided, the CLI sends it to the operator, which stores it in a Kubernetes Secret named `mirrord-branch-creds-{branch_id}`. The branch pod's init container then reads the value from that Secret via `secretKeyRef` to connect to the source database and clone its data.
For example, with branch ID `my-branch`, the operator creates a Secret called `mirrord-branch-creds-my-branch` with a key `PASSWORD` containing the literal value. The init container gets it as an env var:
```yaml
env:
  - name: SECRET_MIRRORD_BRANCH_CREDS_MY_BRANCH_PASSWORD
    valueFrom:
      secretKeyRef:
        name: mirrord-branch-creds-my-branch
        key: PASSWORD
```
The Secret is owned by the BranchDatabase CRD via ownerReferences, so it gets deleted automatically when the branch is deleted.

Here is what we support right now. Because we added the new `value` field we no longer need to have the `variable` field required so we are making it optional.

This is what we have previously for ENV:
```yaml
"params": {
    "password": "DB_PASSWORD"
}
```
Same as above:
```yaml
"params": {
    "password": {
        "variable": "DB_PASSWORD"
    }
}
```
The new field we added:
```yaml
"params": {
    "password": {
        "value": "some-password"
    }
}
```
To be fair, this also will work, but makes no sense anyone to use it, if you have ENV you should just use env:
```yaml
"params": {
    "password": {
        "variable": "DB_PASSWORD",
        "value": "some-password"
    }
}
```
The pure value is only used by the operator for branch creation. It is not injected into the local application's environment.


changelog.d/+db-branching-pure-value.added.md
```diff
@@ -0,0 +1 @@
+Support for pure value credentials in db branching.
```
t4lz: how about plain/raw/in-place
```suggestion
Support for plain value credentials in db branching.
```

mirrord/cli/src/internal_proxy/db_portforwards.rs
```diff
@@ -193,9 +193,31 @@ fn extract_portforward_configs(config: &DatabaseBranchesConfig, key: &str) -> Ha
...
+                        ParamSource::Env {
+                            variable: Some(host),
+                            ..
+                        },
+                        ParamSource::Env {
+                            variable: Some(port),
+                            ..
+                        },
+                    ) => (host.clone(), port.clone()),
+                    _ => continue,
```
t4lz: Looks like before a `_` was deliberately avoided, in order to force people to add cases here when they add variants. Is there a good reason not to respect that and to replace the `Secret` cases with a `_`?

mirrord/cli/src/internal_proxy/db_portforwards.rs
```diff
@@ -193,9 +193,31 @@ fn extract_portforward_configs(config: &DatabaseBranchesConfig, key: &str) -> Ha
...
                 };
```
t4lz: This is a very long way to write something pretty simple. Can you shorten it up a bit?
For example you could use `ParamSource::as_variable`, and then

```suggestion
let (Some(host), Some(port)) = (host.as_variable(), port.as_variable()) else {
    continue
}
```

Or whatever that's pretty and compiles.

t4lz: alternatively


```suggestion
let (host, port) = match (&host, &port) {
    (
        ParamSource::Variable(host)
        | ParamSource::Env {
            variable: Some(host),
            ..
        },
        ParamSource::Variable(port)
        | ParamSource::Env {
            variable: Some(port),
            ..
        },
    ) => (host.clone(), port.clone()),
    _ => continue,
};
```

mirrord/config/src/feature/database_branches.rs
```diff
@@ -377,12 +383,19 @@ pub enum ParamSource {
         name: String,
         key: String,
     },
+    Env {
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        variable: Option<String>,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
```
t4lz: I think these:
```suggestion
        env_var_name: Option<String>,
        #[serde(default, skip_serializing_if = "Option::is_none")]
        fallback_value: Option<String>,
```
are better names

mirrord/config/src/feature/database_branches.rs
```diff
@@ -419,6 +432,11 @@ pub enum TargetEnvironmentVariableSource {
     Env {
         container: Option<String>,
         variable: String,
+        /// Fallback value to use when the env var is not found on the target pod.
+        /// Useful when the app gets credentials from Vault or other non-K8s sources.
+        /// The CLI encrypts this value before it reaches the CRD.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
```
t4lz, line 441: I think
```suggestion
        fallback_value: Option<String>,
```
is a better field name

mirrord/config/src/feature/database_branches.rs
```diff
@@ -740,21 +761,35 @@ mod tests {
...
+                assert_eq!(
+                    config.params.host,
+                    Some(ParamSource::Env {
+                        variable: None,
+                        value: None
+                    })
+                );
+            }
+            other => panic!("expected Params, got {other:?}"),
+        }
```
t4lz: Wait, this isn't ok 😅
The new config schema allows for any json object and doesn't error on it, this test caught this issue, but instead of fixing the issue, the test was changed to be ok with that, and now expects an Env object with both fields empty - which doesn't make sense.
Instead, please make sure in the config (ideally on a schema level):
1. unknown fields are not allowed
2. at least one of the two existing fields has to be present. Though I actually think that `variable` should probably just be always required, see my other comments (also I think we should call it `env_var_name`).

Please make sure the UTs verify both things.

mirrord/operator/src/crd/db_branching/core.rs
```diff
@@ -42,6 +42,12 @@ pub enum ConnectionSourceKind {
     Env {
         container: Option<String>,
         variable: String,
+        /// Literal fallback value. When the CLI receives a `value` in the user config,
+        /// it creates a K8s Secret via the operator and replaces this with a
+        /// `Secret { name, key }` variant before writing the CRD. This field is kept
+        /// for backward compatibility but should not carry sensitive data in the CRD.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        value: Option<String>,
```
t4lz:
```suggestion
        fallback_value: Option<String>,
```

mirrord/operator/src/crd/db_branching/core.rs
```diff
@@ -105,32 +115,45 @@ impl From<&TargetEnvironmentVariableSource> for ConnectionSourceKind {
...
-            host: wrap(&config.params.host),
-            port: wrap(&config.params.port),
-            user: wrap(&config.params.user),
-            password: wrap(&config.params.password),
-            database: wrap(&config.params.database),
+            host: wrap(&config.params.host, "host"),
+            port: wrap(&config.params.port, "port"),
+            user: wrap(&config.params.user, "user"),
+            password: wrap(&config.params.password, "password"),
+            database: wrap(&config.params.database, "database"),
```
t4lz: What's going on here?
Why do we have those hardcoded env var names? Who expects them, where, and why?

mirrord/operator/src/client/database_branches.rs
```diff
@@ -626,6 +626,95 @@ impl AsRef<str> for BranchDatabaseId {
...
+    ) {
+        if let ConnectionSourceKind::Env { variable, .. } = kind
+            && keys.contains_key(variable.as_str())
+        {
+            *kind = ConnectionSourceKind::Secret {
+                name: secret_name.to_string(),
+                key: variable.clone(),
+            };
+        }
+    }
```
t4lz: The docs at multiple places say the value is used as a fallback, if the env var is not present. Given that we switch the connection source kind in advance - is it really used as a fallback only if the env isn't present?

mirrord/config/src/feature/database_branches.rs
```diff
@@ -419,6 +432,11 @@ pub enum TargetEnvironmentVariableSource {
     Env {
         container: Option<String>,
         variable: String,
+        /// Fallback value to use when the env var is not found on the target pod.
+        /// Useful when the app gets credentials from Vault or other non-K8s sources.
+        /// The CLI encrypts this value before it reaches the CRD.
```
t4lz: Does it?

t4lz (changes requested):

mirrord/operator/src/client/database_branches.rs
t4lz: Btw, if it's not actually used as a fallback, then ignore my suggestion to put "fallback" in the field name.

mirrord/config/src/feature/database_branches.rs
t4lz: Though "fallback" only if it actually is a fallback and not an override. See https://github.com/metalbear-co/mirrord/pull/4146#discussion_r3065446548

mirrord/operator/src/crd/db_branching/core.rs
vladrbg: If we do not provide `variable` and only have `value` we will need to create unique keys for the `secret` so we use the `type` of field as part of it. So it will  look like this here, if the user had:

`"password": "some password in config"`

then it will produce (the `PASSWORD` at the end so it is unique name)
```yaml
env:
  - name: SECRET_MIRRORD_BRANCH_CREDS_MY_BRANCH_PASSWORD
    valueFrom:
      secretKeyRef:
        name: mirrord-branch-creds-my-branch
        key: PASSWORD
```

mirrord/config/src/feature/database_branches.rs
vladrbg: The problem is here, with `#[serde(untagged)]`, serde tries each variant and picks the first that succeeds. Since Env has both fields optional  it matches any JSON object:

```rust
#[serde(untagged)]
pub enum ParamSource {
    Variable(String),
    Secret {
        #[serde(rename = "secret")]
        name: String,
        key: String,
    },
    Env {
        #[serde(default, alias = "variable", skip_serializing_if = "Option::is_none")]
        env_var_name: Option<String>,
        value: Option<String>,
    },
}
```

mirrord/config/src/feature/database_branches.rs
vladrbg: oh thats from the first iteration where it was working with encription, leftover.

mirrord/operator/src/client/database_branches.rs
vladrbg: Yeah we dont really fallback, if you have provided value, we create the secret.

mirrord/config/src/feature/database_branches.rs
vladrbg: No longer the case with `env_var_name` being required field now

mirrord/operator/src/crd/db_branching/core.rs
vladrbg: No longer the case with `env_var_name` being required field now

mirrord/operator/src/crd/db_branching/core.rs
t4lz: Since the CRD should actually never be created with that field (if we have a value, we create a Secret source kind instead) it would be way, way better, if this field doesn't exist in the spec, and the types and schemas only allow for valid objects.

mirrord/config/src/feature/database_branches.rs
```diff
@@ -363,12 +363,18 @@ pub struct ConnectionParamsConfig {
...
+/// As a string: `"DB_HOST"` - resolved using the parent `type` field (env or env_from).
 ///
-/// As an object: `{ "secret": "my-secret", "key": "password" }` — read directly from a
+/// As an object with fallback: `{ "variable": "DB_HOST", "value": "myhost.com" }` - tries the
+/// env var first, falls back to the literal value when the var is not on the target pod.
+///
+/// As a value-only object: `{ "value": "myhost.com" }` - provides the value directly without
+/// referencing any env var on the target pod.
+///
+/// As a Secret ref: `{ "secret": "my-secret", "key": "password" }` - read directly from a
```
t4lz, line 377: These docs still call it a fallback, even though it's not a fallback (we don't use the value from the remote env even if it's there).

t4lz (dismissed):
Just fix those docs at that one place plase.

t4lz (approved):
❤️

PR #4151, opened 2026-04-09
COR-232 Feature gate unnecessary dependencies in mirrord-protocol

DmitryDodzin:
### Quality Checklist:

  documentation)


meowjesty (approved):
Would be nice if we could move these optional impls to separate crates, so we can feature gate the whole crate, instead of doing it with a bunch of `cfg`s in the code.

PR #4158, opened 2026-04-12
Remove need of bindep. Introduce cargo xtask for tests

aviramha:
Remove need of bindep. Introduce cargo xtask test-e2e and cargo xtask test-integration


MintSoup (dismissed):
LGTM 👀

xtask/README.md
```diff
@@ -38,6 +39,19 @@ cargo xtask build-cli --release --platform windows
...
+
+### `init`
+
+Installs xtask-managed external tools. `uv` and Python 3 are prerequisites for this command.
+Today it uses `uv` to sync the Python tool environment and exposes `cargo-zigbuild` to Cargo via a
+wrapper in `$CARGO_HOME/bin`.
+
+```bash
+cargo xtask init
 ```
```
iniw: Do we really wanna add this? I don't like our xtask script becoming a pseudo-package-manager. 

Can we delegate that to a proper tool please? I like nix, of course, and we already have a devshell in the project, so I'm inclined to say we could just use that, but if that's out of the question I'd much rather just document the required packages and their respective versions and let the users figure out how to get them.

aviramha: I hate nix but I am removing the xtask init regardless :D

hank-metalbear (approved):

PR #4161, opened 2026-04-13
mirrord up 

MintSoup:
### Quality Checklist:

  documentation)

MVP for mirrord up. See 
https://github.com/metalbear-co/docs/pull/164
https://github.com/metalbear-co/rfcs/pull/27
https://linear.app/metalbear/issue/COR-1326/mirrord-up-accept-mirrord-upyaml-file-and-start-mirrord-session-for
https://www.notion.so/metalbear-co/DRAFT-mirrord-up-32415e39139480abbcb2ff9dd303b60d?source=copy_link#32b15e3913948074a05bd5a0f600866a

for details


mirrord/cli/src/config.rs
```diff
@@ -203,6 +203,10 @@ pub(super) enum Commands {
     #[cfg_attr(target_os = "windows", command(hide = true))]
     Preview(Box<PreviewArgs>),
 
+    /// Run mirrord sessions for all services defined in `mirrord-up.yaml`.
+    #[cfg_attr(target_os = "windows", command(hide = true))]
+    Up(Box<UpArgs>),
```
meowjesty, line 208: alias `mirrord up` to 

Image

mirrord/cli/src/config.rs
```diff
@@ -1399,6 +1403,21 @@ impl PreviewStopArgs {
     }
 }
 
+/// Arguments for the `mirrord up` command.
+#[derive(Args, Debug)]
+pub(super) struct UpArgs {
+    /// Path to the mirrord-up config file.
+    ///
+    /// When using this argument without a value, defaults to `mirrord-up.yaml`
```
meowjesty, line 1411: If only the normies knew how much better `toml` is ...

mirrord/cli/src/error.rs
```diff
@@ -606,6 +607,11 @@ pub(crate) enum CliError {
     #[diagnostic(help("Specify the key using --key <key> or set it in your mirrord config file."))]
     PreviewKeyRequired,
 
+    /// Errors produced by the `mirrord up` command.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    Up(#[from] UpError),
```
meowjesty: Would be hilarious if the error type was named `Down`.

Cargo.toml
```diff
@@ -243,6 +241,11 @@ windows = { version = "0.61", features = [
     "Win32_Foundation",
     "Win32_Networking_WinSock",
 ] }
+
+# Used by `cli`
+whoami = "2.1.1"
```
meowjesty, line 246: The quest for self-improvement, ask not "who am I", ask instead "who I want to be".

mirrord/up/src/lib.rs
```diff
@@ -0,0 +1,119 @@
+use std::{
```
meowjesty, line 4: Add those things like "deny unused dependencies" and "deny missing docs".

mirrord/up/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+serde = { workspace = true, features = ["derive"] }
+serde_yaml.workspace = true
+miette = { workspace = true, features = ["fancy"] }
+thiserror.workspace = true
+strum_macros.workspace = true
+tokio = { workspace = true, features = ["process"] }
+
+
+
```
meowjesty:
```suggestion

```

I know that the number 3 has many symbolic meanings, but we should be culture agnostic here and use only a single empty line.

mirrord/cli/src/main.rs
```diff
@@ -330,6 +330,7 @@ mod port_forward;
 mod preview;
```
meowjesty, line 338: Update the mega docs with the new `mirrord up` command (that we have as the module docs for this file).

mirrord/cli/src/up.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use crate::{config::UpArgs, error::CliResult};
+
+/// The `mirrord up` command handler.
+pub(crate) async fn up_command(args: UpArgs) -> CliResult<()> {
+    let up_config = load_up_config(&args.config_file).map_err(crate::error::CliError::Up)?;
+    let key = args
+        .key
+        .map(EnvKey::Provided)
+        .unwrap_or_else(|| EnvKey::Generated(whoami::username().unwrap()));
```
meowjesty: Nit: I don't think this ever fails, but I think using `expect` with a message indicating that this is a bug and should be reported to us would be better than the default rust `unwrap` crash.

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+};
+use serde::{Deserialize, Serialize};
+use strum_macros::IntoStaticStr;
+
+/// Incoming traffic mode for a service.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum ServiceMode {
+    #[default]
+    Split,
```
meowjesty, line 24: What is split? Docs pls

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Default, IntoStaticStr)]
+#[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "lowercase")]
+pub enum RunType {
```
meowjesty, line 32: Docs pls.

`mirrord up` supports using both `mirrord exec` and `mirrord container`?

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+    #[default]
+    Exec,
+    Container,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub struct RunConfig {
```
meowjesty, line 42: Docs pls.

What's the `RunConfig`?

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+        Self {
+            accept_invalid_certificates: None,
+            operator: None,
+            telemetry: true,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize, Default)]
+pub struct TargetConfig {
```
meowjesty, line 57: Why this config type instead of using the config type we already have?

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+        Self {
+            path: value.path,
+            namespace: value.namespace,
+        }
+    }
+}
+
+/// Per-service configuration.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ServiceConfig {
```
meowjesty, line 74: Why are all of these types and fields `pub`? Aren't we only using them in this crate (with a bunch of these only being used in this file it looks like)?

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub ignore_ports: HashSet<u16>,
+
+    pub run: RunConfig,
+}
+
+/// Resolved top-level `mirrord-up.yaml` configuration.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct UpConfig {
+    #[serde(default)]
+    pub defaults: DefaultsConfig,
```
meowjesty, line 97: Is this `defaults` or more like `shared` configs?

meowjesty: Or `common` configs?

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        cfg.feature.network.incoming.ignore_ports = self.ignore_ports;
+        cfg.key = key;
+
+        (cfg, self.run)
+    }
+}
+
+pub struct SubprocessCfg {
```
meowjesty, line 143: Docs pls.

mirrord/up/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Debug, Error, Diagnostic)]
+pub enum UpError {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+
+    #[error("failed to parse mirrord-up config: {0}")]
+    #[diagnostic(help("Check the YAML syntax and field names in your mirrord-up.yaml."))]
+    Parse(#[from] serde_yaml::Error),
+
+    #[error("mirrord-up config validation failed: {0}")]
```
meowjesty, line 34: Don't we have a nice diagnostic helper for this, or does it come inside the `ConfigError` what's wrong in the user's config?

mirrord/up/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    Panic(JoinError),
+}
+
+/// Load and parse a `mirrord-up.yaml` configuration file.
+pub fn load_up_config(path: &PathBuf) -> Result<UpConfig, UpError> {
+    let content = std::fs::read_to_string(path)?;
+    Ok(serde_yaml::from_str(&content)?)
+}
+
+pub async fn run(up_config: UpConfig, key: EnvKey) -> Result<(), UpError> {
```
meowjesty, line 56: Docs pls.

```
/// Loads the config bla bla bla and `spawn`s a bunch of mirrords ...
```

mirrord/up/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+            let encoded_cfg = config.encode()?;
+
+            let mut cmd = Command::new(std::env::current_exe()?);
+            cmd.env(RESOLVED_CONFIG_ENV, encoded_cfg)
+                .env(MIRRORD_PROGRESS_ENV, "simple")
+                .arg(Into::<&'static str>::into(run.r#type))
+                .arg("--")
+                .args(run.command)
+                .stdin(Stdio::piped())
+                .stdout(Stdio::piped())
```
meowjesty, line 75: Don't we want to give the user an option to write stdio stuff to a file, like we do for `mirrord ci`? The logs here for multiple apps will all be thrown in mixed together, right?

meowjesty (changes requested):
Will Disney sue us?

mirrord/up/src/lib.rs
MintSoup: Yup it's a mess, unfortunately product and Aviram don't want to commit to anything specific just yet for the MVP. Will definitely have a nicer UI in the future tho

mirrord/up/src/config.rs
MintSoup: It's "shared" but I'm just following the PRD

mirrord/up/src/config.rs
MintSoup: I did that initially, however the type we already have parses differently and supports other formats that we don't want here, so we have to use a new one :(

mirrord/cli/src/config.rs
MintSoup: :(

Cargo.toml
MintSoup: lol

mirrord/up/src/lib.rs
MintSoup: Not quite sure what you mean but `serde_yaml::Error` tells us what's wrong, plus I don't think we can reuse `ConfigError` because that's for `mirrord.json` not this

mirrord/up/src/config.rs
MintSoup: Yup

mirrord/up/src/lib.rs
meowjesty: > serde_yaml::Error tells us what's wrong

That's what I was wondering about, if in the error there's some message that helps the user fix their config.

mirrord/up/src/config.rs
meowjesty: I would change this to something that won't cause confusion, `Default` in Rust has a different meaning.

mirrord/up/src/config.rs
meowjesty: Sadness, add some docs explaining this, so people don't try to refactor this out in the future (maybe it becomes a tech-debt thing in the future, to unify these target confis).

mirrord/up/src/lib.rs
meowjesty: Deny missing docs too pls, ideally everything we add comes with docs everywhere, so I don't have to ask people for docs everywhere.

meowjesty (dismissed):
Not a fan of it being named `DefaultsConfig`, and missing the "deny missing docs", but other than that, looks good to me.

mirrord/up/src/lib.rs
MintSoup: IMO it's excessive, that requires docs for even things like `name` and `status` in enum variant fields. Do we really want this?
```rust
    #[error("Service {name} crashed with exit status {status}")]
    ServiceCrashed { name: String, status: ExitStatus },
```

mirrord/up/src/config.rs
```diff
@@ -0,0 +1,354 @@
...
+            ..Default::default()
+        }
+        .generate_config(&mut ConfigContext::default())
+        .unwrap();
+
+        cfg.target = self.target.into();
+        cfg.feature.env = self.env;
+
+        match self.mode {
+            ServiceMode::Split => {
```
0x00A5, line 122: What does service mode equals to `Split` mean? Each service runs stand alone?

mirrord/up/src/lib.rs
```diff
@@ -25,10 +25,92 @@ pub enum UpError {
...
+                                status,
+                            })
+                        }
+                    }
+                }
+            }
+        });
+    }
+
+    let Some(status) = handles.join_next().await else {
```
0x00A5, line 135: We abort everything after the first service/app exits. Have we considered logging that the service has finished, while letting other sub processes continue to run? I think this will play more nicely with subprocesses that are job instead of service.
For reference, `docker-compose up` let services run by default, unless one of `--abort-on-container-exit` or `--abort-on-container-failure` is specified.

MintSoup: Yup, we'll probably want something like that for future releases. But for the MVP it's been decided to keep things minimal.

mirrord/up/src/config.rs
MintSoup:
```rust
/// Incoming traffic mode for a service.
#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
#[serde(rename_all = "lowercase")]
pub enum ServiceMode {
    /// Incoming traffic is split between the local and the original
    /// service. The user-provided HTTP filter is used (if provided),
    /// otherwise we fall back to an autogenerated one (see
    /// [`ServiceConfig::assemble`])
    #[default]
    Split,
}
```
Only mode we support as of now, but more are to come in the future.

0x00A5 (dismissed):

0x00A5 (approved):

PR #4163, opened 2026-04-14
Make iptables chain names dynamic, so mirrord can target multiple containers of the same pod.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1315/platform-api-router-has-two-containers-the-router-and-coprocessor-both

User was trying to run multiple `mirrord exec` targeting the same pod, but different containers:

```
mirrord exec --target "pod/cat/container/meow"
mirrord exec --target "pod/cat/container/rawr"
```

The 1st session stopped working due to iptables chain names being static, now they're dynamic, so the 2nd session doesn't overwrite the 1st session's rules.

### Quality Checklist:

  documentation)


mirrord/agent/iptables/src/lib.rs
Razz4780: This bit looks problematic. The first agent will overwrite the config file and restore it on exit

mirrord/config/src/agent.rs
```diff
@@ -483,6 +483,24 @@ pub struct AgentConfig {
...
+    ///
+    /// ```sh
+    /// mirrord exec --target "pod/cat/container/meow"
+    /// mirrord exec --target "pod/cat/container/rawr"
+    /// ```
+    ///
+    /// Setting it to `true` changes the iptables chain names to be dynamic, and thus allowing
+    /// multiple agents to work in the same network namespace.
+    #[config(default = false)]
+    pub single_pod_multi_container: bool,
```
Razz4780: I don't think we need this.
Agent should always:
1. Check if rules with legacy names exist -> fail if yes
2. Use rules with new names

mirrord/agent/env/src/envs.rs
```diff
@@ -97,3 +97,14 @@ pub const CLEAN_IPTABLES_ON_START: CheckedEnv<bool> =
...
+/// Identifier for the iptables chain names of the agent.
+///
+/// When this env var is set, we avoid conflicts with other agents in the same network namespace
+/// (i.e. when multiple agents are running in the same pod and targeting different containers).
+///
+/// If not set, we default to the legacy chain names (i.e. `MIRRORD_INPUT`, ...).
+///
+/// We decide to set this env var or not by checking `AgentConfig::single_pod_multi_container`.
+pub const IPTABLES_IDENTIFIER: CheckedEnv<String> =
+    CheckedEnv::new("MIRRORD_AGENT_IPTABLES_IDENTIFIER");
```
Razz4780: There's a simple way (and preferred, I think) - embed target container ID in the names.
So the full agent routine would look more or less like:
1. Assert that there are no rules with legacy names
2. Assert that there are no rules with new names derived from our target ID
3. Setup new rules
4. https://www.youtube.com/watch?v=eO6OaVnbqaY
5. Tear down new rules

Razz4780 (changes requested):

changelog.d/+cor-1315.changed.md
```diff
@@ -0,0 +1 @@
+Added new agent config `single_pod_multi_container: bool` that allows the agent to create dynamically named chain names for iptables rules, which allows running multiple mirrord sessions targeting the same pod, but for different containers.
```
Razz4780: Outdated. Also, should be `fixed`, not `changed`

Razz4780 (dismissed):

Razz4780 (approved):

PR #4164, opened 2026-04-14
INT-274 Queue Splitting - Google PubSub support

vladrbg:
### Quality Checklist:

  documentation)


mirrord/config/src/feature/split_queues.rs
```diff
@@ -155,6 +182,17 @@ impl SplitQueuesConfig {
                 QueueFilter::Rmq { message_filter } => {
                     Self::verify_message_attribute_filter(queue_name, message_filter)?;
                 }
+                QueueFilter::GcpPubSub {
```
t4lz: Nitpick, optional: the GcpPubSub and the Sqs cases could be unified with an `|`, and so could the `Kafka` and `Rmq` cases.
This was like that before this PR, of course, but we could improve it now.

mirrord/config/src/feature/split_queues.rs
```diff
@@ -251,6 +289,24 @@ pub enum QueueFilter {
         message_filter: QueueMessageFilter,
     },
 
+    #[serde(rename = "GcpPubSub")]
```
t4lz: Is this supposed to be
```suggestion
    #[serde(rename = "GCPPubSub")]
```
?
I think that would be more consistent with the `SQS` and `RMQ` cases, and also, otherwise this rename is redundant.

mirrord/config/src/feature/split_queues.rs
```diff
@@ -251,6 +289,24 @@ pub enum QueueFilter {
...
+        /// A filter is a mapping between Pub/Sub message attribute names and regexes they
+        /// should match. The local application will only receive messages whose attributes
+        /// match **all** of the given patterns.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        message_filter: Option<QueueMessageFilter>,
+
+        /// A jq filter.
+        ///
+        /// When this is specified, for each Pub/Sub message, the jq filter runs on a JSON
+        /// representation of the message body.
```
t4lz: [Is "message body" a known term with GCP Pub/Sub?](https://docs.cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage)
And to be consistent with the existing jq filter for SQS, and to fulfil the same requirements, the jq matching should be done on the whole message, not just the `data` field, but a JSON that contains, also the other fields - `attributes` etc.

t4lz (approved):

PR #4165, opened 2026-04-14
Deny unknown config fields

t4lz:
Currently there are multiple configuration fields in the mirrord configuration schema that allow unknown fields to be specified under them.
This is bad and could trip users in many different ways (AI hallucinates a new field, human makes a typo, human tries to use a field not supported in the mirrord version they're using, etc.). When a non-existing field is specified, we should error out and exit early, instead of (like now) deserialize to some configuration that is not what the user asked for.

- The first commit adds UTs that fail with the existing code.
- The second commit adds the attribute to deny unknown fields from the conf structs that don't have it yet.

Structs that derive `MirrordConfig` have that setting already, but ones that don't currently need to explicitly add that attribute. We have [a technical-backlog item
](https://linear.app/metalbear/issue/MBE-1730/force-new-config-structs-to-deny-unknown-fields) to find a way for us not to keep adding new fields that allow unknown fields.

### Quality Checklist:

- [ ] ~I have checked and updated the relevant existing docs in code, including removing outdated material~
  documentation)

### DIFF
--- a/.vale/styles/config/vocabularies/Metalbear/accept.txt
+++ b/.vale/styles/config/vocabularies/Metalbear/accept.txt
@@ -32,6 +32,10 @@ gcloud
 Tokio
 tokio
 serde
+serialize
+serialization
+deserialize
+deserialization
 reqwest
 rustls
 bincode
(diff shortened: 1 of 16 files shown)


cristeigabriela (approved):
looks good, no comment!

PR #4167, opened 2026-04-14
Make session kill actually terminate the user application

hank-metalbear:
## Summary

The session monitor's `/kill` endpoint (called by both `mirrord session delete` and the UI Kill button) only cancelled the API server's shutdown token. The user app and intproxy kept running, leaving zombie steal/mirror subscriptions on the cluster pod — users had to fall back to \`killall mirrord\`.

This PR has \`/kill\` SIGTERM every layer-connected PID first. When those processes exit, their layer connections close and the intproxy reaches its idle timeout and exits on its own, tearing down the session cleanly.

## Why

Confirmed as unintentional via Slack:
- **2025-12-18 #solutions:** Aviram asked *\"does killing local intproxy make the session go away?\"* — thread has no answer.
- Multiple threads in #imprint, #metalbear-monday-support, #metalbear-capital-on-tap where the workaround for leftover sessions is \`killall mirrord\`.
- Prior debugging observations (#12054–#12057) identified exactly this: *\"API kill endpoint only shuts down API server, not the session.\"*

## Test plan

Tested end-to-end against staging (\`deploy/app-metalbear-co\`):

**Before /kill:**
\`\`\`
app-server         pid 70493 running
mirrord intproxy   pid 70543 running
session socket     ~/.mirrord/sessions/<id>.sock present
\`\`\`

**After POST /kill:**
- Response: \`{\"signaled_pids\":[70493],\"status\":\"shutting_down\"}\`
- app-server exits immediately (SIGTERM)
- Session socket removed
- Intproxy self-exits within ~3s once the last layer disconnects

No zombies remaining.

- [x] \`cargo clippy -p mirrord-intproxy -- -D warnings\` passes
- [x] End-to-end verified against staging
- [ ] Works on Linux (haven't tested, but \`libc::kill\` + \`SIGTERM\` is portable Unix)

## Base

Stacked on #4152 (han/local-api-security-v2) since the session monitor API is introduced there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4986,6 +4986,7 @@ dependencies = [
  "http-body-util",
  "hyper 1.8.1",
  "hyper-util",
+ "libc",
  "mirrord-analytics",
  "mirrord-config",
  "mirrord-intproxy-protocol",
--- a/changelog.d/+session-kill-processes.fixed.md
+++ b/changelog.d/+session-kill-processes.fixed.md
@@ -1,0 +1,1 @@
+`mirrord session delete` and the session monitor UI's Kill button now actually terminate the user application. Previously they only removed the session socket, leaving the user process and intproxy alive and holding zombie steal/mirror subscriptions on the cluster pod.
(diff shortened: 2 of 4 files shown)


hank-metalbear:
Re-opening against main since the bug already exists there (main has `mirrord session kill`).

PR #4168, opened 2026-04-14
Make session kill actually terminate the user application

hank-metalbear:
## Summary

`mirrord session kill` (and the session monitor UI's Kill button) only cancelled the API server's shutdown token. The user app and intproxy kept running, leaving zombie steal/mirror subscriptions on the cluster pod — users were falling back to `killall mirrord`.

This PR has the `/kill` handler SIGTERM every layer-connected PID first. When those processes exit, their layer connections close and the intproxy reaches its idle timeout and exits on its own, cleanly tearing down the session.

## Test plan

Tested end-to-end against staging (`deploy/app-metalbear-co`):

**Before `/kill`:**
```
app-server         pid 70493 running
mirrord intproxy   pid 70543 running
session socket     ~/.mirrord/sessions/<id>.sock present
```

**After `POST /kill`:**
- Response: `{"signaled_pids":[70493],"status":"shutting_down"}`
- app-server exits immediately (SIGTERM)
- Session socket removed
- Intproxy self-exits within ~3s once the last layer disconnects

- [x] `cargo clippy -p mirrord-intproxy -- -D warnings` passes
- [x] End-to-end verified against staging
- [ ] Not tested on Linux, but `libc::kill` + `SIGTERM` is portable Unix

## UI impact

The `mirrord ui` aggregator (TCP `localhost:59281`) is a separate process from the per-session intproxies. Killing one session tears down only that session's intproxy; the aggregator keeps running and the UI stays up.

### DIFF
--- a/mirrord/intproxy/src/session_monitor/api.rs
+++ b/mirrord/intproxy/src/session_monitor/api.rs
@@ -91,12 +91,33 @@ async fn events(
     )
 }
 
-/// Cancels the API server's cancellation token, triggering graceful shutdown of the API server
-/// only. The mirrord session lifecycle is managed separately by the intproxy.
+/// Tears down the mirrord session: `SIGTERM`s every layer-connected process so the user
+/// application exits, then cancels the API server's shutdown token. Once the user app
+/// processes die, their layer connections close and the intproxy exits on its idle timeout.
+///
+/// Without this, `mirrord session delete` (or the UI Kill button) only removes the socket
+/// while the user app and intproxy keep running, leaving zombie steal/mirror subscriptions
+/// on the cluster pod.
 #[cfg(unix)]
 async fn kill(State(state): State<Arc<AppState>>) -> impl IntoResponse {
+    let pids: Vec<u32> = {
+        let info = state.session_info.read().await;
+        info.processes.iter().map(|p| p.pid).collect()
+    };
+
+    for pid in &pids {
+        // SAFETY: `kill(2)` with a process-group or PID argument is a single syscall that
+        // cannot corrupt our process. If the PID is gone or we lack permission, kill returns
+        // -1 and sets errno; we log and continue.
+        let ret = unsafe { libc::kill(*pid as libc::pid_t, libc::SIGTERM) };
+        if ret != 0 {
+            let err = std::io::Error::last_os_error();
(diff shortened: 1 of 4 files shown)


hank-metalbear:
Hey @MintSoup, looks like kill session wasn't actually killing the mirrord process. This should fix it.

mirrord/intproxy/src/session_monitor/api.rs
```diff
@@ -91,12 +91,33 @@ async fn events(
...
+///
+/// Without this, `mirrord session delete` (or the UI Kill button) only removes the socket
+/// while the user app and intproxy keep running, leaving zombie steal/mirror subscriptions
+/// on the cluster pod.
 #[cfg(unix)]
 async fn kill(State(state): State<Arc<AppState>>) -> impl IntoResponse {
+    let pids: Vec<u32> = {
+        let info = state.session_info.read().await;
+        info.processes.iter().map(|p| p.pid).collect()
+    };
```
MintSoup, line 106: Can't this be a one-liner?

mirrord/intproxy/src/session_monitor/api.rs
```diff
@@ -91,12 +91,33 @@ async fn events(
...
+
+    for pid in &pids {
+        // SAFETY: `kill(2)` with a process-group or PID argument is a single syscall that
+        // cannot corrupt our process. If the PID is gone or we lack permission, kill returns
+        // -1 and sets errno; we log and continue.
+        let ret = unsafe { libc::kill(*pid as libc::pid_t, libc::SIGTERM) };
+        if ret != 0 {
+            let err = std::io::Error::last_os_error();
+            tracing::warn!(pid = *pid, %err, "Failed to SIGTERM layer-connected process");
+        }
```
MintSoup, line 116: I think instead of just logging and moving on we should tell the client that we failed to kill all processes, that way it can react somehow

MintSoup (changes requested):

hank-metalbear:
Closing this as Aviram confirmed this is not desire behaviour here: https://metalbear.slack.com/archives/C08TYT624SV/p1776370087897199?thread_ts=1776366044.842629&cid=C08TYT624SV

PR #4170, opened 2026-04-14
Show config diff (not full resolved config) in session monitor

hank-metalbear:
## Summary

The session monitor's `/info` endpoint returns the full resolved `LayerConfig` (hundreds of lines, mostly defaulted `null` values). The monitor UI's Config tab just `JSON.stringify`s it, so users see an unreadable wall of text instead of what they actually configured.

This PR computes a default `LayerConfig` once (`LayerFileConfig::default().generate_config(&mut ConfigContext::default().strict_env(true))`) and serializes only the fields that differ — so the Config tab shows just the user's (or environment's) customizations.

## Test plan

Tested end-to-end against staging:

Before — hundreds of lines, mostly `null`:
```
"accept_invalid_certificates": null,
"agent": { "annotations": null, "check_out_of_pods": null, ... 60+ fields },
"api": true,
"baggage": null,
"ci": { "output_dir": null },
...
```

After — only what the user set:
```json
{
  "feature": {
    "env": {
      "override": { "ADDRESS": "0.0.0.0:4000", "CRM_USE_AUTH": "false" }
    },
    "network": {
      "incoming": {
        "mode": "steal",
        "http_filter": { "header_filter": "x-mirrord: e2e" },
        "port_mapping": [[4000, 8080]]
      }
    }
  },
  "target": { "namespace": "default", "path": { "deployment": "app-metalbear-co" } },
  "kube_context": "gke_metalbear-staging_us-central1_metalbear-staging"
}
```

- [x] `cargo clippy -p mirrord-intproxy -p mirrord-config -- -D warnings` passes
- [x] Verified with a live session against `deploy/app-metalbear-co`

### DIFF
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -22,7 +22,10 @@ use std::{
 };
 
 use mirrord_analytics::{AnalyticsReporter, CollectAnalytics, Reporter};
-use mirrord_config::LayerConfig;
+use mirrord_config::{
+    LayerConfig, LayerFileConfig,
+    config::{ConfigContext, MirrordConfig},
+};
 use mirrord_intproxy::{
     IntProxy,
     agent_conn::{AgentConnectInfo, AgentConnection},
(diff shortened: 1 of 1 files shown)


hank-metalbear:
Superseded — the config diff commit went into main via #4152.

PR #4172, opened 2026-04-15
Fix mirrord ui 404 and update help text (PRO-90, PRO-91)

hank-metalbear:
## Summary

Two QA fixes from [Aviram's session monitor thread](https://metalbear.slack.com/archives/C0AA637E333/p1776246785412539):

### PRO-90 — `mirrord ui` returns 404

Clicking the token URL `mirrord ui` prints returned 404 for every request. Root cause: `mirrord/cli/build.rs` only created an empty `packages/monitor/dist` directory as a placeholder so rust-embed wouldn't panic at compile time — it never actually built the React app.

Fixed by running `npm install && npm run build` in `packages/monitor` as part of `cargo build`, mirroring how `build_wizard_frontend()` already handles the wizard. The existing `ensure_monitor_frontend_dist()` is kept for the clippy fast-path where we don't need real assets.

Tested: `cargo build -p mirrord --release` → `./target/release/mirrord ui` → root URL returns 200 with real `index.html`.

### PRO-91 — CLI help text

Changed the `mirrord ui` subcommand's clap `about` line from `"Launch the session monitor UI"` to `"Launch the mirrord local UI"`.

## Follow-up

PRO-92 (merge wizard + session monitor frontends) not addressed here — it's a bigger refactor that deserves its own PR.

### DIFF
--- a/changelog.d/+monitor-frontend-build.fixed.md
+++ b/changelog.d/+monitor-frontend-build.fixed.md
@@ -1,0 +1,1 @@
+Build the `packages/monitor` React app as part of `cargo build` so `mirrord ui` serves the real UI. Previously the build script only created an empty `dist/` directory, so opening the UI returned 404.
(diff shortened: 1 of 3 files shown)


hank-metalbear:
Superseded by #4171 (Aviram's xtask-based approach is consistent with how the wizard frontend is already built).

PR #4173, opened 2026-04-15
Fix CLI wording + root process race (PRO-91, PRO-94)

hank-metalbear:
## Summary

Two QA fixes from [Aviram's session monitor thread](https://metalbear.slack.com/archives/C0AA637E333/p1776246785412539).

### PRO-91 — CLI help text

Change the `mirrord ui` subcommand's clap `about` line from `"Launch the session monitor UI"` to `"Launch the mirrord local UI"`. Pure text swap in `mirrord/cli/src/config.rs`.

### PRO-94 — Root layer process not appearing

`mirrord exec sh` showed 0 processes in the session monitor; a nested `sh` inside that shell did show up.

**Root cause:** classic publish-before-subscribe race on the tokio broadcast channel that carries `MonitorEvent`s.

```
start_session_monitor()
   let (tx, _rx) = broadcast::channel(256)   ◄── _rx dropped, 0 subscribers
   spawn(start_api_server(tx))               ◄── queued, runs later
   return tx                                 ◄── intproxy starts using tx
```

Inside the spawned `start_api_server` task, after some async I/O setup:

```
spawn(update_session_info_from_events(state))   ◄── this is where subscribe() finally happens
```

If the layer connects between the channel being created and that nested spawn running its `state.monitor_tx.subscribe()`, the `LayerConnected` event hits `tx.send()` with zero receivers and is silently dropped (tokio broadcast drops sends with no receivers; new subscribers don't see backfill; `MonitorTx::emit` ignores the `Err`).

The root process loses this race the most because it connects immediately after intproxy starts. Nested children connect well after the subscriber is alive, so they're captured. Matches Aviram's exact symptom.

**Fix:** subscribe synchronously in `start_session_monitor` before spawning anything, and pass the receiver into `start_api_server` instead of having `update_session_info_from_events` call `subscribe()` later. This guarantees a live receiver from the moment the publisher exists, closing the race window structurally.

```rust
let (tx, _rx) = broadcast::channel(256);
let api_monitor_rx = tx.subscribe();   // ← live receiver from this point on
let proxy_monitor_tx = MonitorTx::from_sender(tx.clone());
...
spawn(start_api_server(..., api_monitor_rx, ...));
```

## Test plan

- [x] `cargo clippy -p mirrord-intproxy -- -D warnings` passes
- [x] Reproduced the bug locally with the unfixed binary (parent PID missing from `processes` list intermittently)
- [x] Verified the fixed binary tracks the root process consistently across multiple runs

### DIFF
--- a/changelog.d/+monitor-root-process.fixed.md
+++ b/changelog.d/+monitor-root-process.fixed.md
@@ -1,0 +1,1 @@
+Fix the root layer process being absent from the local UI processes list. Also update the `mirrord ui` help text from "Launch the session monitor UI" to "Launch the mirrord local UI".
--- a/deny.toml
+++ b/deny.toml
@@ -18,6 +18,8 @@ ignore = [
   "RUSTSEC-2025-0141", # bincode is unmaintaned, but we cannot remove it yet.
   "RUSTSEC-2026-0049", # aws crates in tests/rust-sqs-printer depend on an old version of rustls-webpki
   "RUSTSEC-2026-0097", # rand: unsound only with a custom logger using `rand::rng()`; we don't use that pattern.
+  "RUSTSEC-2026-0098", # rustls-webpki: name constraints for URI names incorrectly accepted; same old rustls-webpki path as RUSTSEC-2026-0049.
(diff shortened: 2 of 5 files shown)


hank-metalbear:
Superseded by #4175 — bundled with PRO-93 + PRO-94 on top of #4171 to single-PR everything.

PR #4176, opened 2026-04-15
Fix cargo-deny AWS rustls path

t4lz:
Closes INT-351
And also solves a new cargo-deny denial: https://github.com/metalbear-co/mirrord/actions/runs/24453915080/job/71449507715?pr=4165#step:4:9137

## Summary

- remove the legacy AWS rustls path from the SQS test dependencies
- update `Cargo.lock` to resolve `rustls-webpki` to `0.103.12`
- remove the stale RustSec ignore from `deny.toml`

## Root cause

AWS default features were pulling both the modern HTTPS client stack and the legacy `tls-rustls` stack, which kept `hyper-rustls 0.24` / `rustls 0.21` in the dependency graph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -655,8 +655,6 @@ checksum = "11493b0bad143270fb8ad284a096dd529ba91924c5409adeac856cc1bf047dbc"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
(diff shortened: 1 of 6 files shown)


dmzmk (approved):
lgtm!

PR #4180, opened 2026-04-15
Aviram/pro 90 session monitor url returns 404 page

hank-metalbear:
### Quality Checklist:

- [ ] I have documented new code sufficiently
- [ ] I have checked and updated the relevant existing docs in code, including removing outdated material
- [ ] I have written user-facing website docs for new features, or opened a (sub)issue to do so
- [ ] I have checked and updated existing website docs for changed features
- [ ] I have tested this change and know it succeeds and fails as expected
- [ ] I have written unit tests or purposefully omitted them
- [ ] I have written e2e tests or purposefully omitted them
- [ ] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs,
  documentation)
- [ ] I have introduced a short, clear and well-formatted changelog entry

PR #4175, opened 2026-04-15
Bundle: PRO-91 + PRO-93 + PRO-94 (on top of #4171)

hank-metalbear:
## Summary

Bundles the three remaining QA fixes from [Aviram's session monitor thread](https://metalbear.slack.com/archives/C0AA637E333/p1776246785412539) on top of his 404 fix (#4171). Single PR for review.

| Ticket | Change |
|---|---|
| **PRO-91** | `mirrord ui` help text: "Launch the session monitor UI" → "Launch the mirrord local UI" |
| **PRO-93** | Fix the squashed mirrord logo in the header; align typography, color tokens, and header layout with the operator-dashboard design system |
| **PRO-94** | Fix root layer process missing from the local UI processes list (broadcast-channel publish-before-subscribe race) |

### PRO-94 — root process race (the meaty one)

`start_session_monitor()` created a tokio broadcast channel, dropped the initial receiver, and spawned `start_api_server` to subscribe later. The intproxy meanwhile started emitting `LayerConnected` events as soon as a layer connected. Tokio broadcast `send` returns `Err` with no live receivers and `MonitorTx::emit` ignores the `Err`, so events fired in that window were silently dropped. Root processes (which connect immediately after intproxy) lost this race the most; nested children connected later and were captured.

Fix: subscribe synchronously in `start_session_monitor` before spawning anything, and pass the receiver into `start_api_server` instead of having `update_session_info_from_events` re-subscribe later. Channel now has a live receiver from the moment the publisher exists.

### PRO-93 — design system alignment

- Logo: icon-only `MirrordIcon` at `w-8 h-8` with `invert` filter for dark mode (matches operator-dashboard's `AppBar`).
- Typography tokens: `text-h1..h4`, `text-body-lg/md/sm`, `font-regular/medium/semibold/bold` from `@metalbear/ui` design tokens.
- Theming: semantic `bg-background`/`text-foreground`/`border-border`/`via-primary` so the whole app themes together via the existing UI-kit `dark` class.
- Layout: `h-14`, gradient accent line, responsive brand block — mirrors operator-dashboard structurally. Drops `max-w-7xl mx-auto` since the monitor uses a full-width sidebar+detail layout.

## Test plan

- [x] `cargo clippy -p mirrord-intproxy -- -D warnings` passes
- [x] `pnpm run build` clean for `packages/monitor`
- [x] Verified PRO-94 fix on playground: `mirrord exec` against `shop/api-gateway` → root process consistently tracked across multiple runs (was flaky before)
[...]

PR #4189, opened 2026-04-16
ci: build monitor frontend in Linux release jobs

hank-metalbear:
Follow-up to #4184 — extends the `corepack enable pnpm` + monitor frontend build to the Linux cross-build release jobs, so Linux binaries also embed the dist instead of shipping an empty one.

## Why

#4184 fixed the macOS CLI release by enabling corepack so `cargo xtask build-cli` could run `pnpm run build` on the monitor frontend. But the Linux release jobs use `cross build -p mirrord` directly and don't go through the xtask path. Their binaries still embed whatever's in `packages/monitor/dist/` at compile time, which is the empty directory `mirrord/cli/build.rs::ensure_monitor_frontend_dist()` creates when the real one is missing. Result: `mirrord ui` on the Linux binary returns 404 for every path.

Windows is intentionally not covered — the local session-monitor API isn't supported on Windows, so its binary never serves the UI.

## What

In each of the two Linux jobs (`build_binaries_aarch64-unknown-linux-gnu`, `build_binaries_x86_64-unknown-linux-gnu`):
1. `corepack enable pnpm` — only pnpm, so the wizard's `npm install` still works on the bundled Node npm.
2. `pnpm install --ignore-workspace --frozen-lockfile && pnpm run build` in `packages/monitor` — produces the dist that rust-embed consumes at the subsequent cross build.

No `rm -rf packages/monitor/node_modules` cleanup this time because npm (wizard) runs *before* pnpm (monitor) in these jobs, not after, so there's no workspace-resolver collision. The xtask needs the cleanup because its order is reversed.

## Supersedes

Takes the place of the CI workflow changes in the closed #4180. That branch had the right idea but was 95% redundant with main after the rest of its content shipped via #4171, #4175, #4181, #4184.

## Verification

- [x] Local `pnpm install --ignore-workspace --frozen-lockfile && pnpm run build` in `packages/monitor` produces a clean `dist/`.
- [ ] Dispatched `release.yaml` from this branch (run 24528101759) to exercise the Linux release jobs end-to-end. When the artifact lands, the binary should contain the built HTML/JS/CSS (verifiable via `strings` probes for `mirrord Session Monitor` + `assets/index-*.js`). Will report here when complete.


hank-metalbear:
@aviramha we need this PR to avoid 404 on linux. I should've merged your original PR as it had all 404 fixes, but made my own branch and missed it.

hank-metalbear:
release binary here: https://github.com/metalbear-co/mirrord/actions/runs/24528101759

.github/workflows/release.yaml
```diff
@@ -148,10 +160,18 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 24
+      - name: Enable corepack (pnpm-only; leave npm alone for wizard build)
```
aviramha: Why don't we leave the build logic in cargo xtask?

.github/workflows/release.yaml
```diff
@@ -109,10 +109,13 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 24
+      - name: Enable corepack (pnpm-only)
+        run: corepack enable pnpm
       - run: npm install
         working-directory: ./packages/wizard
       - run: npm run build
         working-directory: ./packages/wizard
+      - run: cargo xtask build-monitor
```
aviramha: No. cargo xtask build-cli should have the build monitor logic (or call it)
From user perspective (and CI unless we're optimizing and it doesn't seem like it) just call cargo xtask build-cli

hank-metalbear: Could you clarify how you mean? I'm not familiar with the release yaml. It's my first time touching this file, so I just want to be careful.

aviramha: `cargo xtask build-cli` should be the only point where we build stuff. I am pretty sure the npm run build step before isn't even needed. The idea is that whether I build locally or in release.yaml I'll get almost (as possible) identical result without needing extra steps. The only reason I'd split into steps is for setup of env (installing npm) or cache (see previous steps where we build layer etc and then reuse)
Having a specific step for build-monitor isn't needed, unless it's done for cache purposes.

aviramha: ok I thought I changed to use cargo xtask also for the linux bits but seems not.. blah nvm

aviramha:
addressed in https://github.com/metalbear-co/mirrord/pull/4190

PR #4191, opened 2026-04-17
win ʚɞ 38 — mirrord `pitm` (Process In The Middle) for jetbrains ide ext support

cristeigabriela:
# Linear issue

https://linear.app/metalbear/issue/WIN-38/ide-integration-support-for-windows

## Further context

This is a sister pull-request [to the JetBrains IDE extension support PR](https://github.com/metalbear-co/mirrord-intellij/pull/425). It is also a pre-requisite, because the JetBrains plugin will need to implement a version check for the release which will bundle `pitm`.

To see how `pitm` is going to be used, check out the PR. At the date of writing (though likely not applicable at your time of reading,) that PR does not contain proper documentation, as it's gonna be *loooong*. Yay.

## What The Hell Is This I Don't Do Windows I Don't Wanna Be He--

This PR introduces a subcommand, `pitm` (**P**rocess **I**n **T**he **M**iddle). It is basically exec, but it runs in the same context as [`attach`](https://github.com/metalbear-co/mirrord/pull/3995). That is to say, it is invoked after `ext` has been ran, intproxy is running, etc.

The envvars to set, or unset, that will be necessary for the process to run, are passed through a middle-man envvar to not affect the local environment. This is bundled as JSON then base64'd. After everything's resolved, we build the environment variable map for the new process, and we create it.

We create it *suspended*, and only resume once layer has signaled init. Just like exec. This means that layer instanciation worked!

## Why???

I'll point to [`attach`](https://github.com/metalbear-co/mirrord/pull/3995) again for the vein that caused both of these issues. Windows does not have `LD_PRELOAD`, so we need to do some messed up things to make this work.

## How???

`pitm` is used in scenarios where we can swap which binary will be executed, particularly, by the JetBrains family of IDE's (I'm doing my best not to make jokes at it's expense after the hell it's put me Personally through).

- Sometimes, we can control the entire command line: **process** + **arguments**, separately.
  - In that scenario, what we usually do is a *prepend* (eg. `mirrord.exe pitm -- <original>`)
- Sometimes however, we can't do that.
  - This is particularly an issue in IDEA + Java, and as a response we rename `mirrord.exe` to `java.exe` (will be explained in the sister PR) and directly forward the command-line that we've been given to the actual original process, which is passed to us through an established environment variable (see the code.)

### Quality Checklist:

  documentation)


MintSoup (approved):
I'm not versed in windows black magic but this LGTM

btw great docs, very nice 👍

PR #4193, opened 2026-04-17
Add shear for ci, bump actions in check-lint.yaml

aviramha:


.github/workflows/check-lint-markdown.yaml
iniw, line 1: Why merge these back into the same workflow? I thought we were moving away from these monolithic workflow files

aviramha: I want the workflow to be per category

iniw (approved):

PR #4198, opened 2026-04-20
feat(mirrord-cli): extend mirrord ui with operator-session support

hank-metalbear:
## Summary
- Adds `MirrordClusterSession` CRD type to the `mirrord-operator` crate (wire-compatible with the operator-side type from operator#1466; includes `spec.key`).
- Extends `mirrord ui` daemon (`cli/src/ui.rs`) with a `kube::Api<MirrordClusterSession>` watcher that populates an in-memory map and broadcasts `OperatorSession{Added,Removed,Updated}` events on the existing WebSocket.
- Adds `GET /api/operator-sessions` route returning both flat and grouped-by-key views, plus a `watch_status` diagnostic.
- Degrades gracefully when no cluster is reachable or the CRD isn't installed (empty list + `Unavailable` status).

Unblocks PRO-65 (browser-extension "join operator session" flow, upcoming PR in mirrord-browser).

Depends on operator#1466 for `spec.key` to be populated; without it, the watcher runs but `key` will be `null` for all sessions.

## Test plan
- [x] `cargo check --workspace` passes
- [x] `cargo clippy -p mirrord -p mirrord-operator -- --deny warnings` passes
- [x] `cargo fmt --check` clean
- [x] `cargo test -p mirrord --bin mirrord -- ui::` (new summary + grouping tests pass; 16/16)
- [x] `cargo test -p mirrord-operator` passes
- [x] Manual: `mirrord ui` against a live cluster; `GET /api/operator-sessions` returns 6 live sessions with `watch_status: watching`. All sessions have `key: null` because the cluster is not yet running operator#1466; when the operator PR ships, the `by_key` grouping will populate real buckets.

### DIFF
--- a/mirrord/cli/src/ui.rs
+++ b/mirrord/cli/src/ui.rs
@@ -409,6 +409,40 @@ async fn session_events_sse(
         .into_response()
 }
 
+#[derive(Serialize)]
+struct OperatorSessionsResponse {
+    /// Sessions grouped by key. The key is the map key; value is the list
+    /// of sessions under that key. Sessions whose `spec.key` is `None` are
+    /// collected under a sentinel key `""` (empty string).
+    by_key: std::collections::BTreeMap<String, Vec<OperatorSessionSummary>>,
+    /// Flat list of all sessions, for clients that prefer not to consume
+    /// the grouped view.
+    sessions: Vec<OperatorSessionSummary>,
+    /// Current watch status (for UI diagnostics).
+    watch_status: OperatorWatchStatus,
+}
+
+async fn list_operator_sessions(
+    State(state): State<Arc<AppState>>,
+) -> axum::Json<OperatorSessionsResponse> {
+    let map = state.operator_sessions.read().await;
+    let watch_status = state.operator_watch_status.read().await.clone();
+
+    let mut by_key: std::collections::BTreeMap<String, Vec<OperatorSessionSummary>> =
+        std::collections::BTreeMap::new();
+    let sessions: Vec<OperatorSessionSummary> = map.values().cloned().collect();
+    for s in &sessions {
+        let k = s.key.clone().unwrap_or_default();
+        by_key.entry(k).or_default().push(s.clone());
+    }
+
+    axum::Json(OperatorSessionsResponse {
+        by_key,
+        sessions,
+        watch_status,
+    })
+}
+
 async fn kill_session(State(state): State<Arc<AppState>>, Path(id): Path<String>) -> Response {
     let (client, socket_path) = {
         let sessions = state.sessions.read().await;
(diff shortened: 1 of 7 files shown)


aviramha:
The CRD that the cli uses should be mirrord session and namespaced, not cluster wide.

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +62,100 @@ struct FrontendAssets;
...
+    OperatorSessionRemoved {
+        name: String,
+    },
+    OperatorSessionUpdated {
+        session: Box<OperatorSessionSummary>,
+    },
+}
+
+/// Wire-format summary of an operator-side session that the browser extension
+/// and local UI tab can render. Derived from `MirrordClusterSession`.
```
hank-metalbear: If this is derived from MirrordClusterSession, why isn't this struct called MirrordClusterSession?

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +62,100 @@ struct FrontendAssets;
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionSummary {
+    /// Kubernetes object name of the underlying `MirrordClusterSession` CR.
+    pub name: String,
+    /// Session key (grouping identity). `None` if the session was started by an
```
hank-metalbear: Comments are too noisy. Take out these comments.

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +62,100 @@ struct FrontendAssets;
...
+    /// browser extension parses `header_filter` to build a matching DNR rule).
+    /// `None` if the session was started by an older CLI or operator that
+    /// didn't persist it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub http_filter: Option<OperatorSessionHttpFilter>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionHttpFilter {
```
hank-metalbear, line 91: Looks like this is just a string. Why are we making an entirely new struct for this?

mirrord/cli/src/ui.rs
```diff
@@ -74,10 +174,26 @@ impl Serialize for TrackedSession {
 
 struct AppState {
     sessions: RwLock<HashMap<String, TrackedSession>>,
+    operator_sessions: RwLock<std::collections::BTreeMap<String, OperatorSessionSummary>>,
```
hank-metalbear, line 159: Why are we using a btree map here? What's the purpose? why can't we just use HashMap.

hank-metalbear:
Reset: replaced by a smaller PR that adds `MirrordSession` to mirrord-operator using the `operator.metalbear.co` group for API aggregation, instead of the namespaced `mirrord.metalbear.co` variant.

PR #4200, opened 2026-04-21
INT-371 DB Branching fix how we handle secret env vars

vladrbg:
### Quality Checklist:

  documentation)


mirrord/config/src/feature/database_branches.rs
```diff
@@ -386,6 +389,11 @@ pub enum ParamSource {
         #[serde(rename = "secret")]
         name: String,
         key: String,
+        /// Name of the env var to set on the local process from the resolved
+        /// Secret value. When `None`, the operator only uses the Secret for
+        /// branch provisioning and does not inject anything locally.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        env_var_name: Option<String>,
```
t4lz, line 396: Shouldn't it be required? Does it even do anything for the user if this is not provided?
Without it the local app would use the original DB, wouldn't it?

vladrbg: It should but in case someone has that setup so it is backwards compat, but I doubt someone has it because it doesn't work without that. The only thing it does without it is create the branch but local app wont be using the branch without it.

t4lz: We should probably at least show a warning. Can be a future PR though.

t4lz (dismissed):

mirrord/config/src/feature/database_branches.rs
vladrbg: Added it for now on the operator side it was easier with less change

t4lz (approved):

PR #4201, opened 2026-04-21
DB Branching Support composite env and multiple env vars

vladrbg:
### Quality Checklist:

  documentation)

"Fixes INT-385"
"Fixes INT-376"


mirrord/config/src/feature/database_branches.rs
```diff
@@ -1,12 +1,80 @@
...
 use mirrord_config_derive::MirrordConfig;
-use schemars::JsonSchema;
+use schemars::{JsonSchema, Schema, SchemaGenerator};
 use serde::{Deserialize, Serialize, ser::SerializeMap};
 
 use crate::config::{self, source::MirrordConfigSource};
 
+/// Deserializes from either a single value or a JSON array.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct OneOrMany<T>(pub Vec<T>);
```
gememma: `OneOrMany` sounds like it should be a non-empty vec, but this type serializes to/ from either a value or an array of values. If there was a guarantee of non-emptiness, you should be able to create an instance from a single value _or_ a vec (all those `OneOrMany(vec!(..))` could be simplified) and the `first` method should not return an `Option`. It might be helpful to rename this type to something like `SingleOrVec` to make this less confusing, or you could implement the non-empty guarantee if you think it could be helpful. (although i think you should be able to do `OneOrMany::from(T)` either way)

gememma (dismissed):
Approved in case you want to ignore my comment, but I think it's worth having a nicer type name tbh

mirrord/config/src/feature/database_branches.rs
vladrbg: No I like it, I just went with OneOrMany because serde has something similar, but SingleOrVec is more accurate and also the `from` is nice!

gememma (approved):
🌊

PR #4202, opened 2026-04-23
feat(mirrord-operator): add MirrordSession type for API aggregation [PRO-65]

hank-metalbear:
## Summary

Adds a public `MirrordSession` Resource type to `mirrord-operator` so the operator (in metalbear-co/operator) can serve `mirrordsessions.operator.metalbear.co` via API aggregation rather than as a stored CRD. The CLI's `kube::Api<MirrordSession>` calls work transparently against the aggregated endpoint since the type implements `kube::Resource`.

The type is never persisted in etcd; the operator computes responses from its internal `SessionStore` (which mirrors `MirrordClusterSession`) on every request. This preserves user identity through every request, which is the [team-agreed](https://metalbear.slack.com/archives/C08TYT624SV/p1775646314030139) reason to favor API aggregation over standard CRDs for public resources.

This replaces the previous PR #4198 which took the standard-CRD path.

## Test plan

- [ ] `cargo check -p mirrord-operator --features crd` (already verified locally)
- [ ] Operator-side PR (in metalbear-co/operator) consumes this type and serves `kubectl get mirrordsessions -n <ns>`

### DIFF
--- a/changelog.d/+pro-65-http-filter-ui-plumbing.added.md
+++ b/changelog.d/+pro-65-http-filter-ui-plumbing.added.md
@@ -1,0 +1,1 @@
+`mirrord ui` now surfaces each operator session's `http_filter.header_filter` in the `/api/operator-sessions` response, letting the browser extension derive the header it needs to inject instead of assuming a fixed baggage convention.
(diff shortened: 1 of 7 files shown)


hank-metalbear:
Superseded by #4203. Same MVP-pivot as metalbear-co/operator#1476: drop the new `MirrordSession` Resource type in favor of adding a `key` field to the existing `Session` struct in `MirrordOperator.status`. CLI watcher swapped from `Api<MirrordSession>` LIST to polling `MirrordOperator.status.sessions`.

PR #46, opened 2022-05-10
Small PR to control logs via env variables.

meowjesty:
- Use tracing_subscriber::EnvFilter to have better control over logs from env variables;
- Correct pod in launch.json;
- Fix LD_PRELOAD env variable in launch.json.

PR #4203, opened 2026-04-23
feat(mirrord-cli, mirrord-operator): expose joinable sessions for browser extension [PRO-65]

hank-metalbear:
## Summary

MVP path for letting the browser extension list joinable mirrord sessions, per the [team-aligned approach](https://metalbear.slack.com/archives/C09HK8DDX7U/p1776901241393659): reuse `MirrordOperator.status.sessions` (same source `mirrord ls` already uses), no new Resource type, no API aggregation.

- Adds `key: Option<String>` to the existing `Session` struct in `mirrord/operator/src/crd.rs` (additive, backwards compat). Surfaces the user's `key` from mirrord.json.
- `mirrord ui` polls `MirrordOperator.status.sessions` every 5s, projects each entry to an `OperatorSessionSummary`, and exposes them at `GET /api/operator-sessions` (consumed by the browser extension via the `mirrord ui` daemon).
- Browser extension PR (separate, metalbear-co/mirrord-browser#46) consumes that endpoint and groups sessions by key.

Joining a session is purely client-side (browser injects `baggage: mirrord-session=<key>`). No operator-side state mutation. No `http_filter` exposed (per Aviram + Yuan's pushback). Custom filter support deferred to the future write-capable `MirrordSession` schema design.

## Test plan

- [x] `cargo check -p mirrord` clean
- [x] `cargo clippy -p mirrord -- --deny warnings` clean
- [x] `cargo test --bin mirrord` passes (58 tests)
- [ ] End-to-end: deploy operator with matching PR, run `mirrord exec` with a key, run `mirrord ui`, hit `/api/operator-sessions`, see the session
- [ ] Browser extension lists sessions and join injects baggage

## Coordination

Operator-side change in metalbear-co/operator#TBD adds `spec.key` to `MirrordClusterSession` and populates `status.sessions[].key`.

### DIFF
--- a/mirrord/cli/src/ui.rs
+++ b/mirrord/cli/src/ui.rs
@@ -28,6 +28,8 @@ use axum::{
 use axum_extra::extract::cookie::{Cookie, CookieJar, SameSite};
 use eventsource_stream::Eventsource;
 use futures::stream::StreamExt as _;
+use kube::{Api, Client};
+use mirrord_operator::crd::{MirrordOperatorCrd, OPERATOR_STATUS_NAME, Session};
 use mirrord_session_monitor_client::{
     SessionError, connect_to_session, session_socket_entries, sessions_dir,
 };
@@ -54,8 +56,115 @@ struct FrontendAssets;
 #[derive(Clone, Debug, Serialize)]
 #[serde(tag = "type", rename_all = "snake_case")]
 enum SessionNotification {
-    SessionAdded { session: Box<TrackedSession> },
-    SessionRemoved { session_id: String },
+    SessionAdded {
+        session: Box<TrackedSession>,
+    },
+    SessionRemoved {
+        session_id: String,
+    },
+    OperatorSessionAdded {
+        session: Box<OperatorSessionSummary>,
+    },
+    OperatorSessionRemoved {
+        name: String,
+    },
+    OperatorSessionUpdated {
+        session: Box<OperatorSessionSummary>,
+    },
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionSummary {
+    pub name: String,
+    pub key: Option<String>,
+    pub namespace: String,
+    pub owner: Option<OperatorSessionOwner>,
+    pub target: Option<OperatorSessionTarget>,
+    pub created_at: Option<String>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionOwner {
+    pub username: String,
+    pub k8s_username: String,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionTarget {
+    pub kind: String,
+    pub name: String,
+    pub container: String,
+}
+
+impl OperatorSessionSummary {
+    fn from_session(session: &Session) -> Option<Self> {
+        let name = session.id.clone()?;
+        let namespace = session.namespace.clone().unwrap_or_default();
+        let owner = parse_session_owner(&session.user);
+        let target = parse_session_target(&session.target);
+        let created_at = std::time::SystemTime::now()
+            .checked_sub(std::time::Duration::from_secs(session.duration_secs))
+            .map(|t| {
+                let secs = t
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .map(|d| d.as_secs())
+                    .unwrap_or(0);
+                k8s_openapi::apimachinery::pkg::apis::meta::v1::Time(
+                    k8s_openapi::jiff::Timestamp::from_second(i64::try_from(secs).unwrap_or(0))
+                        .unwrap_or(k8s_openapi::jiff::Timestamp::UNIX_EPOCH),
+                )
+                .0
+                .to_string()
+            });
+        Some(Self {
+            name,
+            key: session.key.clone(),
+            namespace,
+            owner,
+            target,
+            created_at,
+        })
+    }
+}
+
+fn parse_session_owner(user: &str) -> Option<OperatorSessionOwner> {
+    let (head, _hostname) = user.rsplit_once('@')?;
+    let (username, k8s_username) = head.split_once('/')?;
+    Some(OperatorSessionOwner {
+        username: username.to_owned(),
+        k8s_username: k8s_username.to_owned(),
+    })
+}
+
+fn parse_session_target(target: &str) -> Option<OperatorSessionTarget> {
+    if target == "targetless" {
+        return None;
+    }
+    let (kind_short, rest) = target.split_once('.')?;
+    let kind = match kind_short {
+        "deploy" => "Deployment",
+        "pod" => "Pod",
+        "rollout" => "Rollout",
+        "job" => "Job",
+        "cronjob" => "CronJob",
+        "statefulset" => "StatefulSet",
+        "service" => "Service",
+        "replicaset" => "ReplicaSet",
+        other => other,
+    }
+    .to_owned();
+    let (name, container) = rest
+        .split_once(".container.")
+        .map(|(n, c)| (n.to_owned(), c.to_owned()))
+        .unwrap_or_else(|| (rest.to_owned(), String::new()));
+    Some(OperatorSessionTarget {
+        kind,
+        name,
+        container,
+    })
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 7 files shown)


mirrord/cli/src/ui.rs
```diff
@@ -516,6 +658,83 @@ fn start_filesystem_watcher(
...
+
+        *state.operator_watch_status.write().await = OperatorWatchStatus::Watching;
+
+        let mut interval = tokio::time::interval(std::time::Duration::from_secs(5));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        loop {
+            interval.tick().await;
+            match api.get(OPERATOR_STATUS_NAME).await {
+                Ok(operator) => {
+                    let sessions = operator
```
hank-metalbear: Can we refactor what we are doing inside the OK bracket into another function?

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionSummary {
+    pub name: String,
+    pub key: Option<String>,
+    pub namespace: String,
+    pub owner: Option<OperatorSessionOwner>,
+    pub target: Option<OperatorSessionTarget>,
+    pub created_at: Option<String>,
```
hank-metalbear: should be required

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+    OperatorSessionUpdated {
+        session: Box<OperatorSessionSummary>,
+    },
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionSummary {
+    pub name: String,
+    pub key: Option<String>,
```
hank-metalbear: should be required

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+    pub namespace: String,
+    pub owner: Option<OperatorSessionOwner>,
```
hank-metalbear: should be required

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+#[serde(rename_all = "camelCase")]
+pub struct OperatorSessionTarget {
+    pub kind: String,
+    pub name: String,
+    pub container: String,
+}
+
+impl OperatorSessionSummary {
+    fn from_session(session: &Session) -> Option<Self> {
+        let name = session.id.clone()?;
```
hank-metalbear: why name? Shouldn't it be id?

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+    let (head, _hostname) = user.rsplit_once('@')?;
+    let (username, k8s_username) = head.split_once('/')?;
+    Some(OperatorSessionOwner {
+        username: username.to_owned(),
+        k8s_username: k8s_username.to_owned(),
+    })
+}
+
+fn parse_session_target(target: &str) -> Option<OperatorSessionTarget> {
+    if target == "targetless" {
```
hank-metalbear: do you have to hardcode "targetless" isn't there a constant somwhere?

mirrord/cli/src/ui.rs
```diff
@@ -54,8 +56,104 @@ struct FrontendAssets;
...
+        return None;
+    }
+    let (kind_lower, name) = target.split_once('/')?;
+    let mut chars = kind_lower.chars();
+    let kind = chars
+        .next()
+        .map(|c| c.to_ascii_uppercase().to_string() + chars.as_str())
```
hank-metalbear: just keep lowercase

mirrord/cli/src/ui.rs
```diff
@@ -516,6 +658,88 @@ fn start_filesystem_watcher(
...
+        *state.operator_watch_status.write().await = OperatorWatchStatus::Watching;
+
+        let mut interval = tokio::time::interval(std::time::Duration::from_secs(5));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        loop {
+            interval.tick().await;
+            match api.get(OPERATOR_STATUS_NAME).await {
+                Ok(operator) => reconcile_operator_sessions(&state, &operator).await,
+                Err(err) => {
+                    let reason = format!("operator status fetch error: {err}");
```
hank-metalbear, line 699: If this breaks, everything breaks. Shouldn't you panic quit here?

mirrord/cli/src/ui.rs
```diff
@@ -516,6 +658,88 @@ fn start_filesystem_watcher(
...
+        .map(|s| s.sessions.as_slice())
+        .unwrap_or_default()
+        .iter()
+        .filter_map(OperatorSessionSummary::from_session)
+        .map(|s| (s.name.clone(), s))
+        .collect();
+
+    let mut map = state.operator_sessions.write().await;
+
+    for (name, summary) in &observed {
```
hank-metalbear: This could be refactored to another function.

mirrord/cli/src/ui.rs
```diff
@@ -516,6 +658,88 @@ fn start_filesystem_watcher(
...
+            }
+        } else {
+            SessionNotification::OperatorSessionUpdated {
+                session: Box::new(summary.clone()),
+            }
+        };
+        let _ = state.notify_tx.send(event);
+    }
+
+    let stale: Vec<String> = map
```
hank-metalbear, line 750: Same here. This could be another function to process still sessions.

mirrord/cli/src/ui.rs
```diff
@@ -828,4 +1078,75 @@ mod tests {
...
+    mod operator_sessions {
+        use mirrord_operator::crd::Session;
+
+        use super::*;
+
+        fn sample_session(id: &str, key: &str) -> Session {
+            Session {
+                id: Some(id.to_owned()),
+                duration_secs: 60,
+                user: "alice/alice@ex@host".to_owned(),
```
hank-metalbear, line 1107: this probably wouldn't happen in real life, wouldn't there be only 1 @ sign?

mirrord/cli/src/ui.rs
hank-metalbear: Done in `154620b4a` — extracted into `reconcile_operator_sessions`.

mirrord/cli/src/ui.rs
hank-metalbear: Done in `a43eab23f` — `key` is now `String`. `from_session` returns `None` when the operator's `Session.key` is None, so keyless sessions get filtered out at the CLI and the browser only sees joinable ones.

mirrord/cli/src/ui.rs
hank-metalbear: Done in `a43eab23f` — `owner` is now `OperatorSessionOwner` (required). `from_session` returns `None` if the owner string fails to parse.

mirrord/cli/src/ui.rs
hank-metalbear: Done in `a43eab23f` — `created_at` is now `String`. `from_session` returns `None` if timestamp construction fails (shouldn't happen in practice).

mirrord/cli/src/ui.rs
hank-metalbear: Agreed, renamed to `id` in `a43eab23f`. Also renamed `OperatorSessionRemoved { name }` → `{ id }`. Browser PR #46 updated to match.

mirrord/cli/src/ui.rs
hank-metalbear: Good catch. Using `mirrord_operator::crd::TARGETLESS_TARGET_NAME` now, fixed in `a43eab23f`.

mirrord/cli/src/ui.rs
hank-metalbear: Done. Dropped the first-letter capitalization, passing through whatever case the operator emits. Fixed in `a43eab23f`.

mirrord/cli/src/ui.rs
hank-metalbear: Pushing back here. `mirrord ui` serves two independent things: (1) local-session monitoring (filesystem watcher on `~/.mirrord/sessions/`) and (2) this operator-session listing. If the operator is unreachable, (1) still works — panicking would kill it too. The `Unavailable { reason }` status lets the frontend show "operator unavailable: <reason>" while local sessions keep working. Open to reconsidering if you have a reason I'm missing.

mirrord/cli/src/ui.rs
hank-metalbear: Done in `a43eab23f` — extracted into `apply_observed_sessions`.

mirrord/cli/src/ui.rs
hank-metalbear: Done in `a43eab23f` — extracted into `drop_stale_sessions`.

mirrord/cli/src/ui.rs
hank-metalbear: Usually yes, but `k8s_username` is often an email (e.g. `alice@company.com`). `SessionOwner::Display` formats as `"{username}/{k8s_username}@{hostname}"`, so when k8s_username contains its own `@`, the full string ends up with two. `rsplit_once('@')` splits on the last one so `hostname` gets peeled off, leaving `"alice/alice@company.com"` in the head for the username/k8s_username split. `split_once('@')` would break this case.

PR #4206, opened 2026-04-23
make it possible to run the integration tests with the devshell on macos

iniw:
This only required adding the relevant packages to the devshell and forcing the tests to use the system `cat` instead of grabbing it from `$PATH`, which is problematic in this case. See the first commit's description for a bit more context.

### Quality Checklist:

- [-] I have written user-facing website docs for new features, or opened a (sub)issue to do so
- [-] I have checked and updated existing website docs for changed features
  documentation)


meowjesty (approved):
👍

PR #4218, opened 2026-04-24
Config: Support rendering templates in the `key` field

iniw:
## Summary

This PR adds template expansion support to the root `key` config field.

`key` is a bit unusual because it can be set in the config file, and it can
also be used elsewhere in the same file through `{{ key }}`. We already resolve
it early so the rest of the config can use it, but until now that early step
only treated `key` as a plain string.

With this change, that early step also renders template expressions, so configs
like this now work:

```json
{
  "key": "{{ get_env(name=\"USER\", default=\"foo\") }}"
}
```

## Notes

This also fixes a small related bug where auto-generated keys could expose an
internal marker in `{{ key }}`. Templates now always get the real session key.

See the first commit's description for more detail.

### Quality Checklist:

  documentation)


mirrord/config/src/lib.rs
```diff
@@ -993,19 +993,32 @@ impl LayerFileConfig {
...
         let mut tera_context = tera::Context::new();
-        tera_context.insert("key", &key);
+        tera_context.insert(
+            "key",
+            // Auto-generated keys are stored in the context with an internal marker so the later
+            // config resolution step can still tell they were generated, but templates
+            // should only ever see the actual session key value.
+            key.strip_prefix(EnvKey::AUTOGENERATED_MARKER)
+                .unwrap_or(&key),
+        );
```
gememma, line 1021: i feel like the fact that we have the `EnvKey` type should prevent this, no? can we not `let key: EnvKey = match { whatever }.into()` or something and have `EnvKey` do the prefix handling? then just add `key.as_str()` to the context

gememma (approved):
nicenicenice, 1 suggestion 🚲

mirrord/config/src/lib.rs
iniw: But at this point we don't have an `EnvKey` yet, because we haven't resolved the config. It's a bit of a chicken and egg thing because to get an `EnvKey` we need the textual key, because tera needs to know about it.

It's a bit of a mess...

PR #4220, opened 2026-04-27
FIX libuv getaddrinfo EAI errors

t4lz:
## Summary
- return `EAI_*` codes from the Unix `getaddrinfo` detour instead of generic hook error conversions
- keep generic hook-error side effects for non-DNS failures, but collapse them to `EAI_FAIL` for `getaddrinfo` callers
- add a Node regression test that reproduces the empty-remote-DNS-response path without aborting in libuv

## Root Cause
`getaddrinfo(3)` reports failures through its return value, not through `errno`. The Unix mirrord detour was using the generic `Detour::Error -> HookError::into()` path, so `HookError::DNSNoName` and similar non-`DnsLookup` errors could become `-1`/`errno`-style failures instead of valid `EAI_*` results. libuv validates those return codes and aborts when it sees an unknown value.

## Fix
Handle Unix `getaddrinfo` errors explicitly. `ResponseError::DnsLookup` still maps to the native DNS codes, `DNSNoName` now maps to `EAI_NONAME`, and other hook errors preserve their logging/graceful-exit behavior before returning `EAI_FAIL` to the caller.

### DIFF
--- a/mirrord/layer-lib/src/error.rs
+++ b/mirrord/layer-lib/src/error.rs
@@ -552,6 +552,7 @@ pub fn getaddrinfo_error_code(fail: HookError) -> i32 {
         }
         HookError::DNSNoName => libc::EAI_NONAME,
         other => {
+            // The conversion has side effects - logs, possible gracefull exit, etc.
             let _ = i64::from(other);
             libc::EAI_FAIL
         }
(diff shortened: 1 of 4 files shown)


mirrord/layer-lib/src/error.rs
```diff
@@ -539,6 +539,25 @@ fn translate_dns_fail(dns_fail: DnsLookupError) -> i32 {
...
+/// return value.
+#[cfg(unix)]
+pub fn getaddrinfo_error_code(fail: HookError) -> i32 {
+    match fail {
+        HookError::ResponseError(ResponseError::DnsLookup(dns_fail)) => {
+            translate_dns_fail(dns_fail)
+        }
+        HookError::DNSNoName => libc::EAI_NONAME,
+        other => {
+            let _ = i64::from(other);
```
iniw, line 556: Unused?

mirrord/layer-lib/src/error.rs
```diff
@@ -666,3 +685,37 @@ impl From<HookError> for *mut hostent {
...
+    }
+
+    #[test]
+    fn getaddrinfo_maps_other_hook_errors_to_eai_fail() {
+        assert_eq!(
+            getaddrinfo_error_code(HookError::NullPointer),
+            libc::EAI_FAIL
+        );
+    }
+}
```
iniw: I don't see the value in these unit tests since we have the integration test. The function is super tiny anyway.

0x00A5 (dismissed):

mirrord/layer-lib/src/error.rs
t4lz: removed UTs

mirrord/layer-lib/src/error.rs
t4lz: This conversion has side effects - graceful exit, logs, etc.

t4lz: Adding comment to explain

iniw (approved):

PR #4221, opened 2026-04-27
mirrord ui: auto-configure browser extension via externally_connectable

hank-metalbear:
## Summary

- The `mirrord ui` web frontend now sends `{type: 'mirrord-ui-configure', backend, token}` to the mirrord browser extension via `chrome.runtime.sendMessage` on page load. The extension writes the values into `chrome.storage.local` and the popup picks them up immediately, eliminating the `chrome-extension://...configure.html?...` URL paste step.
- Bind on `Ipv4Addr::LOCALHOST` instead of `Ipv6Addr::LOCALHOST`. The browser extension's `externally_connectable.matches` allowlist accepts `http://localhost/*` and `http://127.0.0.1/*` but not `[::1]`, so the IPv6 origin would silently fail the handshake.
- Drops the `chrome-extension://...configure.html?...` URL print and the `build_extension_configure_url` helper from `mirrord/cli/src/ui.rs`. The auto-configure handshake replaces the manual paste flow entirely; no fallback is kept.

Pairs with metalbear-co/mirrord-browser branch `han/pro-65-clean-browser` (which adds `externally_connectable` + the `onMessageExternal` listener).

## Test plan

- [x] `cargo clippy -p mirrord --bin mirrord --no-default-features` — clean
- [x] `cargo fmt`
- [x] `pnpm build` in `packages/monitor`
- [ ] Local: run new CLI, open Web UI in Chrome with extension loaded, observe Sessions tab populates without URL paste


mirrord/cli/src/ui.rs
```diff
@@ -856,18 +854,6 @@ pub async fn ui_command(args: UiArgs) -> Result<(), CliError> {
...
 }
 
-const MIRRORD_EXTENSION_ID: &str = "bijejadnnfgjkfdocgocklekjhnhkhkf";
-
-fn build_extension_configure_url(addr: &SocketAddr, token: &str) -> String {
-    let backend = format!("http://{addr}");
-    let backend_encoded: String =
-        url::form_urlencoded::byte_serialize(backend.as_bytes()).collect();
-    format!(
-        "chrome-extension://{id}/pages/configure.html?backend={backend_encoded}&token={token}",
```
vladrbg, line 866: The summary we have this:

`The chrome-extension:// fallback URL is still printed to stderr so the manual paste flow keeps working with older extension builds.`

But i see it we are removing it here? Is this correct?

hank-metalbear: correct, i will update the pr description.

vladrbg: Oh no I am just checking if correct, if thats intended then no need, we are good!

vladrbg (approved):
lgtm

PR #4222, opened 2026-04-28
INT-374 LocalRedis cleanup is skipped on interrupt

vladrbg:
### Summary
- Added a cleanup guardian process that watches the CLI's PID and removes the local Redis instance when it exits.
- Moved `is_pid_alive` and `cleanup guardian` logic from `db_branches.rs` and `local_redis.rs` to `util.rs` for reuse.

### Quality Checklist:

  documentation)

### DIFF
--- a/changelog.d/+local-redis.fixed.md
+++ b/changelog.d/+local-redis.fixed.md
@@ -1,0 +1,1 @@
+Fixed local Redis containers/processes not being cleaned up.
--- a/changelog.d/+local-redis.fixed.md
+++ b/changelog.d/+local-redis.fixed.md
@@ -1,1 +1,1 @@
-Fixed local Redis containers/processes not being cleaned up.
+Fixed local `Redis` containers/processes not being cleaned up.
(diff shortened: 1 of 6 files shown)


iniw (commented):
Can't we piggyback on the intproxy for this, since it already has a cleanup path for layer closing (EOF on the connection)? Or maybe the finalizer of `MirrordClusterSession`?

hank-metalbear:
I agree with @iniw, is it high effort @vladrbg? Adding a guardian process might introduce more moving parts. If we can leverage intproxy life cycle, it might be net positive.

vladrbg:
> Can't we use piggyback on the intproxy for this, since it already has a cleanup path for layer closing (EOF on the connection)? Or maybe the finalizer of `MirrordClusterSession`?

Not really, the session finalizer runs on the operator but this is about the local redis only, and the intproxy would need extra logic specific to handle local Redis cleanup. So thats why I went this approach.

iniw (approved):
> Not really, the session finalizer runs on the operator but this is about the local redis only, and the intproxy would need extra logic specific to handle local Redis cleanup. So thats why I went this approach.

Yeah, I can see that. We could introduce a new message in the intproxy <-> protocol to register a new Redis to clean up, but it does sound more complex/over-engineered.

I don't think this is super nice, but I don't think it _can_ be super nice unless we add a more generic way to run post-`exec` code. It is what it is.

PR #4226, opened 2026-04-29
Fix image_pull_secrets example in config docs

eyalb181:
## Summary
- Fix the top-level configuration example for `agent.image_pull_secrets` — the key was incorrectly written as `secret-key` instead of `name` (the actual struct field, as shown correctly in the detailed docs further down).
- Reported via website chat: example on https://metalbear.com/mirrord/docs/config showed `[ { "secret-key": "secret" } ]`.
- Updates both `mirrord/config/src/agent.rs` and `mirrord/config/src/lib.rs`.

## Test plan
- [ ] CI passes
- [ ] Verify regenerated docs use `{ "name": "secret" }`


gememma (dismissed):
💂

gememma (approved):

PR #4227, opened 2026-04-29
Cor 1406 support unix socket in mounted directory in the remote pod

aviramha:


mirrord/agent/src/task.rs
```diff
@@ -54,12 +54,15 @@ impl BgTaskRuntime {
...
             if let Some(RuntimeNamespace {
                 target_pid,
-                namespace_type,
+                namespace_types,
             }) = namespace
-                && let Err(error) = namespace::set_namespace(target_pid, namespace_type)
             {
-                let _ = result_tx.send(Err(error.into()));
-                return;
+                for namespace_type in namespace_types {
```
0x00A5, line 60: nit: Don't need a for loop here.
The man page says if fd is the target pid, we can set multiple ns with one call.
```
int fd = pidfd_open(1234, 0);
setns(fd, CLONE_NEWUSER | CLONE_NEWNET | CLONE_NEWUTS);
```

0x00A5 (approved):

PR #4230, opened 2026-04-29
feat(packages/monitor): operator-sessions view in the local UI

hank-metalbear:
## Summary

Adds a TEAM section to the local session monitor showing operator-side sessions visible to the current user, plus the supporting backend and UX work.

### Backend (`mirrord/cli/src/ui.rs`)
- `/api/me` — calls Kubernetes `SelfSubjectReview` via the existing kube-rs client and returns `{ k8sUsername }`. Lets the dashboard classify operator sessions as "yours" vs "team" without any user setup.
- `OperatorSessionSummary` enriched with `durationSecs`, `lockedPorts`, and `queueSplits` (sqs/rmq/kafka counts) so the detail view can show what the session is actually doing cluster-side.

### Frontend (`packages/monitor`)
- **YOURS / TEAM split.** YOURS = local intproxy sessions on this machine + operator sessions whose `owner.k8sUsername` matches `/api/me`. TEAM = everything else.
- **Operator session detail panel** with Overview / Config tabs, matching the local SessionDetail visual structure: identity card, uptime + locked-port + queue-split stat strip, locked-ports list, HTTP filter row, queue splits when non-zero. (Events tab is intentionally local-only — operator API doesn't stream per-session traffic events.)
- **One JoinBar component** used by both local and operator detail views — you can now join your own `mirrord exec` session directly from YOURS, not just teammates' sessions in TEAM.
- **One unified search bar** at the top of the sidebar, replacing the per-section search. Filters both YOURS and TEAM.
- **Stable ordering on heartbeats.** WS `operator_session_updated` used to filter+append, which shuffled positions on every heartbeat; now replaces in place by id index. Within-group sort by session id so two sessions sharing a key hold their positions even when both update simultaneously.
- **No more flicker on Connected → Disconnected.** `selectedKind` was in the WS effect's deps, so every session click tore the WS down and reconnected; moved to a ref so the WS lives once per mount.
- **Subtle scrollbar + resize handle.** The previously solid-purple scrollbar/handle is now muted, so the dividing strip between sidebar and detail no longer dominates the page.
- **"AllAll" namespace filter** dedupe fix.

## Test plan

- [ ] Build CLI with new frontend embedded: `cargo xtask build-cli --release --no-wizard`.
- [ ] `mirrord ui --port 59281`; open the printed URL; YOURS picks up sessions you've started, TEAM picks up teammates' sessions.
- [ ] Run `mirrord exec --target deploy/X -- ...` → session shows up in YOURS with full Overview/Events/Config tabs.
- [ ] Click your own local session → "Join" row appears; click it → browser extension joins the same key.
- [ ] Two sessions sharing a key: positions stay stable through multiple operator heartbeats.
- [ ] Click between sessions: "Connected" indicator stays green throughout (no flicker to "Disconnected").
- [ ] `/api/me` returns the kubectl identity: `curl /api/me?token=...`.


hank-metalbear:
Split into #4245 (Rust slice: `/api/me`, enriched `OperatorSessionSummary`, `parse_session_owner` fallback) and #4246 (TypeScript slice: `packages/monitor` UI work). Same content, smaller surface per reviewer.

PR #4238, opened 2026-04-30
Free disk space before starting E2E tests to avoid running out

MintSoup:


.github/workflows/test-e2e.yaml
```diff
@@ -19,6 +19,16 @@ jobs:
     env:
       MIRRORD_AGENT_RUST_LOG: "warn,mirrord=debug"
     steps:
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
```
0x00A5: should use our fork instead

MintSoup: no idea we had one

iniw (approved):
Hell yeah. Disk space!

PR #4245, opened 2026-05-01
feat(cli): /api/me and enriched OperatorSessionSummary in mirrord ui

hank-metalbear:
Rust slice of the local-UI work, split out of the original combined PR (#4230) so reviewers can review Rust and TypeScript separately.

## What this adds

- **`/api/me` endpoint** in the `mirrord ui` Axum server. Backed by Kubernetes `SelfSubjectReview`. Returns `{ k8sUsername }` so the local UI can render the current user's identity (used for the avatar chip in the header and for filtering YOURS vs TEAM operator sessions).
- **`OperatorSessionSummary` enriched** with three fields the UI renders in the operator-session detail panel:
  - `duration_secs` (how long the session has been alive)
  - `locked_ports` (with new `OperatorLockedPort` type: port + kind + filter)
  - `queue_splits` (with new `OperatorQueueSplits` type: SQS / RabbitMQ / Kafka counts)
- **`parse_session_owner` fallback**: when a `session.user` string doesn't match the standard `username/k8s_username@host` shape, fall back to using the raw string for both fields instead of dropping the session. Standard operator-owned sessions still parse the same way; preview-env sessions and other non-standard owners now make it through. See operator/#1511 for the producer side.

## Test plan

- [ ] `cargo check -p mirrord --keep-going` (with `MIRRORD_LAYER_FILE` set, as CI does)
- [ ] `mirrord ui` against an operator-installed cluster:
  - `GET /api/me` returns `{ k8sUsername: "..." }`
  - `GET /api/operator-sessions` returns sessions including `durationSecs`, `lockedPorts`, `queueSplits`
  - Sessions with non-standard `session.user` strings (e.g. preview-env) appear instead of being dropped

## Related

- TypeScript UI consuming these is **#4246**.
- PreviewSession projection that produces non-standard `session.user` strings is operator/#1511.


deny.toml
```diff
@@ -17,6 +17,7 @@ ignore = [
   "RUSTSEC-2025-0134", # rustls-pemfile is unmaintained, but we cannot remove it yet.
   "RUSTSEC-2025-0141", # bincode is unmaintaned, but we cannot remove it yet.
   "RUSTSEC-2026-0097", # rand: unsound only with a custom logger using `rand::rng()`; we don't use that pattern.
+  "RUSTSEC-2026-0119", # hickory-proto 0.24 DoS via name compression. Pinned by hickory-resolver 0.24; bumping requires a separate dependency upgrade PR. Our DNS path goes through the agent in a controlled cluster, not exposed to untrusted DNS messages.
```
MintSoup: hickory has been bumped, can remove this

MintSoup (changes requested):

deny.toml
hank-metalbear: Done in 6cde1d7a0. Merged main and dropped the ignore. Thanks for the catch.

MintSoup (approved):
looks alright

PR #4250, opened 2026-05-03
HTTP detection timeout improvements

fergusean:
Detection on redirected connections previously waited indefinitely for the client's first byte before the timeout started, which stalled server-first protocols like SMTP. The timer now starts when detection begins, and the value is configurable via `agent.http_detection_timeout` (default 2s).

### Quality Checklist:

- [X] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs,<br>documentation)


fergusean:
See https://github.com/metalbear-co/mirrord/pull/4209#issuecomment-4325537860 for minimal reproduction of server-first protocol stall.

aviramha:
Thanks for contributing! We're having bit of delay from our side due to company retreat so might take longer than usual.

mirrord/agent/src/http.rs
```diff
@@ -135,13 +137,12 @@ where
...
 
     while detected.is_known().not() {
-        let timeout_fut = OptionFuture::from(timeout_at.map(tokio::time::sleep_until));
-
         let result = tokio::select! {
-            Some(..) = timeout_fut => break,
+            biased;
+            _ = tokio::time::sleep_until(timeout_at) => break,
             result = stream.read_buf(&mut buf) => result,
         };
```
iniw: Can this be simplified to something like:
```suggestion
        let result = tokio::time::timeout(timeout, stream.read_buf(&mut_buf)).await;
```

iniw (changes requested):
I'm not sure I agree on the behavior of not checking for an HTTP header at all if timeout is 0.

If we move to using [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) like I suggested, we'll get what I consider to be the correct behavior for free:

> If the provided future completes immediately, then the future returned from this function is guaranteed to complete immediately with an [Ok](https://doc.rust-lang.org/nightly/core/result/enum.Result.html#variant.Ok) variant no matter the provided duration.

That is, if bytes are already written (like in the unit test you added), we simply try determining the version using the bytes we have.

Let me know your thoughts on this, please :)

fergusean:
My thought had been that with a value of 0, the result is timing-dependent (depending on remote RTT, scheduling, etc), so being explicit about the behavior of 0 was the safer option.

That said, it solves my use case (support for server-first protocols without any post-connection delay) regardless, so I'm happy either way!

Changes applied.

iniw (approved):
Cool! Thanks for the contribution

PR #4257, opened 2026-05-10
chore(cli): refresh prompt messages and add successful-usage prompt

eyalb181:
## Summary

Adds a proactive "successful-usage" prompt that suggests mirrord for Teams to repeat OSS users who haven't adopted the operator yet. Two thresholds (10 and 50 runs), gated on `!uses_operator`.

Also refreshes the newsletter prompt: drops the 5-run threshold (too soon to nudge) and ports the body sentences that were already shown by the IntelliJ plugin.

Also fixes a pre-existing rendering bug: both `suggest_newsletter_signup` and the new `suggest_operator_features` were using `progress.add_to_print_buffer(...)`, which only flushes on the parent progress's graceful `success()` call. Because `mirrord exec` ends with an `execve()` that replaces the process image, Rust destructors never run and the buffered prompt was silently dropped. Swapped to `progress.print(...)` (same path the `multipodwarn` / `httpfilter` nudges already use) so the prompts actually render. Also moved the calls to fire before the config-summary subtask, grouping all the Teams-related nudges at the top of the output instead of interleaving them with the config display.

Verified end-to-end against staging: the 10-run prompt now appears in the user's terminal alongside the existing multipod warning.

## Coordinated PRs

Same product surface across the three plugins:
- CLI: this PR
- IntelliJ: metalbear-co/mirrord-intellij#435
- VSCode: metalbear-co/mirrord-vscode#245

## Test plan
- [ ] CI passes
- [ ] Run mirrord 10 times without operator → first successful-usage prompt fires visibly in stderr
- [ ] Run mirrord 50 times without operator → second successful-usage prompt fires
- [ ] Verify nothing fires at 5 runs (was the previous newsletter threshold)
- [ ] Run with operator → no successful-usage prompts


mirrord/cli/src/newsletter.rs
```diff
@@ -54,7 +49,7 @@ pub async fn suggest_newsletter_signup<P: Progress>(user_data: &mut UserData, pr
 
     if let Some(message) = newsletter_invites.get(&current_sessions) {
         // print the chosen invite to the user if progress mode is on
-        progress.add_to_print_buffer(
+        progress.print(
```
meowjesty, line 52: Changing this to `print` looks a bit ugly (imo).

I tried to propose something different for our info messages and user suggestions in: https://linear.app/metalbear/issue/COR-1170/add-oss-to-teams-redirect-messages-in-cli-for-paid-features#comment-10fc7d18

Image

mirrord/cli/src/successful_usage_prompt.rs
```diff
@@ -0,0 +1,54 @@
...
+            PROMPT_AFTER_SECOND,
+            (
+                "mirrord for Teams unlocks team workflow features: DB branching for parallel devs, preview environments for branch testing, and shared targets with queue splitting.",
+                "prompt-2",
+            ),
+        ),
+    ]);
+
+    if let Some((message, content_tag)) = prompts.get(&session_count) {
+        progress.print(
```
meowjesty, line 44: Same comment here.

I think what we should do cli-wise, is to have a header (generic or maybe with context based on the user's config/cluster info) that appears before the progress' messages. And we should give the user a way to disable these _hints_ without having to disable the full progress.

Wdyt?

meowjesty (approved):
👍

PR #4268, opened 2026-05-14
Mirrord up init

MintSoup:
### Quality Checklist:

  documentation)

See 
https://linear.app/metalbear/issue/COR-1397/add-mirrord-up-init-wizard-for-generating-skeleton-mirrord-upyaml
https://github.com/metalbear-co/docs/pull/181

### DIFF
--- a/mirrord/up/src/init.rs
+++ b/mirrord/up/src/init.rs
@@ -1,0 +1,582 @@
+//! Interactive wizard that generates a skeleton `mirrord-up.yaml`.
+//!
+//! The wizard prompts for a small, curated subset of the full schema and
+//! writes the result as YAML. For fields the wizard knows about but the
+//! user did not customize, the output contains a commented-out template
+//! line so the user can discover the field without consulting the docs.
+//!
+//! Keep the prompt set deliberately small. The values that go *into* each
+//! prompt (enum variants, defaults) must always come from the authoritative
+//! type definitions in `config.rs` (via `strum::VariantArray`, `Default`,
+//! etc.), so adding a new variant downstream never requires editing this
+//! file.
+//!
+//! The hand-rolled emitter is necessary because `serde_yaml` cannot emit
+//! comments. A round-trip unit test guards against the wizard's output
+//! ever diverging from what [`crate::load_up_config`] can parse.
+
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Write as _,
+    ops::Not,
+    path::PathBuf,
+    str::FromStr,
+};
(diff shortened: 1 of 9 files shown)


mirrord/up/src/init.rs
0x00A5, line 1: Played with it a bit, a few things I noticed:
1. if user enter they don't use operator, should the questions about operator all get skipped?
2. any way to go back to previous steps if the template preview doesn't look good?

MintSoup: 1. What other questions about the operator do we have? If you mean the other 2 questions in the `common` section, aren't they independent from the operator?
2. Not right now, I was thinking how it could be done nicely. One idea I had was to scratch the current "wizard" approach and go with a "form" approach instead, where the terminal is split, and on one side you fill in the form and on the other a live preview of the generated config is displayed. This would be cool but IMO it's too much, so I'm not inclined to do it.

mirrord/up/src/init.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+fn prompt_http_filter(mode: &ServiceMode) -> Result<HttpFilterConfig, InitError> {
+    match mode {
+        ServiceMode::Split => {
+            let s = Text::new("HTTP header filter (regex; blank for auto session-key filter):")
+                .with_help_message("Example: `session-id: my-session-identifier`")
+                .prompt()?;
+            Ok(match s.trim() {
+                "" => HttpFilterConfig::default(),
```
cristeigabriela, line 207: personal nitpick: hf is hf.is_empty()

cristeigabriela, line 207: id avoid putting multi-line parsing into the return expression personally

mirrord/up/src/init.rs
```diff
@@ -0,0 +1,582 @@
...
+    ];
+
+    let choice = Select::new(
+        "Ports to ignore (sidecar ports, etc.):",
+        presets.iter().map(|(label, _)| *label).collect(),
+    )
+    .prompt()?;
+
+    if choice == "Custom..." {
+        let s = Text::new("Comma-separated ports:")
```
cristeigabriela, line 248: nitpick: not a fan of how the port parsing logic repeats here, id have an utility function that returns Result<Vec<u16>, _> and

in with_validator: map_err
in return: filter_map or unwrap

cristeigabriela (dismissed):
nitpicks! good stuff!

cristeigabriela (approved):
looks great!

PR #4271, opened 2026-05-15
Auto add local outgoing domains to dns filter

MintSoup:
### Quality Checklist:

  documentation)

See https://linear.app/metalbear/issue/COR-1350/auto-add-local-outgoing-domains-to-dns-filter

In addition to the behavior mentioned in the issue, the inverse case (i.e. remote filter ==> remote dns) is also handled.


mirrord/config/src/lib.rs
```diff
@@ -535,9 +537,75 @@ impl LayerConfig {
...
+                _ => None,
+            })
+            .collect();
+
+        if names.is_empty() {
+            return;
+        }
+
+        let dns_filter = &mut self.feature.network.dns.filter;
+        match dbg!((outgoing_is_local, dns_filter.take())) {
```
cristeigabriela: oh this is actually the first time i see this pattern, terrifying

would it make sene to consider the following? this feels a bit wild to read

> Note that the macro is intended as a debugging tool and therefore you should avoid having uses of it in version control for long periods (other than in tests and similar). Debug output from production code is better done with other facilities such as the [debug!](https://docs.rs/log/*/log/macro.debug.html) macro from the [log](https://crates.io/crates/log) crate.

mirrord/config/src/lib.rs
```diff
@@ -964,6 +1032,42 @@ impl LayerConfig {
     }
 }
 
+/// Append `additions` onto `existing`, skipping entries already present and preserving order.
+fn append_dedup(existing: VecOrSingle<String>, additions: Vec<String>) -> VecOrSingle<String> {
+    let mut out: Vec<String> = existing.into();
+    for item in additions {
+        if !out.iter().any(|e| e == &item) {
```
cristeigabriela: 'out.contains'

mirrord/config/src/lib.rs
```diff
@@ -964,6 +1032,42 @@ impl LayerConfig {
...
+            out.push(item);
+        }
+    }
+    VecOrSingle::Multiple(out)
```
cristeigabriela: if we have .into() it would be nice

mirrord/config/src/lib.rs
```diff
@@ -964,6 +1032,42 @@ impl LayerConfig {
     }
 }
 
+/// Append `additions` onto `existing`, skipping entries already present and preserving order.
+fn append_dedup(existing: VecOrSingle<String>, additions: Vec<String>) -> VecOrSingle<String> {
```
cristeigabriela, line 1036: if you wanna get silly and we have IntoIter just as well:

```rs
  existing.into_iter()
      .chain(additions)
      .fold(Vec::<String>::new(), |mut acc, s| {
          if !acc.contains(&s) { acc.push(s); }
          acc
      })
      .into()
```

mirrord/config/src/lib.rs
```diff
@@ -964,6 +1032,42 @@ impl LayerConfig {
...
+fn warn_filter_conflict(
+    context: &mut ConfigContext,
+    existing: &[String],
+    injected: &[String],
+    routed_side: &str,
+    dns_side: &str,
+) {
+    let conflicting: Vec<&str> = injected
+        .iter()
+        .filter(|n| existing.contains(n))
```
cristeigabriela, line 1061: this could be `filter_map(existing.contains(n).then(|| n.as_str())` but i think its purely persoal preference

cristeigabriela (changes requested):
great work! thank you for the documentation, it really helped grasp what was going on quickly!

feel free to ignore most of these, but id like to discuss the `dbg!`

mirrord/config/src/lib.rs
MintSoup: oh yikes this got left in accidentally

mirrord/config/src/lib.rs
MintSoup: cluppy dont like this

mirrord/config/src/lib.rs
iniw: Isn't this just a [`HashSet`](https://doc.rust-lang.org/std/collections/struct.HashSet.html)?

MintSoup: Yeah doing the same thing pretty much. Kept it like this mostly because makes tests slightly easier, can just compare against a vec literal.

iniw: You can do `HashSet::from([...])` in those cases, only slightly less readable IMO. But do whatever your heart desires :3

cristeigabriela (approved):

PR #4275, opened 2026-05-18
win-106 ≽^•⩊•^≼ async fs read

cristeigabriela:
```
                            ╱|、
                          (˚ˎ 。7  
                           |、˜〵          
                          じしˍ,)ノ
```

This pull-request supports cats.

## Linear

https://linear.app/metalbear/issue/WIN-106/readalltextasync-invalid-access-to-memory

---

## Problem summary and analysis

Previously, C# **async** reads (performed, for example, via `File.ReadAllTextAsync`) as illustrated below from a Slack user report:

```c#
while (true)
{
    try
    {
        string content = await File.ReadAllTextAsync("/etc/resolv.conf");
        Console.WriteLine(content);
    }
    catch (Exception ex)
    {
        Console.WriteLine($"Error ReadAllTextAsync /etc/resolv.conf: {ex.Message}");
    }
    await Task.Delay(TimeSpan.FromSeconds(10));
}
```

would fail with the following error:

> Error ReadAllTextAsync /etc/resolv.conf: Invalid access to memory location. (0x800703E6)

Well, that's... unintuitive, isn't it?

> **The parasite in my brain breaking through**: it was actually because of `NtSetInformationFile` on an unimplemented `FileInformationClass` (`FileCompletionInformation`)!

Shut up, parasite in my brain.

> **The parasite in my brain breaking through**: You can't stop me, you know? You read too many of fasterthanlime's blogs.

This was easy to deduce from the logs of running this operation. The last syscall that would run on the handle was that (or `NtClose`, I don't recall) and what would happen is that it would return STATUS_ACCESS_VIOLATION

> **The parasite in my brain breaking through**: But that's a NTSTATUS, and yours is a HRESULT!

Well, not so smart, are you, parasite that's been in my brain for 2 weeks? That's because C# does a 2-stage conversion: it runs `RtlNtStatusToDosError(STATUS_ACCESS_VIOLATION)` first (yielding `ERROR_NOACCESS`), and then it turns that into a `HRESULT` by running it through `HRESULT_FROM_WIN32`,

This is perfectly illustrated by deferring the scribe work to my most-trustworthy Magus Cláudio, the Thinking Man, who would never deceive or betray his Dame, Queen of Computer Realms.

<center>image</center>

GRAND!

But what is `FileCompletionInformation`...?

## Solution summary

Windows does native async by facillitating **IO completion ports**.

<center>Untitled-2026-02-16-2328</center>

> The image above above is a **SVG**, feel free to open it and zoom in.

> **The parasite in my brain breaking through**: this is my space, go away!

> shut up

The basic idea is that, `NtReadFile` is either done on a synchronous handle, or not. The operation is synchronous if synchronicity is opted into from `NtCreateFile`. Otherwise, it will be done through the completion information port.

> The IO completion port has to be **created** (at any time in the process, before binding, through **NtCreateIoCompletion**), and then set explicitly to the file handle (through **NtSetInformationFile** with **FileCompletionInformation**, or **FileReplaceCompletionInformation** afterwards).

> There may only be **one IO port associated with a file handle at any time**.

We will focus on the async scenario.

`NtReadFile` will immediately return `STATUS_PENDING` immediately**!**, and, when the operation has been fulfilled, it will post a completion (closest scalar analogue: `NtSetIoCompletion`, possible to post more through the `-Ex` variant) to the IO port.

> You may **peek**, non-destructively, on the port (which is a **FIFO** queue) using `NtQueryIoCompletion`. You can wait to read **N** `FILE_IO_COMPLETION_INFORMATION`s through `NtRemoveIoCompletion.
> **!** There is the "skip on success" scenario.

The number of bytes read (if any), and the NTSTATUS result of the operation for each packet, is expressed in the IO status block that we get from `NtRemoveIoCompletion(Ex)` packets (`IoCompletionInformation` type: **FILE_IO_COMPLETION_INFORMATION**)

Illustrated:

```cpp
    FILE_IO_COMPLETION_INFORMATION pkt = {};
    EXPECT_OK(nt::DequeueOne(port, &pkt)); // NtRemoveIoCompletion
    note("dequeued: key=%p ctx=%p status=0x%08lX bytes=%zu",
         pkt.KeyContext, pkt.ApcContext,
         (unsigned long)pkt.IoStatusBlock.Status,
         (size_t)pkt.IoStatusBlock.Information);
```

`NtRemoveIoCompletion` can either **poll** or **block** based on the Timeout provided.

> **NULL** => wait until you went through all the packets, **blocking**.
> **QuadPart = 0** => **poll**, `STATUS_TIMEOUT` if empty, can also be used as an actual timeout.

### Win32 question

This API hooks implementation is thorough enough to support Win32 `OVERLAPPED` only through it's NT hooks, as it should be if you dont suk.

### okay getting back

The solution basically implements:

- Support for sync/async handles;
- Support for sync/async reading (with all the caveats like not-really-async, etc);
- A worker thread pool that draws jobs, and posts a completion to the provided (real, Windows-handled) port in **FIFO order**;

## Reorganization in the scope of this PR

The current solution was messy, and was about to get even messier, so I and Claude gave the FS hooks a complete overhaul, split everything in < ~300 line files, and made some load-bearing operations split into definite paths that they will take (eg **read.rs**) to make reviewing each path, and it's conditions for the arm to be reached, easier.

This is a huge refactor, I'm aware. There's more details about it in the Claude implementation summary below.

---

### Claude's implementation summary:

<details>
Adds end-to-end support for async file reads issued through IO completion ports on Windows. Apps that open files with `FILE_FLAG_OVERLAPPED` (Win32) or omit `FILE_SYNCHRONOUS_IO_*` (Nt) and dequeue completion packets via `NtRemoveIoCompletionEx` / `GetQueuedCompletionStatus` -- Node.js, .NET, modern Python asyncio, Go runtime, IIS-style servers -- now have their reads mirrored against the remote pod, matching the sync read behavior that already exists.

## Layer architecture

* `mirrord/layer-win/src/managed/` -- generic `ManagedRegistry<K, V>` over `Lazy<RwLock<HashMap<K, Arc<RwLock<V>>>>>`. `K` is any newtype that implements the `ManagedHandleKey` trait (currently `MirrordFileHandle` for the file domain); the trait carries a `HandleKind` associated constant so a leaked or misused handle can be attributed by kind at a glance from a debugger. The shape is shared so a future IOCP-handle registry can drop in beside the file one without re-deriving the locking pattern.
* `mirrord/layer-win/src/iocp/` (a sibling of `managed`, not a hook):
  * `binding.rs` -- `RwLock<HashMap<FileKey, FileBinding{port, key}>>` indexed by file handle.
  * `packet.rs` -- `IocpPacket {key, apc_context, status, information}`
    + `ORIGINAL_NT_SET_IO_COMPLETION` resolved at init via `GetProcAddress` (no hook on `NtSetIoCompletion`); `enqueue_packet` posts via the captured pointer.
  * `worker.rs` -- a 4-thread pool consuming `Arc<Mutex<mpsc::Receiver<Job>>>` with panic catching.
  * `mod.rs::initialize()` resolves the original pointer; no port is ever managed by mirrord -- we record `(file, port, key)` on bind and post to the real OS port from worker threads.

## FS hooks reorg

`mirrord/layer-win/src/hooks/files/` is reorganized so the 1993-line `mod.rs` becomes a ~390-line dispatch table:

* `types.rs` -- the 17 `NtXxxType` aliases and `OnceLock` originals, previously spread across the top of `mod.rs`.
* `iosb.rs` -- `check_io_pointers` + `write_iosb` / `write_iosb_success` / `write_iosb_eof`, hiding the bindgen `ManuallyDrop<union>` ceremony and the duplicated pointer pre-flight that lived at ~10 sites.
* `ops/{open,read,close,set_info,query_info,cancel,wait,stubs}.rs` -- one module per FS operation; each contains the body that used to live inline in `mod.rs` plus a detailed step-based doc comment. `mod.rs` keeps only the `unsafe extern "system"` shims that delegate to `ops::<op>::handle` (or named entries in `query_info` / `stubs`).
* The "unimplemented" stubs in `ops/stubs.rs` keep their hand-written warn-if-managed bodies on purpose (the warns are load-bearing observability for prioritization, and hand-written bodies make ad-hoc instrumentation trivial).

## NT-level behavior

* `NtCreateFile`: validates the NT contract that `FILE_SYNCHRONOUS_IO_*` requires `SYNCHRONIZE` in `DesiredAccess` -- the layer no longer silently grants a handle the OS would have refused.
* `NtReadFile`: four code paths after pointer pre-flight -- `length == 0` short-circuit, async open-mode validation (rejects NULL `ByteOffset` and the `-2` `FILE_USE_FILE_POINTER_POSITION` sentinel on async handles), async-bound path (worker-driven, returns `STATUS_PENDING`; or inline when `FILE_SKIP_COMPLETION_PORT_ON_SUCCESS` is set; null `ByteOffset` rejected with `STATUS_INVALID_PARAMETER` as a defensive guard for a sync handle that somehow got port-bound), sync/inline path (save/seek/read/restore for explicit-offset reads so the stored position survives, per NT contract).
* `NtSetInformationFile`: handles `FileCompletionInformation`, `FileReplaceCompletionInformation` (Win 8.1+), and `FileIoCompletionNotificationInformation`. Catch-all returns `STATUS_SUCCESS` instead of the former `STATUS_ACCESS_VIOLATION` that was breaking .NET's `FileStream` init with `IOException: Invalid access to memory location`.
* `NtQueryInformationFile`: same shape; catch-all returns `STATUS_INVALID_INFO_CLASS` instead of `STATUS_ACCESS_VIOLATION` for the same reason.
* `NtCancelIoFile` (new hook): preserves the "one IO -> exactly one packet" invariant for managed files; acknowledges with `STATUS_SUCCESS` since the layer has no real IRPs to cancel.
* `NtWaitForSingleObject` (new hook): returns `STATUS_SUCCESS` immediately on synthetic `0x5000_xxxx` file handles -- the data is already in the user's buffer by the time the caller waits.
* `NtClose`: extended to drop any IOCP binding before removing the managed file entry.
</details>

### Quality Checklist:

  documentation)


changelog.d/+iocp-windows-fs.added.md
```diff
@@ -0,0 +1 @@
+Windows: added support for asynchronous file reads (fixes `File.ReadAllTextAsync` etc).
```
meowjesty, line 1: Do we have the technology for adding 2 changelog entries in a single PR? Would be nice to have an internal one saying that this refactored file modules.

tests/cs-e2e/AsyncText/.gitignore
```diff
@@ -0,0 +1,3 @@
+bin/
```
meowjesty: Do we even need this `.gitignore` file?

tests/cs-e2e/AsyncText/AsyncText.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
meowjesty: Do we still need `csproj` for c# apps?

https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps

Can't we just `dotnet build Program.cs` to simplify this whole thing?

tests/cs-e2e/AsyncText/build_test_app.ps1
```diff
@@ -0,0 +1,28 @@
+# Compile the AsyncText.exe e2e test app.
```
meowjesty: Is this how we were building every c# app? I think we could refactor them to be file based apps (no csproj needed), rename each app from `Program.cs` to what they're testing (like in this case `AsyncText.cs`), and then have just 1 ps script that goes into the folder calling `dotnet build` on every app that is in the folder.

For a future PR btw, this PR should just simplify the test that it added.

mirrord/layer-win/src/hooks/files/ops/stubs.rs
```diff
@@ -0,0 +1,318 @@
...
+    output_buffer: PVOID,
+    output_buffer_length: ULONG,
+) -> NTSTATUS {
+    unsafe {
+        if let Ok(handles) = MANAGED_HANDLES.try_read()
+            && let Some(managed_handle) = handles.get(&file)
+            && let Ok(_handle_context) = managed_handle.clone().try_read()
+        {
+            // unsafe to print in
+        }
```
meowjesty, line 237: No need to do this, perform the checks you need and move on, no need for empty blocks for future debugging logs.

meowjesty (changes requested):
Implementation and refactor looks good to me (refactor from skimming it).

I think we can simplify the way the C# tests are being built.

P.S. I've been guilty of this too, but try to separate the refactor from the implementation to make it easier for review. Next time you need to refactor a bunch of stuff, make a PR just for that, and another for the feature on top of it.

mirrord/layer-win/src/hooks/files/ops/stubs.rs
cristeigabriela: the empty block allows for placing breakpoints when the handle is in the handle context

meowjesty: Yeah I got it, but like, you can just add it back again when needed, otherwise we could have this pattern in basically every piece of code.

cristeigabriela: will do.

the reason thats not been picked first is because layer builds take long and it makes debugging terrible.

meowjesty (approved):
👍

PR #4276, opened 2026-05-18
Reject duplicate incoming port mappings

immanuwell:
## What
Reject duplicate ports in incoming `port_mapping` and `listen_ports` before they get collected into a `BiMap`. Tiny config footgun, imo: before this, one entry could just vanish and it was kinda hard to spot.

## Repro
Use a config like:

```json
{ "feature": { "network": { "incoming": { "port_mapping": [[8080, 80], [8080, 81]] } } } }
```

Before: config resolution accepted it, then `BiMap` kept one mapping for `8080` and dropped the other one. same deal for duplicate target ports and `listen_ports`.

Now: config generation errors early with a clear duplicate-port message. no silent yikes.

## Checks
- `cargo fmt`
- `cargo test -p mirrord-config rejects_duplicate_ --lib`
- `cargo test -p mirrord-config --lib`
- `cargo check -p mirrord-config --keep-going`
- `cargo clippy -p mirrord-config -- --deny warnings`
- `git diff --check`


mirrord/config/src/feature/network/incoming.rs
```diff
@@ -142,6 +144,28 @@ impl MirrordConfig for IncomingFileConfig {
...
+
+        if !right_ports.insert(*right) {
+            return Err(ConfigError::Conflict(format!(
+                "`{field}` contains duplicate target port `{right}`"
+            )));
+        }
+    }
+
+    Ok(mappings.into_iter().collect())
+}
```
0x00A5, line 168: nit: we can avoid the sets by doing something like the following

```suggestion
fn bi_map_from_config(mappings: Vec<(u16, u16)>, field: &'static str) -> Result<BiMap<u16, u16>> {
    let mut map = BiMap::new();

    for (left, right) in mappings {
        if map.contains_left(&left) {
            return Err(ConfigError::Conflict(format!(
                "`{field}` contains duplicate source port `{left}`"
            )));
        }

        if map.contains_right(&right) {
            return Err(ConfigError::Conflict(format!(
                "`{field}` contains duplicate target port `{right}`"
            )));
        }

        map.insert(left, right);
    }

    Ok(map)
}
```

0x00A5 (changes requested):
Thank you very much for contributing to mirrord ❤️ . I have requested a small change, otherwise looks great 👍

immanuwell:
@0x00A5 done, I applied your suggestions, PTAL

0x00A5 (approved):

0x00A5:
Thank you for the contribution!

PR #4280, opened 2026-05-19
COR-1458 😼 unescaped regex file filters crash

cristeigabriela:
## Linear
https://linear.app/metalbear/issue/COR-1458/investigate-git-bash-packagejson-execution-error-on-windows

## Summary
Previous code had 2 issues:
```rs
            if let Ok(home) = std::env::var("HOME") {
                let tmpdir = std::env::var("TMPDIR").unwrap_or_else(|_| "/tmp".to_owned());
                let pattern = format!("^{home}/\\.aws(/.*)?");
                // ...
```

1. It used `HOME` on each platform (Windows has `USERPROFILE` as an equivalent);
2. The `home` variable was not Regex-escaped.

But... wait, if Windows doesn't set `HOME`, why was this caught?

Because one of our client uses Git Bash (MSYS), which sets HOME!!!
image

(It uses backslashes, and combined with the unescaped regex clause, we got an invalid regex which caused layer to panic!)

PowerShell, for comparision (proof that it's not the runtime):
image

## Future consideration
Lets please be considerate and begin making sure our regexes are good friends and don't panic layer!

## Claude reaction

image

### Quality Checklist:

  documentation)


tests/src/file_ops.rs
```diff
@@ -353,4 +353,61 @@ mod file_ops_tests {
         };
         assert_eq!(statfs_from_mirrord, statfs_from_pod);
     }
+
```
aviramha: It shouldn't be an e2e test to check it. It can be tested in an integration test.

aviramha (changes requested):

aviramha (approved):

PR #4286, opened 2026-05-21
Updated test guide on CONTRIBUTING.md

giIuis:
* e2e and integration tests now build binary and layer automatically when no args or envvars are presented
* Fix CONTRIBUTING.md to accuratelly describe the steps needed for integration and e2e testing
* Fix scripts/build_go_apps.sh to now always runs to completion even if linux only go apps fail on mac.
Previously, the contributor had to add that by hand

### Quality Checklist:

  documentation)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,11 +29,56 @@ The following guide details the steps to setup a local development environment f
 - [GCC](https://gcc.gnu.org/) - only on Linux, GCC is needed for Go dynamic linking
 - [Rust](https://www.rust-lang.org/)
 - [NodeJS](https://nodejs.org/en/), [ExpressJS](https://expressjs.com/), [portfinder](https://www.npmjs.com/package/portfinder)
-- [Python](https://www.python.org/), [Flask](https://flask.palletsprojects.com/en/2.1.x/), [FastAPI](https://fastapi.tiangolo.com/), [Uvicorn](https://www.uvicorn.org/)
+- [Python (recommended 3.13+)](https://www.python.org/), [Flask](https://flask.palletsprojects.com/en/2.1.x/), [FastAPI](https://fastapi.tiangolo.com/), [Uvicorn](https://www.uvicorn.org/)
 - [Go](https://go.dev/)
 - Kubernetes Cluster (local/remote)
 - [Argo Rollouts CRD](https://argoproj.github.io/argo-rollouts/) - required for rollout-related E2E tests (`kubectl create namespace argo-rollouts && kubectl apply -n argo-rollouts -f https://github.com/argoproj/argo-rollouts/releases/latest/download/install.yaml`)
 
+### Building mirrord layer and binary
+
+If you run integration or e2e tests, both are automatically built.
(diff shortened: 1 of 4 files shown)


changelog.d/+improve-contributing-guide.internal.md
t4lz, line 1: Internal :)

CONTRIBUTING.md
```diff
@@ -160,7 +192,10 @@ IPv6 tests (they currently don't run in the CI):
 The Kubernetes resources created by the E2E tests are automatically deleted when the test exits. However, you can preserve resources from failed tests for debugging. To do this, set the `MIRRORD_E2E_PRESERVE_FAILED` variable to any value.
 
 ```bash
-MIRRORD_E2E_PRESERVE_FAILED=y cargo xtask test-e2e
+MIRRORD_TESTS_USE_BINARY=$(pwd)/target/<arch>/debug/mirrord \
+MIRRORD_LAYER_FILE=$(pwd)/target/<arch>/debug/libmirrord_layer.dylib \
+MIRRORD_E2E_PRESERVE_FAILED=y
+cargo xtask test-e2e
```
hank-metalbear, line 214: should this line be connected with above line?

scripts/build_go_apps.sh
```diff
@@ -23,8 +23,8 @@ do
     cd $directory
 
     1>&2 echo "Building test app $directory/$1.go_test_app"
-    go build -o "$1.go_test_app"
-
+    # This `||` is necessary because //go:build linux prevents certain Go test apps from being built on macOS
```
aviramha: but this also means we'll silently fail to build. perhaps it'd be better to avoid those specific tests on macos also in the build script or find a way that wouldn't make us have silent errors.

giIuis: Oof, that's a big problem. Will investigate a better solution

scripts/build_go_apps.sh
```diff
@@ -22,9 +24,21 @@ do
...
-    1>&2 echo "Building test app $directory/$1.go_test_app"
-    go build -o "$1.go_test_app"
-
+    package_info=$(go list -e -f '{{len .GoFiles}} {{len .CgoFiles}} {{len .IgnoredGoFiles}}' .)
+    set -- $package_info
+    go_files=$1
+    cgo_files=$2
+    ignored_go_files=$3
+
+    if [ "$go_files" -eq 0 ] && [ "$cgo_files" -eq 0 ] && [ "$ignored_go_files" -gt 0 ]
```
giIuis, line 33: @aviramha check this again
how it works: the `go list` commands just returns the number of buildable go / cgo files (compiler diferentiate), as well as the amount of build-time-ignored go files (like //go build linux is ignored on mac). 
A directory is only skipped if there isn't any buildable go/cgo files, and at least one ignored file. This prevents failure on empty directories.

scripts/build_go_apps.sh
giIuis: Check newer diff

aviramha (approved):
🎉

PR #4289, opened 2026-05-21
Add support for outgoing sockets `SOCK_SEQPACKET`.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1401/support-seqstream-type-for-unix-sockets

We support unix sockets, but were missing this (very uncommon) type of unix socket.

Adding support for it seems pretty straightforward, it's mostly a copy paste job from the other outgoing socket stuff we have, except that this socket type is not in `tokio`, so it needs a separate crate [`tokio-seqpacket`](https://docs.rs/tokio-seqpacket/latest/tokio_seqpacket/index.html), and it doesn't have some of the `AsyncWrite`/`AsyncRead` stuff, so it needs to be sort of manually implemented (regular `send` seems to do the trick for write, and for reading we're using `poll_recv`).

### Quality Checklist:

  documentation)


mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+/// Background task that handles [`LayerSeqpacket`] messages from the agent, running in a specific
+/// background runtime.
+struct SeqpacketTask {
+    next_connection_id: ConnectionId,
+    /// Writing handles of peer connections made on layer's requests.
+    writers: HashMap<ConnectionId, Arc<UnixSeqpacket>>,
+    /// Reading handles of peer connections made on layer's requests.
+    readers: StreamMap<ConnectionId, SeqpacketReadStream>,
```
iniw, line 101: Can this be a single `HashMap` instead? They are always operated together (removal, the count on `Drop`, etc.)

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+impl SeqpacketTask {
+    /// Buffer size for reading from the outgoing connections.
+    const READ_BUFFER_SIZE: usize = 64 * 1024;
+    /// How much incoming data we can accumulate in memory, before it's flushed to the client.
+    ///
+    /// This **must** be larger than [`Self::READ_BUFFER_SIZE`].
+    const THROTTLE_PERMITS: usize = 512 * 1024;
```
iniw: If it must be larger can it be calculated using `READ_BUFFER_SIZE` then?
```suggestion
    /// This **must** be larger than [`Self::READ_BUFFER_SIZE`].
    const THROTTLE_PERMITS: usize = Self::READ_BUFFER_SIZE * 8;
```

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+            throttler: Arc::new(Semaphore::new(Self::THROTTLE_PERMITS)),
+        }
+    }
+
+    /// Runs this task as long as the channels connecting it with the [`SeqpacketApi`] are open.
+    #[tracing::instrument(level = Level::TRACE, skip(self))]
+    async fn run(mut self) {
+        loop {
+            let channel_closed = select! {
+                biased;
```
iniw, line 159: Why does it have to be biased?

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+
+                message = self.layer_rx.recv() => match message {
+                    Some(message) => self.handle_layer_msg(message).await.is_err(),
+                    None => true,
+                },
+
+                // Polls the `SeqpacketReadStream` for incoming data.
```
iniw: Pretty useless comment

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+                SEQPACKET_CONNECTION.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
+
+                self.daemon_tx
+                    .send(
+                        DaemonMessage::LogMessage(LogMessage::warn(format!(
+                            "read from outgoing seqpacket connection {connection_id} failed: {error}"
+                        )))
+                        .into(),
+                    )
+                    .await?;
```
iniw, line 226: AFAICT this is this `LogMessage` here and the one on `handle_layer_message` are the only reason why `daemon_tx` can't be a `Sender<Throttled<DaemonSeqpacket>>` instead of `Sender<Throttled<DaemonMessage>>`. 

Limiting this task to produce only `DaemonSeqpacket`s will make it feel more encapsulated, like `UdpOutgoingApi`.

But these logs do seem useful for debugging so feel free to ignore this comment if you think having them is more important than some arbitrary Clean Code™ stuff.

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+                        DaemonMessage::SeqpacketOutgoing(DaemonSeqpacket::Close(connection_id))
+                            .into(),
+                    )
+                    .await?;
+            }
+            Ok(None) => {
+                tracing::trace!(connection_id, "Seqpacket peer connection closed.");
+
+                self.readers.remove(&connection_id);
+                self.writers.remove(&connection_id);
```
iniw: This is what I mean with these two essentially being a single hashmap btw

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+        match remote_address {
+            SocketAddress::Unix(UnixAddr::Pathname(path)) => {
+                if let Some(pid) = target_pid {
+                    Ok(InTargetPathResolver::new(pid).resolve(&path)?)
+                } else {
+                    Ok(path)
+                }
+            }
+            SocketAddress::Unix(UnixAddr::Abstract(mut name)) => {
+                name.insert(0, 0);
```
iniw, line 281: Is this correct? Is the idea here to add a null terminator? If so shouldn't this be more like
```suggestion
                name.push(0);
```

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+                        }),
+                    None => Err(ResponseError::NotFound(connection_id)),
+                };
+
+                match write_result {
+                    Ok(()) => Ok(()),
+                    Err(error) => {
+                        self.writers.remove(&connection_id);
+                        self.readers.remove(&connection_id);
+                        SEQPACKET_CONNECTION.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
```
iniw: These 3 lines are repeated quite a few times (5, from my count). Could maybe be a `dispose_connection()` function or something?

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+        status::{BgTaskStatus, IntoStatus},
+    },
+    util::path_resolver::InTargetPathResolver,
+};
+
+/// The agent uses this api to communicate with the [`SeqpacketTask`].
+///
+/// Handles [`LayerSeqpacket`] messages sending them to the [`SeqpacketTask`] with
+/// [`Self::send_to_task`], and receiving [`DaemonSeqpacket`] with [`Self::recv_from_task`].
+pub(crate) struct SeqpacketApi {
```
iniw, line 45: Are all of these `pub(crate)` specifiers necessary? The module itself is already declared as `pub(crate)` in `mirrord/agent/src/outgoing.rs` so these can just be `pub`, no?

mirrord/agent/src/outgoing/seqpacket.rs
```diff
@@ -0,0 +1,455 @@
...
+                        .await
+                        .map_err(ResponseError::from)
+                        .and_then(|written| {
+                            if written == bytes.len() {
+                                Ok(())
+                            } else {
+                                Err(ResponseError::from(io::Error::new(
+                                    io::ErrorKind::WriteZero,
+                                    "partial seqpacket write",
+                                )))
```
iniw, line 362: Should this be a hard error? Is that how we handle partial writes in other outgoing connections? Why can't we just return the number of bytes written to the client and let them deal with it 🤔

mirrord/layer-lib/src/socket/sockets.rs
```diff
@@ -108,11 +108,30 @@ pub static SOCKETS: LazyLock<Mutex<HashMap<SocketDescriptor, Arc<UserSocket>>>>
             .unwrap_or_default()
     });
 
-// Helper function to convert socket types to SocketKind
+// Helper function to convert socket types to SocketKind (unix only)
+#[cfg(unix)]
 pub fn socket_kind_from_type(socket_type: i32) -> Result<SocketKind, String> {
-    if (socket_type & SOCK_STREAM) == SOCK_STREAM {
+    let raw_socket_type = socket_type & 0xf;
```
iniw: 🚨 magic number alert 🚨

mirrord/layer-lib/src/socket.rs
```diff
@@ -146,24 +149,47 @@ impl SocketKind {
...
     }
 }
 
 impl TryFrom<c_int> for SocketKind {
     type Error = Bypass;
 
+    #[cfg(unix)]
     fn try_from(type_: c_int) -> Result<Self, Self::Error> {
-        if (type_ & SOCK_STREAM) > 0 {
+        let socket_type = type_ & 0xf;
```
iniw, line 175: 🚨 magic number alert 🚨

.gitignore
```diff
@@ -11,6 +11,7 @@ __pycache__
 build/
 .mirrord
 .helix/
+.zed/
```
iniw, line 14: I hate this so much omg please just add it to your local global gitignore instead of to every metalbear project

iniw (changes requested):
Mostly just questions and some suggestions :)

mirrord/agent/src/outgoing/seqpacket.rs
iniw: Seems that we do treat partial writes as errors everywhere else... ignore dis den

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: It definitely can, but we have this pattern in every outgoing task.

I prefer following the pattern (in the past the other ones were a single `HashMap` too, from what I can remember?). Can I keep it or do you want me to change it?

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: I was waiting for someone to complain about this, so I could change the other places where we do exactly this.

541d8889ab48ce9dd36e35a4090cb419bf7ebbe4

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: You want me to remember something from like, 4 years ago. I have a vague memory of some problem when this is not `biased`, I think we want `recv` first to avoid this problem.

Now, what the problem was? Don't remember.
Was there even a problem in the first place, or is this some imaginary memory? Don't remember.

To avoid a potential regression, ctrl-c + ctrl-v.

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: 0c2c137103d58271f81c034e889cf5ee402054ea

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: I'm also with you on this, so I flipped a coin.

- Heads: keep it as-is;
- Tails: change to smaller type;

image

Flipped only once, destiny has spoken.

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: https://github.com/metalbear-co/mirrord/blob/116e83db99b03369fe47feccf6d6f98b03a5ef5a/mirrord/agent/src/outgoing/socket_stream.rs#L80-L85

Lifted from here, it works as-is (I tested it with a sample), so I assume this is correct.

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: da513fbc1d95089307c7f34f90dbce03f8c9d4d4

Changed!

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: Sure, but I like being explicit for things like this, can I keep it?

mirrord/agent/src/outgoing/seqpacket.rs
meowjesty: This might become annoying for seqpacket specifically, but we'll get there when we get there (someone, for some unknown reason is using this thing AND they need partial stuff, I think there's some special case for EOF something too, not 100% sure).

.gitignore
meowjesty: I don't have a local global gitignore, I'm guessing I'm the 1st zed-ish user to need project specific settings?

meowjesty: I'll ping you on every PR I add editor files from now on.

iniw: > I don't have a local global gitignore

Everyone does! You just add stuff to `~/.config/git/ignore`

mirrord/agent/src/outgoing/seqpacket.rs
iniw: What the hell...

mirrord/agent/src/outgoing/seqpacket.rs
iniw: Freaking destiny man

mirrord/agent/src/outgoing/seqpacket.rs
iniw: Fair

mirrord/agent/src/outgoing/seqpacket.rs
iniw: > but we have this pattern in every outgoing task

What happened to ask for forgiveness not for permission?

Time to let the coin decide I guess.

mirrord/agent/src/outgoing/seqpacket.rs
iniw: I think it's just noisy but sure

mirrord/agent/src/outgoing/seqpacket.rs
iniw: Special cased EOF sounds scary 👀
Hopefully we're behaving properly in terms of the semantics of this thing, but I'm sure someone will complain if we aren't 😅

mirrord/layer-lib/src/socket/sockets.rs
meowjesty: 52d8ca39ef352cd77f34ad73fb6c9e7cf587c1af

mirrord/layer-lib/src/socket.rs
meowjesty: 52d8ca39ef352cd77f34ad73fb6c9e7cf587c1af

.gitignore
meowjesty: Sorry, what's git? I use jj.

iniw (approved):

PR #4264, opened 2026-05-12
Make preview environments resilient to crashes and evictions

iniw:
## Summary

See the operator PR :)

## Issues

- [Operator should monitor preview pod health and reschedule within TTL window](https://linear.app/metalbear/issue/COR-1400/operator-should-monitor-preview-pod-health-and-reschedule-within-ttl)
- [Show only live previews in mirrord preview status](https://linear.app/metalbear/issue/COR-1388/show-only-live-previews-in-mirrord-preview-status)

## Related PRs

- https://github.com/metalbear-co/operator/pull/1527

### Quality Checklist:

- [x] I have documented new code sufficiently
- [x] I have checked and updated the relevant existing docs in code, including removing outdated material
- [x] I have written user-facing website docs for new features, or opened a (sub)issue to do so
- [x] I have checked and updated existing website docs for changed features
- [x] I have tested this change and know it succeeds and fails as expected
- [x] I have written unit tests or purposefully omitted them
- [x] I have written e2e tests or purposefully omitted them
- [x] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs,
  documentation)
- [x] I have introduced a short, clear and well-formatted changelog entry

PR #4213, opened 2026-04-24
WIN-108 - Fix DNS error reporting in getaddrinfo hooks and unify Windows detour return handling

itsamegraf:
This PR cleans up Windows detour return handling and fixes DNS error propagation for getaddrinfo hooks.
### Quality Checklist:

- [x] I have documented new code sufficiently
- [x] I have checked and updated the relevant existing docs in code, including removing outdated material
- [ ] I have written user-facing website docs for new features, or opened a (sub)issue to do so
- [ ] I have checked and updated existing website docs for changed features
- [x] I have tested this change and know it succeeds and fails as expected
- [ ] I have written unit tests or purposefully omitted them
- [x] I have written e2e tests or purposefully omitted them
- [x] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs, documentation)
- [x] I have introduced a short, clear and well-formatted changelog entry

### Summary

This PR fixes Windows DNS error reporting for `getaddrinfo`-based flows and aligns Windows detour return/error handling across socket hooks.

### Why

Previously, failing DNS resolution on Windows could surface unrelated errors like `ERROR_INVALID_HANDLE (os error 6)` instead of expected WinSock DNS errors.  
This PR ensures detours return values and error channels according to each Windows API contract, so DNS failures propagate correctly (e.g. `WSAHOST_NOT_FOUND` / 11001).

### What changed

- Added Windows-specific detour return contract in `layer-lib`:
  - Introduced `WindowsDetourReturn` in `mirrord/layer-lib/src/detour.rs`.
  - Added `Detour::unwrap_or_bypass_windows_as` to encode API-specific Windows return behavior.
  - Added marker implementations (`WinSocket`, `WinGetAddrInfoInt`) with shared macro-based impl generation.

- Updated Windows socket hooks to use the new API-specific return path:
  - `socket`, `WSASocketA`, `WSASocketW` now use `unwrap_or_bypass_windows_as::<WinSocket, _>`.
  - `getaddrinfo_detour` and `getaddrinfow_detour` now use `unwrap_or_bypass_windows_as::<WinGetAddrInfoInt, _>`.
  - Moved managed addrinfo storage into the success `map` path for cleaner flow.

- Tests/CI updates:
  - Enabled `dns_resolve` integration test on Windows (`tests/dns_resolve.rs` no longer unix-only).
  - Updated dns_resolve app expectation to Windows-specific message (`No such host is known. (os error 11001)`).
[...]

PR #4302, opened 2026-05-23
Fix Windows mirror/steal HTTP forwarding to wildcard-bound servers

cristeigabriela:
## What

The intproxy HTTP gateway connected to the *resolved* `listening_on` address directly. #4264 (`3817ef2c`) dropped the `normalize_connection_address` call on this path while the raw-TCP path kept it. Connecting to a wildcard address (`0.0.0.0`) routes to loopback on Linux but **fails on Windows**, so every mirrored/stolen HTTP request to a wildcard-bound server failed there.

Rather than just re-add the call on the HTTP path (leaving the two paths free to drift again), normalization is folded into `ListeningOnExt::resolve_addr`: the resolver now always returns a *connectable* address, normalizing a wildcard listen address (`0.0.0.0` / `::`) to loopback. Both callers — the raw-TCP path in `IncomingProxy` and the HTTP gateway — drop their manual `normalize_connection_address` calls, so the two forwarding paths share one source of truth and can't diverge again.

## How this was found

windows-e2e last passed at #4213 and was failing on `main`. Bisecting the 46 commits in between (CI probes at evenly-spaced commits) isolated the regression to **#4264** — id 44 passed, id 45 (#4264) failed. The failing tests are all incoming HTTP (`http::mirror_http_traffic`, `traffic::mirror::mirror_with_http_header_filter::*`, `traffic::steal::steal_tests::*`), matching exactly the gateway forwarding path.

## Notes

- `windows-e2e` label so the Windows e2e suite runs.
- Normalization is now centralized in `ListeningOnExt::resolve_addr` (the follow-up previously noted here is done in this PR), so the TCP and HTTP forwarding paths can't regress independently.

### DIFF
--- a/.vale/styles/config/vocabularies/Metalbear/accept.txt
+++ b/.vale/styles/config/vocabularies/Metalbear/accept.txt
@@ -125,6 +125,7 @@ TTL
 RTT
 VPN
 URI
+loopback
 
 # File formats
 YAML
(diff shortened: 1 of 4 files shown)


iniw (dismissed):
sowwy

iniw (approved):

PR #4320, opened 2026-06-01
Introduce KubeClient

dmzmk:
Bundle kube Client and Config into a KubeClient struct so ResourceGuard::Drop can create a fresh client from the original config rather than using Config::infer()

### Quality Checklist:

  documentation)


tests/src/utils/client.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// External consumers that only need a plain [`Client`] can still import the top-level
+/// `kube_client` from `crate::utils` (or `mirrord_tests::utils`), which wraps this fixture and
+/// returns just the [`Client`].
+#[fixture]
+pub async fn kube_client() -> KubeClient {
+    CRYPTO_PROVIDER_INSTALLED.call_once(|| {
+        if rustls::crypto::CryptoProvider::get_default().is_none() {
+            let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+        }
```
iniw: No need to check the existing provider. If one already exist the function will fail and we ignore the error, so this is equivalent to
```suggestion
        let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```

aviramha (approved):

PR #4339, opened 2026-06-04
preview: hash env keys before placing it in PreviewSession's labels

iniw:
The session key is currently stored in a label in every `PreviewSession` resource to perform server-side filtering through label selectors when listing preview sessions, which happens in `preview start` (to detect conflicting preview sessions) and `preview status` (to show them, duh).

This caused a problem when the session key didn't match [kubernetes label rules](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set), which can happen pretty easily. For example the `/` character is invalid, which is pretty common in branch names.

<details>
<summary>See notes about the previous iteration of this PR</summary>

Initially I thought about fixing this by escaping the key, which could be done when parsing it and storing in `LayerConfig` or when placing/retrieving it from the label, but I thought this was bad and hacky, so I decided instead to simply _not_ inserting the label and using it as a selector.

This does mean we lose server-side filtering, yes, but I expect this to be a complete non-issue in practice since we don't have many `PreviewSession` resources, especially since expired/failed sessions are now automatically deleted, so listing them all and filtering locally should never be noticeably slower than filtering on the server.

</details>

To fix this in a backwards compatible way while still performing server-side filtering I decided to hash the key before placing it in the label, which requires changing the label selector to use a [set-based requirement](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering) to check against either the raw key (for older sessions) or the hashed key.

### Quality Checklist:

  documentation)


changelog.d/+preview-key-label.fixed.md
```diff
@@ -0,0 +1 @@
+Preview session keys are no longer stored as a Kubernetes label on the `PreviewSession` resource, so keys containing characters that aren't valid in a label value can now be used.
```
meowjesty: You could use annotations instead.

CLAUDE.md
```diff
@@ -99,10 +99,3 @@ behavior directly.
 - Prefer `to_owned` for `&str` → `String`.
 - Always use `foo.rs` instead of `foo/mod.rs` for module roots.
 - Run `cargo fmt` and `cargo clippy -- --deny warnings` after every edit.
-
-## Contribution Notes
```
meowjesty, line 103: Random changes lol.

(I hate when the ai auto-adds a changelog, so I welcome this).

meowjesty (dismissed):
👍

changelog.d/+preview-key-label.fixed.md
iniw: We can't because annotations can't be used as a selector when listing resources :(

CLAUDE.md
iniw: Reverted 😅 Will ask if it's okay for me to remove zis

meowjesty (approved):
👍

PR #4345, opened 2026-06-08
>> COR-1542  ʕ•ᴥ•ʔ  GetComputerName* hooks cleanup [Redis]

cristeigabriela:
## Linear
https://linear.app/metalbear/issue/COR-1542/windows-redis-app-fails-related-to-getcomputername

## Summary

On Windows, an app run through mirrord could fail to use the remote computer name. The most visible
symptom was a `StackExchange.Redis.RedisConnectionException` ("It was not possible to connect to the
redis server(s). Error connecting right now.") thrown during DI resolution, even though the Redis
tunnel itself worked.

Root cause: the `GetComputerName*` hooks returned the **full remote pod hostname** (often >15 chars,
e.g. `payment-api-9b7855b77-vlt4r`) and didn't respect the caller's buffer the way Windows does. .NET's
`Environment.MachineName` calls `GetComputerNameW` with a fixed `MAX_COMPUTERNAME_LENGTH + 1` (16)
wchar buffer and **throws** on a `FALSE` return; StackExchange.Redis reads `Environment.MachineName`
while establishing a connection, so the throw aborted the handshake (`InternalFailure`, `0-read,
0-sent`) and surfaced as a connection failure.

This PR makes the hooks behave like the real Win32 APIs.

> ⚠️ **A subtlety worth flagging.** Hooking `GetComputerNameExW` in **kernelbase** (so we catch every
> caller, not just the kernel32 ones) means SChannel/SSPI receives the computer name when it queries
> the **NetBIOS** format during a TLS handshake (`AcquireCredentialsHandle`). A long pod name there
> fails with `SEC_E_SECPKG_NOT_FOUND` and would break **all outgoing TLS** under mirrord — every
> HTTPS/gRPC call (e.g. to `*.googleapis.com`), while plaintext (Redis, Postgres, an in-cluster gRPC
> emulator) keeps working. That is exactly why `GetComputerNameExW` **NetBIOS** formats truncate to
> fit (`ex_w_policy`); **DNS** formats keep the size-probe contract.

## Root cause (verified, not assumed)

- The outgoing Redis tunnel is fine: a raw `TcpClient` `PING` to the in-cluster Redis round-trips
  `+PONG` through mirrord.
- `Environment.MachineName` throws **only under mirrord** (`GetComputerNameW` returns `FALSE` /
  `ERROR_BUFFER_OVERFLOW` because the 27-char pod name doesn't fit the 16-wchar buffer).
- Causation confirmed: setting `ConfigurationOptions.ClientName` explicitly (so StackExchange.Redis
  skips `Environment.MachineName`) makes it connect cleanly through mirrord.

## What changed

<details><summary>What changed, file by file</summary>

`mirrord/layer-win/src/hooks/socket/`:

- **Hook the right exports.** Hook `GetComputerNameW` (exported from **kernel32** — it reads the
  computer name from the registry directly, with an early `_CLUSTER_NETWORK_NAME_` env bail, and does
  **not** funnel through `GetComputerNameExW`) and `GetComputerNameExW` (in **kernelbase**). Drop the
  separate `GetComputerNameA` / `GetComputerNameExA` hooks — they forward to their W counterparts.
- **Respect the caller's buffer.** The shared hostname writers now write what fits the caller's buffer
  instead of imposing a fixed cap:
  - `GetComputerNameW` (`NameTooLong::Truncate`): truncates to fit and returns success. This is the
    path .NET takes; it can't grow its fixed buffer, and a real Windows computer name always fits 16.
  - `gethostname` and `GetComputerNameExW` **DNS** formats (`NameTooLong::Fail`): keep the standard
    size-probe contract — a too-small **or zero-length** buffer reports the required size +
    `ERROR_MORE_DATA` (`WSAEFAULT` for `gethostname`), so *query-size → grow → retry* works.
  - `GetComputerNameExW` **NetBIOS** formats (`NameTooLong::Truncate`, selected by `ex_w_policy`):
    truncate to fit. A NetBIOS name is ≤15 on Windows and callers query it with a fixed buffer they
    can't grow — notably **SChannel/SSPI during a TLS handshake** — so truncating rather than failing
    is what keeps outgoing TLS working (see the regression note above).
- **No buffer-size cap.** Removed `validate_buffer_params` (which rejected buffers `>256` and `==0`
  with `ERROR_INVALID_PARAMETER`). Windows caps neither — only the *name* is bounded, and we only ever
  write the short name — so large / `NI_MAXHOST` (1025) buffers now succeed and get the remote name.
- **Fixed a latent off-by-one.** The old `len > buffer_size` check wrote the null terminator one past
  the end when `len == buffer_size`; the new logic always leaves room for the null.
- NetBIOS trimming (`trim_netbios_name`) is retained only for `GetAddrInfoExW`'s `NS_NETBT` path.

</details>

## Behaviour

| Caller pattern | Before | After |
|---|---|---|
| `GetComputerNameW` / `Environment.MachineName`, fixed 16-buf | `FALSE` → .NET throws | ✅ success, name truncated to fit |
| `GetComputerNameW`, roomy buffer | `FALSE` (too-small) | ✅ success, full remote name |
| `GetComputerNameExW` size-probe (`NULL`/`*nSize == 0`) | `ERROR_INVALID_PARAMETER` | ✅ `ERROR_MORE_DATA` + required size |
| `GetComputerNameExW` / `gethostname`, buffer too small | (varied) | ✅ `ERROR_MORE_DATA` / `WSAEFAULT` + required size (retryable) |
| Large / `NI_MAXHOST` buffer | `ERROR_INVALID_PARAMETER` (>256 cap) | ✅ success, full remote name |
| `len == buffer_size` | one-past-end null write | ✅ truncates, no OOB |

## Testing

**Unit tests** (`hooks::socket::hostname::tests`, 7 passing): name fits / truncate-to-fit / truncate
leaves room for null at exact fit / `Fail` reports required size / zero-length buffer acts as a size
probe / buffer larger than 256 is accepted and returns the full name / `ex_w_policy` truncates NetBIOS
formats and keeps the size-probe contract for DNS formats.

**Manual, through mirrord (Windows)** against a deployment whose pod is
`payment-api-9b7855b77-vlt4r`:

```
# Environment.MachineName + GetComputerNameW
Environment.MachineName = 'payment-api-9b7'
GetComputerNameW(buf= 16) -> ok=True  name='payment-api-9b7'
GetComputerNameW(buf=256) -> ok=True  name='payment-api-9b7855b77-vlt4r'

# GetComputerNameExW size-probe + large buffer
probe(NULL,0)  -> ok=False err=234 (ERROR_MORE_DATA) requiredSize=28
retry(buf=28)  -> ok=True  size=27 name='payment-api-9b7855b77-vlt4r'
big(buf=1025)  -> ok=True  size=27 name='payment-api-9b7855b77-vlt4r'
```

A .NET 8/10 ASP.NET Core API using StackExchange.Redis (`ConnectionMultiplexer.Connect`) now returns
`200 {"redis":"pong"}` through mirrord instead of throwing `RedisConnectionException`.

**Outgoing TLS** (the regression noted above), through mirrord:

```
HttpClient GET https://example.com  -> 200
GCP Secret Manager (gRPC over TLS)  -> 500 Unauthenticated   (TLS handshake succeeds; identical to running without mirrord)
```

Confirmed by build bisection during development: without the NetBIOS truncation, SChannel/SSPI fails
with "the requested security package does not exist" and all outgoing TLS dies; the `ex_w_policy`
truncation restores it (`https://example.com -> 200`, and GCP gRPC-over-TLS reaches the server).

`cargo clippy -p mirrord-layer-win --tests -- -Wclippy::indexing_slicing -D warnings` (the CI lint) is clean.

## Compatibility

No new incompatibilities; the standard Win32 idioms all hold — fixed-buffer `GetComputerName[W/A]`,
`GetComputerNameExW` *query-size → grow → retry*, and large/`NI_MAXHOST` buffers. The hooks return the
**remote** name in every case (no fallback to the local OS name for the buffer-edge cases).

### Quality Checklist:

  documentation)


mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -216,3 +294,178 @@ pub fn is_remote_hostname(hostname: String) -> bool {
...
+    // Bound by *bytes* (the NetBIOS field is 15 OEM bytes), truncating at the largest UTF-8 char
+    // boundary that fits. For the usual ASCII names this is the same as taking 15 characters.
+    let mut end = 0;
+    for (i, ch) in name.char_indices() {
+        if i + ch.len_utf8() > NETBIOS_MAX_BYTES {
+            break;
+        }
+        end = i + ch.len_utf8();
+    }
+    let trimmed = name[..end].to_string();
```
meowjesty: Why is this not just `String::truncate`?

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
+                "{}: using original function (hostname feature disabled or no remote hostname)",
+                function_name
+            );
+            return original_fn();
+        }
+    };
+
+    let buffer_size = unsafe { *nSize } as usize;
+    let bytes = hostname.as_bytes();
+    let needed = bytes.len() + 1; // value + null terminator
```
meowjesty: Doesn't the `hostname` already come with the correct null terminator and length?

So this would make `needed` be actually "what's needed" + 1?

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
+            tracing::debug!(
+                "{}: buffer too small, need {} bytes, have {}",
+                function_name,
+                needed,
+                buffer_size
+            );
+            unsafe {
+                *nSize = needed as u32;
+                SetLastError(err);
+                WSASetLastError(WSAEFAULT.try_into().unwrap());
```
meowjesty: Seeing stuff like this makes me sad that you're not using `Detour` yet.

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
+            }
+            return ret_err;
+        }
+    }
+
+    // Write what fits (`buffer_size >= 1` here) plus the null terminator.
+    let copy_len = bytes.len().min(buffer_size - 1);
```
meowjesty: omg don't we have some `OsStr` or similar abstraction to avoid things like this?

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
+    unsafe {
+        std::ptr::copy_nonoverlapping(bytes.as_ptr() as *const i8, lpBuffer, copy_len);
+        // Null terminator - safe: copy_len <= buffer_size - 1.
+        *(lpBuffer.add(copy_len)) = 0;
+        // Actual length, excluding the null.
+        *nSize = copy_len as u32;
+    }
```
meowjesty: And said abstraction could even implement this stuff too, like `nicer_str.copy_to(lpBuffer)`

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
 
-                    // set error appropriate for the current hooked api
-                    SetLastError(err_buffer_overflow);
-                }
-                // FALSE - Buffer too small
-                return 0;
-            }
+    let buffer_size = unsafe { *nSize } as usize;
+    let name: Vec<u16> = hostname.encode_utf16().collect();
+    let needed = name.len() + 1; // value + null terminator
```
meowjesty: Same question here, isn't `hostname` already null terminated?

mirrord/layer-win/src/hooks/socket/hostname.rs
```diff
@@ -88,124 +121,169 @@ where
...
 
+    // Write what fits (`buffer_size >= 1` here) plus the null terminator.
+    let copy_len = name.len().min(buffer_size - 1);
+    unsafe {
+        std::ptr::copy_nonoverlapping(name.as_ptr(), lpBuffer, copy_len);
+        // UTF-16 null terminator - safe: copy_len <= buffer_size - 1.
+        *(lpBuffer.add(copy_len)) = 0;
+        // Code units written, excluding the null (matches Windows).
+        *nSize = copy_len as u32;
+    }
```
meowjesty: Sadness, if there's no abstraction, this could be a helper function at least.

meowjesty (changes requested):
Have you just tried asking nicely the computer for its name?

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: hostname is always a Rust string, so it means that the null terminator doesn't carry across

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: this was kind of an adhoc implementation of floor_char_boundary and we didnt want to panic on a bad truncate (since the target is a rust string), but it can be replaced by name.truncate(floor_char_boundary(n)). simplified code since floor_char_boundary is available and pushed

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: hostname is always a Rust string, so it means that the null terminator doesn't carry across

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: helper function added

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: should we push to a next pr? sounds worth pursuing

mirrord/layer-win/src/hooks/socket/hostname.rs
cristeigabriela: see prev thread

0x00A5 (approved):

meowjesty (approved):
👍

PR #4346, opened 2026-06-08
COR-1548 — managed handles contention under heavy load

cristeigabriela:
## Linear

https://linear.app/metalbear/issue/COR-1548/managed-handles-contention-under-heavy-load

## Summary

On Windows, mirrord could **drop operations under concurrency**. A `.NET HttpClient` firing many requests surfaced `SocketException: No such host is known` (`WSAHOST_NOT_FOUND`) for names that resolve fine on a single lookup, and concurrent file open/close could drop a file's IOCP completion binding. Both were the **same root pattern**: a single `RwLock<HashMap>` whose non-blocking `try_*` gave up under contention and dropped the op.

## Root cause (bisected)

Concurrent DNS was isolated path by path: only `HttpClient` failed (DNS **plus** sustained concurrent socket I/O), while raw `Dns.GetHostAddressesAsync` / `Socket.ConnectAsync` were 50/50. The async `GetAddrInfoExW` resolution and the socket reads share one 4-worker `task_pool` serialized on a single `PROXY_CONNECTION`, so under load the in-flight DNS queries' `ManagedRegistry` entries pile up, the registry's single `RwLock` stays write-contended past the spin budget, and `try_insert` gives up → `WSAHOST_NOT_FOUND`. The layer log named it 1:1 with each failure. Hammering the **file** hooks the same way surfaced the identical bug in a second map (`iocp::binding`'s `FILE_BINDINGS`).

## What changed

**One sharded map, used everywhere.** `ShardedMap<K, V>` holds the shard array, selection, bounded-spin retry, and contention logging in one place; an entry lives in `hash(key) & 31` so unrelated handles take different locks. `ManagedRegistry` (files + DNS) is now a thin wrapper over it, and the bespoke `FILE_BINDINGS` map is gone.

**Per-handle state unified.** The IOCP `(port, key)` binding moved out of its own map into the file's `HandleContext` (next to `skip_on_success`). `bind`/`unbind`/lookup are operations on the file's existing record, so the binding drops **atomically** when the handle is removed on close — no second lock, no unbind-then-remove leak window. (The old insert-failed → fall-back-to-local-open path was itself buggy — it leaked the remote fd and returned a local handle — so removing it fixes a latent leak.)

**`insert` never drops a registration.** It spins, then under pathological contention **blocks** on the shard write lock rather than failing — a managed handle the caller is about to hand out *must* be tracked. The wait is sub-millisecond and in-memory (a bare `HashMap` op holding no other lock, never the loader lock, off the async-completion path); `get`/`remove` stay non-blocking. This takes the residual from ~1/10,000 inserts at 100-way concurrency to **0**.

**Observability, kept in.** `debug` when an op spins past a threshold (early warning), `warn` when insert has to block, `error` when get/remove bail — each tagged with the map label + spin count.

## Testing

| | result |
|---|---|
| **Concurrent DNS, 50-way × 100 rounds** | **5,000 / 0 fail**, 0 drops |
| **File open/read/close via `fs: read`**, 50-way × 4 | **50/50**, content correct, 200 bind/unbind, 0 errors |
| **Stress, 100-way × 100** | registry held (0 insert drops); the serialized proxy/agent connection is the limit at that level |

### Quality Checklist:

  documentation)


mirrord/layer-win/src/managed/sharded_map.rs
```diff
@@ -0,0 +1,249 @@
...
+                Err(TryLockError::Poisoned(p)) => {
+                    p.into_inner().insert(key, value);
+                    return;
+                }
+                Err(TryLockError::WouldBlock) => std::hint::spin_loop(),
+            }
+        }
+        // The spin budget didn't win. Block to acquire rather than drop a registration the caller
+        // depends on -- loud, because reaching here means genuinely pathological contention.
+        tracing::warn!(
```
0x00A5, line 179: nit: should this be error too?

mirrord/layer-win/src/managed/sharded_map.rs
```diff
@@ -0,0 +1,249 @@
...
+        let shard = self.shard(&key);
+        for attempt in 0..SPIN {
+            match shard.try_write() {
+                Ok(mut map) => {
+                    self.note_contention("insert", attempt);
+                    map.insert(key, value);
+                    return;
+                }
+                Err(TryLockError::Poisoned(p)) => {
+                    p.into_inner().insert(key, value);
```
0x00A5, line 171: nit: do you think we should log a warn here?

0x00A5 (dismissed):

mirrord/layer-win/src/managed/sharded_map.rs
cristeigabriela: do you think this should be an error? personally i just think it's a sign of heavy contention, which should be warned about since it could make the app act a bit weird in the worst of cases, but everything should still work

mirrord/layer-win/src/managed/sharded_map.rs
cristeigabriela: good call

0x00A5 (approved):

PR #4348, opened 2026-06-09
chaos: rule types and effect-selector validation

gememma:
Adds types for chaos rules and requests, see [Linear issue](https://linear.app/metalbear/issue/COR-1512/chaos-rule-selector-validation) for details.

- Validation occurs when a request is turned into a rule, and invalid effect/selector combinations cannot be expressed in a chaos rule
- Unit tests exist to test parsing for malformed requests, invalid rules and valid rules
- Note that lots of `clippy` warnings are still active for code elsewhere, but this PR doesn't attempt to fix them
- Easiest to review commit-by-commit :)


meowjesty (approved):
👍

PR #4349, opened 2026-06-09
mirrord ui: rename "Signed in as" to "Running as" (PRO-122)

hank-metalbear:
## Summary

In the mirrord UI user menu, the identity label said **"Signed in as"**. That identity is the active cluster identity that the CLI queries, not an authenticated session, so the label wrongly implied a sign-in/sign-out flow that does not exist in the UI. This renames it to **"Running as"**, which describes what the value actually is.

Per [PRO-122](https://linear.app/metalbear/issue/PRO-122) (Liron's suggestion, agreed by Aviram).

Before / after (user menu):
```
Signed in as        ->    Running as
alice@cluster             alice@cluster
```

## Test plan

- Open the mirrord UI, click the user menu in the header, confirm the label now reads "Running as" above the identity.
- One-word JSX text change in `packages/monitor/src/components/AppHeader.tsx`; no logic or types touched.

### DIFF
--- a/changelog.d/+running-as-label.changed.md
+++ b/changelog.d/+running-as-label.changed.md
@@ -1,0 +1,1 @@
+Renamed the mirrord UI identity label from "Signed in as" to "Running as", since it reflects the active cluster identity queried by the CLI rather than an authenticated session.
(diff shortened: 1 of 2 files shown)


aviramha (dismissed):

hank-metalbear:
@aviramha can i get another approval?

aviramha (approved):

PR #4351, opened 2026-06-10
Messy chaos

meowjesty:
Kinda messy, but I could not make it look better.

Ignore the `Incoming...` parts, I added it just to help me visualize everything.

This moves the `chaos_rx` from the `message_bus` to the bg tasks, and implements the message and effect applying (still not using the full effects).

Thoughts? Any ideas/code to improve this is welcomed.

### DIFF
--- a/mirrord/cli/src/port_forward.rs
+++ b/mirrord/cli/src/port_forward.rs
@@ -12,7 +12,7 @@ use mirrord_intproxy::{
     background_tasks::{BackgroundTasks, TaskError, TaskSender, TaskUpdate},
     main_tasks::{ProxyMessage, ToLayer},
     proxies::incoming::{IncomingProxy, IncomingProxyError, IncomingProxyMessage},
-    session_monitor::MonitorTx,
+    session_monitor::{MonitorTx, chaos::ChaosWatcherRx},
 };
 use mirrord_intproxy_protocol::{
     IncomingRequest, IncomingResponse, LayerId, PortSubscribe, PortSubscription,
@@ -40,7 +40,7 @@ use tokio::{
     select,
     sync::{
         mpsc::{self, Receiver, Sender},
-        oneshot,
+        oneshot, watch,
     },
 };
 use tokio_stream::{StreamMap, wrappers::TcpListenerStream};
@@ -579,6 +579,8 @@ impl ReversePortForwarder {
         let mut background_tasks: BackgroundTasks<(), ProxyMessage, IncomingProxyError> =
             BackgroundTasks::new(agent_connection.tx_handle());
 
+        let (_chaos_tx, chaos_rx) = watch::channel(Default::default());
+
         let incoming = background_tasks.register(
             IncomingProxy::new(
                 idle_local_http_connection_timeout,
@@ -588,6 +590,7 @@ impl ReversePortForwarder {
                     .or_else(|| network_config.https_delivery.clone())
                     .unwrap_or_default(),
                 MonitorTx::disabled(),
+                ChaosWatcherRx::new(chaos_rx),
             ),
             (),
             512,
--- a/mirrord/config/src/feature/network/filter.rs
+++ b/mirrord/config/src/feature/network/filter.rs
@@ -16,6 +16,7 @@ use nom::{
 };
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tracing::Level;
 
 /// The protocols we support in [`ProtocolAndAddressFilter`].
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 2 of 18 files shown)


mirrord/config/src/feature/network/filter.rs
```diff
@@ -77,6 +78,23 @@ impl AddressFilter {
...
+
+    pub fn matches_socket_address(&self, address: &SocketAddress) -> bool {
+        match self {
+            AddressFilter::Port(port) => {
+                *port == 0 || address.get_port().is_some_and(|p| *port == p)
+            }
+            AddressFilter::Socket(socket_addr) => {
+                socket_addr.ip().is_unspecified()
+                    || address.get_ip().is_some_and(|ip| socket_addr.ip() == ip)
+            }
```
gememma, line 99: doesnt this have to check the port as well? docks for `::Socket` variant say `/// Just a plain old ['SocketAddr'], specified as 'a.b.c.d:e'.`

mirrord/cli/src/port_forward.rs
```diff
@@ -588,6 +590,7 @@ impl ReversePortForwarder {
                     .or_else(|| network_config.https_delivery.clone())
                     .unwrap_or_default(),
                 MonitorTx::disabled(),
+                ChaosWatcherRx::new(chaos_rx),
```
gememma, line 593: not sure if we should let people add rules to the non-exec stuff, maybe we do `ChaosWatcherRx::disabled()` like `MonitorTx`

mirrord/intproxy/src/proxies/incoming/tests.rs
```diff
@@ -61,11 +62,14 @@ async fn http_request_terminates_on_remote_close(#[case] steal_type: StealType)
...
     let local_addr = local_listener.local_addr().unwrap();
 
+    let (_chaos_tx, chaos_rx) = watch::channel(Default::default());
+
     let (conn, _, out) = Connection::dummy();
     let proxy = IncomingProxy::new(
         Duration::from_secs(3),
         Default::default(),
         crate::session_monitor::MonitorTx::disabled(),
+        ChaosWatcherRx::new(chaos_rx),
```
gememma, line 72: oh we could definitely mock this for testing `IncomingProxy` behaviour w/ chaos rules at some point 👀

mirrord/intproxy/src/proxies/outgoing/chaos.rs
```diff
@@ -0,0 +1,91 @@
...
+            layer_id,
+        ) = self
+        else {
+            return None;
+        };
+
+        let Some(rule) = rules
+            .iter()
+            .filter(|rule| rule.applies_to_address(remote_address, *protocol))
+            .max_by_key(|rule| rule.priority)
```
gememma, line 47: ooh, `max_by_key` is cool (i bet this could be an ugly little `fold` too >:3)

mirrord/config/src/feature/network/filter.rs
```diff
@@ -77,6 +78,23 @@ impl AddressFilter {
...
+            AddressFilter::Subnet(ip_net, port) if let Some(ip) = address.get_ip() => {
+                ip_net.contains(&ip)
+                    && (*port == 0 || address.get_port().is_some_and(|p| *port == p))
+            }
+            _ => false,
```
gememma, line 109: `AddressFilter::Name` is valid for the user to give too :/ maybe we resolve it during rule validation? not sure

gememma (approved):
looks good, not sure about `AddressFilter::matches_socket_address()`? maybe i missed something in your reasoning 🤔

mirrord/protocol/src/outgoing.rs
```diff
@@ -37,6 +38,13 @@ impl SocketAddress {
             SocketAddress::Unix(_) => None,
         }
     }
+
+    pub const fn get_ip(&self) -> Option<IpAddr> {
+        match self {
+            SocketAddress::Ip(address) => Some(address.ip()),
+            SocketAddress::Unix(_) => None,
+        }
+    }
```
gememma, line 47: possibly not worth a protocol change? 👀

meowjesty: I think it's fine, should be a small bump in version, and we won't have to put this in some random place as an extension trait.

mirrord/intproxy/src/proxies/outgoing/chaos.rs
meowjesty: everything is just `fold` in the end

mirrord/config/src/feature/network/filter.rs
meowjesty: added the check for port too

mirrord/intproxy/src/proxies/outgoing/chaos.rs
```diff
@@ -55,37 +56,48 @@ impl ApplyChaosRuleLol for OutgoingProxyMessage {
...
-                Some(OutgoingThingToDo::Latency {
-                    effect: effect.clone(),
-                })
+                percentage
+                    .should_apply_for_hit(hit_count)
+                    .then(|| OutgoingThingToDo::Latency {
+                        delay: {
+                            let applied_n_times = percentage.applied_before_hit(hit_count);
+
+                            delay + jitter.saturating_mul(applied_n_times)
```
gememma, line 89: i think this is calculating a backoff, not jitter? it does `delay + jitter`, then next time `delay + jitter*2`, etc. instead of doing `delay + some_random_amount_between(0, jitter)` - will fix in my merge.

PR #4362, opened 2026-06-11
latency in the proper place, and improved where we handle the chaos with messages (proxy itself so we have access to some required proxy stuff)

meowjesty:


gememma (approved):
yay 🦺

PR #4372, opened 2026-06-15
ui: reuse token and detect already-running instance

aviramha:
Store the UI auth token in ~/.mirrord/token and hold an exclusive advisory lock on it while the server runs. A second `mirrord ui` invocation detects the held lock, reads back the published token, and prints the existing URL instead of starting a duplicate server. The lock is released by the OS even on a hard kill, so a stale file is reclaimed on the next run; a graceful shutdown removes the file.


gememma (dismissed):
🔒

gememma (approved):

PR #4374, opened 2026-06-15
Remove comments from `mirrord up init`

MintSoup:
### Quality Checklist:

  documentation)

https://linear.app/metalbear/issue/COR-1545/remove-comments-from-mirrord-up-output-file

### DIFF
--- a/mirrord/up/src/init.rs
+++ b/mirrord/up/src/init.rs
@@ -1,9 +1,7 @@
 //! Interactive wizard that generates a skeleton `mirrord-up.yaml`.
 //!
 //! The wizard prompts for a small, curated subset of the full schema and
-//! writes the result as YAML. For fields the wizard knows about but the
-//! user did not customize, the output contains a commented-out template
-//! line so the user can discover the field without consulting the docs.
+//! writes the result as YAML.
 //!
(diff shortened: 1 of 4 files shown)


mirrord/up/src/init.rs
```diff
@@ -307,172 +308,105 @@ fn prompt_run() -> Result<RunConfig, InitError> {
...
+    for (_, v) in map.iter_mut() {
+        prune(v);
     }
+
+    map.retain(|_, v| match v {
+        serde_yaml::Value::Null => false,
+        serde_yaml::Value::Mapping(m) => m.is_empty().not(),
+        serde_yaml::Value::Sequence(s) => s.is_empty().not(),
+        _ => true,
+    });
```
Razz4780, line 354: I assume this is only fine because of what the skeleton actually contains, right? In general case, we can't remove all empty sequences (e.g. incoming.ports)

Razz4780 (approved):
Looks good, one question

mirrord/up/src/init.rs
MintSoup: Yeah this is fine because `up` config doesn't have any semantics that would get messed up by removing empty sequences

PR #4380, opened 2026-06-16
analytics+cli: add session key information to telemetry events

iniw:
### Quality Checklist:

  documentation)

## Related PRs:

- https://github.com/metalbear-co/operator/pull/1788
- https://github.com/metalbear-co/infra/pull/206
- https://github.com/metalbear-co/infra/pull/226

### DIFF
--- a/mirrord/analytics/src/lib.rs
+++ b/mirrord/analytics/src/lib.rs
@@ -4,6 +4,7 @@ use std::{collections::HashMap, str::FromStr, time::Instant};
 
 use base64::{Engine as _, engine::general_purpose};
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use tracing::{Level, info};
 use uuid::Uuid;
 
@@ -150,6 +151,18 @@ impl AnalyticsHash {
     pub fn from_base64(val: &str) -> Self {
         AnalyticsHash(val.to_owned())
     }
+
+    /// Deterministically hashes a session key with the operator license fingerprint.
+    pub fn for_session_key(key: &str, license_fingerprint: &str) -> Self {
+        let mut hasher = Sha256::new();
+        hasher.update(license_fingerprint.as_bytes());
+        hasher.update(key.as_bytes());
+        Self::from_bytes(&hasher.finalize())
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
 }
 
 /// Structs that collect analytics about themselves should implement this trait
@@ -256,6 +269,7 @@ pub struct AnalyticsReporter {
     operator_properties: Option<AnalyticsOperatorProperties>,
     watch: drain::Watch,
     target: ReportTarget,
+    session_key: Option<String>,
 }
 
 impl AnalyticsReporter {
@@ -264,6 +278,7 @@ impl AnalyticsReporter {
         execution_kind: ExecutionKind,
         watch: drain::Watch,
         machine_id: Uuid,
+        session_key: Option<String>,
     ) -> Self {
         let mut analytics = Analytics::default();
         analytics.add("execution_kind", execution_kind as u32);
@@ -279,6 +294,7 @@ impl AnalyticsReporter {
             start_instant: Instant::now(),
             watch,
             target: ReportTarget::ClientSession,
+            session_key,
         }
     }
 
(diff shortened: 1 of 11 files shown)


mirrord/analytics/src/lib.rs
```diff
@@ -150,6 +151,18 @@ impl AnalyticsHash {
     pub fn from_base64(val: &str) -> Self {
         AnalyticsHash(val.to_owned())
     }
+
+    /// Deterministically hashes a session key with the operator license fingerprint.
+    pub fn for_session_key(key: &str, license_fingerprint: &str) -> Self {
```
meowjesty, line 156: The ticket says to use org name, any special reason to use the fingerprint instead?

I think the fingerprint changes on license updates, wonder if that's ok?

mirrord/cli/src/wizard.rs
```diff
@@ -98,6 +98,7 @@ where
         ExecutionKind::Wizard,
         watch,
         user_data.machine_id(),
+        None,
```
meowjesty, line 101: wizard doesn't have to report it?

I don't even know if the wizard can be run with config, it can right?

mirrord/analytics/src/preview.rs
```diff
@@ -39,9 +43,15 @@ impl PreviewEvent {
...
-        let mut reporter =
-            AnalyticsReporter::new(enabled, ExecutionKind::Preview, watch, machine_id);
+    pub fn cli_report_analytics(&self, enabled: bool, watch: drain::Watch, machine_id: Uuid) {
+        let mut reporter = AnalyticsReporter::new(
+            enabled,
+            ExecutionKind::Preview,
+            watch,
+            machine_id,
+            // We'll report the session key ourselves through `preview_key_identifier`.
+            None,
```
meowjesty, line 53: Why?

meowjesty (commented):
Questions

mirrord/cli/src/wizard.rs
iniw: > wizard doesn't have to report it?

There is no session key to report in this case

> I don't even know if the wizard can be run with config, it can right?

Nope, only kubeconfig apparently:

```rust
#[derive(Args, Debug)]
pub struct WizardArgs {
    /// Accept/reject invalid certificates.
    #[arg(env = "MIRRORD_ACCEPT_INVALID_CERTIFICATES", short = 'c', long, default_missing_value="true", num_args=0..=1, require_equals=true
    )]
    pub accept_invalid_certificates: Option<bool>,

    /// Kube context to use from Kubeconfig.
    #[arg(env = "MIRRORD_KUBE_CONTEXT", long)]
    pub context: Option<String>,

    /// Kubeconfig.
    #[arg(env = "MIRRORD_KUBECONFIG", long)]
    pub kubeconfig: Option<String>,

    /// Controls whether mirrord sends telemetry data to MetalBear cloud. Telemetry sent doesn't
    /// contain personal identifiers or any data that should be considered sensitive. It is used to
    /// improve the product.
    /// [More information](https://github.com/metalbear-co/mirrord/blob/main/TELEMETRY.md).
    #[arg(env = "MIRRORD_TELEMETRY", long, default_value = "true")]
    pub telemetry: bool,
}
```

mirrord/analytics/src/lib.rs
iniw: > any special reason to use the fingerprint instead?

Because it's the only information that we have on the CLI analytics side without having access to an `OperatorApi` (like we have in the preview event):

```rust
/// Extra fields for `AnalyticsReport` when using mirrord with operator.
#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct AnalyticsOperatorProperties {
    /// client certificate public key
    pub client_hash: Option<AnalyticsHash>,

    /// sha256 fingerprint from operator license
    pub license_hash: Option<AnalyticsHash>,
}
```

But I was reading the ticket again and it seems like it only cares about the `operator-session` event, which is done on the operator, so I think I can remove all this stuff and use the license name directly. Will check with the higher powers.

mirrord/analytics/src/preview.rs
iniw: Because `preview_key_identifier` predates `session_key_identifier` and they'll both have the same info, so it doesn't make sense so send both. Ideally we'd have named it `session_key_identifier` from the beginning but renaming schema fields is scary. Will check if we want both or only one.

mirrord/analytics/src/lib.rs
iniw: Okay the higher powers have ok'd the use of license hash, which allows for more events to have a `session_key_identifier` property. So I think this is fine.

meowjesty (approved):
👍

PR #4391, opened 2026-06-17
Adds the chaos feature

meowjesty:
Adds the chaos testing feature (mvp) thingy.

The users interact with the chaos rule system through a local server, that gets started when `mirrord ui` is running, so we're hitching a ride on this whole thing to pass our messages from the user to whatever intproxy they want to interact with (chaos test).

- Adds a new `chaos_router` for both `ui` and `session_monitor`, for dealing with `ChaosRuleRequest` and `ChaosRule`:
  - these are 2 different routers, even though they look samey, and are pretty much the same api, but the `ui` router is just for taking user requests and sending them to the `session_monitor` api;
- Adds analytics for this whole thing, so you know the "hit count" of each rule;
- Refactors some small stuff around the code:
  - changes a `mod.rs` to follow our style of not having `mod.rs`s;
  - adds some helper functions to some types in `protocol`;
  - derives a few more traits (like `Hash`) for some config types that are used by the chaos rule stuff;
  - changes `Arc<AppState>` to be `Clone` (by having its interior be clonable);

### Quality Checklist:

  documentation)


mirrord/cli/src/ui/chaos.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/// Checks if the `session_id` is available in our [`AppState`] (which means that there's some
+/// mirrord session running with this id), and then extracts the [`SessionClient`] that can be used
+/// to make requests to the session monitor server of this particular session. Puts this
+/// `SessionClient` as a request [`Extension`] and passes it along to the chaos route handlers.
+#[tracing::instrument(level = Level::INFO, skip(state, request, next))]
+async fn get_session_client_middleware(
+    State(state): State<AppState>,
+    Path(params): Path<HashMap<String, String>>,
```
0x00A5: Should we use the SessionId type rather than HashMap here?

mirrord/intproxy/src/session_monitor/chaos/api.rs
```diff
@@ -0,0 +1,181 @@
...
+        )
+}
+
+/// - `POST /chaos/rules/{session_id}`: creates a new [`ChaosRule`] based on the
+///   [`ChaosRuleRequest`] that we received. When a `ChaosRule` is built from a `ChaosRuleRequest`
+///   it auto-generates a [`ChaosRule::id`], which makes every request here create a unique
+///   `ChaosRule` (even if all the other fields are the same as another `ChaosRule`, we only compare
+///   ids);
+async fn post_create_rule(
+    Path(_): Path<SessionId>,
```
0x00A5: The session id path param is unused everywhere in intproxy API. I think we should remove it. it's the same session after the request is dispatched to each session client.

mirrord/intproxy/protocol/src/lib.rs
```diff
@@ -145,6 +145,8 @@ pub struct OutgoingConnectRequest {
     pub remote_address: SocketAddress,
     /// The protocol stack the user application wants to use.
     pub protocol: NetProtocol,
+
+    pub hostname: Option<String>,
```
0x00A5, line 184: Can we put this new field under a metadata struct? Or in some other way to make it clear that we are always connecting to the socket addr.
(Also need some doc)

mirrord/intproxy/src/session_monitor/chaos.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Updates a [`ChaosRule`] whose [`ChaosRule::id`] matches `new_rule`'s.
+    ///
+    /// This is possible because of the custom `PartialEq` and `Hash` implementations of
+    /// `ChaosRule`.
+    #[tracing::instrument(level = Level::INFO)]
+    pub(super) fn update_rule(&self, new_rule: ChaosRule) -> Option<ChaosRule> {
+        let mut old_rule = None;
+        self.0
+            .send_modify(|current_rules| old_rule = current_rules.replace(new_rule));
```
0x00A5, line 144: Do we consider them the same rule and let new rule inherit old rule's hit count? Currently it resets it to zero.

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -668,71 +710,89 @@ impl BackgroundTask for OutgoingProxy {
...
+                                connect.connect,
+                                NetProtocol::Seqpacket,
+                                Some(connect.uid),
+                                message_bus,
+                            ).await?,
+                        }
+                        Some(OutgoingProxyMessage::Layer(request, message_id, layer_id)) => {
+                            if let OutgoingRequest::Connect(connect) = &request
+                                && let Some(chaos_reigns) = self.chaos_effect_for_connect(connect, message_id, layer_id)
+                                && Self::apply_chaos_effect(chaos_reigns, message_bus).await.is_break()
```
0x00A5: we call sleep inside apply_chaos_effect(), it will then block the whole `loop { select! {} }`, where it should only delay that one connection only. I think we should spawn here instead, the spawned task sleep for the given amount of time, and send the message to message bus wether it is the connect message or the error to layer.

Also `is_break()` and then `continue` is quite confusing :D

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -668,71 +710,89 @@ impl BackgroundTask for OutgoingProxy {
...
-                    Some(OutgoingProxyMessage::ConnectionRefresh(refresh)) => self.handle_connection_refresh(message_bus, refresh).await,
-                    Some(OutgoingProxyMessage::AgentProtocolVersion(version)) => {
-                        self.protocol_version.replace(version);
                     }
                 },
 
                 Some(task_update) = self.background_tasks.as_mut().unwrap().next() => match task_update {
                     (id, TaskUpdate::Message(bytes)) => {
+                        if let Some(chaos_reigns) = self.chaos_effect_for_write(id)
+                            && Self::apply_chaos_effect(chaos_reigns, message_bus).await.is_break()
```
0x00A5: Same here, we should spawn instead of blocking the main select loop.

mirrord/cli/src/ui/chaos.rs
gememma: this is session id, rule id for some routes, session id only for others, right? since this is middleware for all routes? i dont know if you can change this

mirrord/intproxy/src/session_monitor/chaos.rs
gememma: the id is supposed to stay the same but the hit count is supposed to reset - see https://metalbear.slack.com/archives/C0B4HTCPJLF/p1780560868415119, https://metalbear.slack.com/archives/C0B4HTCPJLF/p1780561229871739?thread_ts=1780560868.415119&cid=C0B4HTCPJLF

it means session IDs are unique to live rules but not all time rules, which i think is kind of weird? especially since you can compare two different rules with Eq/ Hash/ whatever and theyll return true :\

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -110,12 +112,25 @@ impl From<AgentLostOutgoingResponse> for ToLayer {
...
+    hostname: Option<String>,
     requested_at: Instant,
     layer_id: LayerId,
     message_id: MessageId,
     id: u128,
 }
 
+/// Original connection metadata requested by the layer.
+///
+/// Say the user apps wants to make a connection to `https://www.przepisy.pl/`,
```
gememma: 🔍 👀 thanks vale... for catching this...
```suggestion
/// Say the user app wants to make a connection to `https://www.przepisy.pl/`,
```

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -668,71 +710,89 @@ impl BackgroundTask for OutgoingProxy {
...
-                    None => {
-                        tracing::debug!("Message bus closed, exiting");
-                        break Ok(());
-                    },
-                    Some(OutgoingProxyMessage::AgentStream(req)) => match req {
-                        DaemonTcpOutgoing::Close(close) => {
-                            let id = InterceptorId { connection_id: close, protocol: NetProtocol::Stream};
-                            self.txs.remove(&id);
+                msg = message_bus.recv() => {
+                    match msg {
```
gememma: i think changing the indentation here produced this huge diff ☠️  same in [incoming.rs](https://github.com/metalbear-co/mirrord/pull/4391/changes#diff-066a471da780e31ee0a75bbbdb5d76dd843ae3e9c4b6300f283495a1b5a6d6a3)

mirrord/intproxy/src/proxies/outgoing.rs
gememma: if we dont block the loop, does that mean messages could start going out of order? like if you add latency to 50% of requests to the same place, then the regular ones will overtake the slow ones - not sure if this is actually a problem or not

mirrord/intproxy/src/session_monitor/chaos/analytics.rs
```diff
@@ -0,0 +1,309 @@
...
+    time::Instant,
+};
+
+use mirrord_analytics::{AnalyticValue, Analytics, AnalyticsReporter, CollectAnalytics, Reporter};
+
+use crate::session_monitor::chaos::rules::{ChaosRule, ChaosSelector};
+
+/// Wraps AnalyticsReporter for use reporting chaos rule analytics. Stores deleted rules instead of
+/// sending anaytics every time a rule is deleted. Sends all rules that were in effect over the
+/// whole session when the session ends.
```
gememma: c'mon @gememma keep it together (if you want to ignore this, ill fix it post-merge)
```suggestion
/// Wraps [`AnalyticsReporter`] for use reporting chaos rule analytics. Stores deleted rules instead
/// of sending analytics every time a rule is deleted. Sends all rules that were in effect over the
/// whole session when the session ends.
```

mirrord/intproxy/src/session_monitor/chaos/rules.rs
```diff
@@ -0,0 +1,1045 @@
...
+/// [`ChaosEffectType`] specified by the user is applied, for example artificial latency or a
+/// connection error.
+///
+/// Rules can be applied to TCP, HTTP or FS traffic with a given percentage probability. The traffic
+/// type is inferred from a request's `selector` ([`ChaosSelectorRequest`]).
+///
+/// For examples of requests and the rules they correspond to, see [`mod test`].
+///
+/// _WARNING: This type implements `PartialEq` for testing purposes - trying to compare rules
+/// without accounting for their unique `id`s will fail!_
```
gememma:
```suggestion
/// _WARNING: This type implements `PartialEq` but only compares the `id` fields._
```

mirrord/intproxy/src/session_monitor/chaos.rs
```diff
@@ -0,0 +1,149 @@
+use std::{
```
gememma, line 9: docs maybe?

mirrord/intproxy/src/session_monitor/chaos.rs
```diff
@@ -0,0 +1,149 @@
...
+pub type ChaosRuleList = HashSet<ChaosRule>;
+
+/// mirrord sessions can be either a [`Uuid`] when `operator = false`, or this random string when
+/// the operator is responsible for creating a session. To typefy our REST apis, we have this thing.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum SessionId {
+    Uuid(Uuid),
+    VarChar(String),
+}
```
gememma, line 67: hopefully we can remove this in future ;_; https://metalbear.slack.com/archives/C08TYT624SV/p1781868455494629?thread_ts=1781867007.371559&cid=C08TYT624SV

changelog.d/+dns-filter-stable.changed.md
```diff
@@ -0,0 +1 @@
+`feature.network.dns.filter` is no longer marked as unstable in config docs or CLI warnings.
```
gememma: i feel like all these changelog files should not be here 🤔

gememma (commented):
I agree with @0x00A5 's comments, my other comments are all super minor (apart from the changelog files, we should fix those - i assume its a merge commit thing again) but _apart_ from all that i think it looks great, very chaos

mirrord/cli/src/ui/chaos.rs
0x00A5: Ah, I see

mirrord/intproxy/src/proxies/outgoing.rs
0x00A5: Yeah, going out of order is a problem ☠️ . For connect request, we need to make sure to add latency and then proceed with updating intproxy state and sending the message to agent together. Not mutate state first, and then sleep, and then send the connect request. The intproxy state must reflect the sent order of connect requests.
And for writing bytes, outbound bytes could go out of order. (first chunk sleeps X seconds, second chunk sleeps Y seconds, Y < X)

mirrord/intproxy/src/proxies/outgoing.rs
0x00A5: spawning on each write is not a good idea 💩 , it can change outbound bytes order. Maybe a delayed write worker per connection is better. We spawn a delayed write worker per connection. Outgoing proxy's main event loop send the message to the worker in order. The worker receives the bytes in order in a loop, and add latency and send. The worker's loop will preserve outbound byte order.

mirrord/intproxy/src/proxies/outgoing.rs
0x00A5: Regarding outbound write latency and bytes order: https://github.com/metalbear-co/mirrord/pull/4391#discussion_r3443649155

mirrord/cli/src/ui/chaos.rs
meowjesty: I think this is the only way to do many params in middleware. I had only `session_id` before and when you pass session id + rule id the middleware would break and return 500.

mirrord/intproxy/protocol/src/lib.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4407/changes/64d3bc62dd319f782dadd5b6a42fcb729a892991

mirrord/intproxy/src/session_monitor/chaos/api.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4407/changes/20423a5e65f630a0db86e3f15d8846597750f154

mirrord/intproxy/src/proxies/outgoing.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4407/changes/af3f058cf71aa344bdd4bfaff5dc8f48d3f25b2b

mirrord/intproxy/protocol/src/lib.rs
```diff
@@ -145,6 +145,43 @@ pub struct OutgoingConnectRequest {
...
     pub protocol: NetProtocol,
+
+    /// Metadata for this outgoing connection request.
+    ///
+    /// The fields here are not used by the request per se, they're useful for things related to
+    /// the connection, such as applying chaos rules.
+    pub metadata: OutgoingConnectRequestMetadata,
+}
+
+impl OutgoingConnectRequest {
```
0x00A5, line 156: nit: do we need this impl block? everything is public
Up to you to keep this

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -533,12 +575,20 @@ impl OutgoingProxy {
...
 
-        let msg = request
-            .protocol
-            .wrap_agent_connect(request.remote_address, uid);
-        message_bus.send_agent(msg).await;
-
-        Ok(())
+        if let Some(uid) = uid
+            && self
+                .chaos_effect_for_connect_latency(&request, uid, message_bus)
```
0x00A5: handle_connect_request() updates a bunch of state before this point. It's gonna cause problem if we update state and then sleep, and then send message. We should sleep first and then update state and send message together. No state should be tracked for a connect request, until the delay sleep is done.

**Slop machine summarized issues with this approach**:
In `handle_connect_request()`, when chaos latency applies to v2 connect request:
1. connection_id is generated and registered in connections_in_layers
2. State (ConnectInProgress) is inserted into v2_reqs — state tracking happens now
3. chaos_effect_for_connect_latency spawns a task that sleeps, then sends the agent connect message

During the sleep window, other connect requests (NOT affected by chaos) arrive:
1. They register their state into v2_reqs at positions T+1, T+2, etc.
2. They send their agent messages immediately (no delay)

The agent receives messages for T+1, T+2 before T+0. When it responds, it sends responses in the order it processed them. Even though v2 uses UID-based matching, some problems:
1. Connection refresh during sleep: If handle_connection_refresh fires during the sleep window, it drains v2_reqs and sends error responses for all pending connections — including one the agent never heard about. Meanwhile the spawned task still wakes up and fires the connect message to the (now new) agent, which has no matching state.
2. requested_at is wrong: It's set before the sleep, so the elapsed-time tracking includes the artificial latency, not the actual agent round-trip.
3. Orphaned state on channel drop: If the agent tx is dropped before the spawned task fires, the state sits in v2_reqs forever but the agent never sees the request.

**An alternative to consider (slop machine proposed, I think it makes sense)**:
Add an OutgoingProxyMessage::DeferredConnect { message_id, layer_id, request, uid } variant. Before checking chaos, don't register any state. If chaos fires, spawn a task that sleeps then sends this variant back to the proxy's message_bus sender. The existing OutgoingProxyMessage::Layer handler's connect path then runs normally — state registration, then agent send — after the delay, maintaining the invariant that state enters v2_reqs exactly when the agent is told about the connection.

mirrord/intproxy/src/proxies/incoming.rs
```diff
@@ -974,7 +974,7 @@ impl BackgroundTask for IncomingProxy {
 
         loop {
             tokio::select! {
-                msg = message_bus.recv() => match msg {
+                msg = message_bus.recv() =>  match msg {
```
0x00A5, line 977: Huh I see, abusing linter doesn't work in macro :D

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -715,6 +773,12 @@ impl BackgroundTask for OutgoingProxy {
                         ).await?,
                     }
                     Some(OutgoingProxyMessage::Layer(request, message_id, layer_id)) => {
+                        if let OutgoingRequest::Connect(connect) = &request
+                            && self.chaos_effect_for_connect_error(connect, message_id, layer_id, message_bus).await.is_break()
+                        {
+                            continue;
+                        }
```
0x00A5: Please change this `is_break()` and then `continue` thingy ༼;´༎ຶ ۝ ༎ຶ༽ 
Something like returning a `bool`, `true` if affected by chaos, and then
```
if let OutgoingRequest::Connect(connect) = &request
    && self.chaos_effect_for_connect_error(connect, message_id, layer_id, message_bus).await.not() 
{
    self.handle_layer_request(request, layer_id, message_id, message_bus).await?;
}
```

mirrord/intproxy/src/proxies/outgoing.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4391/changes/7465c96cc146e77213bdc5439235f9027d38470a

mirrord/intproxy/src/main_tasks.rs
```diff
@@ -35,7 +35,7 @@ impl ProxyMessage {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
0x00A5: nit: doesn't look like we need this

0x00A5 (dismissed):

mirrord/intproxy/src/main_tasks.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4391/changes/cba6a47735e2adcc05dac72e775b27b8bf99dd98

t4lz (approved):

PR #4397, opened 2026-06-18
chaos: docs and UX for serializing rules

gememma:
- docs for analytics and rules
- switched a random `impl` into a `impl From`
- improved rule serialization to be less ugly and match the request more closely
    - `Duration` is displayed as `field_ms: number `
    - selector is tagged with "type" field instead of nested
    - `Percentage` is displayed as `u32`
    - `AddressFilter` uses `Display`/ `FromStr`

### Example rule in JSON
Screenshot 2026-06-18 at 14 41 54


meowjesty (approved):
👍

PR #4408, opened 2026-06-22
Add feature.magic.turbo to fix Next.js/Turbopack under mirrord

aviramha:
## Problem

Running a Next.js app with **Turbopack** under mirrord crashes when the app compiles CSS (e.g. `next dev` → `TurbopackInternalError: ... node process exited before we could connect to it with exit status: 0`).

Turbopack runs transforms (PostCSS/CSS) in **pooled Node worker subprocesses**. The parent listens on `127.0.0.1:0` and each worker connects back to that port for IPC. When the mirrord layer is loaded into the worker, its loopback `connect("127.0.0.1", <port>)` is routed through the agent (remote) by default — so the worker never reaches the local parent, its socket closes, it `process.exit(0)`s, and the build fails.

## Fix

Add a **`feature.magic.turbo`** flag (on by default, alongside the existing `magic.aws` / `magic.auto_mount`). When enabled, on layer load mirrord detects Next.js processes and keeps their loopback IPC local by turning on `feature.network.outgoing.ignore_localhost`.

Propagation: the Turbopack workers don't look like `next` themselves, and they decode the same pre-baked resolved-config blob, so we can't just mutate their config. Instead the top-level `next` process sets `MIRRORD_OUTGOING_IGNORE_LOCALHOST` in its environment; the workers inherit it; and every process applies that env to its own decoded config. This scopes the workaround to the Next.js process tree — unrelated processes are untouched even though the flag is globally enabled.

Opt out with:
\`\`\`json
{ "feature": { "magic": { "turbo": false } } }
\`\`\`

## Changes

- `mirrord/config/src/feature/magic.rs` — new `turbo: bool` (default `true`) on `MagicConfig` + analytics.
- `mirrord-schema.json` — regenerated.
- `mirrord/layer/src/turbo.rs` — detection + apply logic, with unit tests.
- `mirrord/layer/src/lib.rs` — gate behind `config.feature.magic.turbo` during pre-init.

## Tests

- `cargo test -p mirrord-layer turbo::` — 6 passing (detection heuristic).
- `cargo clippy` / `cargo fmt` clean.
- Config schema up-to-date test passes.


0x00A5 (dismissed):

0x00A5 (approved):

iniw (approved):

gememma (commented):
KA-THUNK✅

gememma (approved):

PR #4412, opened 2026-06-23
Fix macOS 26 Turbopack worker crashes (DNS segfault + EOPNOTSUPP flood)

aviramha:
## Summary

Two issues crash short-lived Node workers (e.g. Next.js / Turbopack PostCSS workers) running under mirrord on **macOS 26 (Tahoe)**. Both are addressed here.

### A) SIGSEGV in the DNS hook (`KERN_INVALID_ADDRESS at 0x4`)

`dns_configuration_copy_detour` returned **NULL** on the `Detour::Error` path: `unwrap_or_bypass_with` only routes `Bypass` through the empty-config fallback, while `Error` went through `impl From<HookError> for *mut dns_config_t`, which returns `null_mut()`. This `Error` happens when `remote_dns_configuration_copy()` can't build a config — e.g. the remote `/etc/resolv.conf` read fails, which races during the bootstrap of short-lived workers.

c-ares (`ares_init_by_sysconfig`) then called `dns_configuration_free(NULL)`, and the free detour read `.resolver` at offset 4 of a null pointer → segfault at `0x4`.

Fix:
- `dns_configuration_copy_detour` now returns the empty `dns_config_t` box on **both** `Bypass` and `Error` — never null.
- `dns_configuration_free_detour` null-checks `config` before reconstructing boxes (defensive; c-ares on macOS 26 can also pass null directly).

### B) `EOPNOTSUPP` flagged as a hard error

The layer surfaced `EOPNOTSUPP` (code 102, "Operation not supported on socket") on the worker's loopback socket as a hard error, flooding logs with `Error occured in Layer >> IO(...)`. Added `EOPNOTSUPP` / `WSAEOPNOTSUPP` to `IGNORE_ERROR_CODES` so it's treated as a soft, pass-through errno like `EADDRINUSE`/`EPERM` already are.

## Testing

- `cargo check`, `cargo clippy -- --deny warnings`, and `cargo fmt` clean on `mirrord-layer` and `mirrord-layer-lib`.

Refs COR-1583 (follow-up to the Turbopack "magic" turbo feature, which did not resolve the macOS 26 crashes).

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -1307,11 +1307,13 @@ pub(super) unsafe fn free_dns_resolver_t(resolver: *mut dns_resolver_t) {
 #[cfg(target_os = "macos")]
 #[mirrord_layer_macro::instrument(level = "trace", ret)]
 pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
+    tracing::debug!("remote_dns_configuration_copy called");
     let remote = read_remote_resolv_conf()?;
 
     // TODO: possibly create a different error rather than the almost correct
     // [`HookError::DNSNoName`]
     let resolv_conf = resolv_conf::Config::parse(remote).map_err(|_error| HookError::DNSNoName)?;
+    tracing::debug!("remote_dns_configuration_copy resolv_conf done");
 
     let options = CString::new(format!(
         "ndots:{ndots} attempts:{attempts}",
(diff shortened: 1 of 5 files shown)


mirrord/layer/src/socket/ops.rs
```diff
@@ -1307,11 +1307,13 @@ pub(super) unsafe fn free_dns_resolver_t(resolver: *mut dns_resolver_t) {
 #[cfg(target_os = "macos")]
 #[mirrord_layer_macro::instrument(level = "trace", ret)]
 pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
+    tracing::debug!("remote_dns_configuration_copy called");
```
iniw, line 1310: leftover debug logs?

aviramha: nope, on purpose in case my PR won't solve the issue.

iniw (approved):

PR #4414, opened 2026-06-23
add deferred connect to handle connection latency

meowjesty:


mirrord/intproxy/src/background_tasks.rs
```diff
@@ -25,6 +25,7 @@ pub type MessageBus<T> =
 /// parents. It allows the tasks to send and receive messages.
 pub struct MessageBusInner<MessageIn, MessageOut> {
     tx: Sender<MessageOut>,
+    self_tx: Sender<MessageIn>,
```
0x00A5: In `MessageBusInner::recv`,
```
    pub async fn recv(&mut self) -> Option<MessageIn> {
        tokio::select! {
            _ = self.tx.closed() => None,
            msg = self.rx.recv() => msg,
        }
    }
```

` msg = self.rx.recv() => msg` will return None if the channel has closed, and in the background task main loop, if msg is None, the task will quit.

If we hold a strong sender here in the message bus itself, the channel will never close. So we should hold a weak sender instead, by calling the [Sender::downgrade](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.downgrade) method. So this teardown path remains valid.

meowjesty: ee9843a107f6004a96c1458c1834e33a3d770143

0x00A5 (approved):

PR #4418, opened 2026-06-24
INT-225 Track and expose SQS queue-splitting status to users

vladrbg:
### Quality Checklist:

  documentation)

FIXED INT-400


mirrord/cli/src/config.rs
```diff
@@ -176,6 +176,11 @@ pub(super) enum Commands {
     #[command(name = "db-branches")]
     DbBranches(Box<DbBranchesArgs>),
 
+    /// Browse the status of active queue-splitting sessions.
+    #[cfg_attr(target_os = "windows", command(hide = true))]
```
Razz4780: No need to hide it on windows, nothing unix-specific going on here

mirrord/cli/src/config.rs
```diff
@@ -1158,6 +1163,23 @@ pub(super) enum DbBranchesCommand {
...
+    pub config_file: Option<PathBuf>,
+
+    #[command(subcommand)]
+    pub command: QueuesCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub(super) enum QueuesCommand {
+    /// Show the status of active queue-splitting sessions.
+    Status,
```
Razz4780: I think we need both `mirrord qs status` and `mirrord qs status <split name>`

mirrord/operator/src/crd/queue_split.rs
```diff
@@ -0,0 +1,119 @@
...
+/// aggregated API, so every request is answered live from the queue-splitting
+/// controller's in-memory state. The `CustomResource` derive is used only to get
+/// the kube `Resource` impl (group/version/kind/plural) and a `metadata`-carrying
+/// wrapper; nothing is persisted in etcd.
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[kube(
+    group = "operator.metalbear.co",
+    version = "v1",
+    kind = "QueueSplit",
+    root = "QueueSplitCrd",
```
Razz4780: `Crd` suffix is a curse from the past, we don't need it 😆

mirrord/operator/src/crd/queue_split.rs
```diff
@@ -0,0 +1,119 @@
...
+)]
+#[serde(rename_all = "camelCase")]
+pub struct QueueSplitSpec {
+    /// mirrord session id this split belongs to, in the same uppercase hex
+    /// format the CLI shows. Matches the prefix of the backing split session.
+    pub session: String,
+    /// Target workload being split.
+    pub target: QueueSplitTarget,
+    /// The developer who started the session.
+    pub owner: QueueSplitOwner,
```
Razz4780: We can use the existing types - `SessionTarget` and `SessionOwner`

Razz4780 (changes requested):
Nits nits

mirrord/cli/src/config.rs
vladrbg: Is this what you imagine, or something else for the detailed one? Also when you say split name is it `session+target` like `138a18e21bfd8f5.sqs-consumer.deployment` or is it only `session` ? 
image

Razz4780: I meant the name of the QueueSplit resource, the one we return in `.metadata.name`

mirrord/cli/src/queues.rs
```diff
@@ -0,0 +1,376 @@
...
+    // We match the resource name shown in the table's NAME column. It encodes
+    // the session and target, so it is unique across all namespaces and lets
+    // the user ask for one without knowing which namespace or cluster it is in.
+    let Some(split) = splits
+        .iter()
+        .find(|split| split.meta().name.as_deref() == Some(&name))
+    else {
+        progress.failure(Some(&format!("No queue-splitting session named '{name}'")));
+        return Ok(());
+    };
```
Razz4780, line 133: Don't list everything if we need just one resource, we should use `Api::get`

mirrord/cli/src/queues.rs
```diff
@@ -0,0 +1,376 @@
...
+        ConfigContext::default().override_env_opt(LayerConfig::FILE_PATH_ENV, args.config_file);
+    let layer_config = LayerConfig::resolve(&mut cfg_context)?;
+
+    let client = kube_client_from_layer_config(&layer_config).await?;
+
+    // The queue-splitting status view is built on the fly by the operator. We
+    // list it across all namespaces so one command shows every active session,
+    // including ones the primary aggregates from other clusters.
+    let api: Api<QueueSplit> = Api::all(client);
+    let splits = list_resource_if_defined(&api, &mut fetch_progress)
```
Razz4780: Huh, I missed this one before, sorry. We should behave like kubectl and by default list resources only from the configured namespace. List all namespaces only if `-A` is passed. I'm not sure whether we should take the namespace from mirrord.json (`.target.namespace`) or kubeconfig default tho 👀 

Also, we should accept `-n` command line argument to specify namespace.

Razz4780 (changes requested):

mirrord/cli/src/config.rs
vladrbg: So thats the one I am using yep. This is how its made. 

```
pub fn session_resource_name(
    target: &ResolvedTarget<false>,
    session_id: &FeatureSessionId,
) -> String {
    let mut resource_name = format!("{session_id}.{}.{}", target.name_any(), target.type_());
    resource_name.truncate(253);
    resource_name
}
```

Razz4780 (dismissed):

iniw (approved):

PR #4423, opened 2026-06-24
Strip instrument and tracing in fn that Go hooks call into

0x00A5:
### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/layer/macro/src/lib.rs
+++ b/mirrord/layer/macro/src/lib.rs
@@ -258,7 +258,7 @@ fn conditional_log(level: &str, input: proc_macro::TokenStream) -> proc_macro::T
     proc_macro::TokenStream::from(output)
 }
 
-/// Wrapper for [`tracing::trace!`] that is only emitted when debug assertions are enabled.
+/// Wrapper for `tracing::trace!` that is only emitted when debug assertions are enabled.
 ///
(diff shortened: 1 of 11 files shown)


mirrord/layer/macro/src/lib.rs
```diff
@@ -236,3 +236,74 @@ pub fn instrument(
...
+    conditional_log("warn", input)
+}
+
+/// Wrapper for `tracing::error!` that is only emitted when debug assertions are enabled.
+///
+/// See [`conditional_log`] for the rationale.
+///
+/// **Warning**: the `tracing` crate must be visible under the name `tracing` at the call site.
+#[proc_macro]
+pub fn error(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
meowjesty: I don't think we ever want these though.

These are log levels that should reach the user, right?

meowjesty (changes requested):
Do we really want to condition `info, warn, error` under debug_assertions? I think these levels should always be available to the user?

mirrord/layer/macro/src/lib.rs
0x00A5: stripped these

meowjesty (approved):
👍

PR #4424, opened 2026-06-25
Tweak post 125 go stack switch

0x00A5:
### Problem
Go programs run under mirrord intermittently SIGSEGV during init — in the GC
sweeper, a map iterator, or other unrelated heap-walking code (`addr=0x0`). The
crashes are non-deterministic surfaced on go1.26.4 in customer CI.

### Hypothesis
mirrord hooks `internal/runtime/syscall.Syscall6` (the raw syscall
leaf) and switches to the `g0` stack to run our handler. That switch was a port of
`runtime.systemstack`, whose return path restores `rsp`/`rbp` from `g.sched` and
then **zeroes `g.sched.sp`/`g.sched.bp`**.

`systemstack` is only legal on a `_Grunning` goroutine. But when the leaf is
reached via `syscall.Syscall6`, the goroutine is `_Gsyscall` — `entersyscall`
already saved its resume continuation into `g.sched`. We were stomping state the
scheduler co-owns, leaving `g.sched.sp == 0` in the window before `exitsyscall`
runs. The goroutine self-heals on its own exit path, so the damage only lands when
a concurrent actor on another core (scheduler P-stealing / `forEachP`) reads that
state mid-window.

### Fix
Switch the `g0` stack-switch to mirror `runtime.asmcgocall` instead of
`runtime.systemstack`. On return `asmcgocall` **never touches `g.sched`** — it restores `rsp` from a
`stack.hi - sp` depth saved on the (non-moving) `g0` stack. The only remaining
write to `g.sched` is the synthetic `gosave_systemstack_switch` frame, exactly as
`asmcgocall` does.

This is also correct for the `RawSyscall6` (`_Grunning`) path: `asmcgocall` is the
strictly more general switch (makes no assumption about who owns `g.sched`), so a
single implementation covers both callers.

### Scope

`post_go_1_25` path, x86-64 Linux only.

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/layer/src/go/linux_x64.rs
+++ b/mirrord/layer/src/go/linux_x64.rs
@@ -611,14 +611,14 @@ fn post_go1_23(hook_manager: &mut HookManager, module_name: Option<&str>) {
 /// Refer:
 ///   - File zsyscall_linux_amd64.go generated using mksyscall.pl.
 ///   - <https://cs.opensource.google/go/go/+/refs/tags/go1.18.5:src/syscall/syscall_unix.go>
-pub(crate) fn enable_hooks(hook_manager: &mut HookManager) {
+pub(crate) fn enable_hooks(hook_manager: &mut HookManager, cgo_stack_switch: bool) {
     let Some(version) = super::get_go_runtime_version(hook_manager) else {
         return;
     };
 
     tracing::trace!(version, "Detected Go");
     if version >= 1.25 {
-        post_go_1_25::hook(hook_manager, version);
+        post_go_1_25::hook(hook_manager, version, cgo_stack_switch);
     } else if version >= 1.23 {
         post_go1_23(hook_manager, None);
     } else if version >= 1.19 {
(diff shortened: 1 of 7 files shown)


mirrord/layer/src/go/linux_x64.rs
```diff
@@ -854,41 +878,57 @@ mod post_go_1_25 {
...
+            // Switch back to curg WITHOUT writing g.sched (the key difference from `systemstack`).
+            // r14 currently holds g0; reach curg through m.
             "mov    rbx,[r14+0x30]",
             "mov    rsi,[rbx+0xb8]",
             "mov    fs:0xfffffffffffffff8,rsi",
             "mov    r14,rsi",
-            "mov    rsp,[rsi+0x38]",
-            "mov    rbp,[rsi+0x60]",
-            "mov    QWORD PTR [rsi+0x38],0x0",
-            "mov    QWORD PTR [rsi+0x60],0x0",
```
0x00A5, line 887: `runtime.systemstack` clears `bp` and `sp` in `g.sched`, but CGO doesn't.

aviramha (dismissed):
LGTM. I'd suggest to duplicate code so we can hide the new implementation behind experimental flag.

aviramha (approved):

PR #4425, opened 2026-06-25
fix(ui): recover token via /api/token so the browser extension always syncs

aviramha:
## Problem

When the browser extension shows *"mirrord ui is up, but the extension doesn't have its token yet — Open the mirrord ui page to configure it automatically"*, clicking **Open mirrord ui** opens the bare origin `http://127.0.0.1:59281` in a new tab, **without** a `?token=` query param (the extension doesn't have the token — that's the whole point). The same happens on any reload.

`autoConfigureExtension()` only forwarded the token to the extension when it could read it from the URL. On a bare-origin load it had no token to forward, because:

- the `mirrord_token` cookie is **`HttpOnly`** → JS can't read it;
- `sessionStorage` is **per-tab** → a freshly opened tab has nothing;
- the URL has no `?token=`.

So the page authenticated fine (via the cookie, sent automatically by the browser) but the extension was never configured. This is why it *"loaded with auth but didn't sync."*

## Fix

Add an authenticated **`GET /api/token`** endpoint that echoes the token back to a request the browser has already authenticated (cookie / `?token=` / `x-auth-token` — gated by the same `token_auth` middleware as every other API route). The frontend's `autoConfigureExtension()` now:

1. uses the `?token=` query param when present (canonical first visit — no extra request); otherwise
2. fetches `/api/token` (the browser attaches the cookie) to recover the token,

then forwards it to the extension. This covers the bare-origin button, reloads, and new tabs.

### Security note

`/api/token` is behind the auth middleware, so an unauthenticated caller can't lift the token. It does expose the token to same-origin JS that already holds the cookie — consistent with the existing design where the token already lives in JS-readable `sessionStorage`/URL during the normal flow. The `HttpOnly`/`SameSite=Strict` cookie protections are unchanged.

## Testing

- New Rust tests: `/api/token` returns 401 without auth, and returns the token when authenticated via cookie. Full `ui::tests` suite passes (20/20).
- `cargo check` (CLI), `tsc --noEmit` + `pnpm build` (monitor) all pass; prettier clean.
- Manual: `mirrord ui --token <t>`, open printed URL once (sets cookie), then click the extension's "Open mirrord ui" button → token now syncs.


iniw (approved):
Looks good 2 meeeeeeeeeeeeeee

PR #4428, opened 2026-06-26
INT-460 Support multiple split queues with the same key

vladrbg:
### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/config/src/feature/split_queues.rs
+++ b/mirrord/config/src/feature/split_queues.rs
@@ -250,6 +250,33 @@ impl SplitQueuesConfig {
         })
     }
 
+    pub fn bullmq(&self) -> impl Iterator<Item = (&str, &QueueMessageFilter)> {
+        self.0.iter().filter_map(|split| match &split.filter {
+            QueueFilter::BullMq {
+                message_filter: Some(message_filter),
+                ..
+            } => Some((split.queue_id.as_str(), message_filter)),
+            _ => None,
+        })
+    }
+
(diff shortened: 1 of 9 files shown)


mirrord/config/src/feature/split_queues.rs
```diff
@@ -1,213 +1,251 @@
...
+/// the filter (instead of using it as a map key) is what lets the same id show up more than once,
+/// which a map cannot do.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub struct QueueSplit {
+    /// The id of the queue to split. Does not have to be unique across entries.
+    pub queue_id: QueueId,
+
+    /// The filter for this queue, tagged by its `queue_type`.
+    #[serde(flatten)]
+    pub filter: QueueFilter,
```
skreborn, line 78: `QueueFilter` has `#[serde(deny_unknown_fields)]`, but [serde silently drops that](https://serde.rs/container-attrs.html#deny_unknown_fields) on a `#[serde(flatten)]` field, so the list form accepts unknown fields that the map form correctly rejects.

The only way I would know how to fix this is with a custom `Deserialize` implementation, but whether that's worth the validation is questionable, so I defer to you in that choice.

skreborn (approved):

mirrord/config/src/feature/split_queues.rs
vladrbg: No sure I like custom deserializer, but the thing is after this change there will be bigger rework on the messaging part, so maybe we do it there, not sure how to approach it yet.

PR #4429, opened 2026-06-26
COR-1614 Chaos latency delayed queue

0x00A5:
### Problem
1. Chaos latency is cumulative.
2. When outgoing proxy to interceptor queue background task's channel is full, `send()` blocks, thus, blocking outgoing proxy's main event loop.

### Solution
Record each message's deadline at enqueue time. Use a delayed queue to keep insertion order, while honoring each message's deadline before releasing.

The current solution uses unbounded delayed queue. To handle layer-to-agent message back pressure, we can use a semaphore backed byte permits shared by interceptor background task and queue background task, so that when delayed queue capacity is full, stop reading from local socket connection.

But that can be a separate PR.

### Quality Checklist:

  documentation)

### DIFF
--- a/changelog.d/+chaos-latency.changed.md
+++ b/changelog.d/+chaos-latency.changed.md
@@ -1,0 +1,2 @@
+Updated the chaos outgoing latency implementation so that busy connection's latency
+will not accumulate.
--- a/changelog.d/+chaos-latency.changed.md
+++ b/changelog.d/+chaos-latency.changed.md
@@ -1,2 +1,3 @@
 Updated the chaos outgoing latency implementation so that busy connection's latency
-will not accumulate.
+will not accumulate. Latency chaos rules will now add the specified latency amount
+more accurately.
--- a/mirrord/intproxy/src/proxies/outgoing/interceptor.rs
+++ b/mirrord/intproxy/src/proxies/outgoing/interceptor.rs
@@ -1,6 +1,7 @@
 //! [`BackgroundTask`] used by [`OutgoingProxy`](super::OutgoingProxy) to manage a single
 //! intercepted connection.
 
+mod delay_queue;
 pub(super) mod read_queue;
 pub(super) mod write_queue;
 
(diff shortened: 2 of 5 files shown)


mirrord/intproxy/src/proxies/outgoing/interceptor.rs
```diff
@@ -1,6 +1,7 @@
 //! [`BackgroundTask`] used by [`OutgoingProxy`](super::OutgoingProxy) to manage a single
 //! intercepted connection.
 
+mod delay_queue;
```
gememma, line 4: `interceptor.rs` and the 3 `*_queue.rs` files are all pretty small, and it seems like there's a lot of similar code between the read and write queues - especially the `new()` functions where you made basically the same changes twice  👀 can we condense all this?

mirrord/intproxy/src/proxies/outgoing/interceptor/delay_queue.rs
gememma, line 1: DelayQueue is very cool B)

changelog.d/+chaos-latency.changed.md
```diff
@@ -0,0 +1,2 @@
+Updated the chaos outgoing latency implementation so that busy connection's latency
+will not accumulate.
```
gememma: nit: maybe something more understandable from user POV 👀 ?

```suggestion
Updated the chaos outgoing latency implementation so that busy connection's latency
will not accumulate. Latency chaos rules will now add the specified latency amount
more accurately.
```

gememma (dismissed):
Nice! The `deadline` approach instead of `sleep(delay)` is v cool 🚜 a couple of minor/ low-impact comments

meowjesty (dismissed):
nice 👍

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
0x00A5: Yeah there are quite some duplicated code right now. I have a followup PR that will only make change to one of the queues. Once finished, I will see if we can unify them a bit.

changelog.d/+chaos-latency.changed.md
0x00A5: Updated with your suggestion 👍

meowjesty (approved):
🔁

PR #4441, opened 2026-06-29
Fix use-after-free in macOS getifaddrs IPv6-hiding hook

aviramha:
## What

Fixes the failing layer test `test_issue2807_with_ipv6_ignore` (`issue2807.rs`), which was failing on macOS CI with:

```
Error: getaddrinfo ENOTFOUND <unknown sa family>
```

## Root cause

`getifaddrs` returns a list that lives in a **single allocation** rooted at the head pointer; each `ifaddrs` node's `ifa_addr` / `ifa_name` / `ifa_netmask` / … pointers point *into that same block*, and `freeifaddrs` reclaims the whole thing with one `free` of the head.

The IPv6-hiding hook (used when `experimental.hide_ipv6_interfaces` is set) introduced in #2905 shallow-copied each kept node into a freshly `malloc`'d array and then called `freeifaddrs(original_head)`. The copies' inner pointers still referenced the now-freed original block — a use-after-free. Applications reading the dangling `ifa_addr->sa_family` saw garbage; node (via `portfinder`) surfaced it as `getaddrinfo ENOTFOUND <unknown sa family>`. It only reproduced when the freed memory was reused before the app read it, which is why it showed up as a flaky CI failure.

## Fix

Hide IPv6 interfaces by re-linking `ifa_next` **in place** so the IPv6 nodes are no longer reachable while still physically living inside the original allocation — the user's eventual `freeifaddrs` reclaims everything at once. No new allocation, no intermediate free, so no dangling pointers. When the first entry is IPv6, the first kept entry's contents are copied over the allocation base so the returned head stays at the base address that `freeifaddrs` needs.

## Testing

- `cargo check -p mirrord-layer` — clean.
- Ran `cargo xtask test-integration test_issue2807` locally with the fix: the `getaddrinfo ENOTFOUND <unknown sa family>` failure no longer occurs. (A separate local run hit an unrelated harness issue where xtask passed a *relative* mirrord binary path that the test's CWD couldn't resolve; that's not related to this change.) Recommend confirming a green run in CI.

### DIFF
--- a/mirrord/layer/src/socket/ops.rs
+++ b/mirrord/layer/src/socket/ops.rs
@@ -1,5 +1,5 @@
 use alloc::ffi::CString;
-use core::{ffi::CStr, mem};
+use core::ffi::CStr;
 #[cfg(target_os = "macos")]
 use std::os::fd::BorrowedFd;
 use std::{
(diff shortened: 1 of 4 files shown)


MintSoup (approved):

mirrord/layer/src/socket/ops.rs
```diff
@@ -1358,57 +1358,40 @@ pub(super) fn remote_dns_configuration_copy() -> Detour<*mut dns_config_t> {
 }
 
 /// Calls [`libc::getifaddrs`] and removes IPv6 addresses from the list.
+///
+/// `getifaddrs` returns a list that lives in a single allocation rooted at the head pointer, and
```
MintSoup, line 1362: I just checked and while this seems to be true on glibc and darwin, I don't think we should rely on this as it's an internal impl detail.
I think better to recreate the list, just correctly.

MintSoup (changes requested):

MintSoup (dismissed):

0x00A5 (approved):

PR #4457, opened 2026-07-01
Ignore flaky test_issue2807_with_ipv6_ignore layer test

aviramha:
## What

Marks the `test_issue2807_with_ipv6_ignore` layer test (`application_1_Application__NodeIssue2807`) as `#[ignore]` so it stops running in CI.

## Why

The test is flaky — it intermittently panics at `test-utils/src/lib.rs:96`. Disabling it until we fix the underlying flakiness so it stops blocking CI.

The test remains in the tree and can still be run explicitly with `cargo test -- --ignored`.


MintSoup (approved):

gememma (approved):
🍨

PR #4459, opened 2026-07-01
Implement Flyway migrations

skreborn:
Resolves INT-478.

See https://github.com/metalbear-co/operator/pull/1883 for implementation details.

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #4470 
- <kbd>&nbsp;1&nbsp;</kbd> #4459 👈

### DIFF
--- a/mirrord/operator/src/client.rs
+++ b/mirrord/operator/src/client.rs
@@ -44,8 +44,8 @@ use crate::{
         database_branches::{
             DatabaseBranchParams, UnifiedDatabaseBranchParams, create_branches,
             create_mongodb_branches, create_mysql_branches, create_pg_branches,
-            list_existing_branches, list_reusable_mongodb_branches, list_reusable_mysql_branches,
-            list_reusable_pg_branches, wait_for_pending_branches,
+            ensure_branch_migrations, list_existing_branches, list_reusable_mongodb_branches,
+            list_reusable_mysql_branches, list_reusable_pg_branches, wait_for_pending_branches,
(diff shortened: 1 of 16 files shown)


mirrord/operator/src/client.rs
```diff
@@ -741,6 +766,21 @@ where
...
+            // re-run the tool (which no-ops, applies the delta, or fails on a conflict); an
+            // unchanged archive is a no-op patch and returns at once.
+            for (id, branch) in existing
+                .ready
+                .iter()
+                .chain(waited_branches.iter())
+                .chain(created_branches.iter())
+            {
+                if let Some(migrations) = desired_migrations.get(id) {
+                    ensure_branch_migrations(&branch_api, branch, migrations, timeout, &subtask)
```
vladrbg, line 779: What happens if we have 2 sessions with same id but have different migrations? They will probably override eachother or?

skreborn: There should be no possibility of corruption. `spec.migrations` is last-writer-wins: the second session's archive bumps the CR generation and the operator runs a fresh generation-keyed job. Flyway is the source of truth via an internal table and its own locking mechanism.

So to directly answer your question: there's no such thing as "same-time" in the real world: they'll be ordered one way or another by Flyway—and if the migrations are incompatible, one of them will get a proper error back.

vladrbg (approved):

PR #4460, opened 2026-07-01
Change mirrord UI to run as background process

gememma:
This PR changes the UI server spawned by running `mirrord ui` to run in the background, detached from the terminal where the command was invoked. This is a UX improvement that will let users leave the server running on their machines for long periods of time.

NB: I use the term "UI server" to refer to the `axum` server that serves the local mirrord UI frontend (also called the dashboard in some places), and not the session monitor server that is started by the intproxy process during a single mirrord session, which is internal and not accessible by the user, but which communicates with the UI server.

### Overview

Old behaviour: The user runs `mirrord ui` to start the server, and can then view the frontend to see session details or use the chaos endpoints to manage chaos rules. The server blocks the terminal where it was started, and runs until the command is cancelled.

New behaviour: The user runs `mirrord ui` to start the server. The terminal is only blocked while the server is being set up, and then the command returns. The server runs for as long as the background task/ process lives. It can be stopped by running `mirrord ui stop` or by killing the process manually by PID (which is printed to the user during startup).

### How does it do that though

1. `ui_start()` decides if the current process is foreground or background by looking for a special env var. The first time it runs, it's the foreground task that the user is seeing in their terminal, and the env var is not present
2. `tokio`'s `Command::spawn()` is used to spawn a child process running `mirrord ui` with the special env var present. `stderr` is inherited, so errors from the server setup will be visible in the terminal
3. The child process finds the env var present, and runs `ui_run_server()` - the first thing it attempts to do is lock the ui.lock file, thus making sure it is the only UI server running on the machine. If it fails, it exits with `Ok` and informs the parent process that a server is already running via `stdout`
4. The child process sets up the server and prints an OK message before starting the server itself
5. The parent process sees that the server started successfully and prints details to the user: the token, the server url and the PID of the server
6. If the server _did not start properly_ and `ui_start()` returns an error, `ui_stop()` is run to clean up
7. After the user is done with the server and runs `mirrord ui stop`, `ui_stop()` attempts to kill the process by its PID (stored in a file next to the token and lockfile) and remove the stale PID & token files

### This PR

I recommend you review this PR either commit by commit (I know there are a few, and I overwrite previous changes a lot) **or** you exclude the final few commits (so you view everything before `meta: move server code to ui::server`) to review the UI server changes first. This is because the commit marked `meta` moves a lot of code into a new file, which makes the git diff very difficult to read.

### DIFF
--- a/changelog.d/+ui-bg-process.changed.md
+++ b/changelog.d/+ui-bg-process.changed.md
@@ -1,2 +1,2 @@
-The `mirror ui` command now starts the local UI server as a separate process that runs in the background on your machine.
+The `mirrord ui` command now starts the local UI server as a separate process that runs in the background on your machine.
 To stop the server, run `mirrord ui stop`.
--- a/changelog.d/+ui-bg-process.changed.md
+++ b/changelog.d/+ui-bg-process.changed.md
@@ -1,0 +1,2 @@
+The `mirror ui` command now starts the local UI server as a separate process that runs in the background on your machine.
+To stop the server, run `mirrord ui stop`.
--- a/mirrord/cli/src/ui.rs
+++ b/mirrord/cli/src/ui.rs
@@ -6,10 +6,11 @@
 //! transport (Unix domain socket or named pipe), and serves a React frontend plus
 //! REST/SSE/WebSocket endpoints on localhost.
 
+#[cfg(unix)]
+use std::num::ParseIntError;
 use std::{
     fs::File,
     net::{Ipv4Addr, SocketAddr},
-    num::ParseIntError,
     path::PathBuf,
     process::Stdio,
     str::FromStr,
@@ -19,6 +20,7 @@ use std::{
 use fs4::fs_std::FileExt;
 use miette::Diagnostic;
 use mirrord_session_monitor_client::sessions_dir;
+#[cfg(unix)]
 use nix::{
     errno::Errno,
     sys::signal::{Signal, kill},
@@ -66,8 +68,8 @@ const PID_FILE_NAME: &str = "server_pid";
 const TOKEN_FILE_NAME: &str = "token";
 
 /// The header key that can be used to set the auth token in requests to the ui server instead of
-/// using the `token` query parameter. It can also be set in a cookie. See the [`token_auth()`]
-/// middleware.
+/// using the `token` query parameter. It can also be set in a cookie. See the
+/// `ui::server::token_auth()` middleware.
 const TOKEN_HEADER_NAME: &str = "x-auth-token";
 
 /// The name of the env var containing the port that the ui server should run on. If present in env,
@@ -88,6 +90,7 @@ pub enum UiCliError {
     Io(#[from] std::io::Error),
 
     /// May occur while trying to kill the existing UI server.
+    #[cfg(unix)]
     #[error("failed to perform an operation on the UI server process: {0}")]
     #[diagnostic(help(
         "To forcefully stop the server process, try killing it manually. On \
@@ -106,6 +109,7 @@ pub enum UiCliError {
     #[error("the new server process ended unexpectedly")]
     ChildExitedUnexpectedly,
 
+    #[cfg(unix)]
     #[error("failed to parse a PID from file contents: {0}")]
     PidParse(ParseIntError),
 }
--- a/mirrord/cli/src/ui.rs
+++ b/mirrord/cli/src/ui.rs
@@ -65,11 +65,37 @@ mod chaos;
 
 const MAX_EVENTS_PER_SESSION: usize = 500;
 
+/// The name of the file that is locked by the currently running ui server. If a new server attempts
+/// to start running, it will fail to lock this fail and exit.
+///
+/// Only ever touched by the *child* process that runs the server, not the parent process running in
+/// the foreground.
 const UI_LOCK_FILE_NAME: &str = "ui.lock";
+
+/// The name of the file where we store the PID of the most recently started ui server. We use this
+/// to be able to kill the currently running server in a standalone command. If there is no lock on
+/// [`UI_LOCK_FILE_NAME`], no server is running: this file is stale.
+///
+/// Only ever written, read or deleted by the parent (foreground) process, not the child process
+/// that runs the ui server. Is not locked.
 const PID_FILE_NAME: &str = "server_pid";
+
+/// The name of the file containing the current auth token for the ui server.
+///
+/// Is written to by the child process (running the server) when it gets the file lock for
+/// [`UI_LOCK_FILE_NAME`]. Is read by the child process for authentication, and the parent
+/// (foreground) process when printing the ui server details to the user. Is not locked.
 const TOKEN_FILE_NAME: &str = "token";
+
+/// The header key that can be used to set the auth token in requests to the ui server instead of
+/// using the `token` query parameter. It can also be set in a cookie. See the [`token_auth()`]
+/// middleware.
 const TOKEN_HEADER_NAME: &str = "x-auth-token";
 
+/// The name of the env var containing the port that the ui server should run on. If present in env,
+/// indicated to the current invokation of `mirrord ui` that this process is a child process and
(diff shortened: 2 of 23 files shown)


mirrord/cli/src/ui.rs
```diff
@@ -1098,414 +256,308 @@ pub async fn setup_ui(port: u16) -> Result<UiSetup, CliError> {
...
+/// $PID`.
+pub async fn ui_start(port: u16, no_browser: bool) -> Result<(), UiCliError> {
+    let mirrord_binary = std::env::current_exe()?;
+
+    let mut child = tokio::process::Command::new(mirrord_binary)
+        .args(vec!["ui"])
+        .envs(vec![(MIRRORD_SERVER_PORT_ENV_NAME, port.to_string())])
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::inherit())
```
Razz4780: Not sure about this. What happens if the parent process exits and the child keeps using inherited stderr? 👀

gememma: it's the same as if its piped and we read it in the parent process, right? like itll just go nowhere? its the default mode for stdio on a spawned command

eventually I want to pipe all the logs into a file like the intproxy does but the code is extremely tied into the intproxy itself, it would take way longer to do that in this PR

gememma: can you think of an easier alternative?

mirrord/cli/src/ui.rs
```diff
@@ -6,923 +6,130 @@
...
-}
-
-#[derive(Serialize)]
-struct TokenResponse {
-    token: String,
-}
-
-/// Returns the auth token to a request that already authenticated.
+/// The name of the file that is locked by the currently running ui server. If a new server attempts
+/// to start running, it will fail to lock this fail and exit.
```
meowjesty:
```suggestion
/// to start running, it will fail to lock this file and exit.
```

mirrord/cli/src/ui.rs
```diff
@@ -953,131 +168,74 @@ enum TokenClaim {
...
-fn claim_token_file_at(lock_path: PathBuf, token_path: PathBuf) -> Result<TokenClaim, CliError> {
-    if let Some(parent) = lock_path.parent() {
-        std::fs::create_dir_all(parent)
-            .map_err(|e| CliError::UiError(format!("failed to create mirrord directory: {e}")))?;
-    }
+impl TokenClaim {
+    /// Tries to become the single running `mirrord ui` instance by taking an exclusive lock on the
+    /// lock file. If another instance already holds it, reads back the token it published.
+    pub fn claim_token_file() -> Result<TokenClaim, std::io::Error> {
+        // ensure ~/.mirrord exists
```
meowjesty, line 182: Hate when apps put random, temporary things in my home dir, why not use tmp?

meowjesty (dismissed):
I think we should use `/tmp/mirrord` instead of `home` for this lock file and this ui stuff.

Also, it should probably be configurable by the user too.

meowjesty:
> foreground or background by looking for a special env var.

What happens if I run `mirrord ui` in one tab from my terminal, and then try it in another tab? The 2nd tab doesn't have the env var (unless it's been spawned after the `mirrord ui` command ran in the 1st tab, and this 2nd tab is spawning from the 1st tab).

Can't we just do a file approach here to detect if the ui server is already running? Write a file to `/tmp/mirrord/ui-server.json` with all the information about the ui server, when the user runs `mirrord ui` we check if this file exists, and create it if there's no file.

Example file:

```json
{
 api: http://127.0.0.1:59281
 token: 34a60502e328f2dd8ad218b36e0c80b453e358432b8e8de2d524194facf92f22
}
```

Send a request to this address to detect if the ui server is actually running (to avoid problems where some crash leaves this file up, but the server is down).

meowjesty:
Also, can we add support for piping std to a file? The only thing I want to see when running `mirrord ui` is that it started/is working and the link I have to click to access it. Error logs will be more useful for us, not sure if the users want to see ui server logs.

mirrord/cli/src/ui.rs
gememma: fine i guess 🤷 it was like this when i got here - i assume its because the sessions dir is also in .mirrord

meowjesty: If it's a file that will always be there, and we keep changing it, I'm fine with it living in home, like the user data thingy.

gememma:
> > foreground or background by looking for a special env var.
> 
> What happens if I run `mirrord ui` in one tab from my terminal, and then try it in another tab? The 2nd tab doesn't have the env var (unless it's been spawned after the `mirrord ui` command ran in the 1st tab, and this 2nd tab is spawning from the 1st tab).
> 
> Can't we just do a file approach here to detect if the ui server is already running? Write a file to `/tmp/mirrord/ui-server.json` with all the information about the ui server, when the user runs `mirrord ui` we check if this file exists, and create it if there's no file.
> 
> Example file:
> 
> ```json
> {
>  api: http://127.0.0.1:59281
>  token: 34a60502e328f2dd8ad218b36e0c80b453e358432b8e8de2d524194facf92f22
> }
> ```
> 
> Send a request to this address to detect if the ui server is actually running (to avoid problems where some crash leaves this file up, but the server is down).

uh... if you run mirrord ui in one tab, it spawns its bg process and starts - then when you run it in a different tab, it acts the exact same (but finds a server is already running because the lock file is locked and then exits) I dont really know what youre asking/ suggesting here? the env var is set by the `Command::spawn()` call

as for the file approach thing, something could accidentally delete or overwrite the file (which leaves a server running that mirrord wont detect) and we cant lock the file but still allow it to be readable because windows ([thread](https://metalbear.slack.com/archives/C08TYT624SV/p1781529323845119))

I agree the whole file system thing is kind of grim, but 1. it works as far as i can tell and 2. I don't reeeaaaally want to change the format of the token file if it'll break for the users who are reading it in scripts (as i assume a specific customer, you know the one, are right now) but I think it might be more acceptable to merge the pid and token files into a json like you suggested, especially if we're moving the files anyway which will _also_ break scripts that read their values

gememma:
> Also, can we add support for piping std to a file? The only thing I want to see when running `mirrord ui` is that it started/is working and the link I have to click to access it. Error logs will be more useful for us, not sure if the users want to see ui server logs.

like I said https://github.com/metalbear-co/mirrord/pull/4460#discussion_r3513926669 , this is too much time/ effort to go in this PR imo - but I agree that it should do that

mirrord/cli/src/ui.rs
meowjesty: Can't you just copy what we do for mirrord ci?

```
let mut child = match tokio::process::Command::new(binary_path)
            .args(binary_args.iter().skip(1))
            .envs(env_vars)
            .stdin(Stdio::null())
            .stdout(File::create(ci_run_output_dir.join("stdout"))?)
            .stderr(File::create(ci_run_output_dir.join("stderr"))?)
            .kill_on_drop(false)
            .spawn()
        {
```

And the stuff for creating the file/dirs.

meowjesty:
> then when you run it in a different tab, it acts the exact same (but finds a server is already running because the lock file is locked and then exits

Ah ok, I misunderstood it thinking that his env var was mixed in how we decide if some ui server is already running or not.

mirrord/cli/src/ui.rs
gememma: nice, will do that for `stderr` - `stdout` has to stay as it is though 👀 nothing interesting for the user gets printed there anyway

changelog.d/+ui-bg-process.changed.md
```diff
@@ -0,0 +1,2 @@
+The `mirror ui` command now starts the local UI server as a separate process that runs in the background on your machine.
```
hank-metalbear: @gememma should this be `mirrord ui` ?

gememma: good catch ☠️ now to wait for CI again T_T

hank-metalbear (approved):

PR #4462, opened 2026-07-01
Add chaos connection error for already established connections.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1623/error-injection-is-not-applied-to-in-flight-connections

### Quality Checklist:

  documentation)


mirrord/intproxy/src/proxies/outgoing/interceptor/read_queue.rs
0x00A5: The name of these two functions `queue_interceptor_message` and `queue_interceptor_command` is getting a lil bit confusing. Maybe we can just rename this one to `queue_interceptor_data`, or maybe even remove this function, and use `queue_interceptor_command` directly. At the call site:
```rust
queue_interceptor_command(
    id,
    InterceptorCommand::Data(bytes),
    delay
)
```

WDYT?

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -826,6 +850,35 @@ impl BackgroundTask for OutgoingProxy {
...
+                                    let close_msg = id.protocol.wrap_agent_close(id.connection_id);
+                                    self.finish_agent_write_queue(id, close_msg, delay).await;
+                                    self.finish_interceptor_read_queue(id, InterceptorCommand::Reset, delay).await;
+                                }
+                                ConnectionErrorType::TimedOut => {
+                                    self.interceptor_connection_info.remove(&id);
+                                    self.abort_agent_write_queue(&id);
+                                    self.queue_interceptor_command(id, InterceptorCommand::Stall, delay).await;
+                                }
+                                ConnectionErrorType::Refused | ConnectionErrorType::Unknown(_) => {
```
0x00A5: What is `ConnectionErrorType::Unknown` for? I looked it up and it is not used anywhere, shall we remove? `ConnectionErrorType` will even become `Copy`  afterwards.

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
```diff
@@ -142,6 +169,20 @@ impl BackgroundTask for Interceptor {
...
+                        connected_socket.shutdown().await?;
+                    }
+                    Some(InterceptorCommand::Reset) => {
+                        tracing::trace!("Resetting connection with layer");
+                        connected_socket.reset()?;
+                        break Ok(());
+                    }
+                    Some(InterceptorCommand::Stall) => {
+                        tracing::trace!("Stalling connection with layer");
+                        message_bus.closed_token().cancelled().await;
```
0x00A5: LOL, this is pure evil. Deadlock the select loop, so it simulates a timeout?

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
```diff
@@ -142,6 +169,20 @@ impl BackgroundTask for Interceptor {
                             connected_socket.send(&bytes).await?;
                         }
                     }
+                    Some(InterceptorCommand::Shutdown) => {
+                        tracing::trace!("Agent shutdown, shutting down connection with layer");
+                        connected_socket.shutdown().await?;
+                    }
+                    Some(InterceptorCommand::Reset) => {
+                        tracing::trace!("Resetting connection with layer");
+                        connected_socket.reset()?;
```
0x00A5, line 185: nit: Can we leave a comment here that explains, after calling reset, dropping the connection will send RST? Or explicitly call `drop(connected_socket)`

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
0x00A5: Given some more thoughts, i think this approach is a bit dangerous. If we are not careful and keep sending message to the interceptor, once the channel fills up, the sender will be blocked, and we may have a deadlock. Right now we are keeping track of `chaos_blocked_interceptors`, and avoid sending everywhere, but let's not rely on this.
I think a safer option is to have a `chaos_stalled` flag, similar to the existing `reading_closed`, and guard the other paths by checking the flag, while keep `message_bus.recv()` arm open, and ignore new data received from message bus.
This also opens the opportunity for us if we ever, in the future, want to make it possible to "resume" a stalled interceptor.

meowjesty: bro the ai has no heart, we're so screwed

mirrord/intproxy/src/proxies/outgoing/interceptor/read_queue.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4462/commits/3e9f2eed3c2001d8730999e87bdbebdd4b2ca8a8

mirrord/intproxy/src/proxies/outgoing.rs
meowjesty: Hmm it's for backwards compat if we ever need to add new errors, but this thing gets updated together with mirrord, so I guess there's not really a point for that?

I'll drop it.

meowjesty: https://github.com/metalbear-co/mirrord/pull/4462/commits/89d1ee2a4b99e06a437947ae48e932596f6b95f1

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4462/commits/a127e2b4348a2033162467aa8925c09ff75cbc8c

mirrord/intproxy/src/proxies/outgoing/interceptor.rs
meowjesty: https://github.com/metalbear-co/mirrord/pull/4462/commits/6f2afee6d6029d577cee623517cf0fdb4b9f35ea

0x00A5 (dismissed):

0x00A5 (approved):

PR #4463, opened 2026-07-02
fix(ui): send flat SessionInfo in ws session_added notifications

hank-metalbear:
## Summary

The `mirrord ui` WebSocket `session_added` payload wrapped `SessionInfo` inside `TrackedSession`, serializing as `{"session": {"info": {...}}}`, while the frontend reads `msg.session` as a flat `SessionInfo` (the same shape `GET /api/sessions` returns).

As a result, every session delivered over the WebSocket (live adds and the per-connect snapshot) rendered as a phantom card: blank target, "NaNs" uptime, grouped under "No session key". It could not be killed or removed (its `session_id` was `undefined`, so `session_removed` never matched), and clicking it crashed the page. The crash comes from `JsonHighlight`: `JSON.stringify(undefined)` returns `undefined`, then `.matchAll` throws. This matches the `ui_crashed` telemetry in PostHog (`Cannot read properties of undefined (reading 'matchAll')`).

Changes:
- `SessionNotification::SessionAdded` now carries `Box<SessionInfo>` (both send sites), matching the REST shape and the frontend `WsMessage` type.
- `TrackedSession` no longer derives `Serialize` (nothing serializes it anymore).
- Frontend hardening so a stale CLI can never crash the page again: `session_added` messages without a `session_id` are ignored, `JsonHighlight` tolerates `undefined`, and `formatUptime` renders an em dash for unparseable timestamps.

## Test plan

- New unit test `session_added_notification_serializes_flat_session_info` asserts the wire shape (written first against the old code, failed with `session_id` nested under `info`, passes after the fix).
- `cargo test -p mirrord`, `cargo fmt`, `cargo clippy` clean; `tsc && vite build` clean for `packages/monitor`.
- Manual end-to-end: built the CLI, started `mirrord ui`, opened the page, then created a fake session socket serving `/info` so the session arrives via the WebSocket path. The card renders with real target, key group, and uptime; the detail view opens with Events and Config; no crash. Kill and `session_removed` now match by real `session_id`.

### DIFF
--- a/changelog.d/+ui-ws-session-shape.fixed.md
+++ b/changelog.d/+ui-ws-session-shape.fixed.md
@@ -1,0 +1,1 @@
+Fixed a phantom session in `mirrord ui` showing an empty target and "NaNs" uptime that could not be removed and crashed the page when clicked. The WebSocket `session_added` payload now carries the same flat session shape as `GET /api/sessions`.
--- a/mirrord/cli/src/ui.rs
+++ b/mirrord/cli/src/ui.rs
@@ -67,7 +67,7 @@ struct FrontendAssets;
 #[serde(tag = "type", rename_all = "snake_case")]
 enum SessionNotification {
     SessionAdded {
-        session: Box<TrackedSession>,
+        session: Box<SessionInfo>,
     },
     SessionRemoved {
         session_id: String,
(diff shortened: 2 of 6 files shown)


hank-metalbear:
Screenshot 2026-07-02 at 9 50 56 AM
Screenshot 2026-07-02 at 9 51 26 AM
Attaching customer screenshot here for context

0x00A5 (dismissed):

0x00A5 (dismissed):

0x00A5 (approved):

PR #4464, opened 2026-07-02
feat(intproxy): session monitor HTTP capture + body-limit config (PRO-174) (1/2)

hank-metalbear:
## Summary

Part 1 of 2 for PRO-174. Adds HTTP capture to the `mirrord ui` session monitor at the intproxy layer, so the event stream carries enough detail to inspect (and, in Part 2, export) a session's traffic. The UI detail view and export land in #4493 (stacked on this).

**What this adds to the session monitor event stream:**

- **Request headers** on `incoming_request`, plus the destination **port** and **HTTP version** (so Part 2 can build faithful HAR entries).
- A new **`incoming_response`** event with status, method, path, HTTP version, and response **headers**. It is emitted at the single point all response modes (basic / framed / chunked / mirror-discard) share, and **exactly once** across gateway retry attempts (emitted only after the point where the attempt can no longer be retried).
- New **`incoming_request_body`** / **`incoming_response_body`** events carrying a body **preview**. Streamed request bodies are accumulated across chunk frames and emitted once the last frame arrives.
- A per-exchange **correlation id** (`epoch:mode:connection:request`) on every HTTP event, so a request, its response, and their bodies pair up. The epoch is bumped on agent reconnect and the mode (steal/mirror) is folded in, because agent connection ids restart on reconnect and the two id spaces are independent.

**Bounding memory:** body capture is limited by the new `experimental.session_monitor_body_limit` option (default 32 KiB; `0` disables body capture entirely, keeping headers and metadata). The limit flows through `MonitorTx` so emit sites skip accumulation when capture is off. Bodies larger than the limit are truncated and flagged.

## Test plan

- `cargo test -p mirrord-intproxy` (62 pass), including:
  - a new gateway test asserting response events fire exactly once when the first send attempt fails with a retryable error *after* the response head was received;
  - a new `IncomingProxy` test covering the streamed chunked request-body path (frame accumulation, single body event, truncation at the limit, and the collision-free exchange id shape).
- `cargo test -p mirrord-config` with the schema regenerated (`mirrord-schema.json` updated), `cargo clippy -- --deny warnings` and `cargo fmt` clean, `cargo check -p mirrord` compiles.

Linear: PRO-174


mirrord/intproxy/src/proxies/incoming/http_gateway.rs
```diff
@@ -846,13 +924,16 @@ mod test {
...
+        let (monitor_raw_tx, mut monitor_rx) = tokio::sync::broadcast::channel(16);
         let _gateway = tasks.register(
             HttpGatewayTask::new(
                 request,
                 ClientStore::new_with_timeout(Duration::from_secs(1), Default::default()),
                 response_mode,
                 ListeningOn::Socket(addr),
                 IncomingTrafficTransportType::Tcp,
+                MonitorTx::from_sender(monitor_raw_tx),
+                "0:s:0:0".to_owned(),
```
hank-metalbear, line 936: This is the exchange correlation id the gateway stamps on every monitor event it emits. In production `IncomingProxy::monitor_exchange_id` builds it as `{epoch}:{s|m}:{connection_id}:{request_id}`, so `0:s:0:0` means epoch 0, steal mode, connection 0, request 0. This test constructs the `HttpGatewayTask` directly with no `IncomingProxy` in the loop, so it passes the literal id a real proxy would have produced. The assertions below then check the emitted `incoming_response` / `incoming_response_body` events carry it back unchanged, which is what lets the UI (and the HAR export in part 2) pair a response with its request.

PR #4466, opened 2026-07-02
config: session key improvements

iniw:
a collection of tiny improvements to the session key, including a change required for `down` (the best program ever).

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -397,7 +397,7 @@ pub(super) struct ExecParams {
     /// Can also be set with the `MIRRORD_KEY` environment variable.
     /// If not provided here, through `MIRRORD_KEY`, or in the config file, a unique key is
     /// generated automatically.
-    #[arg(long, env = "MIRRORD_KEY")]
+    #[arg(long)]
     pub key: Option<String>,
 }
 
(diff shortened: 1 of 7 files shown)


mirrord/cli/src/config.rs
```diff
@@ -397,7 +397,7 @@ pub(super) struct ExecParams {
     /// Can also be set with the `MIRRORD_KEY` environment variable.
     /// If not provided here, through `MIRRORD_KEY`, or in the config file, a unique key is
     /// generated automatically.
-    #[arg(long, env = "MIRRORD_KEY")]
```
0x00A5, line 400: This is removed because it becomes redundant if mirrord uses this env var key internally for passing the value?

iniw: Yeah, `ConfigContext` will look up `MIRRORD_KEY` in the inner hashmap but also in the environment variables:
https://github.com/metalbear-co/mirrord/blob/5d978e6547353a11b5f40ddb3d3acb6fbd133c28/mirrord/config/src/config/context.rs#L106-L128

0x00A5 (dismissed):

0x00A5 (dismissed):

0x00A5 (dismissed):

0x00A5 (dismissed):

0x00A5 (approved):

PR #4478, opened 2026-07-06
Bump kube-rs to mainline 4.0

DmitryDodzin:
Bump kube-rs from our fork to crates.io released 4.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,7 +299,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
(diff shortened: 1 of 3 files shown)


cristeigabriela:
I think this might regress @aviramha's changes in https://github.com/metalbear-co/mirrord/pull/4468 (merged into kube-rs/kube a few days ago, https://github.com/kube-rs/kube/commit/ae49cce192b85db3d734d290a6031aa2d9ac60e0)

As far as I can tell, https://crates.io/crates/kube/4.0.0 has been published 20 days ago so can't contain our fix, which is necessary for a customer issue on Windows, so we might have to keep the fork.

DmitryDodzin:
Will re-open once new release is published

gememma:
https://github.com/kube-rs/kube/releases/tag/4.2.0 👀 we have two in this one

PR #4479, opened 2026-07-06
Fix passthrough requests when app listens on external IP (COR-1643)

aviramha:
## Problem

In some cluster setups the target application resolves its own pod name and binds to the pod's external IP rather than `127.0.0.1`. The agent passes redirected connections (mirror/steal filter misses) through to loopback, so those passthrough requests fail — nothing is listening on `127.0.0.1`.

`ConnectionInfo::pass_through_address()` always rewrote the destination to loopback, deliberately, to avoid an iptables redirection loop (connecting to the pod IP is re-caught by the `-o lo` OUTPUT redirect).

## Fix

New `agent.external_ip_fix` option. When enabled, the agent passes connections through to their **original destination IP** instead of loopback.

To avoid the loop, passthrough sockets are marked with a dedicated fwmark (`0x2000`) via `SO_MARK` before connecting, and a `RETURN` rule is installed at the top of the agent's OUTPUT chain, ahead of the redirect:

```
-m owner --gid-owner {gid} -p tcp -m mark --mark 0x2000/0x2000 -j RETURN
```

- **gid + mask matching** means it only excludes the agent's own marked passthrough traffic — it can't touch or collide with other components' marks (kube-proxy `0x4000/0x8000`, Cilium `0x0F00`, Istio `0x1337`).
- The existing `! -s {pod_ips}` exclusion (which keeps the agent's *outgoing* connections to the pod IP redirected, from #2588) is left intact — only passthrough sockets carry the mark.

### Safety / portability
Marking requires `SO_MARK` (`CAP_NET_ADMIN`). The agent probes once at startup; if unsupported, it logs a single warning and falls back to loopback passthrough — i.e. it degrades to today's behavior rather than looping.

### Rollout
- **OSS**: enabled by default (the CLI's targeted-agent spec sets the env from `agent.external_ip_fix.unwrap_or(true)`).
- **Operator**: opt-in via `agent.external_ip_fix` (defaults off; operator-side wiring is separate).

## Notes for reviewers
- All passthrough connects now route through `ConnectionInfo::pass_through_connect()`; behavior is byte-for-byte unchanged when the flag is off.
- `mirrord-schema.json` regenerated.
- ⚠️ The `mirrord-agent` crate wasn't compiled on my machine (no Linux cross-toolchain locally); relying on CI for the agent build + unit tests. socket2/tokio APIs were verified against the vendored source.
- Worth confirming in a target setup whether the passthrough SYN is actually sourced from the pod IP (`getsockname()` on a connect to the pod IP). If it comes back `127.0.0.1`, the mark/iptables machinery is belt-and-suspenders; the design is safe either way.


aviramha:
Added an e2e regression test: `passthrough_to_app_listening_on_pod_ip` in `tests/src/traffic/steal.rs`, backed by a new `pod_ip_tcp_echo_service` fixture whose deployed app binds **only** to the pod IP (`HOST` from the downward API). It steals with an HTTP filter and sends unmatched raw TCP, so the connection is passed through to the pod-IP-bound app — which fails on the pre-fix (loopback) behavior.

**Ordering dependency:** the test relies on the `tcp-echo` image honoring a `HOST` env var, added in metalbear-co/test-images#42. That PR needs to merge and republish `mirrord-tcp-echo:latest` before this e2e will pass in CI.

aviramha:
Update: dropped the cross-repo dependency. The python test image already binds via `HOST`, so the e2e now reuses `mirrord-pytest` (HTTP passthrough path) with `HOST=status.podIP` — no test-images change needed, and it runs in CI as-is. (test-images#42 closed.)

aviramha:
**Approach change: SO_MARK → gid RETURN.** The e2e I added (`passthrough_to_app_listening_on_pod_ip`) caught that the SO_MARK-based loop avoidance didn't actually keep the pod-IP passthrough out of the redirect chain — the request still 502'd. Rather than iterate blind on the mark/iptables interaction, I switched to reusing the mechanism that already works for loopback passthrough: when `external_ip_fix` is enabled, the agent's OUTPUT `RETURN` rule excludes **all** of the agent's own traffic (drops the `! -s {pod_ips}` qualifier) instead of marking individual sockets.

This removes the SO_MARK socket handling, the capability probe, and the dedicated fwmark. **Trade-off:** while enabled, connections the agent makes to the pod's own IP for the outgoing feature are no longer redirected (the loopback-steal behavior from #2588). Since it's default-on in OSS, flagging that explicitly for review.

MintSoup (approved):

PR #4482, opened 2026-07-06
Add mirrord ui endpoints for kube contexts and namespaces

aviramha:
### What & why

Adds two authenticated `GET` endpoints to the `mirrord ui` local server so clients (e.g. the browser extension / frontend) can discover the user's Kubernetes contexts and namespaces without shelling out to `kubectl`:

- **`GET /api/contexts`** — lists the kube contexts from the merged kubeconfig (honouring `KUBECONFIG`) plus the currently selected context. Reads the kubeconfig directly and makes no cluster call, so it works even when no cluster is reachable.
  ```json
  { "contexts": ["ctx-a", "ctx-b"], "currentContext": "ctx-a" }
  ```
- **`GET /api/namespaces?context=<name>`** — lists namespaces for the given context, or the current context when `context` is omitted. This one queries the API server, so it needs the context reachable and the user authorized to list namespaces.
  ```json
  { "namespaces": ["default", "kube-system"], "context": "ctx-b" }
  ```

Both are registered under the token-authenticated `/api` nest, so they require the same auth as the existing `/api/me` endpoint. On success they return the data directly; on failure they return a non-2xx status (`500` for a kubeconfig read error, `502` when the cluster/context is unreachable or the list fails) with an `{ "error": "..." }` body — the client checks the status code rather than a nullable field in the success payload.

### Quality Checklist:

- [ ] ~I have written unit tests or purposefully omitted them~
- [x] I have explained what this PR introduces and why


mirrord/cli/src/ui.rs
```diff
@@ -6,134 +6,1031 @@
...
+        }
+        Err(err) => axum::Json(CurrentUserResponse {
+            k8s_username: None,
+            error: Some(format!("self-subject-review failed: {err}")),
+        }),
+    }
+}
+
+/// Error body returned by the kube-facing endpoints on failure, alongside a non-2xx status.
+#[derive(Serialize)]
```
meowjesty: Instead of having `error: String`, change this to use `thiserror` like we're doing for the other ui errors (see the chaos errors as an example).

Also, move the error implementation to a module under `ui`, called `error`.

mirrord/cli/src/ui.rs
```diff
@@ -6,134 +6,1031 @@
...
+#[serde(rename_all = "camelCase")]
+struct NamespacesResponse {
+    namespaces: Vec<String>,
+    context: Option<String>,
+}
+
+/// Builds a kube client for the given context, or for the kubeconfig's current context when
+/// `context` is `None`.
+async fn client_for_context(context: Option<&str>) -> Result<Client, String> {
```
meowjesty: Instead of returning `Result<..., String>` here, use the `ApiError` type.

Create a type alias `type UiResult<T> = Result<T, ApiError>` and use it for other top level ui functions.

mirrord/cli/src/ui.rs
```diff
@@ -6,134 +6,1031 @@
...
+                .iter()
+                .filter_map(|namespace| namespace.metadata.name.clone())
+                .collect(),
+            context: query.context,
+        })
+        .into_response(),
+        Err(err) => ApiError::response(
+            StatusCode::BAD_GATEWAY,
+            format!("failed to list namespaces: {err}"),
+        ),
```
meowjesty: Instead of inline converting errors like this here, implement `IntoResponse` for the `ApiError`, giving appropriate status codes for the errors in that implementation.

meowjesty (changes requested):
Improve error type usage.

meowjesty (dismissed):
👍

mirrord/cli/src/ui.rs
aviramha: Done in 2f09d804 — moved the error into a new `ui::error` module as a `thiserror` enum (`ApiError`), following the chaos errors pattern.

mirrord/cli/src/ui.rs
aviramha: Done — `client_for_context` now returns `UiResult<Client>` (`type UiResult<T> = Result<T, ApiError>`), and `list_contexts`/`list_namespaces` use it too and propagate with `?`.

mirrord/cli/src/ui.rs
aviramha: Done — `ApiError` now implements `IntoResponse`, mapping each variant to a status code (kubeconfig read / client init → 500, bad context → 400, cluster API request → 502). The inline conversions are gone.

vladrbg (approved):

PR #4483, opened 2026-07-06
COR-1472 𓇼 ⋆｡˚ 𓆝⋆｡˚ 𓇼 Windows crash and exception diagnostics

cristeigabriela:
# Windows crash and exception diagnostics

<p align="center">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/cristeigabriela/pr-assets/97f26c576df83b53ad9a9a8adc632e67d0bef937/crash_reporter_darkmode.svg">
    mirrord Windows crash-reporter — execution flow, links, lifecycle
  </picture>
</p>

## Fish

<details>

image
image

</details>

## Linear

https://linear.app/metalbear/issue/COR-1472/windows-crash-exception-handler

## Summary

Adds crash and exception diagnostics to the Windows layer.

There are two parts:

- an **in-process crash handler** installed into each layer'd process; and
- a **per-session out-of-process monitor** that dumps crashes from the outside, tracks every layer'd
  process, and surfaces a single copyable report.

An abnormal exit during or after layer injection is now self-explanatory instead of surfacing as an
opaque downstream error.

Extension-managed launches opt out of crash reporting. Both the IntelliJ and VS Code integrations
launch through `mirrord ext`, where an IDE stop abruptly terminates the application and cannot be
reliably distinguished from an unexpected external kill.

A session can hit more than one crash, otherwise known as an ***incident***, in the code (a native fault, an external kill, or an init failure), each a real, independently captured event. The interactive dialog is shown once, for the first, so a busy session is not spammed; nothing is dropped, though — every incident is written to disk, and the single attachable archive, `mirrord-crash-session_<stamp>.zip`, is cumulative: it collects them all.

<p align="center">
  mirrord crash dialog
  mirrord crash dialog
</p>

## Motivation

On Windows the layer is injected into the target process and into every process it spawns. When a
layer'd process died abnormally, the only visible symptom was a downstream error with no attribution:
no indication of whether the process had crashed, been killed from the outside, or failed to
initialize the layer at all, and no artifact to inspect after the fact. A crash inside a broken
process also cannot safely produce its own dump.

This change makes those failures observable. It distinguishes an internal crash (a native fault,
captured with a record and a minidump) from an external termination (a process killed with no handler
firing) from an early layer-initialization failure (a process that never finished coming up under the
layer), and it produces the artifacts from a separate process so the dump is taken from a healthy
address space.

## What it produces and how to collect it

A single session can hit more than one crash — **incidents**, in the code (a native fault, an external
kill, or an init failure). They are independent, real events: a dead parent and its stranded child, for
example, is one session with two incidents. Each incident writes a set of flat files in the crash
directory, sharing one name stem so they cluster:

```
mirrord-crash_<timestamp>_<name>_pid<pid>.report.txt    short, copyable summary (source of truth)
mirrord-crash_<timestamp>_<name>_pid<pid>.record.txt    in-process crash record (code, address, module, stack)
mirrord-crash_<timestamp>_<name>_pid<pid>.modules.txt   the process's module inventory
mirrord-crash_<timestamp>_<name>_pid<pid>.dmp           minidump (kept local until the session ends)
```

The interactive dialog is shown **once per session** — for the first incident — but the attachable
archive is **cumulative**. There is exactly one per session, `mirrord-crash-session_<stamp>.zip`,
rebuilt as each incident lands and finalized when the session ends, and it collects **every** incident
that actually occurred:

```
mirrord-crash-session_<stamp>.zip
├── crash-report.txt              the first incident's report (the whole process tree at a glance)
├── incidents/<name>_pid<pid>/    one folder per incident
│   ├── crash-report.txt
│   ├── crash-record.txt          when a record was captured
│   ├── modules.txt               when modules were enumerated
│   └── dump.dmp                  when a dump was produced
└── layer-logs/                   the session's layer logs, deduplicated
```

The files land in the log directory (`MIRRORD_LAYER_LOG_PATH`, or `%TEMP%/mirrord` by default). The
report is also printed to the console, and the dialog offers three actions: **Copy** (the report
text), **Open folder** (the crash directory), and **Report Crash** (the GitHub issue page).

The native dialog uses the current MetalBear mascot, embedded as a Windows bitmap and pre-composited against the header background because `SS_BITMAP` does not alpha-blend.

To collect a report, send the one `mirrord-crash-session_*.zip` — it is the single file to attach, and
it already holds every incident's report, record, and minidump plus the session's layer logs, so there
is no "which file do I send?" to get wrong. The minidumps live only inside it (their loose files are
removed when the session ends), because a dump can contain sensitive in-memory data; a full-memory dump
is only produced behind an opt-in flag (see [Environment variables](#environment-variables)) for the
same reason.

## Design

Two cooperating pieces.

**In-process crash handler** (`utils-win/src/diagnostics/crash.rs`). Installed early in each layer'd
process. It uses two registrations:

- A chained top-level `SetUnhandledExceptionFilter`. It fires only on a genuinely unhandled
  exception, so it never pre-empts a first-chance fault that a downstream `__try`/`__except` was
  going to handle. It always chains to the previously installed filter, so Windows Error Reporting
  and any host handler still run. A hook on `SetUnhandledExceptionFilter`
  (`layer-win/src/hooks/exception.rs`) keeps the layer's filter installed as the OS top-level filter
  while adopting the target runtime's filter as the chain target (`adopt_previous_filter`), so a
  runtime that installs its own filter does not displace it.
- A narrow vectored exception handler that acts only on stack overflow, where a first-chance fault is
  already fatal and recovery is impossible. Other severe codes are left to the filter.

**The handler runs inside a process whose state may be unstable and even unrecoverable**, so it relies
only on resources it prepared earlier, while the process was still healthy. Everything it needs is
captured at install time (`install`):

- the crash-record path (`CrashState::record_path`)
- a reserved handler stack (`SetThreadStackGuarantee`)
- the `dbghelp` library, pre-loaded (`load_system_library("dbghelp.dll")`)
- a snapshot of the loaded modules (`ModuleTable::capture`)

So at crash time it allocates nothing, calls no loader, and touches no state a fault could have
corrupted. It formats into a fixed stack buffer (`FixedBuf`) ⇒ writes with raw `WriteFile` ⇒
resolves the faulting module from the cached table (`ModuleTable`). A re-entrancy guard turns a fault
inside the handler into a clean bail-out rather than a loop, and the reserved handler stack leaves room
to run even on stack exhaustion.

**Out-of-process monitor** (`cli/src/crash_monitor.rs`, `utils-win/src/diagnostics/monitor*`,
`report.rs`, `dialog*`). One monitor runs per session. It accepts registrations from layer'd
processes, writes the minidump of a crashed process from the outside (the crashing process only
signals; it never runs dump code in its own broken address space), detects processes that die with no
handler firing, composes the report, and shows one native dialog. The in-process dump is a fallback,
used only when no monitor is configured or it does not acknowledge in time.

## Semantics

- **Reported:** native OS faults (a record, a minidump, a report), external kills (no handler fired,
  no dump possible — a death verdict), and early layer-initialization failures.
- **Not caught:** a regular application exception, including a managed .NET exception (the application
  bit — see *NTSTATUS classification* below).
- **One per session tree.** At most one report and one dialog per session — the first reportable
  event claims it, and later deaths in the same session are coalesced. Artifacts (report text, console
  line) are always written even when the dialog is dismissed, and the dialog is suppressed under `CI`.
- **All processes are tracked.** Every layer'd process in the session registers, so the report renders
  the full process tree. A process that has already exited is kept in the tree and marked with its
  exit code, so the ancestry around a failure stays legible.
- **Silent otherwise:** any clean exit at any exit code (the discriminator is the clean-shutdown
  signal, not the code), a managed .NET exception, `Ctrl-C`, and an intentional debugger or
  IDE-extension stop.

## NTSTATUS classification and the application bit

The reporter separates a native OS fault from an application's own exception, and it does so with a
single bit of the 32-bit NTSTATUS rather than a list of per-runtime magic numbers.

```text
   31 30 29  28  27            16 15                     0
  ┌─────┬───┬───┬────────────────┬────────────────────────┐
  │ Sev │ C │ N │    Facility    │          Code          │
  └──┬──┴─┬─┴─┬─┴───────┬────────┴───────────┬────────────┘
     │    │   │         │                    └─ bits 15..0:  the status code proper
     │    │   │         └─ bits 27..16: facility — which subsystem raised it
     │    │   └─ bit 28: reserved, always 0
     │    └─ bit 29: customer / application-defined  ← 0x2000_0000, the bit we test
     └─ bits 31..30: severity (00 success, 01 info, 10 warning, 11 error)
```

The rule is that one bit. The OS leaves it clear on the faults it raises; a language runtime sets it
on the exceptions it defines. So the reporter acts on error-severity codes with the customer bit
**clear** (a native fault) and stays silent on codes with it **set** (an application-level exception).
The one special case is `Ctrl-C` (`0xC000013A`): an OS code, but a user action, so it is treated as
application-level too.

The table is generated by `gen_ntstatus_table.rs` from the same predicates the code uses
(`report.rs`), so it tracks the shipped classifier:

| NTSTATUS | Meaning | Severity | Customer bit | NT_ERROR | Verdict |
| --- | --- | --- | :---: | :---: | --- |
| `0xC0000005` | access violation | error | clear | yes | reported - native fault |
| `0xC00000FD` | stack overflow | error | clear | yes | reported - native fault |
| `0xC0000409` | fail-fast / stack buffer overrun | error | clear | yes | reported - native fault |
| `0xC0000374` | heap corruption | error | clear | yes | reported - native fault |
| `0xC0000094` | integer divide by zero | error | clear | yes | reported - native fault |
| `0xC000013A` | Ctrl-C exit | error | clear | yes | ignored - application/user-level |
| `0xE0434352` | managed .NET exception | error | set | yes | ignored - application/user-level |
| `0xE06D7363` | C++ EH throw | error | set | yes | ignored - application/user-level |
| `0xE0000027` | application-defined exit | error | set | yes | ignored - application/user-level |
| `0x80000003` | breakpoint | warning | clear | no | ignored - not error-severity |
| `0x00000000` | clean exit (0) | success | clear | no | ignored - not error-severity |
| `0x00000001` | clean exit (1) | success | clear | no | ignored - not error-severity |

The practical consequence is in the last column: a regular application exception — a managed .NET
exception, a C++ throw, any runtime-defined code — is never caught. The reporter is for the layer and
OS faults and for external kills, not for the target application's own error handling.

## Communication

The layer and the monitor are linked two ways, by design.

**Registration — TCP.** The CLI spawns the monitor as a subcommand; the monitor binds a localhost
`TcpListener` and prints its address, which the CLI forwards to layer'd processes via
`MIRRORD_LAYER_CRASH_MONITOR_ADDR`. Each layer'd process connects once at startup — while it is
healthy — and registers (`Registration`: pid, parent pid, name, session role, artifact stem, log
path; length-prefixed bincode). The monitor records a process node, creates the per-process kernel
objects, and acknowledges. This is a healthy-time, connection-oriented channel.

**Signalling — named events and shared memory.** A crashing process cannot safely open a socket, so
crash signalling does not use TCP. Registration creates per-pid kernel objects the two sides share:

| Object | Direction | Purpose |
| --- | --- | --- |
| `mirrord_crash_event_<pid>` | handler sets, monitor waits | a crash (or init failure) occurred |
| `mirrord_crash_done_<pid>` | monitor sets, handler waits | the monitor finished (dumped / recorded) |
| `mirrord_crash_clean_<pid>` | layer sets on `DLL_PROCESS_DETACH` | a normal shutdown ran |
| `mirrord_crash_info_<pid>` | shared section | the fault details the handler hands over |

Two capabilities travel this channel:

- **Signal a crash** (`signal_crash`): the handler writes the faulting thread and exception pointers
  into the shared section, sets the crash event, and waits on the done event. The monitor reads the
  section while the process is frozen, writes the dump from the outside, sets the done event, and
  composes the report. Allocation-free and safe from a crash handler.
- **Signal a notable failure** (`signal_init_failure`): the same channel with a reason string, used
  when the layer fails to initialize. Without it, an init failure would exit through
  `DLL_PROCESS_DETACH`, signal a clean shutdown, and the monitor would never learn of the failure.

**Clean-shutdown discriminator.** A normal shutdown runs `DLL_PROCESS_DETACH` and sets
`mirrord_crash_clean_<pid>`. An abrupt kill or crash does not. The monitor infers an abnormal death
from the *absence* of that signal, not from the exit code — an exit code conflates a normal non-zero
exit, a killer-chosen `TerminateProcess` code, and a fault code.

## Lifecycle

The crash-reporter is scoped to a `mirrord exec` session. `mirrord ext` disables it for the entire
layer'd process tree, leaving application runtimes in control of their normal exception filters.

- **One per session.** The CLI spawns exactly one monitor at session start and passes it the root
  process id.
- **Lives the whole session.** The monitor is not tied to any single process. A watchdog waits on the
  root process and tears the monitor down when the session ends. It is deliberately not killed when
  the CLI object drops, because a dialog may still be open.
- **Outlives a crash until dismissed.** When a report has a dialog, the watchdog keeps the monitor
  alive while the dialog is on screen (`DIALOG_SHOWING`) rather than exiting the moment the crashed
  process is gone.
- **Cleans up.** If the CLI created a per-session temp log directory and nothing crashed, the monitor
  removes it on a clean session exit; on a crash it is kept.

## Environment variables

Four new variables, plus one reused. Propagation to children is via the Windows env-forwarding list
`FORWARDED_ENV_VARS` (`layer-lib/src/process/windows/execution/mod.rs`), applied in
`add_mirrord_env_vars` immediately before the `CreateProcess` call, so a forwarded variable reaches
every spawned layer'd process.

> [!NOTE]
> Windows crash reporting is not supported for extension-managed runs. IntelliJ and VS Code launch
> through `mirrord ext`, which sets `MIRRORD_LAYER_CRASH_REPORTING=false` for the entire injected
> process tree. This avoids reporting the IDE's normal stop action as an unexpected external kill.

| Variable | Purpose | Set by | Read by | Forwarded to children |
| --- | --- | --- | --- | :---: |
| `MIRRORD_LAYER_CRASH_REPORTING` | internal session policy: enabled for `mirrord exec`, disabled for `mirrord ext` | CLI | layer | yes |
| `MIRRORD_LAYER_FULL_MEMORY_DUMP` | opt in to a full-memory dump instead of the bounded default | user | layer (`install`) and monitor | yes |
| `MIRRORD_LAYER_CRASH_MONITOR_ADDR` | the per-session monitor's localhost endpoint | CLI, in `spawn_crash_monitor` | layer, to register | yes |
| `MIRRORD_CRASH_EPHEMERAL_DIR` | marks that the CLI created a temp log dir, so the monitor removes it on a clean exit | CLI, on the monitor env only | monitor | no (monitor-only) |
| `MIRRORD_LAYER_LOG_PATH` (reused) | crash-artifact directory; falls back to `%TEMP%/mirrord` | user or CLI | layer and monitor | yes |

## Implementation

A new Windows-only crate, `mirrord-utils-win`, holds the shared pieces:

| Module | Role |
| --- | --- |
| `diagnostics/crash.rs` (+ `crash/codes.rs`) | in-process handler; exception-code classification |
| `diagnostics/monitor.rs` (+ `monitor/channel.rs`, `monitor/server.rs`) | crash-channel protocol; client and server |
| `diagnostics/report.rs` | report composition, NTSTATUS classification, process tree, `.zip` bundle |
| `diagnostics/dump.rs` | `MiniDumpWriteDump` wrappers (out-of-process and in-process) |
| `diagnostics/dialog*` | the native Win32 crash dialog (shown by the monitor only) |
| `diagnostics/handle.rs` | `OwnedHandle` RAII for kernel handles and mapped views |
| `modules.rs` | module inventory and address resolution; security-vendor flagging |
| `process.rs`, `fixed_buf.rs`, `clipboard.rs` | process identity; allocation-free buffer; dialog clipboard |

The layer and CLI wire it in: `layer-win/src/diagnostics.rs` (install / signal / uninstall /
snapshot), `layer-win/src/hooks/exception.rs` (the `SetUnhandledExceptionFilter` hook), and
`cli/src/crash_monitor.rs` (the `mirrord crash-monitor` subcommand).

Kernel handles and mapped views are held through an `OwnedHandle` RAII wrapper, and the dialog renders
its Slack/GitHub/email links as clickable RichEdit links.

### Quality Checklist:

  documentation)


changelog.d/+windows-crash-diagnostics.added.md
```diff
@@ -0,0 +1 @@
...session role and whether its parent is still alive (with its exit code), distinguishing an internal crash from an external termination such as an `EDR` killing the injector. A per-session out-of-process monitor writes minidumps from outside the crashed process, detects processes that are terminated with no handler firing, and surfaces a short, copyable crash report. An early layer-initialization failure (for example a child stranded because its parent died before initialization finished) is now surfaced through the same report instead of vanishing as a silent exit. The reporter stays silent on a normal exit at any exit code, and only treats genuine native faults and external kills as reportable: an application's own uncaught exception (including a managed .NET exception), a `Ctrl-C`, and an intentional debugger or IDE-extension stop are left alone. Each incident's files share one name stem in the log directory, and the attachable `.zip` holds the report, the crash record, and the layer logs, while the memory dump is kept separate and local because it can contain sensitive in-memory data. A one-line module summary at startup flags any loaded security or endpoint-protection products.
```
cristeigabriela: keep comment at max 2 sentences 1 paragraph

mirrord/cli/src/config.rs
```diff
@@ -167,6 +167,30 @@ pub(super) enum Commands {
...
+        port: u16,
+
+        /// The root CLI process id, recorded for the process-tree report.
+        #[arg(long, default_value_t = 0)]
+        root_pid: u32,
+
+        /// Debug arguments.
+        ///
+        /// Passed only so the monitor's command line is self-describing in a process listing
+        /// (Task Manager, Process Explorer), to aid debugging. Never read.
```
cristeigabriela: remove task manager from the list

mirrord/cli/src/crash_monitor.rs
```diff
@@ -0,0 +1,80 @@
...
+use utils_win::diagnostics::{
+    crash_dir, full_memory_dump,
+    monitor::{MonitorConfig, serve},
+};
+
+use crate::error::{CliError, CliResult};
+
+/// Env var the CLI sets when it created an ephemeral per-session log dir. Its presence tells the
+/// monitor to remove that dir on a clean session exit. Set in [`crate::execution`], read here.
+pub(crate) const MIRRORD_CRASH_EPHEMERAL_DIR: &str = "MIRRORD_CRASH_EPHEMERAL_DIR";
```
cristeigabriela: consider if this is the best place for the env

mirrord/layer-lib/src/process/windows/diagnostics.rs
```diff
@@ -0,0 +1,91 @@
...
+                presence(&parent),
+                presence(&layer),
+                presence(&proxy),
+            ),
+        },
+    }
+}
+
+/// Reports whether an optional value is present.
+fn presence(value: &Option<String>) -> &'static str {
```
cristeigabriela: get rid of this utility

mirrord/layer-win/src/diagnostics.rs
```diff
@@ -0,0 +1,190 @@
...
+/// # Arguments
+///
+/// * `process_terminating` - whether the process is exiting, not just unloading the layer.
+pub fn uninstall_crash_handler(process_terminating: bool) {
+    crash::uninstall(process_terminating);
+}
+
+/// The functions we hook. Their prologues are snapshotted before we touch them.
+const HOOKED_FUNCTIONS: &[(&str, &str)] = &[
+    ("kernelbase", "CreateProcessInternalW"),
```
cristeigabriela, line 114: exhaust this a bit more, but this is not an ideal way to approach this functionality in the first place, though we can revise more complexly should the need arise later. this is however still useful as an indicator for now so it shouldnt be fully removed

cristeigabriela (commented):
this and also make sure to make some comments about needing to move more things to utils-win that are currently fo rexample present alongside mirrord ui for token stuff

mirrord/cli/src/execution.rs
```diff
@@ -79,6 +80,15 @@ pub(crate) struct MirrordExecution {
     #[serde(skip)]
     child: Option<Child>,
 
+    /// The per-session Windows crash-dump monitor process.
+    ///
+    /// Not killed on drop: the monitor self-exits when it sees this CLI die, so a crash dialog can
+    /// outlive the session until the user dismisses it. The field is never read directly.
+    #[cfg(windows)]
+    #[serde(skip)]
+    #[allow(dead_code)]
```
0x00A5, line 110: Do we need allow dead code?

mirrord/layer-lib/src/process/windows/diagnostics.rs
```diff
@@ -0,0 +1,82 @@
...
+                    ),
+                },
+            }
+        }
+        (parent, layer, proxy) => SessionRole::MalformedEnv {
+            detail: format!(
+                "partial child env (parent_pid={}, layer_id={}, proxy_addr={})",
+                parent.map_or("missing", |_| "present"),
+                layer.map_or("missing", |_| "present"),
+                proxy.map_or("missing", |_| "present"),
```
0x00A5, line 78: do we not want to record the actual pid, id or addr, but only `present`?

mirrord/utils-win/examples/preview_dialog.rs
0x00A5, line 1: Are these examples leftover from experimenting? or we want to keep them?

mirrord/utils-win/src/diagnostics/monitor/channel.rs
```diff
@@ -0,0 +1,190 @@
...
+/// anything by name and never allocates. Each is an [`OwnedHandle`], so dropping the channel frees
+/// every handle and unmaps the view — no hand-written cleanup.
+///
+/// `view` is the heart of the channel. At registration the monitor created a named shared section
+/// (`mirrord_crash_info_<pid>`) sized to exactly one [`CrashInfo`]; `MapViewOfFile` maps that
+/// section into this process and returns a pointer to its bytes. Because both sides open the *same*
+/// named section and agree on the `#[repr(C)]` [`CrashInfo`] layout, `view.as_ptr() as *mut
+/// CrashInfo` is a valid pointer to a shared `CrashInfo`: what the layer writes here, the monitor
+/// reads from outside.
+pub struct MonitorChannel {
```
0x00A5, line 44: Can we add a safety doc comment on this type or on the methods and explains why `signal_crash` and `signal_init_failure` are safe?

mirrord/utils-win/src/diagnostics/monitor/channel.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn signal_crash(&self, thread_id: u32, exception_pointers: usize) -> bool {
+        unsafe {
+            let info = self.view.as_ptr() as *mut CrashInfo;
+            (*info).thread_id = thread_id;
+            (*info).kind = KIND_CRASH;
+            (*info).exception_pointers = exception_pointers as u64;
+
+            SetEvent(self.crash_event.as_ptr() as HANDLE);
+            WaitForSingleObject(self.done_event.as_ptr() as HANDLE, DUMP_ACK_TIMEOUT_MS)
+                == WAIT_OBJECT_0
```
0x00A5, line 80: Win noob question: we only need to wait for this once ever right? do we need to clear the object or reset it?

mirrord/utils-win/src/diagnostics/monitor/server.rs
```diff
@@ -0,0 +1,722 @@
...
+                            &config,
+                            Outcome::Terminated { exit_code },
+                            None,
+                            None,
+                        );
+                    }
+                    DeathVerdict::Reportable => {}
+                }
+                break;
+            } else if result == WAIT_OBJECT_0 + 1 {
```
0x00A5, line 419: Question: we assert result == 1 rather than >= 1 because with bWaitAll set to false, it is not possible for `WaitForMultipleObjects` to return results > 1?

mirrord/utils-win/src/diagnostics/crash.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+
+/// Reserves crash-handler stack on the current thread, once the handler is installed.
+///
+/// `SetThreadStackGuarantee` is per-thread, so the layer calls this on every `DLL_THREAD_ATTACH`:
+/// without it, a stack overflow on a worker thread that never got the guarantee re-faults inside
+/// the handler and bails via the re-entrancy guard. A no-op until the handler is installed. Note
+/// this cannot cover the process's pre-existing main thread, which never gets a thread-attach
+/// callback.
+pub fn reserve_handler_stack() {
```
0x00A5, line 291: TIL, this is cool 👍

mirrord/utils-win/src/diagnostics/monitor/server.rs
cristeigabriela: I'm not sure what you mean, but what we're doing here is we're waiting on either:

- The `process` handle being signaled (implying that, the process has finished executing)
- The `crash_event` handle being signaled (implying that, execution has broke because of an anomaly [but the process has not terminated yet!])

The way the array of handles passed to `WaitForMultipleObjects` is ordered respects the natural lifecycles we might enocunter here too (eg. the process can't signal finish if a crash report is pending, and a crash report can't signal if the process finished).

We only plan to respond to *one* of these two events (the process will signal done after a crash report, but we don't want to do the regular process termination logic if we already handled it as a crash).

I assume you were asking something along the terms of: what if both those objects are signaled?

> When bWaitAll is FALSE, this function checks the handles in the array in order starting with index 0, until one of the objects is signaled. If multiple objects become signaled, the function returns the index of the first handle in the array whose object was signaled.

It short-circuits on the first signaled handle.

I'm not sure what you mean in the end, but I have 2 suspicions, and I'll address those:
- In case this is what you meant, there's no scenario where we would have to handle multiple objects having signaled in one call
- In case this is what you meant, `WAIT_TIMEOUT`, `WAIT_ABANDONED_`, those are never possible (no timeout, not going to happen outside of conditions where execution is broken beyond fixing)
  - IDK about `WAIT_FAILED` but I've never ran into that in my career

mirrord/utils-win/src/diagnostics/monitor/channel.rs
cristeigabriela: very good instinct! based on the specific exception codes we handle, we do not need to reset crash-event after a dump happened! because the exceptions we catch are always fatal. even if they were for some reason caught and some code made the OS let everything proceed, we don't care to re-do the entire routine again

mirrord/utils-win/examples/preview_dialog.rs
cristeigabriela: it helps to test the crash dialog UI without actually running e2e. it'll be helpful as we actually get the branding in (currently there's no branding embed, see the image in PR)

mirrord/cli/src/execution.rs
cristeigabriela: not sure, i'll check

mirrord/layer-lib/src/process/windows/diagnostics.rs
cristeigabriela: here it's fine i think! but we can change

0x00A5 (dismissed):

mirrord/utils-win/src/diagnostics/monitor/channel.rs
cristeigabriela: will do!

aviramha:
@cristeigabriela note this is approved, I guess just need to solve conflicts and merge?

cristeigabriela:
Yes, also needs a branding update for the logo, that's all

cristeigabriela:
gonna finish up today ish

0x00A5 (dismissed):

0x00A5 (approved):

PR #4493, opened 2026-07-07
feat(ui): session monitor detail view, export, and inspector redesign (PRO-174) (2/2)

hank-metalbear:
## Summary

Part 2 of 2 for PRO-174, stacked on #4464. Builds the `mirrord ui` session monitor detail view and export on top of the richer event stream from part 1, then reworks the surrounding UI into a network-inspector-style experience.

### Events table + inspector
- The event feed becomes a columnar table (time, type, method, path, status, duration, repeat count) with tinted type badges and green/red status pills.
- A **group-repeats** toggle collapses repeating traffic cycles (health probes, polling) into single rows with an `×N` count, using a small lookback window since one HTTP exchange fans out into interleaved request/response/body events.
- Request and response **body events fold into their head row** via the shared exchange id, so one request or response is one row and the inspector shows headers and body together.
- Clicking a row (or `J`/`K`) opens a resizable **inspector pane** on the right with structured Headers and Body sections, a **Copy as cURL** button that reconstructs the request, **Copy JSON**, an exchange **duration**, and position + arrow-key navigation.

### Export (JSON + HAR)
- One click downloads a zip with the redacted session log (every event) and a HAR of the HTTP exchanges, each self-describing: a top-level description with target, session id, timestamps and event count, per-entry stolen/mirrored comments, and a HAR log comment covering redaction and truncation. Sensitive header values are redacted; oversized bodies are truncated and flagged.

### Session chrome
- Config leaves the inspector: it opens as a **centered modal** (session-level, occasional, read-only) so the right side is unambiguously the request inspector. The modal renders the whole config into titled sections plus a collapsed complete-source Raw JSON.
- The session header carries only what changes how the feed reads (target, mode, uptime, Join, Config, kill); the noisy chip strip is gone.

### Keyboard-first
- Shortcuts are taught in-context with visible keycaps (matching the search box's ⌘F): `J`/`K` walk the feed and inspector, `Space` pauses, `G` toggles grouping, `F` cycles filters, `E` export, `C` clear, `Y` copy-cURL, `\` toggles the sidebar, `Esc` peels one layer, and `⌘K` / `?` open a command palette that lists every action with its key.

### Follow-up polish (from design review)
- **Session config moves to the sidebar**: a gear on each local session row opens the config modal directly; the header overflow menu is gone and the header keeps a single kill button, separated from the Join chip by a divider so the destructive action is not a near-miss click. Config only exists for local sessions; cluster-side rows never offer it.
- **Session cards decluttered**: the operator badge and location icon are gone, and a card shows its context/namespace only when it differs from the top-bar selection.
- **Every event type is inspectable**: DNS, file, outgoing and process events open in the inspector with a Details grid (previously only HTTP rows were clickable), and outgoing rows no longer double the port.
- **Errors lead with the message**: 4xx/5xx responses with a captured body show the error message (from conventional keys like `error`/`message`/`detail`) as a callout above the headers.
- **Command palette affordance**: a visible `⌘K` button sits at the end of the events toolbar, and `⌘F` filters events while `⌘⇧F` focuses the sidebar search so each shortcut has one meaning.
- Long values (headers, config paths, MIME types) wrap at natural delimiters instead of mid-word.

## Verification
- `packages/monitor` vitest suite passes (9), `tsc` typecheck clean, `cargo fmt`/`clippy` clean, CLI builds.
- Exercised end to end against a live cluster session (mirror and steal): table, grouping, inspector nav, cURL/JSON copy, errors filter, pause across the event cap, config modal, keyboard shortcuts, and the JSON + HAR export (validated redaction, truncation flag, query strings, and per-entry mode comments).

Linear: PRO-174

> Note: the visual redesign, keyboard affordances and config modal are best confirmed with a quick pass in a browser.


packages/monitor/src/components/events/InspectorPane.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+const STATUS_TONE = (status: number) =>
+  status >= 400
+    ? 'text-red-700 bg-red-100 dark:text-red-400 dark:bg-red-950'
+    : 'text-emerald-700 bg-emerald-100 dark:text-emerald-400 dark:bg-emerald-950'
+
+// Rebuilds a runnable curl command from a captured request event: URL from host/port/path,
+// captured headers minus content-length (curl recomputes it), and the merged body if one was
+// captured. Only request events carry enough to reproduce the call.
+export function buildCurl(raw: unknown): string | null {
```
hank-metalbear, line 68: Are there no packages to do this?

PR #4495, opened 2026-07-08
Container host gateway detection

0x00A5:
Added container host gateway auto detection for `mirrord container`. By using the container argument `--add-host mirrord-external-proxy:host-gateway` for the intproxy sidecar, intproxy is able to resolve `mirrord-external-proxy` to the container runtime's host gateway IP address.
When connecting to the external proxy, which has the actual agent connection running on the host machine, intproxy sidecar no longer needs to rely on the external proxy IP address. Instead, it connects to `mirrord-external-proxy` hostname directly. To make sure the resolved IP address is handled by external proxy, when this flag is set to true, external proxy binds UNSPECIFIED (all interfaces).

The feature flag `container.host_gateway_detection` is set to `true` by default, because `--add-host <myhostname>:host-gateway` is a well documented way to allow containers to connect host machine using custom DNS.

* podman: https://docs.podman.io/en/latest/markdown/podman-run.1.html#add-host-hostname-hostname-ip
* nerdctl: https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md?plain=1#L203
* docker: https://docs.docker.com/compose/how-tos/networking/#custom-dns-with-extra_hosts

Worst case, the feature can be disabled by setting `container.host_gateway_detection` to `false` :D

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/intproxy/src/agent_conn.rs
+++ b/mirrord/intproxy/src/agent_conn.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use mirrord_analytics::{NullReporter, Reporter};
-use mirrord_config::LayerConfig;
+use mirrord_config::{LayerConfig, container::MIRRORD_EXTERNAL_PROXY_HOSTNAME};
 use mirrord_kube::{api::kubernetes::AgentKubernetesConnectInfo, error::KubeApiError, kube};
 use mirrord_operator::{
     client::{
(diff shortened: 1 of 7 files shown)


mirrord/intproxy/src/agent_conn.rs
```diff
@@ -199,11 +199,23 @@ impl AgentConnection {
...
+                            proxy_addr.port(),
+                        ))
+                        .await?;
+                        stream.set_nodelay(true)?;
+                        stream
+                    } else {
+                        let socket = TcpSocket::new_v4()?;
+                        socket.set_keepalive(true)?;
+                        socket.set_nodelay(true)?;
+                        socket.connect(proxy_addr).await?
```
MintSoup, line 217: I'm assuming the split is because in one branch we use hostname and in the other a `SocketAddr`? And why do we not set keepalive in the first branch?

MintSoup (approved):
looks good, one question only

mirrord/intproxy/src/agent_conn.rs
0x00A5: Yeah exactly. `tokio::net::TcpStream` doesn't support setting keepalive, due to a upstream blocker. I looked into a few workaround and decided to drop it. One workaround is manually do a tokio lookup dns and use tcp socket as well. Another using `socket2` crate.
When keepalive is set, kernel waits for a default of 2 hours before sending the first probe packet. With our ping pong mechanism in agent connection, I think it is redundant.

PR #4501, opened 2026-07-08
Support jq filters for Kafka queue splitting (INT-167)

t4lz:
## What

Adds an optional `jq_filter` to the Kafka queue splitting config, matching the existing support for SQS, GCP Pub/Sub, Azure Service Bus, Redis Pub/Sub, Temporal, and BullMQ.

Closes [INT-167](https://linear.app/metalbear/issue/INT-167/support-jq-filter-for-kafka-as-well).

## How

- `QueueFilter::Kafka` gains `jq_filter: Option<String>`, and `message_filter` becomes optional (a queue can now be split with only a jq filter). Both are verified at config-load time like the other brokers.
- The jq program runs on a JSON representation of the Kafka message record: an object with `topic`, `partition`, `offset`, `timestamp` (ms, when present), `key`, `payload`, and `headers` (object mapping header names to values). `key`, `payload`, and header values are UTF-8 strings, or base64 when not valid UTF-8. There is no standard JSON mapping for Kafka records, so this follows the shape the operator already uses for Kafka message events, and is documented in the config docs.
- New `kafka_jq_filters` connect param, gated on the new `NewOperatorFeature::KafkaQueueSplittingWithJqFilter` so older operators are never sent a filter they would silently ignore.
- Preview environments: new additive `kafkaQueueJqFilters` CRD field (the existing `kafkaQueueFilters` value type is a bare header map and can't grow a jq field without breaking stored resources).
- Regenerated `mirrord-schema.json`.

Operator-side implementation: metalbear-co/operator PR (same branch name).


mirrord/config/src/feature/split_queues.rs
```diff
@@ -653,6 +677,10 @@ pub enum QueueFilter {
     /// For **GCP Pub/Sub**, the JSON representation of [`PubsubMessage`](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage)
     /// us used.
```
skreborn: You didn't change this, but you might as well fix it while you're touching this comment.
```suggestion
    /// is used.
```

mirrord/config/src/feature/split_queues.rs
skreborn, line 706: This is missing concrete entries for brokers other than SQS, GCP, and Kafka. Again, shouldn't be for you to fix, because the previous authors should've added them, but here we are.

skreborn (dismissed):
All good on your part, just marked two things you could tidy up unrelated to your changes.

mirrord/config/src/feature/split_queues.rs
t4lz: Added concrete entries for all the remaining jq-capable brokers in 7238a58 — Azure Service Bus (the five-field object the operator actually builds: `body`, `application_properties`, `message_id`, `content_type`, `subject`), Redis Pub/Sub (payload parsed as JSON), Temporal (the per-task document with `task_type` and the activity/workflow field sets), and BullMQ (the job's `data` parsed as JSON). Each description was verified against the operator's forwarder code rather than the existing docs. Also corrected the Azure Service Bus field-level doc, which claimed the full `ServiceBusMessage` object.

skreborn (approved):

PR #4508, opened 2026-07-08
Add context/namespace-aware mirrord ui API v2 with cluster selector

aviramha:
## What

Adds a **context/namespace selector** to the `mirrord ui` session monitor, backed by a new **`/api/v2`** API, and unifies the wizard's cluster reads under it.

### API versioning
- **v1 frozen** — the existing unversioned `/api/*`, `/ws`, and `/chaos/rules/*` are unchanged. The separately-distributed browser extension keeps working against them.
- **`/api/v2/*` (new)** — clean, context/namespace-aware surface used by the bundled frontend (monitor + wizard). Lives in `mirrord/cli/src/ui/server/v2.rs`.

### Why v2 (multi-tab correctness)
v2 has **no shared server-side "current context"**. Every cluster-touching endpoint takes an optional `?context=` (operator sessions also `?namespace=`) and is served per-request, so two browser tabs can view two different clusters/namespaces at once — something the previous single background watcher couldn't express. A kube client is cached per context so the recurring polls don't rebuild TLS.

### v2 routes
- `local/*` — local sessions: `GET` list/one, `DELETE` to stop, SSE events. Host-global, **polled** (no WebSocket), always shown, and labelled with their own context/namespace (`SessionInfo` gains an additive `context` field).
- `operator/sessions?context&namespace` → `{ context, status, reason?, sessions }` (namespace filtered server-side).
- `kube/contexts` (each with its default namespace), `kube/namespaces?context`, `kube/user?context`, `kube/targets?context&namespace&target_type`, `kube/target-types`, `token`.

### Live-data model
Uniform: **HTTP polling for every resource** (local + operator). The "connected" indicator reflects local-poll health.

### Wizard unification
The config wizard's cluster reads move onto the shared `/api/v2/kube` API (`kube/namespaces`, `kube/targets`, `kube/target-types`) instead of its own `/api/v1/cluster-details` + `namespace/{ns}/targets`. The target-enumeration handlers stay in the wizard module (they still flag the returning user); only `/api/v1/is-returning` remains wizard-specific.

### UI
- Context/namespace picker sits in the header next to the user menu; search moved above the sessions list.
- Removed the operator session detail's "Config" panel (it showed session metadata, not config).

## Testing
- `cargo clippy -- --deny warnings` and `cargo fmt` clean.
- Server unit tests (v1 + v2 auth-gating), v2 serialization tests, and the session-monitor round-trip test pass.
- Monitor + wizard typecheck; `packages/ui` builds.


mirrord/cli/src/ui/server/v2.rs
```diff
@@ -0,0 +1,374 @@
...
+/// Lists namespaces visible in a context. Queries the cluster API server, so the context must be
+/// reachable and the user authorized to list namespaces.
+async fn kube_namespaces(
+    State(state): State<AppState>,
+    Query(query): Query<ContextQuery>,
+) -> V2Result<axum::Json<NamespacesResponse>> {
+    let client = cached_client(&state, query.context.as_deref()).await?;
+    let api: Api<Namespace> = Api::all(client);
+    let namespaces = api
+        .list(&ListParams::default())
```
OneCricketeer: What happens when the cluster operators restricts list-namespaces API from developers, and explicitly requires them to statically define their namespace in every k8s operation? 

Is there a possibility to pull this list from external config or in the meantime inject an optional label selector?

packages/monitor/src/App.tsx
```diff
@@ -97,26 +112,91 @@ export default function App({ theme, isDarkMode, onThemeChange }: MonitorProps)
...
+    if (!effectiveContext) return
+    let cancelled = false
+    setNamespacesLoading(true)
+    api.listNamespaces(effectiveContext)
+      .then(({ namespaces }) => {
+        if (!cancelled) setNamespaces(namespaces)
+      })
+      .catch((err) => {
+        if (!cancelled) {
+          console.error(err)
```
OneCricketeer: Suggestion: Some toast pop-up on a list-namespaces error.

mirrord/cli/src/ui/server/v2.rs
aviramha: That'd be evil but I guess we can support open text input (in any case it'd default to current namespace)

OneCricketeer: > That'd be evil 

I've worked on multi tenant clusters with thousands of namespaces. The RBAC policies were quite strict. Potentially to limit leaks from the R&D department on project names before go to market 😅

aviramha: but you'd still be able to see namespaces you have access to afaik.

mirrord/cli/src/ui/server/v2.rs
```diff
@@ -0,0 +1,374 @@
...
+use tracing::warn;
+
+use super::{
+    AppState, OperatorLockedPort, OperatorQueueSplits, OperatorSessionOwner,
+    OperatorSessionSummary, OperatorSessionTarget, client_for_context, get_session, kill_session,
+    list_sessions, session_events_sse,
+};
+use crate::ui::error::ApiError;
+
+type V2Result<T> = Result<T, ApiError>;
```
aviramha: why not just call it Result?

mirrord/cli/src/ui/server/v2.rs
```diff
@@ -0,0 +1,374 @@
...
+            get(get_session).delete(kill_session),
+        )
+        .route("/local/sessions/{id}/events", get(session_events_sse))
+        .route("/operator/sessions", get(operator_sessions))
+        .route("/kube/contexts", get(kube_contexts))
+        .route("/kube/namespaces", get(kube_namespaces))
+        .route("/kube/user", get(kube_user))
+        // Target enumeration for the config wizard lives in the wizard module (it flags the
+        // returning user), but is served here so the monitor and wizard share one kube API.
+        .route("/kube/targets", get(crate::ui::wizard::list_targets))
```
aviramha: why not move impl here instead of calling the wizard crate?

mirrord/cli/src/ui/server/v2.rs
```diff
@@ -0,0 +1,374 @@
...
+        .route(
+            "/kube/target-types",
+            get(crate::ui::wizard::list_target_types),
```
aviramha: why not move impl here instead of calling the wizard crate?

mirrord/cli/src/ui/server/v2.rs
```diff
@@ -0,0 +1,374 @@
...
+        )
+        .route("/token", get(token))
+}
+
+/// Returns a kube client for `context` (or the kubeconfig current context when `None`), caching it
+/// in [`AppState::clients`] so the recurring operator-session polls don't rebuild TLS each time.
+async fn cached_client(state: &AppState, context: Option<&str>) -> V2Result<Client> {
```
aviramha: not bad, but do we have a mechanism to recreate client if it breaks (I.e auth needs to be re run etc?)

packages/monitor/src/App.tsx
```diff
@@ -97,26 +115,94 @@ export default function App({ theme, isDarkMode, onThemeChange }: MonitorProps)
...
+
   const refreshOperatorSessions = useCallback(() => {
-    api.listOperatorSessions()
-      .then((resp: OperatorSessionsResponse) => {
+    api.listOperatorSessions(effectiveContext, selectedNamespace)
+      .then((resp) => {
         setOperatorSessions(resp.sessions)
-        setWatchStatus(resp.watch_status)
-        const fingerprint = resp.operator_license?.fingerprint
-        if (fingerprint) setLicenseGroup(fingerprint, resp.operator_license?.organization)
```
hank-metalbear, line 113: We need this line to know who is opening mirrord ui.

aviramha: Why?

aviramha: nvm claude explained to me

hank-metalbear (approved):
Verified the operator-license attribution restore: same `operator.spec.license` field path as #4454, now fetched once per context with a cancel guard. LGTM.

PR #4515, opened 2026-07-09
Add a retryable agent error variant

skreborn:
Resolves INT-511.

### DIFF
--- a/mirrord/protocol/src/error.rs
+++ b/mirrord/protocol/src/error.rs
@@ -70,6 +70,26 @@ pub enum ResponseError {
         policy_name: Option<String>,
         reason: String,
     },
+
+    #[error("mirrord system error (code {code}): {message}")]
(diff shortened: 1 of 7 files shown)


mirrord/protocol/src/error.rs
```diff
@@ -70,6 +70,26 @@ pub enum ResponseError {
         policy_name: Option<String>,
         reason: String,
     },
+
+    #[error("mirrord system error (code {code}): {message}")]
+    SystemFailure {
+        can_retry: bool,
+        code: u32,
```
Razz4780, line 77: We can already add const for the first code – agent lost while processing request

Razz4780 (dismissed):

Razz4780 (dismissed):

Razz4780 (approved):

PR #4517, opened 2026-07-09
chaos: Allow the latency direction to be defined in rule effect

gememma:
Issue: https://linear.app/metalbear/issue/COR-1627/allow-latency-direction-to-be-defined-in-rule-effect

- chaos rules for latency can specify read, write or both to apply latency in those directions
- jitter is applied only if the latency for a given direction is not 0
- connection latency is read latency + write latency, plus jitter (if defined)
- chaos routes are now nested under `/api`

_As usual, I recommend reviewing this commit by commit_

### DIFF
--- a/mirrord/intproxy/src/proxies/outgoing/chaos.rs
+++ b/mirrord/intproxy/src/proxies/outgoing/chaos.rs
@@ -22,6 +22,13 @@ use crate::{
     },
 };
 
+/// Helper type for distinguishing between `ChaosEffectLatency.read` and `write` directions.
+#[derive(Debug, Clone, Copy)]
+enum ChaosLatencyDirection {
+    Read,
+    Write,
+}
+
 impl OutgoingProxy {
     /// Checks if the `ChaosRule` (stored in `self.chaos_rx`) applies to either of
     /// `remote_address`, `protocol`, or `hostname`.
@@ -140,7 +147,7 @@ impl OutgoingProxy {
                 ChaosSelector::Tcp {
                     effect: TcpChaosEffect::Latency(effect),
                     ..
-                } => Some(effect.latency_duration()),
+                } => effect.connection_latency_duration(),
                 _ => None,
             },
         );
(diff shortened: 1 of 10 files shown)


mirrord/intproxy/src/proxies/outgoing/chaos.rs
```diff
@@ -22,6 +22,13 @@ use crate::{
     },
 };
 
+/// Helper type for distinguishing between `ChaosEffectLatency.read` and `write` directions.
+#[derive(Debug, Clone, Copy)]
+enum ChaosLatencyDirection {
+    Read,
+    Write,
```
meowjesty, line 29: Missed opportunity to have a varian `One` for `OneDirection`

mirrord/intproxy/src/session_monitor/chaos/rules.rs
```diff
@@ -498,27 +509,66 @@ pub enum FsChaosEffect {
...
     #[serde_as(as = "serde_with::DurationMilliSeconds<u64>")]
     #[serde(rename = "jitter_ms")]
+    #[serde(default, skip_serializing_if = "Duration::is_zero")]
     jitter: Duration,
 }
 
 impl ChaosEffectLatency {
-    pub fn new(delay: Duration, jitter: Duration) -> Self {
-        Self { delay, jitter }
+    pub fn new(read: Duration, write: Duration, jitter: Duration) -> Self {
```
meowjesty: In the changelog you say they cannot both be 0, but what's guarding against that here?

meowjesty: Oh it's in the `TryFrom` impl, but I think it should be here instead, check for both being `is_zero` and return error. You could then impl it in `TryFrom` like:

`ChaosEffectLatency::new(read_ms.unwrap_or_default(), write_ms.unwrap_or_default()...)?;` and have the error be some generic thing that says `ChaosEffectLatency has to specify at least one of write_ms, read_ms, and they both cannot be 0`.

Sort of how `NonZero` works in rust.

meowjesty (changes requested):
mega nit

mirrord/intproxy/src/proxies/outgoing/chaos.rs
gememma: ummmm . you dont no youre directionful ❓

meowjesty (approved):
👍

PR #4520, opened 2026-07-09
Stable rust prep work

0x00A5:
### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/cli/src/ci/stop.rs
+++ b/mirrord/cli/src/ci/stop.rs
@@ -1,4 +1,5 @@
 #![cfg_attr(windows, allow(unused))]
+use itertools::Itertools;
 use mirrord_progress::{Progress, ProgressTracker};
 use tokio::process::Command;
 use tracing::Level;
@@ -147,11 +148,11 @@ impl CiStopCommandHandler {
 
         intproxies_killed
             .into_iter()
-            .collect::<Result<(), _>>()
-            .and(sidecars_removed.into_iter().collect::<Result<(), _>>())
-            .and(extproxies_killed.into_iter().collect::<Result<(), _>>())
-            .and(sidecars_killed.into_iter().collect::<Result<(), _>>())
-            .and(users_killed.into_iter().collect::<Result<(), _>>())?;
+            .try_collect::<_, (), _>()
+            .and(sidecars_removed.into_iter().try_collect::<_, (), _>())
+            .and(extproxies_killed.into_iter().try_collect::<_, (), _>())
+            .and(sidecars_killed.into_iter().try_collect::<_, (), _>())
+            .and(users_killed.into_iter().try_collect::<_, (), _>())?;
 
         MirrordCiStore::remove_file().await?;
         progress.success(None);
--- a/mirrord/cli/src/ci/stop.rs
+++ b/mirrord/cli/src/ci/stop.rs
@@ -147,11 +147,11 @@ impl CiStopCommandHandler {
 
         intproxies_killed
             .into_iter()
-            .try_collect::<()>()
-            .and(sidecars_removed.into_iter().try_collect::<()>())
-            .and(extproxies_killed.into_iter().try_collect::<()>())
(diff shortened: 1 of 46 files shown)


mirrord/error-util/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mirrord-error-util"
```
iniw: Do we plan on adding more stuff to this crate? If so we should probably call it `mirrord-nightly-polyfill` or something like that and keep the struct names/paths the same (i.e: `std::error::Reporter` -> `mirrord_nightly_polyfill::error::Reporter`) so that it's easier to port once/if it lands on stable.

mirrord/cli/src/ci/stop.rs
```diff
@@ -147,11 +147,11 @@ impl CiStopCommandHandler {
 
         intproxies_killed
             .into_iter()
-            .try_collect::<()>()
-            .and(sidecars_removed.into_iter().try_collect::<()>())
-            .and(extproxies_killed.into_iter().try_collect::<()>())
-            .and(sidecars_killed.into_iter().try_collect::<()>())
-            .and(users_killed.into_iter().try_collect::<()>())?;
+            .collect::<Result<(), _>>()
```
meowjesty: Why not use `itertools`?

https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.try_collect

rustfmt.toml
```diff
@@ -1,7 +1,2 @@
 edition = "2024"
-group_imports = "StdExternalCrate"
-comment_width = 100
-wrap_comments = true
-format_code_in_doc_comments = true
-imports_granularity = "Crate"
```
meowjesty, line 6: Losing this is the saddest shit ever, wonder if there's a way to keep this nightly, since it's an external tool, not code itself.

mirrord/error-util/Cargo.toml
0x00A5: Yeah, I will need to add try and try residual related hand-roll stuff for layer. A bigger clean up than this PR. This is a good idea, I will update the crate name and paths.

rustfmt.toml
0x00A5: This is indeed sad. Keeping this file as-is requires everyone runs `cargo +nightly fmt` rather than using the stable one, or it will break. rustfmt.toml is not toolchain specific sadly.

mirrord/cli/src/ci/stop.rs
0x00A5: Replaced with try_collect from itertools

mirrord/error-util/Cargo.toml
0x00A5: Done

rustfmt.toml
meowjesty: > Probably everyone already knows, but I'd still like to point out a fun fact: unstable rustfmt options could be enabled by command-line arguments, on stable toolchain, like this.

`cargo fmt --config imports_granularity=Module`

Also, can't we `cargo binstall rustfmt-nightly` and keep it? So everyone runs `rustfmt` instead of doing `cargo fmt`, no need to keep a nightly rust toolchain installed.

meowjesty: `cargo fmt -- --config imports_granularity=Module`

iniw: This would need to be configured on a per-editor basis though, so we'll definitely get misformatted stuff unless we enforce it on CI because people won't configure it properly.

meowjesty: https://web.archive.org/web/20250904025109/https://github.com/nik-rev/nightly-trick

0x00A5: I can revert the rustfmt config change now, and try this linked trick when we are on stable finally.

0x00A5: reverted

meowjesty (approved):
👍

PR #4521, opened 2026-07-10
List active sessions through the operator's active-sessions API

skreborn:
Resolves INT-504.


iniw (approved):
LGTM (Levente Goes To Moon)

iniw (approved):

PR #4536, opened 2026-07-13
layer+cli: use `MIRRORD_LOG` for logging instead of `RUST_LOG`

iniw:
The reasoning behind this change is to not accidentally pollute the console with mirrord logs when the user runs their rust app under mirrord.

As an example, imagine that we're in the operator repo and want to run `operator-service`:

```bash
mirrord exec -f .mirrord/operator.yaml -- cargo run -p operator-service
```

This will run using the default `RUST_LOG` from the deployed operator.

Now let's say we want to increase the log verbosity for one reason or another. We press `↑` on our terminal to re-run
the last command and prefix it with `RUST_LOG` to tune the log filter:

```bash
RUST_LOG=debug mirrord exec -f .mirrord/operator.yaml -- cargo run -p operator-service
```

And, suddenly, a _lot_ of mirrord internal logs show up! Ahhhhh! Help!!!!


cristeigabriela (approved):
vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini vini

PR #4541, opened 2026-07-14
MBE-1975: Add sessions-manager client and wire serverless connection flow

itsamegraf:
This PR adds the sessions-manager client/protocol layer and wires mirrord’s CLI, intproxy, and agent runtime to use it for the new `serverless` target flow.

#### Sessions-manager client and protocol
- Adds `mirrord/sessions-manager/protocol` with the control-plane payloads and role definitions.
- Adds `mirrord/sessions-manager/client` with:
  - sessions-manager room derivation
  - registration / handoff handling
  - websocket transport wrapping for protocol traffic
  - client-side error handling
- Adds support for using sessions-manager-backed connections from both the CLI and the agent side.

#### CLI and target configuration
- Adds a new `serverless` target type in `mirrord/config`.
- Updates target parsing, display, analytics, and config verification to recognize `serverless/{service}[/container/{hostname}]`.
- Updates CLI connection setup so `serverless` targets connect through sessions-manager instead of the existing operator/direct paths.

#### Intproxy and agent runtime
- Updates `intproxy` to consume the sessions-manager-backed agent connection path.
- Updates the agent connection wrapper to support already-established protocol connections.
- Adds a sidecar-aware agent startup path so the agent can serve sessions-manager-backed traffic.
- Extends agent/CLI error handling to surface sessions-manager client failures cleanly.

### Notes
This PR focuses on the baseline sessions-manager connection flow and the runtime wiring needed to use it end-to-end. It does **not** include the serverless bootstrap packaging work or the standalone-agent transport abstraction work.

### Quality Checklist:

  documentation)


itsamegraf:
@greptileai

itsamegraf:
@greptileai

Cargo.toml
```diff
@@ -302,6 +312,16 @@ xmas-elf = "0.10"
...
 
+# used by sessions-manager client & protocol
+# `nextdata-tech/rust-socketio` fork of `tyler/feat-add-features-for-other-tls-types` branch to support rustls
+rust_socketio = {
+    git = "https://github.com/metalbear-co/rust-socketio.git",
+    # HEAD of branch mb-develop
+    rev = "c2157e359e1b0a8f5d96e3b366caeb57feefe7f1",
+    default-features = false,
+    features = ["async", "rustls-tls-native-roots"]
+}
```
Razz4780: Uhm, why do we need this?

mirrord/sessions-manager/protocol/src/lib.rs
```diff
@@ -0,0 +1,233 @@
+use std::borrow::Cow;
```
Razz4780: Crate-level doc would be nice

mirrord/sessions-manager/client/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub mod connection;
```
Razz4780: Crate-level doc would be nice

mirrord/sessions-manager/client/src/connection.rs
```diff
@@ -0,0 +1,341 @@
...
+use tokio_tungstenite::connect_async;
+use tokio_util::sync::CancellationToken;
+
+use crate::{
+    envs::sessions_manager_namespace, error::SessionsManagerClientError,
+    websocket::BinaryWebSocketConnection,
+};
+
+pub const SESSIONS_MANAGER_URL_ENV: &str = "MIRRORD_SESSIONS_MANAGER_URL";
+const SESSIONS_MANAGER_URL_DEFAULT: &str = "http://localhost:4971";
```
Razz4780: Wait what is this 👀 
Why are we running something on localhost

Razz4780:
@greptileai

mirrord/cli/src/connection.rs
```diff
@@ -291,6 +294,33 @@ pub(crate) async fn create_and_connect<R: Reporter>(
...
+        let user_session_id = std::env::var("MIRRORD_SESSION_ID")
+            .unwrap_or_else(|_| uuid::Uuid::new_v4().to_string());
+        let agent_replica_filter = target.sessions_manager_target_replica_id();
+        let connect_info = SessionsManagerConnectInfo {
+            service,
+            environment: config
+                .target
+                .namespace
+                .clone()
+                .unwrap_or_else(|| "default".to_owned()),
```
Razz4780, line 315: Isn't namespace required? Do we have a concept of "default" namespace in serverless?

mirrord/cli/src/main.rs
```diff
@@ -261,6 +261,7 @@
 
 #![warn(clippy::indexing_slicing)]
 #![deny(unused_crate_dependencies)]
+#![recursion_limit = "256"]
```
Razz4780, line 264: 👀

mirrord/protocol-io/src/websocket.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Adapts an already-connected binary WebSocket to [`Connection`].
+pub fn connection<S, E>(socket: WebSocketStream<S>) -> Connection<E>
+where
+    S: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    E: ProtocolEndpoint + Send + Unpin + 'static,
+{
+    Connection::from_channel(WebSocketChannel::<S, E>::new(socket))
+}
+
+/// Adapts a binary WebSocket to the stream-and-sink transport expected by [`Connection`].
```
Razz4780, line 38: Uhm, why do we need this? The codebase should already have all needed protocol adapters for WebSocket – it's how we talk to the operator

mirrord/sessions-manager/client/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+mod assignments;
```
Razz4780, line 1: Crate doc would be nice

mirrord/sessions-manager/protocol/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+pub use control_plane::{
```
Razz4780, line 1: Crate doc would be nice

mirrord/sessions-manager/protocol/src/data_plane.rs
```diff
@@ -0,0 +1,205 @@
...
+
+impl<'de> Deserialize<'de> for DataPlaneAuthorization {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        String::deserialize(deserializer).map(Self::new)
+    }
+}
+/// Relative data-plane URI resolved against the control-plane URL used by the receiving peer.
```
Razz4780, line 67:
```suggestion
}

/// Relative data-plane URI resolved against the control-plane URL used by the receiving peer.
```

mirrord/sessions-manager/protocol/src/data_plane.rs
```diff
@@ -0,0 +1,205 @@
...
+///
+/// The current sessions-manager contract requires this endpoint to be an absolute path on the
+/// control-plane origin. The authorization in an assignment is forwarded to the resolved endpoint.
+/// Supporting another origin requires an explicit protocol change rather than an ambiguous URI.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct DataPlaneEndpoint(Uri);
+
+impl DataPlaneEndpoint {
+    pub fn new(uri: Uri) -> Result<Self, SessionsManagerProtocolError> {
```
Razz4780, line 76: Do we need this validation here? Like, why do we need to verify the url in the client code? It comes from the session manager. If it was hijacked, here we enforce that the url leads to the same hijacked domain. In either way, a doc would be nice (as in – where does this uri come from)

mirrord/sessions-manager/protocol/src/data_plane.rs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            return Err(SessionsManagerProtocolError::InvalidDataPlaneEndpointRecv(
+                uri,
+            ));
+        }
+
+        Ok(Self(uri))
+    }
+
+    pub fn resolve(
```
Razz4780, line 102: A doc would be nice. Also, I don't really get the checks here either. What is the attack scenario?

mirrord/sessions-manager/protocol/src/data_plane.rs
```diff
@@ -0,0 +1,205 @@
...
+        serializer.serialize_str(self.as_str())
+    }
+}
+
+impl<'de> Deserialize<'de> for DataPlaneEndpoint {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let uri: Uri = String::deserialize(deserializer)
```
Razz4780, line 145: There's `http_serde`, we already use it in `mirrord-protocol` 👍

mirrord/sessions-manager/client/src/retry.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+        None => {
+            tokio::select! {
+                _ = cancellation.cancelled() => {
+                    Err(SessionsManagerClientError::Cancelled)
+                }
+                output = future => Ok(output),
+            }
+        }
+    }
```
Razz4780, line 53:
```suggestion
    let timeout = OptionFuture::from(deadline.map(tokio::time::sleep_until));
    tokio::select! {
        _ = cancellation.cancelled() => Err(SessionsManagerClientError::Cancelled),
        Some(()) = timeout => Err(SessionsManagerClientError::OperationTimeout),
        output = future => Ok(output),
    }
```

:D

mirrord/sessions-manager/client/src/retry.rs
```diff
@@ -0,0 +1,160 @@
...
+const INITIAL_RETRY_DELAY_MS: u64 = 100;
+// This caps the base exponential delay; `tokio-retry` adds jitter afterward, so the actual delay
+// can be somewhat larger.
+const MAX_RETRY_DELAY: Duration = Duration::from_secs(5);
+
+pub(crate) type RetryDelays = Box<dyn Iterator<Item = Duration> + Send>;
+
+pub(crate) fn init_retry_policy() -> RetryDelays {
+    Box::new(
+        ExponentialBackoff::from_millis(INITIAL_RETRY_DELAY_MS)
```
Razz4780, line 18: `from_millis` is used both as base and multiplier. The `ExponentialBackoff` constructed here would yield 0.1ms, 10s, 1000s, ... (yes, it is cursed)
With 5s cap, this is 100ms, 5s, 5s, ...

I think you want sth like `ExponentialBackoff::from_millis(2).factor(50).max_delay(Duration::from_secs(5)).map(jitter)` – 0.1s, 0.2s, 0.4s, 0.8s, 1.6s, 3.2s, 5s, 5s, ...

mirrord/sessions-manager/client/src/retry.rs
```diff
@@ -0,0 +1,160 @@
...
+    cancellation: &CancellationToken,
+    deadline: Option<Instant>,
+) -> Result<Duration, SessionsManagerClientError> {
+    let delay = retry_delays
+        .next()
+        .expect("exponential backoff strategy is unbounded");
+    wait_retry(cancellation, deadline, delay).await?;
+    Ok(delay)
+}
+
```
Razz4780, line 80: Still reading, but these things look like you're not using [`RetryIf`](https://docs.rs/tokio-retry/latest/tokio_retry/struct.RetryIf.html) for retrying stuff 👀
Like

```rust
let operation = RetryIf::start(...);
tokio::select! {
    _ = cancellation.cancelled() => Err(SessionsManagerClientError),
    result = operation => result,
}
```

mirrord/sessions-manager/client/src/error.rs
```diff
@@ -0,0 +1,77 @@
+use mirrord_sessions_manager_protocol::SessionsManagerProtocolError;
+use url::Url;
+
+#[derive(thiserror::Error, Debug)]
+pub enum SessionsManagerClientError {
```
Razz4780, line 5: Errors stored as strings are sad :<

mirrord/sessions-manager/client/src/environment.rs
```diff
@@ -0,0 +1,36 @@
+//! Definitions of environment variables used to configure sessions-manager connections.
+//!
+//! If you want to add some more, please do it here.
```
Razz4780, line 3: Heh, we already have a thing like this for the agent configuration env vars, as a separate crate under `.../mirrord/agent`. It also says that all agent configs should be there. Now how do we solve this?

mirrord/sessions-manager/client/src/credentials.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// [`None`] when no token is set, which is the case for a directly reachable
+    /// sessions-manager.
+    pub fn from_env() -> Result<Option<Self>, SessionsManagerClientError> {
+        let token = match std::env::var(SESSIONS_MANAGER_AUTH_TOKEN_ENV) {
+            Ok(token) => token,
+            Err(VarError::NotPresent) => return Ok(None),
+            Err(error) => return Err(error.into()),
+        };
+
+        let name = std::env::var(SESSIONS_MANAGER_AUTH_HEADER_ENV)
```
Razz4780, line 73: This is the only place where `SESSIONS_MANAGER_AUTH_HEADER_ENV` is used. Do we need all this logic and tests for verifying that the header name is not in the arbitrary reserved list? Do you know of any use case where someone would want to pass the auth token in a different header?

mirrord/sessions-manager/client/src/data_plane/websocket.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let (stream, response) =
+        tokio::time::timeout(WEBSOCKET_UPGRADE_TIMEOUT, connect_async(request))
+            .await
+            .map_err(|_| SessionsManagerClientError::WebSocketUpgradeTimeout)??;
+
+    tracing::debug!(
+        status = %response.status(),
+        "WebSocket data-plane connection established"
+    );
```
Razz4780, line 84: Could log url and elapsed time as well

mirrord/sessions-manager/client/src/control_plane/api.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn decode_v1_event(
+        &self,
+        event: Event,
+    ) -> Result<Option<ControlPlaneEvent>, SessionsManagerClientError> {
+        match event.event.parse::<ControlPlaneEventName>() {
+            Ok(ControlPlaneEventName::Assignment) => Ok(Some(ControlPlaneEvent::Assignment(
+                serde_json::from_str(&event.data)?,
+            ))),
+            Ok(ControlPlaneEventName::Superseded) => Ok(Some(ControlPlaneEvent::Superseded)),
+            Err(_) if event.event == "message" && event.data.is_empty() => Ok(None),
```
Razz4780, line 86: What is this 👀

mirrord/sessions-manager/client/src/control_plane/subscriber.rs
```diff
@@ -0,0 +1,502 @@
...
+        &self,
+        client: &HttpControlPlaneClient,
+        cancellation: &CancellationToken,
+    ) -> Result<ControlPlaneEventStream, SessionsManagerClientError>;
+
+    fn extract(&self, event: ControlPlaneEvent)
+    -> Result<Self::Output, SessionsManagerClientError>;
+}
+
+pub(crate) struct ControlPlaneSubscriber<S> {
```
Razz4780, line 33: Doc pls

mirrord/sessions-manager/client/src/control_plane/subscriber.rs
```diff
@@ -0,0 +1,502 @@
...
+            cancellation,
+            events: None,
+            retry_delays: init_retry_policy(),
+            retry_initialization,
+            opened_once: false,
+            terminal: false,
+        }
+    }
+
+    #[allow(dead_code)]
```
Razz4780, line 66: 👀

mirrord/sessions-manager/client/src/control_plane/subscriber.rs
```diff
@@ -0,0 +1,502 @@
...
+        self.next_with_deadline(Some(deadline))
+            .await
+            .unwrap_or_else(|| Err(SessionsManagerClientError::Cancelled))
+    }
+
+    async fn next_with_deadline(
+        &mut self,
+        deadline: Option<Instant>,
+    ) -> Option<Result<S::Output, SessionsManagerClientError>> {
+        if self.terminal || self.cancellation.is_cancelled() {
```
Razz4780, line 100: `CancellationToken` is used so much in this crate, and I'm not sure why. Most of the things here are just polled. Usually the token is passed to a background task and managed at the top level.
Like, consider this struct. Adding a cancellation token as a field only introduces more complexity. It does not spawn background tasks itself. If you no longer care about `next_with_deadline` output, you can just stop polling it 👀

mirrord/sessions-manager/client/src/assignments.rs
```diff
@@ -0,0 +1,242 @@
...
+/// TTL bounds that failure mode instead of requiring it to never happen.
+const CONNECTING_ASSIGNMENT_TTL: Duration = Duration::from_secs(2 * 60);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum LocalAssignmentState {
+    Connecting { accepted_at: Instant },
+    Connected { completed_at: Instant },
+}
+
+#[derive(Debug, Default)]
```
Razz4780, line 66: This really needs a doc. I did not expect to see an lru cache in this PR 😅

Razz4780 (changes requested):

Razz4780 (changes requested):
Hmmm

mirrord/cli/src/connection.rs
itsamegraf: `namespace` comes from reusing the same `TargetFileConfig` from `mirrord-config`. In that configuration, `namespace` is optional, so we need to support cases where it is not provided.

To normalize SM control-plane calls to the `/v1/env/{environment}/service/{service}/assignments` route, the simplest solution was to fall back to `"default"` whenever no namespace is specified. This ensures that the value is always present.

That said, the concept of a Kubernetes-style namespace does not really exist in serverless. In this context, the field is effectively used as a folder-like grouping mechanism, allowing customers to distinguish between different environments running the same mirrord serverless targets—for example, `"dev"` and `"staging"`.

Cargo.toml
itsamegraf: removed

mirrord/cli/src/main.rs
itsamegraf: 🫨  removing

mirrord/protocol-io/src/websocket.rs
itsamegraf: I dont recall anymore why was it originally necessary, it was a few-days worth of fighting with the ai until it agreed this impl has no what-ifs.
I asked chat to do a comparison, it noted a few subtle "improvements" over the operator one's.
https://chatgpt.com/share/e/6aa7ef5d-c448-83ed-a0bd-d1dad974938d

bottom line - will move back to the operator's one, if it's good for operator then it should be good for us. ai might point out some issues but i prefer not duplicating impl unnecesarily.

itsamegraf: while working on implementing this, its preety straight forward, the operator one is only implemented for intproxy->agent, ill expand to to support both directions. (that is what started the horendous websocket.rs of this comment which i wanted to live alongside the operator's one)

itsamegraf: well, this was quite a refactor to make it work for both cases, I've split it into a separate [ticket](https://linear.app/metalbear/issue/MBE-2138/make-operatorconnection-usable-for-both-sides-of-serverless) and [PR](https://github.com/metalbear-co/mirrord/pull/4885)

itsamegraf: well, this was quite a refactor to make it work for both cases, I've split it into a separate [ticket](https://linear.app/metalbear/issue/MBE-2138/make-operatorconnection-usable-for-both-sides-of-serverless) and [PR](https://github.com/metalbear-co/mirrord/pull/4885)

mirrord/sessions-manager/protocol/src/data_plane.rs
itsamegraf: Overall you're right. It guards against malformed or incorrectly constructed assignments causing the client to forward the per-assignment bearer credential to an arbitrary origin. The endpoint comes from the assignment, but under the current protocol it is only allowed to choose a path; the configured sessions-manager URL chooses the origin. This is deliberately not a defense against a compromised sessions-manager origin—if that origin is compromised, the session is already compromised.
I improved the docs there and explained for each check what it guards against.

mirrord/cli/src/main.rs
itsamegraf: found the culprit.
running `cargo clippy --all-targets --all-features -- -D warnings`
caused
```
    Checking mirrord v3.257.0 (/Users/daniel/git/mirrord/mirrord/cli)
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`mirrord`)
  = note: query depth increased by 130 when computing layout of `{async fn body of ci::container::<impl at mirrord/cli/src/ci/container.rs:37:1: 37:39>::handle()}`

error: could not compile `mirrord` (bin "mirrord") due to 1 previous error
```
im checking what caused it

itsamegraf:
```
AgentConnection::restart
  -> RetryIf::start(...)
    -> retry closure
      -> run_until_cancelled(AgentConnection::new(...))
        -> AgentConnectInfo::SessionsManager
          -> IntproxyClient::connect
            -> assignment HTTP/SSE handling
            -> WebSocket connection
```
yup me code pumping cli stack

itsamegraf: boxed the intproxy::connect call sites

mirrord/sessions-manager/client/src/retry.rs
itsamegraf: Thank you for the suggestion, we now use `RetryIf` everywhere, wrapped in your `select!` sketch. It's packaged as one `RetryBudget` type because three things had to come with it:

1. **Backoff must survive across `RetryIf` runs** - the subscriber's retry unit is one `next()` call, not one long operation, so `SharedRetryDelays(Arc<Mutex<…>>)` shares one sequence, reset explicitly only when a retry actually paid off (a decoded event, not just an opened connection).
2. **Cancellation + deadline written once** in `run_until`/`run_interruptible` instead of at each call site.
3. **The `If` predicate is the real win** - `ControlPlaneAttemptError::should_retry` replaced the hand-rolled `handle_error` + `can_retry` bool + `loop { continue }`.

mirrord/sessions-manager/client/src/environment.rs
itsamegraf: env vars will move to agent side only, local side will have to suffice with mirrord.json target settigns

mirrord/sessions-manager/client/src/credentials.rs
itsamegraf: well, it was aviram's work in #4811 that merged in here. I dont see a real reason to keep it, i don't mind removing the customizability of the header (and the checks and tests etc.), anyway its only meant for during our poc states, so ill do that

mirrord/sessions-manager/client/src/control_plane/api.rs
itsamegraf: it was handling of keep alive events in case they slipped, but its not necessary, ill remove it

mirrord/sessions-manager/client/src/control_plane/subscriber.rs
itsamegraf: cancellationtoken was purged to one usage at the top level :)

mirrord/agent/src/entrypoint.rs
```diff
@@ -1137,6 +1176,110 @@ async fn start_agent(args: Args) -> AgentResult<()> {
...
+
+    let service = envs::REMOTE_SERVICE
+        .try_from_env()
+        .expect("String environment variables are infallible")
+        .ok_or::<AgentError>(SessionsManagerClientError::MissingConfigService.into())?;
+    let environment = envs::REMOTE_ENVIRONMENT
+        .try_from_env()
+        .expect("String environment variables are infallible")
+        .unwrap_or_else(|| "default".to_owned());
+    let replica_id = envs::REMOTE_SERVICE_REPLICA
```
Razz4780: Shouldn't this be a random UUID? Who's going to assign IDs to companion instances and set the env vars, for example when the ECS task is scaled?

changelog.d/+mbe-2075-companion-file-paths.internal.md
```diff
@@ -0,0 +1,2 @@
+Fixed remote file operations in the agent's workload companion mode, where absolute
+paths were resolved against the agent process working directory.
```
Razz4780, line 2: I don't get this. Companion was using a `FileManager` without any path resolver. So absolute paths from the layer were just used as they were. Now the companion's `FileManager` uses a path resolver with `/` as base. Which should be noop. What's the fix? 👀

mirrord/cli/src/connection.rs
Razz4780: Yeah I get that. But why does this simple solution make sense?
Why not error out?

mirrord/agent/env/src/envs.rs
```diff
@@ -59,6 +59,18 @@ pub const POD_IPS: CheckedEnv<Vec<IpAddr>> = CheckedEnv::new("MIRRORD_AGENT_POD_
 /// Should follow `tracing` format, e.g `mirrord=trace`.
 pub const LOG_LEVEL: CheckedEnv<String> = CheckedEnv::new("RUST_LOG");
 
+/// Identifies the serverless service represented by a workload-companion agent.
+pub const REMOTE_SERVICE: CheckedEnv<String> = CheckedEnv::new("MIRRORD_REMOTE_SERVICE");
+
+/// Identifies the workload-companion replica represented by this agent.
```
Razz4780: Aight, I'm getting lost a bit. What is workload-companion replica? I see that instance id is a v4 UUID generated when connecting to the sessions manager. What is this then? 👀

mirrord/sessions-manager/client/src/client/intproxy.rs
```diff
@@ -0,0 +1,157 @@
...
+    error::SessionsManagerClientError,
+    retry::{RetryBudget, with_deadline},
+};
+
+/// Identifies the sessions-manager allocation requested by an intproxy.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SessionsManagerConnectInfo {
+    pub service: String,
+    pub environment: String,
+    pub agent_replica_filter: Option<String>,
```
Razz4780: This needs a doc

mirrord/config/src/target/serverless.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use super::{FAIL_PARSE_DEPLOYMENT_OR_POD, FromSplit};
+use crate::config::{ConfigError, Result};
+
+/// - `serverless/{service-name}[/container/{replica-id}]`;
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Hash, Debug, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct ServerlessTarget {
+    pub serverless: String,
+    // container holds the hostname of the container
```
Razz4780: I don't really understand this

Razz4780 (commented):
Not finished yet. I see that the PR has grown too :V

mirrord/agent/env/src/envs.rs
itsamegraf: replica is one instance of a workload, e.g. ecs task.
workload companion is the agent in serverless scenarios running within the container.
that agent runs within each of these task replicas and therefore refered as "workload companion replica".

"REMOTE_SERVICE_REPLICA" (`replica_id`) is a user customizable id per replica, defaults to "hostname".
`agent_instance_id` is a v4 uuid that is generated on workload companion init, its just a way to differentiate between connections of the same replica id (workload restarted/image replaced/etc.).
these 2 are used as a unique id for a serverless agent against the sm control plane.

these are explain in the baseline RFC0014:
> In RFC 0013's identity taxonomy, (environment, service) plays the role of service identity, user_session_id plays session identity, (replica_id, agent_instance_id) identifies an agent instance, and intproxy_connection_id identifies one intproxy connection.

from [RFC0014#control-plane-subscription-and-assignment-flow](https://github.com/metalbear-co/rfcs/blob/main/text/0014-serverless-remote-execution-baseline.md#control-plane-subscription-and-assignment-flow)

Razz4780: Ok, thanks. So who's going to handle assigning these replica ids automatically when the target task is horizontally scaled? Like, we expect it to be set in the env at startup

itsamegraf: I understand from your review its quite complex to follow through this solely from the code, i'll introduce some Identifier type with proper explanations in the code to make it easier

itsamegraf: > Ok, thanks. So who's going to handle assigning these replica ids automatically when the target task is horizontally scaled? Like, we expect it to be set in the env at startup

We don't have an implicit support for when the task will be horizontally scaled. In the AWS jargon, its an ecs **Service** that controls the scale-out and scale-in, whether static or dynamic. ECS Tasks are these replicas. 
after a bit of research, the `TaskARN` is the ultimate "replica-unique-id", I will keep `REMOTE_SERVICE_REPLICA` as a "customer controlled" per replica id, but will fallback to retrieving the TaskARN as implicit replica_id

mirrord/config/src/target/serverless.rs
itsamegraf: old doc 😞

changelog.d/+mbe-2075-companion-file-paths.internal.md
itsamegraf: This code came from https://github.com/metalbear-co/mirrord/pull/4864/changes
It did have a FileManager, through the shared `ClientConnectionHandler`, which was using `FileManager::new(file_pid)` with implicit file_pid=1. it should've worked, but for some reason resulted with the initial `/` stripped out 🤷 explicitly noting the `/` fixed it - you can read the full discussion [here](https://linear.app/metalbear/issue/MBE-2075/fix-file-reads-in-workload-companion-mode#comment-df0cbc38)

mirrord/cli/src/connection.rs
itsamegraf: to make "foldering" an opt-in feature while maintaining a single control plane endpoint `/v1/env/{environment}/service/{service}/assignments`

mirrord/agent/src/workload_companion/replica_id.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let mut request = hyper::Request::get(uri);
+    for (name, value) in headers {
+        request = request.header(*name, *value);
+    }
+    let request = request
+        .body(Empty::<Bytes>::new())
+        .expect("URI is valid and headers are static");
+
+    let response = Client::builder(TokioExecutor::new())
```
Razz4780, line 114: Kinda weird to use the legacy hyper Client, but ok 😅

mirrord/sessions-manager/client/src/data_plane/websocket.rs
```diff
@@ -0,0 +1,108 @@
...
+        .map_err(|_| SessionsManagerClientError::InvalidAuthorization)?;
+    authorization.set_sensitive(true);
+    request.headers_mut().insert(AUTHORIZATION, authorization);
+
+    let connector = HttpsConnectorBuilder::new()
+        .with_webpki_roots()
+        .https_or_http()
+        .enable_http1()
+        .build();
+    let client: HyperClient<_, Empty<Bytes>> =
```
Razz4780, line 78: You already depend on reqwest, why use legacy hyper client?

Razz4780 (approved):

Razz4780 (approved):
All right let's roll

mirrord/agent/src/workload_companion/replica_id.rs
itsamegraf: only because the ai insisted not to add reqwest dependancy 😓 but hyper legacy feature, sure...

mirrord/sessions-manager/client/src/data_plane/websocket.rs
itsamegraf: legit, will swap

itsamegraf: didnt swap, it requires additional changes to operatorconnection, which i prefer not to do but use it as-is

PR #4542, opened 2026-07-14
MBE-1977: Package the remote bootstrap flow

itsamegraf:
## Description

This PR adds the packaging and build pipeline for the `mirrord-remote-bootstrap` library, which is responsible for launching the sidecar agent from the serverless/remote bootstrap flow.

#### New remote bootstrap crate
- Adds `mirrord/remote/bootstrap` as a Linux-only `cdylib`.
- The bootstrap:
  - initializes tracing
  - removes `LD_PRELOAD` before spawning the agent
  - extracts the embedded agent binary to disk
  - spawns the agent in `sidecar` mode
  - ensures the spawned process receives `SIGTERM` if the bootstrap exits

#### Embedded agent artifact extraction
- Adds build-time wiring so the bootstrap crate gets the path to the built agent binary through `MIRRORD_AGENT_BINARY`.
- Adds extraction helpers that:
  - reuse an existing extracted binary if present
  - otherwise write the embedded bytes to the configured target path
  - set executable permissions on the extracted binary

#### Workspace/build integration
- Adds `mirrord/remote/bootstrap` to the workspace.
- Excludes `mirrord/remote` from the main workspace path so the bootstrap packaging flow can be built separately.
- Updates `Cargo.lock` for the new crate dependency set.

#### xtask support
- Adds a new `xtask` command for building the remote bootstrap flow.
- The command:
  - builds `mirrord-agent` for the chosen Linux target
  - passes the resulting agent binary path into the bootstrap build
  - builds `mirrord-remote-bootstrap` as a shared library artifact
- Supports both debug and release builds.

### Quality Checklist:

  documentation)


itsamegraf:
CI passes - https://github.com/metalbear-co/mirrord/actions/runs/29422087130

itsamegraf:
https://github.com/metalbear-co/mirrord/actions/runs/29432353483

itsamegraf:
@greptileai

.github/workflows/ci.yaml
```diff
@@ -1,3 +1,4 @@
+# this change will be removed - intended to make mbe-1977 run the full ci
```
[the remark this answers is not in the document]
itsamegraf: will be removed before merge

mirrord/remote/bootstrap/build.rs
```diff
@@ -0,0 +1,22 @@
+fn should_skip_build_req() -> bool {
+    std::env::var("RUSTDOCFLAGS").is_ok() || std::env::var("CLIPPY_ARGS").is_ok()
+}
+
+fn binary_env(var_name: &str) -> String {
+    if should_skip_build_req() {
+        // use bs value
```
Razz4780: Use dummy value maybe? 😆

mirrord/remote/bootstrap/build.rs
```diff
@@ -0,0 +1,22 @@
+fn should_skip_build_req() -> bool {
```
Razz4780: Not sure what's the purpose of this function

mirrord/remote/bootstrap/src/error.rs
```diff
@@ -0,0 +1,10 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum RemoteBootstrapError {
+    #[error("IO Error: {0}")]
+    IO(#[from] std::io::Error),
+    #[error("Null error {0}")]
+    Null(#[from] std::ffi::NulError),
+}
+pub(crate) type Result<T> = std::result::Result<T, RemoteBootstrapError>;
```
Razz4780, line 13:
```suggestion
}

pub(crate) type Result<T> = std::result::Result<T, RemoteBootstrapError>;
```

mirrord/remote/bootstrap/src/error.rs
```diff
@@ -0,0 +1,10 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum RemoteBootstrapError {
+    #[error("IO Error: {0}")]
+    IO(#[from] std::io::Error),
+    #[error("Null error {0}")]
+    Null(#[from] std::ffi::NulError),
+}
```
Razz4780, line 11:
```suggestion
#[derive(Debug, Error)]
pub enum RemoteBootstrapError {
    #[error(transparent)]
    IO(#[from] std::io::Error),
    #[error(transparent)]
    Null(#[from] std::ffi::NulError),
}
```

mirrord/remote/bootstrap/src/extract.rs
```diff
@@ -0,0 +1,77 @@
...
+    io::Write,
+    os::unix::fs::PermissionsExt,
+    path::{Path, PathBuf},
+    process,
+};
+
+use crate::error::Result;
+
+const AGENT_BINARY_ENV: &str = "MIRRORD_REMOTE_AGENT_SIDECAR_BINARY";
+const DEFAULT_AGENT_BINARY: &str = "/tmp/mirrord/mirrord-agent";
```
Razz4780: Can we use [`NamedTempFile`](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html) instead?

mirrord/remote/bootstrap/src/extract.rs
```diff
@@ -0,0 +1,77 @@
...
+        .unwrap_or_else(|_| PathBuf::from(default_path));
+
+    write_binary(&target_path, bytes, binary_name)?;
+
+    Ok(target_path)
+}
+
+fn write_binary(target_path: &Path, bytes: &[u8], binary_name: &str) -> Result<()> {
+    if target_path.exists() {
+        tracing::debug!(binary = %target_path.display(), %binary_name, "Reusing existing binary");
```
Razz4780: I'm not sure what's the reuse for. Is it for testing? If it's prod flow, we need to verify that the agent binary we found has the version we expect (like, just compare raw bytes)

mirrord/remote/bootstrap/src/extract.rs
```diff
@@ -0,0 +1,77 @@
...
+    tracing::info!(binary = %target_path.display(), %binary_name, "Extracting embedded binary");
+
+    write_binary_atomically(target_path, bytes)?;
+
+    tracing::debug!(binary = %target_path.display(), %binary_name, "Finished extracting embedded binary");
+    Ok(())
+}
+
+fn write_binary_atomically(target_path: &Path, bytes: &[u8]) -> std::io::Result<()> {
+    let temporary_path = target_path.with_extension(format!("{}.tmp", process::id()));
```
Razz4780: Should be `tempfile`. Given that we accept target path from env, it might not be within temporary directory

mirrord/remote/bootstrap/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+mod error;
+mod extract;
+
+#[ctor]
+fn mirrord_layer_bootstrap_entry_point() {
+    init_tracing();
+    tracing::info!("Starting serverless bootstrap");
+
+    // make sure no other process loads bootstrap
+    unsafe { std::env::remove_var("LD_PRELOAD") };
```
Razz4780, line 17: Wait, won't this library be used for intercepting socket ops as well? 👀 
It should be injected in all process tree. I'd use a custom env to mark "agent already spawned"

mirrord/remote/bootstrap/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+    let mut child = command.spawn()?;
+    tracing::info!(pid = child.id(), "Spawned sidecar agent");
+
+    // Reap the agent asynchronously so bootstrap startup does not block and the host cannot retain
+    // a zombie.
+    std::thread::spawn(move || {
+        if let Err(error) = child.wait() {
+            tracing::warn!(%error, "Failed to reap workload-companion agent");
+        }
+    });
```
Razz4780: 1. Not sure about the deathsig. This process might spawn a child and die, agent should still serve the child.
2. Spawning a thread from an injected library will hurt, we learned the hard way ☠️ 

I think you should:
1. Do the double fork trick to make the agent an isolated process
2. Open a pair of anonymous unix sockets, both without CLOEXEC. Agent will close the first one and use the second one to communicate with the lib. Lib will close the second one and use the first one (sharing it across the whole process tree). Agent will exit when the socket is closed, meaning that all processes with lib have exited

Razz4780 (changes requested):

mirrord/remote/bootstrap/build.rs
itsamegraf: this is how its done in `cli/build.rs`, thought its the convention for dummy val 😜. Ill change to "DUMMY"

mirrord/remote/bootstrap/build.rs
itsamegraf: for clippy and rustdoc builds, like in ci, we dont need the actual agent binary (and later remote_layer as well), so this function accumulates all the cases of "if this env var exists, dont require binary build and use dummy instead"

mirrord/remote/bootstrap/src/lib.rs
itsamegraf: The bootstrap is only responsible for starting the workload-companion and loading `mirrord-remote-layer` (https://github.com/metalbear-co/mirrord/pull/4559) into the initial process. The remote layer is the component propagated to descendants and can assume that the bootstrap already started the agent; an “agent already spawned” guard is therefore not needed in the remote layer. Loading the remote layer without the bootstrap is outside its contract.

mirrord/remote/bootstrap/src/extract.rs
itsamegraf: Done. An existing artifact is reused only when its contents exactly match the embedded bytes; otherwise it is atomically replaced.

mirrord/remote/bootstrap/build.rs
itsamegraf: on second thought, its being consumed by `include_bytes` so we need to give it a file path.. CARGO_MANIFEST_PATH is a good "always-exists" dummy path. let me know if you have other suggestion 🙏

mirrord/remote/bootstrap/src/lib.rs
itsamegraf: Implemented the double-fork approach with `setsid`. The bootstrap synchronously reaps the intermediate child, avoiding a thread in the injected library. The detached workload-companion is intentionally scoped to the container lifetime rather than the lifetime of the process tree, so it remains available even if the process that initially loaded the bootstrap exits.

mirrord/remote/bootstrap/src/extract.rs
itsamegraf: Done. The temporary file is created with NamedTempFile in the destination’s parent directory, then persisted atomically to the requested target path.

itsamegraf:
@greptileai please review

itsamegraf:
@greptileai please rereview

mirrord/remote/bootstrap/build.rs
Razz4780: Oh I only meant "use dummy value" instead of "use bullshit value" :laughing:

mirrord/remote/bootstrap/src/lib.rs
Razz4780: Hmm, I see 🤔 
How do we want to propagate the layer into children processes?

itsamegraf: Very good point - lets discuss this subject in #4688 please 
currently it only has fork support, hooks are supposed to implicitly be propagated, with global locks safety handled by the hook.
yesterday I figured im missing execve support, Ill work on it later today.

mirrord/remote/bootstrap/src/error.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Debug, Error)]
+pub enum RemoteBootstrapError {
+    #[error(transparent)]
+    IO(#[from] std::io::Error),
+    #[error(transparent)]
+    Null(#[from] std::ffi::NulError),
+    #[error("Failed locating bootstrap location: {0}")]
+    DlAddr(String),
+}
```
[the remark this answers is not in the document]
itsamegraf: good catch, the code depending on this missing variant belong to a pr later in the stack, I removed it from this pr.

Razz4780 (dismissed):

itsamegraf:
will merge together with the entire stack
- <kbd>&nbsp;3&nbsp;</kbd> #4688
- <kbd>&nbsp;2&nbsp;</kbd> #4542 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #4541

itsamegraf:
gonna get chopped soon

PR #4548, opened 2026-07-16
Fix error `not found` being shown when mirrord ui stop with no mirrord ui running.

meowjesty:
- Issue: n/a

### Quality Checklist:

  documentation)


mirrord/cli/src/ui.rs
```diff
@@ -446,14 +446,24 @@ fn ui_start_printout(
 }
 
 /// Kills the UI server that is currently running by reading the contents of the file
-/// [`PID_FILE_NAME`]. First checks a server is running by attempting to lock the file
-/// [`UI_LOCK_FILE_NAME`]. Releases the lock after deleting stale files.
+/// [`PID_FILE_NAME`]. A missing PID file means there is no server to stop. Otherwise, checks
+/// whether the server is running by attempting to lock [`UI_LOCK_FILE_NAME`], and releases the
+/// lock after deleting stale files.
```
gememma: > A missing PID file means there is no server to stop

not necessarily true - PID file writing may have failed during start (unlikely but possible) and the server starts anyway. The source of truth for "is there a server running?" is if the `ui.lock` file is locked. Something like this is more accurate (this will fail CI lint if you apply the suggestion directly) -


```suggestion
/// [`PID_FILE_NAME`]. First checks a server is running by attempting to lock the file
/// [`UI_LOCK_FILE_NAME`]. Releases the lock after deleting stale files.
///
/// If there is no PID file but there _is_ a lock file: the server is running but we cannot kill it
/// from this command. In this case, instruct the user to kill it manually.
```

mirrord/cli/src/ui.rs
gememma: I structured this function badly when i wrote it 🤦

It should try to get the guard _first_, so we know if there's a server process running (and stopping new ones from running) before we decide how to handle the PID file not found error - this way around, if there's no server running, the user gets "* No running instance of `mirrord ui` was found. If you think this is incorrect ..." and if there is a server running but the file is missing, they get "* Couldn't kill the server process. Try ..."

```suggestion
    let guard = match TokenClaim::claim_token_file()? {
        TokenClaim::AlreadyRunning => {
            let pid = match std::fs::read_to_string(&pid_file) {
                Ok(pid) => pid,
                Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
                    if with_printouts {
                        println!(
                            "* Couldn't kill the server process. Try killing the process manually, \
                            for example by running `ps aux | grep mirrord` and then `kill $PID` in a \
                            terminal."
                        );
                    }
                    return Ok(());
                }
                Err(error) => return Err(error.into()),
            };

            debug!(
                ?pid,
                ?pid_file,
                "UI server process ID read from file successfully"
            );

```

changelog.d/+small-mirrord-ui-stop-fix.changed.md
```diff
@@ -0,0 +1 @@
+Fix error `not found` being shown when mirrord ui stop with no mirrord ui running.
```
gememma:
```suggestion
Fix error `File not found` being shown when running `mirrord ui stop` with no mirrord ui running.
```

gememma (changes requested):
No idea why I wrote it like this in the first place 🫠

mirrord/cli/src/ui.rs
meowjesty: bbf62f04b5def9021fe94286e74b73f7eb574b98

mirrord/cli/src/ui.rs
meowjesty: bbf62f04b5def9021fe94286e74b73f7eb574b98

meowjesty:
```
mirrord ui stop
* No running instance of `mirrord ui` was found. If you think this is incorrect, try killing the process manually, for example by running `ps aux | grep mirrord` and then `kill $PID` in a terminal.
* Cleaned up stale files
```

It looks like this now.

gememma (approved):
🧑‍🍳 😚 🧑‍🍳 👌

PR #4552, opened 2026-07-16
agent: add env flag to skip conntrack -D during connection flush

aviramha:
## What

Adds `MIRRORD_AGENT_STEALER_FLUSH_CONNECTIONS_CONNTRACK`, an agent environment variable that controls whether the connection flush uses the `conntrack -D` command.

- **Default `true`** — existing behaviour, `conntrack -D` + `ss -K`.
- **`false`** — flush with `ss -K` only, skipping `conntrack -D`.

## Why

When stealing starts, `add_redirect` flushes existing connections so already-established flows reconnect into the redirect. Part of that flush runs:

```
conntrack -D -p tcp --dport <port> --state ESTABLISHED
```

That filter matches on the original-direction destination port, so it also deletes the conntrack entries of connections that were *just* redirected to the agent but not yet accepted. The agent resolves a redirected connection's original destination with `getsockopt(SO_ORIGINAL_DST)`, which reads from conntrack — with the entry gone, the lookup fails with `ENOENT` and the connection is dropped:

```
ERROR mirrord_agent::incoming::iptables: Failed to obtain the original destination
of a redirected TCP connection. Dropping the connection., error: ENOENT
```

On busy stolen ports (e.g. preview envs fronting a service with many health-check / mesh / pooled connections) this shows up as a **burst** of dropped connections each time stealing starts. This flag lets operators opt out of the `conntrack -D` call while keeping `ss -K`, avoiding the burst.

## Notes

- Only an explicit `false` disables `conntrack`; unset or malformed values keep it enabled, so behaviour is unchanged unless deliberately opted out.
- Scoped to the agent's `flush_connections` path; `ss -K` (kernel ≥ 5.1) still handles the intended flush of pre-existing local connections.

## Testing

- `cargo check` / `cargo clippy --deny warnings` on `mirrord-agent-iptables` (`x86_64-unknown-linux-gnu`) and `mirrord-agent-env`.

### DIFF
--- a/mirrord/agent/iptables/src/flush_connections.rs
+++ b/mirrord/agent/iptables/src/flush_connections.rs
@@ -9,14 +9,29 @@
 use std::ops::Not;
 
 use async_trait::async_trait;
+use mirrord_agent_env::envs;
 use tokio::process::Command;
 use tracing::{Level, warn};
 
 use crate::{error::IPTablesResult, redirect::Redirect};
 
+/// Reads whether the flush should use `conntrack -D`.
+///
+/// Defaults to `true`; only an explicit `false` disables it. See
+/// [`STEALER_FLUSH_CONNECTIONS_CONNTRACK`](envs::STEALER_FLUSH_CONNECTIONS_CONNTRACK).
+fn conntrack_enabled() -> bool {
+    envs::STEALER_FLUSH_CONNECTIONS_CONNTRACK
+        .try_from_env()
+        .ok()
+        .flatten()
+        .unwrap_or(true)
+}
+
 #[derive(Debug)]
 pub struct FlushConnections<T> {
     inner: Box<T>,
+    /// Whether to flush existing connections with a `conntrack -D` command in addition to `ss -K`.
+    conntrack: bool,
 }
 
 impl<T> FlushConnections<T>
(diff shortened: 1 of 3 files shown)


mirrord/agent/iptables/src/flush_connections.rs
```diff
@@ -9,14 +9,29 @@
...
+        .ok()
+        .flatten()
+        .unwrap_or(true)
+}
+
 #[derive(Debug)]
 pub struct FlushConnections<T> {
     inner: Box<T>,
+    /// Whether to flush existing connections with a `conntrack -D` command in addition to `ss -K`.
+    conntrack: bool,
```
meowjesty, line 34: Nit: this function is global anyways, so why add it here? Just call `conntrack_enabled()` when you need to check that.

meowjesty (changes requested):
niiiit

mirrord/agent/iptables/src/flush_connections.rs
aviramha: mm it would need to do it each time port is added instead ofo ne, but I guess it's a total nit so just ignorign :D

meowjesty (approved):
👍

PR #4553, opened 2026-07-16
ci+tests: properly run clippy on all targets

iniw:
- **tests: fix `str_to_string` lint**
- **ci: check all targets when running clippy**


meowjesty (approved):
UI not ui

PR #4554, opened 2026-07-16
feat(ui): report unhandled frontend errors when telemetry is enabled

hank-metalbear:
## Summary

The `mirrord ui` frontend initializes analytics (telemetry-gated, opt-out respected) but never captures exceptions, so frontend crashes in the wild are invisible. One-line change: enable exception autocapture in the existing init in the monitor package. That init runs page wide for the merged app, so the shell, session monitor, and wizard are all covered; nothing changes for users with telemetry disabled.

The session-replay masking already in this init (`maskTextSelector: '*'`, `maskAllInputs`) is unrelated to exception events; exception messages come from our own frontend code paths.

## Test plan

- [x] `pnpm --filter session-monitor-frontend run lint` and `run typecheck` green (CI `ui-frontend-lint` job runs these plus format and build)

### DIFF
--- a/changelog.d/+ui-error-tracking.internal.md
+++ b/changelog.d/+ui-error-tracking.internal.md
@@ -1,0 +1,1 @@
+The `mirrord ui` frontend now reports unhandled errors to our product analytics when telemetry is enabled, so crashes in the wild surface in error tracking.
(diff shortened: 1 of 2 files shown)


hank-metalbear:
Closing: the PostHog project-level exception autocapture toggle is already enabled and reaches this SDK via remote config — exception events from the mirrord ui frontend are already flowing (verified in the project data). The in-code flag would be redundant.

PR #4558, opened 2026-07-17
COR-1668 Preview env HTTP filtering does not work with multi-cluster

vladrbg:
### Quality Checklist:

  documentation)


mirrord/cli/src/preview.rs
```diff
@@ -442,6 +454,146 @@ async fn preview_start(
...
+                        "could not read the preview's multicluster status: {error}"
+                    ));
+                    return;
+                }
+                tokio::time::sleep(REPLICA_CLUSTERS_POLL).await;
+            }
+        }
+    };
+
+    let status = view.status.unwrap_or_default();
```
iniw: Does using a default status even make sense? Should it be something like this instead?
```suggestion
    let Some(status) = &view.status else {
        return;
    }
```
Should `PreviewStatus` even implement `Default`? I don't think so.

mirrord/cli/src/preview.rs
```diff
@@ -442,6 +454,146 @@ async fn preview_start(
...
+    loop {
+        let status = match api.get_opt(session_name).await {
+            Ok(Some(view)) => {
+                consecutive_errors = 0;
+                view.status.unwrap_or_default()
+            }
+            Ok(None) => {
+                subtask.warning(
+                    "the preview disappeared while waiting for its replicas - was it stopped, \
+                     or did it fail and get cleaned up?",
```
iniw: I don't think it's very nice to ask the user questions 😆 Just give them suggestions of what may have happened and ways that they can investigate further. For example: https://github.com/metalbear-co/mirrord/blob/30c04f44548522d9ebbf7d348c8dba86eb357553/mirrord/cli/src/error.rs#L636-L642

mirrord/cli/src/preview.rs
```diff
@@ -442,6 +454,146 @@ async fn preview_start(
...
+    }
+    if status.clusters.is_empty() {
+        return;
+    }
+
+    let mut subtask = progress.subtask(&format!(
+        "waiting for the preview on {} cluster(s)",
+        status.clusters.len()
+    ));
+    let deadline = Instant::now() + REPLICA_CLUSTERS_TIMEOUT;
```
iniw: Instead of checking this deadline manually can we wrap this whole thing in a `tokio::timeout` call?

mirrord/cli/src/preview.rs
```diff
@@ -442,6 +454,146 @@ async fn preview_start(
...
+/// the only place the CLI can see all clusters without holding their credentials.
+///
+/// Deliberately best-effort for SLOW clusters: replicas exist for availability, so a cluster
+/// that lags must not fail (or block forever) the whole start - on timeout the lagging
+/// clusters are named and the command proceeds. But not blind: a preview that FAILS while
+/// waiting surfaces immediately with its failure message, a degraded fleet (replicas off,
+/// credential unavailable) is announced instead of waited on, and persistent API errors end
+/// with a warning, never a silent success. Operators without the previews route (genuine 404
+/// on the first fetch) skip the wait silently.
+async fn wait_for_replica_clusters(
```
iniw: Let's move this to a `cli/preview/multicluster.rs` module plz

mirrord/cli/src/preview.rs
```diff
@@ -442,6 +454,146 @@ async fn preview_start(
...
+                .collect::<Vec<_>>()
+                .join(", ");
+            subtask.success(Some(&format!("preview serving on: {names}")));
+            return;
+        }
+
+        if Instant::now() >= deadline {
+            subtask.warning(&format!(
+                "some clusters are not serving the preview yet: {} - they keep converging in \
+                 the background (`mirrord preview status` to re-check)",
```
iniw: What does "they keep converging in the background means? It's worded as if we're complaining about it to the user.

mirrord/cli/src/preview.rs
```diff
@@ -592,6 +744,30 @@ async fn preview_status(
...
+                    let clusters = view_status
+                        .clusters
+                        .iter()
+                        .map(|(cluster, phase)| format!("{cluster}: {}", phase.to_lowercase()))
+                        .collect::<Vec<_>>()
+                        .join(", ");
+                    println!("        clusters: {clusters}");
+                }
+                if let Some(message) = &view_status.message {
+                    println!("        note: {}", message.text);
```
iniw: Isn't this a failure message? "note" makes it seem like it's a normal thing, but it'll only appear on errors, no?

mirrord/operator/src/crd/preview/view.rs
iniw, line 1: This module exists purely for multicluster purposes, right? If so let's move it to `crd/preview/multicluster.rs`. Otheriwse, if the idea behind this is to be a "generic preview session view" then we should move it to `crd/preview/view.rs`.

mirrord/operator/src/crd/preview_view.rs
```diff
@@ -0,0 +1,72 @@
...
+/// (multicluster replica copies are folded in, never listed), and on a multicluster primary the
+/// per-cluster replica phases are aggregated by reading each workload cluster's copy at request
+/// time. Statuses stay in the copies' own status subresources; nothing here is persisted. The
+/// `CustomResource` derive is used only to get the kube `Resource` impl (group/version/kind/
+/// plural) and a `metadata`-carrying wrapper.
+#[derive(CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[kube(
+    group = "operator.metalbear.co",
+    version = "v1",
+    kind = "Preview",
```
iniw: This makes the struct really easy to confuse with `PreviewSession`. I think a better `kind` would be `PreviewSessionMulticlusterView`/`PreviewSessionMulticlusterViewSpec` (or remove the `Multicluster` part if it's meant to be a generic thing).

mirrord/operator/src/crd/preview_view.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub key: String,
+    /// Target workload, e.g. `deployment/my-app`.
+    pub target: String,
+    /// The user's container image running in the preview pods.
+    pub image: String,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PreviewStatus {
```
iniw: Same thing, `PreviewSessionMulticlusterViewStatus` would be better IMO

mirrord/operator/src/crd/preview_view.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Lifecycle phase of the preview, e.g. `Ready` or `Failed`.
+    pub phase: String,
```
iniw: Can we re-use `PreviewSessionPhase`?

mirrord/cli/src/preview.rs
```diff
@@ -592,6 +744,30 @@ async fn preview_status(
...
+                        .get_opt(session_name)
+                        .await
+                && let Some(view_status) = view.status
+            {
+                if !view_status.clusters.is_empty() {
+                    let clusters = view_status
+                        .clusters
+                        .iter()
+                        .map(|(cluster, phase)| format!("{cluster}: {}", phase.to_lowercase()))
+                        .collect::<Vec<_>>()
```
iniw: Can use [`Itertools::join`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join) here to avoid the collect.

mirrord/operator/src/crd/preview_view.rs
```diff
@@ -0,0 +1,72 @@
...
+/// (`GET /apis/operator.metalbear.co/v1/previews`).
+///
+/// This is not a stored Kubernetes object and there is no `CustomResourceDefinition` for it.
+/// The whole `operator.metalbear.co` group is served through the operator's aggregated API, so
+/// every request is answered live from the `PreviewSession` CRs: one entry per logical preview
+/// (multicluster replica copies are folded in, never listed), and on a multicluster primary the
+/// per-cluster replica phases are aggregated by reading each workload cluster's copy at request
+/// time. Statuses stay in the copies' own status subresources; nothing here is persisted. The
+/// `CustomResource` derive is used only to get the kube `Resource` impl (group/version/kind/
+/// plural) and a `metadata`-carrying wrapper.
```
iniw: I think that this goes into some of the mechanical details of why we're using API extensions here instead of proper k8s objects but it doesn't really explain _why_. AFAICT the problem with using k8s objects is needing to synchronize them all across the clusters, is that right?

iniw (changes requested):

mirrord/operator/src/crd/preview_view.rs
vladrbg: Changed to PreviewSessionView and PreviewSessionViewStatus as it's also for single cluster but yea makes more sense like this.

mirrord/cli/src/preview/multicluster.rs
```diff
@@ -0,0 +1,199 @@
...
+                if consecutive_errors >= REPLICA_CLUSTERS_MAX_ERRORS {
+                    progress.warning(&format!(
+                        "could not read the preview's multicluster status: {error}"
+                    ));
+                    return;
+                }
+                tokio::time::sleep(REPLICA_CLUSTERS_POLL).await;
+            }
+        }
+    };
```
iniw: Shouldn't this loop not be part of the timeout too?

iniw (commented):
Minor question but otherwise looks good, will defer approving after the operator review since that might require changes here

mirrord/cli/src/preview/multicluster.rs
vladrbg: oh yep your right

mirrord/cli/src/preview/multicluster.rs
```diff
@@ -135,8 +134,36 @@ pub(super) async fn wait_for_replica_clusters(
...
+/// First fetch of the preview view, tolerating up to [`REPLICA_CLUSTERS_MAX_ERRORS`]
+/// consecutive flakes. `None` means there is nothing to wait on: an operator without the
+/// previews route (genuine 404), or persistent errors - those warn on `progress` here.
+/// The caller bounds this with the wait's shared deadline.
+async fn first_view(
+    api: &Api<PreviewSessionView>,
+    session_name: &str,
+    progress: &mut ProgressTracker,
+) -> Option<PreviewSessionView> {
+    let mut consecutive_errors = 0u32;
```
iniw: Can we restructure these loops to use one of the `tokio_retry` strategies? I think the current code can be replicated with [`FixedInterval`](https://docs.rs/tokio-retry/latest/tokio_retry/strategy/struct.FixedInterval.html), but maybe [`ExponentialBackoff`](https://docs.rs/tokio-retry/latest/tokio_retry/strategy/struct.ExponentialBackoff.html) makes more sense?

iniw (changes requested):
Sorry I just remembered about the existence of `tokio_retry` 😓

mirrord/cli/src/preview/multicluster.rs
vladrbg: yep makes sense

mirrord/operator/src/crd/preview/view.rs
```diff
@@ -0,0 +1,176 @@
+use std::{borrow::Cow, collections::BTreeMap, fmt};
+
+use kube::CustomResource;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use super::PreviewSessionPhase;
+use crate::crd::session::SessionTarget;
+
+/// Read-only view of a preview environment, served by the operator's preview status API
```
Razz4780, line 10: Maybe let's not call it a `View`. We'll want to migrate preview envs to be an API extension, so we might as well extend this resource when the time comes (add POST/PATCH/DELETE endpoints)
This resource should be called `PreviewEnv` I think.

mirrord/config/src/feature/preview.rs
```diff
@@ -526,6 +526,11 @@ pub struct PreviewIdleConfig {
     ///
     /// How long a waking session holds incoming requests while waiting for a preview pod to
     /// become ready, before letting them fail. Defaults to the operator's default (90 seconds).
+    ///
+    /// For previews using database branching on a multicluster fleet, the wake also boots the
+    /// branch-proxy pods before the preview pod - on a node that must pull the operator image
+    /// first, the whole wake can exceed the default and the held requests fail; raise this
+    /// value (or pre-pull the image) for such setups.
```
Razz4780, line 533: This does not seem relevant here

vladrbg:
This is moved to another PR closing this one

PR #4565, opened 2026-07-17
monitor: report frontend errors with stack traces, stop chaos-load alert flood

hank-metalbear:
## Summary

The `mirrord ui` session monitor recorded frontend failures as plain analytics events carrying only a `reason` and an error message string. No stack trace was captured, so these errors could not be grouped or debugged in error tracking. On top of that, a chaos-rule load failure was re-reported on **every** 2.5s poll, so a single ended session produced a continuous stream of duplicate telemetry.

This PR:

- **Captures stack traces.** Real errors are now routed through `posthog.captureException` at every monitor error site — unhandled window errors, unhandled promise rejections, the React crash boundary, and chaos-rule load failures — so they land in error tracking with a stack, alongside the existing `monitor_user_blocked` analytics events (which are unchanged, so usage metrics keep working). `emitUserBlocked` gains an optional raw-error argument for this.
- **Stops the alert flood.** `listChaosRules` now treats a `404` as a benign "session ended" (matching `getSession`) and returns an empty list instead of throwing, and the poll loop reports a load failure **at most once per failure episode** instead of once per poll.

Stack traces are minified until source-map upload is wired into the monitor build (separate follow-up).

## Test plan

- [x] `pnpm run typecheck`, `pnpm run lint`, `pnpm run format:check` all green
- [x] Exercised the real `api` / `emitUserBlocked` / `useChaosRules` modules in a browser against mocked responses:
  - [x] A persistent `404` (session ended) → **0** `monitor_user_blocked` events and **0** captured exceptions across multiple poll cycles (was: one per poll)
  - [x] A persistent non-`404` error → **exactly one** `monitor_user_blocked` event **plus one captured exception carrying a stack trace**, across multiple poll cycles

### DIFF
--- a/packages/monitor/src/components/ErrorBoundary.tsx
+++ b/packages/monitor/src/components/ErrorBoundary.tsx
@@ -22,11 +22,16 @@ export class ErrorBoundary extends Component<Props, State> {
   }
 
   override componentDidCatch(error: Error, info: ErrorInfo): void {
-    emitUserBlocked('ui_crashed', 'user_action', {
-      error: error.message,
-      component: this.props.component,
-      stack: info.componentStack?.slice(0, STACK_TRACE_MAX_LEN),
-    })
+    emitUserBlocked(
+      'ui_crashed',
+      'user_action',
+      {
+        error: error.message,
+        component: this.props.component,
+        stack: info.componentStack?.slice(0, STACK_TRACE_MAX_LEN),
+      },
+      error,
+    )
   }
 
   override render(): ReactNode {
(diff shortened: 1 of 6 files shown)


packages/monitor/src/components/ErrorBoundary.tsx
```diff
@@ -22,11 +22,16 @@ export class ErrorBoundary extends Component<Props, State> {
...
-      component: this.props.component,
-      stack: info.componentStack?.slice(0, STACK_TRACE_MAX_LEN),
-    })
+    emitUserBlocked(
+      'ui_crashed',
+      'user_action',
+      {
+        error: error.message,
+        component: this.props.component,
+        stack: info.componentStack?.slice(0, STACK_TRACE_MAX_LEN),
```
hank-metalbear, line 31: shouldn't this be error.stack?

0x00A5 (approved):

PR #4570, opened 2026-07-20
COR-1685 Terminate loaded process on unrecoverable agent connection loss

### LINKED ISSUE #4563: Session silently enters failover when agent connection is lost; app keeps running "meshed" as an unresponsive zombie
## Summary

When the agent connection drops mid-session (target pod replaced, agent exit, port-forward closed), the internal proxy logs `Proxy encountered a critical error, and is entering the failover state...` **and nothing is surfaced to the user**: `mirrord exec` prints nothing, the wrapped application keeps running, and every mirrord-hooked path in it (including its own listen/accept sockets) is silently broken. The developer only notices because remote traffic stops arriving; `ps` shows a perfectly healthy-looking process tree.

This compounds badly in daily use:

1. Session dies silently → developer's app becomes a zombie that must be killed by hand.
2. Killing the tree produces a confusing Node crash (`Error: read EIO` from the TTY stdin ReadStream) in every Node process in the chain (yarn/tsx watchers etc.), which looks like the app crashed rather than mirrord dying.
3. Zombies keep holding the app's local port. On the **next** mirrord session, the layer's hooked `listen()` **silently swallows** `EADDRINUSE` and binds a random ephemeral port instead — the app prints "listening on 8005" while actually listening on e.g. 57441. The new session's stolen traffic is then forwarded to the dead zombie on the real port (`failed to send the request to the local application's HTTP server: connection closed before message completed` / `Connection refused`), so the new session is *also* dead on arrival, with no error shown to the user.
4. Repeat: each restart creates another zombie and another dead session. Recovering requires manually hunting down every orphaned process.

We hit this constantly with Node.js services whose request handlers fork short-lived child processes (worker pools fork `node --inspect=4321x` children and SIGKILL them within ~300ms; other services spawn `ffmpeg`). Each such child gets the layer injected, opens a layer connection, steal-subscribes its inspector port on the agent, and dies — so a busy session generates constant layer-connection + port-subscription churn (log excerpt below), and the failure correlates strongly with "a new process just started".

## Environment

* mirrord CLI 3.234.0 (Homebrew, macOS), also reproduced on 3.181.0
* agent image `ghcr.io/metalbear-co/mirrord:3.234.0`, ephemeral agent, no operator
* macOS 26.5.1 (arm64), Node.js v20.17.0
* EKS, Kubernetes 1.34, agent picks `iptables-legacy` backend
* incoming steal with HTTP path filter (config below)

## Minimal reproduction

`repro.json`

```json
{
  "target": { "path": { "deployment": "mirrord-repro-target", "container": "app" }, "namespace": "<ns>" },
  "feature": {
    "env": false,
    "fs": { "mode": "read" },
    "network": {
      "dns": true,
      "incoming": {
        "mode": "steal",
        "http_filter": { "path_filter": "^(?!/health)", "ports": [8080] }
      },
      "outgoing": { "tcp": true, "udp": true, "ignore_localhost": true }
    }
  },
  "agent": { "ephemeral": true, "check_out_of_pods": false },
  "internal_proxy": { "log_destination": "/tmp/intproxy-repro.log", "log_level": "mirrord=info" }
}
```

`target-deployment.yaml` — any probed HTTP deployment:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mirrord-repro-target
spec:
  replicas: 1
  selector: { matchLabels: { app: mirrord-repro-target } }
  template:
    metadata: { labels: { app: mirrord-repro-target } }
    spec:
      containers:
        - name: app
          image: registry.k8s.io/e2e-test-images/agnhost:2.45
          args: ["netexec", "--http-port=8080"]
          ports: [{ containerPort: 8080 }]
          readinessProbe: { httpGet: { path: /, port: 8080 }, periodSeconds: 5, failureThreshold: 2 }
          livenessProbe:  { httpGet: { path: /, port: 8080 }, periodSeconds: 5, failureThreshold: 3 }
```

`server.mjs` — a plain HTTP server that forks worker-pool-style children on demand:

```js
import http from 'node:http';
import { fork } from 'node:child_process';
import { fileURLToPath } from 'node:url';
import path from 'node:path';

const childPath = path.join(path.dirname(fileURLToPath(import.meta.url)), 'child.mjs');
let n = 0;
http.createServer((req, res) => {
  if (req.url.startsWith('/fork')) {
    // mimics workerpool's default process worker: fork with an inspector port, SIGKILL shortly after
    const port = 43210 + (n++ % 20);
    const c = fork(childPath, [], { execArgv: [`--inspect=${port}`], silent: true });
    setTimeout(() => c.kill('SIGKILL'), 300);
    res.end(`forked pid=${c.pid}\n`);
    return;
  }
  res.end('hi\n');
}).listen(process.env.PORT || 8080, () => console.log('listening'));
```

```js
// child.mjs
setInterval(() => {}, 1000);
```

**Steps**

1. `kubectl apply -f target-deployment.yaml`
2. `PORT=8080 mirrord exec -f repro.json -- node server.mjs`
3. (optional churn, mirrors real Node worker pools) `for i in $(seq 30); do curl -s localhost:8080/fork; sleep 0.7; done`
4. Kill the agent the way it dies in real life: `kubectl delete pod -l app=mirrord-repro-target`  <br>(organically this happens whenever the deployment rolls, the node drains, or — see "Aggravating factors" — the kubelet kills the pod because its probes were stolen)

**Observed**

* `/tmp/intproxy-repro.log` (timestamps from an actual run):

```
15:56:41.741 WARN  port-forward connection closed without any error published in relevant channel
15:56:41.741 ERROR failed to receive message from the agent, inner task down
15:56:41.741 ERROR agent unexpectedly closed connection
15:56:41.742 ERROR One of the main tasks failed        task_id=AGENT_CONNECTION
15:56:41.742 ERROR Proxy encountered a critical error, and is entering the failover state...
15:56:41.742 ERROR One of the main tasks finished unexpectedly  task_id=INCOMING_PROXY (+ OUTGOING/PING_PONG/SIMPLE/FILES)
```

* **The console shows nothing.** No output from `mirrord exec`, no exit. The `node server.mjs` process stays alive indefinitely but no longer answers even plain local `curl localhost:8080/` (accept path goes through the dead proxy). The intproxy process also stays alive in "failover".
* Killing the process tree afterwards, any Node process in the chain that had a TTY stdin ReadStream dies with:

```
Error: read EIO
    at TTY.onStreamRead (node:internal/stream_base_commons:218:20)
Emitted 'error' event on ReadStream instance ...
{ errno: -5, code: 'EIO', syscall: 'read' }
```

* Starting a **new** `mirrord exec` session while the zombie still holds the port: the app logs `listening on 8080` but `lsof` shows the layer actually bound `*:57441` (random); the zombie still owns `*:8080`; the new session's intproxy forwards stolen requests to `0.0.0.0:8080` → the zombie → `failed to send the request to the local application's HTTP server: ...`. No `EADDRINUSE` is ever raised to the app and no warning is shown.

**Expected**

* Losing the agent connection should be **loud**: kill the wrapped process (or at minimum print a prominent error to the terminal and exit non-zero), the way most people assume `mirrord exec` behaves. An opt-in "keep running unmeshed" mode would be fine, but silent failover as the default makes every downstream symptom look like an application bug.
* `listen()` failing locally with `EADDRINUSE` should be surfaced to the app (or at least warn), not silently rebound to an ephemeral port.

## Aggravating factors seen in the same sessions

These make the silent death occur *organically* (no manual pod delete needed), typically "whenever a new process starts" in the app:

1. **Worker-pool fork churn**: every `/fork` above produces on the agent a full layer lifecycle within ~500ms — `New layer connected` → `TcpSteal(PortSubscribe(All(43210)))` → SIGKILL → `Layer connection closed` → unsubscribe. Real Node services using `workerpool` (default process workers, debug port auto-allocated from 43210) do this on *every job*. The dev server's own `--inspect` port (e.g. 18005) is also steal-subscribed on the target pod. Abruptly killed children show up intproxy-side as `Layer connection failed: io failed: Connection reset by peer`.
2. **Stolen health probes** (ref [#3121](<https://github.com/metalbear-co/mirrord/issues/3121>)): with `path_filter: "^(?!/health)"`, kubelet probes on any other path (`/`, `/ready`, ...) are stolen. If the local app is slow, hung, or a zombie from a previous session, probes fail → kubelet kills the target pod → the ephemeral agent dies with it → silent failover (above) → next session's zombie makes it worse. We can also see it non-destructively: in the repro, every stolen kube-probe request fails intproxy-side with `failed to send the request to the local application's HTTP server: channel closed` at exactly the probe rate (~24/min, indefinitely, starting seconds after session start), preceded by a single agent-side error:

```
ERROR mirrord_agent::incoming::iptables: Failed to obtain the original destination of a redirected
TCP connection. Dropping the connection., error: ENOENT: No such file or directory,
connection_source: <node-ip>:49894        (mirrord/agent/src/incoming/iptables.rs:198)
```

while non-probe traffic to the same stolen port is delivered to the local app just fine. The kubelet meanwhile receives responses (apparently from the original container), so the pod stays Ready — i.e. probe traffic never reaches the local app and nobody is told. Happy to file this one separately if useful.  <br>3. **Dirty leftovers** (ref metalbear-co/mirrord#4115): after a few of these cycles, fresh sessions die within seconds of starting (`agent unexpectedly closed connection` 2–8s after `Starting mirrord-intproxy`) until the target pod is recycled.

## Impact

A team of developers using mirrord for daily "mesh into the dev cluster" workflows loses the mesh several times per hour, each time silently. Because the failure mode is "everything looks alive but no traffic arrives", each occurrence costs a long confused debugging session before someone thinks to blame mirrord, and recovering requires manually killing orphaned process trees (which themselves crash with EIO noise) before restarting.

aviramha:
## Problem

Fixes metalbear-co/mirrord#4563.

When the agent connection drops mid-session on a non-reconnectable session (OSS direct Kubernetes / external proxy — target pod replaced, agent exit, port-forward closed), the internal proxy logs `entering the failover state...` and **nothing is surfaced to the user**: `mirrord exec` prints nothing and does not exit, and the wrapped application keeps running with every mirrord-hooked path in it silently broken.

The failover strategy only delivers the error to a layer on its *next* request. An application idle in `accept()` never makes that call (no traffic arrives once the agent is gone), so it hangs indefinitely as a zombie holding its local ports, while `mirrord exec` never exits.

## Root cause

`mirrord exec` replaces the CLI with the user binary via `execv`, so once a session is running the internal proxy is the only mirrord-controlled process left that observes the agent dropping. The user binary only finds out lazily, on its next hooked syscall.

## Change

On entering failover, the internal proxy now terminates every connected layer process — `SIGTERM` first (so e.g. Node runs its exit handlers), then `SIGKILL` after a short grace for anything that ignored it — so the failure is loud and nothing lingers.

Failover is only reached on terminal, non-recoverable failures: reconnectable operator sessions reconnect inside `AgentConnection` and never get here, so no extra gating is required. Signalling is compiled out under `cfg(test)` (unit tests never signal real pids) and on non-unix targets.

## Verification

Reproduced on a local k3s cluster with an OSS (`operator: false`) steal session, then confirmed the fix with a freshly built CLI. After deleting the target pod, hands-off:

|  | Before | After |
| -- | -- | -- |
| App process | zombie, alive indefinitely | terminated (clean `SIGTERM`, exit 143) |
| `mirrord exec` | never exits | exits |
| Local port | held by zombie | released |
| Diagnostics | silent | log names the process and the reason |

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and all internal-proxy tests pass.


mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +76,74 @@ impl FailoverStrategy {
...
+    async fn terminate_connected_processes(&self) {
+        let pids = self
+            .connected_layers
+            .values()
+            .map(|info| {
+                tracing::error!(
+                    pid = info.pid,
+                    process = info.name,
+                    cause = %self.fail_cause,
+                    "Agent connection was lost and cannot be recovered. Terminating the meshed \
```
aviramha: meshed process isn't a mirrord/MetalBear term.
Use injected/loaded.

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +76,74 @@ impl FailoverStrategy {
...
+    /// hangs forever as a zombie holding its ports. Rather than fail silently, we terminate every
+    /// connected process so the failure is loud and nothing lingers.
+    ///
+    /// See [`Self::signal_processes`] for the per-platform termination.
+    async fn terminate_connected_processes(&self) {
+        let pids = self
+            .connected_layers
+            .values()
+            .map(|info| {
+                tracing::error!(
```
gememma: feels a bit weird to have an error log emitted for every layer instead of a single log at the top of the function which includes all this info 👀 it logs `cause = %self.fail_cause` individually for each layer even though it never changes, as well as the message

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +76,74 @@ impl FailoverStrategy {
...
+            // SAFETY: FFI. An invalid or exited pid yields a null handle, which we skip; every
+            // opened handle is closed. `TerminateProcess` on a valid handle cannot fail us into UB.
+            unsafe {
+                let handle = OpenProcess(PROCESS_TERMINATE, FALSE, pid as u32);
+                if handle.is_null() {
+                    continue;
+                }
+                TerminateProcess(handle, 1);
+                CloseHandle(handle);
+            }
```
gememma: can we avoid the unsafe with something like this?
```rust
#[cfg(windows)]
std::process::Command::new("taskkill")
    .args(["/pid", &pid, "/t"])
    .output()?;
```

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -30,8 +45,24 @@ pub(super) struct FailoverStrategy {
...
+    connected_layers: HashMap<LayerId, ProcessInfo>,
 }
 
+/// Grace period between the `SIGTERM` and the `SIGKILL` we send to the meshed processes on a
+/// terminal failure, giving them a chance to run their own shutdown before we force the issue.
+/// Shortened under test so the real termination path can be exercised without a slow wait.
+#[cfg(all(unix, not(test)))]
+const TERMINATION_GRACE: Duration = Duration::from_secs(2);
+#[cfg(all(unix, test))]
+const TERMINATION_GRACE: Duration = Duration::from_millis(50);
```
gememma: do we need this? I don't think we normally have test gated code (outside of test utils) right?

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +76,74 @@ impl FailoverStrategy {
...
+    /// [`TERMINATION_GRACE`], so well-behaved processes get to run their shutdown first.
+    #[cfg(unix)]
+    async fn signal_processes(pids: Vec<i32>) {
+        if pids.is_empty() {
+            return;
+        }
+
+        for pid in &pids {
+            // An invalid or already-reaped pid just yields `ESRCH`, which we ignore.
+            let _ = kill(Pid::from_raw(*pid), Signal::SIGTERM);
```
gememma: this is ignoring `ESRCH` but also ignoring legit errors, do we want to at least log them to the user? otherwise it's just another silent failure

mirrord/intproxy/src/failover_strategy.rs
aviramha: I'm against invoking external processes, on Windows it always complicates AVs

mirrord/intproxy/src/failover_strategy.rs
aviramha: It's just the termination grace so test will be faster.

gememma: 🤷 its only 2 seconds on one test, feels unnecessary to me

aviramha: It compounds. There is limit on concurrent tests

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -30,8 +45,24 @@ pub(super) struct FailoverStrategy {
     pending_layers: Vec<(LayerId, MessageId)>,
     any_connection_accepted: bool,
     fail_cause: ProxyRuntimeError,
+    /// Processes of the layers that were connected when the proxy failed.
```
aviramha: This should be two sentences.

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +77,95 @@ impl FailoverStrategy {
...
+            return;
+        }
+
+        let processes = self
+            .connected_layers
+            .values()
+            .map(|info| (info.pid, info.name.as_str()))
+            .collect::<Vec<_>>();
+
+        tracing::error!(
```
aviramha: change to warn

gememma (dismissed):
LGTM, ping if you need re-approval 🫑

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +77,95 @@ impl FailoverStrategy {
...
+        for pid in pids {
+            // SAFETY: FFI. An invalid or exited pid yields a null handle, which we skip; every
+            // opened handle is closed. `TerminateProcess` on a valid handle cannot fail us into UB.
+            unsafe {
+                let handle = OpenProcess(PROCESS_TERMINATE, FALSE, pid as u32);
+                if handle.is_null() {
+                    continue;
+                }
+                if TerminateProcess(handle, 1) == 0 {
+                    tracing::warn!(
```
aviramha, line 170: how about providing getlasterror or something so we'll know why?!

mirrord/intproxy/src/failover_strategy.rs
```diff
@@ -45,6 +77,95 @@ impl FailoverStrategy {
...
+    /// arbitrary process here, so this matches the unix `SIGKILL` with no grace phase.
+    #[cfg(windows)]
+    async fn signal_processes(pids: Vec<i32>) {
+        for pid in pids {
+            // SAFETY: FFI. An invalid or exited pid yields a null handle, which we skip; every
+            // opened handle is closed. `TerminateProcess` on a valid handle cannot fail us into UB.
+            unsafe {
+                let handle = OpenProcess(PROCESS_TERMINATE, FALSE, pid as u32);
+                if handle.is_null() {
+                    continue;
```
aviramha, line 167: also warn here, probably this is like ESRCH (no process) but should verify this too.

gememma (approved):

PR #4571, opened 2026-07-20
Retry flaky Linux Homebrew install in release monitor

aviramha:
## What

The **Release Monitor** workflow (`releases-test.yaml`) fails intermittently — ~47% of runs (7 of the last 15) — always on the same leg: `test_install_paths (ubuntu-24.04)` → the Linux `brew install` step. It dies right after `==> Installing mirrord from metalbear-co/mirrord` with `##[error]Broken pipe`, then fires a false `install-path check FAILED` alert to the prod alerts channel.

## Root cause

It's a flaky Linuxbrew failure, **not** a real release regression. During its Linux install, Homebrew runs `ldd` on the binary (linkage audit) and pipes it into a reader that closes early, so `ldd`/`printf` get SIGPIPE. It's a race:

- Usually the broken pipe lands harmlessly on `ldd`'s stderr and the install succeeds — visible even in *successful* runs:
  ```
  mirrord 3.234.0
  /usr/bin/ldd: line 41: printf: write error: Broken pipe
  brew install succeeded. Version: mirrord 3.234.0
  ```
- ~half the time the EPIPE bubbles up and aborts `brew install` with a non-zero exit → the step reports failure.

Everything else is healthy across the failing runs: the curl|bash installer works, macOS `brew install` works, all required assets are present, and the version endpoint is fine. The formula itself just downloads the zip and does `bin.install "mirrord"` — the broken pipe is inside Homebrew's own Linux install machinery. The earlier `HOMEBREW_NO_SANDBOX_LINUX=1` change (#4364) reduced but didn't eliminate this.

## Fix

Retry the Linux `brew install` up to 3 times, resetting any partial install between attempts, before declaring failure. A transient EPIPE almost never recurs 3 times; a genuinely broken formula or missing asset still fails every attempt, so real regressions are still caught.


meowjesty (dismissed):

meowjesty (approved):

gememma (commented):
hmm, i wonder if the runner is using a very old version of brew 👀 it _seems_ like someone fixed this issue [upstream in january](https://github.com/homebrew/brew/issues/21365), maybe you can set a minimum brew version?

aviramha:
> hmm, i wonder if the runner is using a very old version of brew 👀 it _seems_ like someone fixed this issue [upstream in january](https://github.com/homebrew/brew/issues/21365), maybe you can set a minimum brew version?

we're using latest stable..

PR #4575, opened 2026-07-21
INT-527 Automatically enable copy_target for Job/CronJob targets

t4lz:
## What

`Job` and `CronJob` targets no longer require the user to enable the `copy_target` feature manually. mirrord now enables copy target automatically for them, instead of failing config verification.

Closes INT-527.

## Why

Job and CronJob targets have no long-running pod for the agent to attach to, so they can only be used through a copied pod. Previously, targeting one without `"copy_target": true` in the config failed verification with:

```
A Job or CronJob target has been specified, but the feature `copy_target` has not been enabled!
```

This was a papercut — the requirement is an implementation detail the user shouldn't have to know. It came up when setting up database branching against a CronJob, where the target was rejected before branching could run.

We already auto-enable copy target for other cases where it is needed but not explicitly requested (empty deployments, queue splitting). This extends the same treatment to Job/CronJob targets, and reuses the existing message so the user is told it happened:

> The copy target feature is used for a Job or CronJob target (even though `copy_target` was not explicitly set).

## Changes

- **`mirrord-config`**: removed the `TargetJobWithoutCopyTarget` error and the `verify()` check that raised it. The operator-disabled case is still caught by the more accurate `TargetRequiresOperator` error (Job/CronJob already `requires_operator`). `TargetType::compatible_with` now lists Job/CronJob regardless of `copy_target`, so `mirrord ls` and the IDEs offer them. Doc comments updated.
- **`mirrord-operator`**: `should_copy_target` and its multi-cluster variant now return `true` for Job/CronJob targets. Because the copy-target branch is taken, `ResolvedTarget::assert_valid_mirrord_target` (which still rejects these targets) is bypassed — it remains the backstop for the non-copy path.

The OSS/no-operator rejection in `mirrord-kube` (`resolved.rs`) is intentionally left in place.


mirrord/config/src/lib.rs
```diff
@@ -2263,6 +2252,41 @@ mod tests {
         );
     }
 
+    /// A `Job` target no longer requires the user to enable `copy_target` explicitly: it is
+    /// enabled automatically later in the session flow, so verification must pass without it.
+    #[test]
+    fn job_target_without_copy_target_passes_verification() {
```
aviramha: I feel like this is redundant.. no real logic is being tested.

aviramha (dismissed):

mirrord/config/src/lib.rs
t4lz: [removed](https://github.com/metalbear-co/mirrord/commit/345e339b4edc179d74778e61d71bbdb7b824d0ba)

gememma (commented):
i think youll need to update the schema since you changed a config comment 👀

gememma (approved):
⛵

PR #4576, opened 2026-07-21
feat(monitor): attach mirrord version to session monitor telemetry

hank-metalbear:
## Summary

Monitor telemetry carries no indication of which mirrord release produced an event. Because the UI ships inside the CLI and users upgrade on their own schedule, a crash that was fixed and released weeks ago keeps firing alerts from older clients, and there is no way to tell that apart from a live regression on the current release. Triage stalls there: you cannot answer "is this still broken?" without the version.

Today the only way to answer it at all is an accident. React's `componentStack` happens to embed the bundle's asset URL, so a crash event carries a hashed bundle name that can be used as a rough release fingerprint. That worked once, to retire a long-running crash signature by showing that every occurrence since the fix came from bundles that predate it. It does not generalize: the fingerprint only exists on crash events, it is a build hash rather than a version, and nothing maps it back to a release. Every non-crash blocked reason, including the most common one, has no attribution at all.

The session API already reports `mirrord_version` (from `CARGO_PKG_VERSION` in the intproxy session info), and the UI already parses it into `SessionInfo`. It was simply never sent anywhere.

This registers it as a PostHog super property, so every monitor event, including `monitor_user_blocked` and the captured exception, is attributable to a release. It is registered before the first capture, so `session_monitor_opened` carries it too and the blocked-vs-opened ratio can be broken down by version.

No new data is collected from the user's machine: this is the version string the CLI already reports, and it rides the existing telemetry opt-in.

## Test plan

Gates, all pass:

- `pnpm --filter session-monitor-frontend typecheck`, `lint`, `format:check`
- `pnpm --filter mirrord-ui typecheck`, `lint`, `format:check`, `build`

Wire verification, observed rather than reasoned about. Served the built `packages/ui/dist` bundle against a stub session API returning a session with a known `mirrord_version`, drove it with a headless browser, and intercepted the outgoing telemetry requests to decode the actual event payloads:

| build | events captured | carrying `mirrord_version` |
|---|---|---|
| this branch | 8 | 8 |
| `main`, same driver and stub | 8 | 0 |

The 8 are the same events in both runs (`session_monitor_opened`, `monitor_user_succeeded`, `monitor_user_blocked`, `$exception`, and opt-in events). On this branch every one of them carries the version from the stubbed session API; on `main` none do. `monitor_user_blocked` and `$exception` are the two that matter for triage, and both were confirmed on the wire, not inferred from posthog-js source.

Not exercised: a real operator install. The session payload was stubbed, so the version string came from the stub rather than from a live intproxy.


packages/monitor/src/analytics.ts
```diff
@@ -36,9 +39,25 @@ export function initAnalytics(telemetryEnabled: boolean) {
...
+let registeredVersion: string | null = null
+
+/**
+ * Register the mirrord CLI version serving this UI as a super property, so every captured
+ * event carries it. Users run whatever CLI version they installed and upgrade on their own
+ * schedule, so without this a crash fixed and released weeks ago is indistinguishable from a
+ * live regression: the alerts keep firing from old clients and there is no way to tell which
+ * releases are still affected.
+ */
+export function setMirrordVersion(version: string | undefined) {
```
wg4197763-eng: Will I ever be free

PR #4585, opened 2026-07-21
COR-1690 Fix credential store save

0x00A5:
In [`CredentialStoreSync::get_client_certificate`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/auth/src/credential_store.rs#L299-L316), we lock the `~/.mirorrd/credentials` file, call [`CredentialStoreSync::access_credential`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/auth/src/credential_store.rs#L235), and then unlock the file.

**Every call** to `CredentialStoreSync::access_credential` will load, wipe out, and then completely rewrite the `~/.mirrord/credentials` content.

The [`‎CredentialStore::save‎`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/auth/src/credential_store.rs#L84), however, calls `write_all` only. Note that `write_all` returns immediately after all bytes are **queued** for writing. This opens a window where bytes are partially written, but the file is already unlocked, allowing another process starting a mirrord session in parallel to view a malformed `~/.mirrord/credentials` file, thus, triggering a re-generation of the client key pair.

This issue can cause unexpected seat count creep for users who runs many mirrord session in parallel, such as using a coding agent to run test suite.

The fix is just adding a flush after write_all, so we make sure all bytes land in the file before unlocking it.

### Quality Checklist:

  documentation)

### DIFF
--- a/changelog.d/+cred-wipe.fixed.md
+++ b/changelog.d/+cred-wipe.fixed.md
@@ -1,0 +1,2 @@
+Fixed a bug in `mirrord-auth` causing seat counting client key pair being re-generated when running
+a burst of mirrord sessions concurrently.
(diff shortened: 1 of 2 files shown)


meowjesty (approved):
👍

PR #4586, opened 2026-07-21
Support targeting labels.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1671/support-one-local-session-with-multi-target-in-a-single-cluster
- Related PRs;
  - https://github.com/metalbear-co/operator/pull/2043 

With this PR now you can target labels like this:

```toml
labels = { issue = "cor-1671", "cor.metalbear.co/target-group" = "meow" }
```

It'll target all the pods that have these labels in a namespace, with some caveats: most operations are limited to 1 targeted pod (file ops, and some others), meaning that whichever pod we selected as the 1st target, will be the one where these operations happen in, but every pod that matches the label will have it's traffic stolen (this was the feature request).

tl;dr: file ops and some other stuff works only in the pod we select, the other pods only get incoming traffic ops.

Some features are not implemented here: copy target, queue splitting, db branching, and multi-cluster.

### Quality Checklist:

  documentation)


mirrord/operator/src/crd/session.rs
```diff
@@ -115,12 +117,34 @@ impl SessionTarget {
                 name: t.replica_set,
                 container: t.container?,
             }),
+            Target::Label(t) => Some(Self {
+                api_version: LABEL_TARGET_API_VERSION.to_owned(),
+                kind: LABEL_TARGET_KIND.to_owned(),
+                name: t.selector(),
+                container: t.container?,
+            }),
```
Razz4780: Not a fan of this hack :V

Can we do:

```rust
    use std::cmp::Ordering;

    use kube::core::{Expression, Selector};
    use schemars::JsonSchema;
    use serde::{Deserialize, Serialize};

    /// Target of a mirrord session.
    #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
    #[serde(untagged, rename_all = "camelCase")]
    pub enum SessionTarget {
        KubeResource(KubeResourceTarget),
        PodSet(PodSetTarget),
    }

    /// Target of a mirrord session that is a single Kubernetes resource.
    #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
    pub struct KubeResourceTarget {
        /// Kubernetes resource apiVersion.
        pub api_version: String,
        /// Kubernetes resource kind.
        pub kind: String,
        /// Kubernetes resource name.
        pub name: String,
        /// Name of the container defined in the Pod spec.
        pub container: String,
    }

    /// Target of a mirrord session that is a set of pods.
    #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
    pub struct PodSetTarget {
        /// Optional user-provided name of this set.
        ///
        /// Should **only** be used for display purposes.
        pub name: Option<String>,
        /// Label selector for the pods.
        ///
        /// Expressions in this selector must always be sorted according to
        /// [`Self::cmp_expressions`]. This is always true for freshly deserialized values.
        /// If the expressions are not sorted, derived [`PartialEq`] and [`Eq`] on this struct
        /// might yield false negatives.
        #[schemars(schema_with = "label_selector_serde::schema")]
        #[serde(with = "label_selector_serde")]
        pub label_selector: Selector,
    }

    impl PodSetTarget {
        /// Alternatively, derive [`PartialOrd`] and [`Ord`] on [`Expression`] in our fork.
        /// Or just change [`Selector`] to keep a `BTreeSet` of expressions instead of `Vec`.
        pub fn cmp_expressions(lhs: &Expression, rhs: &Expression) -> Ordering {
            todo!()
        }
    }

    mod label_selector_serde {
        use k8s_openapi::apimachinery::pkg::apis::meta::v1::LabelSelector;
        use kube::core::Selector;
        use schemars::{JsonSchema, Schema, SchemaGenerator};
        use serde::{Deserialize, Deserializer, Serialize, Serializer, de::Error};

        use crate::crd::session::new::PodSetTarget;

        pub fn schema(generator: &mut SchemaGenerator) -> Schema {
            crate::crd::label_selector::LabelSelector::json_schema(generator)
        }

        pub fn deserialize<'de, D: Deserializer<'de>>(
            deserializer: D,
        ) -> Result<Selector, D::Error> {
            let selector = LabelSelector::deserialize(deserializer)?;
            let mut expressions = Selector::try_from(selector)
                .map_err(D::Error::custom)?
                .into_iter();
            expressions
                .as_mut_slice()
                .sort_unstable_by(PodSetTarget::cmp_expressions);
            Ok(expressions.collect())
        }

        pub fn serialize<S: Serializer>(
            selector: &Selector,
            serializer: S,
        ) -> Result<S::Ok, S::Error> {
            LabelSelector::from(selector.clone()).serialize(serializer)
        }
    }

```

This would also be safer in terms of updating the operator code - will trigger compilation errors across the codebase.

meowjesty:
@greptile-apps @greptile review this

mirrord/cli/src/error.rs
```diff
@@ -771,6 +775,7 @@ impl From<OperatorApiError> for CliError {
                 Self::friendlier_error_or_else(e, Self::CreateKubeApiFailed)
             }
             OperatorApiError::ConnectRequestBuildError(e) => Self::ConnectRequestBuildError(e),
+            OperatorApiError::InvalidBaggageHeader(error) => Self::InvalidBaggageHeader(error),
```
0x00A5, line 782: What is this new error variant for?

mirrord/config/src/target/label.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    fn container(&self) -> Option<&String> {
+        self.container.as_ref()
+    }
+}
+
+impl fmt::Display for LabelTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "label/{}", self.selector())?;
```
0x00A5, line 101: The delimiter collide with the values here.
Label key or value could contain `/`. For example, `"app.kubernetes.io/name": "app"`. It serializes to `label/app.kubernetes.io/name=api/container/app`. If we rely on this `to_string()` and some `from_str()` implementation to do serde roundtrip, it will break.

mirrord/config/src/target.rs
0x00A5, line 368: The `Label` variant is currently silently failing? Mentioned in another comment, but there are places relying on the `to_string()` , `from_str()` roundtrip to work and produce the same value.

mirrord/kube/src/resolved.rs
```diff
@@ -51,12 +51,39 @@ pub enum ResolvedTarget<const CHECKED: bool> {
...
+#[derive(Debug, Clone)]
+pub struct ResolvedLabelTarget {
+    pub labels: BTreeMap<String, String>,
+    /// We're selecting one [`Pod`] to be the "representative" pod for all the pods that match the
+    /// label selector.
+    ///
+    /// Translation: we select 1 pod out of a list (whichever 1st one we find that is ready and
+    /// good to go), and we use its env vars, fileops, dns and outgoing network. Traffic stealing
+    /// happens for every pod we find that matches the label selector, but these other operations
+    /// are performed on the representative pod.
```
0x00A5: nitpick: just use the post-translation part alone is fine :D

mirrord/kube/src/resolved.rs
```diff
@@ -51,12 +51,39 @@ pub enum ResolvedTarget<const CHECKED: bool> {
...
+    pub representative: ResolvedResource<Pod>,
+}
+
+impl ResolvedLabelTarget {
+    pub fn name_any(&self) -> String {
```
0x00A5, line 78: Double check where this is used in operator. We maybe use the returned string to construct resource names, and the value this returns may not work.

mirrord/operator/src/client/database_branches.rs
```diff
@@ -1363,8 +1363,14 @@ impl UnifiedDatabaseBranchParams {
...
         }
         let target_display = target_with_container.to_string();
-        let session_target = SessionTarget::from_config(target_with_container)
-            .ok_or_else(|| OperatorApiError::TargetResolutionFailed(target_display))?;
+        let session_target = match SessionTarget::from_config(target_with_container)
+            .ok_or_else(|| OperatorApiError::TargetResolutionFailed(target_display.clone()))?
+        {
+            SessionTarget::KubeResource(target) => target,
+            SessionTarget::PodSet(_) => {
+                return Err(OperatorApiError::TargetResolutionFailed(target_display));
```
0x00A5: The error here maybe misleading, more like unsupported target config?

mirrord/operator/src/client/error.rs
```diff
@@ -49,6 +49,9 @@ pub enum OperatorApiError {
     #[error("failed to build a websocket connect request: {0}")]
     ConnectRequestBuildError(HttpError),
 
+    #[error("configured baggage is not a valid HTTP header value: {0}")]
+    InvalidBaggageHeader(#[from] http::header::InvalidHeaderValue),
```
0x00A5, line 53: Same question for this error variant, what is this for?

mirrord/operator/src/crd/session.rs
```diff
@@ -38,20 +40,77 @@ impl fmt::Display for SessionOwner {
...
+///
+/// Its optional name is only used for display; selection and equality are based on the label
+/// selector and container.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PodSetTarget {
+    /// Target name.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[schemars(with = "String")]
+    pub name: Option<String>,
```
0x00A5: What's this field for? It seems that it's only initiated as `None` (except in unit tests).

mirrord/operator/src/client.rs
0x00A5, line 1: what are the baggage header related changes in this file for?

mirrord/operator/src/crd/session.rs
```diff
@@ -38,20 +40,77 @@ impl fmt::Display for SessionOwner {
...
+
+    /// Label selector for the targeted pods.
+    #[schemars(schema_with = "label_selector_serde::schema")]
+    #[serde(with = "label_selector_serde")]
+    label_selector: Selector,
```
0x00A5, line 71: Why don't we use the type `LabelSelector` from mirrord/operator/src/crd/label_selector.rs directly here?

meowjesty: Either michal forgot this exists, or he wants us to use kube's `Selector` here instead. @Razz4780

Razz4780: Cuz Selector is nice : D
Just one ToString away from using it in GetParams::label_selector

mirrord/operator/src/client.rs
meowjesty: It's so we can test some stuff in staging. I remember not being able to test this label targeting without adding that, like some of the initial requests we made reached the original staging operator, instead of the one I was running.

- ai:

> Previously, baggage was attached only to the final WebSocket connection request. Earlier requests made through the base Kubernetes client—operator discovery, feature checks, and session setup—did not carry it. Consequently, those requests reached the deployed staging operator instead of the local operator running under mirrord, producing the misleading “label targeting is not supported” error.

mirrord/operator/src/crd/session.rs
meowjesty: Useless right now. It was part of the suggested changes, and I remember ai telling me it could be used in the future if we fill something something.

Tbh I would rather just remove it, we can right?

mirrord/operator/src/client/database_branches.rs
meowjesty: 5aea994b03a12858b015adf7cdcb7080846bf780

mirrord/operator/src/client/error.rs
meowjesty: It's for the baggage change, see the other comment.

mirrord/kube/src/resolved.rs
meowjesty: 3cdba32afc0633ba87c4e090083726b8b14957f8

mirrord/config/src/target.rs
meowjesty: 8c64dc943ec72825de310ecedb49f1c27a1df7ec

I had it implemented just in the config first to test, and forgot to port it to be supported as a real target everywhere.

mirrord/config/src/target/label.rs
meowjesty: 8c64dc943ec72825de310ecedb49f1c27a1df7ec

Should've been addressed by this one too.

mirrord/cli/src/error.rs
meowjesty: Addressed in the other error comment.

mirrord/kube/src/resolved.rs
meowjesty: [8134d2c](https://github.com/metalbear-co/mirrord/pull/4586/commits/8134d2c12443cfbc5b540d83c68ddea82489963c)

meowjesty:
@greptile-apps hey friend, can you pls review it again?

mirrord/operator/src/crd/session.rs
0x00A5: it's only initiated as `None` so I think we remove. Can add it in the future when it's needed.

meowjesty: [49379e4](https://github.com/metalbear-co/mirrord/pull/4586/commits/49379e478e23562e1a3448f06b7191f7ebe0aa97)

0x00A5 (approved):

PR #4596, opened 2026-07-23
Use mirrordclient from mirrord protocol api in mirrord cli

MintSoup:
### Quality Checklist:

  documentation)


mirrord/protocol-api/Cargo.toml
```diff
@@ -15,6 +15,7 @@ categories.workspace = true
 
 [dependencies]
 mirrord-protocol = { path = "../protocol" }
+mirrord-progress = { path = "../progress", features = ["implementations"] }
```
Razz4780: Do we need `implementations` feature? 👀 
You could accept `impl Progress` instead of `ProgressTracker`

mirrord/protocol-api/src/client/retrying.rs
```diff
@@ -183,26 +190,53 @@ impl MirrordClientRetry for MirrordClient {
...
 async fn retry_op<F, Fut, R>(mut make_fut: F, timeout: Duration) -> ClientResult<R>
 where
     F: FnMut() -> Fut,
     Fut: Future<Output = ClientResult<R>>,
 {
     let mut timeout_fut = std::pin::pin!(tokio::time::sleep(timeout));
+    let mut retryable_error_backoff = ExponentialBackoff::from_millis(2)
+        .factor(125)
+        .max_delay(Duration::from_secs(2))
+        .take(MAX_RETRYABLE_ERROR_ATTEMPTS);
```
Razz4780, line 208: I always like to add jitter. I doesn't cost anything, and you don't have to wonder if thundering herd will be problem or not

mirrord/protocol-api/src/client/error.rs
```diff
@@ -88,10 +88,10 @@ impl TaskError {
 
     pub(super) fn can_reconnect(&self) -> bool {
         match self {
-            Self::Io(..)
-            | Self::ServerClosed(..)
-            | Self::MissedPing
-            | Self::CommunicationTimeout { .. } => true,
+            // Can't reconnect when there's an error explicitly
+            // provided (e.g. dirty iptables)
+            Self::ServerClosed(error) => error.is_none(),
```
Razz4780: Afaik we also provide error if the operator is shutting down. The can/cannot reconnect state should be privy to the `ProtocolConnector` implementation.
In the operator scenario, if the session is not reconnectable, the operator responds with a special HTTP code and reason. Until this happens, we must assume that reconnect is possible

mirrord/kube/Cargo.toml
```diff
@@ -36,7 +36,7 @@ k8s-openapi.workspace = true
 kube.workspace = true
 rand.workspace = true
 regex.workspace = true
-serde.workspace = true
+serde = { workspace = true, features = ["rc"]}
```
Razz4780: Why? 👀

mirrord/cli/src/connector.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl AgentConnector {
+    pub async fn into_client(
+        self,
+        progress: &mut ProgressTracker,
+    ) -> Result<MirrordClient, ClientError> {
+        MirrordClient::new(
+            self,
+            ClientConfig::cli(),
+            NonZeroUsize::new(1024).expect("channel size is nonzero"),
```
Razz4780: More like 16 I think 😆 
We don't need to buffer 1024 concurrent requests

mirrord/cli/src/connector.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+/// Connects to an operator session that was prepared during setup.
+///
+/// The first connection is established while the session is set up and parked in
+/// [`Self::first_conn`], to be handed out on the first [`connect`](AgentConnector::connect) call.
+/// Reconnects go through [`OperatorApi::connect_to_session`], reusing [`Self::session`].
+#[derive(Debug)]
+pub(crate) struct OperatorConnector {
```
Razz4780, line 70: This connector has to enter a permanent "failed" state when it receives session-non-reconnectable error from `OperatorApi` (see how we handle this in intproxy)

mirrord/cli/src/main.rs
```diff
@@ -963,36 +962,29 @@ async fn port_forward(
...
-                kube_api_error,
-                CliError::PortForwardingSetupError,
-            ),
-            AgentConnectionError::Tls(connection_tls_error) => connection_tls_error.into(),
-            AgentConnectionError::ProtocolError(protocol_error) => protocol_error.into(),
-        })?;
-
-    let connection_2 = agent_conn.connection;
+    let connection_2 =
+        Connection::from_channel(connector.connect(&mut progress).await.map_err(friendly)?);
```
Razz4780: Did you omit port-forward on purpose? This PR is already quite big, so we could as well take care of it separately.

Razz4780 (changes requested):

mirrord/protocol-api/Cargo.toml
MintSoup: Making it generic over a `Progress` impl would have to make a nontrivial volume of code also generic (think all the functions in the call graph that pass it around). This also includes the trait methods on `MirrordClient`. The enum dispatch overhead is negligible and it saves the monomorphization costs (compile time and binary size), so IMO it's a net win.

Another thing we could do is use `dyn Progress` everywhere but `ProgressTracker` achieves (mostly) the same thing and I don't see a reason not to use it.

mirrord/kube/Cargo.toml
MintSoup: ah I think it's a leftover from some experimentation, will remove

mirrord/cli/src/main.rs
MintSoup: Portforward is used in the intproxy process as well, for the DB branch auto portforwarding feature. It shares the agent connection with the main intproxy logic and that uses the legacy style API, so we can't refactor it atm. That being said [we are looking towards moving PF into its own separate process](https://metalbear.slack.com/archives/C0959M0306L/p1774450102745689) so that might help.

mirrord/cli/src/connector.rs
MintSoup: copied from the tests and forgot to change 💀

mirrord/cli/src/connector.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// Sets [`Self::failed`] when the error implies we can no longer
+    /// reconnect.
+    fn handle_error(&mut self, error: &OperatorApiError) {
+        if let OperatorApiError::KubeError {
+            error: kube::Error::Api(error),
+            operation: OperatorOperation::WebsocketConnection,
+        } = error
+        {
+            self.failed |= error.code == RECONNECT_NOT_POSSIBLE_CODE;
+            self.failed |= error.reason == RECONNECT_NOT_POSSIBLE_REASON;
```
Razz4780:
```suggestion
            self.failed |= error.code == RECONNECT_NOT_POSSIBLE_CODE && error.reason == RECONNECT_NOT_POSSIBLE_REASON;
```

Code is just generic 410 Gone, we need to check the reason as well

mirrord/cli/src/connector.rs
```diff
@@ -0,0 +1,307 @@
...
+            }
+        }
+    }
+
+    fn can_reconnect(&self) -> bool {
+        match self {
+            AgentConnector::Operator(operator) => operator.can_reconnect(),
+            AgentConnector::Direct(_) => false,
+        }
+    }
```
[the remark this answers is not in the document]
Razz4780: This is intended behavior. Could be documented here though - we only support reconnect when the operator is used

mirrord/protocol-api/Cargo.toml
Razz4780: I think you no longer use the progress at all, after disabling reconnects in OSS

Razz4780 (changes requested):

Razz4780 (approved):

PR #4605, opened 2026-07-26
INT-483 Add query by routing key to the active sessions api and cli

DmitryDodzin:
This adds the `--key` arg to `ls` subcommand under `mirrord session` command, it filters out the sessions that don't have a specific key that was provided

### Quality Checklist:

  documentation)


mirrord/cli/src/config.rs
```diff
@@ -1678,13 +1678,32 @@ pub struct SessionCommonArgs {
 pub enum LocalSessionCommand {
     /// List mirrord sessions currently running locally and in cluster (in same namespace).
     #[command(visible_alias = "ls")]
-    List,
+    List(SessionListArgs),
```
meowjesty, line 1853: This enum doesn't support `#[default]`?

mirrord/operator/src/crd.rs
```diff
@@ -528,6 +531,179 @@ pub struct SessionSpec {
...
+/// enough:
+///
+/// - A field path is dot-separated, with `\.` escaping a literal dot in a key and `\\` a literal
+///   backslash. A JSON Pointer is `/`-separated with no backslash escaping.
+/// - Within a JSON Pointer token, `~` must be encoded as `~0` and `/` as `~1` (escaping `~` first).
+///   This matters for keys that contain slashes, e.g. Kubernetes annotations/labels like
+///   `operator.metalbear.co/session-id`.
+///
+/// [RFC 6901]: https://www.rfc-editor.org/rfc/rfc6901
+fn field_path_to_json_pointer(path: &str) -> String {
```
meowjesty: What the heck, our keys can be some crazy shit like this, or is this a thing that kubernetes does to our key when it's like, too big or has special characters?

meowjesty (commented):
Question about why `SessionCrd` needs so much stuff, rather than just doing some substring matching like `label.contains(&key)` for each label (I don't remember where our key ends up, is it in a label for the session or something else? Anyways, why do we need all this json and string-fu).

DmitryDodzin:
> Question about why `SessionCrd` needs so much stuff, rather than just doing some substring matching like `label.contains(&key)` for each label (I don't remember where our key ends up, is it in a label for the session or something else? Anyways, why do we need all this json and string-fu).

Well right now we mainly store it in spec, I was thinking of adding a label but on the other hand I could solve this and any other field needed if I support `field_selector` query param for api-service so the filter can be done with any field like k8s intends

mirrord/operator/src/crd.rs
DmitryDodzin: It shouldn't be, but because a user can set this field to any value I think we should support it (or at least not break by accident)

mirrord/cli/src/config.rs
DmitryDodzin: `#[default]` for enum must be on a unit type variants =| so manual implementation it is

t4lz:
@greptileai review please

mirrord/operator/src/crd.rs
```diff
@@ -528,6 +531,179 @@ pub struct SessionSpec {
...
 
+impl SessionCrd {
+    /// Resolves a Kubernetes `fieldSelector`-style field path against this session, returning the
+    /// value at that path (or [`None`] if the path doesn't resolve to anything).
+    ///
+    /// The path is dot-separated (e.g. `spec.session.key`), with `\.` escaping a literal dot in a
+    /// key; see [`field_path_to_json_pointer`] for the exact path grammar. Resolution is done by
+    /// serializing to [`serde_json::Value`] and walking it, so paths address the serialized shape
+    /// of the CRD, not its Rust field names.
+    pub fn get_field(&self, path: &str) -> Option<serde_json::Value> {
```
meowjesty: These `pub` functions here will be used by the operator, right? Any reason why they're here and not in the operator code? We could implement a `SessionCrdExt` trait if you want them to appear in a `SessionCrd`.

mirrord/operator/src/crd.rs
```diff
@@ -528,6 +531,179 @@ pub struct SessionSpec {
...
+    /// Checks whether this session satisfies a Kubernetes `fieldSelector`.
+    ///
+    /// `selector` is the raw value of the `fieldSelector` query parameter: a comma-separated list
+    /// of `key=value`, `key==value`, and `key!=value` requirements that are ANDed together (all
+    /// must hold). An empty selector matches everything.
+    ///
+    /// Each `key` is a field path resolved with [`Self::get_field`]. Following Kubernetes' own
+    /// semantics, a missing field compares equal to the empty string (so `key=` and `key!=value`
+    /// both match an absent field). A term that can't be parsed (no operator) matches nothing.
+    pub fn matches_field_selector(&self, selector: &str) -> bool {
```
meowjesty: Same pub question here.

meowjesty (changes requested):
Not sure about the whole parser thingy.

I think some parts of it are operator only, so why not implement it there?

mirrord/operator/src/crd.rs
DmitryDodzin: Ehmm well you are correct, I can move this to operator because the usage will be only there (the client side can be hardcoded filter on key only)

meowjesty (approved):
👍

PR #4609, opened 2026-07-27
Make AsyncDecoder::receive cancel safe

aviramha:
Sessions died with `Proxy error, connectivity issue or a bug: connection closed.` when an
application opened a large number of concurrent outgoing connections. The intproxy log showed:

```
ERROR mirrord_intproxy::layer_conn: io failed: Invalid argument (os error 22)
ERROR mirrord_intproxy: Layer connection failed
```

## Cause

`AsyncDecoder::receive` was not cancel safe, and `LayerConnection::run` selects over it:

```rust
tokio::select! {
    res = self.layer_codec_rx.receive() => ...,
    msg = message_bus.recv() => self.send_and_flush(&msg).await?,
}
```

`receive` kept its progress on the stack — it read the 4-byte length prefix, then awaited the
payload. When the outgoing branch won the race, `select!` dropped the receive future and the
already-consumed prefix bytes went with it. The next call then read a message *payload* as a
length prefix.

The resulting length is garbage. It either fails to decode, or asks for a multi-gigabyte read,
which on macOS returns `EINVAL` — `read` rejects `nbyte > INT_MAX`. That is the "Invalid argument"
above. Either way the layer connection dies and takes the session with it.

This needs both directions busy at the same time, which is why it only shows up under a burst of
concurrent connections and why extra logging makes it disappear.

## Evidence

Instrumenting both sides of the codec and comparing the ledgers ruled the writer out: declared
length always matched payload length, and every frame went out in a single `write()` syscall — no
partial writes, no `EINTR`, no errors, all under the `sender` mutex. On the reader side the phase
shift is visible directly. Where a frame prefix belonged, the decoder read `a5 03 00 00`, which is
the payload head of the *previous* message, with the frames after it still well-formed:

```
76 03 00 00 00 0a ... <payload, no prefix>
00 00 00 38 | fb f6 01 03 ...    <- next frame, framed correctly
00 00 00 36 | 78 03 ...          <- and the one after
```

Bytes were being consumed and discarded, not mis-written.

## Fix

Move the read progress into `AsyncDecoder` (`prefix`, `prefix_read`, `payload_len`,
`payload_read`) and loop on cancel-safe `read` instead of `read_exact`, so a dropped future
resumes where it stopped. EOF semantics are unchanged.

## Testing

Reproduced with a Go program opening 1000 concurrent TCP connections to a cluster service through
`mirrord exec`.

- Before: failed on every run at 1000 connections; never at 300.
- After: 1000 connections pass 3/3. 2000 and 4000 no longer desynchronize either — the remaining
  failures there are the single target replica resetting connections, reported per connection with
  the session intact.
- `cargo test -p mirrord-intproxy`: 72 passed. Clippy clean with `--deny warnings`.

Added `receive_resumes_after_being_cancelled_mid_message` in `layer_conn.rs`, next to the
`select!` that motivates the requirement. Confirmed it catches the bug by reverting the fix — it
hangs on the old decoder, so the test is wrapped in a timeout to fail fast rather than hang CI.

### Quality Checklist:

### DIFF
--- a/mirrord/intproxy/protocol/src/codec/codec_async.rs
+++ b/mirrord/intproxy/protocol/src/codec/codec_async.rs
@@ -1,4 +1,4 @@
-use std::io::ErrorKind;
+use std::io::{self, ErrorKind};
 
 use bincode::{Decode, Encode};
 use tokio::{
@@ -64,6 +64,16 @@ where
 pub struct AsyncDecoder<T, R> {
     buffer: Vec<u8>,
     reader: R,
+    /// Length prefix bytes read so far for the message currently being decoded.
+    ///
+    /// Retained across [`AsyncDecoder::receive`] calls to keep it cancel safe.
+    prefix: [u8; PREFIX_BYTES],
+    /// How much of [`AsyncDecoder::prefix`] is filled.
+    prefix_read: usize,
+    /// Payload length, known once the whole prefix has been read.
+    payload_len: Option<usize>,
+    /// How much of the payload was already read into [`AsyncDecoder::buffer`].
+    payload_read: usize,
     _phantom: std::marker::PhantomData<fn() -> T>,
 }
 
@@ -73,6 +83,10 @@ impl<T, R> AsyncDecoder<T, R> {
         Self {
             buffer: Vec::with_capacity(BUFFER_SIZE),
             reader,
+            prefix: [0; PREFIX_BYTES],
+            prefix_read: 0,
+            payload_len: None,
+            payload_read: 0,
             _phantom: Default::default(),
         }
     }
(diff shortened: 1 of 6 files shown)


mirrord/intproxy/protocol/src/codec/codec_async.rs
```diff
@@ -88,23 +107,83 @@ where
...
+                0 => return Poll::Ready(Ok(None)),
+                read => self.prefix_read += read,
+            }
+        }
+
+        let len = match self.payload_len {
+            Some(len) => len,
+            None => {
+                let len = u32::from_be_bytes(self.prefix) as usize;
+                self.buffer.resize(len, 0);
```
aviramha: can just resize if it's bigger, and keep it as max seen so far that is possible to avoid resizing each time.

aviramha: Done in 3f8e8516 — the buffer now only grows, settling at the largest message the connection has carried.

One consequence worth calling out: the buffer is now usually longer than the message sitting in it, so two spots had to become length-bounded rather than buffer-bounded.

- The payload read is now `&mut self.buffer[self.payload_read..len]`. Left unbounded, `poll_read` could fill past the current message and swallow the head of the next frame — the same desync this PR is fixing, just from the other direction.
- The decode is now `&self.buffer[..len]`, so a shorter message that follows a longer one does not pick up the previous one's tail.

Added `decoder_reuses_its_buffer_across_message_sizes` to cover exactly that ordering (512-byte message, then a 1-byte one).

aviramha:
Reopened as #4610 — renaming the branch to pick up the issue ID closed this one instead of retargeting it. Same commits, same branch.

PR #4614, opened 2026-07-27
COR-1706 — detect Java debugger ports from Windows launchers and argfiles

cristeigabriela:
# Windows Java debugger-port detection

## Linear

https://linear.app/metalbear/issue/COR-1706/windows-layer-javaagent-debugger-port-detection-accept-javaexe-expand

## Stack

1 of 4 in the Windows `pitm` and bind-fix stack. This PR is based on `main` and should be reviewed first.

## Summary

Makes the layer's `javaagent` debugger-port detection work for JVMs launched through Windows `mirrord pitm`.

The detector now:

- recognizes `java` and `javaw` launchers with or without the `.exe` suffix, case-insensitively;
- expands Java `@argfile` arguments before looking for `-agentlib:jdwp=...`;
- handles quoted tokens, whitespace, and comments used by IntelliJ and Gradle-generated argfiles;
- keeps the original token when an argfile cannot be read, so detection remains best-effort.

## Motivation

On Windows, `pitm` starts `java.exe` and Gradle commonly places the JDWP argument in an `@argfile`. The existing detector required the first argument to end in `java` and inspected only `std::env::args()`, so it missed both the `.exe` launcher and the hidden JDWP option.

Without the detected port in the debugger-port set, mirrord manages the debugger's loopback socket as application traffic. The JDWP transport then resets during startup with `JDWP TRANSPORT_INIT(510)`.

## Implementation

The change stays in the shared debugger-port module used by the layers. Launcher recognition is basename-aware, and the Java-only argument expansion implements the subset of Java argfile parsing needed by the launchers without changing argument handling for other debugger types.

Focused tests cover Windows launcher spellings, quoted and commented argfiles, unreadable files, non-UTF-8 input, and successful JDWP-port recovery.

## Verification

- `cargo fmt --check`
- `cargo test -p mirrord-layer-lib debugger_ports -- --nocapture` — 27 passed
- `cargo check -p mirrord-layer-win --all-targets --keep-going`
- `cargo xtask build-cli`

### Quality Checklist

- [x] I have documented the non-obvious argfile behavior
- [x] I have tested this change and know it succeeds and fails as expected
- [x] I have written unit tests
- [x] E2E coverage is provided by the IntelliJ Windows Gradle Run/Debug validation for this stack
- [x] I have linked the relevant Linear issue
- [x] I have introduced a short, clear changelog entry

PR #4615, opened 2026-07-27
COR-1708 — kill Windows pitm child trees with a Job Object

cristeigabriela:
# Bind the Windows `pitm` child lifetime to mirrord

## Linear

https://linear.app/metalbear/issue/COR-1708/windows-pitm-bind-the-child-jvms-lifetime-via-a-job-object-orphan-on

## Stack

2 of 4 in the Windows `pitm` and bind-fix stack. Based on #4614; review only the Job Object commit in this PR.

## Summary

Places Windows processes launched by `mirrord pitm` and `mirrord exec` in a Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.

The child is assigned while still suspended, before it can create descendants, and the owning handle is kept alive for the managed process lifetime. If the mirrord wrapper exits normally, crashes, or is terminated by the IDE, Windows tears down the full child tree.

## Motivation

IntelliJ stops the `mirrord pitm` wrapper with `TerminateProcess`. That bypasses Rust destructors, so the previous `terminate_on_drop` cleanup never ran. The injected JVM could remain orphaned and keep its layer connection alive; that kept the intproxy and agent alive as well, leaving agent iptables behind for the next session.

The failure was intermittent because it depended on how the IDE stopped the process tree. The next run then surfaced the secondary symptom: `Detected dirty iptables.`

## Implementation

`LayerManagedProcess::execute` receives an explicit `kill_children_on_exit` policy. On Windows, enabled callers create and configure a Job Object, assign the suspended child to it, and retain the handle in the managed process.

`pitm` and `mirrord exec` enable the policy. The layer's child-process hook disables it because that process is already covered by job inheritance and is intentionally released from the local process wrapper.

The Windows API feature set is updated only for the Job Object calls required by this path.

## Verification

- `cargo fmt --check`
- `cargo xtask build-cli`
- Windows IntelliJ Gradle Run/Debug stop behavior exercised with the combined stack

### Quality Checklist

- [x] I have documented why the Job Object is required
- [x] Unit-test coverage is purposefully omitted for the OS-owned process-tree lifetime behavior
- [x] The behavior was exercised through the Windows IDE integration path
- [x] I have linked the relevant Linear issue and preceding stack PR

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -267,6 +267,7 @@ winapi = { version = "0.3.9", features = [
     "sysinfoapi",
     "synchapi",
     "handleapi",
+    "jobapi2",
     "processenv",
     "winioctl",
     "memoryapi",
--- a/changelog.d/+windows-pitm-kill-children.fixed.md
+++ b/changelog.d/+windows-pitm-kill-children.fixed.md
@@ -1,0 +1,1 @@
+Fixed Windows-native runs (IntelliJ `pitm` and `mirrord exec`) leaking the injected target process when the parent was killed abruptly — e.g. IntelliJ/Gradle stopping a run. The child is now bound to a job object and torn down with the parent, so the layer disconnects and the agent cleans up instead of leaving "dirty iptables" that break the next session.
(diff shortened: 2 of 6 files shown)


cristeigabriela:

PR #4422, opened 2026-06-24
Fix broken windows tests

MintSoup:
### Quality Checklist:

- [X] I have documented new code sufficiently
- [X] I have checked and updated the relevant existing docs in code, including removing outdated material
- [X] I have written user-facing website docs for new features, or opened a (sub)issue to do so
- [X] I have checked and updated existing website docs for changed features
- [X] I have tested this change and know it succeeds and fails as expected
- [X] I have written unit tests or purposefully omitted them
- [X] I have written e2e tests or purposefully omitted them
- [-] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs,
  documentation)
- [x] I have introduced a short, clear and well-formatted changelog entry

https://linear.app/metalbear/issue/COR-1507/add-windows-specific-low-level-integration-tests

PR #4617, opened 2026-07-27
COR-1729 — preserve WinSock bind errors and complete random-port fallback

cristeigabriela:
# Windows managed-bind collision handling

## Linear

https://linear.app/metalbear/issue/COR-1729/preserve-wsagetlasterror-across-the-windows-bind-hook

Follow-up to COR-1496 and https://github.com/metalbear-co/mirrord/pull/4422.

## Stack

4 of 4 in the Windows `pitm` and bind-fix stack. Based on #4616; this PR contains only the Windows bind changes and the corrected stack-size comment.

## Summary

Completes the Windows managed-bind collision behavior and preserves the original WinSock failure that must be returned to the application.

For an unpinned incoming port, a collision on the local requested port now retries on port `0`, mirroring Unix `bind_similar_address`: the remote/requested port remains unchanged while mirrord privately uses an available local port. An explicit `incoming.listen_ports` mapping remains pinned and fails instead of silently changing the configured local port.

The hook also captures `WSAGetLastError` immediately after every failed original `bind` and restores the final error only after the instrumented implementation and its tracing span have returned.

## Root cause

The Windows layer previously had two related problems:

1. A managed socket could fail outright when its local port was occupied, even though the local port is an implementation detail and the remote subscription must continue using the application's requested port.
2. After a real bind failure, tracing and managed-socket bookkeeping ran before control returned to the application. WinSock's last-error value is thread-local, so an intervening Windows call could replace `WSAEADDRINUSE (10048)`. The customer JVM then received the unrelated error `6` and reported `Unrecognized Windows Sockets error: 6: bind`.

## Behavior

- No explicit local-port mapping: retry the local bind with port `0`, record the actual local address from `getsockname`, and retain the user's requested address in managed socket state.
- Explicit `listen_ports` mapping: do not fall back; return the original collision.
- Two mirrord-managed sockets requesting the same address: preserve user-visible bind semantics and return `WSAEADDRINUSE`.
- Any final original-bind failure: restore its captured WinSock error as the last uninstrumented action in the detour.

The exported detour is intentionally uninstrumented. Its inner implementation owns tracing and returns both the result and captured error; only after that span is fully closed does the outer detour call `WSASetLastError` and return.

The same commit corrects the nearby Windows linker comment: `8388608` bytes is an 8 MiB stack, not 32 MiB.

## Verification

A Java `BindCollision` Gradle reproduction was run through a real mirrord agent against fresh BusyBox pods in the `gabriela-debug` namespace.

Validated cases:

- occupied default local port: bind succeeds through an opaque random-local-port fallback while the requested/remote port remains unchanged;
- occupied explicitly mapped local port: bind fails with Java `BindException: Address already in use` and WinSock error `10048`, not error `6`;
- managed socket state continues to expose the user's requested address separately from mirrord's private local binding.

Also passed:

- `cargo fmt --check`
- `cargo xtask build-cli`

### Quality Checklist

- [x] I have documented the requested-address versus private-local-address behavior
- [x] I have tested success and failure paths against a real agent
- [x] The Java bind-collision reproduction covers the regression end to end
- [x] I have explained the root cause and linked the related issue and PR

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,5 +18,5 @@ incremental = false
 
 [target.'cfg(windows)']
 rustflags = [
-    "-C", "link-args=/STACK:8388608", # STACKSIZE=32MB
+    "-C", "link-args=/STACK:8388608", # STACKSIZE=8MB
 ]
--- a/changelog.d/+windows-bind-last-error.fixed.md
+++ b/changelog.d/+windows-bind-last-error.fixed.md
@@ -1,0 +1,1 @@
+Fixed Windows socket hooks replacing a failed `bind` call's WinSock error while closing the tracing span, which made applications receive an unrelated error code.
--- a/mirrord/layer-win/src/hooks/socket/mod.rs
+++ b/mirrord/layer-win/src/hooks/socket/mod.rs
@@ -309,33 +309,60 @@ unsafe extern "system" fn wsa_socket_w_detour(
 }
 
 /// Windows socket hook for bind
-#[mirrord_layer_macro::instrument(level = "trace", ret)]
+///
+/// The instrumented implementation may call Windows APIs while its tracing span is being closed.
+/// Restore the original bind error only after it returns so the caller receives the correct
+/// WinSock error.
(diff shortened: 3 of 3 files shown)


cristeigabriela:

PR #4618, opened 2026-07-27
mirrord chaos CLI command

gememma:
Adds the `mirrord chaos` set of commands, as an alternative way to manage chaos rules without manually curling the chaos endpoints exposed by the local UI server.

Linear: [COR-1537](https://linear.app/metalbear/issue/COR-1537/chaos-cli)

### Quality Checklist:

- [X] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs,<br>documentation)

### DIFF
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -1651,6 +1651,17 @@ pub struct ChaosArgs {
     pub command: ChaosSubcommand,
 }
 
+impl ChaosArgs {
+    pub fn session_id(&self) -> &str {
+        match &self.command {
+            ChaosSubcommand::List { session_id, .. }
+            | ChaosSubcommand::Add { session_id, .. }
+            | ChaosSubcommand::Edit { session_id, .. }
+            | ChaosSubcommand::Delete { session_id, .. } => session_id,
+        }
+    }
+}
+
 /// `mirrord chaos` subcommands.
 #[derive(Subcommand, Debug)]
 pub enum ChaosSubcommand {
--- a/mirrord/cli/src/config.rs
+++ b/mirrord/cli/src/config.rs
@@ -271,8 +271,11 @@ pub(super) enum Commands {
         no_telemetry: bool,
 
         #[clap(flatten)]
-        args: Box<UiCommonArgs>,
+        args: UiCommonArgs,
     },
+
+    /// Manage per-session chaos rules for local chaos testing.
+    Chaos(ChaosArgs),
 
     /// Manage local mirrord sessions.
     #[command(visible_alias = "sessions")]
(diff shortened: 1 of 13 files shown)


mirrord/cli/src/config.rs
```diff
@@ -1640,6 +1644,142 @@ pub enum UiSubcommand {
...
+        match &self.command {
+            ChaosSubcommand::List { rule_id, .. } | ChaosSubcommand::Delete { rule_id, .. } => {
+                rule_id.as_deref()
+            }
+            ChaosSubcommand::Edit { rule_id, .. } => Some(rule_id),
+            ChaosSubcommand::Add { .. } => None,
+        }
+    }
+
+    /// Returns `true` if this command expects a JSON repsonse body.
```
Razz4780: Ah yes, the repsonse 🤔

mirrord/cli/src/config.rs
```diff
@@ -1640,6 +1644,142 @@ pub enum UiSubcommand {
...
+        #[arg(short = 's', long)]
+        session_id: String,
+
+        /// Specific rule to delete. If absent, all rules for this session are cleared.
+        #[arg(short = 'r', long)]
+        rule_id: Option<String>,
+    },
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum, Default, Display)]
```
Razz4780, line 1769: Did you verify that all derives use the same lowercase format? I think `Display` might just use the original casing

mirrord/cli/src/ui.rs
```diff
@@ -591,11 +630,162 @@ pub async fn wizard_command(
...
+        .send()
+        .await
+        .map_err(ChaosApiError::SessionMonitor)?;
+
+    let status = response.status();
+    if !status.is_success() {
+        let _: () = response
+            .json()
+            .await
+            .map_err(ChaosApiError::SessionMonitor)?;
```
Razz4780: This is kinda weird 👀 
If the response is a valid json of `()` (whatever it is, null maybe?), we'll return Ok. If it's not, well return a cryptic response parsing error

mirrord/cli/src/ui.rs
```diff
@@ -591,11 +630,162 @@ pub async fn wizard_command(
...
+                    .unwrap()
+                    .iter()
+                    .for_each(ChaosRule::pretty_print);
+            }
+            (ChaosFormat::Pretty, false) => {
+                println!("Command sucess: status code {status}")
+            }
+            (ChaosFormat::Json, true) => {
+                let thing = response.bytes().await.unwrap();
+                println!("{}", str::from_utf8(&thing).unwrap());
```
Razz4780: Given that the UI server responds with JSONs, this will print make the subcommand print NDJSON instead of promised JSON (since it can make multiple requests, and for each we print a JSON in a new line).

Razz4780 (dismissed):

mirrord/cli/src/ui.rs
Razz4780: Like, someone will specify `ChaosFormat::Json` and expect the output to deserialize into JSON. Won't work if what we actually print is NDJSON

mirrord/cli/src/config.rs
gememma: ~~omg vale you're making me look like an idiot~~ TIL we don't run vale on rust docs 😐

mirrord/cli/src/config.rs
gememma: Image

that `Display` is `strum_macros::Display`, which prefers a `serialize` attribute to using the variant name

mirrord/cli/src/ui.rs
gememma: ugh... yeah ☠️ handling multiple commands in one command sure was a choice that I made

mirrord/cli/src/ui.rs
gememma: Image
Image

I changed it so errors are returned like this instead 👀 in these im calling `add` on two rules, one of which is invalid and fails to be added in the session monitor

aviramha (approved):

PR #4627, opened 2026-07-28
Guard std fds in the layer against processes exec'd with closed stdin

### LINKED ISSUE #4622: Layer's internal-proxy socket can be allocated to fd 0, then broken when `libuv` sets `0_NONBLOCK` on `stdin`
### Bug Description

On MacOS, if a process is `exec`'d with fd 0 closed, `ProxyConnection::new`'s `TcpStream::connect` is handed fd 0. The layer runs in the `dylib` constructor before runtime init, so fd 0 is the lowest free descriptor. The codec's `try_clone` dups it to fd 3.

Node then treats fd 0 as `stdin`. Once anything touches `process.stdin`, `libuv` wraps that fd and sets `0_NONBLOCK` on it. `0_NONBLOCK` lives on the open file description, so the dup on fd 3 is affected too - the layer's socket is now non-blocking on both fds.

The layer's next blocking read returns `EAGAIN`, which surfaces as:

```
Proxy error, connectivity issue or a bug: io failed: Resource temporarily unavailable (os error 35).
```

`graceful_exit!` then SIGKILLs the process.

This needs a parent that closes fd 0 before `exec`, and a child that touches `stdin`.

### Steps to Reproduce

`repro.js`:

```js
if (process.env.TOUCH_STDIN === "1") {
  process.stdin.resume();
  console.log("touched process.stdin");
}
setTimeout(() => {
  require("dns").lookup("example.com", () => { console.log("layer survived"); process.exit(0); });
}, 400);
setTimeout(() => { console.log("timeout"); process.exit(2); }, 8000);
```

`exec 0<&-` closes fd 0 immediately before `exec`, which is what matters - closing `stdin` on the `mirrord` process itself doesn't seem to be enough.

```bash
#fails: exit 137, "os error 35"
mirrord exec -f config.json -- sh -c 'exec 0<&- ; exec env TOUCH_STDIN=1 node repro.js'

#passes: exit 0, "layer survived"
mirrord exec -f config.json -- sh -c 'exec 0<&- ; exec env TOUCH_STDIN=0 node repro.js'
```

### Evidence:

`lsof` on the affected process -  the layer's connection to the internal proxy is on fd 0, dup'd to fd 3 (same device, so one open description):

```
node  77504  gray   0u   IPv4  0xd45788bcd32edac6  0t0  TCP 127.0.0.1:51839->127.0.0.1:51836 (ESTABLISHED)
node  77504  gray   1w    REG  ...
node  77504  gray   2w    REG  ...
node  77504  gray   3u   IPv4  0xd45788bcd32edac6  0t0  TCP 127.0.0.1:51839->127.0.0.1:51836 (ESTABLISHED)
```

From the real-world case (Next.js 16 + Turbopack, ~1000 `lsof` snapshots of transform workers per run, all consistent):

| | layer socket fd | fd 0 holds |
| --- | --- | --- |
| failing | 0 and 3 (977/977 snapshots) | the layer's own socket |
| `MIRRORD_LAYER_LOG_PATH` set | 3 and 4 (1006/1006) | the log file |

### Workaround:
Setting `MIRRORD_LAYER_LOG_PATH` makes the failure disappear entirely: `init_tracing` opens the log file before the proxy connection, so the log file takes fd 0 and the socket lands at fd 3+, where the runtime never touches it.

| condition | fd consumed? | result |
| --- | --- | --- |
| var unset | no | fails |
| var set, pointed at an unwritable dir so `open()` fails | no | fails identically |
| var set, `MIRRORD_LOG=off` so 0 bytes are written | yes | passes |

### mirrord layer logs

Can't be attached - enabling them w/ `MIRRORD_LAYER_LOG_PATH` consumes the fd and suppresses the bug (see workaround above).

### mirrord intproxy logs

These were clean, no `ERROR` lines even when the failure occurs.

### mirrord config

Something simple, like:

```json
{
  "target": { "path": "deployment/frontend/container/frontend", "namespace": "<ns>" },
  "agent": { "namespace": "<ns>", "privileged": true },
  "feature": {
    "fs": { "mode": "local" },
    "network": {
      "incoming": false,
      "outgoing": { "ignore_localhost": true },
      "dns": { "enabled": true }
    }
  }
}
```

### mirrord CLI version

3.238 (also 3.240.0 and 6 older releases)

### mirrord-agent version

_No response_

N/A

N/A

MacOS 26.5.2, arm64 (Darwin 25.5.0)

### Local process

node - Mach-0 64-bit executable arm64

### Local process version

Node v24.14.1

### Additional Info

Found via Next.js 16.2.10 w/ Turbopack, where `@vanilla-extract/next-plugin` registers a webpack-loader rule over every `*.css.ts`. Turbopack runs those loaders in short-lived Node subprocesses spawned w/ fd 0 closed, so every worker hits this and gets SIGKILLed mid-transform. Turbopack then reports:

```
- Execution of evaluate_webpack_loader failed
- failed to receive message
- reading packet length
- unexpected end of file
```

Surfacing as `FATAL: An unexpected Turbopack error occured` and HTTP 500 on every route.

hank-metalbear:
## Summary

Fixes #4622 (Linear: MBE-1990).

When a parent closes a standard fd before `exec` (Next.js 16 with Turbopack does this with stdin for every transform worker), the layer's dylib constructor runs while fd 0 is free. `ProxyConnection::new`'s `TcpStream::connect` is then assigned fd 0, and the codec's `try_clone` dups a second fd sharing the same open file description. Node treats fd 0 as stdin: as soon as the app touches `process.stdin`, libuv sets `O_NONBLOCK` on the shared open file description, the layer's next blocking read fails with `EAGAIN` (os error 35), and the layer kills the process.

The fix adds `guard_std_fds()`, which opens `/dev/null` onto any closed fd in 0-2. It is called at the top of `layer_start` and `load_only_layer_start`, before `init_tracing` and `ProxyConnection::new`, so no long-lived layer fd can be assigned a std fd number.

Per review, the guard is gated behind a new `experimental.guard_std_fds` config, off by default, so current behavior is unchanged until the flag is enabled (standard experimental rollout: affected user enables the flag, then OSS default, then paid). Implemented with the `nix` crate rather than raw `libc`.

Design notes:
- The guard runs only when the layer actually initializes in an exec'd process. It deliberately does not run on the fork-child reconnect path: daemon-style apps legitimately close std fds before forking and may rely on lowest-fd semantics. At exec time no such contract exists.
- Dup'ing the socket above fd 2 at the connect site alone would not fix the bug: the codec's `try_clone` (`make_sync_framed` in `intproxy/protocol/src/codec.rs`) would still land on the free fd 0 and share the same open file description.

## Why this matters for adoption

Turbopack has been Next.js's default dev bundler since v16, so this fires on a mainstream setup rather than an exotic one. A developer trying mirrord on such an app sees every route fail, and the error names the bundler rather than mirrord, so the reasonable conclusion is that mirrord is incompatible with their stack and the evaluation stops there. Nothing in the failure points back to us, so it is lost silently: the reporter here, who spent hours with `lsof` to find the real cause, is the exception rather than the rule.

The trigger is also broader than Next.js. Any parent that closes a standard fd before `exec`, and any child that later touches that fd, hits the same path. Fixing it removes a first-run blocker for a large part of the JS/TS ecosystem.

## Test plan

- New integration test `issue4622` (enables `experimental.guard_std_fds` in its config): runs `sh -c "exec 0<&-; exec node issue4622.js"` through the real `mirrord exec`, matching the reporter's repro. The app touches `process.stdin` and then resolves a name through remote DNS; the test asserts the process survives to a successful exit.
- Verified red-then-green locally on macOS arm64 (debug build): pre-fix the layer dies mid-lookup and the test fails with a dropped intproxy connection; post-fix it passes. In the debug build the pre-fix crash surfaces as a null-pointer UB check in the layer rather than the reporter's release-mode `EAGAIN` + kill; both stem from the intproxy socket landing on fd 0. As a control, the near-identical `node_getaddrinfo_no_name` test (same flow, stdin open) passes on the same pre-fix build.
- Neighboring layer tests pass locally post-fix: `fork`, `spawn`, `exec_functionality`, `bash`, `sip`, `skipping_process`, `dns_resolve`, `node_getaddrinfo_no_name`.
- `cargo clippy -p mirrord-layer -p mirrord-layer-tests --all-targets` reports no issues; `cargo fmt` applied.
- Flag gating verified on the same build: with `guard_std_fds: false` the test fails with the pre-fix layer death, with `true` it passes.
- Verified against a real Next.js 16.2.12 + Turbopack app with a webpack-loader rule (vanilla-extract), targetless against a dev cluster, red-then-green: pre-fix every route returns 500 with worker deaths blamed on Turbopack, post-fix all routes 200 and a live worker shows fd 0 on /dev/null with the layer socket at 3+.
- `mirrord-schema.json` regenerated; `cargo clippy -p mirrord-config -p mirrord-layer -p mirrord-layer-tests --all-targets -- --deny warnings` clean; `cargo fmt` applied.
- Not exercised: a real Next.js 16 + Turbopack app against a cluster (verified via the integration test harness only), and the Linux integration suite (left to CI).


mirrord/layer/src/lib.rs
```diff
@@ -268,9 +268,30 @@ fn layer_pre_initialization() -> Result<(), LayerError> {
...
+/// When the process is `exec`'d with a standard fd closed (Next.js with Turbopack does this with
+/// stdin for its transform workers), any socket or file the layer opens can be assigned fds 0-2.
+/// The application's runtime may then reconfigure the layer's fd as if it was a std stream (e.g.
+/// `libuv` setting `O_NONBLOCK` on stdin), breaking the layer's connection to the internal proxy.
+/// Must be called before the layer creates any long-lived fd, most importantly the
+/// [`PROXY_CONNECTION`] socket.
+/// See [#4622](https://github.com/metalbear-co/mirrord/issues/4622).
+fn guard_std_fds() {
+    for fd in 0..=2 {
+        if unsafe { libc::fcntl(fd, libc::F_GETFD) } == -1 {
```
aviramha: use nix crate instead of libc directly

mirrord/layer/src/lib.rs
```diff
@@ -337,17 +358,20 @@ fn mirrord_layer_entry_point() {
...
+/// 4. Global [`PROXY_CONNECTION`];
 ///
-/// 4. Replaces the [`libc`] calls with our hooks with [`enable_hooks`];
+/// 5. Replaces the [`libc`] calls with our hooks with [`enable_hooks`];
 ///
-/// 5. Fetches remote environment from the agent (if enabled with
+/// 6. Fetches remote environment from the agent (if enabled with
 ///    [`EnvFileConfig::load_from_process`](mirrord_config::feature::env::EnvFileConfig::load_from_process)).
 fn layer_start(config: LayerConfig) {
+    guard_std_fds();
```
aviramha: I don't see a reason not to make this experimental and follow our standard that we make it experimental, turned off - see it works for customer by them enabling flag, then enabling for OSS only then for paid.

aviramha (changes requested):

mirrord/layer/src/lib.rs
hank-metalbear: Done in 563730f63, switched to `nix::fcntl::{fcntl, open}`.

mirrord/layer/src/lib.rs
hank-metalbear: Done in 563730f63. Added `experimental.guard_std_fds`, off by default, gating both call sites, so behavior is unchanged unless the flag is enabled. Schema regenerated and the issue4622 integration test enables the flag in its config. Verified both states locally on macOS arm64: with the flag off the test fails with the pre-fix layer death, with it on it passes.

aviramha (dismissed):

aviramha (approved):

PR #4630, opened 2026-07-29
Forward variadic mode in openat64 and openat$NOCANCEL hooks

aviramha:
## Problem

`openat64_detour` and `openat_nocancel_detour` declared the hooked libc function as taking three fixed arguments. `openat64`/`openat$NOCANCEL` are variadic, so the `mode` passed alongside `O_CREAT` was dropped.

When the call was bypassed and executed locally, the original function was called without `mode` and read whatever occupied that argument slot, so the file was created with arbitrary permission bits.

The other hooks in the family (`open`, `open64`, `open$NOCANCEL`, `openat`) already handled this correctly.

## Fix

Declare both hooks variadic, read `mode`, and pass it through on every path — including the early return taken when the detour guard is unavailable. This required moving them from `hook_guard_fn` to `hook_fn` with a manual guard, same as `openat_detour`.

## Verification

New integration test `cor_1734_file_mode` runs a C app under `fs.mode: localwithoverrides` that creates files on a locally-handled path through every hooked open entry point (`open`, `openat`, `open64`/`openat64` on Linux, `open$NOCANCEL`/`openat$NOCANCEL` on macOS) and asserts the resulting permission bits.

Confirmed on macOS/aarch64 that the test fails before the change and passes after:

```
before: /tmp/cor_1734_openat_nocancel created with mode 0600, expected 0640  FAIL
after:  PASS
```

`mkdir_rmdir`, `readlink`, `realpath`, `statfs`, `pwrite`, `node_close` and `read_from_mirrord_bin` also pass.

Closes COR-1734.


mirrord/layer-tests/tests/configs/cor_1734_local_fs.json
gememma, line 1: instead of a new file, can we do the thing where it's defined in the test and put in a temp file like https://github.com/metalbear-co/mirrord/blob/main/mirrord/layer-tests/tests/dlopen_cgo.rs#L22 ? i think it makes sense when only one test uses it

mirrord/layer-tests/tests/apps/cor_1734_file_mode/cor_1734_file_mode.c
```diff
@@ -0,0 +1,103 @@
...
+#include <sys/stat.h>
+#include <unistd.h>
+
+/// Every `open`-family hook must forward the variadic `mode` argument when it bypasses the call
+/// and runs it locally. A hook that declares the libc function as non-variadic silently drops
+/// `mode`, and the file ends up created with whatever garbage occupied the argument slot.
+///
+/// All paths here live under `/tmp`, which mirrord treats as local, so each call bypasses to libc
+/// and the resulting permission bits must match exactly what was requested.
+#define EXPECTED_MODE 0640
```
gememma: is there any point in running tests with different expected modes? 👀 not sure why it would matter but then again, why not

gememma (dismissed):
LGTM ⚔️ a couple of suggestions but nothing important

mirrord/layer-tests/tests/configs/cor_1734_local_fs.json
aviramha: Done in 5beb345b.

mirrord/layer-tests/tests/apps/cor_1734_file_mode/cor_1734_file_mode.c
aviramha: Done in 5beb345b.

gememma (approved):

PR #4634, opened 2026-07-29
Add `auto_queue_splitting` for copy target

0x00A5:
Introduce `auto_queue_splitting` flag for copy target CRD. When set to true (by `mirrord up` only for now), operator should dismiss queue split configs of disabled queue kinds rather than denying copy target creation.

Downstream operator PR: https://github.com/metalbear-co/operator/pull/2109
Doc update: https://github.com/metalbear-co/docs/pull/292

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/operator/src/client.rs
+++ b/mirrord/operator/src/client.rs
@@ -1356,10 +1356,16 @@ impl OperatorApi<PreparedClientCert> {
             }
 
             let (copied, reused) = {
-                let reused = self.try_reuse_copy_target(layer_config, progress).await?;
+                let reused = self
+                    .try_reuse_copy_target(layer_config, auto_queue_splitting, progress)
+                    .await?;
                 match reused {
                     Some(reused) => (reused, true),
-                    None => (self.copy_target(layer_config, progress).await?, false),
+                    None => (
+                        self.copy_target(layer_config, auto_queue_splitting, progress)
+                            .await?,
+                        false,
+                    ),
                 }
             };
             copy_subtask.success(None);
@@ -1467,7 +1473,9 @@ impl OperatorApi<PreparedClientCert> {
                 operation: OperatorOperation::WebsocketConnection,
             }) if response.code == 404 && reused_copy => {
                 connection_subtask.failure(Some("copied target is gone"));
-                let copied = self.copy_target(layer_config, progress).await?;
+                let copied = self
+                    .copy_target(layer_config, auto_queue_splitting, progress)
+                    .await?;
 
                 let connect_url = Self::copy_target_connect_url(
                     &copied,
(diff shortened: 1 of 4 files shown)


mirrord/operator/src/client.rs
```diff
@@ -2086,6 +2101,7 @@ impl OperatorApi<PreparedClientCert> {
             idle_ttl: Some(Self::COPIED_POD_IDLE_TTL),
             scale_down,
             split_queues,
+            auto_queue_splitting: auto_queue_splitting.then_some(true),
             exclude_containers,
             exclude_init_containers,
         };
```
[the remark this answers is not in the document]
0x00A5: https://github.com/metalbear-co/operator/pull/2109

mirrord/operator/src/client.rs
```diff
@@ -2086,6 +2101,7 @@ impl OperatorApi<PreparedClientCert> {
             idle_ttl: Some(Self::COPIED_POD_IDLE_TTL),
             scale_down,
             split_queues,
+            auto_queue_splitting: auto_queue_splitting.then_some(true),
```
gememma, line 2104: why not `auto_queue_splitting: Some(auto_queue_splitting)` ? 👀 this returns `None` if param is false, right?

mirrord/operator/src/crd/copy_target.rs
```diff
@@ -34,6 +34,11 @@ pub struct CopyTargetSpec {
     /// Init containers that are ignored by copy target.
     #[serde(default)]
     pub exclude_init_containers: Vec<String>,
+    /// When set to `true`, `split_queues` contains wildcard config for all queue kinds.
+    /// Resource creation handler must dismiss unsupported kinds rather than rejecting the
+    /// creation.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub auto_queue_splitting: Option<bool>,
```
gememma, line 41: can we just use bool instead? since it defaults to false anyway

mirrord/operator/src/client.rs
```diff
@@ -2145,6 +2162,7 @@ impl OperatorApi<PreparedClientCert> {
...
```
gememma, line 2165: same as above comment

gememma (approved):
looks good, nits only 👁️

mirrord/operator/src/crd/copy_target.rs
0x00A5: I remember we have some compatibility rules that recommends when adding fields to a CRD, we should prefer `OptionT>` unless it's collections like Vec, HashMap, etc. And we prefer skipping skip serializing if none or if empty, so that new version of CLI and old version of CLI, by default produce the same CRD after serializing.

mirrord/operator/src/client.rs
0x00A5: This is to skip serializing the field intentionally, so the default case produces the same CR after serializing as before (a already installed operator sees the same resource spec before and after a CLI upgrade).

mirrord/operator/src/client.rs
0x00A5: https://github.com/metalbear-co/mirrord/pull/4634#discussion_r3684368161

PR #4635, opened 2026-07-30
Set TCP_NODELAY on all proxying sockets

aviramha:
Every socket in a mirrord session is a relay hop: data written to it was already
read from another socket, where the application on the far end decided how to
frame it. Nagle's algorithm has nothing useful to coalesce on such a socket — it
just holds small writes back waiting for an ACK, adding latency to each hop.
`TCP_NODELAY` was only set on two of them (intproxy → external proxy, and the
console connections).

This sets it on the rest. Accepted sockets do not inherit the option from their
listener, so both ends of each hop needed it.

| Hop | Where |
|---|---|
| layer → internal proxy | `layer-lib/src/proxy_connection.rs`, `intproxy/src/layer_initializer.rs` |
| internal proxy → external proxy | `cli/src/external_proxy.rs` (accept side; connect side already had it) |
| internal / external proxy → agent | `kube/src/api/kubernetes.rs`, `intproxy/src/agent_conn.rs`, `agent/src/entrypoint.rs` |
| agent → outgoing connections | `agent/src/outgoing/socket_stream.rs` |
| redirected incoming → agent | `agent/src/incoming/iptables.rs` |
| agent → passthrough to the target app | `agent/src/incoming/connection.rs` |
| internal proxy → local app (steal / mirror) | `intproxy/src/proxies/incoming/bound_socket.rs`, `.../incoming/http/client_store.rs` |
| local app → internal proxy (outgoing interception) | `intproxy/src/proxies/outgoing/net_protocol_ext.rs` |
| `mirrord port-forward` local sockets | `cli/src/port_forward.rs` |

Not covered, deliberately: unix-domain sockets (outgoing UDS, session monitor,
container runtime) and UDP, where the option does not apply, and the Kubernetes
port-forward path, which rides the API server websocket inside `kube`/`hyper`
rather than a socket we own.

Failing to set the option costs latency, never correctness, so no call site
treats it as fatal — every one logs a warning and carries on. A `setsockopt`
failure cannot break a session or take down a listener.

The two pre-existing `?` propagations in `agent_conn.rs` (internal proxy →
external proxy) are left as they are, since they have already shipped. Happy to
soften those too if we want the rule to hold everywhere.

### Quality Checklist:

- ~[ ] I have written unit tests or purposefully omitted them~ — no behaviour to assert beyond the socket option itself
- ~[ ] I have written e2e tests or purposefully omitted them~


MintSoup (approved):
begone nagle

PR #4641, opened 2026-07-30
Honor AI_NUMERICHOST in getaddrinfo instead of resolving remotely

aviramha:
## What

`getaddrinfo` hooks now bypass when the caller passes `AI_NUMERICHOST`.

## Why

`AI_NUMERICHOST` is not a request for name resolution. It asks whether the given string is *already* a numeric address, and it must fail with `EAI_NONAME` (`WSAHOST_NOT_FOUND` on Windows) when it is not. The hooks passed the flag straight through to the remote lookup and never acted on it, so any hostname came back as a successful "numeric" result. Callers that use this as a cheap "do I need to resolve this at all?" test were told yes for every hostname.

Bypassing costs nothing here. With the flag set, the local `getaddrinfo` performs the same string parse the remote one would and issues no DNS query of its own, so the answer is identical either way and no traffic escapes to the local network.

## How

- `mirrord/layer-lib/src/socket/dns/unix.rs`: bypass before any resolution when the flag is set.
- `mirrord/layer-lib/src/socket/dns/windows.rs`: same check. This needed the caller's `ai_flags` exposed, so `WindowsAddrInfo` gains `get_flags()` alongside `get_family_socktype_protocol()`, implemented for `ADDRINFOA`, `ADDRINFOW` and `ADDRINFOEXW`.
- New `Bypass::NumericHostLookup` so the reason is visible in traces.

## Testing

Unit test asserting a hostname with `AI_NUMERICHOST` bypasses rather than resolving. Verified against a live cluster that the bypass fires for a real application making these calls.

The Windows half could not be compiled locally (no Windows target on this machine) — it is a mechanical addition of one trait method and its three impls, and relies on CI for verification.

## Scope

This was found while investigating COR-1738, but **it does not fix that issue** — the affected runtime still fails to resolve with this applied, so the cause there is something else. Filing it on its own merits as a spec-compliance bug.

### DIFF
--- a/changelog.d/+getaddrinfo-numerichost.fixed.md
+++ b/changelog.d/+getaddrinfo-numerichost.fixed.md
@@ -1,0 +1,1 @@
+`getaddrinfo` calls that pass `AI_NUMERICHOST` are no longer resolved remotely. That flag asks whether the given string is already a numeric address rather than for a name lookup, and must fail when it is not, so resolving it reported every hostname as a literal address.
--- a/mirrord/layer-lib/src/detour.rs
+++ b/mirrord/layer-lib/src/detour.rs
@@ -205,6 +205,9 @@ pub enum Bypass {
     /// Called `getaddrinfo` with `rawish_node` being [`None`].
     NullNode,
 
+    /// Called `getaddrinfo` with `AI_NUMERICHOST`, which asks for no name resolution at all.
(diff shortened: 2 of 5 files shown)


meowjesty (approved):
👍

0x00A5 (approved):

PR #4651, opened 2026-08-02
Sign Windows release artifacts with Azure Artifact Signing

aviramha:
Replaces the DigiCert Software Trust Manager flow in the Windows release job with the Azure artifact-signing action.

## What changed

- **Auth is now GitHub OIDC** instead of a stored client certificate and API key. The job gets `id-token: write`, `azure/login` exchanges the token, and there is no long-lived signing credential to store or rotate. The service principal holds the signer role scoped to a single certificate profile.
- **Removed the two hand-rolled signtool discovery blocks.** The action locates and invokes signtool itself.
- **Binary paths are now absolute** (`Resolve-Path`), so the action resolves them independently of its working directory. The uploaded artifact layout is unchanged — `upload-artifact`'s common ancestor is still the runner work root, so the paths `release_gh` moves from are the same.
- **Added a `Verify signatures` step** that checks the Authenticode status of the EXE, DLL and MSI after signing. The certificates are short-lived and re-issued every few days, so without this a broken chain would only surface once someone tries to install a published artifact.

Both new actions are SHA-pinned.

## Testing

The signing steps only run in the release workflow, so PR CI can't exercise them. To verify, run `Release` via `workflow_dispatch` **selecting this branch** — `workflow_dispatch` uses the workflow definition from the ref you pick, so running it from `main` would exercise the old flow. Manual dispatch builds and signs but publishes nothing.

A federated credential for this branch exists to make that possible; it is temporary and should be deleted once this merges.

## Follow-up

- Delete the temporary branch federated credential after merge.
- The DigiCert secrets are now unused but intentionally left in place until a release has published and installed cleanly.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -156,6 +156,10 @@ jobs:
     needs: [validate_release_request]
     if: github.event_name == 'workflow_dispatch' || needs.validate_release_request.outputs.should_release == 'true'
     runs-on: windows-latest
+    permissions:
+      # Needed to mint the OIDC token that authenticates against Azure Artifact Signing.
+      id-token: write
+      contents: read
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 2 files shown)


cristeigabriela (approved):
code changes look good

binaries produced from action against this branch (https://github.com/metalbear-co/mirrord/actions/runs/30745112698) are correctly signed with the new cert (installer, layer, cli)

PR #4653, opened 2026-08-02
ci: merge approved release PRs as the release bot, and lock manual merges

aviramha:
## What this adds

Two workflows, in both `mirrord` and `operator`.

### `merge-release-pr`

Merges an **approved** release PR as the release bot. The `main-merge-method` ruleset restricts everyone to squash and exempts the bot, so the bot is the only identity that can produce a merge commit — and the release only stays taggable if the PR lands as one.

So **approval replaces merging as the human gate on a release.** Same judgment, different button.

It merges outright when required checks are already green, and falls back to auto-merge when approval arrives first. The bot does **not** bypass `main-protection`, so required checks and the review requirement still bind it — this cannot ship a release with red CI.

### `release-merge-lock`

Stops the mistake happening at all. Reports a `release-merge-lock` commit status that `main-merge-method` requires:

| | status | effect |
|---|---|---|
| ordinary PR | success | merges as usual |
| release PR | **pending** | merge button disabled, reason shown |
| release bot | — | bypasses the ruleset, merges through |

Pending rather than failure: it blocks just as hard without painting every release PR red, and carries a description — *"Release PRs are merged by automation. Approve instead of merging."*

## Why a commit status and not a job result

A skipped job is treated as **passing** for required checks in several configurations. Reporting via the API is deterministic — the status is exactly what we set, with no skipped-job semantics involved. Given this is the thing standing between a human and a broken release, that mattered.

## Context

mirrord 3.244.0 failed exactly this way: release PR squash-merged by hand, branch tip orphaned, `validate_release_request` aborted. Nothing shipped — the guard worked — but main is bumped to 3.244.0 with no release, and every future release would repeat it.

Requires the infra PR adding `release-merge-lock` to `main-merge-method`. Order: land these two first so the status is being reported, then the infra one — otherwise the required check has nothing reporting it and blocks every PR.

## Testing

Both workflows parse. Not exercised end to end — that needs a real release PR, which is the next scheduled run.

### DIFF
--- a/.github/workflows/merge-release-pr.yaml
+++ b/.github/workflows/merge-release-pr.yaml
@@ -23,8 +23,29 @@ permissions:
 
 jobs:
   merge:
+    # `pull_request_review` runs in the context of the base repository, with
+    # access to secrets, even when the pull request comes from a fork. On a
+    # public repository anyone can open a pull request from a fork with any head
+    # branch name they like — including `releases/1.2.3` — and anyone can submit
+    # an approving review on it. Both of those are attacker-controlled, so
+    # neither the branch name nor the approval is worth anything on its own.
+    #
+    # Two conditions make this safe:
+    #
+    #   - the pull request must come from this repository, not a fork. Only the
+    #     release bot can create a `releases/*` branch here, because the
+    #     release-branches-bot-only ruleset blocks creation by anyone else, so a
+    #     same-repository release branch is necessarily one the bot generated.
+    #   - the approval must come from someone with write access. Reviews from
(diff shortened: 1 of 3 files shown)


aviramha:
Superseded by metalbear-co/infra#388, which disables squash merging entirely. The workflows here existed to work around GitHub not being able to require merge commits for release branches specifically — rulesets condition on the base ref, not the head. Making every PR a merge commit gets the same property without a bot-merge trigger, a lock status, or the fork-PR and self-approval guards that trigger needed on a public repo.

PR #4659, opened 2026-08-03
Make CI great again

skreborn:
See https://github.com/metalbear-co/operator/pull/2165 for details.

### DIFF
--- a/.github/scripts/flaky-issue.sh
+++ b/.github/scripts/flaky-issue.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+#
+# Files (or updates) the Linear issue tracking tests that retry too often on `main`.
+#
+# Creates one urgent issue per repository and comments on it thereafter, so a daily report updates a
+# single issue rather than opening one per run.
+#
+# Usage: flaky-issue.sh <repo> <team-key> <body-file>
+#
+# Requires $LINEAR_ACCESS_KEY.
+
+set -euo pipefail
+
+repo=$1
+team_key=$2
+body_file=$3
+
+title="Flaky tests on main: $repo"
+body=$(cat "$body_file")
+
+# GraphQL reports failures in the body with HTTP 200, so the response has to be inspected rather
+# than left to curl's status handling.
+api() {
+  local response
+  response=$(curl --fail-with-body --silent --show-error \
+    -X POST https://api.linear.app/graphql \
+    -H "Authorization: $LINEAR_ACCESS_KEY" \
+    -H 'Content-Type: application/json' \
+    --data "$1")
+
+  if printf '%s' "$response" | jq -e 'has("errors")' > /dev/null; then
+    printf 'Linear API error: %s\n' "$(printf '%s' "$response" | jq -c '.errors')" >&2
+    return 1
+  fi
+
+  printf '%s' "$response"
+}
+
+query=$(jq -n --arg title "$title" --arg key "$team_key" '{
+  query: "query($title: String!, $key: String!) {
+    issues(filter: { title: { eq: $title }, team: { key: { eq: $key } } }) {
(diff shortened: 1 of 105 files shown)


tests/src/utils/cluster_resource.rs
```diff
@@ -67,6 +67,12 @@ pub(super) fn get_pod_template_json_value(
                             host: None,
                             port: IntOrString::Int(80),
                         }),
+                        // Kubernetes' defaults give a startup 30s at 10s granularity, which a test
+                        // app on a loaded e2e node can exceed while still being healthy. Poll more
+                        // often, and allow longer before declaring the pod dead.
+                        period_seconds: Some(2),
+                        failure_threshold: Some(60),
+                        timeout_seconds: Some(5),
```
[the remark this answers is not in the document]
skreborn: Fixed.

.github/scripts/flaky-tests.sh
```diff
@@ -0,0 +1,96 @@
...
+threshold=$4
+urgent_threshold=${5:-10}
+
+work=$(mktemp -d)
+trap 'rm -rf "$work"' EXIT
+
+mkdir -p "$work/reports"
+
+gh api "repos/$repo/actions/workflows/$workflow/runs?per_page=$runs&status=completed&branch=main" \
+  --jq '.workflow_runs[].id' > "$work/runs.txt"
```
vladrbg: The run query filters on `status`, which is the run's *lifecycle* state, not its result:
```bash
gh api "repos/$repo/actions/workflows/$workflow/runs?per_page=$runs&status=completed&branch=main"
```
GitHub run objects carry two separate fields:
| Field | Meaning | Values |
|---|---|---|
| `status` | Did the run finish? | `queued`, `in_progress`, `completed` |
| `conclusion` | How did it end? | `success`, `failure`, `cancelled`, `timed_out`, `skipped` |
`status=completed` includes **failed** runs (they're `completed` + `conclusion=failure`). So a genuinely broken test that fails all 3 attempts on a red `main` run gets its attempts counted as "retries", climbs to the top of the flaky table, and can open an urgent Linear issue - even though it's broken, not flaky.
Fix, either:
- add `&conclusion=success` to the query (only green runs, where a retry means failed-then-passed = truly flaky), or
- in the Python, count `flakyFailure` (passed after retry = flaky) separately from `rerunFailure` (never passed = broken).

.github/scripts/flaky-issue.sh
```diff
@@ -0,0 +1,76 @@
...
+title="Flaky tests on main: $repo"
+body=$(cat "$body_file")
+
+api() {
+  curl --fail-with-body --silent --show-error \
+    -X POST https://api.linear.app/graphql \
+    -H "Authorization: $LINEAR_ACCESS_KEY" \
+    -H 'Content-Type: application/json' \
+    --data "$1"
+}
```
vladrbg, line 37: The helper posts to Linear's GraphQL API:
```bash
curl --fail-with-body --silent --show-error \
  -X POST https://api.linear.app/graphql \
  -H "Authorization: $LINEAR_ACCESS_KEY" ...
```
`--fail-with-body` only fails on HTTP >= 400. GraphQL returns **HTTP 200** even on errors, putting them in an `"errors": [...]` field (or `success: false`) in the body. So a bad token or rejected mutation replies `200 OK`, curl is happy, the step goes green, and the issue is silently never created/updated.
Fix: check the response body (`.errors` / `.data.issueCreate.success`), not just the HTTP status.

vladrbg (dismissed):
Nothing from me, but AI complained about some stuff, so commented them they seem legit but feel free to check them as well.

vladrbg (approved):

vladrbg (dismissed):

vladrbg (approved):

PR #4668, opened 2026-08-04
Name the kubeconfig and context when the operator is not found (COR-445)

hank-metalbear:
## Summary

`mirrord operator was not found in the cluster.` told the user that the lookup failed, but never which cluster mirrord actually searched. The most common cause of that error is not a missing operator, it is a kubeconfig or a `current-context` pointing at a cluster the developer did not intend.

The error help now names the kubeconfig and the context mirrord resolved, and says how to change them:

```
Error:   × mirrord operator was not found in the cluster.
  help: mirrord looked in the cluster reached with kubeconfig
        `/Users/dev/.kube/config` with context `my-staging-cluster`.
        If that is not the cluster you expected, select the right one with
        `kubectl config use-context <context>`, or set `kubeconfig` /
        `kube_context` in your mirrord config.
        Otherwise, this command requires the mirrord operator, or operator
        usage was explicitly enabled in the configuration file.
```

`KubeContextInfo::resolve` follows the same precedence `create_kube_config` already uses: the mirrord config wins over `KUBECONFIG`, which wins over the default path, and an explicitly configured context wins over the kubeconfig's `current-context`. Resolution is best-effort, so an unreadable kubeconfig leaves the field empty rather than replacing the original error.

The path splitting and merging that `create_kube_config` did inline is now shared with the resolver, so the two cannot disagree about which file was read. All seven sites that raise this error carry the information, not just `mirrord exec`.

Closes metalbear-co/private#319.

## Why this matters for adoption

Pointing at the wrong cluster is one of the most common first-run failures, and it is invisible: kubeconfig and context are ambient state set by other tools, IDEs, and shell wrappers, so nothing on screen contradicts the developer's belief about where they are connected. The old message actively pushed the wrong conclusion, that the operator was never installed, which sends people back to re-run the install or open a support ticket instead of switching context. It fails silently in the sense that mirrord is behaving correctly and reporting truthfully, just not with the one fact that resolves the situation. Teams evaluating the operator hit this on day one, before they have any intuition for where to look.

## Test plan

Reproduced first against a local API server returning 404 for every request, which is exactly what a cluster without the operator looks like to the client.

**Before** (on `origin/main`, built with `cargo xtask build-cli`), naming neither source:

```
Error:   × mirrord operator was not found in the cluster.
  help: Command requires the mirrord operator or operator usage was explicitly
        enabled in the configuration file.
```

**After**, same reproduction: names `kubeconfig /tmp/.../kubeconfig-fake` and `context my-staging-cluster`.

Exercised live with the built CLI across every precedence branch:

* `KUBECONFIG` env var set, context from `current-context`
* `kubeconfig` set in the mirrord config file with `KUBECONFIG` unset
* `kube_context` set explicitly in the mirrord config
* `mirrord operator status` against the same 404 cluster, covering a second call site

Happy path checked for regression: `mirrord operator status` against a local cluster that does have the operator still reports version, license, and the session table normally.

Regression tests fail on base and pass with the fix, both states observed. Reverting only the help text (keeping the type change) makes `operator_not_installed_help_names_the_cluster_source` fail with the old wording printed verbatim; restoring it passes.

| Gate | Result |
| -- | -- |
| `cargo xtask test-ut` | 96 passed, 0 failed |
| `cargo test -p mirrord-kube` | 29 passed, 0 failed (7 new) |
| `cargo clippy --all-targets --keep-going -- --deny warnings` | clean, whole workspace |
| `cargo fmt --check` | clean |

Not exercised: a real cluster genuinely missing the operator. The 404 API server reproduces the client-side condition exactly (`OperatorApi::try_new` returning `None`), but no cluster was torn down to confirm it end to end. Docker Desktop's API server was tried first as a realistic target and abandoned because the kube client hung against it, unrelated to this change.

### DIFF
--- a/mirrord/cli/src/error.rs
+++ b/mirrord/cli/src/error.rs
@@ -12,7 +12,7 @@ use mirrord_intproxy::{
     agent_conn::{AgentConnectionError, ConnectionTlsError},
     error::ProxyStartupError,
 };
-use mirrord_kube::error::KubeApiError;
+use mirrord_kube::{api::kubernetes::KubeContextInfo, error::KubeApiError};
 use mirrord_operator::client::error::{HttpError, OperatorApiError, OperatorOperation};
 use mirrord_protocol_io::ProtocolError;
 use mirrord_tls_util::SecureChannelError;
@@ -430,10 +430,12 @@ pub(crate) enum CliError {
 
     #[error("mirrord operator was not found in the cluster.")]
     #[diagnostic(help(
-        "Command requires the mirrord operator or operator usage was explicitly enabled in the configuration file.
+        "mirrord looked in the cluster reached with {0}.
+        If that is not the cluster you expected, select the right one with `kubectl config use-context <context>`, or set `kubeconfig` / `kube_context` in your mirrord config.
+        Otherwise, this command requires the mirrord operator, or operator usage was explicitly enabled in the configuration file.
         Read more here: https://metalbear.com/mirrord/docs/overview/quick-start/#operator.{GENERAL_HELP}"
     ))]
-    OperatorNotInstalled,
+    OperatorNotInstalled(KubeContextInfo),
 
     #[error("mirrord returned a target resource of unknown type: {0}")]
     #[diagnostic(help("{GENERAL_BUG}"))]
@@ -868,6 +870,37 @@ mod tests {
     };
     use tokio::{net::TcpListener, sync::Notify};
     use tokio_rustls::TlsAcceptor;
+
+    use super::*;
+
+    /// A missing operator is usually a kubeconfig pointed at the wrong cluster rather than an
+    /// operator that was never installed, so the help text has to name the kubeconfig and the
+    /// context that mirrord actually used.
+    #[test]
+    fn operator_not_installed_help_names_the_cluster_source() {
+        let error = CliError::OperatorNotInstalled(KubeContextInfo {
+            kubeconfig: Some("/home/dev/.kube/config".to_owned()),
+            context: Some("staging-eu".to_owned()),
+        });
+
+        let help = error
+            .help()
+            .expect("`OperatorNotInstalled` carries a `help` diagnostic")
+            .to_string();
+
+        assert!(
+            help.contains("/home/dev/.kube/config"),
+            "help should name the kubeconfig, got: {help}"
+        );
+        assert!(
+            help.contains("staging-eu"),
+            "help should name the context, got: {help}"
+        );
+        assert!(
+            help.contains("use-context"),
+            "help should tell the user how to switch cluster, got: {help}"
+        );
+    }
 
     /// With this test we're trying to `assert` that our [`kube`] crate is (somewhat)
     /// version-synced with [`rustls`]. To give a friendlier error message on kube requests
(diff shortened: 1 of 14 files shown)


mirrord/cli/src/error.rs
```diff
@@ -869,6 +871,37 @@ mod tests {
...
+        assert!(
+            help.contains("staging-eu"),
+            "help should name the context, got: {help}"
+        );
+        assert!(
+            help.contains("use-context"),
+            "help should tell the user how to switch cluster, got: {help}"
+        );
+    }
+
```
gememma:
```suggestion
```

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -429,3 +517,97 @@ where
...
+    /// An unreadable kubeconfig must not turn into a hard failure, because this type only ever
+    /// decorates an error that is already on its way to the user.
+    #[test]
+    fn resolve_tolerates_an_unreadable_kubeconfig() {
+        let info = KubeContextInfo::resolve(Some("/nonexistent/kubeconfig".to_owned()), None);
+
+        assert_eq!(info.kubeconfig, Some("/nonexistent/kubeconfig".to_owned()));
+        assert_eq!(info.context, None);
+    }
+}
```
gememma:
```suggestion
```

mirrord/kube/Cargo.toml
```diff
@@ -48,6 +48,7 @@ itertools.workspace = true
 
 [dev-dependencies]
 rstest.workspace = true
+tempfile.workspace = true
```
gememma:
```suggestion
```

mirrord/kube/src/api/kubernetes.rs
```diff
@@ -363,6 +365,110 @@ pub struct AgentKubernetesConnectInfo {
...
+            }
+            (Some(kubeconfig), None) => {
+                write!(f, "kubeconfig `{kubeconfig}` with no context selected")
+            }
+            (None, Some(context)) => write!(f, "context `{context}`"),
+            (None, None) => f.write_str("no kubeconfig or context that it could identify"),
+        }
+    }
+}
+
```
gememma, line 413: this whole approach is strange to me - to create the arror, the layer config is needed anyway, so why not just call `mirrord_kube::api::kubernetes::create_kube_config()` and print the `config.cluster_url` instead? or extract out the code that resolves kubeconfig/ context from that function so it can be called from elsewhere if really necessary - it would only be a few lines changed

gememma (commented):
I like the idea to add which cluster mirrord is looking in to this error message, but i think its like a sub-hundred LOC change - also we dont need all these comments repeating the same stuff, its just confusing, and i dont think this these tests are super useful
also wow that description is long, it could probably just be "Added cluster context to operator not found error, linear issue: ABC-123"

hank-metalbear:
Pushed d4546f9: dropped `KubeContextInfo` and its tests, and `OperatorNotInstalled` now calls `create_kube_config` directly and names `config.cluster_url` instead of re-resolving the kubeconfig/context precedence separately. Also removed the low-value test modules and the now-unneeded `tempfile` dev-dependency.

PR #4669, opened 2026-08-04
Expose the current git branch to config templating

aviramha:
Closes COR-1763. Adds a `git_branch` variable to config templating (`mirrord.json` and `mirrord-up.yaml`), so a committed config can derive values from the current branch. The useful case is `key`, which propagates into the generated HTTP filter and gives every branch its own slice of traffic with nobody editing the file.

`get_user_git_branch` moves from the CLI (where analytics already used it) into `mirrord-config`, turns sync, and gains a `OnceLock` cache, so templating and analytics now share one `git` invocation instead of spawning one each.

Templating was documented as a paragraph in the Getting Started preamble; this promotes it to its own `Templating` section in the config reference, covering the available variables, the fact that `key` is read before templating runs (hence single quotes in `key` templates), and the undefined-variable behaviour outside a git checkout.

Verified end to end against a local cluster: on branch `feature-beta`, a request carrying `baggage: mirrord-session=feature-beta` reached the local process while unfiltered requests and ones carrying a different branch stayed on the remote pod. `mirrord up` renders both `git_branch` and `key` in the same filter.


mirrord/up/src/lib.rs
```diff
@@ -143,6 +144,12 @@ fn template(content: &str, key: &EnvKey) -> Result<UpConfig, UpError> {
 
     let mut ctx = tera::Context::new();
     ctx.insert("key", key.as_str());
+    // Same contract as `mirrord.json` templating: left out of the context entirely when the branch
+    // can't be determined, so that `{{ git_branch | default(value='...') }}` gives users a
+    // fallback.
```
meowjesty: Useless comments of the ai explaining to you why it changed something.

mirrord/config/src/util.rs
```diff
@@ -339,3 +348,57 @@ pub fn home_dir_for_path_mapping() -> Option<String> {
...
+///
+/// The result is resolved once and cached for the lifetime of the process, so the config
+/// templating and the analytics reporter share a single `git` invocation.
+pub fn get_user_git_branch() -> Option<String> {
+    static GIT_BRANCH: OnceLock<Option<String>> = OnceLock::new();
+
+    GIT_BRANCH.get_or_init(resolve_user_git_branch).clone()
+}
+
+fn resolve_user_git_branch() -> Option<String> {
```
meowjesty: Why is this not async anymore (not using tokio)?

mirrord/config/src/util.rs
```diff
@@ -339,3 +348,57 @@ pub fn home_dir_for_path_mapping() -> Option<String> {
...
+/// Returns the user's current git branch, or [`None`] if it could not be determined.
+///
+/// Taken from [`MIRRORD_BRANCH_NAME_ENV`] when set, otherwise resolved by running
+/// `git branch --show-current`. A detached HEAD, a working directory outside any repository, or a
+/// missing `git` binary all produce [`None`] rather than an error - the branch is a convenience,
+/// never something a mirrord session depends on.
+///
+/// The result is resolved once and cached for the lifetime of the process, so the config
+/// templating and the analytics reporter share a single `git` invocation.
+pub fn get_user_git_branch() -> Option<String> {
```
meowjesty: oh lol this tricked me.

We java now making getters and setters? Why not just make the `GIT_BRANCH` public.

mirrord/config/src/lib.rs
```diff
@@ -1965,6 +2044,50 @@ mod tests {
...
+                            "override": {
+                                "BRANCH": "{{ git_branch | default(value='no-branch') }}"
+                            }
+                        }
+                    }
+                }"#,
+            )
+            .unwrap();
+
+        let expected = util::get_user_git_branch().unwrap_or_else(|| FALLBACK.to_owned());
```
meowjesty: I'm not sure what we want to test here in this function?

With this `unwrap_or_else`, aren't we always passing this test? If we get a git branch, we pass, and if we don't, then we pass it to because we'll expect the `FALLBACK`, makes no sense to me to test it like this.

mirrord/config/src/lib.rs
```diff
@@ -84,14 +84,78 @@ pub const MIRRORD_LAYER_WAIT_FOR_DEBUGGER: &str = "MIRRORD_LAYER_WAIT_FOR_DEBUGG
...
-/// Currently we don't provide additional values to the context, if you have anything you want us to
-/// provide please let us know.
+/// The configuration supports [templating](#root-templating), so values can be derived at runtime
+/// instead of hardcoded.
 ///
 /// To use a configuration file in the CLI, use the `-f <CONFIG_PATH>` flag.
 /// Or if using VSCode Extension or JetBrains plugin, simply create a `.mirrord/mirrord.json` file
 /// or use the UI.
 ///
+/// ## Templating {#root-templating}
```
meowjesty, line 94: These docs make it seem like the `git_branch` is the "blessed" way of setting a key, is that actually the case? We want everyone to default to using `git_branch`?

If that's not the case, then the 1st example should be whatever we want people to use first, and the `git_branch` example should be shown as an alternative.

meowjesty (changes requested):
Some changes, cause some code doesn't make some sense.

mirrord/config/src/util.rs
aviramha: since it's being used in up where we don't have async contex.

mirrord/config/src/util.rs
aviramha: I think it's nicer to hide "caching" but okay changing it if you insist

mirrord/config/src/lib.rs
aviramha: sorry, slop

mirrord/up/src/lib.rs
aviramha: Removed.

mirrord/config/src/util.rs
aviramha: Config rendering is synchronous, and `mirrord up` resolves its config outside any async context, so an async version would have to be blocked on anyway. Put that reasoning in the doc comment so it doesn't come up again.

mirrord/config/src/util.rs
aviramha: Fair. It's a `pub static GIT_BRANCH: LazyLock<Option<String>>` now, no wrapper - callers do `GIT_BRANCH.clone()` / `GIT_BRANCH.as_ref()`. Caching still hidden, just by `LazyLock` rather than by a function.

mirrord/config/src/lib.rs
aviramha: You're right, it asserted nothing. The branch for templating now resolves through `ConfigContext`, so the test overrides `MIRRORD_BRANCH_NAME` on the context and asserts the literal `my-feature-branch` in both render passes.

Also added `test_quote_style_in_key_template`, which pins down something the manual testing turned up: `key` is read out of the raw file before templating, so a double-quoted filter argument makes the file invalid JSON and the `key` field is *silently* ignored - you get a generated key and a session that looks healthy while filtering on the wrong value. Single quotes work. Documented, but it's arguably worth making that read failure loud in a follow-up.

mirrord/config/src/lib.rs
aviramha: No, it's not the blessed way. Reordered: the templating section now leads with `{{ key }}` in an HTTP filter and says the key is normally set per run via `--key`/`MIRRORD_KEY`, with `git_branch` introduced after as one way to source it from the environment instead. Same on the `key` field docs.

mirrord/config/src/util.rs
meowjesty: Hmm I don't know how I feel about that, everywhere else we just let the static be pub, cause it's already behind a `get_or_init`. I think this is a bit too java-y imo.

mirrord/config/src/lib.rs
```diff
@@ -123,6 +121,16 @@ pub const MIRRORD_LAYER_WAIT_FOR_DEBUGGER: &str = "MIRRORD_LAYER_WAIT_FOR_DEBUGG
 /// }
```
meowjesty, line 121: This reads a bit weird.

It should be something more like:

> mirrord will automatically generate a key, and you can use the templating {{ key }} in your HTTP filter like so:

Then the example.

And then say something like:

> It also supports setting your git branch as the key using templating, like this:

Show the same example as before, but with the `key: git_branch` now.

meowjesty (changes requested):
One last comment on docs.

mirrord/config/src/util.rs
aviramha: Inlined the initialiser so it's just the static and a closure, no named resolver function - I think that named fn was the java-y bit.

On the type itself I kept `LazyLock`: `pub static X: LazyLock<..>` is what the workspace already uses for computed statics (`protocol::VERSION`, `vpn::MINIMAL_PROTOCOL_VERSION`, the `VersionReq`s in `protocol/src/error.rs` - 15 of them). The two `pub static .. OnceLock` cases I found, `layer-lib`'s `SETUP` and `PROXY_CONNECTION`, are the inject-once kind that someone `set`s, which is a different job from compute-on-first-read. Happy to switch to `OnceLock` + caller-side `get_or_init` if you'd rather, just say.

mirrord/config/src/lib.rs
aviramha: Restructured to that shape: generated key with `{{ key }}` in the filter first, then the same config repeated with `"key": "{{ git_branch }}"` added.

meowjesty (approved):
👍

PR #4671, opened 2026-08-04
COR-1729 — preserve Windows bind errors

cristeigabriela:
# Preserve Windows bind errors

## Linear

- https://linear.app/metalbear/issue/COR-1729

## Summary

Preserves the real WinSock error returned by the Windows `bind` hook and lets ordinary managed binds complete the existing random-local-port fallback before considering whether elevation is required.

The requested port remains the remote subscription identity. When `incoming.listen_ports` does not pin the local port, mirrord may bind the local socket to an ephemeral port without exposing that implementation detail to the application.

## Motivation

The Windows bind detour called the original `bind`, then performed tracing and managed-socket work before returning. WinSock's last error is thread-local, so intervening Windows calls could replace `WSAEADDRINUSE (10048)` with error `6`. Java consequently reported:

```text
java.net.SocketException: Unrecognized Windows Sockets error: 6: bind
```

Preserving the actual `WSAEACCES` also made an ordering bug observable in the Windows HTTP tests. An ordinary managed bind to a privileged requested port logged that it would retry on a random local port, but the UAC diagnostic terminated the process before the fallback ran.

This is a deterministic bind-path correction rather than a test-flake workaround. Host port and elevation state only determine whether the affected path is reached.

## Implementation

- keeps the exported bind detour uninstrumented;
- captures each failed `bind` immediately;
- restores the final error after the instrumented implementation and tracing span return;
- distinguishes retryable primary failures from final failures;
- allows ordinary managed binds to complete the existing random-local-port fallback;
- retains the UAC diagnostic when `incoming.listen_ports` pins a privileged local port and prevents fallback;
- keeps the requested remote subscription port and managed user-visible socket address unchanged;
- complements the random-local-port fallback from #4422;
- corrects the Windows linker comment from 32 MB to the configured 8 MB stack size.

## Verification

- `cargo fmt --all -- --check`;
- `cargo check -p mirrord-layer-win --all-targets --keep-going`;
- `cargo clippy -p mirrord-layer-win --all-targets -- --deny warnings`;
- `cargo xtask build-cli` on Windows from the independent split branch;
- Java bind-collision reproduction through a real mirrord agent: fixed local-port failures return `BindException: Address already in use`, not error `6`;
- occupied managed port: mirrord selects a free local port while the remote subscription retains the requested port and managed `getsockname` returns the user-visible address;
- focused Windows HTTP mirroring with the same bind commit: the Node application requested port 80, subscribed to `Tcp(PortSubscribe(80))`, completed GET/POST/PUT/DELETE, and passed without elevation.

### Quality Checklist:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,5 +18,5 @@ incremental = false
 
 [target.'cfg(windows)']
 rustflags = [
-    "-C", "link-args=/STACK:8388608", # STACKSIZE=32MB
+    "-C", "link-args=/STACK:8388608", # STACKSIZE=8MB
 ]
--- a/changelog.d/+windows-bind-last-error.fixed.md
+++ b/changelog.d/+windows-bind-last-error.fixed.md
@@ -1,0 +1,1 @@
+Fixed Windows applications reporting the wrong error when a requested local port was unavailable.
--- a/mirrord/layer-win/src/hooks/socket/mod.rs
+++ b/mirrord/layer-win/src/hooks/socket/mod.rs
@@ -309,33 +309,60 @@ unsafe extern "system" fn wsa_socket_w_detour(
 }
 
 /// Windows socket hook for bind
-#[mirrord_layer_macro::instrument(level = "trace", ret)]
+///
+/// The instrumented implementation may call Windows APIs while its tracing span is being closed.
+/// Restore the original bind error only after it returns so the caller receives the correct
+/// WinSock error.
 unsafe extern "system" fn bind_detour(s: SOCKET, name: *const SOCKADDR, namelen: INT) -> INT {
+    let (result, last_error) = unsafe { bind_detour_impl(s, name, namelen) };
+    if let Some(last_error) = last_error {
+        unsafe { WSASetLastError(last_error) };
+    }
+
+    result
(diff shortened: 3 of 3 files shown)


changelog.d/+windows-bind-last-error.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed Windows socket hooks replacing a failed `bind` call's WinSock error while closing the tracing span, which made applications receive an unrelated error code.
```
cristeigabriela: this is too technical

.cargo/config.toml
```diff
@@ -18,5 +18,5 @@ incremental = false
 
 [target.'cfg(windows)']
 rustflags = [
-    "-C", "link-args=/STACK:8388608", # STACKSIZE=32MB
```
0x00A5, line 21: Huh, thanks for catching this. Was supposed to be unchanged as 8Mb in the previous PR.

0x00A5 (approved):

.cargo/config.toml
cristeigabriela: no worries, me and areg were aware of this

PR #4673, opened 2026-08-05
INT-494: Longer lived port forwarding

giIuis:
This PR introduces a daemon which holds ui server, chaos and DB port forwarding management.

The original ticket required that DB port forwards weren't session specific. So, after this PR DB port forwards are handled by the daemon., that is, a single DB port forward is created (per forward), which mirrord sessions can attach to. Once no mirrord sessions are attached, the portforward is dropped. 

My initial implementation just created a "DB port forwarding manager" process, but since ui-server already provided all the needed functionality (separate process that outlives mirrord session that creates it, already handles authentication), then it made sense that those two concepts shared a home, in the form of the mirrord daemon.
The deamon is started automatically anytime a mirrord session is started, and persists it. 

Created MBE-2074 to handle later simplification of the daemon shutdown.

### Quality Checklist:

  documentation)


giIuis:
@greptileai have at it

giIuis:
@greptileai review again

mirrord/cli/src/internal_proxy/db_portforwards.rs
```diff
@@ -36,8 +35,8 @@ pub(crate) enum SetupError {
     #[error("agent connection dropped unexpectedly")]
     AgentConnectionDropped,
 
-    #[error("failed to set up port forwarder: {0}")]
-    PortForwarder(#[from] port_forward::PortForwardError),
+    #[error("failed to attach DB branch port forward to the local daemon: {0}")]
+    Daemon(#[from] UiCliError),
```
meowjesty, line 39: Any way we can print something like this to the user when the error happens?

Image

mirrord/cli/src/ui/server.rs
```diff
@@ -311,6 +314,10 @@ pub struct AppState {
     /// poll. Only the v2 (per-request, context-aware) handlers use this; the v1 background
     /// watcher keeps its own current-context client.
     pub(crate) clients: Arc<RwLock<HashMap<Option<String>, Client>>>,
+    /// DB branch forwards owned by the local daemon and shared by active mirrord sessions.
+    pub(crate) db_portforwards: db_portforwards::DbPortForwards,
+    /// Whether browser-facing UI routes are enabled. Internal daemon APIs remain available.
+    pub(crate) ui_enabled: Arc<AtomicBool>,
```
meowjesty: Any reason to not always enable it?

mirrord/cli/src/ui/server.rs
```diff
@@ -1059,6 +1073,36 @@ async fn health() -> impl IntoResponse {
...
+    State(state): State<AppState>,
+    request: Request,
+    next: Next,
+) -> Response {
+    if state.ui_enabled.load(Ordering::Acquire) {
+        next.run(request).await
+    } else {
+        StatusCode::SERVICE_UNAVAILABLE.into_response()
+    }
+}
```
meowjesty: Yeah I don't see what we get from disabling/enabling the ui routes like this?

mirrord/cli/src/ui.rs
```diff
@@ -511,12 +572,131 @@ fn ui_start_printout(
...
+        std::fs::read_to_string(mirrord_dir::get_path_or_fallback().join(TOKEN_FILE_NAME)),
+    ) {
+        let client = reqwest::Client::new();
+        if client
+            .get(format!("http://{}/api/internal/ping", info.addr))
+            .header(TOKEN_HEADER_NAME, token.trim())
+            .timeout(Duration::from_secs(1))
+            .send()
+            .await
+            .is_ok_and(|response| response.status().is_success())
```
meowjesty: So if this ping fails, we assume no ui running, and start it?

mirrord/cli/src/ui.rs
```diff
@@ -511,12 +572,131 @@ fn ui_start_printout(
...
+/// ping succeeds. Otherwise this starts a daemon without enabling its browser UI, then reads the
+/// newly published discovery information. Incompatible protocol versions are never reused.
+pub(crate) async fn ensure_daemon() -> Result<DaemonClient, UiCliError> {
+    if let (Some(info), Ok(token)) = (
+        read_daemon_info(),
+        std::fs::read_to_string(mirrord_dir::get_path_or_fallback().join(TOKEN_FILE_NAME)),
+    ) {
+        let client = reqwest::Client::new();
+        if client
+            .get(format!("http://{}/api/internal/ping", info.addr))
```
meowjesty: Do we need a special route for this? Can't we just hit `/`?

mirrord/cli/src/ui/server.rs
meowjesty: Hmm I guess it's because the daemon gets started when you start a session, and `ui` is an explicit start?

`mirrord ui start` should be pretty quick to always start though, so what do you think of always starting it? Starting the whole thing doesn't seem to me like it would be any slower than just the daemon part?

I think it's fine to treat the ui as the daemon, thoughts?

meowjesty (commented):
Commenting, since I want to understand some stuff first.

I wonder why not just treat the whole ui as the daemon, start it if it's not running (instead of just the daemon), and don't allow `mirrord ui stop` to work if there's an active session ongoing, that needs the daemon (show some error to the user saying for db-branch we need the mirrord daemon running, yadda yadda yadda, stop the sessions [list of sessions we know are running and need the daemon] before running `mirrord ui stop` again [show command on how to kill the sessions]).

Something along those lines.

mirrord/cli/src/internal_proxy/db_portforwards.rs
giIuis: Only if you find a wh40k.rs crate that can turn that into asciiart

mirrord/cli/src/ui/server.rs
giIuis: The reason I left that in was to preserve the old behavoiur of "activating" and "deactivating" the ui from the command line, in case any user was counting on that / had that as part of their workflow.
After thinking a bit more, imo the best option is this:
- ui starts with daemon, stops with daemon. as long as daemon runs, ui server is up.
- `mirrord ui start` still works, but just opens the browser. (which I guess is what most users actually expect from it, rather than starting a background process, server etc.)
- Daemon (and ui server) can be killed with `mirrord ui stop` 
        - if any sessions are still using db port forwarding, `mirrord ui stop` just says "hey, can't stop because x sessions are still attached to db port forwarding" 
        - alternativelly, maybe `mirrord ui stop` could be removed, and a new command `mirrord daemon stop` could be added? guess it's more descriptive of what's actually happening

What do you think about this?

mirrord/cli/src/ui/server.rs
giIuis: Yeah, now it don't make sense to me either. Check above comment

DmitryDodzin (dismissed):

mirrord/cli/src/ui.rs
DmitryDodzin: I would keep a special route because fail on `/` potentially can be missed (if the / returns a user readable error aka a 200 response just with an error screen) not sure it's the case now but can change at anytime so I think it's better to use a special endpoint just for it

mirrord/cli/src/ui/server.rs
meowjesty: > alternativelly, maybe mirrord ui stop could be removed

Instead of removing, I think we could alias it. But imo, at least initially, it's fine to leave it under `mirrord ui stop` (don't think we should remove it right now either).

About the rest: I like this approach, imo it's simpler to just run the whole thing and it makes sense to have `mirrord ui start` just open the browser, but hmm, what happens if the user is not running any sessions yet (so no daemon started, right?) and they run `mirrord ui start`? So I guess it still has to be able to start the whole thing.

Basically, a couple of cases:

- The user starts their first mirrord session of the day.

1. We start the daemon (an equivalent to `mirrord ui start`, but without automatically opening the browser);
2. The user does their thing, and **stops** their mirrord session;
3. [we keep the whole backend running?]

- User starts another mirrord session.

1. No need to start the backend process, it's already running;
2. They run `mirrord ui start`, cause they want to see some stuff in the browser;
3. [I think we already support running that command when the backend is already running, so we just keep doing what we're doing, not restarting anything, just shows the links and stuff and open the browser];

And if they have no sessions running, `mirrord ui stop` kills the whole backend.

Is that it?

giIuis: yes, that's it. Will implement in a bit, ping when done

mirrord/cli/src/ui/server.rs
```diff
@@ -1061,6 +1073,27 @@ async fn health() -> impl IntoResponse {
...
+async fn daemon_shutdown(State(state): State<AppState>) -> Response {
+    match db_portforwards::request_daemon_shutdown(&state.db_portforwards, &state.shutdown).await {
+        Ok(()) => StatusCode::NO_CONTENT.into_response(),
+        Err(sessions) => (
+            StatusCode::CONFLICT,
+            axum::Json(DaemonShutdownBlocked { sessions }),
+        )
+            .into_response(),
+    }
+}
```
meowjesty: Nit: move the daemon stuff (and the daemon `Router::new` to its own module.

So in the `build_router` we can just call it like `.nest("/api/internal|daemon|whatever", daemon_router());`

mirrord/cli/src/ui.rs
```diff
@@ -588,11 +764,69 @@ pub async fn ui_stop(with_printouts: bool) -> Result<(), UiCliError> {
...
+    for _ in 0..200 {
+        if read_daemon_info().is_none() {
+            let _ = std::fs::remove_file(mirrord_dir::get_path_or_fallback().join(PID_FILE_NAME));
+            if with_printouts {
+                println!("* Local mirrord daemon stopped");
+            }
+            return Ok(());
+        }
+        tokio::time::sleep(Duration::from_millis(25)).await;
+    }
```
meowjesty: What's going on here?

The server is shutting down, but is keeping the guard file locked, and thus we cannot remove it yet in here? So we keep checking if the daemon file doesn't exist with `read_daemon_info`, and sleeping while it still exists?

Only when we detect that the daemon file is gone, we can actually remove server locking file. Is that it?

mirrord/cli/src/ui.rs
```diff
@@ -588,11 +764,69 @@ pub async fn ui_stop(with_printouts: bool) -> Result<(), UiCliError> {
...
 }
 
-/// Runs the `mirrord ui` command to either start or stop the UI server.
-/// If starting a new server fails, runs `mirrord ui stop` silently to eliminate any leftover
-/// process or files, ignoring errors.
+/// Stops the local daemon unless active sessions still claim daemon-owned DB port forwards.
+///
+/// The shutdown decision is made inside the daemon while holding the forward registry lock. This
+/// prevents a session from attaching between the safety check and process termination.
+pub async fn ui_stop(with_printouts: bool) -> Result<(), UiCliError> {
```
meowjesty: Why has this function changed so much?

Shouldn't it be pretty much the same as it was, except that now we check if there's any mirrord session that needs the ui (daemon) running, and if that's the case, we don't let it shutdown?

mirrord/cli/src/ui.rs
```diff
@@ -588,11 +764,69 @@ pub async fn ui_stop(with_printouts: bool) -> Result<(), UiCliError> {
...
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
+                if with_printouts {
+                    println!("* No running local mirrord daemon was found");
+                }
+                return Ok(());
+            }
+            Err(error) => return Err(error.into()),
+        };
+    let response = reqwest::Client::new()
+        .post(format!("http://{}/api/internal/shutdown", info.addr))
```
meowjesty: When the server is going down, won't the whole process go down with it? Why do we need an explicit "shutdown" sent to the server itself here? Won't all the resources get cleaned up on drop?

Or am I misunderstanding what's going on here?

mirrord/cli/src/ui.rs
```diff
@@ -500,23 +554,143 @@ fn ui_start_printout(
...
+    }
+}
+
+/// Discovers or starts the local daemon and returns an authenticated internal client.
+///
+/// `daemon.json` supplies the loopback address and compatible protocol version, while the token
+/// file supplies authentication. Existing discovery data is trusted only after an authenticated
+/// ping succeeds. Otherwise this starts a daemon without enabling its browser UI, then reads the
+/// newly published discovery information. Incompatible protocol versions are never reused.
+pub(crate) async fn ensure_daemon() -> Result<DaemonClient, UiCliError> {
```
meowjesty: Nit: move all the daemon stuff to its own module.

meowjesty (changes requested):
Some more comments:

I think the whole shutdown thing can still be simplified further (unless I'm misunderstanding some stuff here, which is highly plausible).

But it feels like `ui_stop` is more complicated than it has to be? I would assume that things would clean up after themselves when they're dropped (the server is getting shutdown).

And I don't really understand why we need 2 lockfiles, 1 for the server and 1 for the daemon? Aren't they the same process?

meowjesty (dismissed):
Good luck with CI!

meowjesty (approved):

PR #4675, opened 2026-08-05
deps: update kube fork to 4.2.0

aviramha:
Moves the `kube` fork pin from `ae49cce` to `6cbb2e8` (the fork's current `main`), taking it from 4.0.0 to 4.2.0.

## What this brings in

- `TCP_NODELAY` is set on the default connector, so requests to the Kubernetes API server are not held back by Nagle's algorithm.
- Ordering derives on label selector expressions.
- Upstream `kube` 4.0.0 → 4.2.0.

`k8s-openapi` stays on 0.28, so no API surface change there. The fork's MSRV moves to 1.89, which is well under the pinned toolchain.

## Verification

- `cargo check --workspace --all-targets` passes (CLI checked separately with a built layer via `MIRRORD_LAYER_FILE`).
- `cargo clippy` clean on every kube-dependent crate: `mirrord-kube`, `mirrord-operator`, `mirrord-auth`, `mirrord-vpn`, `mirrord-up`, `mirrord-intproxy`, `mirrord`.

No code changes were needed — the bump is source-compatible.

## Note for the operator

The operator pins `kube` to the same rev as the one used by its mirrord dependencies, so its bump also has to move its mirrord pin to this commit. That PR is up and blocked on this one merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,7 +307,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


cristeigabriela (approved):
haha 4.2.0

PR #4677, opened 2026-08-05
Send session analytics at session end instead of relying on runtime teardown

hank-metalbear:
## Summary

Since 3.222.0 (#4391), the internal proxy's `AnalyticsReporter` is moved into the session monitor API server task. Nothing stops that task when the session ends, so the reporter is only dropped during tokio runtime teardown, where the send task it spawns from `Drop` never runs. Two user-visible effects:

- Successful `mirrord exec` sessions are never reported to telemetry (the CLI-side reporter is error-only, so the intproxy report was the only success signal). Error reports still go out, which heavily skews any error-rate reading of `client_session_v1`.
- Every intproxy exit stalls for the full 10-second drain timeout (`Failed to drain in a timely manner, ongoing tasks dropped.`), because the reporter's `drain::Watch` is never released.

Reference counting alone cannot release the reporter: axum spawns a task per connection, and a consumer holding `/events` open (e.g. `mirrord ui`) keeps its connection task, and with it the app state, alive indefinitely.

Changes:

- `SessionAnalytics`: shared handle around the chaos analytics reporter; the intproxy takes the reporter back at session end (`take_reporter`, folding chaos rule analytics in exactly like `Drop`) and drops it while the runtime is live, so the report is sent regardless of API server state.
- `proxy()` shuts the session monitor down after `IntProxy::run` returns: cancels the server's token, joins the server task with a bounded wait, then drops the reporter.
- `serve_session` hard-stops the server one second after cancellation if graceful drain is still blocked by open SSE streams; the session info updater task also exits on cancellation. This also makes the socket cleanup guard actually run at session end.
- With the API server disabled (`api: false`), the reporter is now held until session end instead of being dropped (and reporting) at session start.

## Test plan

- `cargo clippy -p mirrord -p mirrord-intproxy --all-targets --all-features -- --deny warnings` clean, `cargo fmt`.
- `session_monitor_round_trip` tests pass, including a new regression test: external cancellation stops the server within the grace period while a consumer holds an open `/events` SSE stream.
- Live A/B against minikube (real `mirrord exec` with a pod target, `mirrord ui` running so an SSE consumer is attached), observing outbound requests through a local logging proxy:
  - Released 3.221.1: analytics request sent at intproxy shutdown, no drain warning.
  - Released 3.222.0: session succeeds, no analytics request ever, intproxy's last log line is the drain-timeout warning 15s after the session.
  - This branch: analytics request sent immediately after "Session monitor API server stopped", no drain warning, intproxy exits ~6s after the session (5s idle timeout + 1s connection grace) instead of 16s.

## Follow-up (second commit)

An independent adversarial review flagged that an `IntProxy::run` startup error (first-connection timeout) would be reported as a successful session, a quirk carried over from the pre-3.222.0 behavior. The second commit maps `ProxyStartupError::ConnectionAcceptTimeout` to the previously unused `AnalyticsError::IntProxyFirstConnection` before the report is sent, and logs session monitor server task join failures instead of ignoring them.

### DIFF
--- a/changelog.d/+intproxy-session-analytics.fixed.md
+++ b/changelog.d/+intproxy-session-analytics.fixed.md
@@ -1,0 +1,1 @@
+Fixed successful sessions not being reported to mirrord telemetry: the session monitor API server kept the analytics reporter alive past runtime shutdown, which also stalled internal proxy exit for the 10-second drain timeout.
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -23,7 +23,7 @@ use std::{
 };
 
 use mirrord_analytics::{
-    AnalyticsReporter, CollectAnalytics, Reporter, read_correlation_id_from_env,
+    AnalyticsError, AnalyticsReporter, CollectAnalytics, Reporter, read_correlation_id_from_env,
     read_kube_version_from_env,
 };
 use mirrord_config::{
@@ -33,6 +33,7 @@ use mirrord_config::{
 use mirrord_intproxy::{
     IntProxy, IntProxyIntervals,
     agent_conn::{AgentConnectInfo, AgentConnection},
+    error::ProxyStartupError,
     session_monitor::{
         MonitorTx,
         api::SessionAnalytics,
@@ -129,16 +130,24 @@ struct SessionMonitorHandle {
 }
 
 impl SessionMonitorHandle {
-    async fn shutdown(self) {
+    /// `error` is attached to the session report before it is sent, since the reporter is out
+    /// of the intproxy's reach for the whole session.
+    async fn shutdown(self, error: Option<AnalyticsError>) {
         self.shutdown.cancel();
-        if let Some(server) = self.server
-            && tokio::time::timeout(Duration::from_secs(5), server)
-                .await
-                .is_err()
-        {
-            tracing::warn!("Session monitor API server did not shut down in time");
+        if let Some(server) = self.server {
+            match tokio::time::timeout(Duration::from_secs(5), server).await {
+                Ok(Ok(())) => {}
+                Ok(Err(join_error)) => {
+                    tracing::warn!(%join_error, "Session monitor API server task failed")
+                }
+                Err(_) => tracing::warn!("Session monitor API server did not shut down in time"),
+            }
         }
-        drop(self.analytics.take_reporter().await);
+        if let Some(mut reporter) = self.analytics.take_reporter().await
+            && let Some(error) = error
+        {
+            reporter.set_error(error);
+        }
     }
 }
 
@@ -405,7 +414,10 @@ pub(crate) async fn proxy(
     .run(first_connection_timeout, consecutive_connection_timeout)
     .await;
 
-    session_monitor.shutdown().await;
+    let analytics_error = result.as_ref().err().map(|error| match error {
+        ProxyStartupError::ConnectionAcceptTimeout => AnalyticsError::IntProxyFirstConnection,
+    });
+    session_monitor.shutdown(analytics_error).await;
 
     result.map_err(From::from)
 }
--- a/mirrord/cli/src/internal_proxy.rs
+++ b/mirrord/cli/src/internal_proxy.rs
@@ -35,6 +35,7 @@ use mirrord_intproxy::{
     agent_conn::{AgentConnectInfo, AgentConnection},
     session_monitor::{
         MonitorTx,
+        api::SessionAnalytics,
         chaos::{ChaosWatcherRx, ChaosWatcherTx},
     },
 };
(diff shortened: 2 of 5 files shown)


changelog.d/+intproxy-session-analytics.fixed.md
gememma, line 1: should be `internal` instead of `fixed` imo

mirrord/cli/src/internal_proxy.rs
```diff
@@ -114,21 +116,70 @@ fn print_addr(listener: &TcpListener) -> io::Result<()> {
     Ok(())
 }
 
+/// Handle for tearing down the session monitor API server when the session ends.
+///
+/// The session's [`AnalyticsReporter`] sends the session report from its `Drop`, which spawns
+/// onto the runtime, so [`SessionMonitorHandle::shutdown`] must run at session end, while the
+/// runtime is still live. Otherwise the reporter is dropped during runtime teardown, the
+/// report is lost, and the held [`drain::Watch`] stalls process exit until the drain timeout
+/// expires.
```
gememma, line 125:
```suggestion
/// The session's [`AnalyticsReporter`] sends analytics via `Drop` which requires the runtime to
/// still be live. Otherwise, `Drop` is ignored during teardown and analytics will be missing.
```

mirrord/cli/src/internal_proxy.rs
```diff
@@ -114,21 +116,70 @@ fn print_addr(listener: &TcpListener) -> io::Result<()> {
...
+struct SessionMonitorHandle {
+    shutdown: CancellationToken,
+    server: Option<tokio::task::JoinHandle<()>>,
+    analytics: SessionAnalytics,
+}
+
+impl SessionMonitorHandle {
+    /// `error` is attached to the session report before it is sent, since the reporter is out
+    /// of the intproxy's reach for the whole session.
```
gememma, line 134:
```suggestion
    /// Shuts down the [`SessionMonitor`] by cancelling `self.shutdown` with 5 seconds grace, then
    /// explicitly attaches `error` to the [`AnalyticsReporter`].
```

mirrord/cli/src/internal_proxy.rs
```diff
@@ -114,21 +116,70 @@ fn print_addr(listener: &TcpListener) -> io::Result<()> {
...
+    }
+}
+
 /// Starts the session monitor API server if enabled.
 ///
 /// `@analytics`: optionally, pass a reporter in to be used by the chaos router for chaos metrics
 /// reporting. If `None`, the chaos router will work as normal but will not report metrics.
+///
+/// The reporter is shared with the API server through [`SessionAnalytics`]; the returned
+/// [`SessionMonitorHandle`] must be shut down at session end for the session report to be sent.
```
gememma, line 160:
```suggestion
///
/// [`SessionMonitorHandle`] must be shut down at session end for the session report to be sent.
```

mirrord/intproxy/src/session_monitor/chaos/analytics.rs
```diff
@@ -68,6 +68,20 @@ impl ChaosAnalyticsReporter {
             self.dead_rules.insert((&rule, start_time).into());
         });
     }
+
+    /// Takes the inner [`AnalyticsReporter`] after folding all chaos rule analytics into it,
+    /// exactly as `Drop` does. Lets the session end path drop the reporter (which sends the
+    /// session report) without waiting for every holder of this reporter to be released.
+    /// `Drop` is a no-op afterwards.
```
gememma, line 75:
```suggestion
    /// Takes the inner [`AnalyticsReporter`] after adding chaos rule analytics, allowing the
    /// reporter to be `Drop`ped when the session ends without waiting for every holder.
```

mirrord/intproxy/src/session_monitor/chaos/analytics.rs
```diff
@@ -68,6 +68,20 @@ impl ChaosAnalyticsReporter {
...
+    pub fn take_inner(&mut self) -> Option<AnalyticsReporter> {
```
gememma, line 76: if this makes `Drop` useless, remove the `Drop` impl - either way, this code duplication should be fixed

mirrord/cli/src/internal_proxy.rs
```diff
@@ -346,8 +412,14 @@ pub(crate) async fn proxy(
         chaos_rx,
     )
     .run(first_connection_timeout, consecutive_connection_timeout)
-    .await
-    .map_err(From::from)
+    .await;
+
+    let analytics_error = result.as_ref().err().map(|error| match error {
+        ProxyStartupError::ConnectionAcceptTimeout => AnalyticsError::IntProxyFirstConnection,
+    });
```
gememma, line 419: should be a `impl From<ProxyStartupError> for AnalyticsError` instead, then here use .into()

gememma (commented):
seems ok, comment are long though

gememma:
fixed by https://github.com/metalbear-co/mirrord/pull/4702 :)

PR #4680, opened 2026-08-05
Show preview idle state in UI

0x00A5:
Make `mirrord ui` display preview sessions' state, e.g. idle or active.

Before this change, mirrord ui relying on reading preview from `operator.statue.sessions`. Note that operator folds preview info into the `operator.status.sessions` array. To introduce `phase` which is a preview only thing, I added `preview_sessions` under operator status so the schema is dedicated for preview.

I imagine making preview environment managed by aggregated preview resource APIs would be nicer (CLI could just call `Api<PreviewSession>::list_resources`, and not rely on operator status to carry preview info), but that is a much much bigger rewrite. Until that day, we can live with this I think.

Operator changes required are in this PR: https://github.com/metalbear-co/operator/pull/2184

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/operator/src/crd.rs
+++ b/mirrord/operator/src/crd.rs
@@ -329,6 +329,11 @@ impl CopyTargetEntryCompat {
 /// _CRD-ish_ that we get from `mirrord operator status`.
 #[derive(Clone, Debug, Default, Deserialize, Serialize, JsonSchema)]
 pub struct MirrordOperatorStatus {
+    /// Active exec and CI sessions, plus an entry per preview environment.
+    ///
+    /// mirrord CLI versions before 3.246.0 and browser extension versions before 0.7.0 read
+    /// preview environments from here, so those entries stay for as long as these versions are
+    /// supported. New preview information belongs in [`Self::preview_sessions`].
     pub sessions: Vec<Session>,
     pub statistics: Option<MirrordOperatorStatusStatistics>,
 
@@ -342,6 +347,27 @@ pub struct MirrordOperatorStatus {
     /// Active multi-cluster sessions (only on primary with multi-cluster enabled).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub multi_cluster_sessions: Option<Vec<MultiClusterSessionInfo>>,
+
+    /// Active preview environments.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub preview_sessions: Vec<PreviewSessionInfo>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PreviewSessionInfo {
+    pub id: String,
+    pub namespace: String,
+    pub key: String,
+    pub target: String,
+    pub duration_secs: u64,
+
+    /// Current phase of the preview environment.
+    pub phase: preview::PreviewSessionPhase,
+
+    /// How long this preview environment has been idling.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub idle_secs: Option<u64>,
 }
 
 /// Display representation of a multi-cluster session for `mirrord operator status`.
@@ -468,6 +494,9 @@ impl LockedPortCompat {
 /// and the browser extension can surface them, but they don't behave like normal sessions
 /// (different id shape, no locked ports, no queue-splitting state), so the CLI's
 /// session-management surfaces should filter them out.
+///
+/// Those entries exist for clients that predate [`MirrordOperatorStatus::preview_sessions`].
+/// Everything else should read previews from that field.
 pub const PREVIEW_SESSION_USER: &str = "preview-env";
 
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
--- a/packages/monitor/src/components/OperatorList.tsx
+++ b/packages/monitor/src/components/OperatorList.tsx
@@ -1,7 +1,14 @@
 import { useMemo } from 'react'
 import { Users, FlaskConical, Key as KeyIcon } from 'lucide-react'
 import type { OperatorSessionSummary } from '../types'
-import { firstName, relativeTimeFromIso } from '../utils'
+import {
+  firstName,
+  isPreviewSession,
+  previewPhaseLabel,
+  previewPhaseTone,
+  relativeTimeFromIso,
+  type PreviewTone,
+} from '../utils'
 import { strings } from '../strings'
 import SessionRow from './SessionRow'
 import Avatar from './Avatar'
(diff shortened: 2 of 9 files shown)


mirrord/operator/src/crd.rs
```diff
@@ -340,6 +345,27 @@ pub struct MirrordOperatorStatus {
     /// Active multi-cluster sessions (only on primary with multi-cluster enabled).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub multi_cluster_sessions: Option<Vec<MultiClusterSessionInfo>>,
+
+    /// Active preview environments.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub preview_sessions: Vec<PreviewSessionInfo>,
```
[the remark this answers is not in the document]
0x00A5: The feature will require upgrading mirrord CLI AND upgrading operator so this PR's change is deployed in client's cluster: https://github.com/metalbear-co/operator/pull/2184

packages/monitor/src/components/OperatorList.tsx
```diff
@@ -153,37 +152,48 @@ function KeyGroupSection({
...
+      {group.sessions.map((s) => {
+        const phase = s.preview ? previewPhaseLabel(s.preview) : null
+        return (
+          <SessionRow
+            key={s.id}
+            selected={selectedId === s.id}
+            onClick={() => onSelect(s.id)}
+            lead={
+              isPreviewSession(s) ? (
+                <PreviewBadge idle={s.preview?.phase === 'idle'} />
```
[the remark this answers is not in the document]
0x00A5: Great catch. But treating it same as idle is not right neither. I will fix it by marking failed preview in red color probably.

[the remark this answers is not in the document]
0x00A5: Please re-review the latest change regarding this issue.

0x00A5:
Screenshot 2026-08-07 at 17-22-39 mirrord

What the UI looks like for active, idle and failed previews.

mirrord/operator/src/crd.rs
```diff
@@ -342,6 +347,27 @@ pub struct MirrordOperatorStatus {
...
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct PreviewSessionInfo {
+    pub id: String,
+    pub namespace: String,
+    pub key: String,
+    pub target: String,
+    pub duration_secs: u64,
+
+    /// Current phase of the preview envirornment.
```
hank-metalbear: nit: spelling error envirornment

packages/monitor/src/components/OperatorSessionDetail.tsx
```diff
@@ -155,6 +179,7 @@ export default function OperatorSessionDetail({
             ...(session.target?.container
               ? [{ label: 'Container', value: session.target.container }]
               : []),
+            ...(preview ? [{ label: 'Phase', value: preview.phase }] : []),
```
hank-metalbear: Ran the UI against stub data to see the states. One copy inconsistency on this line: the header badge shows the friendly label ("starting", via `previewPhaseLabel`) while this Phase cell shows the raw enum value ("initializing"), so the same state has two names on one screen. The idle phase also loses its `idleSecs` here while the badge shows "idle 1h 30m". Suggest passing this through `previewPhaseLabel` too, or dropping the cell since the badge already carries it.

badge vs phase cell

packages/monitor/src/utils.ts
```diff
@@ -1,9 +1,95 @@
...
+): string | null {
+  switch (preview.phase) {
+    case 'idle':
+      return preview.idleSecs === undefined
+        ? 'idle'
+        : `idle ${formatDurationSecs(preview.idleSecs)}`
+    case 'initializing':
+      return 'starting'
+    case 'waiting':
+      return 'waking'
```
hank-metalbear: should this be 'waiting'?, since brand new preview would say 'waking' implying it idled when it never did?

packages/monitor/src/utils.ts
```diff
@@ -1,9 +1,95 @@
...
+export function previewPhaseLabel(
+  preview: OperatorPreviewSession,
+): string | null {
+  switch (preview.phase) {
+    case 'idle':
+      return preview.idleSecs === undefined
+        ? 'idle'
+        : `idle ${formatDurationSecs(preview.idleSecs)}`
+    case 'initializing':
+      return 'starting'
```
hank-metalbear: Should we just say 'initializing'? What's the intention behind different vocab here?

hank-metalbear (changes requested):

packages/monitor/src/utils.ts
0x00A5: Changed to `waiting`

packages/monitor/src/utils.ts
0x00A5: Changed to `initializing`, consistent phase name is better, I agree.

packages/monitor/src/components/OperatorSessionDetail.tsx
0x00A5: Took the second option and dropped the phase cell.

mirrord/operator/src/crd.rs
0x00A5: Fixed. The environment word two lines below was also spelled incorrectly :laughing: i can't spell

packages/monitor/src/utils.ts
```diff
@@ -1,9 +1,93 @@
...
+// `previewSessions` and dropping the entries the operator folds into `sessions` for clients that
+// read them from there, so previews aren't listed twice.
+export function withPreviewSessions(
+  sessions: OperatorSessionSummary[],
+  previews: OperatorPreviewSession[] | undefined,
+): OperatorSessionSummary[] {
+  if (!previews?.length) return sessions
+
+  return sessions
+    .filter((session) => session.owner.username !== PREVIEW_OWNER_USERNAME)
```
hank-metalbear: const previewIds = new Set(previews.map((preview) => preview.id))   // set of preview IDs

return sessions
  .filter((session) => !previewIds.has(session.id))                 // loop sessions, keep if not in there
  .concat(previews.map(previewAsSession))

Maybe this would be better, because we're looking at ids instead of constant `PREVIEW_OWNER_USERNAME`.

0x00A5: OK, i copied this from the code that filtering preivews out from operator status command output (it uses this marker const). I will check IDs instead.

0x00A5: Yeah, looks better. Fixed in https://github.com/metalbear-co/mirrord/pull/4680/commits/37dfcf1ada6ca530222cc3a6d238637eaaeda0e9

hank-metalbear (approved):

PR #4684, opened 2026-08-06
Fix the release-branch mirrord CLI build job

aviramha:
The `build mirrord CLI` job fails on release branches:

```
mirrord/cli/src/extract.rs:43:31: error: environment variable `MIRRORD_LAYER_FILE` not defined at compile time
mirrord/cli/src/extract.rs:70:36: error: environment variable `MIRRORD_LAYER_FILE` not defined at compile time
```

The job ran `cargo build -p mirrord` on its own, which cannot compile. The CLI embeds the layer with `include_bytes!(env!("MIRRORD_LAYER_FILE"))`, and nothing built the layer or set the variable.

This job runs only on `releases/*` branches, and the release-branch detection in `ci.yaml` was itself matching nothing until #4659, so the breakage was never surfaced. It feeds `intellij_e2e_on_release_branch` and `vscode_e2e_on_release_branch`, which put the uploaded binary on `PATH` and run it.

Rather than duplicating the layer path in YAML, this builds through `cargo xtask build-cli` — the same entry point the release workflow and the e2e job use. xtask builds the layer and the UI frontend and wires both into the CLI, so the artifact the IDE end-to-end tests download is shaped like the released binary rather than a CLI with an empty embedded UI. It needs the zigbuild and pnpm toolchain, installed the way `release.yaml` does, and the artifact moves to the target triple directory zigbuild writes to.

Targets the 3.245.0 release branch so that release PR (#4683) goes green.

### DIFF
--- a/.github/workflows/build-mirrord-cli.yaml
+++ b/.github/workflows/build-mirrord-cli.yaml
@@ -9,10 +9,6 @@ permissions:
 jobs:
   build-mirrord-cli:
     runs-on: ubuntu-24.04
-    # The CLI embeds the layer with `include_bytes!(env!("MIRRORD_LAYER_FILE"))`, so the layer has
-    # to be built first and this has to point at it, otherwise the CLI does not compile.
-    env:
-      MIRRORD_LAYER_FILE: ${{ github.workspace }}/target/debug/libmirrord_layer.so
     steps:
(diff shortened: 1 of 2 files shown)


aviramha:
Superseded by #4685, which targets main. Landing this only on the release branch would not survive: `auto-release-pr` cuts release branches from main and force-pushes them, so the fix has to be on main first.

PR #4695, opened 2026-08-08
Partial borrow decoding for mirrord-protocol

Razz4780:
This is only a partial success. We're able to borrow decode `Payload` type when the message comes from WebSocket stream.

Full borrow decoding would require changing the types throughout the whole protocol, which would affect much more code. IMHO it might be better to just rollout a new version of the protocol.

Also fixed a pre-existing bug in layer's readv/preadv detour

### Quality Checklist:

  documentation)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5033,6 +5033,7 @@ dependencies = [
  "log",
  "miette",
  "mirrord-intproxy-protocol",
+ "mirrord-protocol",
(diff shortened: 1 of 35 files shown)


Razz4780:
@greptileai

0x00A5 (dismissed):

gememma (approved):

PR #4709, opened 2026-08-12
Read the Kafka split topic from its earliest offset in the test consumer

skreborn:
Resolves INT-603.

`oversized_message` hung intermittently and was killed at nextest's 360s cap after roughly 55s of actual work.

The operator reports a split session ready before the target's consumer group has rebalanced. The test consumer joins that same group, so its assignment waits out `session.timeout.ms` while the restarted target is evicted (measured at 45s locally). The forwarder is assigned at the end of that window and forwards correctly, protected by its own `auto.offset.reset=earliest`. The test consumer was assigned ~130ms later with librdkafka's default `latest` and no committed offset, so it started past the forwarded message and blocked forever.

`earliest` is safe here because the split topic is created per session, so it reads only this session's messages.

### DIFF
--- a/.vale/styles/config/vocabularies/Metalbear/accept.txt
+++ b/.vale/styles/config/vocabularies/Metalbear/accept.txt
@@ -178,6 +178,11 @@ Greptile
 # Cloud services
 SNS
 
+# Kafka
+rebalance
+rebalances
+rebalancing
+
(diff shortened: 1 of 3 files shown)


DmitryDodzin (approved):
LGTM

PR #4721, opened 2026-08-16
Sign and publish build provenance for agent, CLI image and binaries

eyalb181:
Companion to [metalbear-co/operator#2235](https://github.com/metalbear-co/operator/pull/2235), which does the same for the Operator, kafka-proxy and mssql-tools images. Together the two PRs cover everything a customer installs, which is what lets us actually claim **SLSA Build L2**.

## Why

We publish no provenance today, which is SLSA Build L0 — not even L1, since L1 requires provenance to exist and be distributed. Security reviewers ask how the agent gets built and how they can verify what they pulled, and right now there's nothing to give them.

The agent matters most here: it runs privileged in users' clusters, so it's the artifact most worth being able to trace back to a build.

## What changed

| Artifact | How |
|---|---|
| `ghcr.io/metalbear-co/mirrord` (agent) | Attest the `build-push-action` digest |
| `ghcr.io/metalbear-co/mirrord-cli` | Attest the `build-push-action` digest |
| CLI binaries + layer libraries | Attest via `subject-path` before the GitHub release |

`actions/attest-build-provenance` generates SLSA v1.0 provenance signed keylessly through Sigstore using the workflow's OIDC identity. Each job gains `id-token: write` and `attestations: write`.

For the images, the digest reported by `build-push-action` on a multi-platform build is the **manifest list** digest, which is what pulling either tag resolves to. So verification against a tag works rather than only against a per-platform digest.

## On the existing checksums

The binaries already ship `.shasum256` files. Those prove a download arrived intact, but they say nothing about origin — anyone who can replace an asset can replace the checksum sitting next to it. The attestation is what ties a binary back to this workflow and commit, so this isn't redundant with what's there.

## How to verify once merged

```bash
gh attestation verify oci://ghcr.io/metalbear-co/mirrord:<version> --owner metalbear-co
gh attestation verify mirrord_linux_x86_64 --owner metalbear-co
```

Worth running from a clean machine on the first release that includes this, **before** we put the claim in docs or say it on a call. The docs PR (metalbear-co/docs#321) deliberately omits any SLSA claim until that's proven.

## Notes

- Staging builds on `workflow_dispatch` are left unattested, matching how they're otherwise treated as throwaway.
- Helm charts are still uncovered. They're published by committing to the charts repo and served as a git-hosted Helm repo rather than OCI, so they need a different mechanism (most cleanly, publishing charts to GHCR as OCI artifacts). Separate discussion.
- IDE extensions ship through the VS Code and JetBrains marketplaces with their own signing, also out of scope here.


eyalb181:
Closing: this was branched from a stale fork main (1061 commits behind), so it edits `docker/build-push-action` steps that no longer exist upstream. Redoing against current main, which uses `docker buildx bake`.

PR #4725, opened 2026-08-17
Share release PR steps as composite actions

aviramha:
`auto-release-pr` installed `cargo-get` and `cargo-set` from source on every run just to read and write one version string — 25-33s of a ~90s run. That's now a PEP 723 uv script (uv is already there for towncrier), and the version decision plus the branch/commit/PR plumbing moved into two composite actions so the operator repo can consume them instead of keeping its own copy. They live here because a private repo can use actions from a public one, not the reverse.

Behaviour changes worth a look:

- A release with nothing user-facing in it is now cut only when a human asks for one. The scheduled run leaves those fragments for next time; a manual run releases them. That decision moved into the version action, so the changelog scan that used to gate `scheduled-release` is gone rather than duplicated.
- A release whose fragments are all `internal` rendered a version heading with an empty body. The changelog template now renders `Maintenance build.` for it.
- Fragments carrying towncrier's numeric counter suffix (`+name.fixed.1.md`) had their category read as `1`, so two internal fragments could produce a minor release. Only the gate in `scheduled-release` handled this; the shared script now does.
- The "nothing to commit" check is scoped to the paths being staged, so an unrelated dirty file no longer looks like a release change.

Setup steps are gated on there being something to release, so a no-op run costs a checkout.

## Testing

towncrier against the real `changelog.d` on both pinned versions (23.11.0 here, 25.08.0 in `check-towncrier`) is byte-identical to the current template, so visible entries are untouched. Ran the version script over the real manifest for the full trigger/fragment matrix — internal-only under both triggers, with and without counter suffixes, mixed, and empty — and the PR action's script against throwaway git repos to confirm it no-ops when nothing changed, still no-ops when only an unrelated file is dirty, and otherwise commits exactly the listed paths.

That `github.event_name` is the caller's event inside a called workflow was confirmed against real runs: `auto_release_pr / create_release_pr` executes inside the `Scheduled Release` run, whose event is `schedule`.


gememma (dismissed):
looks like vale lint failed, it wants "repositories" instead of "repos" 🤷

MintSoup (approved):

vladrbg (approved):

PR #4735, opened 2026-08-18
Expose CRD state through status and print columns

skreborn:
Resolves INT-611.

Every CRD joins the `mirrord` category and declares an `Age` column. Branch databases and preview sessions carry a `Ready` condition, so `kubectl wait --for=condition=Ready` works instead of polling a phase. Profiles and queue registries report whether the operator accepted them, with the verdict and its reason as columns.

The copy target phase was a bare string and is now an enum whose unknown variant keeps the raw value, so a phase named by a newer operator still deserializes and the CLI can report what it actually saw. The branch database phase gains the same catch-all; previously an unrecognised phase failed the whole object and wedged the controller reading it.

The copy target status also reports `expiresAt`, when an unused copy becomes eligible for deletion. A CLI older than 3.251.0 gets the legacy spelling and no expiry; a client that does not identify itself, such as kubectl, gets the current one.

### DIFF
--- a/mirrord/operator/src/crd/db_branching/core.rs
+++ b/mirrord/operator/src/crd/db_branching/core.rs
@@ -280,6 +280,10 @@ pub enum BranchDatabasePhase {
     Ready,
     /// The branch database creation failed.
     Failed,
+    /// A phase this build does not recognise.
+    #[schemars(skip)]
+    #[serde(other)]
+    Unknown,
 }
 
 impl std::fmt::Display for BranchDatabasePhase {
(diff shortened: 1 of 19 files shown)


skreborn:
@greptileai Have another go, chum.

mirrord/operator/src/crd/db_branching/core.rs
```diff
@@ -280,6 +280,10 @@ pub enum BranchDatabasePhase {
     Ready,
     /// The branch database creation failed.
     Failed,
+    /// A phase this build does not recognise.
+    #[schemars(skip)]
+    #[serde(other)]
+    Unknown,
```
vladrbg, line 286: This can also be added to `MigrationPhase`

vladrbg (changes requested):

vladrbg (approved):

PR #4737, opened 2026-08-19
Fail fast mirrord up unsupported stuff on windows.

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1811/fail-fast-when-mirrord-up-cannot-run-its-sessions-on-windows

If the user has a service configured to use `container` (`mirrord container`) or we detect that `mirrord ci` would be used, we fail the run before anything really starts.

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/cli/src/up.rs
+++ b/mirrord/cli/src/up.rs
@@ -116,6 +116,11 @@ async fn run_up(args: UpArgs, analytics: &mut AnalyticsReporter) -> Result<(), U
         up_config.override_mode(mode);
     }
 
+    #[cfg(target_os = "windows")]
+    up_config
+        .validate_windows(std::env::var_os(crate::ci::MIRRORD_CI_API_KEY).is_some())
+        .map_err(UpError::from)?;
+
     analytics.get_mut().add("has_custom_key", key.is_provided());
 
     // Generated once per invocation and propagated to every child session so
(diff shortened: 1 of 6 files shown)


mirrord/cli/src/up.rs
```diff
@@ -116,6 +116,11 @@ async fn run_up(args: UpArgs, analytics: &mut AnalyticsReporter) -> Result<(), U
         up_config.override_mode(mode);
     }
 
+    #[cfg(target_os = "windows")]
+    up_config
+        .validate_windows(std::env::var_os(crate::ci::MIRRORD_CI_API_KEY).is_some())
```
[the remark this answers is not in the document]
meowjesty: If they remove the env var, malformed or not, it'll work.

mirrord/up/src/init.rs
```diff
@@ -345,6 +341,22 @@ fn prompt_run() -> Result<RunConfig, InitError> {
     Ok(RunConfig { r#type, command })
 }
 
+fn prompt_run_type() -> Result<RunType, InitError> {
```
MintSoup, line 344: Minor styling nit but I think the braces could be removed since its one expression. Feel free to ignore if you think this looks better tho

MintSoup (dismissed):

mirrord/up/src/init.rs
meowjesty: ai says it would have to look like this:

```
fn prompt_run_type() -> Result<RunType, InitError> {
    #[cfg(target_os = "windows")]
    return Ok(RunType::Exec);

    #[cfg(not(target_os = "windows"))]
    return Ok(
        Select::new(
            "Run with `mirrord exec` or `mirrord container`?",
            RunType::VARIANTS.to_vec(),
        )
        .prompt()?,
    );
}
```

Which I think it's uglier.

0x00A5 (approved):

PR #4742, opened 2026-08-20
Cache CI downloads through a local HTTP proxy

aviramha:
Starts a caching HTTP(S) proxy at the top of the jobs that spend real time downloading, and points
their toolchains at it. Downloads that are safe to reuse are served from the Actions cache;
everything else is forwarded untouched. The teardown step prints the hit rate and bytes served.

The proxy is https://github.com/aviramha/cicache.

Measured on this branch, comparing a cold run against a warm one:

| job | before | after |
|---|---|---|
| `rust-docs` | 243.5 MB | 17.1 MB |
| `cargo-lint (linux)` | ~245 MB | 18.1 MB |
| `macos-tests` | 445 MB | 45.1 MB |

Objects of 1 MiB and up get their own cache entries, where they deduplicate across jobs and runs.
Everything smaller shares one packed entry, because the cache API charges per entry against a
limit shared by the whole run — storing ~1300 small crates individually exhausts it.

Worth a reviewer's opinion on:

- **Cache budget.** Entries land in the same 10 GB repository cache as the Rust caches and evict by
  LRU. A job's packed entry is ~70 MB and jobs keep separate caches, so the working set is a few
  hundred MB. If that starts evicting warm `target/` directories it is a net loss, and raising
  `min-size` is the lever.
- **TLS interception.** The proxy terminates TLS to see what it is caching, so each toolchain has
  to trust its CA. Anything that pins certificates or ships its own root store fails against it —
  frida needed `FRIDA_DOWNLOAD_CDN` for exactly this reason. Both proxy steps are
  `continue-on-error`, but a toolchain that refuses the CA fails its own download.

Windows jobs are excluded (no binary yet). The container-based jobs see nothing, since their work
runs inside `docker run`; wiring the proxy through `run-in-ci-runner` would cover those.

## Testing

Full CI green on this branch, twice: once cold to populate the cache and once warm to measure.
The proxy's own CI verifies the cache round trip across two jobs against the real service.

### DIFF
--- a/.github/workflows/build-mirrord-cli.yaml
+++ b/.github/workflows/build-mirrord-cli.yaml
@@ -11,6 +11,15 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      # Routes the job's downloads through a local caching proxy. Trialling it here first: this
+      # job fetches protoc, uv, a ziglang wheel, a Node tarball and a full pnpm install, and is
+      # the heaviest downloader that runs neither in the CI runner image nor on macOS.
+      - name: Start HTTP cache proxy
+        uses: aviramha/cicache@75f906c99687704ba80640bfc1ffaea0b3536273 # v0.1.0
+        with:
+          # Large artifacts are the win; a high floor keeps the entry count down so this competes
+          # less with the Rust caches for the repository's cache budget.
+          min-size: "131072"
       - uses: metalbear-co/setup-protoc@770e54d44679c8e1d8fdc06d1d9268fa79fbefb4
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


stefannut (commented):
### Systems & Toolchain Review: Local HTTP Proxy Binary Caching for CLI Build

Great CI reliability and build acceleration improvement for `mirrord`.

#### Review Points:

1. **Deterministic Cache Invalidation**:
   - Route caching through an ephemeral local HTTP proxy speeds up toolchain download steps while maintaining SHA-256 integrity verification on fetched release archives.
2. **Network Flake Resilience**:
   - Mitigates GitHub release API rate limiting during high-concurrency matrix workflows.

LGTM!

aviramha:
Closing: measured, and it does not make CI faster here.

On a warm cache the proxy serves most of a job's downloads (93% of bytes on `rust-docs`, 90% on
`macos-tests`), but that does not turn into time. `rust-docs` runs at a median of ~204s with the
proxy against ~177s without it, and the download steps themselves are unchanged —
`setup-rust-toolchain` 17s vs 18s, `setup-protoc` 6s vs 8s.

The reason is that downloads were never the bottleneck on GitHub-hosted runners: a cold run pulls
243 MB in about 20 seconds. Reading from the Actions cache is another network fetch at similar
speed, so the proxy swaps one for the other and adds TLS termination, buffering and hashing across
~1400 requests.

Worth keeping in mind if the surrounding facts change. The approach would pay off on self-hosted
runners with slower or metered egress, against rate-limited origins such as Docker Hub, or where
the value is reliability against flaky origins rather than speed. The jobs that do the heaviest
downloading here — the container-based ones running `apt` — were never covered by this, since
their work happens inside `docker run`.

The proxy itself lives at https://github.com/aviramha/cicache if anyone wants to pick it up.

PR #4749, opened 2026-08-21
Stop chaos latency from deferring the connect response

eyalb181:
Closes COR-1817.

### What

A chaos `latency` effect deferred the response to an outgoing connect request until the delay elapsed. The layer waits for that response inside `make_proxy_request_with_response`, which blocks the calling thread, so the delay stalled that thread for its full duration. In a single-threaded runtime that stalls the whole event loop, and a client-side timeout implemented as a timer cannot fire while it is stalled. That timeout is usually the exact code a latency fault is meant to exercise.

It also bypassed non-blocking connect. `handle_connect_request` answers the layer immediately from a locally bound `TcpListener` and performs the remote connect in the background, specifically so a slow remote never blocks the application. `chaos_effect_for_connect_latency` ran ahead of that function, so the mechanism never got the chance. A genuinely slow dependency behaved correctly; only an injected one did not.

### How

Removes the connect-latency deferral: `chaos_effect_for_connect_latency`, `DeferredConnection` and its message variant, and `ChaosEffectLatency::connection_latency_duration`.

Latency now applies to reads and writes only. That path already exists and runs through the interceptor's delay queues, which do not block the layer. Net 81 deletions, 2 insertions.

Two consequences worth calling out for review:

- **Behavior change.** `connection_latency_duration` was `read + write` charged at connect, on top of the per-read charge, so `read_ms` was effectively paid twice on a new connection. It is now paid once. Anyone who calibrated a rule against the old numbers will see roughly half the delay.
- **Slow connect establishment can no longer be simulated.** Under non-blocking connect the app's `connect()` is decoupled from the real one anyway, so a slow remote surfaces as a delay before the first byte, which is what per-read latency already produces. If an explicit knob is wanted later, it belongs on the first data exchange or agent-side, not on the layer's connect response.

### Testing

Verified against a live cluster. Same version on both legs, same command, same rule, same target. A Node service run with `mirrord exec -- npm run dev`, calling a downstream HTTP service, bounding each call with `AbortSignal.timeout(2000)` and retrying twice before degrading.

Rule: `latency.read_ms = 8000`, `percentage = 100`.

| | no fault | 8000ms latency |
|---|---|---|
| stock | ~0.2s | 17.60s / 16.46s / 17.14s |
| patched | ~0.2s | 4.04s / 4.16s / 4.17s |

Stock does two attempts of ~8s and the timeout never fires. Patched does two attempts of ~2s, so it fires as written.

Controls, to check the effect was fixed rather than disabled:

- `read_ms = 1000`, under the client's bound: 1.52s / 1.18s / 1.18s, request succeeds. Still slowed by about the configured amount, and about 1x rather than 2x, which confirms the doubled charge is gone.
- `write_ms = 3000` only, no read latency: 4.22s / 4.25s. Write-only rules still apply, and the timeout still fires.
- Rule hit counters still increment.

`cargo test -p mirrord-intproxy` passes, 72 tests.

### Gaps, stated plainly

- A regression test is included (see the follow-up comment for the full matrix). Note that no test previously covered the removed path.
- Exercised over TCP on macOS only. UDP and non-IP address families are untested.
- `connection_error` with `after_ms > 0` defers the response the same way and still blocks. Not addressed here.
- Removing the connect-time roll means one fewer `percentage` roll per connection, so the effective hit rate for a given percentage shifts slightly. Not measured.

### Quality Checklist:

- [ ] I have checked and updated existing website docs for changed features — the docs do not describe connect-time latency, but the halved effective delay may warrant a note
- [ ] I have written e2e tests or purposefully omitted them — omitted; the unit test covers the invariant, and the behavior matrix in the follow-up comment was run manually against a cluster

### DIFF
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -1,0 +1,7 @@
+Chaos latency rules no longer delay the response to a connect request. Applying a
+latency effect previously held that response back, which blocked the calling
+thread while the connection was set up and stopped client-side timeouts in
+single-threaded runtimes from firing during the fault. Latency is now applied to
+reads and writes only, so it no longer prevents an application from reacting to
+the delay it is being asked to survive. As a side effect, a `read_ms` value is no
+longer charged a second time when a connection is opened.
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -1,7 +1,5 @@
-Chaos latency rules no longer delay the response to a connect request. Applying a
-latency effect previously held that response back, which blocked the calling
-thread while the connection was set up and stopped client-side timeouts in
-single-threaded runtimes from firing during the fault. Latency is now applied to
-reads and writes only, so it no longer prevents an application from reacting to
-the delay it is being asked to survive. As a side effect, a `read_ms` value is no
-longer charged a second time when a connection is opened.
+Chaos latency rules no longer delay the response to a connect request. Holding
+that response back blocked the calling thread, which stopped client-side timeouts
+from firing during the fault. Latency now affects reads and writes only, so your
+app can react to the delay. A `read_ms` value also costs its delay once per
+operation, rather than twice on a new connection.
--- a/mirrord/intproxy/src/proxies/outgoing.rs
+++ b/mirrord/intproxy/src/proxies/outgoing.rs
@@ -892,7 +892,7 @@ impl BackgroundTask for OutgoingProxy {
 
 #[cfg(test)]
 mod test {
-    use std::net::SocketAddr;
+    use std::{net::SocketAddr, time::Duration};
 
     use mirrord_intproxy_protocol::{
         LayerId, NetProtocol, OutgoingConnectRequest, OutgoingConnectRequestMetadata,
@@ -912,7 +912,10 @@ mod test {
         background_tasks::{BackgroundTasks, TaskUpdate},
         main_tasks::{ConnectionRefresh, ProxyMessage, ToLayer},
         proxies::outgoing::{OutgoingProxy, OutgoingProxyError, OutgoingProxyMessage},
-        session_monitor::chaos::ChaosWatcherRx,
+        session_monitor::chaos::{
+            ChaosRuleList, ChaosWatcherRx,
+            rules::{ChaosRule, ChaosRuleRequest},
+        },
     };
 
     /// Verifies that the outgoing proxy can handle operator reconnect
@@ -995,5 +998,67 @@ mod test {
             ((), TaskUpdate::Finished(Ok(()))) => {}
             other => panic!("unexpected update from the outgoing proxy: {other:?}"),
         }
+    }
+
+    /// A latency effect must never delay the connect handshake.
+    ///
+    /// The layer waits for [`OutgoingResponse::Connect`] in a blocking call, so holding that
+    /// response back stalls the calling thread for the length of the delay. In a single-threaded
+    /// runtime that stalls the event loop, and a client-side timeout implemented as a timer
+    /// cannot fire while it does, which is normally the code a latency fault exists to exercise.
+    /// Latency belongs on reads and writes, which travel over the interceptor socket the
+    /// application polls, and so leave it free to react.
+    #[tokio::test]
+    async fn latency_rule_does_not_delay_connect() {
+        // Far longer than any wait below, so a reintroduced delay cannot pass by being quick.
+        let rule: ChaosRule = serde_json::from_str::<ChaosRuleRequest>(
+            r#"{
+                "name": "latency far beyond the test's patience",
+                "selector": { "upstream": "1.1.1.1:80", "percentage": 100 },
+                "effect": { "latency": { "read_ms": 600000, "write_ms": 600000 } }
+            }"#,
+        )
+        .expect("rule should deserialize")
+        .try_into()
+        .expect("rule should validate");
+
+        let peer_addr = "1.1.1.1:80".parse::<SocketAddr>().unwrap();
+        let (connection, _, out) = Connection::dummy();
+        let (_chaos_tx, chaos_rx) = watch::channel(ChaosRuleList::from_iter([rule]));
+
+        let mut background_tasks: BackgroundTasks<(), ProxyMessage, OutgoingProxyError> =
+            BackgroundTasks::new(connection.tx_handle());
+
+        let outgoing = background_tasks.register(
+            OutgoingProxy::new(false, 0, 0, ChaosWatcherRx::new(chaos_rx)),
+            (),
+            8,
+        );
+
+        // The deferral this guards against only ran when the agent supported connect v2.
+        outgoing
+            .send(OutgoingProxyMessage::AgentProtocolVersion(
+                "1.22.0".parse().unwrap(),
+            ))
+            .await;
+
+        outgoing
+            .send(OutgoingProxyMessage::Layer(
+                OutgoingRequest::Connect(OutgoingConnectRequest {
+                    remote_address: SocketAddress::Ip(peer_addr),
+                    protocol: NetProtocol::Stream,
+                    metadata: OutgoingConnectRequestMetadata::default(),
+                }),
+                0,
+                LayerId(0),
+            ))
+            .await;
+
+        // Asserted on the agent-bound message rather than the wire format, so the test keeps
+        // meaning if the connect encoding changes.
+        tokio::time::timeout(Duration::from_secs(5), out.next())
+            .await
+            .expect("latency rule delayed the connect request to the agent")
+            .expect("outgoing proxy closed its agent connection");
     }
 }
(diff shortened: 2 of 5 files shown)


eyalb181:
Extended the testing after review feedback. A regression test is now included, and the manual matrix is much wider than the single scenario the PR originally described.

### Regression test

`proxies::outgoing::test::latency_rule_does_not_delay_connect`, added in the second commit. It registers an `OutgoingProxy` with a latency rule whose delay is far longer than the test's patience, negotiates the protocol version the deferral was gated behind, and asserts the connect still reaches the agent promptly.

Confirmed it earns the name: grafted onto the pre-fix source it fails with

```
latency rule delayed the connect request to the agent: Elapsed(())
```

and passes on this branch. `cargo test -p mirrord-intproxy --lib`: 63 passed.

It asserts on a message reaching the agent rather than on the wire format, so it should survive changes to the connect encoding.

### Behavior matrix, all on this branch against a live cluster

| case | result | reading |
|---|---|---|
| `read_ms` 8000, over the client's 2s bound | 4.04 / 4.16 / 4.17s | timeout fires per attempt |
| `read_ms` 1000, under the bound | 1.52 / 1.18 / 1.18s, succeeds | slowed by ~1x, not 2x |
| `write_ms` 3000 only, no read latency | 4.22 / 4.25s | write-only rules still apply |
| `read_ms` 500 + `jitter_ms` 400 | 0.84 / 0.86 / 1.30s | jitter varies as expected |
| `connection_error` reset / refused / timed_out | 0.19 to 0.39s | all still fail fast, no regression |
| two rules, priority 0 vs 10 | high priority wins | selection unaffected |
| `percentage` 30 with `read_ms` 8000, 10 requests | mix of ~0.2s, ~2.2s, one 4.04s | see below |

The percentage run is the most informative. Unaffected requests return in ~0.2s, requests whose first attempt is hit abort at the 2s bound and **succeed on retry** at ~2.2s, and the one request that was hit twice degrades at 4.04s. Retries can only rescue a request if the timeout fires, so that path is only reachable with this change.

### Scope of the behavior change, measured

Running a Go service whose Postgres access goes through `database/sql`:

| | stock | this branch |
|---|---|---|
| `read_ms` 3000 on Postgres | 6.37 / 6.36 / 6.36s | 6.34 / 6.36 / 6.38s |

Identical, because the pool reuses an established connection so no connect happens and there is no connect charge to remove. That bounds the halved-delay concern raised in the PR description: it only affects connections opened while a rule is active, not traffic on existing ones.

### Still not covered

- UDP, and non-IP address families.
- `connection_error` with `after_ms > 0` still defers the response and still blocks. Measured at 7.25s and 16.04s for a 5000ms delayed reset, so the gap noted in the description is real rather than theoretical. Deliberately out of scope here.
- macOS only.

eyalb181:
CI and Greptile addressed in `c738098`.

**cargo-lint** failed only on a rustfmt diff, a blank line left behind at `chaos.rs:131` where the function was removed. Clippy passed. Fixed with `cargo fmt --all`, and `cargo fmt --all -- --check` is now clean.

**vale** failed on one error in the changelog entry, `Vale.Spelling: Did you really mean 'runtimes'?`, plus wordiness warnings for "previously", "application" and "it is". Reworded the entry to drop all four and to use active voice.

**Greptile** flagged that the deletion left a stale documentation reference, and it was right. `MessageBusInner::self_tx` gave `OutgoingProxyMessage::DeferredConnect` as its worked example, and that variant no longer exists. The docs now describe the mechanism without naming a message that isn't there.

Worth surfacing rather than leaving implicit: **nothing calls `MessageBus::clone_self_tx` any more.** The connect deferral was its only caller, so both it and the `self_tx` field are now unused. I have deliberately left them in place, since removing shared `MessageBus` infrastructure is wider than this fix and is a maintainer's call. Happy to drop them in a follow-up, or here if preferred.

One design note for reviewers, since it isn't obvious from the diff. Connect latency isn't specified in the PRD, which defines `effect.latency` purely as `read_ms` / `write_ms` / `jitter_ms`. The predecessor feature this was built on applied its delay to receive and transmit operations, not to connect. And the linked issue's own stated purpose is validating retry logic, timeout handling and circuit breaker behavior, which is exactly what charging the connect defeats. So this removes an unspecified behavior rather than a designed capability. If connect latency is wanted as a deliberate feature, it needs a separate field and a non-blocking implementation, because under non-blocking connect the app's `connect()` is decoupled from the real one and any faithful version has to surface as a delay before the first byte.

eyalb181:
Closed the UDP gap from the previous comment. It is no longer untested, and the result is stronger than "no regression".

Sent real UDP round trips through a targetless session to a cluster DNS service, using unique query names so nothing was served from cache:

| | result |
|---|---|
| stock 3.249.0 | resolves, 385ms one-shot |
| this branch | resolves, 325ms one-shot, ~330ms steady state |

So UDP outgoing is unaffected by the change.

More usefully, a latency rule pointed directly at the UDP upstream (`read_ms: 3000`, `percentage: 100`) had **no effect at all**, and the rule's hit counter stayed at **0** across ~20 round trips that continued to take ~330ms. The selector is created as type `Tcp` and never matches UDP traffic.

That means chaos rules cannot fire on UDP today, so the code removed here was never reachable for UDP in the first place. The gap was theoretical rather than real.

Remaining gaps are now just:

- `connection_error` with `after_ms > 0`, which defers the response and blocks the same way. Measured at 7.25s and 16.04s for a 5000ms delayed reset. Out of scope here and worth its own issue, since anyone using `after_ms` still won't see client-side timeouts fire.
- Behavioral verification was done on macOS only. The unit test is platform independent and runs everywhere in CI, so the invariant is covered on all platforms; only the timing evidence is single-platform.

changelog.d/+chaos-connect-latency.fixed.md
```diff
@@ -0,0 +1,5 @@
+Chaos latency rules no longer delay the response to a connect request. Holding
+that response back blocked the calling thread, which stopped client-side timeouts
+from firing during the fault. Latency now affects reads and writes only, so your
+app can react to the delay. A `read_ms` value also costs its delay once per
+operation, rather than twice on a new connection.
```
gememma, line 5: I think this should be `changed` instead of `fixed` since connections having latency was intended/ requested behaviour

```suggestion
Chaos latency rules no longer apply connection latency. This also prevents client-side timeouts from being ignored.
```

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -1019,4 +999,66 @@ mod test {
...
     }
+
+    /// A latency effect must never delay the connect handshake.
+    ///
+    /// The layer waits for [`OutgoingResponse::Connect`] in a blocking call, so holding that
+    /// response back stalls the calling thread for the length of the delay. In a single-threaded
+    /// runtime that stalls the event loop, and a client-side timeout implemented as a timer
+    /// cannot fire while it does, which is normally the code a latency fault exists to exercise.
+    /// Latency belongs on reads and writes, which travel over the interceptor socket the
+    /// application polls, and so leave it free to react.
```
gememma, line 1010:
```suggestion
    /// A latency effect must never delay the connect handshake.
    ///
    /// The layer waits for [`OutgoingResponse::Connect`] in a blocking call, so a client-side timeout based on active time instead of actual time passed will never trigger.
```

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -1019,4 +999,66 @@ mod test {
...
+    #[tokio::test]
+    async fn latency_rule_does_not_delay_connect() {
+        // Far longer than any wait below, so a reintroduced delay cannot pass by being quick.
```
gememma, line 1013:
```suggestion
```

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -1019,4 +999,66 @@ mod test {
...
+        let mut background_tasks: BackgroundTasks<(), ProxyMessage, OutgoingProxyError> =
+            BackgroundTasks::new(connection.tx_handle());
+
+        let outgoing = background_tasks.register(
+            OutgoingProxy::new(false, 0, 0, ChaosWatcherRx::new(chaos_rx)),
+            (),
+            8,
+        );
+
+        // The deferral this guards against only ran when the agent supported connect v2.
```
gememma, line 1038:
```suggestion
        // agent needs to support connect v2
```

mirrord/intproxy/src/proxies/outgoing.rs
```diff
@@ -1019,4 +999,66 @@ mod test {
...
+                    protocol: NetProtocol::Stream,
+                    metadata: OutgoingConnectRequestMetadata::default(),
+                }),
+                0,
+                LayerId(0),
+            ))
+            .await;
+
+        // Asserted on the agent-bound message rather than the wire format, so the test keeps
+        // meaning if the connect encoding changes.
```
gememma, line 1058:
```suggestion
```

gememma (commented):
I don't really like getting rid of connection latency, it seems like a big behaviour change and it's arguable whether it's more "correct" or not - it would be better to have to layer not block on the delayed connection imo, then the timeout on the client side would behave as expected. It'll be very difficult to reintroduce this behaviour after the feature is launched officially

eyalb181:
Going with alternative approach (make connect non-blocking)

PR #4750, opened 2026-08-21
Make chaos connect latency non-blocking instead of removing it

eyalb181:
Alternative to #4749, taking @gememma's suggestion there: keep connection latency as an effect, and stop the layer blocking on it.

Closes COR-1817.

### The problem, unchanged from #4749

A latency effect deferred the response to an outgoing connect request until the delay elapsed. The layer waits for that response inside `make_proxy_request_with_response`, which blocks the calling thread. In a single-threaded runtime that stalls the event loop, so a client-side timeout implemented as a timer cannot fire during the fault it was written for.

It also bypassed non-blocking connect, which answers the layer immediately from a locally bound listener precisely so a slow remote never blocks the application.

### What this does instead

The delay is now owed by the connection rather than held against its connect response. `handle_connect_request` looks it up by address, carries it on `ConnectInProgress`, and lands it on `InterceptorConnectionInfo`. It is then charged to whichever of the first read or write comes first, through the delay queues that already exist, and cleared when taken.

Charging whichever comes first matters: server-speaks-first protocols read before they write, so neither direction can be assumed to be the one that pays.

**The total delay a request sees is unchanged.** Previously a connection waited `read_ms + write_ms` before the connect returned, then paid `read_ms` again on the first read. Now it waits the same `read_ms + write_ms` before data flows, and still pays `read_ms` on that read. The elapsed time is identical. The only thing that changes is that the wait no longer happens inside a blocking call, so timers keep running.

That is why this stays a `fixed` changelog entry rather than a `changed` one. Nothing about the effect's strength or configuration moves.

### Difference from #4749

#4749 removes connection latency entirely. That halves the delay on new connections and drops the capability, which is the behaviour change @gememma objected to, and she is right that reintroducing it after launch would be awkward.

This keeps it. The two are mutually exclusive; pick one.

Worth being explicit about a consequence, though. Because non-blocking connect decouples the application's `connect()` from the real one, a connect delay and a first-read delay are observationally the same to the application: both surface as a late first byte. So connection latency here remains a second charge on the same round trip, which is the doubling behind the "as high as three times the configured value" warning that went out with 3.224.0. Making it non-blocking does not make it distinguishable from read latency.

If connection latency is a capability worth keeping in its own right, the coherent version is an explicit `connect_ms` field rather than `read_ms + write_ms` summed, and that is better decided before launch than after. Happy to follow up with it.

### Testing

`cargo test -p mirrord-intproxy --lib`: 63 passed, including the regression test from #4749, which still holds because the connect handshake is still never deferred.

Not yet re-verified end to end against a cluster. #4749 was, with before and after measurements, and I will post the same for this branch.

### Review comments from #4749

Carried over: the terser test doc comment and the three comment trims. The changelog category question is answered differently here, for the reason above.

### DIFF
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -1,0 +1,7 @@
+Chaos latency rules no longer delay the response to a connect request. Applying a
+latency effect previously held that response back, which blocked the calling
+thread while the connection was set up and stopped client-side timeouts in
+single-threaded runtimes from firing during the fault. Latency is now applied to
+reads and writes only, so it no longer prevents an application from reacting to
+the delay it is being asked to survive. As a side effect, a `read_ms` value is no
+longer charged a second time when a connection is opened.
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -1,6 +1,6 @@
 Chaos latency rules no longer hold back the response to a connect request.
 Holding it back blocked the calling thread, which in a single-threaded runtime
-stalled the event loop, so a client-side timeout could not fire during the
-fault it was written for. A connection now waits for its delay before it
-carries data, on whichever of the first read or write comes first, which leaves
-the application free to act on it. The total delay a request sees is unchanged.
+stalled the event loop, so a client-side timeout could not fire during the fault
+it was written for. A connection now waits for its delay before it carries data,
+on whichever of the first read or write comes first, which leaves the
+application free to act on it.
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -1,5 +1,6 @@
-Chaos latency rules no longer delay the response to a connect request. Holding
-that response back blocked the calling thread, which stopped client-side timeouts
-from firing during the fault. Latency now affects reads and writes only, so your
-app can react to the delay. A `read_ms` value also costs its delay once per
-operation, rather than twice on a new connection.
+Chaos latency rules no longer hold back the response to a connect request.
+Holding it back blocked the calling thread, which in a single-threaded runtime
+stalled the event loop, so a client-side timeout could not fire during the
+fault it was written for. A connection now waits for its delay before it
+carries data, on whichever of the first read or write comes first, which leaves
+the application free to act on it. The total delay a request sees is unchanged.
--- a/changelog.d/+chaos-connect-latency.fixed.md
+++ b/changelog.d/+chaos-connect-latency.fixed.md
@@ -3,4 +3,4 @@ Holding it back blocked the calling thread, which in a single-threaded runtime
 stalled the event loop, so a client-side timeout could not fire during the fault
 it was written for. A connection now waits for its delay before it carries data,
 on whichever of the first read or write comes first, which leaves the
-application free to act on it.
+application free to act on it. The total delay a request sees is unchanged.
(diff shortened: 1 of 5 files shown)


eyalb181:
Verified end to end against a live cluster. One result confirms the fix, and one **contradicts a claim I made in the description**, which I have removed from the changelog in `f86410a`.

Setup as in #4749: a Node service under `mirrord exec`, calling a downstream HTTP service, bounding each call with `AbortSignal.timeout(2000)` and retrying twice before degrading.

### The fix works, and works more consistently than stock

`latency.read_ms = 8000`, `percentage = 100`, well over the client's 2s bound:

| | result |
|---|---|
| stock 3.249.0 | 10.18 / 18.22 / 18.21s |
| this branch | 4.03 / 4.15 / 4.19 / 4.35 / 4.11 / 4.16s |

Six samples, all landing at roughly two attempts of 2s each. The timeout fires as written, and the spread is tight where stock's is not, because stock's outcome depends on whether the request happened to reuse a pooled connection.

### The correction

I claimed the total delay a request sees is unchanged, reasoning that the same charges are applied in the same quantity, just not inside a blocking call. That is not what the measurements show.

`latency.read_ms = 700`, deliberately under the bound so nothing times out, with requests spaced 7s apart so the connection pool expires and each request opens a fresh connection:

| | fresh-connection request |
|---|---|
| baseline, no rule | ~0.18s |
| stock 3.249.0 | 1.93 / 1.94 / 1.93s |
| this branch | 1.21 / 1.22 / 1.23 / 1.22s |

Stock adds about 1.75s over baseline, this branch about 1.04s. The difference is close to exactly one 700ms charge, so this branch appears to apply one fewer than stock does.

I can't currently account for that. On paper both should apply the connect-derived delay once and `read_ms` once on the first read. Either stock charges something a third time, or the debt here is being absorbed somewhere I have not traced, possibly overlapping with the concurrent call the same request makes to the other dependency.

Worth pinning down before this merges, because it decides how the change should be described. If the delay genuinely drops, this is a `changed` entry after all, and @gememma's original point about the changelog category applies here too rather than only to #4749.

Pooled connections show no difference at all between the two, which is consistent with the connect charge only ever applying to new connections.

### Unrelated anomaly

One request in an earlier run failed with `credit-check-service unreachable after 3 attempts`, on a rule that targets the fraud service only. It did not recur across the later runs. Recording it rather than attributing it to this change, since I have no evidence either way.

### Other checks

`cargo test -p mirrord-intproxy --lib`: 63 passed, including #4749's regression test, which still holds because the handshake is never deferred. Clippy and fmt clean.

eyalb181:
Correcting my previous comment: the delay is **not** different, and I have restored that statement to the changelog in `b0dcda6`.

I traced the three delay sites and ran a single request on a genuinely fresh connection with `read_ms: 700`. Both charges land:

```
connect debt   Some(700ms)   computed at connect, response not deferred
write delay    700ms         debt taken by the first write
read  delay    700ms         charged separately on the first read
```

Total 1400ms on the faulted path, and the request measured 2.01s against stock's 1.93s for the same rule. The other dependency the same request calls is an in-run control: no rule matches it, so its debt is `None` and both its delays are `0ns`.

**Where my earlier numbers went wrong.** The 1.21s figures were pooled connections, not fresh ones. Spacing requests 7s apart did not reliably expire the HTTP client's keep-alive pool, so those requests reused an existing connection, skipped the connect charge entirely, and paid only `read_ms`. Stock happened to get fresh connections in its run, so I was comparing fresh against pooled and reading the difference as a missing charge.

That also explains the shape of it: the gap was almost exactly one 700ms charge, which is what a pooled connection saves.

So the position in the description stands. The same charges are applied in the same quantity; only the blocking goes away. Pooled connections are unaffected either way, because there is no connect for the charge to attach to.

Sorry for the noise. Retracting a correct claim on the strength of an uncontrolled measurement was the wrong call, and the instrumented run is what should have settled it in the first place.

The instrumentation was temporary and is not part of the branch.

PR #4757, opened 2026-08-23
Unify session monitor header across own and shared-operator views

liron-sel:
## Summary
- The own-session view (`SessionDetail`) showed Session ID / License key / Port / Mode, and never surfaced the actual session key.
- The shared-operator (team member) view (`OperatorSessionDetail`) showed Namespace / Session ID / Key / Container / HTTP filter / Locked ports / Queue splits, with no explicit Mode field.
- Both views now lead with the same Session ID / Key / Mode / Port fields, in the same order and under the same labels.
- Auto-generated session keys (`EnvKey::is_generated()`) are threaded through `SessionInfo` (`session-monitor-protocol`) and marked `(auto-generated)` only in the viewer's own session view — a teammate looking at your session on a shared operator does not see the marker.

## Test plan
- [x] `cargo check` / `cargo clippy --all-targets -- --deny warnings` / `cargo fmt --check` on `mirrord`, `mirrord-session-monitor-protocol`, `mirrord-intproxy`
- [x] `pnpm typecheck` / `pnpm lint` / `pnpm format:check` on `packages/monitor`
- [ ] Manual check in `mirrord ui`: start a session with an explicit `--key`, confirm the header shows Session ID / Key / Mode / Port with no auto-generated marker; start one without `--key`, confirm the marker appears only in your own view, not when viewed via a teammate's operator session

### DIFF
--- a/changelog.d/+session-monitor-unify-header.changed.md
+++ b/changelog.d/+session-monitor-unify-header.changed.md
@@ -1,0 +1,1 @@
+`mirrord ui`'s session detail view shows a consistent Session ID / Key / Mode / Port line for both your own sessions and sessions viewed on a shared operator, and now displays the actual session key (previously shown only for the shared-operator view). Auto-generated keys are marked as such only in your own session view.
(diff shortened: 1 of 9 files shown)


liron-sel:
Superseded — reworking this to unify both views completely (identical fields/order/labels) rather than the partial unification here. New PR to follow.

PR #4758, opened 2026-08-23
Unify session monitor header: same 8 fields, both views

liron-sel:
## What changed

1. **Both views now show exactly 8 fields, always, in the same order**: Key, Session ID, Port, Mode, Namespace, Container, HTTP filter, Queue splits. Previously the own-session and shared-operator views had different field sets, different orders, and different labels (own view was even missing `Key` entirely).
2. **Auto-generated keys are marked in the viewer's own session view only.** Threaded `EnvKey::is_generated()` through `mirrord-session-monitor-protocol::SessionInfo` (new `key_generated` field) so the CLI reports it; the shared-operator view never shows this marker, since the operator's session model has no concept of it (see edge case below).
3. **Ports render identically in both views**, via a new shared `PortModeChip` component (extracted from the shared-operator view's existing chip) — same chip, same mode/filter styling, used by both.
4. **HTTP filter is now shown for own sessions too**, extracted client-side from the local session's config diff (`feature.network.incoming.http_filter`), matching the same header/path/all_of/any_of logic the shared-operator view already used.
5. **Queue splits keeps the shared-operator view's existing live operator-reported count**; the own-session view derives the equivalent from the local config's *requested* `feature.split_queues` entries. These are different numbers in nature (live vs. requested) — noted in a code comment since it's a real semantic gap, not just a relabeling.
6. **A field with nothing to show renders `—`, never disappears.** Where a feature is fundamentally unavailable rather than just unused — Queue splits on a direct/OSS (non-operator) session, since queue splitting requires the mirrord Operator — it shows "Requires higher tier" instead.
7. Removed now-dead code from the earlier per-view fields: `extractLicenseKey`/`stringifyPrimitive`, the old per-view `describeFilter`/`splitSummary` duplication (now shared via `utils.formatQueueSplits`), and the local `PortChip`.

## Known edge case, flagged not solved

The auto-generated-key marker is only meaningful for own sessions because the mirrord Operator's session model has no field for "was this key auto-generated" anywhere — the CLI never even tells the operator this fact today. Making it show up identically in the shared-operator view would need a new field threaded through operator session creation → the operator's session CRD/status (private `operator` repo) → its session-listing API → this repo's operator client. Out of scope here; the views intentionally diverge on this one field.

## Test plan
- [x] `cargo check` / `cargo clippy --all-targets -- --deny warnings` / `cargo fmt --check` on `mirrord`, `mirrord-session-monitor-protocol`, `mirrord-intproxy`
- [x] `pnpm typecheck` / `pnpm lint` / `pnpm format:check` on `packages/monitor`
- [x] Verified visually via a temporary local harness rendering both components with mock data covering: fully-populated operator-backed session, direct/OSS session with nothing set (edge cases), and a teammate session — confirmed identical field sets/order/labels, `—` fallbacks, and tier-gating message
- [ ] Manual pass in a live `mirrord ui` against a real operator session recommended before merge


liron-sel:
@greptileai recheck

PR #4759, opened 2026-08-23
Match chaos rules to events on the hostname, not the resolved address

eyalb181:
## Problem

The session monitor's `Affected` event filter never selects anything, for any rule. With a latency rule armed and firing, its card showed 22 hits while `Affected` showed `0 events`.

Chaos rule selectors are written as hostnames, but the only destination an `outgoing_connection` event carried was `address`, which by then is a resolved IP. So the comparison in `matchChaosRule` comes down to:

```
"fraud-check-service.loan-demo.svc.cluster.local" === "10.96.27.156"
```

which is false for every rule. Nothing else in the stream can bridge it either: `dns_query` carries the short service name with no resolved address, and no event carries a chaos marker.

## Fix

The name is already available where the event is emitted. The connect request the layer sends carries the hostname the app originally asked for, and the proxy uses exactly that to decide whether a rule applies (`chaos_effect_for_address` → `applies_to_address`). This carries it on the event too and matches against it, keeping the resolved address as a fallback so a selector written as a literal address still works.

## Scope

This preserves the targeting semantics the existing comment describes, rather than changing them: a rule with a `percentage` still marks every connection it targets, not the subset actually faulted. Reporting only rules that genuinely fired would mean stamping the matched rule id on the event from the point where the effect is applied, which is a larger change and not attempted here.

## Verification

- `cargo check -p mirrord-intproxy` clean
- `pnpm typecheck` and `pnpm format:check` clean in `packages/monitor`
- `pnpm lint` reports one pre-existing error in `App.tsx:272`, identical before and after this change

COR-1821

### DIFF
--- a/changelog.d/+chaos-affected-filter-hostname.fixed.md
+++ b/changelog.d/+chaos-affected-filter-hostname.fixed.md
@@ -1,0 +1,2 @@
+Fixed the session monitor's `Affected` event filter, which compared chaos rule
+selectors against resolved IP addresses and so never matched anything.
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -704,6 +704,7 @@ impl IntProxy {
                     self.monitor_tx.emit(MonitorEvent::OutgoingConnection {
                         address: format!("{}", connect_req.remote_address),
                         port,
+                        hostname: connect_req.hostname().cloned(),
                     });
                 }
                 self.task_txs
--- a/mirrord/intproxy/src/session_monitor.rs
+++ b/mirrord/intproxy/src/session_monitor.rs
@@ -36,6 +36,13 @@ pub enum MonitorEvent {
     OutgoingConnection {
         address: String,
         port: u16,
+        /// Hostname the app originally asked for, before it was resolved to `address`.
+        ///
+        /// Chaos rule selectors are written as hostnames, so this is what the session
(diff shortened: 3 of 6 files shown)


eyalb181:
Tested end to end against a kind cluster, which turned up a real problem with the first version of this change.

Built this branch with `scripts/build_fat_mac.sh`, ran a service under `mirrord exec`, and read the session monitor's own SSE stream. The event now carries the name:

```json
{"address":"10.96.27.156:8000","hostname":"fraud-check-service.loan-demo.svc.cluster.local.","port":8000,"type":"outgoing_connection"}
```

Note the **trailing dot**. The first commit here compared the selector to that by equality, which would still have failed. The proxy does not compare by equality either — a name filter matches when the resolved hostname *contains* the selector (`AddressFilter::matches_socket_address`), which is also what lets a bare service name match its qualified form. The second commit matches the same way.

Running the shipped `matchChaosRule` against the captured event and a real rule:

```
PASS  FQDN selector matches (trailing dot)
PASS  short selector matches FQDN event
PASS  literal IP selector matches
PASS  does not match the other upstream
PASS  disarmed rule ignored
PASS  wrong port ignored
PASS  no hostname falls back to address
PASS  no hostname, name selector misses
```

And the same rule and event against the matcher as it exists on main returns `null`, which is the reported bug.

eyalb181:
Reopened from an upstream branch as #4761, since we no longer use forks. Same two commits.

PR #4761, opened 2026-08-24
Match chaos rules to events on the hostname, not the resolved address

eyalb181:
## Problem

The session monitor's `Affected` event filter never selects anything, for any rule. With a latency rule armed and firing, its card showed 22 hits while `Affected` showed `0 events`.

Chaos rule selectors are written as hostnames, but the only destination an `outgoing_connection` event carried was `address`, which by then is a resolved IP. So the comparison in `matchChaosRule` came down to:

```
"fraud-check-service.loan-demo.svc.cluster.local" === "10.96.27.156"
```

which is false for every rule. Nothing else in the stream can bridge it either: `dns_query` carries the short service name with no resolved address, and no event carries a chaos marker.

## Fix

The name is already available where the event is emitted. The connect request the layer sends carries the hostname the app originally asked for, and the proxy uses exactly that to decide whether a rule applies (`chaos_effect_for_address` → `applies_to_address`). This carries it on the event too and matches against it.

Matching is a substring test rather than equality, mirroring `AddressFilter::matches_socket_address`. That matters: the hostname on the event is fully qualified **with a trailing dot**, so equality fails, and the substring form is also what lets a bare service name match its qualified form. The resolved address is kept as an exact-match fallback so a selector written as a literal address still works.

## Testing

Built this branch with `scripts/build_fat_mac.sh`, ran a service under `mirrord exec` against a kind cluster, and read the session monitor's SSE stream. The event now carries the name:

```json
{"address":"10.96.27.156:8000","hostname":"fraud-check-service.loan-demo.svc.cluster.local.","port":8000,"type":"outgoing_connection"}
```

Confirmed in the UI: with a rule armed, `Affected` now lists the outgoing events and each one renders the matched rule name underneath. Before the change it showed "No events match the current filter."

The shipped `matchChaosRule` against that captured event and real rules:

```
PASS  FQDN selector matches (trailing dot)
PASS  short selector matches FQDN event
PASS  literal IP selector matches
PASS  does not match the other upstream
PASS  disarmed rule ignored
PASS  wrong port ignored
PASS  no hostname falls back to address
PASS  no hostname, name selector misses
```

The same rule and event against the matcher on main returns `null`, which is the reported bug.

Also clean: `cargo check -p mirrord-intproxy`, and `typecheck` / `format:check` in `packages/monitor`. `lint` reports one error at `App.tsx:272` which is identical on main and unrelated to this change.

## Scope

This preserves the targeting semantics the existing comment describes, rather than changing them: a rule with a `percentage` still marks every connection it targets, not the subset actually faulted. So the filter can list more events than the hit counter. Reporting only rules that genuinely fired would mean stamping the matched rule id on the event from the point where the effect is applied, which is a larger change and not attempted here.

COR-1821

### DIFF
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -16,8 +16,7 @@ use mirrord_config::{
     experimental::ExperimentalConfig, feature::network::incoming::tls_delivery::LocalTlsDelivery,
 };
 use mirrord_intproxy_protocol::{
-    IncomingRequest, LayerId, LayerToProxyMessage, LocalMessage, MessageId,
-    ProcessInfo,
+    IncomingRequest, LayerId, LayerToProxyMessage, LocalMessage, MessageId, ProcessInfo,
 };
 use mirrord_protocol::{
     CLIENT_READY_FOR_LOGS, ClientMessage, DaemonMessage, FileRequest, LogLevel,
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -704,6 +704,7 @@ impl IntProxy {
                     self.monitor_tx.emit(MonitorEvent::OutgoingConnection {
                         address: format!("{}", connect_req.remote_address),
                         port,
+                        hostname: connect_req.hostname().cloned(),
                     });
                 }
                 self.task_txs
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -149,6 +149,8 @@ pub struct IntProxy {
 
     /// Session monitor event sender
     monitor_tx: MonitorTx,
+    /// Live chaos rules, used to tag outgoing events with the rules that target them.
+    chaos_rx: ChaosWatcherRx,
 }
 
 /// Timing configuration for [`IntProxy`] maintenance tasks.
@@ -275,6 +277,7 @@ impl IntProxy {
             process_logging_interval,
             agent_tx,
             monitor_tx,
+            chaos_rx,
         }
     }
 
@@ -701,10 +704,23 @@ impl IntProxy {
             LayerToProxyMessage::Outgoing(req) => {
                 if let OutgoingRequest::Connect(ref connect_req) = req {
                     let port = connect_req.remote_address.get_port().unwrap_or(0);
+                    let chaos_rules = self
+                        .chaos_rx
+                        .borrow()
+                        .iter()
+                        .filter(|rule| {
+                            rule.applies_to_address(
+                                &connect_req.remote_address,
+                                connect_req.protocol,
+                                connect_req.hostname(),
+                            )
+                        })
+                        .map(|rule| rule.id)
+                        .collect();
                     self.monitor_tx.emit(MonitorEvent::OutgoingConnection {
                         address: format!("{}", connect_req.remote_address),
                         port,
-                        hostname: connect_req.hostname().cloned(),
+                        chaos_rules,
                     });
                 }
                 self.task_txs
--- a/mirrord/intproxy/src/lib.rs
+++ b/mirrord/intproxy/src/lib.rs
@@ -16,7 +16,7 @@ use mirrord_config::{
     experimental::ExperimentalConfig, feature::network::incoming::tls_delivery::LocalTlsDelivery,
 };
 use mirrord_intproxy_protocol::{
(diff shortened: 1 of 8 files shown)


gememma (changes requested):
I tried it (OSS) and the UI doesn't work for me, I added a rule with selector "www.google.com" and it shows the events for a second before editing the rule selector (???) to "www.google.com:0", and then it doesnt show any.

Image


Looks like this change doesn't work if you don't specify a port number, which is represented with `:0` internally. Easiest fix would be to chop off any trailing `:0`s from the new hostname string in the frontend, as long as its not an ipv6 address (`2001:db8::1:0` is a valid ipv6, for example)

packages/monitor/src/components/chaos/chaosMatch.ts
```diff
@@ -12,21 +12,43 @@ export function splitUpstream(upstream: string): {
...
-    if (target.host !== host) continue
-    if (target.port !== null && target.port !== port) continue
+    // A selector renders as `host:port`, and a rule that named no port renders as
+    // port 0, which the proxy reads as any port.
+    if (target.port !== null && target.port !== 0 && target.port !== port)
+      continue
+    const matched = isLiteralAddress(target.host)
+      ? target.host === resolved
+      : Boolean(hostname?.includes(target.host))
+    if (!matched) continue
```
gememma: this will still allow UDP traffic to count as matching the chaos rule, even though the chaos rule only matches TCP traffic - not sure how common this would be

gememma (commented):
hmm there are still ways which this matching is different from how its done in `AddressFilter` - like `0.0.0.0` should match all addresses but this doesnt handle that, and also subnets wont be handled correctly - see [`AddressFilter::matches_socket_address`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/feature/network/filter.rs#L95). I think the easiest way would be to replicate this logic as closely as possible.

The complete relationship between the user-requested address and traffic address in the backend is: 
1. in the chaos code, we create the AddressFilter with [`AddressFilter::from_str(requested_upstream)`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/feature/network/filter.rs#L185)
2. we use [`AddressFilter::matches_socket_address`](https://github.com/metalbear-co/mirrord/blob/main/mirrord/config/src/feature/network/filter.rs#L95) against the traffic `SocketAddress`.

mirrord/intproxy/src/lib.rs
```diff
@@ -701,9 +704,23 @@ impl IntProxy {
...
+                        .iter()
+                        .filter(|rule| {
+                            rule.applies_to_address(
+                                &connect_req.remote_address,
+                                connect_req.protocol,
+                                connect_req.hostname(),
+                            )
+                        })
+                        .map(|rule| rule.id)
+                        .collect();
```
gememma: This is returning multiple rules that could apply to this address, rather than the one we chose (we only ever apply one chaos rule) - you want the highest priority one and if there are multiple, a deterministic choice -

```suggestion
                        .max_by_key(|rule| rule.priority)
                        .map(|rule| rule.id);
```

^ this will return `Option<Uuid>` (Option is functionally a vec of 0 or 1, so it can be turned into a Vec if that's easier to deal with in frontend - in that case, keep the `collect()` call).

As mentioned in a comment elsewhere, this doesn't account for rules that don't apply to 100% of traffic, but that's ok - we choose randomly for those rules, so no way to replicate the logic here regardless

mirrord/intproxy/src/session_monitor.rs
```diff
@@ -36,6 +36,15 @@ pub enum MonitorEvent {
     OutgoingConnection {
         address: String,
         port: u16,
+        /// Ids of the chaos rules that target this connection.
+        ///
+        /// Resolved here because the proxy owns that decision: a selector is matched
+        /// against the hostname the app asked for as well as the resolved address, and
+        /// only the address survives into `address`. Targeting a connection is not the
+        /// same as faulting it, since a rule with a percentage targets every connection
+        /// it selects and faults a share of them.
```
gememma:
```suggestion
        /// The chaos rule that targets this connection, if there is one.
```

gememma (commented):
Last thing I think 🙏

gememma (dismissed):
nice 🙇 just need a lint fix for CI to be happy

gememma (approved):

PR #4764, opened 2026-08-24
feat(monitor): report unhandled errors as user-facing failures

aviramha:
Unhandled errors and promise rejections were emitted under a different event kind than the
error boundary's crash, so any breakdown by kind counted one and missed the other. Both
blank the UI for the person in front of it, so they're now reported the same way.

That left `kind` with exactly one legal value at every call site, so it's gone — the
parameter, the type, and the emitted property. `reason` already carries the distinction a
breakdown needs.

The part worth an opinion: events from here stop carrying `kind`, so a saved query
filtering on it matches only events recorded before this ships. I don't believe anything
worth keeping filters on it, but that should be a deliberate call rather than a side
effect — say so if you'd rather keep emitting it as a constant.

Second of three in the stack, based on #4763. `packages/monitor` has no test harness, so
verification is `tsc -b`, `eslint src`, and `prettier --check src`.

### DIFF
--- a/packages/monitor/src/analytics.ts
+++ b/packages/monitor/src/analytics.ts
@@ -98,7 +98,12 @@ export function trackEvent(
   posthog.capture(event, { source: 'session-monitor', ...properties })
 }
 
-export type EventKind = 'user_action' | 'health'
+// Every emitted event describes something a person was trying to do, including a crash,
+// which stops them mid-task just as surely as a failed request does. Background liveness
+// signals are deliberately not reported through here: they track how long a tab stayed
+// open rather than whether anyone was affected, so mixing them in makes the
+// blocked-versus-succeeded ratio unreadable.
+export type EventKind = 'user_action'
 
 export function emitUserBlocked(
   reason: string,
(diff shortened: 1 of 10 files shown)


packages/monitor/src/analytics.ts
```diff
@@ -82,7 +82,12 @@ export function trackEvent(
   posthog.capture(event, { source: 'session-monitor', ...properties })
 }
 
-export type EventKind = 'user_action' | 'health'
+// Every emitted event describes something a person was trying to do, including a crash,
+// which stops them mid-task just as surely as a failed request does. Background liveness
+// signals are deliberately not reported through here: they track how long a tab stayed
+// open rather than whether anyone was affected, so mixing them in makes the
+// blocked-versus-succeeded ratio unreadable.
+export type EventKind = 'user_action'
```
aviramha: why do we need an enum then? why not just send unhandled_error always?

aviramha: Agreed — dropped it entirely. `kind` had one legal value at all twenty-odd call sites once the liveness events were gone, so it discriminated nothing in a breakdown; `reason` already carries the distinction. The parameter, the type, and the emitted property are all gone.

One consequence worth naming: events emitted from here no longer carry `kind`, so any saved query filtering on it will match only events recorded before this ships.

hank-metalbear (approved):

hank-metalbear:
Nice, Thanks for helping to cleanup frontend observability!

PR #4765, opened 2026-08-24
feat(monitor): report each unhandled error once per page load

aviramha:
A component that throws on every render re-fires the window error handler for as long as
the tab stays open, at whatever rate the render loop runs. Every repeat carries the same
message, so the extras add nothing diagnostic while one stuck tab can outweigh every
healthy session in any aggregate built on them.

Each distinct message is now reported once per page load. The set of seen messages is
capped so a page producing an unbounded variety of them can't grow it without limit.

The part worth an opinion: this gates `captureException` as well as the analytics event.
The error-tracking issue is still created on the first occurrence, but occurrence counts
for a crash loop collapse to one per page load. That's the cost of not letting a single
tab flood the data; if the volume signal matters more, the dedupe can cover only the
analytics event.

Top of three in the stack, based on #4764. `packages/monitor` has no test harness, so
verification is `tsc -b`, `eslint src`, and `prettier --check src`.


packages/monitor/src/index.tsx
```diff
@@ -10,21 +10,34 @@ import { emitUserBlocked } from './analytics'
 
 let bootstrapped = false
 
+// A crash loop re-throws the same error for as long as the tab stays open, so one stuck
```
aviramha: shouldn't we just use posthog's error reporting capabilities with extra steps to make sure we don't get PII sent to us?

aviramha: assuming they have already that logic.

aviramha: Checked this against what the project has actually recorded rather than the config, and it doesn't hold up today — worth knowing why.

Exception autocapture *is* enabled here via remote config (`$exception_capture_enabled_server_side` is true on essentially every event this UI emits). But every `$exception` recorded from this surface is `handled: true` — i.e. the explicit `captureException` calls our own `emitUserBlocked` makes. There are no autocaptured unhandled exceptions from this surface at all, including across a real multi-hour crash loop that the `window.addEventListener('error')` path reported throughout. Something about the autocapture extension isn't working on a local origin with a custom `api_host`. So relying on it today wouldn't be simplification, it'd be losing crash visibility. Worth a separate issue.

On dedupe: 1.386.6 has remote-config suppression rules (rule-based, matched per exception) and a global events-per-second rate limiter, but nothing that collapses repeats of the same message. The loop's rate was well under the limiter, so autocapture wouldn't have prevented the flood either.

On PII, though — the concern is right and it doesn't distinguish the two paths. We send `error: event.message`, which is the same string autocapture would send, and we hand the throwable to `captureException` on top. This UI renders pod names, paths, URLs and request bodies, so a message can carry customer data either way, and replay here is already fully masked for exactly that reason. That says the scrubbing belongs in a global `before_send`, where it covers both paths, rather than in either one. Happy to do that as its own PR — it's a real gap and independent of this change.

packages/monitor/src/index.tsx
```diff
@@ -10,21 +10,34 @@ import { emitUserBlocked } from './analytics'
...
+const reportedErrors = new Set<string>()
+
+function reportUnhandledError(
+  error: string,
+  source: 'error' | 'unhandledrejection',
+  thrown: unknown,
+): void {
+  if (reportedErrors.has(error) || reportedErrors.size >= MAX_REPORTED_ERRORS) {
+    return
+  }
```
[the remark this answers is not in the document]
aviramha: Valid, fixed in 6870f479. The key is now the handler, the message, and the stack rather than the message alone, so two bugs that happen to share wording stay distinct and the second one's `captureException` still gets through. A loop re-throws from one site, so all three stay stable and repeats still collapse to one.

The `unknown rejection` case you flagged is only partly addressed, deliberately. A rejection carrying a non-Error has no stack to key on, so those still collapse together within a page load. Deriving something more specific would mean changing the `error` string itself, which is pre-existing behaviour on main and would shift what the property means for anything already reading it — worth doing, but not inside this change. Called out in the commit message so it isn't silently forgotten.

Worth noting for anyone reading this thread later: the latch pattern from #4565 (report on entering the failed state, reset on recovery) doesn't transfer here, since an unhandled window error has no recovered state to reset on. Hence the seen-set.

hank-metalbear (approved):

PR #4767, opened 2026-08-24
Unify session monitor fields, keeping container and process (PRO-220)

hank-metalbear:
Supersedes #4758. PRO-220.

## Summary

- My View and Team View render the same fields, same labels, same order: Key, Session ID, Port, Mode, Namespace, Container, HTTP filter, Queue splits.
- "License key" was showing the session key (`config.key`). Renamed to Key.
- Auto-generated keys read `<key> (auto-generated)` in your own view. Team view cannot show it: the operator session model has no equivalent.
- Container was always blank on operator and preview sessions (`OperatorSessionTarget::from_str` discarded it). Now carried through; a target with no container is `None`, not `""`.
- Process is back, owner view only. #4758 dropped it and the strip was the only place attached PIDs appeared.
- Empty fields are left out instead of rendering `-`. An ordinary session was showing four placeholders at once. `Requires higher tier` on Queue splits still shows.
- Values wrap instead of truncating, so a long key stays readable.

Out of scope, separate issues: "Join" vs "Share" wording (Aviram disputed the premise), context dropdown filtering the left-hand list.

## Test plan

- [x] `cargo xtask test-ut`, 124 pass, 3 new for target parsing
- [x] `cargo clippy -p mirrord --all-targets -- --deny warnings`
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm format:check` on `packages/monitor`; `vale` on the changelog
- [x] Live on minikube, operator 3.193.0: `mirrord exec` with steal + HTTP filter, no explicit key. Key renders `07e77 (auto-generated)` without truncating, session id in full, absent fields omitted
- [x] Teammate session on the same operator (second cluster identity via a ServiceAccount under a separate `HOME`): Team view renders Key, Session ID, Namespace, HTTP filter from the operator's session list


changelog.d/+session-monitor-unify-fields.changed.md
```diff
@@ -0,0 +1 @@
+`mirrord ui`'s session detail view now describes a session the same way whether it is your own or a teammate's on a shared operator. Both views draw from the same ordered field list, `Key`, `Session ID`, `Port`, `Mode`, `Namespace`, `Container`, `HTTP filter` and `Queue splits`, so a field that has a value always appears under the same label in the same place. Fields with nothing to report are left out rather than printed as placeholders. On a direct session `Queue splits` reads `Requires higher tier`, since queue splitting needs the mirrord operator. Your own view also flags an auto-generated key and lists the processes attached to the session. `Container` no longer renders blank for an operator session whose target names one.
```
cristeigabriela: this is quite long

PR #4772, opened 2026-08-25
Make install.sh POSIX sh compatible

### LINKED ISSUE #4762: Teams trial onboarding uses incompatible `sudo sh` installer command
### Bug Description

Hello! The teams trial onboarding uses a "sudo sh" installer command:

```curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | sudo sh```

This fails on Ubuntu 24.04 because /bin/sh resolves to Dash. I was able to get around this with:

```curl -fsSL https://raw.githubusercontent.com/metalbear-co/mirrord/main/scripts/install.sh | bash```

I'd recommend updating the onboarding page to invoke Bash explicitly!

### Steps to Reproduce

1. Sign up for a free trial here; https://app.metalbear.com/account/sign-up?_gl=1*1yvpoqp*_gcl_au*NTY3ODA4MzIuMTc4NzU2MTUyNg..
2. Add your details
3. Verify Account
4. Follow the steps until you get to the point where you install the mirrord cli above. When selecting Linux this appears.

### Backtrace

```shell

```

### mirrord layer logs

```shell

```

### mirrord intproxy logs

```shell

```

### mirrord agent logs

```shell

```

### mirrord config

```json

```

### mirrord CLI version

mirrord 3.247.0

### mirrord-agent version

_No response_

_No response_

_No response_

Ubuntu 24.04

### Local process

Install mirrord cli

### Local process version

_No response_

### Additional Info

_No response_

aviramha:
The installer is documented as `curl ... | sh` but was written in bash. On distros where `/bin/sh` is dash (Ubuntu, Debian) the bashisms break it — most importantly `$OSTYPE`, which is bash-only, so platform detection falls through to the "unsupported platform" branch.

Rewrote it in POSIX sh: `case` instead of `[[ ]]` pattern matching, `uname -s` instead of `$OSTYPE`, plain `[ ]` tests, and no `function` keyword. Behaviour is otherwise unchanged.

Second commit adds CI for this, gated on `scripts/install.sh` (or the workflow) changing. It pipes the script into each shell on each platform we support — sh/bash/dash/zsh/busybox ash on linux x86_64 and aarch64, sh/bash/zsh on mac aarch64 and x86_64 — and runs the resulting binary. Plus a shellcheck `-s sh` and `dash -n` job for the branches the install jobs never reach, and one cell covering version-selection precedence.

Confirmed the new job would have caught this: the pre-fix script piped into dash dies with `Syntax error: "(" unexpected`.

Closes #4762

### DIFF
--- a/.github/workflows/test-install-script.yaml
+++ b/.github/workflows/test-install-script.yaml
@@ -1,0 +1,102 @@
+# Verifies `scripts/install.sh` against every platform it claims to support, under
+# every shell a user might pipe it into.
+#
+# The documented invocation is `curl ... | sh`, so the script has to be POSIX sh:
+# on Debian/Ubuntu `/bin/sh` is dash, and a bashism there breaks the install with a
+# syntax error rather than anything actionable. See #4762.
+#
+# The install jobs pipe the script into the shell from stdin, the same way the
+# documented command does, and then run the binary that comes out.
+name: Test install script
+
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          sparse-checkout: scripts/install.sh
+          sparse-checkout-cone-mode: false
+
+      # `-s sh` uses the POSIX dialect regardless of the shebang.
+      - name: shellcheck, POSIX sh dialect
+        run: shellcheck -s sh scripts/install.sh
+
+      - name: parse under dash
+        run: dash -n scripts/install.sh
+
+  install:
+    name: install (${{ matrix.name }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      # The full set of shells runs on one arch per platform and the second arch runs
+      # a single sanity test. `sh` is listed separately from the shell it points at to test
+      # the indirection specifically.
+      matrix:
+        include:
+          - { name: "linux x86_64 / sh", os: ubuntu-24.04, sh: "sh" }
+          - { name: "linux x86_64 / bash", os: ubuntu-24.04, sh: "bash" }
+          - { name: "linux x86_64 / dash", os: ubuntu-24.04, sh: "dash" }
+          - { name: "linux x86_64 / zsh", os: ubuntu-24.04, sh: "zsh" }
+          - { name: "linux x86_64 / busybox ash", os: ubuntu-24.04, sh: "busybox sh" }
+          - { name: "linux aarch64 / sh", os: ubuntu-24.04-arm, sh: "sh" }
+          - { name: "mac aarch64 / sh", os: macos-latest, sh: "sh" }
+          - { name: "mac aarch64 / bash", os: macos-latest, sh: "bash" }
+          - { name: "mac aarch64 / zsh", os: macos-latest, sh: "zsh" }
+          - { name: "mac x86_64 / sh", os: macos-15-intel, sh: "sh" }
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          sparse-checkout: scripts/install.sh
+          sparse-checkout-cone-mode: false
+
+      - name: Install the shell under test
+        if: ${{ runner.os == 'Linux' && (matrix.sh == 'zsh' || matrix.sh == 'busybox sh') }}
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq zsh busybox-static
+
+      # $SH is deliberately unquoted, so that "busybox sh" splits into a command
+      # and its argument.
+      - name: Install, piped into the shell the same way the docs pipe it
+        env:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  # Static pass over the whole script. The install jobs only ever take the happy
+  # path, so this is what covers the branches they don't reach - unsupported
+  # platform, unsupported architecture, the sudo fallback.
```
gememma: don't think we need this? up to you
```suggestion
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+  lint:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          sparse-checkout: scripts/install.sh
+          sparse-checkout-cone-mode: false
+
+      # `-s sh` forces the POSIX dialect regardless of the shebang, so this fails
+      # on a bashism even if someone points the shebang back at bash.
```
gememma:
```suggestion
      # `-s sh` uses the POSIX dialect regardless of the shebang.
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+    name: install (${{ matrix.name }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      # The shells are the point, the architectures are not: the only thing arch
+      # changes is which asset the script downloads, so the full set of shells
+      # runs on one arch per platform and the second arch gets a single smoke
+      # test. `sh` is listed separately from the shell it points at because that
+      # indirection is the thing that broke: on Ubuntu it is dash, on macOS it is
+      # bash in sh mode.
```
gememma:
```suggestion
      # The full set of shells runs on one arch per platform and the second arch runs
      # a single sanity test. `sh` is listed separately from the shell it points at to test
      # the indirection specifically.
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+      matrix:
+        include:
```
gememma, line 43: no fish 🐟 ? 🥺

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+      # and its argument.
+      - name: Install, piped into the shell the same way the docs pipe it
+        env:
+          SH: ${{ matrix.sh }}
+        run: |
+          set -euo pipefail
+          $SH --version 2>&1 | head -n 1 || true
+          cat scripts/install.sh | $SH
+          # The script only installs into ~/.local/bin when that is already on
+          # PATH, so whichever branch it took, the binary is reachable by name.
```
gememma:
```suggestion
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+          command -v mirrord
+          mirrord --version
+
+      # Version selection is `VERSION` env var, then positional argument, then
+      # whatever is latest. It is plain POSIX test syntax, so one platform and one
+      # shell is enough to cover it.
+      - name: Check version selection
```
gememma:
```suggestion
      - name: Check version selection (one platform/ shell only)
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+        if: ${{ matrix.os == 'ubuntu-24.04' && matrix.sh == 'sh' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # The release before the latest one, so that asking for it specifically
+          # is distinguishable from getting the default.
```
gememma:
```suggestion
          # Specify a non-latest release
```

.github/workflows/test-install-script.yaml
```diff
@@ -0,0 +1,116 @@
...
+          echo "Pinning to $pinned"
+
+          uninstall() { sudo rm -f "$(command -v mirrord)"; }
+
+          uninstall
+          sh scripts/install.sh "$pinned"
+          [ "$(mirrord --version)" = "mirrord $pinned" ]
+
+          # Piping needs `-s --` to get an argument through to the script, which is
+          # what a user passing a version to `curl ... | sh` has to write.
```
gememma:
```suggestion
          # use `-s --` to pass the argument to the script when using pipes
```

gememma (dismissed):
nice 🐟 nits/ comments only, resolve any you don't agree with

.github/workflows/test-install-script.yaml
aviramha: adding

aviramha: removing, found it it's not POSIX compliant 💀

gememma: 😭 amazing

gememma (approved):
🍥

PR #4774, opened 2026-08-25
Authenticate sessions-manager connections with a shared-secret header

aviramha:
Prerequisite for exposing sessions-manager over the internet — metalbear-co/infra#467 puts it behind a load balancer that refuses requests without a shared secret, and the client currently has no way to send one.

Neither the Socket.IO control-plane handshake nor the `connect_async` data-plane upgrade sent any headers a proxy could key on, so turning on that check would have locked mirrord itself out. Both now attach an optional header, which lets a proxy make the decision without understanding either protocol.

Off unless `MIRRORD_SESSIONS_MANAGER_AUTH_TOKEN` is set, so nothing changes for a directly reachable sessions-manager. `MIRRORD_SESSIONS_MANAGER_AUTH_HEADER` overrides the default `x-mirrord-sm-auth` name.

**On the base branch:** this belongs to Graf's demo work, which lives on `danielg/demo-snapshot-260804` on his fork — that branch isn't on origin, so this targets the packaging branch instead, where the sessions-manager client files are identical. Happy to retarget if the snapshot lands on origin.

The demo-side companion changes (a slim image stage carrying just the bootstrap library, for deployments that mount it from an init container, plus the ECS README) only apply to files that exist on the snapshot branch, so they're held back rather than included here.

Verified with `cargo clippy --deny warnings` on this base. Not exercised against a live proxy yet.

### DIFF
--- a/changelog.d/+sessions-manager-auth-header.internal.md
+++ b/changelog.d/+sessions-manager-auth-header.internal.md
@@ -1,0 +1,5 @@
+Send an optional shared-secret header on sessions-manager control-plane and
+data-plane connections, so a deployment can put an authenticating proxy in front
+of sessions-manager. Set `MIRRORD_SESSIONS_MANAGER_AUTH_TOKEN` to enable it, and
+`MIRRORD_SESSIONS_MANAGER_AUTH_HEADER` to override the default `x-mirrord-sm-auth`
+header name.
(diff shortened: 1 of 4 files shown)


itsamegraf:
target branch is stale and will be force pushed into another sessions-manager solution (non-socketio based).
pushed the commit to https://github.com/itsamegraf/mirrord/tree/demo-snapshot-260804 holding the last stable socketio based code

PR #4778, opened 2026-08-26
Update nix and tun2 (now just tun).

meowjesty:
- Issue: https://linear.app/metalbear/issue/COR-1801/investigate-future-rust-incompatibility-warnings

Drop `tun2` now that `tun` is updated again.
Update `nix`. Had to change 1 nix usage in the layer to `unsafe libc` as it's not really possible to do it without `unsafe` even with `nix` (and using rust primitives would also be kinda wrong there due to what's expected of `OwnedFd` and `BorrowedFd`).

The clanker found a (minor) potential issue in the agent.

### Quality Checklist:

  documentation)

### DIFF
--- a/mirrord/agent/src/file.rs
+++ b/mirrord/agent/src/file.rs
@@ -7,7 +7,7 @@ use std::{
     iter::{Enumerate, Peekable},
     ops::RangeInclusive,
     os::{
-        fd::{AsRawFd, RawFd},
+        fd::AsRawFd,
         unix::{ffi::OsStrExt, fs::MetadataExt, prelude::FileExt},
     },
     path::{Path, PathBuf, StripPrefixError},
@@ -17,7 +17,7 @@ use std::{
 use faccess::{AccessMode, PathExt as _};
 use libc::DT_DIR;
 use mirrord_protocol::{FileRequest, FileResponse, RemoteResult, ResponseError, file::*};
-use nix::unistd::UnlinkatFlags;
+use nix::{fcntl::AT_FDCWD, unistd::UnlinkatFlags};
 use tracing::{Level, error, trace};
 
 use crate::{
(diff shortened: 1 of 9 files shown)


mirrord/agent/src/file.rs
```diff
@@ -549,9 +549,12 @@ impl FileManager {
...
         };
 
-        let fd: Option<RawFd> = dirfd.map(|fd| fd as RawFd);
-
-        nix::unistd::unlinkat(fd, path.as_ref(), flags)
+        // `path` is already resolved against `dirfd`, so we can just use it with `AT_FDCWD`.
+        //
+        // TODO(alex): clanker says that what we're doing here is not really equivalent to
+        // `unlinkat`, renaming a dir or replacing the original path could cause issues, also
+        // potential issues with symlinks and permissions.
```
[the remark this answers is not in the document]
meowjesty: The TODO is not about the change itself, it's about a potential issue.

mirrord/agent/src/file.rs
```diff
@@ -545,9 +545,12 @@ impl FileManager {
             }
         };
 
-        let fd: Option<RawFd> = dirfd.map(|fd| fd as RawFd);
-
-        nix::unistd::unlinkat(fd, path.as_ref(), flags)
+        // `path` is already resolved against `dirfd`, so we can just use it with `AT_FDCWD`.
+        //
+        // TODO(alex): clanker says that what we're doing here is not really equivalent to
```
0x00A5, line 550: Don't really understand this note. which part is not `unlinkat` tho?
The previous code seems a legit bug, that it maps a `dirfd` (which is not an actual fd) to a raw fd, and pass it to `unlinkat`. It works because `path` is already absolute and `dirfd` gets ignored 🤔

0x00A5 (approved):

mirrord/agent/src/file.rs
meowjesty: This one might be a bit annoying, so let's track it in https://linear.app/metalbear/issue/COR-1834/preserve-directory-descriptor-semantics-for-unlinkat

It's not really related to this PR, just something the clanker found thanks to the `nix` changes we had to make here.

PR #4781, opened 2026-08-27
CI: verify mirrord-tui on CI

t4lz:
Stacked on #4779 `import-mirrord-tui` this diff is only the CI change.

Three additions, no new jobs:

```yaml
# test-integration.yaml, next to the tests it belongs with
- cd mirrord/tui && cargo build --target x86_64-unknown-linux-gnu
- cargo nextest run --target x86_64-unknown-linux-gnu -p mirrord-tui

# test-macos.yaml
- cargo nextest run -p mirrord-tui
```

## Already covered by existing checks:

- **clippy** — `--all-targets --all-features -- "-Wclippy::indexing_slicing" -D warnings` reaches it on all five targets: `--workspace` on linux, and via the `mirrord/*` `default-members` glob on the other four.
- **rustfmt, cargo shear, typos, cargo deny, rust-docs** — all workspace-wide.

## UT fix

`shell_is_constrained_to_the_pane_and_the_prefix_takes_the_keyboard_back` types `echo q` at the pane's shell and waits for it on screen, which assumed both the character the prompt prints and that it leaves six columns on the line. Neither holds in CI:

- `sh` running as root prompts with `#`, not `$`.
- On macOS `/bin/sh` is bash, started as a login shell, and its system rc files rebuild the prompt out of the hostname. A runner's hostname is 61 characters, so in the 78-column pane the echoed command wrapped onto the next line and no needle spanning the break could match it.

The `PS1` the test puts in the environment settles neither, because those rc files run after it is inherited. It now asks for a short prompt from inside the shell, where nothing runs afterwards to override it, and matches the echo together with that prompt — a stronger assertion than the bare command.

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -126,11 +126,7 @@ jobs:
         with:
           command: |
             cargo nextest run --target x86_64-unknown-linux-gnu -p mirrord-vpn
-      # The command someone working on the interface actually types, rather than reaching it with
-      # `-p` from the root: from here cargo selects this package alone, and only because the root
-      # manifest is virtual. `cargo run` itself cannot finish on a runner - there is no terminal for
-      # the interface to take over - and the smoke tests below cover launching the built binary.
-      - name: mirrord tui builds from its own directory
+      - name: verify mirrord tui builds as its own binary
         uses: ./.github/actions/run-in-ci-runner
         with:
           command: |
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -126,6 +126,15 @@ jobs:
         with:
           command: |
             cargo nextest run --target x86_64-unknown-linux-gnu -p mirrord-vpn
+      # The command someone working on the interface actually types, rather than reaching it with
+      # `-p` from the root: from here cargo selects this package alone, and only because the root
+      # manifest is virtual. `cargo run` itself cannot finish on a runner - there is no terminal for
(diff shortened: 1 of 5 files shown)


t4lz:
@greptileai review again please

.github/workflows/test-integration.yaml
```diff
@@ -126,6 +126,16 @@ jobs:
         with:
           command: |
             cargo nextest run --target x86_64-unknown-linux-gnu -p mirrord-vpn
+      - name: verify mirrord tui builds as its own binary
+        uses: ./.github/actions/run-in-ci-runner
+        with:
+          command: |
+            cd mirrord/tui && cargo build --target x86_64-unknown-linux-gnu
+      - name: mirrord tui UT
```
aviramha: unrelevant but important - does it reuse same container between invocations? if not, I think it's slower than just running both as part of same step inside same container.

aviramha (approved):

.github/workflows/test-integration.yaml
t4lz: It does not reuse the container — `run-in-ci-runner` is a `docker run --rm` per step, so each step pays a container start and then a `sudo chown -R` back over the whole workspace, `target/` included. Only the cargo caches and the target dir carry over, through the bind mounts.

So you are right that one step is cheaper, and they are now merged into one. `set -e` is there because the action runs the block as `bash -c`, where without it the step exit status would be the last command's and a build failure would be masked by passing tests. The build stays in a subshell so the tests still run from the workspace root.

skreborn (approved):

PR #4783, opened 2026-08-27
Report anonymous usage of the terminal interface

t4lz:
Third in the stack: #4779 (imports the crate) → #4781 (CI for it) → this. The base here is `tui-ci-standalone`, so this diff is only the analytics.

**Do not merge before metalbear-co/operator#2313 is deployed.**

## What is reported

Each action as it happens, rather than one summary when the interface exits:

| When | Extra |
| --- | --- |
| the interface started | — |
| a connection attempt to the cluster failed | — |
| a session was launched from the targets view | — |
| preview environments were stopped | how many that one command stopped |
| the interface was quit | how many times each tab was switched to |

All five carry the same `run_id`.

## Why not one summary at the end

A TUI stays open a long time and is often closed by closing its terminal window — a `SIGHUP` that kills the process before anything could be sent. A summary would often not arrive.

Reporting as things happen costs one request per major action and loses only the last one. The `run_id` keeps a run reassemblable from its events, and its length readable from their timestamps. The tab counts are the one thing that still rides on a clean exit, which is a deliberate trade for not sending a report per tab switch.

## Notes for review

- **The action is a number, not a name.** `AnalyticValue` deliberately has no string variant — that is what keeps target names and namespaces out of telemetry by construction — so the action is a `#[repr(u32)]` discriminant, the same way `execution_kind` is already reported. The mapping lives in the dashboard.
- **Off by default for the standalone binary.** `run` takes an `Option<TelemetrySession>`; `mirrord-tui` passes `None`, since it exists for working on the interface. `mirrord tui` passes one only when the resolved config leaves `telemetry` on.
- **`EVENT_KIND_HEADER`'s doc comment was wrong** and is corrected here: the server rejects an unrecognized value rather than treating it as a client session. Only a *missing* header falls back, for older CLIs.
- **`mirrord tui` now fails fast on a broken config**, like every other subcommand: `tui_command` resolves it up front and lets the error out, instead of opening the interface and reporting the same problem later as a failure to connect. A missing config is not an error — it resolves to defaults, and telemetry defaults to on.
- The interface reaches the reporter through `Context`, which every screen already clones, so the two action call sites needed no new plumbing.

`TELEMETRY.md` gains a section describing all of the above, and states the opt-out covers it.

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -18,6 +18,19 @@ When there's an error, we send the name of the error (out of a hard-coded list, 
 
 The opt-out below disables this event as well; setting `common.telemetry: false` in `mirrord-up.yaml` is honored the same way as `telemetry: false` in a regular mirrord config.
 
+## `mirrord tui`
+
+The terminal interface reports each of these as it happens, rather than one summary when it exits: that it
+started, that a connection to the cluster failed, that a mirrord session was launched from the targets view,
+and that preview environments were stopped (with how many that one command stopped). When the interface is
+quit rather than killed, it also reports how many times each tab was switched to.
+
+Every one of these carries a random identifier for the run, so the events of a single run can be grouped;
+it is generated per run and is not stored or reused. No target names, namespaces, cluster identifiers or
+config values are included. The standalone `mirrord-tui` development binary reports nothing at all.
+
+The opt-out below disables all of it.
+
 ## Disabling
 
 Telemetry can be disabled by specifying the following in the mirrord config file:
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -22,7 +22,6 @@ The opt-out below disables this event as well; setting `common.telemetry: false`
 
 The terminal interface sends reports on the following events:
 - TUI start
-- a connection to the cluster has failed
 - a mirrord session was launched from the targets view
 - preview environments were stopped - with how many that one command stopped
 - TUI exit - with how many times each tab was switched to
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -20,14 +20,16 @@ The opt-out below disables this event as well; setting `common.telemetry: false`
 
 ## `mirrord tui`
 
-The terminal interface reports each of these as it happens, rather than one summary when it exits: that it
-started, that a connection to the cluster failed, that a mirrord session was launched from the targets view,
-and that preview environments were stopped (with how many that one command stopped). When the interface is
-quit rather than killed, it also reports how many times each tab was switched to.
+The terminal interface sends reports on the following events:
+- TUI start
+- a connection to the cluster has failed
+- a mirrord session was launched from the targets view
+- preview environments were stopped - with how many that one command stopped
+- TUI exit - with how many times each tab was switched to
 
 Every one of these carries a random identifier for the run, so the events of a single run can be grouped;
 it is generated per run and is not stored or reused. No target names, namespaces, cluster identifiers or
-config values are included. The standalone `mirrord-tui` development binary reports nothing at all.
+config values are included.
 
 The opt-out below disables all of it.
 
--- a/mirrord/analytics/src/lib.rs
+++ b/mirrord/analytics/src/lib.rs
@@ -273,11 +273,16 @@ pub trait Reporter: Sized {
 pub enum ReportTarget {
     ClientSession,
     UpSession,
+    /// One action inside the mirrord terminal interface. Several are reported per run, tied
+    /// together by a `run_id` property.
+    TuiEvent,
 }
 
 /// Header the client sets to tell the analytics-server which event a report is.
-/// The server maps known values to PostHog event names and treats anything
-/// unrecognized (or missing) as a client session.
+///
+/// The server maps each known value to its own PostHog event name. A *missing* header is taken as
+/// a client session, for older CLIs; an *unrecognized* one is rejected with 400, so every value
+/// here has to exist server-side before a build that sends it ships.
 pub const EVENT_KIND_HEADER: &str = "x-mirrord-event-kind";
 
 impl ReportTarget {
@@ -286,6 +291,7 @@ impl ReportTarget {
         match self {
             ReportTarget::ClientSession => "client-session",
             ReportTarget::UpSession => "up-session",
+            ReportTarget::TuiEvent => "tui-event",
         }
     }
 }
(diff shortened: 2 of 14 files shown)


mirrord/cli/src/tui.rs
```diff
@@ -24,8 +28,23 @@ pub enum TuiCliError {
...
 
 /// The `mirrord tui` command handler.
-pub(crate) async fn tui_command() -> Result<(), TuiCliError> {
-    mirrord_tui::run()
+pub(crate) async fn tui_command(watch: drain::Watch, user_data: &UserData) -> CliResult<()> {
+    // Read here for the `telemetry` opt-out; the interface resolves the config again per
+    // connection, for the settings that a change of scope can change while it runs. Resolving up
+    // front is also what reports a broken config as the config error it is, the way every other
+    // subcommand does - left to the interface, the same error would surface as a failure to
+    // connect, once the alternate screen is already up and pointing at the wrong problem.
```
t4lz:
```suggestion
```

mirrord/cli/src/tui.rs
```diff
@@ -24,8 +28,23 @@ pub enum TuiCliError {
...
+    let config = LayerConfig::resolve(&mut ConfigContext::default())?;
+
+    let telemetry = TelemetrySession {
+        enabled: config.telemetry,
```
t4lz:
```suggestion
    let telemetry_enabled = LayerConfig::resolve(&mut ConfigContext::default())?.telemetry;

    let telemetry = TelemetrySession {
        enabled: telemetry_enabled,
```

mirrord/tui/src/telemetry.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        if let Ok(mut visits) = inner.tab_visits.lock() {
+            *visits.entry(tab).or_default() += 1;
+        }
+    }
+
+    /// The interface exited on its own terms, with how often each tab was visited.
+    ///
+    /// Lost whenever the interface is killed rather than quit, which is the common case - the
+    /// counts describe the runs that ended in `q` or `Ctrl+C`, not all of them.
```
t4lz:
```suggestion
    /// The interface is exiting gracefully. Report how often each tab was visited.
    ///
    /// This event will be lost whenever the TUI is killed rather than quit with `q` or `Ctrl+C`.
```

mirrord/cli/src/tui.rs
t4lz: Agreed and applied in 63aea0c28ac — it names why the config is resolved here at all, and does not hold a whole `LayerConfig` in scope for one bool.

Verified rather than assumed, since this sits on the config path: `cargo fmt --all --check`, clippy on both `-p mirrord` and the interface with `-Wclippy::indexing_slicing -D warnings`, and the interface's 89 tests all pass. Rebuilt the CLI and re-ran the three config cases — missing file and malformed file both still exit 1 with the config error, and no config at all still opens and quits cleanly. Behaviour unchanged.

mirrord/tui/src/screens/preview_envs.rs
```diff
@@ -66,6 +67,8 @@ impl PreviewEnvsScreen {
             }
         };
 
+        self.telemetry.preview_envs_stopped(candidates.len() as u32);
```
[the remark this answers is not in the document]
t4lz: Yeah, good comment, but it's analytics, we'll leave it like this, and maybe we'll add a separate event on preview envs failed to delete.

mirrord/tui/src/telemetry.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Clone, Default)]
+pub struct Telemetry(Option<Arc<Inner>>);
+
+struct Inner {
+    machine_id: Uuid,
+    watch: drain::Watch,
+    /// Ties this run's reports together.
+    run_id: Uuid,
+    /// How many times each tab was switched to, reported when the interface closes. Ordered so
+    /// the reported object is stable rather than however the map happened to hash.
```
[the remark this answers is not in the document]
t4lz: Valid — it is the repo's own rule in AGENTS.md, and several of those did say no more than the declaration below them. Fixed in d602f85a7cc.

Removed: `Action`'s "What a report is about" heading, `Session`'s doc and its `enabled` field, `disabled`, `started`, and the first sentence of `connection_failed`.

Kept deliberately, since these are the "why" the rule asks for rather than restatements:

- why the action is a number (`AnalyticValue` has no string variant, which is what keeps target names out of telemetry by construction)
- that `watch` is what keeps the runtime alive until an in-flight report finishes
- that `tab_visits` is ordered so the reported object is stable
- that `report` sends when the reporter drops at the end of the function
- `session_started` naming the targets view, which the method name does not

`cargo fmt --check`, clippy with `-Wclippy::indexing_slicing -D warnings`, `cargo doc` and the 89 tests are all clean afterwards.

mirrord/analytics/src/lib.rs
```diff
@@ -330,6 +336,35 @@ impl AnalyticsReporter {
...
+    pub fn for_tui_event(
+        enabled: bool,
+        watch: drain::Watch,
+        machine_id: Uuid,
+        run_id: Uuid,
+    ) -> Self {
+        let mut analytics = Analytics::default();
+        analytics.add("machine_id", machine_id);
+        analytics.add("run_id", run_id);
+        analytics.add("is_ci", ci_info::is_ci());
```
aviramha: can drop is_ci

TELEMETRY.md
```diff
@@ -18,6 +18,21 @@ When there's an error, we send the name of the error (out of a hard-coded list,
 
 The opt-out below disables this event as well; setting `common.telemetry: false` in `mirrord-up.yaml` is honored the same way as `telemetry: false` in a regular mirrord config.
 
+## `mirrord tui`
+
+The terminal interface sends reports on the following events:
+- TUI start
+- a connection to the cluster has failed
```
aviramha: I don't think we need that (connection to cluster has failed)

mirrord/analytics/src/lib.rs
t4lz: Dropped from `for_tui_event`. Left as-is on `new` and `for_up_event`.

TELEMETRY.md
t4lz: Removed — the line here, plus the `ConnectionFailed` action and its call site in `app.rs`.

The `Action` discriminants keep the gap at 2 rather than renumbering, so the numbers stay usable as a wire format when querying; there is a note on the enum saying so.

aviramha (approved):

mirrord/cli/src/tui.rs
```diff
@@ -24,8 +28,18 @@ pub enum TuiCliError {
 }
 
 /// The `mirrord tui` command handler.
-pub(crate) async fn tui_command() -> Result<(), TuiCliError> {
-    mirrord_tui::run()
+pub(crate) async fn tui_command(watch: drain::Watch, user_data: &UserData) -> CliResult<()> {
+    let telemetry_enabled = LayerConfig::resolve(&mut ConfigContext::default())?.telemetry;
```
skreborn, line 32: Does this take into account `-f <config>` or only `MIRRORD_CONFIG_FILE`?

skreborn (approved):

mirrord/cli/src/tui.rs
t4lz: mirrord tui doesn't take `-f`.

PR #4787, opened 2026-08-28
Send the attachment title Linear requires when counting a flake

skreborn:
`AttachmentUpdateInput.title` is required.

A tally that cannot be advanced now fails the step, but only after the Slack alert has gone out.


iniw (approved):
Gah damn we gotta rewrite this thing as a python script or something.

PR #4790, opened 2026-08-28
Keep docs of items that carry no docs of their own

### LINKED ISSUE #3460: medschool might not be handling config enums properly.
### Bug Description

https://github.com/metalbear-co/mirrord/blob/2f65071136bae286092cd2672be56f360f972e5b/mirrord/config/src/feature/fs/mode.rs#L31

This should be in `configuration.md`, but somehow it's not being added, so there's something weird going on.

An easy way to try this out is to modify the other docs for this enum https://github.com/metalbear-co/mirrord/blob/2f65071136bae286092cd2672be56f360f972e5b/mirrord/config/src/feature/fs/mode.rs#L25 run medschool and see if it's working properly.

### Steps to Reproduce

Run `medschool` in mirrord repo

### Local process

medschool

leopepe:
Closes #3460.

### Quality Checklist:

## What

`docs_from_attributes` returns `None` for two unrelated situations: an item explicitly hidden with `<!--${internal}-->`, and an item that simply carries no doc comment of its own. The `TryFrom<syn::Item*>` impls turn that `None` into an `Err` via `.ok_or(())?` and drop the item, so an **undocumented enum or struct is discarded together with the docs on its variants and fields**.

That is the behaviour behind this issue. `FsModeConfig` has an enum-level doc comment, so its variants survive; an enum without one disappears completely. It also explains why the repro suggested in the issue works — editing the enum's own docs is what toggles whether its variants show up at all.

Only `<!--${internal}-->` yields `None` now. An item with no docs of its own yields an empty `Vec` and keeps whatever is documented beneath it.

## Effect on `configuration.md`

Regenerated with `cargo run -p medschool -- --input ./mirrord/config/src --output configuration.md`, the diff is **+280 lines, -0**. Nothing that was previously documented is lost; two config areas that were silently missing come back:

- **`InnerFilter`** (`feature/network/incoming/http_filter.rs`) — an enum with no doc comment of its own, whose variants document `body_filter`, `body_filter.json`, `header_filter`, `header_filter_jq` and `path_filter`. All of it was absent from the reference.
- **`PreviewLabelsConfig`** (`feature/preview.rs`) — a struct with no doc comment of its own, documenting `feature.preview.labels.include` and `feature.preview.labels.exclude`.

## One thing worth flagging

Restoring `InnerFilter` adds 5 duplicate anchors, because `HttpFilterConfig` references it from both `all_of` and `any_of`, and medschool inlines a referenced type's subtree once per reference. The anchors in those doc comments are hardcoded, so both copies claim `{#feature-network-incoming-inner-body-filter}` and friends.

This is pre-existing behaviour rather than something introduced here — `LocalTlsDelivery` is duplicated the same way today via `https_delivery`/`tls_delivery`, and the generated file already contains 7 duplicate anchors on `main`. I left it alone to keep this PR on the issue, and I'd argue documenting the inner filters twice beats not documenting them at all, but happy to follow up separately if you'd like the inlining deduplicated.

I also asked in the issue about the `{#feature-fs-mode}` duplication (the `mode` field in `feature/fs/advanced.rs` and the `FsModeConfig` enum both carry a heading with that anchor) and about three copy-paste anchor typos in the config docs. Those are source-doc fixes rather than medschool ones, so they're not in this PR — tell me which way you want them and I'll open another.

## Tests

Two regression tests in `medschool/src/main.rs`, alongside the existing fixture-based ones:

- `undocumented_enum_keeps_variant_docs` — fails on `main` with `"# Root\n\n## Root - mode\n\n"` against the expected output that includes both variants.
- `internal_enum_is_still_hidden` — pins the other half of the contract, so `<!--${internal}-->` keeps dropping the whole item.

The three pre-existing tests are unchanged and still pass. Verified with the pinned `nightly-2026-08-13` toolchain:

```
cargo test -p medschool                  # 5 passed
cargo clippy -p medschool --all-targets --all-features -- -Wclippy::indexing_slicing -D warnings
cargo fmt -p medschool -- --check
```

### DIFF
--- a/changelog.d/3460.fixed.md
+++ b/changelog.d/3460.fixed.md
@@ -1,0 +1,1 @@
+Fixed `medschool` dropping items that carry no documentation of their own, which kept documented enum variants and struct fields out of the generated configuration reference.
--- a/medschool/src/main.rs
+++ b/medschool/src/main.rs
@@ -282,6 +282,50 @@ mod test {
     "#,
     ];
 
+    /// An enum that carries no docs of its own, but whose variants are documented.
+    ///
+    /// This mirrors `TlsDeliveryProtocol` in `mirrord-config`, where the heading lives on the
+    /// referencing field and the enum itself is left undocumented.
+    const UNDOCUMENTED_ENUM_EXPECTED: &str =
+        "# Root\n\n## Root - mode\n\n### Root - mode - local\n\n### Root - mode - remote\n\n";
+
+    const UNDOCUMENTED_ENUM_FILES: [&str; 1] = [r#"
+    /// # Root
+    struct UndocumentedEnumRoot {
+        /// ## Root - mode
+        mode: UndocumentedEnum,
+    }
+
+    enum UndocumentedEnum {
+        /// ### Root - mode - local
+        Local,
+
+        /// ### Root - mode - remote
+        Remote,
+    }
+"#];
+
+    /// An enum marked as internal must stay hidden, docs on its variants notwithstanding.
+    const INTERNAL_ENUM_EXPECTED: &str = "# Root\n\n## Root - mode\n\n";
+
+    const INTERNAL_ENUM_FILES: [&str; 1] = [r#"
+    /// # Root
+    struct InternalEnumRoot {
+        /// ## Root - mode
+        mode: InternalEnum,
+    }
+
+    /// <!--${internal}-->
+    /// # Should not show up
+    enum InternalEnum {
+        /// ### Should not show up either
+        Local,
+
+        /// ### Nor this one
+        Remote,
+    }
+"#];
+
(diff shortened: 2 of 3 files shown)


meowjesty:
Hey thanks for the PR. I'll try to get to it on monday. I know that we fixed some other medschool issue that was also related to `enum`s, but I guess some bugs were still lurking.

meowjesty:
> Restoring InnerFilter adds 5 duplicate anchors

Can we do something about this? If we just merge this as-is the docs will be a bit confusing with this duplication. I was wondering if we could just dedup docs in `produce_docs` (maybe some intermediate step that puts the string docs into a `HashMap` or something like that).

I would say this can't be merged without fixing the duplicates.

PR #4792, opened 2026-08-28
Resolve Kubernetes context with client configuration

meowjesty:
## Summary

- Resolve the selected Kubernetes context while loading the Kubernetes client configuration.
- Carry that context forward with the client so later context-scoped behavior does not read kubeconfig again and drift if the active context changes during startup.

This is the foundation for storing user-wide settings safely per Kubernetes context in COR-1798.

- Translation:

To support setting a kube context on the new user global config (not a mirrord config, it's a helper config for some things that can be shared), we want to keep the kube context around from when we resolve the user's proper kube config. Previously we would just keep the kube config around, but it lacks the name of the context (that we want to save in this new user config), so this PR changes it to keeping both the kube config and the context name.

Without this, we would need to resolve the user's cluster info again at some point during start-up (when the context name becomes relevant).

## Testing

- `cargo test --package mirrord-kube created_config_retains_the_context_used_for_resolution --target x86_64-unknown-linux-gnu`

### Quality Checklist

- [x] User-facing website documentation is not applicable to this internal context-resolution layer
- [x] E2E coverage is provided by the complete COR-1798 stack

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,7 +168,7 @@ dependencies = [
  "apple-bundles",
  "apple-flat-package",
  "apple-xar",
- "base64",
+ "base64 0.22.1",
  "bcder",
  "bitflags 2.13.1",
  "bytes",
@@ -196,7 +196,7 @@ dependencies = [
  "once_cell",
  "p12",
  "p256",
- "pem",
+ "pem 3.0.6",
  "pkcs1",
  "pkcs8",
  "plist",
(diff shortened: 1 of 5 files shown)


0x00A5 (commented):
The changes in this PR looks fine, it seems to borrow a lot of kube_client internals.
I don't fully understand the purpose of carrying the context. How is it going to be used?

0x00A5 (approved):

PR #4794, opened 2026-08-28
Apply user-wide configuration to all CLI commands

meowjesty:
## Summary

- Apply user-wide defaults to every CLI command that resolves mirrord configuration, not only `mirrord exec`.
- Thread the same user data through CI, container, listing, operator, session, preview, database branch, queue, VPN,
  `mirrord up`, and configuration verification flows.
- Keep command, environment, and project configuration authoritative over user-wide defaults.

This PR is stacked on #4796, #4804, #4793, and #4792.

- Translation:

Applies the new global user config to all the other commands.

## Testing

- `cargo clippy --package mirrord-kube --package mirrord-operator --package mirrord --all-targets --all-features --target x86_64-unknown-linux-gnu --keep-going -- --deny warnings`
- `nu manage.nu test-sample` against `metalbear-staging`, covering real operator and OSS sessions, automatic persistence,
  explicit and invalid contexts, project-over-user precedence, default context application, and unsetting values.

### Quality Checklist

- [x] User-facing website documentation can follow the stacked CLI review; behavior is covered by command help and E2E tests
- [x] I have written an E2E sample and run it against staging
- [x] I have explained what this PR introduces and why, and linked to relevant context and lower PRs


mirrord/cli/src/main.rs
```diff
@@ -1059,6 +1056,10 @@ fn main() -> miette::Result<()> {
             .await
             .inspect_err(|fail| trace!(?fail, "Failed initializing `UserData`!"))
             .unwrap_or_default();
+        let global_config = GlobalConfig::from_default_path()
```
0x00A5: nit: not all command needs this. Bonus point if we lazily load a static, we will not load this at all unless we need to read it (intproxy, extproxy never needs it, for examlpe)

0x00A5 (commented):
I think we should avoid passing a global_config reference everywhere, feels hacky. How about loading it lazily in a static?
Something like the following.

```rust
static GLOBAL: LazyLock<GlobalConfig> = LazyLock::new(|| GlobalConfig::load().unwrap_or_default());

impl GlobalConfig {
    fn apply(config: &mut LayerConfig) {
        config.operator = config.operator.or(GLOBAL.config.operator);
    }
}

fn resolve_layer_config(ctx: &mut ConfigContext) -> CliResult<LayerConfig> {
    let mut config = LayerConfig::resolve(ctx)?;
    GlobalConfig::apply(&mut config);
    Ok(config)
}
```

We only ever need to load it once and make it available in the CLI process. If we mutate the global config, it will go straight to the file and only affects future runs.

mirrord/cli/src/main.rs
meowjesty: Changed to the global global thingy.

MintSoup (approved):

PR #4800, opened 2026-08-30
chore(frontends): bump js-yaml + nanoid to patched versions (Dependabot highs)

eyalb181:
Resolves the open Dependabot **high** alerts in the mirrord frontends (`packages/ui`, `monitor`, `wizard`). Lockfile-only, both are transitive so no `package.json` changes.

| Package | Was | Now | Advisory |
|---|---|---|---|
| js-yaml | 4.2.0 | 4.3.2 | DoS (quadratic CPU on `!!omap`) |
| nanoid | 3.3.12 | 3.3.18 | DoS (infinite loop on invalid size) |

Both are DoS-only and reached via build/UI tooling. Part of the 2026 ISO/SOC2 vulnerability remediation.


hank-metalbear (approved):
Lockfile-only, both bumps land on the first patched versions of every open high advisory for the 4.x / 3.x lines. Frozen install, ui build, and typecheck across ui/monitor/wizard pass on the PR head.

PR #4804, opened 2026-08-31
Add schema-validated user configuration editing

meowjesty:
## Summary

- Add `mirrord user-config show`, `set`, and `unset` for inspecting and changing user-wide configuration.
- Address settings with RFC 6901 JSON Pointers so new fields do not require field-specific CLI arguments or handlers.
- Accept multiple set or unset operations in one command and apply them atomically.
- Validate the resulting JSON against `UserConfig` before updating `~/.mirrord/data.json`.

This PR is stacked on #4793. The import/export and apply-to-all PRs are stacked above it.

- Translation:

The user needs a way of (un)setting things in the global user config, and this PR introduces a cli interface for that. It uses JSON pointer syntax, which means that you set it with something like `feature/network/incoming=enabled` (an example, we don't really support this particular config), instead of "dot notation" (`feature.network.incoming=enabled`), as this seems to be a better spec when `.` could actually be part of the value we're trying to set.

## Testing

- `cargo test --package mirrord user_config --target x86_64-unknown-linux-gnu` (20 passed at this stack layer)
- `cargo clippy --package mirrord --all-targets --all-features --target x86_64-unknown-linux-gnu --keep-going -- --deny warnings`

### Quality Checklist

- [x] User-facing website documentation can follow the stacked CLI review; command behavior is documented by `--help`
- [x] E2E coverage is provided by the complete COR-1798 stack
- [x] I have explained what this PR introduces and why, and linked to relevant context and the lower PR
- [x] The COR-1798 changelog is owned by the lower user-configuration PR


meowjesty:
@greptile-apps re-review this please!

mirrord/cli/src/config/global_config.rs
0x00A5, line 1: I asked AI to summarize other CLI's config set / unset commands. I think we should follow the majority and make two adjustments.

1. Make the command `mirrord confg set` and `mirrord config unset`. It should default to editing global config. If we ever need the CLI to edit local config, we can introduce a `--path` or `--file` arg for specifying the local mirrord.json path
2. Instead of `mirrord config set /{json}/{path}={value}`, dotted path + space-separated value seems much more common for config editing. Examples:
* `jj config set signing.key {key-id}`
* `gh config set hosts.aliases '{"mb":"metalbear.com"}'`
*`mirrord config set feature.network.incoming.mode 'steal'`

meowjesty: 376bbc10d38959edb199a4d9f78ce51939eee6d0

There's a harmless quirk with `unset`, in which it sets the optional value to `null`, can we live with that? I tried it out on a couple of configs and it seems fine.

meowjesty:
@greptile-apps can you review this again, pls my beautiful.

mirrord/cli/src/data/global_config.rs
```diff
@@ -46,6 +79,60 @@ impl GlobalConfig {
...
+        match tokio::fs::read(path).await {
+            Ok(contents) => Ok(serde_json::from_slice(&contents)?),
+            Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(BTreeMap::new()),
+            Err(error) => Err(error.into()),
+        }
+    }
+
+    /// Atomically mutates a strictly parsed sparse global configuration document.
+    async fn mutate<E>(
+        update: impl FnOnce(&mut BTreeMap<String, Value>) -> Result<(), E> + Send + 'static,
```
0x00A5: We shouldn't be dealing with BTreeMap anymore, we should use `LayerFileConfig` for file content serde. This should resolve the issue you described in another comment too - unset command writes `null` value. The `LayerFileConfig` skip serializing if none for optional fields.

Cargo.toml
```diff
@@ -156,6 +156,7 @@ jaq-core = "2.2.1"
 jaq-json = "1.1.3"
 jaq-std = "2.1.2"
 json-patch = { version = "4.2", default-features = false }
+jsonptr = "0.7.1"
```
0x00A5: We should drop jsonptr completely. The only two data types we need is `LayerFileConfig` and `serde_json::Value`. `LayerFileConfig` for serde raw file content. `serde_json::Value` for dot path mutation (split by dot, and traverse key by key, and mutate the value).

0x00A5: Right the mutation does something like file -> BTreeMap -> serde_json::Value, mutate -> BTreeMap -> file. It should be file -> LayerFileConfig -> serde_json::Value, mutate -> LayerFileConfig -> file.

mirrord/cli/src/data/global_config.rs
```diff
@@ -46,6 +79,60 @@ impl GlobalConfig {
...
+
+    /// Validates a candidate as a regular unresolved mirrord file configuration.
+    pub(crate) fn validate_value(value: Value) -> Result<(), GlobalConfigValidationError> {
+        if value.is_object().not() {
+            return Err(GlobalConfigValidationError::ExpectedObject);
+        }
+
+        let mut unknown_field = None;
+        let config: LayerFileConfig =
+            serde_ignored::deserialize(value.into_deserializer(), |path| {
```
0x00A5: The generated type `LayerFileConfig` already have `#[serde(deny_unknown_fields)]`, we can drop `serde_ignore`.

mirrord/cli/src/data/global_config.rs
```diff
@@ -66,6 +154,162 @@ impl GlobalConfig {
...
+/// Handles all `mirrord config` subcommands.
+pub(crate) async fn global_config_command(args: GlobalConfigArgs) -> Result<(), GlobalConfigError> {
+    match args.command {
+        GlobalConfigCommand::Show => show().await,
+        GlobalConfigCommand::Set(args) => set(args).await,
+        GlobalConfigCommand::Unset(args) => unset(args).await,
+    }
+}
+
+async fn show() -> Result<(), GlobalConfigError> {
```
0x00A5: This show the global config unconditionally. But usually tools would show the merged result depending on whether there is an override file in the working directory.
For example, if $PWD/.mirrord/mirrord.json or $PWD/mirrord.json is present, it should show the merged results rather than the global.

If the issue doesn't require this, I think we can drop it. And implement only when we need it.

mirrord/cli/src/data/global_config.rs
```diff
@@ -66,6 +127,209 @@ impl GlobalConfig {
...
+            if next.is_array() {
+                return Err(GlobalConfigError::mutation_error(
+                    &self.dotted,
+                    "cannot traverse an array",
+                ));
+            }
+
+            if next.is_object().not() {
+                *next = Value::Object(Map::new());
+            }
```
0x00A5: I think if any of the parent segment is not object, we should throw that `cannot traverse a non-object value` error, rather than replacing non-object value silently with an empty object, and continue. 
For example, if I have config:
```
{
  "agent": {
    "image": "internal.repo/images/mirrord:latest"
  }
}
```

`agent.image` can be an string or an object according to our schema.

and if I run a command `mirrord config set agent.image.tag 'my-tag'`, the current code will silently replace the image string with an empty object, and then insert `agent.image.tag` with `'my-tag'`.

meowjesty: 20c9dfa79f535fbabb202ffc536d6f2bcc0ff741

0x00A5 (approved):

PR #4807, opened 2026-09-01
Make user data persistence atomic

meowjesty:
## Summary

- Replace in-place truncation of `~/.mirrord/data.json` with locked atomic file replacement.
- Avoid rewriting an up-to-date data file while retaining eager migration of valid older files that are missing newly introduced fields.
- Preserve the established recovery behavior for malformed data by replacing it atomically with defaults.
- Serialize concurrent updates through a stable lock file so independent mirrord processes do not lose each other's changes.

This PR is stacked on #4792. The user-wide mirrord configuration PRs are stacked above it and use this safer persistence path.

- Translation:

greptile found some annoying issues in how we were dealing with the `data.json` `UserData` file, I think we were not properly locking it and writes were not really atomic? Anyways, this PR introduces the `atomic-write-file` crate and improves some of the stuff we were doing.

## Testing

- `cargo test --package mirrord user_data --target x86_64-unknown-linux-gnu`
- `cargo clippy --package mirrord --all-targets --all-features --target x86_64-unknown-linux-gnu --keep-going -- --deny warnings`

### Quality Checklist

- [x] I have documented the persistence invariants and recovery behavior
- [x] Existing user data remains forwards compatible through eager migration
- [x] Malformed data retains the previous reset-to-default behavior
- [x] I have tested concurrent updates, interrupted writes, migration, and malformed data recovery
- [x] I have explained why this PR is a separate foundational layer in the COR-1798 stack

### DIFF
--- a/mirrord/cli/src/user_data.rs
+++ b/mirrord/cli/src/user_data.rs
@@ -1,11 +1,15 @@
-use std::{env::home_dir, ops::Not, path::PathBuf, sync::LazyLock};
-
-use fs4::tokio::AsyncFileExt;
-use serde::{Deserialize, Serialize};
-use tokio::{
-    fs,
-    io::{self, AsyncWriteExt},
+use std::{
+    env::home_dir,
+    fs::{self, OpenOptions},
+    io::{self, Write},
+    path::{Path, PathBuf},
+    sync::LazyLock,
 };
+
+use atomic_write_file::AtomicWriteFile;
+use fs4::fs_std::FileExt;
+use serde::{Deserialize, Serialize};
+use tokio::task;
 use tracing::trace;
 use uuid::Uuid;
 
@@ -22,10 +26,8 @@ static DATA_STORE_PATH: LazyLock<PathBuf> = LazyLock::new(|| DATA_STORE_DIR.join
 /// Data that we store in the user's machine at `~/.mirrord/data.json` that might be used
 /// for a variety of purposes.
 ///
-/// Whenever we deserialize the `UserData` json file, if there are any errors we generate
-/// a new one using [`UserData::default`]. To avoid overwriting the file with all new default
-/// values in case we got a deserialization error due to a missing field, each field here
-/// gets a `default` annotation, so only the missing fields will be updated.
+/// Missing fields use their defaults so older files remain compatible. Loading rewrites only data
+/// that needs migration or replacement, keeping the file current without rewriting every run.
 #[derive(Debug, Serialize, Deserialize)]
 pub(crate) struct UserData {
     /// Amount of times this user has run mirrord.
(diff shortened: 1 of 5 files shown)


mirrord/cli/src/user_data.rs
```diff
@@ -146,4 +98,174 @@ impl UserData {
     pub(crate) fn machine_id(&self) -> Uuid {
         self.machine_id
     }
+
+    async fn update_at_path(
+        path: &Path,
+        update: impl FnOnce(&mut Self) + Send + 'static,
+    ) -> io::Result<Self> {
+        Self::try_update_at_path(path, |data| {
```
0x00A5: What's the purpose of adding another inner function call `try_update_at_path`. Can it be inlined in `update_at_path`? Is it for unit test?

mirrord/cli/src/user_data.rs
```diff
@@ -146,4 +98,174 @@ impl UserData {
...
+
+    async fn try_update_at_path<E>(
+        path: &Path,
+        update: impl FnOnce(&mut Self) -> Result<(), E> + Send + 'static,
+    ) -> Result<Self, E>
+    where
+        E: From<io::Error> + Send + 'static,
+    {
+        let path = path.to_owned();
+        task::spawn_blocking(move || {
```
0x00A5, line 107: Seems the only await point is this spawn_blocking. Possible to make this function sync rather than `async`? Therefore the whole call chain becomes sync - `try_update_at_path` <- `update_at_path` <- `from_path` | `update_is_returning_wizard`

mirrord/cli/src/user_data.rs
```diff
@@ -146,4 +98,174 @@ impl UserData {
...
+            // Atomic replacement changes the data file's inode, so writers coordinate through a
+            // sidecar whose identity remains stable across commits.
+            let lock_path = path.with_extension("lock");
+            let lock_file = OpenOptions::new()
+                .read(true)
+                .write(true)
+                .create(true)
+                .truncate(false)
+                .open(lock_path)?;
+            lock_file.lock_exclusive()?;
```
0x00A5, line 121: Where do we unlock?

meowjesty: clanker says on drop, even if it fails.

mirrord/cli/src/user_data.rs
meowjesty: slop. In another PR we sort of need the `try` version, but I'll get the clanker to change the function itself rather than this weird duplication.

mirrord/cli/src/user_data.rs
meowjesty: Wouldn't that block other tokio runtime things, if this is made fully sync? I think with `spawn_blocking` the tokio runtime can keep going (for the other cli tokio stuff we're doing).

mirrord/cli/src/user_data.rs
```diff
@@ -1,11 +1,15 @@
-use std::{env::home_dir, ops::Not, path::PathBuf, sync::LazyLock};
+use std::{
+    env::home_dir,
+    fs::{self as blocking_fs, OpenOptions},
```
0x00A5: do we still need `self as blocking_fs`? `use tokio::fs;` is removed, nothing clashing

mirrord/cli/src/user_data.rs
0x00A5: oh it's unlocked when fs is closed by `File`'s drop

0x00A5 (approved):

PR #4814, opened 2026-09-01
Set Host from the authority when sending a stolen HTTP/2 request over HTTP/1

aviramha:
An HTTP/2 request carries its target in pseudo-headers and has no `Host` header, so converting one to HTTP/1 means recreating `Host` from the authority, as [RFC 9113 section 8.3.1](https://www.rfc-editor.org/rfc/rfc9113#section-8.3.1) requires. We were sending neither, and a server that enforces the HTTP/1.1 host requirement (Tomcat, for one) answers such a request with a bare 400 before the application is involved.

The target is also rewritten from absolute to origin form, which is what a server that is not a proxy expects.

Tested with unit tests over the conversion, plus one that asserts the bytes that reach a local HTTP/1 server.

Part of COR-1838.

### DIFF
--- a/mirrord/intproxy/src/proxies/incoming/http.rs
+++ b/mirrord/intproxy/src/proxies/incoming/http.rs
@@ -1,9 +1,11 @@
 use std::{fmt, io, net::SocketAddr, ops::Not};
 
 use hyper::{
-    Request, Response, StatusCode, Version,
+    Method, Request, Response, StatusCode, Uri, Version,
     body::Incoming,
     client::conn::{http1, http2},
+    header::{HOST, HeaderValue},
+    http::uri::PathAndQuery,
 };
 use hyper_util::rt::{TokioExecutor, TokioIo};
 use mirrord_protocol::{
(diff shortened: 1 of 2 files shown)


mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -174,6 +176,60 @@ pub fn mirrord_error_response<M: fmt::Display>(
...
+///
+/// [RFC 9113 section 8.3.1]: https://www.rfc-editor.org/rfc/rfc9113#section-8.3.1
+fn downgrade_to_http1<B>(request: &mut Request<B>) {
+    if request.version() != Version::HTTP_2 {
+        return;
+    }
+
+    // In a CONNECT request the authority is the target itself, and there is no path to fall back
+    // on.
+    if request.method() == Method::CONNECT {
```
aviramha, line 208: shouldn't this error?

aviramha: Good catch — it errors now (`LocalHttpError::UnsupportedHttp2Connect`, not retryable). The authority of a CONNECT is its target rather than a host to put in a header, and the tunnel differs between the protocols: HTTP/2 CONNECT tunnels one stream, HTTP/1 CONNECT tunnels the whole connection. Sending it through unconverted would mean something else than the client asked for.

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -174,6 +180,65 @@ pub fn mirrord_error_response<M: fmt::Display>(
...
+        return Ok(());
+    }
+
+    if request.method() == Method::CONNECT {
+        return Err(LocalHttpError::UnsupportedHttp2Connect);
+    }
+
+    *request.version_mut() = Version::HTTP_11;
+
+    let Some(authority) = request.uri().authority().cloned() else {
```
Razz4780: Unnecessary clone I think?

Razz4780 (approved):

mirrord/intproxy/src/proxies/incoming/http.rs
```diff
@@ -283,3 +351,161 @@ impl HttpSender {
...
+    use rstest::rstest;
+    use tokio::{
+        io::{AsyncReadExt, AsyncWriteExt},
+        net::{TcpListener, TcpStream},
+    };
+
+    use super::{HttpSender, LocalHttpError, StreamingBody, downgrade_to_http1};
+
+    /// Builds a request as [`hyper`]'s HTTP/2 server produces it: version [`Version::HTTP_2`], the
+    /// target in absolute form, and no `Host` header.
```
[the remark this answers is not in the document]
aviramha: Fixed — the lead lines that restated the test names are gone, and what remains on each test is only the rationale a name cannot carry (the RFC 9113 rule on authority vs `Host`, why an HTTP/1 request is left alone, why CONNECT has no equivalent, why the assertion is made on the wire bytes). One test lost its doc comment entirely, since the name and the cases already said everything.

mirrord/intproxy/src/proxies/incoming/http.rs
aviramha: Right, dropped it. The clone was there because `authority` was held across `headers_mut()`, but building the `HeaderValue` ends the borrow, so the immutable borrow is over before the mutable one starts.

PR #4815, opened 2026-09-01
Pick the local HTTP protocol from the request, not from the connection pool

aviramha:
Stacked on #4814.

An idle HTTP/1 client was reused for an HTTP/2 request, so whether the local application saw HTTP/2 or a silent conversion to HTTP/1 depended on what happened to be in the connection pool. Client reuse is keyed by version: a client serves only the protocol it made its handshake with.

A local server that speaks only HTTP/1 will now reject the cleartext HTTP/2 connection preface rather than being handed a converted request, which is the honest outcome — the protocol is the remote server's, not ours to change.

Tested with a store-level test: an idle HTTP/1 client is left alone while an HTTP/2 request gets an HTTP/2 client.

Fixes COR-1838.


Razz4780 (approved):

aviramha:
Re the Greptile finding on `client_store.rs` (service reused after move): false positive — this compiles and the test runs. `service_fn` returns `ServiceFn`, which is `Copy` when the closure is (`impl<F, R> Copy for ServiceFn<F, R> where F: Copy` in hyper 1.11), and this closure captures nothing, so the HTTP/1 `serve_connection` copies it rather than moving it. Verified locally: `cargo test -p mirrord-intproxy --lib` passes with `does_not_reuse_http1_client_for_http2_request` included.

(Posting here rather than in the thread — a pending review draft on this PR blocks threaded replies.)

PR #4817, opened 2026-09-01
feat(monitor): attach the mirrord version to session monitor telemetry

hank-metalbear:
## Summary

Stamps the mirrord version on session monitor telemetry as the person property `version`, matching what the CLI already sets on `client_session_v1` so a single breakdown key spans both surfaces.

The version is injected at build time from the workspace Cargo.toml (the bundle is embedded into the CLI binary built from the same checkout, so the bundle version is exactly the mirrord version it ships in) and applied with `setPersonProperties` before the first capture. That call is also what creates the person profile: under `person_profiles: 'identified_only'` the monitor has none, and `$set` on an anonymous event is dropped server-side. A build that never injected a version reports `unknown`, so the gap surfaces as its own bucket rather than as absent traffic.

Reading `sessions[0].mirrord_version` from the local session API was rejected: each session reports its own intproxy's version, and sessions from different CLI binaries (terminal CLI vs an IDE-bundled one) can coexist in one list, so the pick is arbitrary and can differ from the CLI actually serving the UI.

The UI ships inside the CLI and users upgrade on their own schedule; without the version, an issue already fixed in a newer release is indistinguishable from a live regression.

Breakdown: [Session monitor events by mirrord version](https://us.posthog.com/project/29024/insights/JygnMvkC)

Closes PRO-232.

### DIFF
--- a/packages/monitor/src/analytics.ts
+++ b/packages/monitor/src/analytics.ts
@@ -62,11 +62,6 @@ export function initAnalytics(telemetryEnabled: boolean) {
   // `posthog.init` leaves the client opted in, and init only runs with telemetry enabled, so
   // capturing is already in the desired state before any `setTelemetryEnabled` call arrives.
   appliedTelemetry = true
-  // Under `identified_only` the monitor has no person profile until something asks for one,
-  // and anonymous events drop `$set` server-side. `setPersonProperties` creates the profile,
-  // so this must run before the first capture below. A build that never injected the version
-  // reports `unknown` rather than nothing, so the gap is visible as its own bucket in a
-  // version breakdown instead of looking like an absence of traffic.
   posthog.setPersonProperties({ version: MIRRORD_VERSION ?? 'unknown' })
   posthog.capture('session_monitor_opened', { source: 'session-monitor' })
 }
--- a/packages/monitor/src/analytics.ts
+++ b/packages/monitor/src/analytics.ts
@@ -11,8 +11,8 @@ declare const __MIRRORD_VERSION__: string
  * Users upgrade on their own schedule, so without it a crash fixed in a newer release is
  * indistinguishable from a live regression.
  *
- * Reported as the person properties `version` / `initial_version`, matching what the CLI
- * sets on `client_session_v1` so one breakdown key spans both surfaces.
+ * Reported as the person property `version`, matching what the CLI sets on
+ * `client_session_v1` so one breakdown key spans both surfaces.
  */
 const MIRRORD_VERSION: string | undefined =
   typeof __MIRRORD_VERSION__ === 'undefined' ? undefined : __MIRRORD_VERSION__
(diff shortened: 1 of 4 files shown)


hank-metalbear:
Screenshot 2026-09-01 at 2 51 28 PM

packages/monitor/src/analytics.ts
```diff
@@ -22,6 +33,18 @@ export function initAnalytics(telemetryEnabled: boolean) {
...
     autocapture: false,
     capture_pageview: false,
+    // Stamped here rather than with `posthog.register`: registered super properties persist
+    // in localStorage on this fixed localhost origin, where a UI served later by a
+    // different CLI version would inherit them and misreport its version.
+    before_send: (event) => {
+      if (event && MIRRORD_VERSION) {
+        event.properties = {
+          ...event.properties,
+          mirrord_version: MIRRORD_VERSION,
```
aviramha: shouldn't it be person property?

aviramha: yeah it should in general

packages/monitor/src/analytics.ts
```diff
@@ -71,6 +62,15 @@ export function initAnalytics(telemetryEnabled: boolean) {
   // `posthog.init` leaves the client opted in, and init only runs with telemetry enabled, so
   // capturing is already in the desired state before any `setTelemetryEnabled` call arrives.
   appliedTelemetry = true
+  // Under `identified_only` the monitor has no person profile until something asks for one,
+  // and anonymous events drop `$set` server-side. `setPersonProperties` creates the profile,
+  // so this must run before the first capture below.
+  if (MIRRORD_VERSION) {
+    posthog.setPersonProperties(
+      { version: MIRRORD_VERSION },
+      { initial_version: MIRRORD_VERSION },
```
aviramha: why do we need initial_version?

packages/monitor/src/analytics.ts
```diff
@@ -48,6 +62,15 @@ export function initAnalytics(telemetryEnabled: boolean) {
   // `posthog.init` leaves the client opted in, and init only runs with telemetry enabled, so
   // capturing is already in the desired state before any `setTelemetryEnabled` call arrives.
   appliedTelemetry = true
+  // Under `identified_only` the monitor has no person profile until something asks for one,
+  // and anonymous events drop `$set` server-side. `setPersonProperties` creates the profile,
+  // so this must run before the first capture below.
+  if (MIRRORD_VERSION) {
+    posthog.setPersonProperties(
+      { version: MIRRORD_VERSION },
```
hank-metalbear: @aviramha let me know if this is what you meant

packages/monitor/src/analytics.ts
```diff
@@ -48,6 +62,12 @@ export function initAnalytics(telemetryEnabled: boolean) {
   // `posthog.init` leaves the client opted in, and init only runs with telemetry enabled, so
   // capturing is already in the desired state before any `setTelemetryEnabled` call arrives.
   appliedTelemetry = true
+  // Under `identified_only` the monitor has no person profile until something asks for one,
+  // and anonymous events drop `$set` server-side. `setPersonProperties` creates the profile,
+  // so this must run before the first capture below.
+  if (MIRRORD_VERSION) {
+    posthog.setPersonProperties({ version: MIRRORD_VERSION })
+  }
```
hank-metalbear: I think there should be a posthog metric if there is no version so we're aware

packages/monitor/src/analytics.ts
hank-metalbear: Ok, pushed under person property.

packages/monitor/src/analytics.ts
hank-metalbear: Took out intial_version

aviramha (approved):

PR #4821, opened 2026-09-02
feat(remote): route Go runtime accept syscalls into remote-layer

DmitryDodzin:
Go issues syscalls directly from its own runtime rather than through libc, so remote-layer's libc hooks never fire in a Go application: every connection was accepted locally and the agent saw no incoming traffic.

Extract the Go syscall trampolines, runtime version detection and symbol hooking out of mirrord-layer into a shared mirrord-layer-go crate, with each layer plugging in its own dispatch table, and give remote-layer a table routing accept, accept4, getsockname and getpeername into the same detours the libc hooks use.

close and the dup family are still passed through, so claimed-socket bookkeeping does not yet see fds a Go application closes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-# this change will be removed - intended to make mbe-1977 run the full ci
+# this change will be removed - intended to make mbe-1979 run the full ci
 # Hello traveler!
 # This is the CI workflow for mirrord.
 # It is a bit complicated, but it is also very powerful.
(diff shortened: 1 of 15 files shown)


itsamegraf:
@greptileai please rereview

itsamegraf (approved):

itsamegraf:
merged manually into base branch - https://github.com/metalbear-co/mirrord/commit/4c35439737b7814831c43e677fd07dd6c931d58f

PR #4826, opened 2026-09-02
intproxy: properly track remote fds across disconnects

iniw:
## Problem

Intproxy adds an offset to agent file descriptors before it sends them to the layer. It maps these descriptors back to their agent values when it receives file requests from the layer.

However, intproxy also stored the offseted, layer-facing descriptors in its internal resource and buffer tables. During normal operation, the offset is zero, so both values are equal and the problem stays hidden.

After a reconnect however, the offset changes. Internal tables then store layer-facing descriptors, while file operations use agent-facing descriptors. This causes explicit close operations to fail. When the layer disconnects, intproxy can also send the wrong descriptors to the agent. The agent then logs `fd not found` and `close_dir -> fd not found` errors.

An Operator restart can also expose the issue because it causes intproxy to reconnect.

## Solution

Keep agent-facing file descriptors in all internal resource and buffer tables.

Apply the offset only when intproxy sends a response to the layer. Continue to map descriptors back before processing requests from the layer.

This keeps one descriptor range inside intproxy and makes explicit close operations, buffered reads, directory reads, and layer cleanup work after a reconnect.

## Testing

The first commit in this PR adds a regression test that fails without the second commit, which has the fix for the issue.

Given that, I recommend reviewing commit by commit :) As always :)

### DIFF
--- a/mirrord/intproxy/src/proxies/files.rs
+++ b/mirrord/intproxy/src/proxies/files.rs
@@ -347,7 +347,7 @@ impl RouterFileOps {
         Ok(Some(request))
     }
 
-    /// Return a response to be sent to the client.
+    /// Maps agent-facing file descriptors in a response before sending it to the layer.
     #[tracing::instrument(level = Level::TRACE, ret)]
     pub fn map_response(&mut self, mut response: FileResponse) -> FileResponse {
         match &mut response {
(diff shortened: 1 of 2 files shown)


mirrord/intproxy/src/proxies/files.rs
0x00A5, line 905: Wonder if we can just call `self.reconnect_tracker.map_response(response)` before this line. So we don't need to edit each variant separately.
Otherwise, someone adds a new variant may forget to handle the offset here properly.

0x00A5 (approved):

PR #4851, opened 2026-09-07
sip: skip patching only when the DYLD entitlement is granted

MaxFreedomPollard:
### Quality Checklist:

- [x] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs, documentation)

### The bug

`has_dyld_entitlement` (`mirrord/sip/src/lib.rs:573`, the `.contains(...)` on line 582) asks whether the entitlements plist *mentions* `com.apple.security.cs.allow-dyld-environment-variables`. An entitlement is a key plus a value, and it can be listed with the value `false`, so the key being present does not mean the binary keeps its `DYLD_*` variables.

`is_binary_sip` returns `is_restricted && !has_dyld_entitlement(data)` (line 607), so a hardened binary whose entitlement is turned off comes back as `NoSip`. mirrord runs it unpatched, macOS strips `DYLD_INSERT_LIBRARIES` from it, the layer never loads, and nothing in the logs says why.

macOS 26 signs `/usr/bin/aa` exactly like that:

```
$ codesign -dvvv /usr/bin/aa 2>&1 | grep flags=
CodeDirectory v=20500 size=1485 flags=0x10000(runtime) hashes=36+7 location=embedded

$ codesign -d --entitlements :- /usr/bin/aa
...<key>com.apple.security.cs.allow-dyld-environment-variables</key><false/><key>com.apple.security.cs.disable-library-validation</key><false/>...
```

`/usr/bin/aa` is the binary `entitled_binary_is_not_sip` uses. On macOS 26.2 that test fails on `main` today: `sip_patch` returns `None`, and then `DYLD_PRINT_LIBRARIES=1 /usr/bin/aa` prints nothing, because the entitlement is off. It passes on the `macos-latest` runner, whose `/usr/bin/aa` is granted the entitlement, so I left the test as it is.

### The fix

`grants_dyld_entitlement` reads the value element that follows the entitlement key and accepts only `<true>`. `has_dyld_entitlement` calls it instead of `contains`. Nothing else about the SIP decision moves: a binary that really is granted the entitlement is still left unpatched, and one that never lists it is still patched.

Three tests come with it. Two sign `/bin/wait4path` into a temp file with the hardened runtime, once with the entitlement set to `false` and once to `true`, and check `get_sip_status`. The third feeds the value forms (`<true/>`, `<true />`, `<true></true>`, `<false/>`, `<string>true</string>`) straight to the parser. None of them depend on the macOS version or on which binaries Apple happens to entitle.

### Verification

macOS 26.2, arm64, reproducing the `macos-latest` job for this crate: `rm rust-toolchain.toml`, Homebrew `protoc` on `PATH`, `cargo-nextest` 0.9.143, stable rustc 1.95.

The new regression test against the unmodified `has_dyld_entitlement`:

```
$ cargo nextest run -p mirrord-sip -E 'test(binary_with_disabled_dyld_entitlement_is_sip)'
FAIL  mirrord-sip main::tests::binary_with_disabled_dyld_entitlement_is_sip
assertion failed: matches!(get_sip_status(...).unwrap(), SipBinary(_))
```

With the fix:

```
$ cargo nextest run -p mirrord-sip -E 'not test(entitled_binary_is_not_sip)'
Summary [0.975s] 32 tests run: 32 passed, 1 skipped
```

`entitled_binary_is_not_sip` is excluded above because it fails on this machine before and after the change, for the reason described in the first section. On `main` it fails on the `DYLD_PRINT_LIBRARIES` assertion; with the fix it fails on `assert!(result.is_none())`, since `/usr/bin/aa` on macOS 26 does need patching.

Lint, exactly as `check-lint.yaml` runs it for this target:

```
$ cargo clippy -p mirrord-sip --all-targets --all-features --target aarch64-apple-darwin --keep-going --quiet -- "-Wclippy::indexing_slicing" -D warnings
$ cargo fmt --all -- --check
```

Both clean. No new dependencies, no changes to the crate's public API.

### DIFF
--- a/changelog.d/+sip-dyld-entitlement-value.fixed.md
+++ b/changelog.d/+sip-dyld-entitlement-value.fixed.md
@@ -1,0 +1,1 @@
+mirrord now SIP-patches a macOS binary that lists the `com.apple.security.cs.allow-dyld-environment-variables` entitlement with a `false` value, such as `/usr/bin/aa` on macOS 26. macOS strips `DYLD_INSERT_LIBRARIES` from such a binary, so the layer could not load into it.
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -566,17 +566,13 @@ mod main {
         false
     }
 
-    /// Key element that names the DYLD environment variables entitlement inside an entitlements
-    /// plist.
     const DYLD_ENTITLEMENT_KEY: &str =
         "<key>com.apple.security.cs.allow-dyld-environment-variables</key>";
 
-    /// Whether an entitlements plist grants the DYLD environment variables entitlement.
-    ///
-    /// An entitlements blob is an XML plist in which every entitlement carries a value, and Apple
-    /// ships binaries that list an entitlement turned off: on macOS 26 `/usr/bin/aa` carries this
-    /// one with a `false` value, and macOS still strips its DYLD variables. The key alone
-    /// therefore decides nothing, only the value element right after it does.
+    /// An entitlements plist lists each entitlement with a value, and Apple ships binaries that
+    /// list this one turned off: macOS 26 signs `/usr/bin/aa` with a `false` value, and macOS
+    /// still strips its DYLD variables. The key alone therefore decides nothing, only the value
+    /// element right after it does.
     fn grants_dyld_entitlement(entitlements: &str) -> bool {
         entitlements
             .split(DYLD_ENTITLEMENT_KEY)
--- a/mirrord/sip/src/lib.rs
+++ b/mirrord/sip/src/lib.rs
@@ -566,8 +566,32 @@ mod main {
         false
     }
 
-    /// Checks if the binary has the `com.apple.security.cs.allow-dyld-environment-variables`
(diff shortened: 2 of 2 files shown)


MaxFreedomPollard:
Still relevant and ready for review; the Greptile note was addressed in `ef03a9df5`.

aviramha:
Thank you for the contribution @MaxFreedomPollard ! Did you find such ocurrence in any system binary or is it just a theoretically possible thing?

MaxFreedomPollard:
Yes, it shows up in real system binaries. On macOS 26.2 (25C56), `/usr/bin/aa`, `/usr/bin/aea` and `/usr/bin/yaa` are all signed with the hardened runtime and list the entitlement as `false`:

```
$ codesign -d --entitlements :- /usr/bin/aa
...<key>com.apple.security.cs.allow-dyld-environment-variables</key><false/>...
```

`/usr/bin/aa` is also the binary `entitled_binary_is_not_sip` uses as its entitled example, so that test's premise no longer holds on macOS 26. On this machine, no binary in `/bin`, `/sbin`, `/usr/bin`, `/usr/sbin` or `/usr/libexec` lists the entitlement as `true`.

PR #4871, opened 2026-09-11
up: add 'mirror' mode and apply to traffic, queue splits

gememma:
- `mirrord up` allows mirror mode, where http traffic and queue messages are mirrored
- the default mode remains split mode
- the default http filter and queue split filters remain as-is (if the user sets no filter one is chosen automatically and used to select which traffic to mirror how to filter queues)


mirrord/up/src/config.rs
```diff
@@ -74,6 +78,28 @@ impl ValueEnum for ServiceMode {
...
+impl Into<IncomingMode> for ServiceMode {
+    fn into(self) -> IncomingMode {
+        match self {
+            ServiceMode::Split | ServiceMode::Replace => IncomingMode::Steal,
+            ServiceMode::Mirror => IncomingMode::Mirror,
+        }
+    }
+}
+
+#[allow(clippy::from_over_into)]
```
MintSoup, line 97: why not `From` tho

MintSoup (approved):

mirrord/up/src/config.rs
gememma: because it would have to be in the crate of the other types, which would mean importing these types, which is a circular dependency :((( nicer to have them here

PR #4646, opened 2026-07-31
Move from `serde_yaml` to `serde_saphyr`

gememma:
### Summary

`serde_yaml` is unmaintained, and `kube-rs` completed the same migration [in this PR](https://github.com/kube-rs/kube/pull/1975).

Notes:
- the mirrord-up and mirrord crates cannot be migrated (without a disproportionate amount of work) because they rely on **editing YAML in-place without changing the structure** - `serde_saphyr` does not support this, and `rust-yaml` does not yet support this
- the [`yaml_edit` crate](https://github.com/jelmer/yaml-edit) supports in-place YAML editing but has no license
- additionally, `yamlpatch::Op::MergeInto` takes `serde_yaml::Value` as an argument (for some reason) or, in newer versions, takes `yaml_serde::Value`
- `apple-codesign` still depends on `serde_yaml` and has not had a new release in 2 years
- `rust-yaml` has also been added as a dependency because it has an equivalent to `serde_yaml::Value` which is used in the `mirrord-vpn` crate - I don't love adding both crates because there is a lot of overlap, but I think this is OK

### `cargo tree`

After migration:
```
❯ cargo tree -i serde_yaml --depth 1 --target=all
serde_yaml v0.9.34+deprecated
├── apple-codesign v0.29.0
│   [dev-dependencies]
├── mirrord v3.242.0 (/.../mirrord/mirrord/cli)
├── mirrord-up v3.242.0 (/.../mirrord/mirrord/up)
└── yamlpatch v0.9.0
```

Before migration:
```
❯ cargo tree -i serde_yaml --depth 1 --target=all
serde_yaml v0.9.34+deprecated
├── apple-codesign v0.29.0
│   [dev-dependencies]
├── mirrord v3.242.0 (/.../mirrord/mirrord/cli)
├── mirrord-auth v3.242.0 (/.../mirrord/mirrord/auth)
├── mirrord-config v3.242.0 (/.../mirrord/mirrord/config)
├── mirrord-up v3.242.0 (/.../mirrord/mirrord/up)
├── mirrord-vpn v3.242.0 (/.../mirrord/mirrord/vpn)
└── yamlpatch v0.9.0
[dev-dependencies]
└── mirrord-operator v3.242.0 (/.../mirrord/mirrord/operator)
```

PR #4875, opened 2026-09-14
Remove more `serde_yaml` dependencies and uses

gememma:
Linear issues: [MBE-1715](https://linear.app/metalbear/issue/MBE-1715/check-kube-rs-yaml-serde-migration-and-upgrade-away-from-serde-yaml), [MBE-2005](https://linear.app/metalbear/issue/MBE-2005/completely-remove-serde-yaml-in-mirrord)

Previous PR: #4646 

- removed new dependencies on `serde_yaml` from tui, kafka consumer crates
- remove dependency from mirrord crate
- removed most mentions from mirrord-up crate, only some manual yaml editing remains
- added a ban to cargo deny config to prevent re-introducing deps


mirrord/cli/src/up.rs
```diff
@@ -252,7 +252,8 @@ mod tests {
 
     #[test]
     fn parse_error_buckets_as_config_validation() {
-        let parse_err: serde_yaml::Error = serde_yaml::from_str::<i32>("not a number").unwrap_err();
+        let parse_err: serde_saphyr::Error =
+            serde_saphyr::from_str::<i32>("not a number").unwrap_err();
```
meowjesty, line 256: That's not a number, (unsheathes pi) that's a number. 🐊

mirrord/up/src/init.rs
```diff
@@ -47,9 +46,13 @@ pub enum InitError {
...
 
-    /// Failed to serialize the assembled config to YAML.
+    /// Failed to deserialize the assembled config to YAML.
     #[error("failed to render config: {0}")]
-    Yaml(#[from] serde_yaml::Error),
+    YamlDeser(#[from] serde_yaml::Error),
+
+    /// Failed to serialize config YAML.
+    #[error("failed to render config: {0}")]
+    YamlSer(#[from] serde_saphyr::ser::Error),
```
meowjesty, line 55: Ser ou não ser, eis a questão.

tests/kafka-consumer/Cargo.toml
```diff
@@ -11,7 +11,7 @@ figment = { version = "0.10", features = ["env"] }
 rdkafka.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-serde_yaml.workspace = true
+rust-yaml.workspace = true
```
meowjesty, line 14: rust-yaml instead of saphyr?

meowjesty (approved):
Yamln't

tests/kafka-consumer/Cargo.toml
gememma: saphyr has no generic yaml `Value` ☠️ their docs say "just use `serde_json::Value`!" lol - see the last point of my rant in the PR description [here](https://github.com/metalbear-co/mirrord/pull/4646#issue-5030503695)

PR #4878, opened 2026-09-14
MBE-2092 Serverless agent api key

DmitryDodzin:
Exchange a MetalBear API key for the token sessions-manager verifies

Agents and intproxies authenticate to sessions-manager itself by trading the long-lived API key in MIRRORD_SESSIONS_MANAGER_API_KEY for a short-lived token from the MetalBear cloud, sent as an `Authorization: Bearer` credential on control-plane requests. The token is cached until a minute before its `exp`, which is read from the payload without verifying the signature: the client holds no public key and takes authenticity from TLS, so the claim only schedules the refresh.

This is independent of MIRRORD_SESSIONS_MANAGER_AUTH_TOKEN, which keeps meaning the secret expected by whatever fronts sessions-manager. Both may be set, and both headers then go out.

The cloud token is offered only to the control plane. The data-plane upgrade presents the single-use per-assignment credential under the same `authorization` name, so the per-plane headers keep it off that request.

### Quality Checklist:

  documentation)


changelog.d/+sessions-manager-cloud-token.internal.md
```diff
@@ -0,0 +1,9 @@
+Authenticate agents and intproxies to sessions-manager itself, by exchanging a
+long-lived MetalBear API key for a short-lived token and sending it as
+`Authorization: Bearer <token>` on control-plane requests. Set
```
itsamegraf, line 3: theyll need to be sent on all sessions-manager requests, as in, including data-plane.

changelog.d/+sessions-manager-cloud-token.internal.md
```diff
@@ -0,0 +1,9 @@
...
+`MIRRORD_SESSIONS_MANAGER_API_KEY` to enable it, and `MIRRORD_METALBEAR_CLOUD_URL`
+to point the exchange at somewhere other than `https://app.metalbear.com`. A token
+whose `scope` is not `serverless_agent` is logged as a warning but still used;
```
itsamegraf, line 6: once well have mirrord login in place, this warning will need to be removed as users will mint tokens which use the regular scope and access sessions manager through it.
either add note in code and/or comment in the relevant ticket in linear

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+///
+/// The exchange endpoint is unauthenticated apart from the key in the request body, so the key
+/// must never leave this process by any other route — neither it nor the tokens it yields are
+/// ever logged.
+///
+/// A token is fetched on first use and reused until [`TOKEN_REFRESH_MARGIN`] before its `exp`.
+/// The mutex is held across the exchange so that concurrent callers waiting on a cold or stale
+/// cache collapse into a single request rather than each starting one of their own.
+///
+/// Control plane only: the data-plane upgrade already sends the single-use per-assignment
```
itsamegraf, line 85: IMO the dataplane upgrade single use per assignment should not replace the cloudtokencredentials but come in addition to that, its meant to make sure only one consumer of that dataplane assignment can use it, once single-use token was consumed by succesful upgrade, it is revoked.

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+            return Err(SessionsManagerClientError::InvalidBaseUrlScheme(endpoint));
+        }
+        endpoint
+            .path_segments_mut()
+            .map_err(|_| SessionsManagerClientError::InvalidBaseUrl)?
+            .pop_if_empty()
+            .extend(TOKEN_EXCHANGE_SEGMENTS);
+
+        Ok(Self {
+            client: reqwest::Client::builder()
```
itsamegraf, line 128: all hail lamlam
**P1 — Following a 307/308 redirect can exfiltrate the long-lived API key**
   [`cloud_token.rs:134-139`](https://github.com/metalbear-co/mirrord/blob/8c05e502c3afaadab52720fd53b10b3eb9c6d1e4/mirrord/sessions-manager/client/src/credentials/cloud_token.rs#L134-L139), [`cloud_token.rs:174-179`](https://github.com/metalbear-co/mirrord/blob/8c05e502c3afaadab52720fd53b10b3eb9c6d1e4/mirrord/sessions-manager/client/src/credentials/cloud_token.rs#L174-L179)

   The exchange client uses reqwest’s default redirect policy. The API key is in the POST JSON body, and `307 Temporary Redirect` / `308 Permanent Redirect` preserve the method and body. Reqwest’s cross-origin protections for sensitive headers do not protect this body, so a redirect from the configured HTTPS exchange endpoint can forward `{"apiKey": ...}` to an arbitrary origin.

   This remains exploitable even after fixing the plain-HTTP configuration issue. Construct the exchange client with `redirect(reqwest::redirect::Policy::none())` (the fixed token endpoint should not require redirects), and test that a redirect response does not cause a second request.

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+        if claims.scope.as_deref() != Some(EXPECTED_TOKEN_SCOPE) {
+            tracing::warn!(
+                scope = ?claims.scope,
+                expected = EXPECTED_TOKEN_SCOPE,
+                "the MetalBear API key is not a serverless agent key, sessions-manager is \
+                 expected to reject the token it was exchanged for; check \
+                 {SESSIONS_MANAGER_API_KEY_ENV}"
+            );
+        }
+        let expires_at = SystemTime::UNIX_EPOCH + Duration::from_secs(claims.exp);
```
itsamegraf, line 196: me: we should be able to trust claims.exp validity but, just in case something goes wacky, I think we should cap it to something (like no more than a day or a week).
lamlam:
**P2 — A validly encoded JWT with an oversized `exp` crashes the long-running client**
   [`cloud_token.rs:211-220`](https://github.com/metalbear-co/mirrord/blob/8c05e502c3afaadab52720fd53b10b3eb9c6d1e4/mirrord/sessions-manager/client/src/credentials/cloud_token.rs#L211-L220)

   The token payload is intentionally parsed without signature verification, then its `u64` expiry is evaluated with:

   ```rust
   SystemTime::UNIX_EPOCH + Duration::from_secs(claims.exp)
   ```

   `SystemTime` addition panics when the resulting value is unrepresentable. A token containing `exp: 18446744073709551615` passes decoding and deserialization but terminates the agent/intproxy instead of returning a `TokenExchange` error. This also contradicts the comment that an untrusted payload can only cause an ill-timed refresh or warning.

   Use `SystemTime::UNIX_EPOCH.checked_add(...)` and map overflow to a non-retryable `SessionsManagerClientError::TokenExchange`. Add a regression test with an out-of-range `exp`.

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+
+        let mut header = HeaderValue::try_from(format!("Bearer {token}")).map_err(|_| {
+            SessionsManagerClientError::TokenExchange(
+                "token is not a valid header value".to_owned(),
```
itsamegraf, line 200: swap string based errors with specific variants.

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+                .unwrap_or(SystemTime::UNIX_EPOCH),
+        })
+    }
+}
+
+/// Reads [`TokenClaims`] out of a JWT payload *without verifying the signature*.
+///
+/// The client holds no public key and could not verify one; authenticity comes from TLS to the
+/// MetalBear cloud. The claims are read only to decide when to ask for a new token and whether to
+/// warn about the key's scope, so a payload this client mis-trusts costs at most an ill-timed
```
itsamegraf, line 227: lets discuss this, I was under the impression the minted tokens has proper utils in operator crate to verify them against public key, similar to how operator does for api keys. we cannot rely soly on tls :(

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
```diff
@@ -0,0 +1,588 @@
...
+            headers.insert(reqwest::header::AUTHORIZATION, self.authorization().await?);
+            Ok(headers)
+        }
+        .boxed()
+    }
+
+    /// Nothing: the data-plane upgrade authenticates with the single-use credential minted for
+    /// its assignment, which occupies this same header name.
+    fn data_plane_headers(&self) -> BoxFuture<'_, Result<HeaderMap, SessionsManagerClientError>> {
+        ready_headers(HeaderMap::new())
```
itsamegraf, line 267: note previous comments about it, single use credential minted should be data plane specific additional guard and not the unique one

itsamegraf (changes requested):

mirrord/sessions-manager/client/src/credentials/cloud_token.rs
itsamegraf: thinking back at this, we are on the client side. we dont need any verification or parsing here of the token - client side should just pass them to sm to perform the actual checks of the token.

PR #4880, opened 2026-09-14
MBE-2092 cloud baggage session

DmitryDodzin:
Route the token exchange to a local app-server for development

Testing the API-key exchange against a locally-run app-server needs the request to carry the baggage mirrord's steal filter matches on, or it reaches the deployed one instead. MIRRORD_METALBEAR_CLOUD_BAGGAGE_SESSION tags the exchange with `baggage: mirrord-session=<key>`, the operator's OPERATOR_CLOUD_BAGGAGE_SESSION for the same purpose. Unset in production, where no baggage header is sent at all.

The key is turned into a header value when the provider is built rather than on each exchange, so one that cannot be a header value is a config error at startup instead of a failure on every token the process ever asks for.

Only the exchange is tagged. Requests to sessions-manager itself are addressed by MIRRORD_SESSIONS_MANAGER_URL, which already points wherever a developer wants them.

### Quality Checklist:

  documentation)


itsamegraf (approved):
~~please reimplement it as a generic credential provider addon on top of other credential provider, well also need it for operator connection later~~
disregard, operator connection has its own impl of this. if needed well address it later

PR #4881, opened 2026-09-14
Add an Events tab to mirrord UI

skreborn:
Resolves INT-551.

The Events tab shows the operator's interception events for every session in the selected context at once.

There is no graceful fallback: for a context with an older operator (or one with no operator at all), the tab notes the incompatibility instead of showing an empty table.

An HTTP call is one row, not two. The operator emits the request and the response separately, and the response completes the row its request opened.

`--session-key-field` and `--unmatched` are off by default on `mirrord subscribe`, so existing output and any assertions on it are unaffected.


.github/workflows/check-lint.yaml
```diff
@@ -100,6 +100,17 @@ jobs:
       - run: pnpm --filter mirrord-ui run format:check
       - run: pnpm --filter mirrord-ui run typecheck
       - run: pnpm --filter mirrord-ui run build
+  ui-frontend-test:
```
gememma, line 103: huh, weird that this and `wizard-frontend-test` are in the `check-lint` workflow 🤔 i guess there isn't necessarily a nicer place for them 🤷

mirrord/cli/src/config.rs
```diff
@@ -1381,6 +1381,18 @@ pub(super) struct SubscribeArgs {
...
+    pub session_key_field: bool,
+
+    /// Also report queue messages that matched no session's filter, as `"mode": "filtered"`.
+    ///
+    /// These cover every session splitting the queue, including messages that went to a teammate's
+    /// session or straight to the workload. HTTP requests that matched nothing are never reported:
+    /// the agent passes those to the workload before the operator sees them.
+    #[arg(long)]
+    pub unmatched: bool,
+
```
gememma: why not flatten `EventStreamOptions` for these? is there a good reason to keep them separate?
```rust
#[clap(flatten)]
pub event_stream_options: EventStreamOptions,
```

mirrord/cli/src/subscribe.rs
```diff
@@ -37,36 +86,23 @@ pub(crate) async fn subscribe_command(args: SubscribeArgs) -> CliResult<()> {
...
-    .map_err(|error| CliError::SubscribeError(error.to_string()))?;
-
-    let stream = client
-        .request_stream(request)
-        .await
-        .map_err(|error| CliError::SubscribeError(error.to_string()))?;
+    let options = EventStreamOptions {
+        session_key_field: args.session_key_field,
+        unmatched: args.unmatched,
+    };
```
gememma: see https://github.com/metalbear-co/mirrord/pull/4881/changes#r4045415287 ☝️

mirrord/cli/src/subscribe.rs
```diff
@@ -17,10 +18,58 @@ use crate::{
     util::remove_proxy_env,
 };
 
-/// Streams interception events for a session key from the operator to stdout as JSON.
+/// What an event stream asks the operator to include beyond the default payload. An operator that
+/// predates an option ignores it.
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct EventStreamOptions {
+    /// Name the session on each event. Implied by a stream that named no key.
```
gememma: this comment confused me 😅 potentially a skill issue on my part
```suggestion
    /// Name the session on each event. Always `true` when a stream has no key.
```

mirrord/cli/src/subscribe.rs
```diff
@@ -17,10 +18,58 @@ use crate::{
...
+    let stream = client
+        .request_stream(request)
+        .await
+        .map_err(|error| CliError::SubscribeError(error.to_string()))?;
+
+    Ok(stream.lines().filter_map(|line| async move {
+        match line {
+            Ok(line) => {
+                let payload = line.strip_prefix("data:")?;
+                Some(Ok(payload.strip_prefix(' ').unwrap_or(payload).to_owned()))
```
gememma, line 75: as long as you want to trim all whitespace at the front and not just one space specifically-
```suggestion
                Some(Ok(line.strip_prefix("data:")?.trim_start().to_owned()))
```

mirrord/operator/src/crd.rs
```diff
@@ -784,6 +784,11 @@ pub enum NewOperatorFeature {
     /// targets at resolution time and would only report it as a failed session.
     PreviewCronJobTarget,
 
+    /// The interception event stream serves every session at once when given no key, honours
```
gememma: 🇺🇸 
```suggestion
    /// The interception event stream serves every session at once when given no key, honors
```

mirrord/cli/src/ui/server/v2.rs
```diff
@@ -182,6 +184,10 @@ struct OperatorStatusSummary {
     sessions: Vec<OperatorSessionSummary>,
     preview_sessions: Vec<OperatorPreviewSession>,
     license: OperatorLicense,
+    /// Whether this operator can name and pair the events the events view reads.
+    supports_subscribe_event_options: bool,
+    /// The operator's own version, so a view it can't serve can say what is installed.
+    version: String,
```
gememma: why do we need both? 🤔 surely the `supports_...` is enough to show the user, and they dont need to see the version?

mirrord/cli/src/ui/server/v2/events.rs
```diff
@@ -0,0 +1,169 @@
...
+    Query(query): Query<EventsQuery>,
+) -> Response {
+    let (tx, rx) = sse_channel();
+
+    tokio::spawn(forward(state, query.context, query.unmatched, tx));
+
+    sse_response(rx)
+}
+
+/// Reports reachability whenever it changes and forwards the operator's events, reopening the
```
gememma, line 91: why would it change? operator upgrade during a session?

mirrord/cli/src/ui/server/v2/events.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// Whether the operator can be read for a context, and whether it can serve this view.
+#[derive(PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Reachability {
+    available: bool,
+
+    /// Whether the operator advertises [`NewOperatorFeature::SubscribeEventOptions`].
+    ///
+    /// [`NewOperatorFeature::SubscribeEventOptions`]: mirrord_operator::crd::NewOperatorFeature::SubscribeEventOptions
```
gememma:
```suggestion
    /// Whether the operator advertises [`NewOperatorFeature::SubscribeEventOptions`](mirrord_operator::crd::NewOperatorFeature::SubscribeEventOptions).
```

mirrord/cli/src/ui/server/v2/events.rs
```diff
@@ -0,0 +1,169 @@
...
+async fn forward(state: AppState, context: Option<String>, unmatched: bool, tx: SseSender) {
+    let mut reported = None;
+
+    loop {
+        let status = Reachability::of(&fetch_operator(&state, context.as_deref()).await);
+        let supported = status.supported;
+
+        if reported.as_ref() != Some(&status) {
+            let event = sse::Event::default()
+                .event(STATUS_EVENT)
```
gememma, line 102: neater to inline this? up to you

gememma (approved):
only nits and comment suggestions 👍

mirrord/cli/src/config.rs
skreborn: No good reason, I'm afraid, just an oversight.

mirrord/cli/src/subscribe.rs
skreborn: :point_up_2:

mirrord/cli/src/subscribe.rs
skreborn: SSE strips exactly one optional space after the colon, and anything beyond it is part of the value, so that's entirely intentional. Is that a silly standard? Yes, but I'm not on the committee, so here we are.

.github/workflows/check-lint.yaml
skreborn: Odd indeed. Just trying to avoid duplication. Maybe it just needs a new name.

mirrord/cli/src/ui/server/v2/events.rs
skreborn: I'm exercising my rights to "up to me" and leaving it so. :sparkles:

mirrord/cli/src/ui/server/v2/events.rs
skreborn: Operator upgrade is one case, yes, but there's also cluster sleeps and wakes, generic network issues, and operator restarts.

mirrord/cli/src/ui/server/v2.rs
skreborn: So the idea here is that `supports_subscribe_event_options` decides, `version` explains. The concrete message is "This operator is 3.209.0. Showing events here needs 3.210.0 or newer." Without it you get "This operator is too old" with no way to tell whether the upgrade you just did took effect. Perhaps that's a bit too much hand-holding, but I'd keep it.

mirrord/cli/src/subscribe.rs
skreborn: Fair enough.

skreborn:
@gememma Apologies for the confusing multi-commit push after your review. I already started separating them before your review came in and didn't want to walk it back. Anything to follow up with, or am I good to merge?

mirrord/cli/src/ui/server/v2.rs
gememma: mm, that could definitely be condensed into an `Option<String>` 🦀 but that's fair enough - I guess the only thing is it could get out of hand if we keep adding features - we'd end up with a bunch of bool fields kicking around. Seems fine though 🙇

gememma (approved):
all good 🪑

PR #4883, opened 2026-09-14
Route subscribe and ui requests with the configured baggage

skreborn:
`baggage` from the config was applied to intercepted traffic but not to the CLI's own requests to the operator, so neither `mirrord subscribe` nor `mirrord ui` could reach an operator you are running locally behind a header filter on that key. Both now send it.


mirrord/cli/src/ui/server.rs
```diff
@@ -758,22 +761,39 @@ pub(super) async fn client_for_context(context: Option<&str>) -> UiResult<Client
...
+        .map_err(ApiError::from)
+}
+
+/// The baggage the daemon was started with: `mirrord ui -f <config>` resolves the config file's
+/// `baggage` into `MIRRORD_BAGGAGE` before spawning it, and setting the variable directly works
+/// too.
+fn baggage_from_env() -> Option<String> {
+    std::env::var("MIRRORD_BAGGAGE")
+        .ok()
+        .filter(|baggage| !baggage.is_empty())
```
meowjesty, line 796: Shame on people that don't use `.not()`

meowjesty (approved):
Deja vu

mirrord/cli/src/ui/server.rs
skreborn: I think `.not()`! :angry:

PR #4895, opened 2026-09-15
Fix mirrord up shutdown on user signals

meowjesty:
## Summary

- supervise all `mirrord up` services until shutdown instead of returning after the first child exits
- forward SIGINT, SIGTERM, and SIGHUP to Unix service process groups, allow graceful termination, and escalate after the shutdown timeout
- clean up descendants that outlive their direct child and preserve conventional forced-exit statuses
- treat deliberate wizard cancellation and user-ended sessions as successful outcomes instead of reporting them as failures
- add focused coverage for cancellation, signal forwarding, forced shutdown, and descendant cleanup

## Why

A user ending `mirrord up` could leave service processes running and could be recorded as a failed session. The command now owns service shutdown through completion, distinguishes deliberate cancellation from failures, and reliably cleans up the complete process tree.

## Clankerspeak

`mirrord up` now handles `ctrl-c` (and friends) a bit better, sending the correct telemetry "session end" to us.

There was a similar problem for the `mirrord up` prompt config thingy that we have, where if you cancelled what you were doing mid setup, it would send the wrong telemetry.

## Testing

- `cargo test -p mirrord-up process::tests` — 12 passed, 1 ignored subprocess entry point
- `samples/cor-1860-mirrord-up-user-ended-sessions/manage.nu test-sample` — passed against the staging cluster; both services handled SIGINT, stopped, and disappeared

### Quality Checklist

- [x] Documented the new supervision and signal-handling code
- [x] Tested successful user shutdown and process cleanup
- [x] Added focused unit and subprocess tests
- [x] Explained the behavior and motivation
- [x] User-facing website documentation is not required for this bug fix

Fixes COR-1860


mirrord/up/src/lib.rs
```diff
@@ -38,23 +35,28 @@ use mirrord_kube::{
...
 use crate::kube_context::UpKubeContext;
 
+/// Silences `deny(unused_crate_dependencies)`.
+///
+/// Windows test builds do not compile the Unix-only process tests that otherwise use tempfile.
+#[cfg(all(test, windows))]
+mod tempfile_marker {
+    use tempfile as _;
+}
+
```
MintSoup: can we not make this a dev-dependency on windows only? this doesnt look right

mirrord/up/src/process.rs
```diff
@@ -0,0 +1,814 @@
...
+
+use mirrord_progress::messages::SESSION_READY_MESSAGE;
+#[cfg(unix)]
+use nix::{
+    errno::Errno,
+    sys::signal::{Signal, killpg},
+    unistd::Pid,
+};
+#[cfg(windows)]
+use process_wrap::tokio::JobObject;
```
MintSoup, line 20: so this works on 🪟 too ?

mirrord/up/src/process.rs
```diff
@@ -0,0 +1,814 @@
...
+            self.child.try_wait()?;
+            group_exists = self.group_exists()?;
+            if group_exists {
+                tokio::time::sleep(Duration::from_millis(20)).await;
+            }
+        }
+        if group_exists {
+            self.kill_group()?;
+        }
+        self.child.wait().await?;
```
MintSoup, line 121: We expect this to finish immediately, right?

mirrord/up/src/process.rs
```diff
@@ -0,0 +1,814 @@
...
+    /// The grace period and received event cannot be forwarded as Windows
+    /// console events, so Job Object termination preserves forceful shutdown
+    /// while covering every descendant in the service tree.
+    #[cfg(windows)]
+    async fn stop(
+        &mut self,
+        _received_signal: Option<ShutdownSignal>,
+        _grace: Duration,
+    ) -> io::Result<()> {
+        Box::into_pin(self.child.kill()).await?;
```
MintSoup: why box?

mirrord/up/src/process.rs
```diff
@@ -0,0 +1,814 @@
...
+        .send(Ok(SignalDelivery { signal, accepted }))
+        .is_err()
+        || acceptance.await.is_err()
+    {
+        std::process::exit(signal.forced_exit_code());
+    }
+
+    if let Ok(signal) = receive_signal(&mut signals).await {
+        std::process::exit(signal.forced_exit_code());
+    }
```
MintSoup: So if user sends two `SIGINT`s in a row without waiting for the children to exit it kills everything ? Not sure what the correct thing would be from a UX perspective but I think this makes it too easy to accidentally to vaporize the session without graceful exit.

(not implying what we have here is wrong, just brainstorming)

mirrord/up/src/process.rs
```diff
@@ -0,0 +1,814 @@
...
+        received = signals.hangup.recv() => received.map(|_| ShutdownSignal::Hangup),
+    }
+    .ok_or_else(|| io::Error::other("shutdown signal stream closed"))
+}
+
+/// Windows console-event listeners installed before any service is spawned.
+///
+/// Ctrl-C and Ctrl-Break use distinct tokio listener types, so both must remain
+/// alive while [`receive_signal`] waits for whichever event arrives first.
+#[cfg(windows)]
```
MintSoup: I think it'd be cleaner to move the unix and windows code into their own submodules. Not sure if it's easily doable for all the code but maybe at least for signals it'd be ok?

MintSoup (changes requested):
looks good mostly

mirrord/up/src/process.rs
meowjesty: It should, but I haven't tested it.

mirrord/up/src/process.rs
meowjesty: - clanker says the following (because I don't remember anymore):

```
self.child.wait().await should remain. It reaps the direct child and lets process-wrap complete wrapper-specific process-tree bookkeeping.                                                                                                                                                                                                          

It should normally finish quickly after the group disappears or receives SIGKILL, but “immediately” should not be promised. Removing the wait risks zombies or incomplete cleanup.
```

mirrord/up/src/process.rs
meowjesty: I (vaguely) remember this has something to do with the `process-wrap` crate.

- clanker:

```
ChildWrapper::kill() returns an unpinned:                                                                                                                                                                                                                                                                                                           

 ```rust
   Box<dyn Future<Output = io::Result<()>> + Send + '_>
 ```

 The dynamic future is not guaranteed to implement Unpin, so it must be converted into a pinned box before awaiting. The boxed child itself is also expected because process-wrap returns different concrete wrapper types on Unix and Windows.
```

mirrord/up/src/process.rs
meowjesty: I think this is similar to the issue greptile was reporting on.

I'm not sure what to do either, I don't think we can keep "not quitting" if the user is pressing ctrl-c a bunch of times.

Clanker added a 2nd signal handler that vaporizes everything, cancelling the graceful exit.

mirrord/up/src/process.rs
MintSoup:
```rust
	fn kill(&mut self) -> Box<dyn Future<Output = Result<()>> + Send + '_> {
		Box::new(async {
			self.start_kill()?;
			self.wait().await?;
			Ok(())
		})
	}
```
huh, weird.
feel free to resolve ig

mirrord/up/src/process.rs
MintSoup: A quick smoke test would be nice, although IIRC we originally shipped `up` without testing on windows

mirrord/up/src/process.rs
MintSoup: I think a better UX would be to have a counter, with a status bar at the bottom. Like
`Keep pressing Ctrl-C to forcefully exit (1/3)` and have it increment every time until it reaches the top value.

Might be too much of a gimmick tho, not sure. If the clankers can do it quickly I'd say it's worth implementing, otherwise not.

MintSoup (approved):

mirrord/up/src/process.rs
meowjesty: A bit too gimmicky for this PR.

I'll open an issue for a follow-up, maybe we can come up with some status bar or something for `mirrord up`.

mirrord/up/src/process.rs
meowjesty: :no_smoking:

meowjesty: clanker said this will be mildly troublesome to add, do we have an issue for `mirrord up` tests for windows? I'll check and create one.

PR #4896, opened 2026-09-16
Stop release monitor paging on transient GitHub 5xx and resets

vladrbg:
### Quality Checklist:

- [ ] Documented new code sufficiently and updated existing comments and docs
- [ ] Written or updated user-facing website docs for new features, or opened a (sub)issue to do so
- [ ] Tested this change and know it succeeds and fails as expected
- [ ] Added tests
- [ ] Explained what this PR introduces and why, and linked to relevant context and added a changelog entry


giIuis (commented):
Did you consider tracking whether any retry returned a 404?
With --retry-all-errors, a sequence like 404 → 404 → 500 leaves code=500, so it could be classified as transient despite the earlier 404s.

Otherwise seems good :)

PR #4921, opened 2026-09-21
MBE-2080 Always hand off remote-layer incoming connections

itsamegraf:
This PR makes workload-companion remote-layer incoming traffic always enter the agent’s incoming-traffic pipeline, including connections accepted before a steal or mirror subscription exists.

#### Always accept remote-layer handoffs
- Removes the per-port subscription check from workload-companion connection-handoff negotiation.
- Always returns an accepted handoff with a localhost placeholder listener, allowing the remote layer to preserve the original connection’s passthrough path.
- Removes the remote-layer redirector’s shared subscription bookkeeping, since it no longer controls whether a remote-layer accept is handed off.

#### Handle connections before subscriptions
- Extends `PortRedirector` with a capability indicating that it can deliver connections before their destination port is subscribed.
- Makes the generic incoming redirector retain an empty port state for redirectors with that capability instead of dropping the connection.
- Routes these connections through HTTP detection even while no client is subscribed:
  - requests received before subscription continue to pass through to the target;
  - a later steal or mirror subscription can receive subsequent requests on the same HTTP/1 keep-alive or HTTP/2 connection.
- Keeps the normal redirector lifecycle intact by reconciling a later subscription through `add_redirection`.

#### Tests
- Updates handoff negotiation tests to expect all remote-layer handoffs to be accepted.
- Adds an HTTP/2 regression test covering a connection accepted before subscription, where:
  1. the first request is passed through; and
  2. a request after stealing begins is delivered to the stealer.

### Notes

This is part of the remote-layer incoming-connection work. The remote layer cannot know whether the agent has completed its per-port subscription setup when it accepts a connection. Accepting every handoff and deferring the subscription decision to the shared incoming pipeline avoids dropping that connection while preserving passthrough behavior until a subscription is active.

### Quality Checklist:

- [x] I have explained what this PR introduces and why, and linked to relevant context (e.g. linear issues, related PRs, documentation)


itsamegraf:
split off to https://github.com/metalbear-co/mirrord/pull/4867 (adjustments of workload companion) and https://github.com/metalbear-co/mirrord/pull/4922 (agent infra changes)

